Issue #25, opened 2019-09-23
Help Translate Appwrite Locale and Email templates


eldadfux:
Currently, Appwrite locale and email templates support 44 locales. We need help supporting more languages and improving current translations.

Translating is as easy as adding a few PHP file with new locale as prefix in file names.

## Update (#Hacktoberfest 20)
Please submit any PRs against branch `0.7.x` with the updated file structure, as shown below ⬇️

## Structure
List of file required for new locale are:

**Email Templates**
app/config/locale/translations/templates/en.email.auth.confirm.tpl
app/config/locale/translations/templates/en.email.auth.invitation.tpl
app/config/locale/translations/templates/en.email.auth.recovery.tpl

**Locale Lists and Settings**
app/config/locale/translations/en.php
app/config/locale/translations/en.continents.php
app/config/locale/translations/en.countries.php

**Configuration and Loading**
Add the new language to Appwrite config file: `/app/config/locale/codes.php` (order by ABC)
Load the new translation here: `./app/init.php (line: 224)` (order by ABC)

Make sure to keep PHP array keys exactly the same.

## Template Status

✅ = Updated - You can still submit PRs with fixes or improvements
⛔️ = Outdated - Update to match languages that are marked as updated
🆘 = Missing - No translation has yet been contributed (fallback to english)

New locales are always welcomed.

| Locale  | Phrases | Countries | Continents | Template - confirm | Template - invite | Template - recovery
|--|--|--|--|--|--|--|
| af | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| ar | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| bn | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| cat | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| cz | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| de | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| en | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| es | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| fi | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| fo | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| fr | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
[...]

PR #27, opened 2019-09-26
#25 Translate Locale and templates to Norwegian

Chaaarles:
Translated the files to Norwegian Bokmål


eldadfux:
@Chaaarles, Amazing work and you done it so quickly! Just updated the issue #25 . We also need this file if you have the time:

https://github.com/appwrite/appwrite/blob/master/app/config/locale/en.php

In your branch its probably a lot longer but we only need the list in the updated file as in the link above.

eldadfux:
Happy Hacktoberfest @Chaaarles! Changes merged. Please reach out on Gitter if you would like me to send you some cool Appwrite stickers and magnets! 🥳

PR #55, opened 2019-09-30
Fix code style in views

lex111:
cc @eldadfux


eldadfux:
Hey @lex111 thanks for the PR. Reach out on Gitter if you would like to get some Appwrite stickers and magnets!

PR #62, opened 2019-10-01
Use `echo` shortcut tag

lex111:
> echo also has a shortcut syntax, where you can immediately follow the opening tag with an equals sign. **Prior to PHP 5.4.0, this short syntax only works with the short_open_tag configuration setting enabled.**

According to my favorite PHP docs.


eldadfux:
🤔Actually, this is a conscious decision we made against the short_tags in order to be more consistent with the echo command as there are some place where some display logic is needed and its hard to avoid using echo.

And maybe also a bit of self bias against them 😅
Its not you @lex111 its the sort tags  😅

lex111:
A strange decision in fact, but okay. :open_mouth:

PR #67, opened 2019-10-01
Added Hindi-India Locale

mudit458:


eldadfux:
There is only 1 file in the PR 😬

eldadfux:
@mudit458 I have to close the PR because we already got a full Hinidi translations in another PR.

Please feel free to go over the repo issues to see if there is anything else you would like to help with.

Anyway, thank you for all the effort 🙏!

PR #76, opened 2019-10-01
added si.countries.php to support the country names in Sinhala language

IshankaDSenevirathne:


RizkyRajitha:
@eldadfux  i added other files nad made a pr just above this.

PR #86, opened 2019-10-02
updated phpunit.xml

SanyuktaSaha:
added extra code in it. The <filter> element and its children can be used to configure the whitelist for the code coverage reporting. The <php> element and its children can be used to configure PHP settings, constants, and global variables. It can also be used to prepend the include_path.


eldadfux:
@SanyuktaSaha sorry again, but this is not the kind of PR's that we need at the moment.

PR #95, opened 2019-10-02
Added Microsoft & Dropbox OAuth Support

christyjacob4:
1) Added Microsoft OAuth Support as per #5  
2) Added Dropbox OAuth Support as per #5


eldadfux:
Great as always! Thank you @christyjacob4 !

PR #109, opened 2019-10-03
introduce multi-stage build

mostafahussein:
This PR introduces multi-stage build which reduces the final size of the image and also makes use of the LTS version of Debian the slim edition.

Note that there is no vendor directory so that I have removed it from the Dockerfile.
If there are any modifications needed let me know.

Thanks


eldadfux:
This is a great feature I haven't found the time to get to, Thank you @mostafahussein ! ❤️

Only 2 little requests:
1. Please keep the ubuntu base image, as I am the only maintainer of this section of the project and I feel more comfortable with Ubuntu.
2. The vendor folder is generated by the PHP package manager in build.sh, please add it back or move the composer installation to the first stage of the build, whatever you like.

And thanks again!

mostafahussein:
Hell @eldadfux ,

1- The Debian image is not that different from ubuntu as Ubuntu is based on it, but the advantage here is to have a minimal image that contains the necessary packages only which is one of the reasons that we have saved about 100MB from the final image size and its more suitable for production, what do you think ? If you are okay with this, I can work on the other repositories

2- I have moved the composer step inside the Dockerfile.

Regarding being the only maintainer of this section of the project, I would be happy to be an additional maintainer for this section too :D

eldadfux:
Hey, @mostafahussein, thanks for adding composer to build stage, fantastic work!

Just checked, and the docker hub page for debian-slim image states that its not a stable release yet. I understand your motivation, but I still prefer to 'play it safe' and stay with the Ubuntu image. I am open to discussing this issue again in the future as the project progress, and we will have more focus on other things. I hope you understand.

I was also wondering if you could help us with another docker related issue we need for a different feature and include DBs and clustering. Let me know if that sounds interesting.

Please reach out to me on Gitter if you would like me to send you some cool Appwrite stickers and magnets!

mostafahussein:
D you mean this line in dockerhub docs about the slim editions?
> These tags are an experiment in providing a slimmer base
I understand your concerns, after revising the base image size it won't be a huge problem to move to ubuntu. I will work on moving to it then I will send a commit with it.

I will reach you on Gitter to discuss the other issues

Thanks!

mostafahussein:
@eldadfux, I have updated the Dockerfile, can you take a look when you have time? And let me know if you need anything else.

mostafahussein:
@eldadfux this PR is ready to be reviewed.

eldadfux:
@mostafahussein Haven't forgotten about this! Going to review it tomorrow. Just want to make sure it works well for dev mode.

eldadfux:
@mostafahussein as you suggested a few weeks ago, I moved all the volumes to use built-in docker volumes, this seems to solve a lot of permissions issues developers reported

tests/resources/docker-compose.yml
```diff
@@ -2,15 +2,19 @@ version: '3'
...
-    build: ../../
+    build:
+      context: ../../
+      args:
+        - TESTING=true
     restart: unless-stopped
     volumes:
     - appwrite-uploads:/storage/uploads:rw
     - appwrite-cache:/storage/cache:rw
+    - ./../../phpunit.xml:/usr/share/nginx/html/phpunit.xml
```
eldadfux: Any reason why we need the tests inside the container?

Dockerfile
```diff
@@ -1,66 +1,97 @@
...
+    DEBIAN_FRONTEND=noninteractive \
+    PHP_VERSION=7.3 \
+    PHP_REDIS_VERSION=3.1.2
+
+RUN \
+  apt-get update && \
+  apt-get install -y --no-install-recommends --no-install-suggests ca-certificates software-properties-common wget curl git openssl && \
+  LC_ALL=C.UTF-8 add-apt-repository -y ppa:ondrej/php && \
+  apt-get update && \
+  apt-get install -y --no-install-recommends --no-install-suggests make php$PHP_VERSION php$PHP_VERSION-dev zip unzip php$PHP_VERSION-zip && \
```
eldadfux: is the zip & unzip required because of --no-install-recommends?

Dockerfile
```diff
@@ -1,66 +1,97 @@
...
+  curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/bin --filename=composer
+
+WORKDIR /usr/local/src/
+# Updating PHP dependencies and auto-loading...
+ENV TESTING=$TESTING
+COPY composer.* /usr/local/src/
+RUN composer update --ignore-platform-reqs --optimize-autoloader \
+    --no-plugins --no-scripts --prefer-dist \
+    `if [ "$TESTING" != "true" ]; then echo "--no-dev"; fi`
+
```
eldadfux: We can skip this and just use production settings. All current tests are actually e2e so they don't have to be part of the container.

Dockerfile
```diff
@@ -1,66 +1,97 @@
+FROM ubuntu:18.04 AS builder
+
+LABEL maintainer="team@appwrite.io"
+
+ARG TESTING=false
```
eldadfux: We can skip this all tests will run from outside the container

Dockerfile
mostafahussein: Yes because in this case, we install what we need or what is actually required as a dependency.

Dockerfile
```diff
@@ -1,66 +1,97 @@
...
 
+COPY --from=builder /phpredis-3.1.2/modules/redis.so /usr/lib/php/20180731/
+
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
 RUN \
-  apt-get update --fix-missing && \
-  apt-get install -y software-properties-common && \
+  apt-get update && \
+  apt-get install -y --no-install-recommends --no-install-suggests wget curl ca-certificates software-properties-common openssl gnupg && \
```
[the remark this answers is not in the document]
mostafahussein: both for being able to install packages through apt as it require to accept keys or connect to https mirrors

eldadfux:
Hey @mostafahussein let me know if you have any more questions about my feedback we can also work on this together online on Discord. We really want to push this feature for the upcoming 0.4 release.

This will also help me and @itaymelamed, push the CI optimisation task much faster.

tests/resources/docker-compose.yml
mostafahussein: Because the environment outside docker is not prepared to run anything related to tests, so for example, all packages are installed inside the container itself so I decided to move the tests as well to avoid any repeated steps and so on

Dockerfile
```diff
@@ -1,66 +1,97 @@
...
+ARG TESTING=false
+
+ENV TZ=Asia/Tel_Aviv \
+    DEBIAN_FRONTEND=noninteractive \
+    PHP_VERSION=7.3 \
+    PHP_REDIS_VERSION=3.1.2
+
+RUN \
+  apt-get update && \
+  apt-get install -y --no-install-recommends --no-install-suggests ca-certificates software-properties-common wget curl git openssl && \
```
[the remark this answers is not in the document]
mostafahussein: `ca-certificates` is needed so you can handle https protocol because we are dealing with https links used by apt and wget.
for `git` I cannot remember the reason behind adding it but I assume that composer might need it if it's going to install packages through git itself

Dockerfile
mostafahussein: For this comment and other comments below that are about the usage of `TESTING=false/true`, all of these modifications related to the running tests inside the container.
If we are going to run them outside then we need to prepare the environment itself to run the tests

mostafahussein: Related comment: https://github.com/appwrite/appwrite/pull/109#discussion_r349878947

Dockerfile
mostafahussein: Related comments:
- https://github.com/appwrite/appwrite/pull/109#discussion_r349879353
- https://github.com/appwrite/appwrite/pull/109#discussion_r349879353

mostafahussein:
Sorry for the late follow-up, I have replied to all discussions.

eldadfux:
@mostafahussein thanks for the follow up, what should we do with the travis.yml? Should all tests run from the Appwrite container with a different entrypoint? Have you got a chance to make sure all tests still work that way?

mostafahussein:
If recall this correclty they tests was not passed until I have modified
the build.sh and adding TESTING parameter. But I have not run any travis
tests on my own. Do you suggest that I should run the travis test under my
own repository and check ?
Or do you have other scenario in mind ?

eldadfux:
I think that every time you update this PR the travis tests run again. If not you can integrate travis with your fork.

image

mostafahussein:
@eldadfux, can you please review the latest modifications? I have updated `travis.yml` file

.travis.yml
```diff
@@ -19,8 +19,7 @@ services:
 #   - docker_images
 
 before_install:
-- composer update --ignore-platform-reqs --optimize-autoloader --no-plugins --no-scripts
-- sudo service mysql stop
```
eldadfux: I originally added this because I had port conflicts, I am surprised it didn't break the build 🤔

eldadfux: Any idea what was I missing?

.travis.yml
```diff
@@ -19,8 +19,7 @@ services:
...
+- docker load -i docker_images/images.tar || true
```
eldadfux: I think this should enable caching in combination with this lines:

```
# cache:
#   directories:
#   - docker_images

before_install:
- docker load -i docker_images/images.tar || true

# before_cache:
# - docker save -o docker_images/images.tar $(docker images -a -q)
```
Any chance you tested it? @itaymelamed and I tried to use this to improve the build performance with no luck so far, but it seems your latests build with multi-stage support is now 3 minutes faster.

.travis.yml
mostafahussein: I am not sure but I expect to make appwrite container connect to the database through the docker network itself and not through the host network. Where I can find the database configuration related to the test case to check the host URL?

mostafahussein: Sorry I missed a point, the test cases will use the same database communication as defined in the application config itself which should be within the docker network that's why there is no need to check or disable the local mysql instance

eldadfux:
🥳

eldadfux:
@mostafahussein Thanks you so much for this amazing pull-request, its actually our biggest ever PR!

eldadfux:
I have also added a new build arg for our app versioning variable:
https://github.com/appwrite/appwrite/commit/05b51bd700923905066a802e037073a8855c2a02

eldadfux:
@mostafahussein we are talking about you over at Discord 😉
https://discord.gg/GSeTUeA you are welcome to join us 🙏

eldadfux:
@mostafahussein New build is now only 127MB thanks to this PR (24MB smaller!) 🥳

image

PR #121, opened 2019-10-04
added sinhala locale country sri lanka

RizkyRajitha:
added all the required files .


eldadfux:
Hey @RizkyRajitha, I really appreciate all your hard work, but it seems like we had duplicated PRs, and I have just merged another Sri Lanka locale earlier today.

If you would still like to help, it would be great if you could review the current locale and see if there is room for improvement and send a new PR.

PR #127, opened 2019-10-05
Add follow new releases badge

scopsy:
Added a badge so people can be notified when appwrite releases new versions and awesome features.


eldadfux:
Thank you ❤️

Let me know if you want some Hacktoberfest swags!

PR #135, opened 2019-10-05
Added Amazon & VK OAuth

christyjacob4:
1) Added Amazon OAuth as per #5  
However, it works with a small hack. The response returned by amazon is html encoded and it has to be html decoded for it to work correctly. This is not the case with any other provider. 

```php
// Uncomment this while testing amazon oAuth
// $state = html_entity_decode($state);
```


eldadfux:
I think I have an idea for solving the amazon issue without the risk of hurting other provider's implementation.

1. Add to abstract OAuth class 'appwrite/src/Auth/OAuth.php' a method like this:

```php
public function parseState($state) {
    return json_decode($state, true);
}
```

2. In the Amazon adapter, override this method, to handle the wrong way, they are passing the scope param (with your html_entity_decode solution).

3. In the controller itself, use the current adapter to parse the scope param. That way, we will ensure all adapters work correctly.

Don't love the solution, but considering the way Amazon is passing the data, it's pretty solid.

eldadfux:
* I haven't tested the code example.

Please mention in the code comments explaining why this implementation has been done.

christyjacob4:
Okay @eldad I'll implement it and let you know how it goes! Cheers!

eldadfux:
Great additions!

christyjacob4:
More to come!

PR #151, opened 2019-10-08
delete remove-orphans option

mostafahussein:
`--remove-orphans` is only needed when you have old containers e.g. when you renaming a service while its container is running then the container considered as an orphan that should be removed which might not the case as you start a new deployment

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ The easiest way to start running your Appwrite server is by running our docker c
 mkdir appwrite-ce && \
 cd appwrite-ce && \
 curl -o docker-compose.yml https://appwrite.io/docker-compose.yml?version=0.2.0 && \
-docker-compose up -d --remove-orphans
+docker-compose up -d
 ```
 
 


eldadfux:
Yeh, actually this option was really helpful when I just started development, probably can let it go.

 Just one thing to consider, in case we change the container structure how should we force developers that are updating their stack to stay only with relevant containers?

mostafahussein:
In that case we might change the documentation to have two ways/methods of
installation, you know something similar to how to add your code to a new
repository either by create filesh and push it or by adding an existing
project to your new repository.

So we will have a way for fresh installation which contains `mkdir` and
another way in case you have the project already and you want to update it
where you need to use `--remove-orphan` and you don't need to create a
directory unless you intend to have a separate installation (which will
more suitable to follow the first method)

eldadfux:
@mostafahussein changing the documentation in a way that will add more complexity/options to developers is an option I really want to avoid if possible.

I am trying to implement in this project a philosophy of treating the documentation and code the same way designers, and UX people look at the UI. Adding more stages, options, complexity, or just making the docs longer when we can avoid it will hurt developers' experience just as much as a bad UI will hurt app usability.

* All of the above doesn't suggest we have perfect documentation, but we do strive to make it outstanding by v1.0.0 is released. 😉

Let's think of different solutions that hides this complexity from the end developer.

PR #156, opened 2019-10-08
update

anandprabhakar0507:
http://localhost/${PORT} 

In case your port 80 is already taken, change the port number in the above command and in the URL also replace ${PORT} with listening port.


eldadfux:
Hey @anandprabhakar0507, we really appreciate your time and willingness to contribute to open-source, but this is not the kind of PR's that we need at the moment.

PR #158, opened 2019-10-08
Feature/node sdk file upload

kimuraz:


eldadfux:
Hey @kimuraz , really appreciate the effort, but I think you misunderstood the scope of this issue.

All of Appwrite SDKs are auto-generated by our SDK Generator, who has a separated repo: https://github.com/appwrite/sdk-generator.

This issue is meant for adding support to the NodeJS HTTP client template to allow it to send multipart/form requests while handling payload and attached files properly. I guess the best way to understand this requirement is to take a look at the PHP client template who is already implementing this feature.

We should also make sure that file upload is working, by generating the templates to a functioning SDK code and testing the 'createFile' method.

kimuraz:
Sorry for that, I'll take a further look at the generator... 

Although, if it does have any usage, the script I used to test it was:
```js
const sdk = require('../appwrite/app/sdks/node/index.js');                                                                                                                        
const fs = require('fs');
// Init SDK
let client = new sdk.Client();
client.setEndpoint('http://appwrite.test/v1');

let storage = new sdk.Storage(client);

client.setProject('').setKey('');


let promise = storage.createFile(fs.createReadStream('./img.jpg'), [], [], '1');

promise.then(function (response) {
  console.log(response);
  console.log(error.statusCode);
}, function (error) {
  //console.log(error);
  console.log(error.statusCode, error.message, 'ERR');
});
```

PR #172, opened 2019-10-12
adding translation to pt

Mukarillo:


Doges:
is that PT-BR or PT-PT?

eldadfux:
We already have pt-br and pt-pt

Mukarillo:
Oh, I see... Sorry about that

PR #185, opened 2019-10-13
fix: semantic errors in hindi locale

HellBender2011:
changed the content and it's placement, so that it would be semantically better.

### DIFF
--- a/app/config/locale/hi.php
+++ b/app/config/locale/hi.php
@@ -1,17 +1,17 @@
 <?php
 
 return [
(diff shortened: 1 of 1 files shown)


christyjacob4:
Hey @HellBender2011 I'm really grateful to you for having taken the time to contribute to our project. I went through your PR and found that it's not really a meaningful addition to what is already present. We strongly encourage you to take a look at our issues and pick up one that suits your interests or raise an issue for anything you are having trouble with. 
Thanks for your time!

PR #187, opened 2019-10-13
added stack exchange

nilmanichandrakar:


christyjacob4:
@eldadfux @nilmanichandrakar  This is not a functional adapter. You can see that the return value of getLoginURL does not return a valid page

test

PR #204, opened 2019-10-17
Format status.json so it's readable

jrcamelo:


eldadfux:
Hey @jrcamelo, we really appreciate your time and willingness to contribute to open-source, but this is not the kind of PR's that we need at the moment.

The status.json file is a cache result from an API call to status robot. Being readable has no value here as it not edited or read by anyone,

PR #213, opened 2019-10-20
Fix for typos in es translations

asperduti:
Update on es.continents.php. 'AN' with 'Antártida'(Related to  #25). It's the most used form.
The previous translation was 'Antártica' which is used in Chile. Sorry, but when I read it, I thought that it was a wrong translation, here in Argentina it's not used. I checked it, It's valid but it's not very common to use it in Latin America and it's not used in Spain. Only in Chile.

Refer: http://lema.rae.es/dpd/?key=Ant%C3%A1rtida

### DIFF
--- a/app/config/locale/es.continents.php
+++ b/app/config/locale/es.continents.php
@@ -2,7 +2,7 @@
 
 return [
     'AF' => 'África',
-    'AN' => 'Antártica',
+    'AN' => 'Antártida',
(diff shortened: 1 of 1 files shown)


eldadfux:
THANK YOU! Change merged 🥳

Please reach out to me on Discord or Gitter if you would like me to send you some cool Appwrite stickers and magnets!

PR #218, opened 2019-10-20
Add Thai template

Vinatk:


eldadfux:
Hey @Vinatk, please combine all Thai pull requests into one request with all commits, so we will be able to review it properly.

PR #232, opened 2019-10-25
Changed OAuth paths and some route titles

eldadfux:
Updated the path of our OAuth login route to allow better API predictability when we introduce new login methods in the future.

New login method mights include:

/auth/login/sms
/auth/login/passwordless || /auth/login/email || /auth/login/otp
/auth/login/qrcode

etc...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,7 +17,7 @@
 
 ## Breaking Changes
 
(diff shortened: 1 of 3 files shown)


christyjacob4:
@eldadfux this sounds great!

PR #242, opened 2019-10-28
Remove executable bit from non-executable files

friederbluemle:
A bunch of non-executable files in the repo erroneously had the executable bit set. This fixes is by changing the file mode from `755` to `644`.


friederbluemle:
Rebased onto latest default branch.

eldadfux:
Hey @friederbluemle, can you please explain the change in the image files? How can I validate other and future files we might add to the project?

friederbluemle:
Hi @eldadfux - I'm not sure what caused the image files to have the executable bit set, but one possibility is an extracted .zip archive file on macOS. There seems to be a bug with the default extracting tool on macOS that sets +x on *all* files.

I am not aware of an "easy" way to validate, except to check the correct file permissions for each newly added file to the repo (the executable bit should _only_ be set for executable shell scripts). Unfortunately it seems the GitHub UI only displays file permissions _after_ the file has been added to the repo:

image

eldadfux:
Thank you for the info and for improving Appwrite @friederbluemle!

I don't remember if you already signed up to get some Hacktoberfest swags, but if not, please contact me on Discord! 

And we will be thrilled if you will join us on our Discord server, you'll be a great addition to our community!

PR #253, opened 2019-11-04
Test enhancement

peter279k:
# Changed log
- Removing unused white spaces.
- Adding native type hints for every method arguments and return type hints.
- Using the `assertCount` to assert expected count is same as result.
- The PHPUnit fixtures are `protected setUp(): void` and `protected tearDown(): void`.


eldadfux:
Thanks again @peter279k! If you haven’t signed up for swags yet, please contCt me on Discord.

PR #256, opened 2019-11-18
Added multistage building

itaymelamed:


eldadfux:
Seems the composer command isn't available on the default PHP docker image.

Take a look at this:
https://hub.docker.com/_/composer

PR #259, opened 2019-11-20
Update sdks.php

PedroCisnerosSantana:
Fixing license for python lang due to classifiers in order to upload to pypi


eldadfux:
Great work pedro!

Issue #261, opened 2019-11-23
Custom OAuth Scopes


eldadfux:
Currently, Appwrite's OAuth endpoint is requesting our predefined OAuth scopes from each service. This predefines scopes are usually used for getting the user name and email for allowing us to verify the user identity and successfully completing its registration or login process.

After completing the user authentication, we save the user OAuth token on his user account to allow app developers to interact with the user's OAuth provider API.

If a user has logged with facebook as his OAuth provider, this means we can allow the app developer to send Facebook API requests to get more of the users' data. Of course, this entire process is subjected to user consent.

Currently, Appwrite API is limited to only a few predefined scopes for each OAuth provider. Allowing custom scopes will allow app developers to make requests to OAuth providers with extra permissions on behalf of the user.

To allow custom scopes, we need to introduce a new parameter to the OAuth endpoint (https://appwrite.io/docs/auth#oauth). The new parameter, called 'scopes', will accept an array of strings and will be empty and not required by default.

After accepting and validating the new parameter, we should merge its data with our existing list of scopes. Our current list of scopes is located on each OAuth provider class. An example can be viewed in our Facebook adapter on line 32:
https://github.com/appwrite/appwrite/blob/master/src/Auth/OAuth/Facebook.php#L32

To enforce Appwrite maintainers to manage scopes in an organized and controlled method, we should add two new methods to our OAuth provider's abstract class:
https://github.com/appwrite/appwrite/blob/master/src/Auth/OAuth.php

1. addScope(string $scope)
2. getScopes():array

After adding those two new methods, we should update each OAuth provider to actually use them for parsing the scopes when sending his OAuth request.

eldadfux:
We have made big progress on this thanks to @christyjacob4 fantastic work on this during the weekend. This feature should be ready for the next version release (0.5).

@christyjacob4 do you think you could write a small post about how to use this feature that we could publish on our Medium blog?

christyjacob4:
Yea I can write the blog post. Let me know by when you require it.
I also need some details regarding the same.

CodeWithKamal:
link to blog post if you have one, I need to use Facebook Marketing API and need to custom scopes with Facebook OAuth.

PR #273, opened 2020-01-12
Custom OAuth Scopes

christyjacob4:
Added support for providing custom scopes as per #261


src/Auth/OAuth.php
```diff
@@ -24,20 +24,27 @@ abstract class OAuth
...
      * OAuth constructor.
      *
      * @param string $appId
      * @param string $appSecret
      * @param string $callback
      * @param array  $state
+     * @param array $userScopes
      */
-    public function __construct(string $appId, string $appSecret, string $callback, $state = [])
+    public function __construct(string $appId, string $appSecret, string $callback, $state = [], $userScopes = [])
```
eldadfux: When passing scopes directly to the protected $userScopes attribute we skip the duplication check. Maybe we can use the addScope internally from inside the object constructor.

Also I think that in the context of this class the 'userScopes' variable can be named only 'scopes'.

src/Auth/OAuth.php
```diff
@@ -78,14 +85,33 @@ public function __construct(string $appId, string $appSecret, string $callback,
...
+    /**
+     * @param $scope
+     * 
+     * @return array
+     */
+    protected function addScope(string $scope){
+        // Add a scope to the scopes array if it isn't already present 
+        if (!in_array($scope, $this->userScopes)){
+            $this->userScopes[] = $scope; 
+        }
```
eldadfux: We can use here:
```
return $this;
```

to allow chaining.

eldadfux: Also need to change docblock return value

src/Auth/OAuth/Bitbucket.php
```diff
@@ -27,10 +32,20 @@ public function getName(): string
...
+        foreach ($this->requiredScope as $item) {
+            $this->addScope($item);
+        }
+        // var_dump($this->getScopes());
+        // exit();
+
+        return 'https://bitbucket.org/site/oauth2/authorize?' .http_build_query([
+                'response_type' => 'code',
+                'client_id' => $this->appID,
+                'scope' => implode(' ', $this->getScopes()),
```
eldadfux: perfect 👌🏻

app/controllers/auth.php
```diff
@@ -433,8 +435,8 @@ function ($provider, $success, $failure) use ($response, $request, $project) {
     ->desc('OAuth Callback')
     ->label('error', __DIR__.'/../views/general/error.phtml')
     ->label('scope', 'auth')
-    ->label('abuse-limit', 50)
-    ->label('abuse-key', 'ip:{ip}')
+    // ->label('abuse-limit', 100)
```
eldadfux: You can completely remove this 2 commented lines, no need for abuse check here as the route only do a redirect anyway.

app/controllers/auth.php
```diff
@@ -452,8 +454,8 @@ function ($projectId, $provider, $code, $state) use ($response, $request, $domai
     ->label('error', __DIR__.'/../views/general/error.phtml')
     ->label('webhook', 'auth.oauth')
     ->label('scope', 'auth')
-    ->label('abuse-limit', 50)
-    ->label('abuse-key', 'ip:{ip}')
+    // ->label('abuse-limit', 100)
```
eldadfux: This 2 lines can be uncommented, they should work and the abuse check here actually is important

src/Auth/OAuth.php
christyjacob4: could you please elaborate this? I dont think I understood

src/Auth/OAuth.php
christyjacob4: could you please elaborate this? I dont think I understood

eldadfux: instead of this assignment:
```
$this->userScopes = $userScopes;
```

we should have something like this:
```
foreach($scopes as $scope) {
    $this->addScope($scope);
}
```

src/Auth/OAuth.php
eldadfux: when adding this statement:
```
return $this;
```

we can use this function this way (chaining):
```
$oauth
    ->addScope('x')
    ->addScope('x')
;
```

its very common to do this in PHP set and add methods

app/controllers/auth.php
```diff
@@ -395,13 +396,14 @@ function ($email, $password, $success, $failure) use ($response, $request, $proj
     ->label('sdk.description', '/docs/references/auth/login-oauth.md')
     ->label('sdk.location', true)
     ->label('sdk.cookies', true)
-    ->label('abuse-limit', 50)
-    ->label('abuse-key', 'ip:{ip}')
+    // ->label('abuse-limit', 100)
```
eldadfux: Please note we should uncomment and keep lines 399-400
Please note we should uncomment and keep lines 457-458

Lines: 438-439 can be deleted

src/Auth/OAuth.php
christyjacob4: okay noted.

src/Auth/OAuth.php
christyjacob4: Noted and done.

PR #274, opened 2020-01-12
Custom OAuth Scopes

christyjacob4:
Added support for Custom OAuth Scopes as per #261


src/Auth/OAuth/Bitbucket.php
```diff
@@ -14,6 +14,11 @@ class Bitbucket extends OAuth
      */
     protected $user = [];
 
+    /**
+     * @var array
+     */
+    protected $requiredScope = [];
+
```
eldadfux: Instead of having a new attribute '$requiredScope' we can overwrite '$userScope'.

This will save you the writing of this loop in every adapter:
```
foreach ($this->requiredScope as $item) {
           $this->addScope($item);
 }
```

For example:

    /**
     * @var array
     */
    protected $userScope = ['email', 'job_title'];

eldadfux: You can also name the abstract attribute 'userScope' with just **'scope'**

christyjacob4: can you elaborate how this works?

christyjacob4: Also, I have renamed the variable in OAuth from userScopes to scopes

eldadfux:
Merged 🎉🎉🎉🎉

PR #278, opened 2020-01-15
Add twitch oauth provider

armino-dev:
In ref to #5 
I finished the twitch.tv OAuth provider integration 😅.


eldadfux:
@armino-dev great work! Thank you so much! Please let me know if there are any other provider or issues you would like to work on!

armino-dev:
I was thinking on going for spotify

PR #283, opened 2020-01-17
hide scrollbar if not needed in menu

TorstenDittmann:
scrollbar


eldadfux:
Thank you @TorstenDittmann !

In case you want to see the changes take place in your setup, you can just use:
```bash
npm install
gulp less
```

PR #284, opened 2020-01-17
Added Yahoo OAuth provider

armino-dev:
In ref to #5
I managed to get some time to finish the Yahoo provider 😄, so here it is.


eldadfux:
❤️🎉

PR #289, opened 2020-01-20
Added PayPal OAuth provider

armino-dev:
In ref to #5 
PayPal OAuth is ready, so here it is.


src/Auth/OAuth/Paypal.php
```diff
@@ -0,0 +1,164 @@
...
+        'sandbox' => 'https://www.sandbox.paypal.com/',
+        'live' => 'https://www.paypal.com/',
+    ];
+
+    private $resourceEndpoint = [
+        'sandbox' => 'https://api.sandbox.paypal.com/v1/',
+        'live' => 'https://api.paypal.com/v1/',
+    ];
+
+    private $environment = 'sandbox';
```
eldadfux: As we currently don’t have any support for multiple environments we should probably have the live one as default.

armino-dev: Changed environment to live. Shall I remove entirely the sandbox?

eldadfux:
👏🏻

PR #290, opened 2020-01-21
Added Bitly OAuth provider

armino-dev:
In ref to #5 
Hi guys, I know it wasn't on the list, but here it is the Bitly OAuth provider.


eldadfux:
Great work as always!

Issue #303, opened 2020-03-25
Some API parameters are not in camelCase 


bartektartanus:
Currently in API there are 3 methods that uses order type parameter. Two are named `orderType` and one is `order-type` which is inconsistent. 
`GET/storage/files` has `orderType`
`GET/database/collections/{collectionId}/documents` has `order-type`
I think this should be fixed to `orderType` as most of the parameters are written in camelCase.
Other parameters are: `order-field`, `order-cast`, `old-password`, `password-a`, `password-b`.

eldadfux:
This is a really bad API design issue and we'll push a fix for version 0.6. This will be introduced as an API breaking change. It should not affect users of Appwrite SDKs, but new versions of all SDKs will be pushed anyway.

eldadfux:
This is fixed now, will be released in next version

PR #304, opened 2020-03-25
Updated all the REST API query params to be in camelCase

eldadfux:
As reported in this issue: #303


public/scripts/dependencies/appwrite.js
```diff
@@ -665,11 +665,11 @@
                 }
 
                 if(passwordA) {
```
bartektartanus: Maybe these variables also should be updated to `password` and `passwordAgain`?

eldadfux: 👌🏻

eldadfux: Fixed

Issue #373, opened 2020-06-02
UI improvement


AnatoleLucet:
So first of all; in my opinion the ui feels nice (it's the main reason why I've decided to learn Appwrite instead of Parse, Kuzzle, and more... (yes, uis makes a big diff for me 🙃)), it always feels like the product is finely tuned when the user experience a good ui. But I believe some improvement can be made.

Watch-out the animations feel a bit slower on the gif then it really is.

## Modals animations
- [ ] Done
### Before:
![](https://i.gyazo.com/03c9c6b78f0a1f84a89d17d2df2cd595.gif)
### After:
![](https://i.gyazo.com/42c2920e518e10e697ce56be8c90c2fd.gif)

## Buttons animations
- [ ] Done
### Before:
![](https://i.gyazo.com/ede6a93074995e41d1eff0f32c8d240b.gif)
### After:
Unfortunately we can't really see this one on the gif. 😕
But it's clearly smoother.
![](https://i.gyazo.com/b6a3b7d3394244b9ee71ddba9b5376e2.gif)

## Go-back link animations
- [x] Done  (#392)
### Before:
![](https://i.gyazo.com/db5c8274faf228c3eccff4276649d764.gif)
### After:
![](https://i.gyazo.com/902accadcf02f1a130da7dac751a03f3.gif)

## Nav links animations
- [x] Done (#397)
### Before:
![](https://i.gyazo.com/e2e8b546e1fa4e54b5067d418c29fddb.gif)
### After:
Same here, we can't really see the diff, but I simply removed the weight change and added a transition.
![](https://i.gyazo.com/fd0bf189da72d6b9fde72e4ed4f4b3a0.gif)

## Links animations
- [x] Done (#392)
### Before:
![](https://i.gyazo.com/49f7c4c6107e00898840a5249157fa74.gif)
### After:
![](https://i.gyazo.com/97c13c3d4287fff7d4198e75d0f5dff8.gif)

## Toggle between light and dark theme transition
- [ ] Done

I haven't tried anything yet, but it could be nice to have a transition when switching theme.

## Fix the scroll when menu is open
- [ ] Done

For now when the menu is open, you can still scroll the page in the background (see the scroll bar on the right). I don't think this is wanted.

![](https://i.gyazo.com/46c5d96d1b540f4bc538b9cae8e8ae0f.gif)

## The icon in the `add` button isn't PERFECTLY centered
- [x] Done (#396)
[...]

PR #392, opened 2020-06-09
Adding links animation

AnatoleLucet:
This PR adds the links animations (see #373).

### 1. Bouncy links
In #373 I've proposed color changing links, but I found out that bouncy animations on this kind of link looks nice. I've created a class to remove this animation if needed (`.link-animation-disabled`).


eldadfux:
@AnatoleLucet The stylesheet should support future plans to add RTL language translations. To be able for this changes to support this, we need to change naming conventions from left & right to start & end.

For example: `link-return-animation--right` should be `link-return-animation--end`

You can also use or configure new env vars in default-ltr.less and default-rtl.less

AnatoleLucet:
@eldadfux 
>  we need to change naming conventions from left & right to start & end.

Sure, I'll do.

> You can also use or configure new env vars in default-ltr.less and default-rtl.less

Shall I move `.link-return-animation()` in `default-ltr.less` or `default-rtl.less`? I guess `ltr` means *left to right* (same for `rtl`), but I don't understand what they're used for.

eldadfux:
You can take a look at both default-ltr.less (left-to-right) and default-rtl.less (right-to-left) they hold config vars that helps us use left and right values as params instead of explicitly referring right or left attribute (we use end and start instead) and switch between them when changing between languages.

In RTL design right will be the start and left will be the end. With LTR design left will be considered as the start and right as the end

AnatoleLucet:
Oh, well thank you about the explanation. I've never worked with LTR languages before.

But by what I understand, `link-return-animation--right` and `link-return-animation--left` doesn't change the animation to support LTR languages. `link-return-animation--right` makes the arrow move to the right on hover, and `link-return-animation--left` makes the arrow move to the left on hover.

For example, this link is using `link-return-animation--left`:

and this one `link-return-animation--right`:

Because the arrow direction&position doesn't change between LTR and RTL languages, I don't think they should belong in `default-ltr.less`&`default-rtl.less` (same for the `start` and `end`, the animation direction doesn't change on LTR languages).

eldadfux:
@AnatoleLucet this is how Appwrite look when changing to RTL mode (no translations have yet been implemented, so its still in English):
image

eldadfux:
All the arrows are being switched to the other direction..

eldadfux:
You can test how things work on RTL. by manually changing the CSS path in your devtools: 
image

AnatoleLucet:
Great! My bad then, I'm going to work on that.

PR #401, opened 2020-06-11
Database key fix

eldadfux:
Fixed an issue where database rules keys have been limited to 40 chars instead of 32. The console UI was also missing proper validation.


eldadfux:
Documents list UI has also been limited to 10 cells as performance was starting to degrade when tested with a lot of cells (45). Figured 10 should be more than enough for document preview.

PR #411, opened 2020-06-13
[Feature] Mail background worker

TorstenDittmann:
To speed up endpoints in which an email is sent, this PR adds a background worker which allows asynchronous sending of emails.


eldadfux:
Great work!! 🥳

PR #422, opened 2020-06-17
add container_name to all of the compose containers

drandell:
All of the containers the compose file creates should be named the same regardless of where the docker-compose file is ran from. 

i.e running this currently from a folder called appwrite defines the container name for redis as;
```
appwrite_redis_1 
```
This change will now make the redis container name;

I can see the main appwrite container already has this but the rest do not. 
```
appwrite_redis
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,6 +3,7 @@ version: '3'
 services:
   traefik:
     image: traefik:v2.1.4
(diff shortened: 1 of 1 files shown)


eldadfux:
All changes merged! 🥳 @drandell, congratz on your first Appwrite merged PR!

PR #452, opened 2020-06-26
Preloader

eldadfux:
Still need to add an option to disable preloading in dev mode and create a production-like compose file for CI testing.


eldadfux:
The Preloader feature is already merged to 0.7.0 as part of the Swoole integration.

Issue #410, opened 2020-06-13
[HELP-WANTED] Adding Support for More OAuth Providers


eldadfux:
New issue to replace #5 . Help us add new OAuth providers for Appwrite authentication. Adding a new OAuth provider is pretty simple and can be a great way to start adding contributions to the project. We have created a step by step guide to add a new adapter, and you can ask for support here or on our Discord server #maintainers channel (https://discord.gg/GSeTUeA).

Tutorial available here:
https://github.com/appwrite/appwrite/blob/master/docs/tutorials/add-oauth2-provider.md

- [x] Amazon
- [x] Apple
- [x] Bitbucket
- [x] Bitly
- [x] Discord
- [x] Dropbox
- [x] Facebook
- [x] GitHub
- [x] GitLab
- [x] Google
- [x] Linkedin
- [x] Microsoft
- [x] PayPal
- [x] SalesForce
- [x] Slack
- [x] Spotify
- [x] Twitch
- [x] VK
- [x] Yahoo
- [x] Yandex
- [ ] Baidu
- [x] Box
- [ ] Evernote
- [ ] Fitbit - **in progress (@NidhxGupta)**
- [x] PayPal Sandbox (Should extend the current PayPal provider...)
- [x] Wordpress (@rsneh)
- [ ] Weibo
- [x] Yammer
- [ ] Mail.ru
- [ ] Reddit (Reddit has some technical difficulties since their API doesn't return an email ID or a UUID or a full name, which we require) 
- [ ] Twitter (Not supporting OAuth2 yet)
- [ ] Instagram (Instagram has some technical difficulties since their API doesn't return an email ID, which we require) 
- [ ] Steam
- [ ] Zoho
- [x] Stripe
- [ ] Imgur
- [ ] Disqus
- [ ] Netflix
- [ ] Telegram

komemi:
@eldadfux is someone assigned to wordpress api?

eldadfux:
@komemi you are :) congratz 😅

eldadfux:
Box OAuth adapter has been merged! Great work @armino-dev #467

armino-dev:
Thanks @eldadfux, you guys are moving fast. I see a lot of improvements every day. 😄

eldadfux:
@armino-dev a lot thanks to amazing contributors like you ❤️

eldadfux:
I have added a new 'name' attributes in the providers config file to have better representation of providers names inside the Appwrite console.

[...]

PR #469, opened 2020-07-06
Extending Paypal OAuth2 provider to Paypal Sandbox

armino-dev:
Here is a small PR for #410 to extend support for Paypal Sandbox.


eldadfux:
Merged 🎊

PR #473, opened 2020-07-09
Corrected filename from DomianTest.php to DomainTest.php

armino-dev:
I've made a sync with the upstream and noticed some typo.
So here is a quick fix to correct the filename from DomianTest.php to DomainTest.php ;)


eldadfux:
Thanks!

PR #536, opened 2020-09-13
Feat new preview options

eldadfux:
- Changed naming conventions to be similar to CSS conventions.
- Changed all image properties methods to start with set* for consistency.
- `setImageOpacity` method is deprecated, used an alternative method.
- No need for try -> catch to hide error, if something is wrong we better know about it.
- Added default black color to the borderColor param
- Added strict typing to PHP Resize lib

TODO:

- [ ] Add unit test to all new image options.
- [ ] Add new e2e tests to all the new API endpoints params.
- [ ] `roundCorners` is not available on latest imagemagick version. Find an alternative.


app/controllers/api/storage.php
```diff
@@ -257,7 +262,7 @@
         $fileLogos = Config::getParam('storage-logos');
 
         $date = \date('D, d M Y H:i:s', \time() + (60 * 60 * 24 * 45)).' GMT';  // 45 days cache
-        $key = \md5($fileId.$width.$height.$quality.$background.$storage.$output);
+        $key = \md5(rand(0, 10000000).$fileId.$width.$height.$borderColor.$borderWidth.$opacity.$rotation.$borderRadius.$quality.$background.$storage.$output);
```
christyjacob4, line 265: You might want to remove `rand(0, 10000000)`  in the final version

eldadfux:
Closed in favor of #1008

PR #573, opened 2020-09-30
Updated dependencies, compressed some imgs, and README edit.

Mou-Ikkai:
An assortment of changes, including updating your REALLY outdated dependencies. Some were even 1.x behind!


Mou-Ikkai:
I'll try to fix this if... my change is at fault.

Mou-Ikkai:
Found it. Latest ubuntu version breaks something.

Mou-Ikkai:
Right... PHP likes changing things each version so you have to rewrite code...

eldadfux:
@Mou-Ikkai please note that some of your changes are already applied to our latest in-development version located at branch 0.7.x.

You should work against that branch and not the master.

Also, please don't manipulate any images inside the `./tests` directory as it might break the tests and has no performance advantage.

Mou-Ikkai:
Will do.

PR #583, opened 2020-10-01
build.sh: Quote this to prevent word splitting

Yash-Garg:


PedroCisnerosSantana:
Hello @Yash-Garg 

In this case It's not needed, if you check the find command you will find out we are checking if the file was modified in the last 6 months.

PedroCisnerosSantana (changes requested):
unnecessary quotes

PR #610, opened 2020-10-01
Hey, I added in a head tag to make your html look more professional.

Rishi-coder11:


Rishi-coder11:
Yea now that I think about it... Fixed it though!

eldadfux:
@Rishi-coder11 thanks for submitting this PR, but we don't need a head tag in this file, as it isn't viewable by browsers and only uses for bypassing CORS restrictions. Please check other open issues if you would like to submit another PR.

PR #636, opened 2020-10-02
Gave si.php more context

manula2004:
the Sinhalese translation did not make a lot of sense. being a Sinhalese myself, I changed a few words in order to make sure it sounds right


spielers:
@manula2004 I would You should make 1 PR with all these minor changes that will help us to review it correctly,
You can have look at the description of the issue for more 

best,
spielers

PR #637, opened 2020-10-02
Update si.continents.php

manula2004:
In Sinhalese, the bare translation of Oceania sounds incomplete. Therefore I added the proper word to give it some sense


spielers:
@manula2004  Thanks for your contribution. I would suggest you make PR to 0.7.x branch rather than master.
the translation work currently going on there.

best,
spielers

manula2004:
@spielers  I changed the branch to 0.7x as you told. Thanks for the notice

PR #639, opened 2020-10-02
Edited the wrong translations

manula2004:
In Sinhalese, some of these country names were actually different than the way we address them . Therefore I added the proper letters and terms to make them sound normal


spielers:
hey, @manula2004  Thanks for your contributions I would suggest you make PR to the 0.7.x branch because we are currently working on that branch regarding translation.

best,
spielers

spielers:
@manula2004  You  have a conflict please resolve this 
Thanks

PR #646, opened 2020-10-04
Updated Dockerfile

SoftCreatR:
- Updated PHP-Redis to 5.3.1
- Moved PHP-API-Version and PHP-Redis-Version to separate variables for easier maintenance
- Removed manual nginx compilation in favor of [nginx-ee](https://github.com/VirtuBox/nginx-ee), which does all the hard work for us (Compile nginx + brotli, etc.)

Update: Looks like something broke. I'll perform my changes step by step to find the issue and fix it. If that doesn't work, I'll close the PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ ARG TESTING=false
 ENV TZ=Asia/Tel_Aviv \
     DEBIAN_FRONTEND=noninteractive \
     PHP_VERSION=7.4 \
-    PHP_REDIS_VERSION=5.2.1
+    PHP_REDIS_VERSION=5.3.1
 
 RUN \
   apt-get update && \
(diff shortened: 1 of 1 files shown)


eldadfux:
@SoftCreatR this is an outdated version of our Dockerfile 🙏🏻 please check branch `0.7.x` and check if their's anything to improve their.

SoftCreatR:
> @SoftCreatR this is an outdated version of our Dockerfile 🙏🏻 please check branch `0.7.x` and check if their's anything to improve their.

Explains a lot 😸

eldadfux:
+ we have some problems compiling php8 + swoole on our Dockerfile in the branch named `php8`. Seems like tests are failing randomly when the HTTP server stop to respond, would be great if you could look into it 🙏🏻

SoftCreatR:
> * we have some problems compiling php8 + swoole on our Dockerfile in the branch named `php8`. Seems like tests are failing randomly when the HTTP server stop to respond, would be great if you could look into it 🙏🏻

That also seems to happen here.

eldadfux:
This is the first time I see this error. Might be related to changes in the PR?

SoftCreatR:
> This is the first time I see this error. Might be related to changes in the PR?

That's somewhat unlikely because all other tests were successful. However, I'll keep trying (using the 0.7.x branch).

PR #647, opened 2020-10-04
Fix grammar.

jbampton:
Changed `it's` to `its`.


spielers:
Hey, @jbampton Thanks or your contribution but This is not the proper way of fixing grammar, According to a newer update of hacktoberfest that would be counted against spam. I would suggest you have a look at our open issues if you want to contribute

Thanks,
spielers

PR #652, opened 2020-10-05
Update my-first-issue.md

abhaypawar:
@spielers 
Hey, Its my pleasure to go for my first translation. Check it out.
I would continue translating important docs further. Thanks!


spielers:
Hey, @abhaypawar Thanks for your contribution but we don't need the translation of these files

The translation is only for Email locale templates #25
best,
spielers

PR #654, opened 2020-10-05
Update oauth.md

abhaypawar:
oauth.md translated. The reviewers @spielers check it.


spielers:
Hey, @abhaypawar  Thanks but we dont need the translation of oauth.md files

Translation is only for Email locale templates #25 
best,
spielers

PR #655, opened 2020-10-05
Update register.md

abhaypawar:
register.md translated. @spielers reviewers please check.
Thanks again!


spielers:
Hey, @abhaypawar Thanks but we don't need the translation of register.md files

The translation is only for Email locale templates #25
best,
spielers

Issue #511, opened 2020-08-23
Support For Verifying JWT Token using the Server API


ereio:
I can extract the token from the x-fallback-cookie in headers for the client, but there's no way to validate these access tokens in the node sdk. If you allow the node-appwrite getSessions query to include each sessions access token, one could authenticate custom endpoints through our own backends and still use the same authentication platform through appwrite. 

A common pattern I use when developing backends is to create a proxy graphql server as an abstraction for every service available. When doing this, I don't want to roll my own authentication each time as it's a liability, but I want to abstract it in case I need to switch auth providers in the future. Firebase allows you to verify tokens using their node sdk and thus check the token being passed by the user in the headers.

Let me know if this makes sense, if you want further clarification, or if I'm missing something here.

Thanks for all your hard work!

eldadfux:
@ereio Would a JWT implementation (https://jwt.io/) in the Appwrite users/account APIs help you accomplish this use case?

ereio:
yes, that would be great

eldadfux:
Added this feature request to our roadmap

punitdiwan:
I always wanted to have this JWT auth in addition to cookie based auth. Its way easier to manage the JWT token from Local Storage. I am excited to learn that it is being implemented. Moreover, its equally secure as cookie based auth.

eldadfux:
JWT draft PR is ready: https://github.com/appwrite/appwrite/pull/784/files 🎉

eldadfux:
JWT is available in a closed beta, if anyone wants to try it please reach me on Discord.

TorstenDittmann:
This got fixed with https://github.com/appwrite/appwrite/releases/tag/0.8.0 👏🏻

jdub113113:
Help me

jdub113113:
Jdizzle11383@gmail.com

Issue #509, opened 2020-08-19
[FEATURE] Add MQTT communication 


raul-parada:
Communication with IoT devices using MQTT will provide an additional attention to use Appwrite.

eldadfux:
Hey Raul, this sounds extremely interesting. I would love to get some other use cases that you might have in mind and how would you see that kind of feature interacting with Appwrite's APIs (auth, storage, database).

Appwrite already has plans to create more network based protocol for interacting with stuff like GraphQL, GRPC and Protobuf. MQTT sounds interesting, but we could definitely use more community feedback to  better understand how this could be implemented and work well for Appwrite developers.

raul-parada:
Yes, it is a powerful feature. One use case is the connected car: collect, process and display data directly from the on-board unit of a vehicle.

codeagencybe:
Another general use case is to use mqtt or similar like Kafka as a message broker queue. 
Basically it can catch workloads and process them at a later time and return back. 

Typically use cases can be eg request reports that require big data/processing and you don't want to another process to keep waiting for it. So instead you shove it in a queue and return it at a later time. Banking eg that generate financial reports. You can request a report from a dashboard and you receive it a later time by email   
This prevents overloading the system by having too many parallel jobs running. 

Other uses cases: big import / export tasks

eldadfux:
@codeagencybe I'm finding it a little bit difficult to understand how you view this as part of the current Appwrite architecture. Currently we are using Redis to achieve queueing or pub/sub capabilities.

I think cloud functions feature can be an easier solution to the use cases you mentioned. Would be great if you could visualize this over the current architecture diagram, to help me better understand this:
https://github.com/appwrite/appwrite/blob/master/docs/specs/overview.drawio.svg

![image](https://user-images.githubusercontent.com/1297371/90654943-a2339180-e249-11ea-8327-c3e2fc5c40cd.png)

brandonroberts:
Thanks @raul-parada. Our functions have come a long way since this conversation. Let us know if this is still relevant and feel free to reopen if needed.

Issue #265, opened 2019-12-11
Add realtime capabilities 


FOUADFAISAL:
Dear ,
is this product support real time database and backend as a service ?
and  is service provides application developers an API that allows application data to be synchronised across clients and stored on appwrite's cloud ?
and when you think it will release android version ?
thank you to reply me.

eldadfux:
You can use Appwrite database to sync data between clients on different devices. You can learn more here: https://appwrite.io/docs/database. Appwrite currently only offers support for self-hosted setups, we get a lot of requests for a hosted cloud service, and it's something we are definitely aiming for.

Real time support is not something we have built-in yet. We are planing to add this feature in the future, but for now you can use any custom websocket server and integrate it with the Appwrite API or webhooks events to get this functionality working and customized exactly for your needs. I'll be happy to help on this if it something you think you'll want to try.

We have ongoing work on both Java and Kotlin SDKs. You can follow the progress on this issues:

https://github.com/appwrite/sdk-generator/issues/24
https://github.com/appwrite/sdk-generator/issues/5

In the meantime you could integrate directly to Appwrite HTTP API with any native HTTP client.

eldadfux:
I created a simple flowchart to describe how the realtime server will work with other Appwrite components: https://github.com/appwrite/appwrite/blob/feat-265-realtime-server/docs/specs/realtime.drawio.svg

The spec is aimed to allow the setup to have a cluster of realtime server behind Appwrite loadbalancer.

Any feedback will be great.

eldadfux:
I have created a new branch with a POC websocket sever using our already in-place Swoole extension that supports both Websockets and MQTT protocols out of the box. You can see the small POC code here: https://github.com/appwrite/appwrite/compare/0.7.x...feat-265-realtime-support

As we can't add any custom headers to the native Websocket client on browsers, and 3rd party cookies don't seem to pass with requests, we might be required to add JWT support for connection authorization. (#511) 

Feedback is welcomed.

eldadfux:
JWT draft PR is ready: https://github.com/appwrite/appwrite/pull/784/files 🎉

hbakhtiyor:
https://github.com/resgateio/resgate
https://github.com/dunglas/mercure

designlook:
Would love to get realtime support in Appwrite. I tried resgate but seems hard to have it running along with Appwrite. Using docker versions of both. Port conflicts and seems hard to get it working together with Nginx.

[...]

PR #692, opened 2020-10-20
Feat 265 *platform wide* realtime support

eldadfux:
## What does this PR do?

Adding realtime server for allowing live updates and interactions with the Appwrite platform.

The server will include multiple realtime protocols support including (in order of priority):
- [x] Websocket - **in progress**
- [ ] MQTT
- [ ] Socket.io
- [ ] SSE

This PR is still a work in progress. Stuff still yet to be implemented:
- [ ] Abuse control check on connection start (X connections per Y hours, X connection per user)
- [ ] Limit payload size
- [ ] Defining message structure: { status: "ok"|"error", event: EVENT_NAME, data: <any arbitrary data> }
- [ ] JWT Authentication (passed in path / or in message) **related to #511 **
- [ ] List of channels
- [x] Channels subscriptions (per resource / group of resources (collections / files))
- [ ] Logout on current session delete event
- [ ] Integrate messaging worker for event triggering
- [ ] Add e2e tests
- [ ] Update installation script compose file

## Test Plan

TODO: add new set of tests to both test connection establishment and stream functionality. Trigger mock event, assert payload is correct on each listener.

## Related PRs and Issues

#511 
#265 
#509


app/controllers/api/health.php
```diff
@@ -30,6 +30,19 @@
         $response->json(['version' => APP_VERSION_STABLE]);
     }, ['response']);
 
+App::get('/v1/health/realtime')
```
eldadfux, line 33: This is only used for testing and will be deleted later.

TorstenDittmann:
May I ask where the need for multiple real-time protocols comes from?

eldadfux:
@TorstenDittmann not much of a need, but more of a possibility. Seems to be pretty easy to support multiple protocols using Swoole.

This also have a very good fit with Appwrite agenda of being tech and platform agonistic, and might allow new use-cases. There were a few mentions in the community about MQTT and Websocket, I guess that will allow easier integration with Appwrite for platform that are already comfortable with a specific protocol or client SDK.

TorstenDittmann:
I see nothing explicitly wrong with this code for now, since it is a starting point anyway. 

The plan is that all possible functions that are normally requested via HTTP are also accessible via the realtime protocol, right?

Also, have you figured out a plan to manage connections and subscriptions yet? 

Amazing work! 👍

eldadfux:
> The plan is that all possible functions that are normally requested via HTTP are also accessible via the realtime protocol, right?

@TorstenDittmann I'm still thinking about it, not sure how the message request will look like, maybe something similar to GraphQL or GraphQL itself? I was thinking of starting with a read-only mode where users can subscribe and receive events and use it in combination with the REST API. What do you think?

> Also, have you figured out a plan to manage connections and subscriptions yet?

This is what I'm thinking about. I'm gonna use a hashmap of all the users where the key is going to be the connected user ID and the value is going to be a hashmap of channel names and the user connection ID, like this: `user_id` => [channels => ['channel1' => true, 'channel2' => true], connection => x].

Every new event message will do a hash lookup, which should be pretty fast, find all relevant users, check if they have subscribed to the messages, and then send the message if relevant (if the user has subscribed for it).

The performance here should be O(N), where N is the number of "read" permissions in a resource [file/document/collection] - (@TorstenDittmann let me know if I'm missing something). Maybe we should add a limit to the max number of permissions?

On each connection establishment, we'll verify the user has privileges to access that channel. We'll remove the connection ID on every connection close event or on specific failure errors.

For channel logic, I had this in mind: (but I'd love some feedback)

| Channel | Description |
|---|---|
| account | All account related events (session create, name update...) |
| collections | Any update/delete/create events for collections where user has read access |
| collections.[ID] | Any update/delete/create events to a given collection where user has read access |
| collections.[ID].documents | Any update/delete/create events to any document in a given collection where user has read access |
| documents | Any update/delete/create events to documents where user has read access |
| documents.[ID] | Any update/delete/create events to a given document where user has read access |
| files | Any update/delete/create events to file where user has read access |
| files.[ID] | Any update/delete/create events to a given file where user has read access |

joshdvir:
@eldadfux I think that adding realtime feature to Appwrite will be a great addition, just a few things to keep in mind:

1. MQTT is not really a realtime protocol, it's a lightweight messaging protocol I'm not sure what use case are you trying to add here with it.
2. SSE only allows 1 way communication, only the server is sending messages to the client which can be limiting, I prefer Websockets by far.
3. I like the approach you took when you wanted to add a new feature you added a new component to complement that feature, I think this should also apply here, Scaling Websockets is different than scaling HTTP services so I think the component should be a new one that can be scaled independently and not part of the main application.
Moreover in the past I used https://centrifugal.github.io/centrifugo/ as a websockets server with much success and adding this type of component will save a lot of development time since the main application would just need to "manage" it and define the protocol to use.
Centrifugo also support SockJS which is like Socket.io.

Let me know what you think

Josh

eldadfux:
Thanks for the feedback @joshdvir!

About the points you raised:
1. MQTT can be used to communicate between Appwrite and a developer custom backend located on the same network or datacenter. Also one of our main agendas is to be tech and platform agnostic in order to allow max customisation and flexibility with Appwrite.

2. Actually that can be good enough for a lot of use cases where you just need a listener. Even with Websocket will probably start by supporting on way communication and on phase 2 we'll add 2-way communication, but that hasn't been decided yet.

3. Completely agree. The realtime server is going to be an independent & stateless Docker container which will be super easy to both scale and debug. We're using Swoole for the Websocket server mainly because we already use it for HTTP (with great performance) and it allow us to easily reuse a lot of our codebase.

eldadfux:
@TorstenDittmann are we OK to close this PR in favor of #948 ?

TorstenDittmann:
> @TorstenDittmann are we OK to close this PR in favor of #948 ?

Yes, we can close this one.

Issue #516, opened 2020-08-29
Docker based hot-reload solution for dev environment 


eldadfux:
From version 0.7 the Appwrite containers requires a reload  for any changes to get applied. This is due to the fact that most of the codebase is loaded into memory for performance boosting.

We need help setting up a hot reload solution to our code when making changes during development.

From my short research this looks like a nice idea for a Docker based solution:
https://github.com/apogiatzis/docker-compose-livereloader
https://medium.com/better-programming/live-reloading-with-docker-compose-for-efficient-development-356d50e91e39

Any similar Docker based solution that gets the job done and is language agnostic can be considered for this issue.

lightlace:
I'd like to tackle this issue.

eldadfux:
Please do 🙏🏻

eldadfux:
Hey @lightlace are you still working on this issue?

Glide7:
Hey @lightlace , are you still working on this?

eldadfux:
If anyone wants to help with this issue please let us know.

KShivendu:
Hey @eldadfux, I would like to give it a try. Can you assign it to me :)

Emiliaaah:
I'd like to take on this issue. @eldadfux feel free to assign me to this issue.

Emiliaaah:
Kind of messed up the first one since I thought I did something wrong, but this should be good.

PR #695, opened 2020-10-20
Feat 516 Add code hot-reload functionality for docker

katsadim:
## What does this PR do?

Adds code hot-reload functionality.

Took advantage of [docker-compose-livereloader](https://github.com/apogiatzis/docker-compose-livereloader) project where it allows live reloading of code for efficient development.

In order to make it work the following command should be run:

```bash
docker-compose -f docker-compose.yml -f docker-compose-reload.yml up
```

## Test Plan

I have only added the public directory to be reloaded every time there is a change. I tested it and it works fine. This is a work in progress and you should advice which other directory I should add for hot reloading (i guess `/app` is a good candidate).

The reload delay is currently set to 20 seconds but it is configurable.

## Related PRs and Issues

#516 

Yes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,6 +74,12 @@ docker-compose up -d
 
 After finishing the installation process, you can start writing and editing code. To compile new CSS and JS distribution files, use 'less' and 'build' tasks using gulp as a task manager.
 
+Note that code hot-reloading is supported. Therefore instead of `docker-compose up -d`, you may run the following:
+
+```bash
+docker-compose -f docker-compose.yml -f docker-compose-reload.yml up
+``` 
+
 ## Architecture
 
 Appwrite's current structure is a combination of both [Monolithic](https://en.wikipedia.org/wiki/Monolithic_application) and [Microservice](https://en.wikipedia.org/wiki/Microservices) architectures, but our final goal, as we grow, is to be using only microservices.
--- a/docker-compose-reload.yml
+++ b/docker-compose-reload.yml
@@ -1,0 +1,19 @@
+version: '3'
+
+services:
+
+  appwrite:
+    container_name: appwrite
+
+  live-reloader:
+    image: apogiatzis/livereloading
+    container_name: livereloader
+    privileged: true
+      environment:
+        - RELOAD_CONTAINER=appwrite
+        - RELOAD_DELAY=20              # seconds
+        - RESTART_TIMEOUT=10         # seconds
+        - RELOAD_DIR=/usr/share/nginx/html/public
+      volumes:
+            - "/var/run/docker.sock:/var/run/docker.sock"
+


eldadfux:
Sorry for the slow response, I labeled the PR as `hacktoberfest-accepted`, I will try to fully review as soon as possible!

CONTRIBUTING.md
```diff
@@ -74,6 +74,12 @@ docker-compose up -d
 
 After finishing the installation process, you can start writing and editing code. To compile new CSS and JS distribution files, use 'less' and 'build' tasks using gulp as a task manager.
 
+Note that code hot-reloading is supported. Therefore instead of `docker-compose up -d`, you may run the following:
```
eldadfux, line 77: Hot-reloading shouldn't be optional for maintainers, let's just add it to the main Appwrite container as a POC. If it will work well, we can work to make it suitable for all other important containers which we spend time working on.

Remember that the docker-compose.yml in this repo is only used by Appwrite maintainers, it's not the one used when developer install Appwrite for building their apps.

docker-compose-reload.yml
```diff
@@ -0,0 +1,19 @@
...
+
+  appwrite:
+    container_name: appwrite
+
+  live-reloader:
+    image: apogiatzis/livereloading
+    container_name: livereloader
+    privileged: true
+      environment:
+        - RELOAD_CONTAINER=appwrite
```
eldadfux, line 13: I think the the delay here should be much lower, waiting 20 seconds for container to reload after a change can be forever. Maybe 1 second to start with?

docker-compose-reload.yml
```diff
@@ -0,0 +1,19 @@
...
+        - RELOAD_DELAY=20              # seconds
+        - RESTART_TIMEOUT=10         # seconds
+        - RELOAD_DIR=/usr/share/nginx/html/public
+      volumes:
+            - "/var/run/docker.sock:/var/run/docker.sock"
```
eldadfux, line 18: Looking at the `apogiatzis/livereloading` docs, it seems we also need to mount the dirs we're watching to this container.

docker-compose-reload.yml
```diff
@@ -0,0 +1,19 @@
...
+
+  live-reloader:
+    image: apogiatzis/livereloading
+    container_name: livereloader
+    privileged: true
+      environment:
+        - RELOAD_CONTAINER=appwrite
+        - RELOAD_DELAY=20              # seconds
+        - RESTART_TIMEOUT=10         # seconds
+        - RELOAD_DIR=/usr/share/nginx/html/public
```
eldadfux, line 16: Let's change the target branch for this change from master to `0.7.x`, we had lots of architecture changes between them, and hot-reloading will be much more suitable for 0.7 version.

After that, this are the 3 main libraries we need to listen for changes that will reload the main `appwrite` container:
- ./app:/usr/src/code/app
- ./docs:/usr/src/code/docs
- ./src:/usr/src/code/src

Reading the docs, multiple DIRS can be supplied comma separated.

eldadfux (changes requested):
Finally got some time for a proper review. First of all, Great work! I have added some minor changes that will help to start enjoying this feature :) Let me know if you need any help!

Issue #475, opened 2020-07-12
[HELP-WANTED] Replace our QR code generator library


eldadfux:
We are currently using: https://github.com/Bacon/BaconQrCode for generating QR codes. This library is great but introduces a lot of 3rd party dependencies we would like to reduce, and don't seemed to be actively maintained anymore. We want to replace it with this lib: https://github.com/chillerlan/php-qrcode which seems a lot lighter and more active.

PedroCisnerosSantana:
Hello @eldadfux I think this issue could be good for me to make some real php code :P 
Could you please assign it to me?

Thanks

eldadfux:
Sure @PedroCisnerosSantana that will be very helpful. Please work on the 0.7 branch, and let me know if you have any question :)

PedroCisnerosSantana:
Thanks @eldadfux I'd love to be your young padawan :P

eldadfux:
@PedroCisnerosSantana Got any chance to work on this? let me know if you need help

PedroCisnerosSantana:
@eldadfux will contact you in private later

anuragagarwal561994:
Is this issue being worked upon, or up for grab? I was thinking to take this up if possible.

PedroCisnerosSantana:
Hello @anuragagarwal561994 I'm working on it, if you want to help you can reach me out on our discord server :D

ealvarez968:
@PedroCisnerosSantana whats your discord server I would like to help

eldadfux:
Hey @PedroCisnerosSantana, any updates? Would you like me to re-assign this?

PedroCisnerosSantana:
Hello @eldadfux I think it's completed, I just need your help to test it out :D

eldadfux:
@PedroCisnerosSantana No problem, submit a PR and lets check it out together.

PedroCisnerosSantana:
Hello @ealvarez968 the link is on the README, but here you have it https://discord.gg/GSeTUeA :)

PedroCisnerosSantana:
@eldadfux As soon as I arrive home will make a PR, do you want me to make the PR to a specific branch?

eldadfux:
@PedroCisnerosSantana awesome work on this one!! this fix is now available with the release of Appwrite 0.7! 🚀

PR #700, opened 2020-10-22
Changes BaconQRCode for QRCode

PedroCisnerosSantana:
## What does this PR do?

It changes BaconQRCode for QRCode, much more lighter 😉 

## Test Plan

This code might have errors, I need @eldadfux to review and test it ❤️ 

## Related PRs and Issues

#475 

 Yes


composer.json
```diff
@@ -50,7 +50,8 @@
         "domnikl/statsd": "3.0.*",
         "influxdb/influxdb-php": "1.15.*",
         "bacon/bacon-qr-code": "2.0.2",
```
eldadfux, line 52: We need to remove this dependency `"bacon/bacon-qr-code": "2.0.2",`.

composer.json
```diff
@@ -50,7 +50,8 @@
...
-        "phpmailer/phpmailer": "6.1.7"
+        "phpmailer/phpmailer": "6.1.7",
+        "chillerlan/php-qrcode": "^4.1"
```
eldadfux, line 54: 4.2 seem to be the latest version.

app/controllers/api/avatars.php
```diff
@@ -11,12 +11,15 @@
 use Utopia\Cache\Adapter\Filesystem;
```
eldadfux, line 11: Please remove all comments from the code, we can always use other branches or tags to revert.

app/controllers/api/avatars.php
```diff
@@ -366,12 +369,17 @@
...
+       /*  $renderer = new ImageRenderer(
             new RendererStyle($size, $margin),
             new ImagickImageBackEnd('png', 100)
-        );
+        ); */
+        $qropts = new QROptions([
+            'quietzone'         => $size
+        ]);
+        $qrcode = new QRCode($qropts);
+        $qrcode->render($text);
```
eldadfux, line 380: This line is not doing anything: `$qrcode->render($text);` just return the result output to no var. What you should do is use the `->send` method in the bottom of the route to attach the image data to our response payload. Example:

```php   
        $response
            ->addHeader('Expires', \date('D, d M Y H:i:s', \time() + (60 * 60 * 24 * 45)).' GMT') // 45 days cache
            ->setContentType('image/png')
            ->send($qrcode->render($text))
        ;
```

eldadfux (changes requested):

composer.json
PedroCisnerosSantana: Yes, forgot to apply the change on the VM env, sorry

eldadfux (approved):

Issue #474, opened 2020-07-11
[HELP-WANTED] Use Maxmind native C extension


eldadfux:
https://github.com/maxmind/libmaxminddb -> Maxmind native C extension for PHP should introduce improved performance. 

This change should be merged to the `swoole` [branch](https://github.com/appwrite/appwrite/tree/swoole) which contains the most up-to-date versions of our Dockerfiles. This changes should be made for both the [Nginx Dockerfile](https://github.com/appwrite/appwrite/blob/swoole/Dockerfile.nginx) and the [Swoole Dockerfile](https://github.com/appwrite/appwrite/blob/swoole/Dockerfile.swoole) (including the [default Dockerfile](https://github.com/appwrite/appwrite/blob/swoole/Dockerfile) in that branch).

The C extension should replace the [geoip2/geoip2](https://github.com/appwrite/appwrite/blob/swoole/composer.json#L47) package.

Would be great to get some benchmark results for this change.

snehal199:
Can I pick this up @eldadfux ?

eldadfux:
@snehal199 yes sure. Please submit the PR against branch `0.7.x`.

adi-mohan:
Hi! Is this issue still open? Can you give me some guidance on how to do this. Most of the links provided here are broken. I saw geoip2 listed in composer.json. Am still a beginner in open-source. Thanks!

hack3r-0m:
I would like to work on this issue

eldadfux:
Hey @adi-mohan, I have just spoke with @hack3r-0m about this issue minutes before you message. Would love to assign you any other issue you might be able to help with. Feel free to reach me out on Discord: https://appwrite.io/discord

adi-mohan:
Ok. Np. I would like to work on #547 if it is still open.

eldadfux:
@adi-mohan checking for you on the issue page...

eldadfux:
Changes merged, thanks to @hack3r-0m for the amazing work on this. This update will be included in the next 0.7 release 🥳

eldadfux:
@hack3r-0m awesome work on this one! This upgrade is now available as part of v0.7 🥳

❤️

PR #702, opened 2020-10-23
Replace .mmdb reader with maxmind and native C extension

hack3r-0m:
## What does this PR do?

Replaces current geoip2 mmdb reader with [maxmind reader](https://github.com/maxmind/MaxMind-DB-Reader-php) along with [native C extension](https://github.com/maxmind/libmaxminddb) which is noticeably faster.

## Test Plan

Compatible with previous test cases

## Related PRs and Issues

#474 

Yes

## Doubtful Regarding :

I had to add a block from lines 142 to 153 in Dockerfile (which is already repeated at line 22 to 33 and 107 to 127) otherwise I was getting error no autoconf, no compiler found, etc, anything wrong here?

In line 46 Dockerfile, according to libmaxmindb docs, `ldconfig` should be run but in alpine `ldconfig` needs to be run with config directory(https://stackoverflow.com/questions/36990951/ldconfig-seems-no-functional-under-alpine-3-3), is that correct path on that line?

In line 164 Dockerfile, is that path correct to load extension for ini file?

I am not sure if any changes are needed in `docker-compose.yml`, `docker-compose.nginx.yml` and `Dockerfile.nginx`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,6 +34,8 @@ RUN \
 
 RUN docker-php-ext-install sockets
 
+# install C extension for .mmdb reader
+# https://github.com/maxmind/libmaxminddb
 RUN wget https://github.com/maxmind/libmaxminddb/releases/download/1.4.3/libmaxminddb-1.4.3.tar.gz
 RUN tar -xvf libmaxminddb-1.4.3.tar.gz && \
     cd ./libmaxminddb-1.4.3 && \
@@ -42,7 +44,6 @@ RUN tar -xvf libmaxminddb-1.4.3.tar.gz && \
     make check && \
     make install && \
     ldconfig ./libmaxminddb-1.4.3
-
 
 RUN \
   # Redis Extension
@@ -151,6 +152,7 @@ RUN \
   zlib-dev \
   brotli-dev
 
+# Enabling C extension for .mmdb reader and adding loading it
 RUN apk add libmaxminddb-dev && \
     cd vendor/maxmind-db/reader/ext && ls -a && \
     phpize && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,23 +30,14 @@ RUN \
   wget \
   git \
   zlib-dev \
-  brotli-dev
+  brotli-dev \
+  libmaxminddb-dev
 
 RUN docker-php-ext-install sockets
 
-# install C extension for .mmdb reader
-# https://github.com/maxmind/libmaxminddb
-RUN wget https://github.com/maxmind/libmaxminddb/releases/download/1.4.3/libmaxminddb-1.4.3.tar.gz
-RUN tar -xvf libmaxminddb-1.4.3.tar.gz && \
-    cd ./libmaxminddb-1.4.3 && \
-    ./configure && \
-    make && \
-    make check && \
-    make install && \
-    ldconfig ./libmaxminddb-1.4.3
-
 RUN \
   # Redis Extension
+
   wget -q https://github.com/phpredis/phpredis/archive/$PHP_REDIS_VERSION.tar.gz && \
   tar -xf $PHP_REDIS_VERSION.tar.gz && \
   cd phpredis-$PHP_REDIS_VERSION && \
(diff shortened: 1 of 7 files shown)


eldadfux:
Hey @hack3r-0m great work, this looks like a very promising start!

As for your question no need to make any changes in `docker-compose.yml`, `docker-compose.nginx.yml` and `Dockerfile.nginx`.

About the duplicated code part. We use a multistage Docker build, the first stage install all the composer dependencies. The second stage install all the PHP extensions and compile them. In the final stage we copy the vendor directory from the first stage and the `.so` files from the second stage and add them to the correct php.ini path.

It would be much easier for you to take a look at how we install the phpredis extension and replicate that to make the maxmind one work.

If you need more advice, let me know on Discord.

eldadfux:
This is the section where we download and compile extensions:
https://github.com/appwrite/appwrite/blob/198b6e02b27e73868952f2a95cd0841a5c435884/Dockerfile#L48

This is the section where we copy the compiled .so files:
https://github.com/appwrite/appwrite/blob/198b6e02b27e73868952f2a95cd0841a5c435884/Dockerfile#L132

This is the section where we enable the extenstions in php.ini:
https://github.com/appwrite/appwrite/blob/198b6e02b27e73868952f2a95cd0841a5c435884/Dockerfile#L201

This way our final image is much smaller and don't include build tools that are unnecessary and considered not secure for production work.

hack3r-0m:
alright! and PHP test cases which are failing will be solved by just making method change to api/account.php as it also had some geoip2 usage

eldadfux:
Exactly... this also reminds me that api/users.php also has some uses of the geodb

eldadfux:
When you have some time, please also submit the results of the benchmarks on this PR. It's important that we'll have records of the change effect, and I will also want to share (and credit 😉) the huge performance improvement you managed to have with our community.

composer.json
```diff
@@ -45,7 +45,7 @@
         "utopia-php/domains": "1.1.*",
 
         "resque/php-resque": "1.3.6",
-        "geoip2/geoip2": "2.10.0",
+        "maxmind-db/reader": "1.8.*",
```
eldadfux, line 48: Why do we need this dependency when using the native C extension?

Dockerfile
```diff
@@ -128,6 +139,30 @@ COPY ./docs /usr/src/code/docs
 COPY ./public /usr/src/code/public
 COPY ./src /usr/src/code/src
 
+RUN \
```
eldadfux, line 142: We don't need to re-build the extension again on this image stage. Please check my previous comment about the phpredis extension and how we copy .so file between stages to reduce dependencies and final image size.

Dockerfile
```diff
@@ -34,6 +34,17 @@ RUN \
 
 RUN docker-php-ext-install sockets
 
+# install C extension for .mmdb reader
+# https://github.com/maxmind/libmaxminddb
+RUN wget https://github.com/maxmind/libmaxminddb/releases/download/1.4.3/libmaxminddb-1.4.3.tar.gz
```
eldadfux, line 39: We should combine all this `RUN` statements to better manage the image caching. Take a look at the `RUN` statement where we install `phpredis` and `swoole` extensions.

eldadfux (changes requested):
Great work! Added some comments to make build a bit cleaner and more consistent.

composer.json
hack3r-0m: [As mentioned in their reop](https://github.com/maxmind/MaxMind-DB-Reader-php#optional-php-c-extension), it works along with reader API. so after the composer installs reader in vendor extension can be enabled from vendor/maxmind-db/reader/ext (from their README). is it the wrong approach?

Dockerfile
hack3r-0m: yes, i am still working on Dockerfile

composer.json
eldadfux: I just went through their docs. Your approach seems OK 👍🏻

hack3r-0m:
## BENCHMARKS

OLD: geoip2 reader without any extension
AVG(5 runs): 36.2 seconds for 100k searches (running after breaks, less cache) and 34.7 seconds (running frequently, more cache)

NEW: maxmind PHP mmdb reader with native C extension
AVG(5 runs): 29.1 seconds for 100k searches (running after breaks, less cache) and 28.9 seconds (running frequently, more cache)

eldadfux (approved):

eldadfux:
Great work @hack3r-0m  ❤️❤️

eldadfux:
Please reach out to me on our [Discord server](https://appwrite.io/discord) if you would like me to send you some cool Appwrite stickers and magnets! Please note that shipping might take longer times because of COVID19.

As a way of saying thank you, we have just invited you to join the Appwrite organization on GitHub. You can accept the invite by visiting https://github.com/orgs/appwrite/invitation. By joining our team, you will officially be an Appwrite maintainer on GitHub.

You can change your membership [visibility settings](https://help.github.com/en/github/setting-up-and-managing-your-github-user-account/publicizing-or-hiding-organization-membership), so your new Appwrite team membership badge will show up on your personal GitHub profile.

Issue #730, opened 2020-11-04
Endpoint to get a password reset link or token to be able to implement a custom password reset flow 


saibotma:
## 🚀 Feature

It would be great to have an endpoint where a server could request a password reset link or token for a specific user.
The server could then send this link to the user using his own emailing system, SMS, etc. and would not be limited to the password reset feature currently available in Appwrite.

Firebase can be used as an orientation: https://firebase.google.com/docs/auth/admin/email-action-links

@eldadfux suggested a temporary, easy to implement workaround solution on Discord: 

> I think the only way to do this would be to parse Appwrite recovery email for the generated secret token. If you could raise this as an issue we can add that token to the POST /v1/account/recovery response object, and then you can just take the userID and the token from the response and transmit them to the user anyway you like

## Have you spent some time to check if this issue has been raised before?

yes

### Have you read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/master/CODE_OF_CONDUCT.md)?

yes

## Pitch

As stated above.

saibotma:
We @tappeddev need this feature for a next software project. Without this feature it is not possible for us to use appwrite. 
We would love to contribute and create a PR.
Do you have any quick tips on where to start and what we have to keep in mind?

eldadfux:
@saibotma I think there are only 2 missing pieces to help you make this feature possible, but please confirm I'm getting it right.

1. You need a way to get the secret hash after using the password token creation endpoint (https://appwrite.io/docs/client/account?sdk=web#createRecovery) response, as this is the only thing you're missing in order to not have to parse the email message for the token itself and submit the confirmation endpoint (https://appwrite.io/docs/client/account#updateRecovery) successfully?

2. You need a way to make sure that Appwrite doesn't send any emails by its own on password reset?

And just a general question, is this a workaround for another issue with the email template or way to overcome this missing SMS recovery?

JulianBissekkou:
Hey @eldadfux I am a developer at @tappeddev as well and I am helping @saibotma with this project.


I would like to explain the exact problem again to make sure that we find the correct solution: 

**Problem:**
We have multiple users that are connected to one E-Mail address. 
In appwrite, every user has a unique email since we can't link multiple users to one E-Mail. _The unique emails do not exist and are generated by us._
In case the user forgets the password, we have to send the password reset link to the correct E-Mail address which is stored in _our database._ 

**Proposed Solution:**
[...]

PR #737, opened 2020-11-20
Feat 730 new webhook events

eldadfux:
## What does this PR do?

Adds to new system events: `account.recovery.create` and `account.recovery.update`, the new event will allow to customize Appwrite password recovery process. The new event are supported for usage with both Appwrite webhooks and functions.

## Test Plan

Added a new test class to check all Webhooks payload are sent as expected to any 3rd party HTTP server.

## Related PRs and Issues

#730 

Yes.


app/config/events.php
```diff
@@ -19,6 +19,18 @@
     'account.update.prefs' => [
         'description' => 'This event triggers when the account preferences are updated.',
     ],
+    'account.recovery.create' => [
+        'description' => 'This event triggers when the account recovery token is created.',
+    ],
+    'account.recovery.update' => [
+        'description' => 'This event triggers when the account recovery token is validated.',
```
JulianBissekkou, line 26: Is there a specific reason why you named this `.update` instead of `validate`? 
Just curious since i was confused when reading the name of the hook, but it was clear after i read the description 🤔

JulianBissekkou: same for `account.verification.update`. Does it makes sense to call it `account.verification.validated`?

JulianBissekkou (commented):
Nice! 
I am not sure were the actual event is send because my PHP skills are pretty bad tbh :P 
Can you quickly tell me where the payload for those events are send?
I just want to make sure that i contains everything we need. 
Thanks a lot ☺️

app/config/events.php
eldadfux: Because we're trying to stay consisted with REST standards. We're using POST request (create) to create token X, and we use PUT request (update) to verify it.

JulianBissekkou: I understand, makes sense :)

eldadfux:
> Nice!
> I am not sure were the actual event is send because my PHP skills are pretty bad tbh :P
> Can you quickly tell me where the payload for those events are send?
> I just want to make sure that i contains everything we need.
> Thanks a lot ☺️

This is the structure of the object passed as the request payload: https://github.com/appwrite/appwrite/blob/ea0f51a4dadb218b6c2f0127daf1aca7568e182f/src%2FAppwrite%2FUtopia%2FResponse%2FModel%2FToken.php after the new version will be released we'll also have this available in the [website docs](https://twitter.com/appwrite_io/status/1328432348938063879).

The request itself is being sent from this part of code, which runs as the `appwrite-worker-webhooks` container: https://github.com/appwrite/appwrite/blob/2596c2c1f89551aa2624652cbd4c1f80366117f2/app%2Fworkers%2Fwebhooks.php#L78

I think what still missing is the user ID, but I want to add it as the `X-Appwrite-Webhook-UserID` header

eldadfux:
Finished adding all missing tests and added multiple new events!

app/config/events.php
```diff
@@ -40,8 +52,8 @@
     'database.documents.create' => [
         'description' => 'This event triggers when a database document is created.',
     ],
-    'database.documents.patch' => [
-        'description' => 'This event triggers when a database document is patched.',
+    'database.documents.update' => [
```
christyjacob4, line 55: Is the reasoning for `update` same as here? https://github.com/appwrite/appwrite/pull/737/files#r528358275

app/config/events.php
```diff
@@ -67,4 +79,22 @@
...
+    'teams.update' => [
+        'description' => 'This event triggers when a team is updated.',
+    ],
+    'teams.delete' => [
+        'description' => 'This event triggers when a team is deleted.',
+    ],
+    'teams.memberships.create' => [
+        'description' => 'This event triggers when a team memberships is created.',
+    ],
+    'teams.memberships.update.status' => [
```
christyjacob4, line 94: Can this not be 

```suggestion
    'teams.memberships.update' => [
```

src/Appwrite/Utopia/Response/Model/User.php
```diff
@@ -47,13 +47,6 @@ public function __construct()
                 'default' => new \stdClass,
                 'example' => ['theme' => 'pink', 'timezone' => 'UTC'],
             ])
-            ->addRule('roles', [
```
christyjacob4, line 50: Any reason this was removed?

christyjacob4 (approved):
LGTM @eldadfux 👍  🎉

src/Appwrite/Utopia/Response/Model/User.php
eldadfux: Yeh, it was too complected to handle the return of this output, because we only get it on runtime and don't have it in the DB. We might create a dedicated route for it or save it in the DB later.

app/config/events.php
eldadfux: Might need to re-think the naming conventions of the events name, might require extra work for migration from old setup.

app/config/events.php
eldadfux: Might need to re-think the naming conventions of the events name, might require extra work for migration from old setup.

PR #762, opened 2020-12-12
Feat remove default order field where possible

eldadfux:
## What does this PR do?

Optimized system queries by using the internal Appwrite numeric ID as sort field instead of using an extra join/s for custom field. This change was only applied to queries where the sort was done by according to creation date which can be replaced by our incremental numeric ID (the internal ID, not the public UID that looks like: "5e63e0a61d9c2").

## Test Plan

Leveraging current test plan.

## Related PRs and Issues

None.

Yes.


christyjacob4 (approved):
LGTM 👍

eldadfux:
🚀

PR #766, opened 2020-12-14
feat: starter for maintenance task

christyjacob4:
## What does this PR do?
This PR contains the implementation of a maintenance container which is supposed to schedule maintenance tasks across the Appwrite system

## Test Plan
Will be writing e2e tests for the same. 

## Related PRs and Issues
https://github.com/appwrite/rfc/blob/main/001-maintenance-worker.md

Yes


app/tasks/maintenance.php
```diff
@@ -0,0 +1,19 @@
...
+
+$cli
+    ->task('maintenance')
+    ->desc('Schedules maintenance tasks and publishes them to resque')
+    ->action(function () {
+        $interval = App::getEnv('_APP_MAINTENANCE_INTERVAL', '');
+
+        for($i = 0; $i <= 10; ++$i){
+            Console::log('Starting the maintenance worker every '.$interval.' seconds');
+            sleep($interval);
```
christyjacob4, line 16: @eldadfux is the `Console::loop` function available to be consumed here ?

eldadfux: Yes. Update the composer dependency to the new utopia/cli:0.8.0 version and it will be available. Appwrite container rebuild is required.

docker-compose.yml
```diff
@@ -277,6 +277,22 @@ services:
...
+    networks:
+      - appwrite
+    depends_on:
+      - redis
+      - mariadb
+    environment:
+      - _APP_ENV
+      - _APP_REDIS_HOST
+      - _APP_REDIS_PORT
+
```
christyjacob4, line 295: Something wierd I noticed in the appwrite-maintenance container. 
Even though I have not passed `_APP_MAINTENANCE_INTERVAL` environment variable in the docker-compose, Im able to access it in `maintenance.php`

Dockerfile
```diff
@@ -93,7 +93,8 @@ ENV _APP_SERVER=swoole \
     _APP_FUNCTIONS_TIMEOUT=900 \
     _APP_FUNCTIONS_CONTAINERS=10 \
     _APP_SETUP=self-hosted \
-    _APP_VERSION=$VERSION
+    _APP_VERSION=$VERSION \
+    _APP_MAINTENANCE_INTERVAL=10
```
eldadfux, line 97: Let me know what time units are we using here, we need to update this option in the homepage docs.

docker-compose.yml
```diff
@@ -277,6 +277,22 @@ services:
...
+    entrypoint: maintenance
+    container_name: appwrite-maintenance
+    build:
+      context: .
+    restart: unless-stopped
+    networks:
+      - appwrite
+    depends_on:
+      - redis
+      - mariadb
```
eldadfux, line 290: I think you don't need MariaDB here, if you do make sure to pass DB env vars as well as the Redis ones.

Dockerfile
christyjacob4: I believe if we use seconds, we can stay consistent with the epoch time conventions

app/tasks/maintenance.php
```diff
@@ -0,0 +1,39 @@
...
+    ->task('maintenance')
+    ->desc('Schedules maintenance tasks and publishes them to resque')
+    ->action(function () {
+        // Convert string to integer
+        $interval = App::getEnv('_APP_MAINTENANCE_INTERVAL', '') + 0;
+        //Convert Seconds to microseconds
+        $interval = $interval * 1000000;
+
+        Console::loop(function() {
+            $projects = $consoleDB->getCollection([
```
christyjacob4, line 23: @eldadfux At this point , Im unable to use `$consoleDB`
Let me know how I can access it.

eldadfux: You can check for an example here: https://github.com/appwrite/appwrite/blob/d0f7558ddf1c58ecaeffa9503ac84d0ccc11daee/app%2Fworkers%2Fdeletes.php#L189

Note that you'll need the DB env vars available in the container, pass them in the docker-compose file (my previous comment :)).

app/tasks/maintenance.php
christyjacob4: Done

docker-compose.yml
christyjacob4: @eldadfux

docker-compose.yml
christyjacob4: Hmm.. I need to be making database calls to get the list of project IDs right?

app/tasks/maintenance.php
christyjacob4: Thank you 😄

docker-compose.yml
eldadfux: Did you define a default value in the main Dockerfile?

docker-compose.yml
eldadfux: Yes. no problem here.

Dockerfile
```diff
@@ -93,7 +93,8 @@ ENV _APP_SERVER=swoole \
     _APP_FUNCTIONS_TIMEOUT=900 \
     _APP_FUNCTIONS_CONTAINERS=10 \
     _APP_SETUP=self-hosted \
-    _APP_VERSION=$VERSION
+    _APP_VERSION=$VERSION \
+    _APP_MAINTENANCE_INTERVAL=864000
```
eldadfux, line 97: Add a comment here with the # of days

docker-compose.yml
christyjacob4: https://github.com/appwrite/appwrite/pull/766/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R97

This one

app/workers/deletes.php
```diff
@@ -95,6 +98,23 @@ protected function deleteUser(Document $document, $projectId)
...
+        
+        $projectIds = $document->getAttribute('projectIds', []);
+        foreach ($projectIds as $projectId) {
+            if (!($projectDB = $this->getProjectDB($projectId))) {
+                throw new Exception('Failed to get projectDB for project '.$projectId, 500);
+            }
+
+            // Delete Executions
+            $this->deleteByGroup([
+                '$collection='.$document->getCollection(),
```
eldadfux, line 112: What document is passed here? Maybe we should use Database::SYSTEM_COLLECTION_EXECUTIONS to avoid any mistakes that could lead to data loss?

app/tasks/maintenance.php
```diff
@@ -0,0 +1,59 @@
...
+use Appwrite\Database\Document;
+use Appwrite\Database\Adapter\MySQL as MySQLAdapter;
+use Appwrite\Database\Adapter\Redis as RedisAdapter;
+use Appwrite\Database\Validator\Authorization;
+use Utopia\App;
+use Utopia\CLI\Console;
+use Utopia\Config\Config;
+
+
+function getConsoleDB() {
```
eldadfux, line 17: I'll try and find a better way to manage these dependencies. Don't worry about it for now, but maybe add a //TODO comment

app/tasks/maintenance.php
```diff
@@ -0,0 +1,59 @@
...
+}
+
+$cli
+    ->task('maintenance')
+    ->desc('Schedules maintenance tasks and publishes them to resque')
+    ->action(function () {
+        // Convert string to integer
+        $interval = App::getEnv('_APP_MAINTENANCE_INTERVAL', '') + 0;
+        //Convert Seconds to microseconds
+        $interval = $interval * 1000000;
```
eldadfux, line 33: Add a comment with the # of days to $interval value

eldadfux (changes requested):

app/tasks/maintenance.php
christyjacob4: Actually interval represents # days in seconds (1 day = 86400s)

app/workers/deletes.php
christyjacob4: Screenshot 2020-12-17 at 9 08 15 PM

This is the document passed

docker-compose.yml
eldadfux: So that's the value you get.

app/workers/deletes.php
eldadfux: We discussed changing the way we trigger that worker anyway... So this is probably not relevant anymore.

app/tasks/maintenance.php
eldadfux: Yeh, I got it, just would be easier to read the code if you had a comment beside the large number.

docker-compose.yml
christyjacob4: If any env variable that you declare in the main docker file becomes visible in the services , then what is the significance of lines https://github.com/appwrite/appwrite/pull/766/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R291-R294

Dockerfile
christyjacob4: Done

app/tasks/maintenance.php
christyjacob4: Done

app/tasks/maintenance.php
christyjacob4: Sure

app/workers/deletes.php
christyjacob4: sure 👍

app/tasks/maintenance.php
```diff
@@ -0,0 +1,95 @@
...
+use Appwrite\Database\Database;
+use Appwrite\Database\Document;
+use Appwrite\Database\Adapter\MySQL as MySQLAdapter;
+use Appwrite\Database\Adapter\Redis as RedisAdapter;
+use Appwrite\Database\Validator\Authorization;
+use Utopia\App;
+use Utopia\CLI\Console;
+use Utopia\Config\Config;
+
+define("DELETE_QUEUE_NAME", "v1-deletes");
```
eldadfux, line 16: For consistency, let's add new constants for all the queues, better add them in our init.php constants list if no relevant class is available. Actually maybe we can add them in the Event class. Let me know what you think.

When setting new constants, we also better be consistent and use the same syntax. Right now we have preferred the `const X = 'x';` over `define('X', 'x');`. +Single quotes have better performance than double quotes in PHP.

docker-compose.yml
eldadfux: It allow us to edit the vars using the .env file, we used to edit them from the docker-compose.yml in the past, but the env allow as to reuse vars and avoid duplicated values in the main file.

eldadfux (changes requested):
Added a few more comments and answers.

app/workers/deletes.php
```diff
@@ -95,6 +120,70 @@ protected function deleteUser(Document $document, $projectId)
...
+        });
+        
+        foreach ($projectIds as $projectId) {
+            Console::success("Deleting abuse logs for Project: ", $projectId);
+            $timeLimit->setNamespace('app_'.$projectId);
+            $abuse = new Abuse($timeLimit); 
+
+            $status = $abuse->cleanup($timestamp);
+            if (!$status) {
+                throw new Exception('Failed to delete Abuse logs for project '.$projectId, 500);
```
eldadfux, line 160: You don't need HTTP related error codes here, because this is not the HTTP API scope. You can just skip that param.

app/workers/deletes.php
```diff
@@ -95,6 +120,70 @@ protected function deleteUser(Document $document, $projectId)
...
+            throw new Exception('Failed to delete audit logs. No timestamp provided');
+        }
+
+        foreach ($projectIds as $projectId) {
+            $adapter = new AuditAdapter($register->get('db'));
+            $adapter->setNamespace('app_'.$projectId);
+            $audit = new Audit($adapter);
+            $status = $audit->cleanup($timestamp);
+            if (!$status) {
+                throw new Exception('Failed to delete Audit logs for project'.$projectId, 500);
```
eldadfux, line 182: You don't need HTTP related error codes here, because this is not the HTTP API scope. You can just skip that param.

app/workers/deletes.php
```diff
@@ -95,6 +120,70 @@ protected function deleteUser(Document $document, $projectId)
         ], $this->getProjectDB($projectId));
     }
 
+    protected function deleteExecutionLogs(Document $document) 
+    {
+        $projectIds = $document->getAttribute('projectIds', []);
+        foreach ($projectIds as $projectId) {
+            if (!($projectDB = $this->getProjectDB($projectId))) {
+                throw new Exception('Failed to get projectDB for project '.$projectId, 500);
```
eldadfux, line 128: You don't need HTTP related error codes here, because this is not the HTTP API scope. You can just skip that param.

app/tasks/maintenance.php
christyjacob4: Thanks eldad. Those are some useful insights! I will add them in Events. That is more relevant

docker-compose.yml
```diff
@@ -55,7 +55,7 @@ services:
       - ./psalm.xml:/usr/src/code/psalm.xml
       - ./tests:/usr/src/code/tests
       - ./app:/usr/src/code/app
-      # - ./vendor:/usr/src/code/vendor
+      - ./vendor:/usr/src/code/vendor
```
eldadfux, line 58: Might be better to comment this by default, it would make new contributors life harder as they will have to install all the dependencies with composer locally, especially if they don't use PHP often.

eldadfux: I think this line also makes the tests to fail, makes sure to comment all vendor mounts (travis don't install dependencies locally - which is OK because otherwise the build will take longer time to run)

app/workers/deletes.php
```diff
@@ -142,6 +222,49 @@ protected function deleteById(Document $document, Database $database, callable $
...
+
+        $executionStart = \microtime(true);
+        
+        while($sum === $limit) {
+            $chunk++;
+
+            Authorization::disable();
+            $projects = $this->getConsoleDB()->getCollection([
+                'limit' => $limit,
+                'offset' => 0,
```
eldadfux, line 241: If the offset is not changing you don't move forward in the list, and will end in an endless loop. Unlike the original code, here projects don't get deleted and the list isn't gettin smaller on every iteration.

eldadfux: Changing the limit with no offset will only return a bigger list at every loop. Old items + new items.

app/workers/deletes.php
```diff
@@ -95,6 +121,60 @@ protected function deleteUser(Document $document, $projectId)
...
+    {
+        $this->deleteForProjectIds(function($projectId) {
+            if (!($projectDB = $this->getProjectDB($projectId))) {
+                throw new Exception('Failed to get projectDB for project '.$projectId);
+            }
+
+            // Delete Executions
+            $this->deleteByGroup([
+                '$collection='.Database::SYSTEM_COLLECTION_EXECUTIONS,
+                '$projectId='.$projectId
```
eldadfux, line 134: Execution logs don't have the projectId attribute. You can check system collections structures here: `app/config/collections.php`. This filter will always return zero documents. Because Appwrite has separate tables for each projects you can just delete all execution logs with not special filters.

docker-compose.yml
christyjacob4: Okay commenting it 👍

app/workers/deletes.php
christyjacob4: Yes I thought about it too. I referred to the `deleteByGroup()` function which had this same issue. I will fix it in both places now.

app/workers/deletes.php
christyjacob4: Okay understood 👍

app/controllers/api/health.php
```diff
@@ -164,7 +164,7 @@
     ->action(function ($response) {
         /** @var Appwrite\Utopia\Response $response */
 
-        $response->json(['size' => Resque::size('v1-usage')]);
+        $response->json(['size' => Resque::size(Event::USAGE_QUEUE_NAME)]);
```
eldadfux, line 167: Your'e missing a namespace declaration in this file, which cause the tests to fail.

Please add constants for all the other `Resque::size` calls in this controller.

app/init.php
```diff
@@ -52,7 +52,12 @@
 const APP_SOCIAL_GITHUB = 'https://github.com/appwrite';
 const APP_SOCIAL_DISCORD = 'https://appwrite.io/discord';
 const APP_SOCIAL_DEV = 'https://dev.to/appwrite';
-const APP_SOCIAL_STACKSHARE = 'https://stackshare.io/appwrite';
+const APP_SOCIAL_STACKSHARE = 'https://stackshare.io/appwrite'; 
+// Deletion Types
+const DELETE_TYPE_DOCUMENT = 'document';
+const DELETE_TYPE_EXECUTION_LOGS = 'execution_logs';
```
eldadfux, line 58: This can probably be named just "executions"

app/controllers/api/health.php
christyjacob4: Done 👍

app/init.php
christyjacob4: Done 👍

eldadfux:
@christyjacob4 LGTM.

@TorstenDittmann this is a pretty bug change, can you please double check us?

eldadfux (approved):

TorstenDittmann:
> @christyjacob4 LGTM.
> 
> @TorstenDittmann this is a pretty bug change, can you please double check us?

Will do during the day 👍

eldadfux:
big** hehe

app/tasks/maintenance.php
```diff
@@ -0,0 +1,67 @@
...
+        'timestamp' => time() - $interval
+    ]);
+}
+
+$cli
+    ->task('maintenance')
+    ->desc('Schedules maintenance tasks and publishes them to resque')
+    ->action(function () {
+        // # of days in seconds (1 day = 86400s)
+        $interval = App::getEnv('_APP_MAINTENANCE_INTERVAL', '') + 0;
```
TorstenDittmann, line 53: We could add 86400 as a fallback value here if something gets misconfigured. The type casting with `+ 0` looks a little bit off to me. How about this:

`$interval = (int) App::getEnv('_APP_MAINTENANCE_INTERVAL', '86400');`

app/workers/deletes.php
```diff
@@ -142,6 +221,49 @@ protected function deleteById(Document $document, Database $database, callable $
...
+        $executionStart = \microtime(true);
+        
+        while($sum === $limit) {
+            $chunk++;
+
+            Authorization::disable();
+            $projects = $this->getConsoleDB()->getCollection([
+                'limit' => $limit,
+                'offset' => $count,
+                'orderField' => '$id',
```
TorstenDittmann, line 241: Why do we order here? 🤔

TorstenDittmann (approved):
Only stumbled upon some minor things 🙂 

lgtm 👍

app/workers/deletes.php
eldadfux: @christyjacob4, @TorstenDittmann is right, no need for this anymore. Actually, the default order  (when passing no order setting) has a slightly better performance.

app/tasks/maintenance.php
christyjacob4: Yea that looks better 👍

app/workers/deletes.php
christyjacob4: Okay understood. Thanks 👍

eldadfux:
Great work @christyjacob4 👏🏻

PR #772, opened 2020-12-22
feat: Response Filters

christyjacob4:
## What does this PR do?

Starter code for Response filters as mentioned in https://github.com/appwrite/rfc/pull/4

## Test Plan

Wil write unit tests to test response filters

## Related PRs and Issues
https://github.com/appwrite/rfc/pull/4

Yes


eldadfux:
Looks good so far.

app/controllers/general.php
```diff
@@ -16,12 +16,23 @@
...
 use Utopia\CLI\Console;
 
 Config::setParam('domainVerification', false);
 Config::setParam('cookieDomain', 'localhost');
 Config::setParam('cookieSamesite', Response::COOKIE_SAMESITE_NONE);
 
+function mapResponseFormatToClass(string $responseFormat): Filter {
+    switch($responseFormat) {
+        case preg_match($responseFormat, "/0\.[0-6]\.\d?/"):
+            return new V06();
```
christyjacob4, line 30: Does this cause a new instance of the filter to be created upon each request? 
If yes, I should probably implement the Filter class as a Singleton. 
WDYT @eldadfux

app/controllers/general.php
```diff
@@ -92,6 +103,18 @@
...
+    */
+    $responseFormatEnvVar = App::getEnv('_APP_SYSTEM_RESPONSE_FORMAT', '');
+    $responseFormatHeader = $request->getHeader('x-appwrite-response-format', '');
+    $responseFormat = empty($responseFormatHeader) ? $responseFormatEnvVar : $responseFormatHeader;
+    if (empty($responseFormat) || ($filter = mapResponseFormatToClass($responseFormat)) == null) {
+        throw new Exception('No filter available for response format : '.$responseFormat, 404);
+    } else {
+        Response::setFilter($filter);   
+    }
+
```
eldadfux, line 117: You can simplify this logic by doing something like this:

```php
/* 
 * Response format
 */
$responseFormat = $request->getHeader('x-appwrite-response-format', 
    App::getEnv('_APP_SYSTEM_RESPONSE_FORMAT', ''));

switch($responseFormat) {
    case '0.6.0':
    case '0.6.1':
    case '0.6.2':
        Response::setFilter(new V06());
    break;
}
```

We get the header first or fallback to the env var if no header provided. If no filter is requested we just use the default response models with no parsing. This also help us avoid any check on the server startup and avoid the use of any global functions.

Creating a new instance of the filter on every request shouldn't be significant and the cost of complexity might not worth a singleton anyway.

app/controllers/general.php
eldadfux: see my second comment about this.

app/controllers/general.php
eldadfux: Another option here is to use the native `version_compare` function:

```php
switch($responseFormat) {
    case version_compare ($responseFormat , '0.6.2', '=<') :
        Response::setFilter(new V06());
    break;
}

```

christyjacob4: Awesome. Ill make these changes

christyjacob4: @eldadfux 

> `$responseFormat = $request->getHeader('x-appwrite-response-format', 
    App::getEnv('_APP_SYSTEM_RESPONSE_FORMAT', ''));`

This is really elegant! Love this

christyjacob4: and this 

> `switch($responseFormat) {
    case version_compare ($responseFormat , '0.6.2', '=<') :
        Response::setFilter(new V06());
    break;
}`

❤️

app/controllers/general.php
```diff
@@ -108,14 +99,15 @@ function mapResponseFormatToClass(string $responseFormat): Filter {
...
-        throw new Exception('No filter available for response format : '.$responseFormat, 404);
-    } else {
-        Response::setFilter($filter);   
+    $responseFormat = $request->getHeader('x-appwrite-response-format', App::getEnv('_APP_SYSTEM_RESPONSE_FORMAT', ''));
+    switch($responseFormat) {
+        case version_compare ($responseFormat , '0.6.2', '=<') :
+            Response::setFilter(new V06());
+            break;
+        default:
+            throw new Exception('No filter available for response format : '.$responseFormat, 404);
```
eldadfux, line 108: Not sure I would throw an exception here, maybe just let the response class use the latest format?

eldadfux (changes requested):
Test are failing, you can remove the validation on the server startup

christyjacob4:
>Tests are still failing, you can remove the validation on the server startup

@eldadfux But don't we require the env variable validation?

app/controllers/general.php
christyjacob4: But what if we are sent some unrecognised version? Something like 1.3.2 or v1.3.2 . I feel we must inform the user that such a response format doesn't exist.

TorstenDittmann:
Do I see that correctly, that always the current answer is gonna be passed to a filter?

Then we would have to revise this again and again with every breaking change.

So let's say we have version 0.7 and the 0.6 filter is used. Now we update to to version 0.8 with a lot of breaking changes again, does that mean we would have adapt the 0.6 filter again to process a 0.8 response?

eldadfux:
Most chances are that the 0.6 filters will work with v0.8 with no changes or just minimal ones from our side. @christyjacob4 have added test to make sure new changes will still be validated against older filters with any new version.

I guess at some point we'll start to drop support for versions older than X where X should give enough time for app developers to migrate users.

app/http.php
```diff
@@ -23,6 +23,7 @@
 $http = new Server("0.0.0.0", 80);
 
 $payloadSize = max(4000000 /* 4mb */, App::getEnv('_APP_STORAGE_LIMIT', 10000000 /* 10mb */));
+$responseFormat = App::getEnv('_APP_SYSTEM_RESPONSE_FORMAT', '');
```
eldadfux, line 26: No need for this line

.env
```diff
@@ -31,3 +31,4 @@ _APP_FUNCTIONS_CPUS=1
 _APP_FUNCTIONS_MEMORY=128
 _APP_FUNCTIONS_MEMORY_SWAP=128
 _APP_MAINTENANCE_INTERVAL=86400
+_APP_SYSTEM_RESPONSE_FORMAT=0.6.2
```
eldadfux, line 34: should be empty by default

eldadfux (changes requested):

Dockerfile
```diff
@@ -97,6 +97,7 @@ ENV _APP_SERVER=swoole \
     _APP_FUNCTIONS_MEMORY_SWAP=128 \
     _APP_SETUP=self-hosted \
     _APP_VERSION=$VERSION \
+    _APP_SYSTEM_RESPONSE_FORMAT=0.6.2 \
```
eldadfux, line 100: Should be empty by default

app/config/variables.php
```diff
@@ -175,4 +175,10 @@
         'required' => false,
         'question' => '',
     ],
+    [
+        'name' => '_APP_SYSTEM_RESPONSE_FORMAT',
+        'default' => '0.7',
```
eldadfux, line 180: Should be empty by default

eldadfux (changes requested):

eldadfux:
Tests are green! 🟢🟢🟢

@TorstenDittmann, would be great to get your help double checking the solution and PR.

src/Appwrite/Utopia/Response/Filters/V06.php
```diff
@@ -0,0 +1,288 @@
...
+        foreach($countries as $country) {
+            $parsedResponse[$country['code']] = $country['name'];
+        }
+        $content['countries'] = $parsedResponse;
+        return $content;
+    }
+
+    private function parseLocale(array $content) 
+    {
+        $content['ip'] = empty($content['ip']) ? '' : $content['ip'];
```
TorstenDittmann, line 194: Now it is nitpicking 😅 don't know if it has any impact on performance, but wouldn't this get the same result without a method call:

`$content['ip'] = $content['ip'] ? $content['ip'] : ''`

TorstenDittmann (approved):
Just some minor things that catched my eye, they might not even be worth looking at. Awesome work @christyjacob4 :clap:

src/Appwrite/Utopia/Response/Filters/V06.php
eldadfux: `$content['ip'] = $content['ip'] ?? '';` can also do the trick

TorstenDittmann: Does `??` also takes care of an empty string and not just `null`?

app/controllers/general.php
```diff
@@ -94,6 +96,22 @@
     Storage::setDevice('files', new Local(APP_STORAGE_UPLOADS.'/app-'.$project->getId()));
     Storage::setDevice('functions', new Local(APP_STORAGE_FUNCTIONS.'/app-'.$project->getId()));
 
+    /* 
+    * Response format
+    */
+    $responseFormat = $request->getHeader('x-appwrite-response-format', App::getEnv('_APP_SYSTEM_RESPONSE_FORMAT', ''));
+    if (!empty($responseFormat)) {
```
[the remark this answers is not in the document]
christyjacob4: Good catch Torsten 👍

src/Appwrite/Utopia/Response/Filters/V06.php
TorstenDittmann: Yeah, totally right. Checking against an empty string would be stupid, since we make it an empty string anyway on the condition 😅

eldadfux:
@christyjacob4 2 last errors to fix and we can merge this PR 🤞🏻

eldadfux:
🎉

Issue #773, opened 2020-12-24
Linkedin Provider is not working(Flutter)


udcode:
Getting Error 500 - provider is not supported 
When trying to log in with LinkedIn.

Basic setup(flutter):
   Client _client = new Client();
    _client.setEndpoint(...).setProject(...):
   Account _account = new Account(_client)
   ....
    var result = _account.createOAuth2Session(provider:'linkedIn');

christyjacob4:
Hey @udcode thanks for raising this. We are working on this and will be out with a fix soon on the 0.7.x branch

Let me know if you need help setting up the development version of appwrite and I'd be happy to assist.

eldadfux:
@christyjacob4 we should probably also check why this is raising error 500 and not 400

christyjacob4:
Sure @eldadfux I'm testing the entire flow just to be sure

christyjacob4:
@udcode  Your error appears to be different from the fix that we have provided. 
Consider replacing your code with this and try again. I think it's got to do with the uppercase I in `linkedIn`

```dart
var result = _account.createOAuth2Session(provider:'linkedin');
```

udcode:
i tried with lower vase

christyjacob4:
@udcode We have merged the fix to the 0.7.x branch. Do you need help setting it up?

udcode:
Tnx!!
Help would be greatly appreciated 🙏

christyjacob4:
Okay you can follow these steps

```bash
$ git clone https://github.com/appwrite/appwrite/tree/master
$ cd appwrite 
$ git checkout 0.7.x

# Assuming you're coming from a previous version of appwrite and you are just testing the new version
# We recommend deleting the volumes from 0.6.x or prior 
$ docker volume rm $(docker volume ls -q)  

$ docker-compose up -d
```

Once the containers are up and running, head over to your dashboard at localhost (default). 
Create a project and head over to the users section 

![Screenshot 2020-12-24 at 8 00 39 PM](https://user-images.githubusercontent.com/20852629/103094193-b423a780-4622-11eb-93e3-65b52bdfdc97.png)

Head over to OAuth providers and enable Linkedin
![Screenshot 2020-12-24 at 8 00 57 PM](https://user-images.githubusercontent.com/20852629/103094199-bede3c80-4622-11eb-8306-9ed672cd7e50.png)

Create a project on [Linkedin](http://linkedin.com/developers) and get the appID and appSecret from Linkedin's console. 
[...]

PR #774, opened 2020-12-24
feat: updated new Linkedin scopes

christyjacob4:
## What does this PR do?
Updated the OAuth scopes for linkedin

## Test Plan
Created a Linkedin App and tested the oauth flow. The user was successfully added.

## Related PRs and Issues
#773 

Yes


app/controllers/api/account.php
```diff
@@ -292,8 +292,6 @@
         $appId = $project->getAttribute('usersOauth2'.\ucfirst($provider).'Appid', '');
         $appSecret = $project->getAttribute('usersOauth2'.\ucfirst($provider).'Secret', '{}');
 
-        $appSecret = \json_decode($appSecret, true);
```
christyjacob4, line 295: @eldadfux Is there any reason this line was added? It was breaking the oAuth flow in the Linkedin case while testing

eldadfux: Maybe because of the Apple OAuth adapter? can you check if Facebook or other providers has the same issue?

eldadfux: Yep, this line can be removed, not needed with the intro of DB filters.

eldadfux (approved):

PR #784, opened 2020-12-28
Feat 511 jwt support

eldadfux:
## What does this PR do?

Adds JWT authentication method for the Appwrite server API. Implementation of RFC#3: https://github.com/appwrite/rfc/blob/main/003-jwt-support.md

## Test Plan

Create a new test suit for validating the creation of the JWT and using it for authenticating on behalf of the user.

## Related PRs and Issues

#511 

Yes


punitdiwan:
This is much awaited feature request. I am really excited to see it. Waiting to get it merged to give a try.

eldadfux:
Updated the docs for server integration:

image

app/controllers/general.php
```diff
@@ -162,27 +160,31 @@
...
+
+    if (!empty($authKey)) { // API Key authentication
+        // Check if given key match project API keys
+        $key = $project->search('secret', $authKey, $project->getAttribute('keys', []));
+            
+        /*
+         * Try app auth when we have project key and no user
+         *  Mock user to app and grant API key scopes in addition to default app scopes
+         */
+        if (null !== $key && $user->isEmpty()) {
```
TorstenDittmann, line 174: Is there a difference to using just 
`if ($key && $user->isEmpty()) {`

I suppose `null` is a falsy value in PHP as well right?

TorstenDittmann (approved):
Looks goot do me 😃

app/controllers/general.php
eldadfux: yeh, no difference null is falsy

TorstenDittmann (approved):
lgtm 👍

lohanidamodar:
lgtm 👍

eldadfux:
Merging to 0.7.x on hidden mode, we'll release it later alongside the SDKs support.

fightforlife:
In the screenshot above it is written, that to create a valid jwt I need to authenticate the user with a valid session. That means I need to use a Cookie once to get the jwt, right?
(Doesn't this kind of defeats the purpose for clients that cannot use cookie authentication? Maybe I am just understanding it wrong)

eldadfux:
@fightforlife yes, the main use-case in mind right now  is to allow servers and cloud functions to extend Appwrite functionality using JWT. We could try and make the auth method produce JWT alongside/instead of cookies.

What kind of clients did you had in mind?

fightforlife:
@eldadfux thanks for the quick response! 
I am currently building a Android/iOS App which should be used to organize and track events for a hobby. (Minimal social network)
Since I am neither a frontend developer nor fluent in Javascript I am using a Service called Appgyver. (There are hundreds of these, but Appgyver is the only one which allows very complex logic and also connection to a REST API)
I am thinking of adding a simple Web frontend later.

Currently I am storing the cookie in persistent storage and add it to the header in every manual/hacky http request. Since this is kind of forbidden, it is not working as consistend as I hoped. (E.g. I can not use the official REST library but have to write my own http request.)


I think it would be really cool if the 'create session' endpoint could optionally return a jwt **instead** of a Cookie. It should not return both at the same time. If another client (web vs. App) needs another authentication it should request a new session with a Cookie as standard or instead _optionally_ a jwt.
This way each frontend could have his own authentication in his own session.

eldadfux:
Got it, have you considered a Cookie Jar solution for handling the cookie? this has worked very well for us in our Flutter SDK under similar circumstances.

How would you store and manage JWT (valid for 15m) renewal if we made that possible without a cookie?

fightforlife:
You are right this would be a problem. One would need 2 JWTs. One for access (short lived) and one to refresh (long lived) the token.

The process would look like in this picture:
https://supertokens.io/static/webflow/blog/jwt-right-way/images/image223x-p-800.png
Taken from https://supertokens.io/blog/are-you-using-jwts-for-user-sessions-in-the-correct-waw

This would enable clients to be used which absolutely don't support cookies.

But I think the client side implementation is similar complex to a manual cookie jar.

eldadfux:
Agree about the complexity. Honestly the server implementation itself is going to be even a bit more complex (not that this should be the main consideration).

On a browser context this implementation is much less secure then a server-only cookie in the context of XSS attack / malicious 3rd party script / evil browser extension.

smartlot:
First, where is the JWT secret? Second? How do I add claims to the JWT Token? Issuer? The primary reason for JWT tokens is to verify the client to third party services on the back-end. Without the ability to set the properties, this is not a real JWT implementation, no?

Why can we not set registered claims when generating the JWT token so that it can be used to verify the user to third party services?



https://jwt.io/introduction

TorstenDittmann:
Hey @smartlot 

We don't support custom claims yet, the current JWT implementation will allow you to authenticate to Appwrite on server side with the client's scope.

What exactly would be your use case? 🙂

smartlot:
I am trying to implement Talk from Cora (https://docs.coralproject.net/sso)l, a comments server, but without being able to implement the minimum requirement, based on jwt.io, I cannot set iss,exp, aud, or any claims in order to fully implement what JWT proposes; authentication on the backend for client services.

PR #787, opened 2020-12-29
Fix: deprecate the $as option

christyjacob4:
## What does this PR do?
The storage controller uses $as param for converting between PDF and text files. This is being deprecated since it is not secure. 

## Test Plan
N/A

## Related PRs and Issues
None

Yes


eldadfux:
🚀

Issue #307, opened 2020-04-01
Appwrite 'Cloud Functions'


eldadfux:
Adding cloud functions is one of the most requested Appwrite features to date. Before we finalize the feature spec, we would love to get as many community feedback and suggestions as possible.

## Goal

The goal of this feature is to create a sandbox where user-generated code can execute both on-demand and on existing system event (user.create, document.update, file.create, etc...). Each sandbox will be provided with both event-specific data and user provider env vars. This feature will help users trying to avoid the overhead of setting up a full backend server alongside Appwrite for specific custom tasks.

## Cross Platform & Language Agnostic

On of Appwrite main goals is to stay cross-platform and framework agnostic. We treat all tech stacks equally, so this feature should allow us to run code from multiple languages. We would probably support Node.js, PHP, Ruby, and Python first, but potentially any language can be supported.

## Security & Isolation

We are going to use Docker and Appwrite built-in queuing engine to execute the tasks in a completely and isolated environment.

## Performance & Scalability

To make sure performance is optimized, we will review the possibility to pre-download the core docker images so run time should only be depended on the code running.

All tasks will run in async way using our job workers. This will also allow us to scale very easily by adding new workers as demand grow.

## Abuse Protection

We should possibly also create some sort of a timeout to help protect smaller Appwrite instances running on a single server from being abused.

## Packaging

This is one of the bigger question marks. Since we need to support multiple languages and make sure the final package can be easily executed using a base Docker image, we need to allow developers to package and upload their code easily.

Some open questions:
* Should we create a CLI for uploading code packages?
* Can we create a single CLI that will be able to deal with different coding languages? (Should we use Docker to package the CLI?)
* Should we integrate with VCS directly? (can be very time consuming for v1 of this feature)
* Should we support version control?
* How do other cloud function services manage code uploading?
* Should we allow built-in IDE in Appwrite dashboard for simple functions?
* Should we store the code in an encrypted way? How will this influence performance and time-to-start?

TorstenDittmann:
Then I'll add my mustard, as we say in Germany.

What bothers me about most cloud functions is the unnecessary complexity. In my experience, the requirements for most cloud functions are relatively simple.

Maybe it is because I try not to push complex functions to serverless or cloud functions.

## Languages
To simplify maintainability, I would advise to stick with Node.js, PHP, Ruby, and Python only. These are probably the most common languages in the environment, except Java maybe.

## Because of CLI
[...]

PR #788, opened 2020-12-29
Enabled client endpoints for CF

eldadfux:
## What does this PR do?

Allow cloud functions to be executed from the client side by authorized users.

## Test Plan

Added new tests to simulate client execution of cloud funtions.

## Related PRs and Issues

#307 

Yes.


christyjacob4:
@eldadfux by client, do you mean the server SDK ?

app/workers/functions.php
```diff
@@ -507,7 +505,7 @@ public function cleanup(): void
                 $stdout = '';
                 $stderr = '';
 
-                if(Console::execute("docker stop {$first['name']}", '', $stdout, $stderr, 30) !== 0) {
+                if(Console::execute("docker rm -f {$first['name']}", '', $stdout, $stderr, 30) !== 0) {
```
christyjacob4, line 508: @eldadfux  we should probably stop the container before we remove it else `docker rm -f` would throw an error.

christyjacob4 (changes requested):
Looks good 👍

app/workers/functions.php
eldadfux: The -f option allow us to remove containers without needing to stop them. Its just work faster than sending 2 CLI commands.

eldadfux:
> @eldadfux by client, do you mean the server SDK ?

More like opening cloud function API endpoints to be available from client side apps, mainly for executing pre-configured functions.

christyjacob4:
If it is a client side app, Im confused by the usage of an API key in the request.
https://github.com/appwrite/appwrite/pull/788/files#diff-4a6c9e676aa23dca41808a3da457256a06c81ab23512ad7ced740696fd11541cR53

eldadfux:
> If it is a client side app, Im confused by the usage of an API key in the request.
> https://github.com/appwrite/appwrite/pull/788/files#diff-4a6c9e676aa23dca41808a3da457256a06c81ab23512ad7ced740696fd11541cR53

Only the execution routes are available to the client (/v1/functions/[id]/executions). I have to create a function using the server API before I can execute it on the client.

christyjacob4:
Okay makes sense now 👍  Thank you

app/workers/functions.php
christyjacob4: Okay awesome 👍

christyjacob4 (approved):

TorstenDittmann (commented):
Was a big one, but didn't see anything wrong 👍 

Great job!

Edit: Well, didn't see it was already merged 😆

Issue #477, opened 2020-07-13
Add Select All Checkbox for on Console API key Scopes Screen.


punitdiwan:
Please add Select All Checkbox for on API key Scopes Screen. We have now 13 scopes and will definitely increase in near future. It will be good to have **Select All Checkbox**  
for API key Scopes Screen in Console.

Thanks
Punit

eldadfux:
Work in progress:
![image](https://user-images.githubusercontent.com/1297371/88369815-80a1d000-cd99-11ea-8cd4-72265a92faac.png)

eldadfux:
@punitdiwan thanks for the idea, this feature is now available on v0.7 🥳

punitdiwan:
@eldadfux You are always Welcome

PR #812, opened 2021-01-10
Fix select all button bug

eldadfux:
## What does this PR do?

Fixed bug where select/unselect all button had weird behavior.

## Test Plan

N/A

## Related PRs and Issues

#477 

Yes


src/Appwrite/Utopia/Response.php
```diff
@@ -257,8 +258,7 @@ public function dynamic(Document $document, string $model): void
             $item = self::getFilter()->parse($output, $model);
         }
 
-        $this->json($output);
-
+        $this->json(!empty($output) ? $output : new stdClass());
```
christyjacob4, line 261: Can we not do the following here 
```suggestion
        $this->json($output ?? new stdClass());
```

christyjacob4 (changes requested):
Rest Looks good @eldadfux

src/Appwrite/Utopia/Response.php
eldadfux: I wasn't 100% sure so I doubled check, and the output is not the same for both lines. ?? is similar to `isset` and check if the assoc array has a specific key. Empty will also return true for empty string or arrays.

This is the code I used for the test:

```php
$output = [];

echo "\n";

echo '1:'.json_encode(!empty($output) ? $output : new \stdClass());

echo "\n";

echo '2:'.json_encode($output ?? new \stdClass());
```

results:
```php
// 1:{}
// 2:[]
```

christyjacob4: okay eldad 👍

PR #818, opened 2021-01-12
Fix for console custom domain

eldadfux:
## What does this PR do?

Fixed a bug in 0.7 where the console project don't detect the custom console domain as it should.

## Test Plan

N/A

## Related PRs and Issues

N/A

Yes


eldadfux:
Instead of adding the current hostname used for the console as a whitelisted platform on the API init callback, I have moved it to the callback that creates the allowed clients list for different validators in our API routes. This makes the current hostname be available in all app scopes.

lohanidamodar (approved):
Hey, everything looks good toe me!

PR #823, opened 2021-01-14
Feat new public permissions

eldadfux:
## What does this PR do?

Minor code refactoring. Changing role names from numbers to meaningful names that could be used to set new permission for Appwrite resources (documents, collections, files and more) both by end users and internally. I'll later update the docs here: https://appwrite.io/docs/permissions with the new changes. The first new permission we'll introduce will be: `role:guest`, `role:member`, and `role:app`. Let me know if naming conventions are good enough and consisted with other parts of the system.

## Test Plan

Updated the database tests to support this changes. No breaking changes for end-users as this roles were never publicly introduced.

## Related PRs and Issues

None

Yes


eldadfux:
@TorstenDittmann note that I have move the `Background Jobs` classes initing from the general controller to the api controller because they are not relevant for the web based controller (home & console controllers).

TorstenDittmann (approved):

PR #860, opened 2021-01-30
Fix migration comparison bug

eldadfux:
## What does this PR do?

Fix for comparison bug not detecting changes in object keys thinking both are the same. This was specifically detected on the `passwordUpdate` migration. The value stayed the same and only its key was changed. The final object structure was correct, but the save never happened because the comparison was only looking for changes in values, while ignoring keys. 

Yes


TorstenDittmann (approved):
Good catch 👏

Issue #547, opened 2020-09-21
Appwrite [Lite / Slim / Mini / Zero]


eldadfux:
We need to design a single Docker container that can run all of the Appwrite services and internal processes using Supervisord. This should be a feature with Appwrite minimum, must-have features (ClamAV disabled). This is something we need to allow developers to run Appwrite easily on low-end systems and 1-click deployment services (Raspberry PI, Free hosting tiers, Heroku, KintoHub, and others).

The new container can extend/copy the Dockerfile we're using in branch `0.7.x` and should add on top of its installations for all our 3rd party containers and run all internal process (scheduler + workers) as shown in the docker-compose file: https://github.com/appwrite/appwrite/blob/e4ff592eacc29000d04d0ba8c433b25e8a28c9bd/docker-compose.yml

This issue should be managed as a standalone repo that we'll later merge under the `appwrite` namespace.

qt-sergal:
Hi! Can I take this issue? Would be glad to start contributing to Appwrite and this issue looks like a great beginning.

eldadfux:
Yeh sure @qt-sergal. I have assigned you with the issue. I'll open a separate repo for you to submit a PR with the new Dockerfile. Please reach out to me on Discord if you have any question.

https://appwrite.io/discord

qt-sergal:
Okay! Great, I will start working on it right away.

eldadfux:
Great! this is the new repo: https://github.com/appwrite/lite the name is temporary as we want to put it to a community vote :)

Jacimartini98:
Appwrite

Jacimartini98:
Write

ritwikchakraborty123:
Sir I also want to contribute or may be collaborate with other contributer

Arunoday-bhattacharjee:
Hey I would also like to contribute to this issue . Can I help in any way?

eldadfux:
Hey @qt-sergal, would be great if you could post an update about your progress with this issue. 🙏🏻

qt-sergal:
Hi! Sorry, I am already done working on it, just got busy and wasn't sure if i should commit things part-by-part or all at once.
Will create a pull request to the other repository later today.

eldadfux:
@qt-sergal That's great, looking forward!

eldadfux:
Hey @qt-sergal any updates? sorry for nudging...

qt-sergal:
[...]

Issue #461, opened 2020-06-28
One-click Deploy to Heroku


TJMusiitwa:
Hey there, while the install and deploy with Docker is great. I have been having issues with the Docker installation on my computer so if there was support for a 1-click deploy to Heroku that would be a great developer experience.
Or maybe I am just asking something that is a bit-far fetched.
Thanks

eldadfux:
Actually we were just talking about working on something like this. @PedroCisnerosSantana do you have any updates on this or any estimated time table?

PedroCisnerosSantana:
Hey @eldadfux I started working on this some days ago and had some problems, still trying to solve them, will keep you up to date in Discord

eldadfux:
@PedroCisnerosSantana great, let me know if I can help.

PedroCisnerosSantana:
@eldadfux sure, I will contact you if not today, tomorrow :D

BananaEgg:
I'm also trying to deploy to Heroku,
I need to know what container is the front end so I can set the process type to web.

eldadfux:
@BananaEgg We use Traefik as the entrypoint on a "normal" setup: https://appwrite.io/docker-compose.yml

You can also try and use the Appwrite container on port 80 and 443. Traefik is mainly there for the auto-ssl certificates.

BananaEgg:
Iv'e deployed it following this: https://devcenter.heroku.com/articles/container-registry-and-runtime
meaning that i cloned Appwrite locally and created an app with it.
Now i'm getting this error " Error R10 (Boot timeout) -> Web process failed to bind to $PORT within 60 seconds of launch"

I understand that the issue is because there's no option to bind with the environment variable port.
I found the format in their documentation: https://help.heroku.com/PPBPA231/how-do-i-use-the-port-environment-variable-in-container-based-apps

I'm quite new to dockers and I'm not sure how to implement it without braking anything else.
This is the current format " 
command:
      - --log.level=DEBUG
      - --api.insecure=true
      - --providers.file.directory=/storage/config
      - --providers.file.watch=true
      - --providers.docker=true
      - --entrypoints.web.address=:80
      - --entrypoints.websecure.address=:443
      - --accesslog=true
    restart: unless-stopped
    ports:
      - 80:80
      - 443:443
      - 8080:8080
" 
This is what they are expecting - For example, instead of this:

PORT=8080
EXPOSE $PORT
CMD gunicorn --bind 0.0.0.0:$PORT wsgi

[...]

PR #863, opened 2021-01-31
Feat redis connection password

lohanidamodar:
## What does this PR do?

This PR introduces _APP_REDIS_AUTH environment variable, whose value can be (user:password) for a redis server. If user is empty and only password is available _APP_REDIS_AUTH should still be (:password)

## Test Plan

Existing tests will cover

## Related PRs and Issues

Its related to #461 , #547 

Yes


app/config/variables.php
```diff
@@ -133,6 +133,14 @@
                 'required' => false,
                 'question' => '',
             ],
+            [
+                'name' => '_APP_REDIS_AUTH',
```
eldadfux, line 137: We better keep consistency with other connector and use 2 vars instead of one. _APP_REDIS_USER and _APP_REDIS_PASS

app/init.php
```diff
@@ -90,9 +90,12 @@
 Config::load('storage-inputs', __DIR__.'/config/storage/inputs.php'); 
 Config::load('storage-outputs', __DIR__.'/config/storage/outputs.php'); 
 
-Resque::setBackend(App::getEnv('_APP_REDIS_HOST', '')
-    .':'.App::getEnv('_APP_REDIS_PORT', ''));
-
+$auth = App::getEnv('_APP_REDIS_AUTH',null);
+if($auth != null) {
```
eldadfux, line 94: Not sure default value is actually null. Docker might treat it as an empty string.

app/init.php
```diff
@@ -175,6 +178,15 @@ function($value) {
 $register->set('cache', function () { // Register cache connection
     $redis = new Redis();
     $redis->pconnect(App::getEnv('_APP_REDIS_HOST', ''), App::getEnv('_APP_REDIS_PORT', ''));
+    $auth = App::getEnv('_APP_REDIS_AUTH',null);
```
eldadfux, line 181: Use 2 env vars, if one of them is not empty use authentication.

app/init.php
lohanidamodar: oh, got it!

app/config/variables.php
lohanidamodar: Ok!

app/init.php
```diff
@@ -90,9 +90,10 @@
 Config::load('storage-inputs', __DIR__.'/config/storage/inputs.php'); 
 Config::load('storage-outputs', __DIR__.'/config/storage/outputs.php'); 
 
-$auth = App::getEnv('_APP_REDIS_AUTH',null);
-if($auth != null) {
-    Resque::setBackend('redis://'.App::getEnv('_APP_REDIS_AUTH', '').'@'.App::getEnv('_APP_REDIS_HOST', '').':'.App::getEnv('_APP_REDIS_PORT', ''));
+$user = App::getEnv('_APP_REDIS_USER','');
+$pass = App::getEnv('_APP_REDIS_PASS','');
+if(!empty($pass)) {
```
eldadfux, line 95: We should allow authentication if user *or* password have been set.

app/init.php
```diff
@@ -178,13 +179,13 @@ function($value) {
...
     $redis = new Redis();
     $redis->pconnect(App::getEnv('_APP_REDIS_HOST', ''), App::getEnv('_APP_REDIS_PORT', ''));
-    $auth = App::getEnv('_APP_REDIS_AUTH',null);
-    if($auth != null) {
-        $auth = explode(':',$auth);
-        if(!empty(trim($auth[0]))) {
-            $redis->auth($auth);
+    $user = App::getEnv('_APP_REDIS_USER','') ?? '';
+    $pass = App::getEnv('_APP_REDIS_PASS','') ?? '';
+    if(!empty($pass)) {
```
eldadfux, line 184: We should allow authentication if user or password have been set.

No need for `??` check. `getEnv` already takes care of that internally and return a default value (second param).

bin/schedule
```diff
@@ -1,10 +1,10 @@
 #!/bin/sh
 
-if [ -z "$_APP_REDIS_AUTH" ]
+if [ -z "$_APP_REDIS_PASS" ]
```
eldadfux, line 3: User or password

app/init.php
lohanidamodar: Updated!

app/init.php
lohanidamodar: Updated

app/init.php
lohanidamodar: updated

bin/schedule
lohanidamodar: updated

lohanidamodar:
@eldadfux Tests are green :smile:

eldadfux:
Awesome work @lohanidamodar 🥳

PR #873, opened 2021-02-02
Feat docker login from fucntions

eldadfux:
## What does this PR do?

Added docker login for the Appwrite cloud functions worker to make sure we don't reach docker pulls limit.

## Test Plan

Leverage existing tests.

## Related PRs and Issues

#307 

Yes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ install:
 script:
 - docker ps
 - docker-compose logs appwrite
+- docker-compose logs appwrite-worker-functions
 - docker-compose exec appwrite doctor
 - docker-compose exec appwrite vars
 - docker-compose exec appwrite test
--- a/app/workers/functions.php
+++ b/app/workers/functions.php
@@ -27,6 +27,17 @@ $environments = Config::getParam('environments');
  */
 $warmupStart = \microtime(true);
 
+$dockerUser = App::getEnv('DOCKERHUB_PULL_USERNAME', null);
+$dockerPass = App::getEnv('DOCKERHUB_PULL_PASSWORD', null);
+
+if($dockerUser) {
+    $stdout = '';
+    $stderr = '';
+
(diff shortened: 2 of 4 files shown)


TorstenDittmann:
Wouldn't this be extremely vulnerable for user to create a PR and expose our Docker credentials with this setup?

eldadfux:
I would probably choose to avoid this if we had a choice, not sure we do.

Anyway Travis is doing some effort to avoid this data from being exposed, but obviously we have to do some manual checking before accepting any potential malicious PR.

> By default, the value of these new environment variables is hidden from the export line in the logs. This corresponds to the behavior of encrypted variables in your .travis.yml. The variables are stored encrypted in our systems, and get decrypted when the build script is generated.

> Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository.

> As an alternative to the web interface, you can also use the CLI’s env command.

lohanidamodar (approved):
LGTM. I don't understand though how one can reverse engineer the credentials with this if data is encrypted in hidden in Travis. Does it expose anywhere else?

TorstenDittmann:
Can't I just add

```
- DOCKERHUB_PULL_USERNAME
- DOCKERHUB_PULL_PASSWORD
```

to my Appwrite container and insert following assert in a test:

```
$login = App::getEnv('DOCKERHUB_PULL_USERNAME', null).' - '.App::getEnv('DOCKERHUB_PULL_PASSWORD', null)
$this->assertEquals("thisFailsOnPurpose", $login);
```

And so the Travis logs will expose our password 👍🏻 

Do

TorstenDittmann (approved):

eldadfux:
Anyway, we should make sure no incoming PR is trying print this vars is accepted.
@christyjacob4 @TorstenDittmann @lohanidamodar

PR #739, opened 2020-11-21
Feat 730 disable smtp server

eldadfux:
## What does this PR do?

Adds option to disable all Appwrite system emails by passing an empty SMTP hostname value.

## Test Plan

Use existing tests.

## Related PRs and Issues

#730 

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

Yes.

PR #935, opened 2021-02-22
Feat disable SMTP by Default

eldadfux:
## What does this PR do?

Having a dummy SMTP server with the default setup is confusing. Appwrite can work well with no SMTP server attached and only a limited amount of features are disabled. We already allowed to completely disable the SMTP functionality in PR #739 .

In this PR we use that new feature by default. I also added a note on the dashboard and blocked the ability to send team members invites without configuring an SMTP server. We should follow up with this issue and update the email delivery page on the [documentation](https://appwrite.io/docs/email-delivery).  

## Test Plan

N/A

## Related PRs and Issues

#739 

Yes.


eldadfux:
For the record, this is the UI change:
image

TorstenDittmann (approved):

PR #939, opened 2021-02-23
Fixed XSS in the console

eldadfux:
## What does this PR do?

Fixed XSS in the console

## Test Plan

N/A

## Related PRs and Issues

N/A

Yes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,6 +26,10 @@
 
 - Updated missing storage env vars
 
+## Security
+
+- Fixed an XSS vulnerability in the Appwrite console
+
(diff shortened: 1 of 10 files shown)


TorstenDittmann (approved):

hack3r-0m:
it would be great if somebody can confirm fix against other fields too, for e.g update name, create api key field, etc.

eldadfux:
> it would be great if somebody can confirm fix against other fields too, for e.g update name, create api key field, etc.

Yeh I did that. It seems the problem was in a centralize place, so one fix took care to all different occurrences we had.

PR #948, opened 2021-03-03
Feat 265 - Realtime Implementation

TorstenDittmann:
## What does this PR do?

Adds a Realtime Endpoint supporting WebSocket protocol as described here: https://github.com/appwrite/rfc/pull/17
This endpoint will let you subscribe to specific channels and will send you realtime updates.

## Test Plan

This PR includes all the Unit-Tests. E2E tests can be found #949 here.

## Related PRs and Issues

#948 
https://github.com/appwrite/appwrite/pull/692
https://github.com/appwrite/appwrite/issues/265
https://github.com/appwrite/appwrite/issues/509
https://github.com/appwrite/rfc/pull/17

✅


app/realtime.php
```diff
@@ -0,0 +1,304 @@
...
+    $server->push($connection, json_encode($channels));
+});
+
+$server->on('message', function (Server $server, Frame $frame) {
+    $server->push($frame->fd, 'Sending messages is not allowed.');
+    $server->close($frame->fd);
+});
+
+$server->on('close', function (Server $server, int $fd) use (&$connections, &$subscriptions) {
+    Realtime::unsubscribe($fd, $subscriptions, $connections);
```
eldadfux, line 300: Let's stick with `$connection` instead of `$fd`? (or the other way around)

src/Appwrite/Database/Adapter/Redis.php
```diff
@@ -261,7 +268,7 @@ public function getDebug()
      */
     protected function getRedis(): Client
     {
-        return $this->register->get('cache');
+        return $this->isPool ? $this->register->get('redis')->get() : $this->register->get('cache');
```
eldadfux, line 271: Reminder to drop this and use one method.

src/Appwrite/Realtime/Realtime.php
```diff
@@ -0,0 +1,193 @@
...
+                        foreach (array_keys($subscriptions[$event['project']][$role][$channel]) as $ids) {
+                            $receivers[] = $ids;
+                        }
+                        break;
+                    }
+                }
+            }
+        }
+
+        return array_keys(array_flip($receivers));
```
eldadfux, line 113: I think iterating the `$receivers` array once and building the list might be faster.

src/Appwrite/Realtime/Realtime.php
```diff
@@ -0,0 +1,193 @@
+<?php
```
eldadfux, line 1: I wondering if it will make sense to combine both Realtime classes into one class. I don't see any major concept differentiation between them.

src/Appwrite/Realtime/Realtime.php
```diff
@@ -0,0 +1,193 @@
...
+    /**
+     * Adds Subscription. 
+     * 
+     * @param string $projectId
+     * @param mixed $connection
+     * @param array $subscriptions
+     * @param array $roles
+     * @param array $channels
+     */
+    static function subscribe($projectId, $connection, $roles, &$subscriptions, &$connections, &$channels)
```
eldadfux, line 125: Do we have any results for testing on small vs large amount of data?
Might be useful to list them here for future reference.

eldadfux: Same for the unsubscribe.

eldadfux (changes requested):
+ This `docs/specs/realtime.drawio.svg` diagram is probably not very accurate anymore

src/Appwrite/Realtime/Realtime.php
TorstenDittmann: I tried that first and this approach was faster for me.

I just add all the receivers to an array, without checking for duplicates. In the end I remove all duplicates at one time and return it.

app/realtime.php
```diff
@@ -205,6 +208,21 @@
         Realtime::subscribe($project->getId(), $connection, $roles, $subscriptions, $connections, $channels);
 
         $server->push($connection, json_encode($channels));
+
```
eldadfux, line 211: We should probably return the connection to the pull on an exception as well. What about the HTTP server do we want to try and fully leverage the connection pool for both APIs?

app/http.php
```diff
@@ -31,6 +31,13 @@
     'validateCNAME' => false,
 ]);
 
+$register->set('cache', function () use ($register) { // Register cache connection
```
eldadfux, line 34: If we keep this set callback, we can just replace the original one.

app/init.php
```diff
@@ -165,7 +165,7 @@ function($value) {
             PDONative::ATTR_ERRMODE => PDONative::ERRMODE_EXCEPTION,
         ]);
 
-    $pool = new PDOPool($config, 4096); // TODO: Investigate pool size
+    $pool = new PDOPool($config, 16384); // TODO: Investigate pool size
```
eldadfux, line 168: Did we open an issue about this on the Swoole repo?

app/http.php
TorstenDittmann: The original one in `init.php` is used by the workers 👍🏻

app/realtime.php
TorstenDittmann: Definitely should be in the exception 👍🏻 had it like this for testing.

I still not 100% understand the behavior behind the Connection Pools. Will make this part of the Issue I will raise over at the Swoole repo over the weekend 🙂

app/init.php
TorstenDittmann: Not yet, see previous comment.

src/Appwrite/Realtime/Realtime.php
TorstenDittmann: Okay, understood what you meant. Improved now 👍🏻

app/controllers/api/teams.php
```diff
@@ -90,6 +92,10 @@
             }
         }
 
+        if (!empty($user->getId())) {
```
eldadfux, line 95: What is the cause behind this addition?

app/init.php
```diff
@@ -398,10 +418,10 @@ function($value) {
             $request->getCookie(Auth::$cookieName.'_legacy', '')));// Get fallback session from old clients (no SameSite support)
 
     // Get fallback session from clients who block 3rd-party cookies
-    $response->addHeader('X-Debug-Fallback', 'false');
+    if($response) $response->addHeader('X-Debug-Fallback', 'false');
```
eldadfux, line 421: Let's add a comment explaining this condition for future reference.

src/Appwrite/Database/Pool.php
```diff
@@ -1,11 +1,12 @@
 <?php
```
eldadfux, line 1: I guess we don't need this file anymore.

src/Appwrite/Event/Realtime.php
```diff
@@ -0,0 +1,228 @@
+<?php
```
eldadfux, line 1: Big parts here might be a good fit for the Messaging library.

src/Appwrite/Utopia/Response/Model/Execution.php
```diff
@@ -16,6 +16,13 @@ public function __construct()
                 'default' => '',
                 'example' => '5e5ea5c16897e',
             ])
+            ->addRule('$permissions', [
```
eldadfux, line 19: Why is the reason behind this addition?

eldadfux (commented):
Reviewed everything except of the tests classes, waiting for the integration with the messaging lib for another review 👍

src/Appwrite/Utopia/Response/Model/Execution.php
TorstenDittmann: I need to have some type of `permissions` value seeing who is to receive an event for Executions 🙂

app/controllers/api/teams.php
TorstenDittmann: If a user is creating a team, I need his user id to re-fetch his roles 👍🏻

src/Appwrite/Event/Realtime.php
TorstenDittmann: Got this class removed 👏🏻

app/init.php
TorstenDittmann: Actually not sure anymore where this change comes from.

Do you remember? https://github.com/appwrite/appwrite/commit/619781a4ad567b6def0add82e8a9fcf9cb48f83a

rlee1990:
@TorstenDittmann I have a question about the WebSocket is there a reason why it has to run from a fork? I have other packages that also use the WebSocket and so I can't test this with that.

app/controllers/api/projects.php
```diff
@@ -260,11 +262,34 @@
                         'date' => \strtotime($point['time']),
                     ];
                 }
+
+                // Realtime Connections
+                $result = $database->query('SELECT sum(value) AS "value" FROM "appwrite_usage_realtime_clients" WHERE time > \''.$start.'\' AND time < \''.$end.'\' AND "metric_type"=\'counter\' AND "project"=\''.$project->getId().'\' GROUP BY time('.$period[$range]['group'].') FILL(null)');
```
christyjacob4, line 267: These queries will need to be part of the usage worker i believe . Are we planning to do this refactor after the release?

app/controllers/general.php
```diff
@@ -240,33 +240,22 @@
...
 
+            Authorization::setRole('role:'.Auth::USER_ROLE_APP);
             Authorization::setDefaultStatus(false);  // Cancel security segmentation for API keys.
         }
     }
 
-    if ($user->getId()) {
-        Authorization::setRole('user:'.$user->getId());
+    foreach (Auth::getRoles($user) as $authRole) {
+        Authorization::setRole($authRole);
```
christyjacob4, line 249: Is there any effect of iterating over this list and using `setRole` over and over again ? 

Isn't this the same as doing `Authorization::setRole(end(Auth::getRoles($user)))`

TorstenDittmann: I removed code later that would iterate anyway, I just optimized it here 👍🏻

app/controllers/api/projects.php
TorstenDittmann: Yes, once 0.10.0 is out - I will need to refactor lots of stuff to adapt to the DB refactor and new usage.

src/Appwrite/Messaging/Adapter/Realtime.php
```diff
@@ -0,0 +1,313 @@
...
+     *          [CHANNEL_NAME_Z] -> [CONNECTION_ID]
+     *      [ROLE_Y] -> 
+     *          [CHANNEL_NAME_X] -> [CONNECTION_ID]
+     *          [CHANNEL_NAME_Y] -> [CONNECTION_ID]
+     *          [CHANNEL_NAME_Z] -> [CONNECTION_ID]
+     */
+    public array $subscriptions = [];
+
+    /**
+     * Adds a subscribtion.
```
christyjacob4, line 38:
```suggestion
     * Adds a subscription.
```

app/tasks/maintenance.php
```diff
@@ -39,6 +39,14 @@ function notifyDeleteAuditLogs(int $interval)
             ]);
         }
 
+        function notifyDeleteConnections() 
+        {
+            Resque::enqueue(Event::DELETE_QUEUE_NAME, Event::DELETE_CLASS_NAME, [
+                'type' => DELETE_TYPE_REALTIME,
+                'timestamp' => time() - 60
```
christyjacob4, line 46: Is this the expected behaviour ? 
"Every _APP_MAINTENANCE_INTERVAL ( or 86400 ) seconds we delete realtime connections older than 60 seconds?  "

christyjacob4 (changes requested):
Looks Good @TorstenDittmann

app/tasks/maintenance.php
TorstenDittmann: Yes, this will just clean up unused data that is used to have concurrent connection stats across all containers.

app/init.php
eldadfux: Maybe for realtime, workers or CLI where we use this init, but don't have a request

PR #974, opened 2021-03-12
Feat graphql support

christyjacob4:
## What does this PR do?

Add GraphQL support

## Test Plan

New tests covering all GraphQL queries/mutations

## Related PRs and Issues

https://github.com/appwrite/docs/pull/287
https://github.com/appwrite/homepage/pull/113
https://github.com/appwrite/sdk-generator/pull/577

Yes


src/Appwrite/Utopia/Response.php
```diff
@@ -342,10 +349,16 @@ public function __construct(SwooleHTTPResponse $response)
...
+    public function getSwoole(): SwooleHTTPResponse
+    {
+        return $this->swoole;
+    }
+
     /**
      * HTTP content types
      */
     public const CONTENT_TYPE_YAML = 'application/x-yaml';
+    public const CONTENT_TYPE_NULL = 'null';
```
eldadfux, line 361: What is content type `null`? This doesn't seem HTTP compliant, how do we use it?

eldadfux (changes requested):
The PR looks great. I'm already leaving some of my comments, but I'm yet to review the controller, api.php, init.php and src/Appwrite/GraphQL.

app/controllers/api/databases.php
```diff
@@ -2567,6 +2565,7 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
     });
 
 App::get('/v1/databases/:databaseId/usage')
+    ->alias('/v1/database/usage', ['databaseId' => 'default'])
```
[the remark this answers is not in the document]
abnegate: Can't remember 🤔 not necessary though, will remove it

docker-compose.yml
```diff
@@ -176,6 +176,9 @@ services:
       - _APP_MAINTENANCE_RETENTION_AUDIT
       - _APP_SMS_PROVIDER
       - _APP_SMS_FROM
+      - _APP_GRAPHQL_MAX_COMPLEXITY
+      - _APP_GRAPHQL_MAX_DEPTH
+      - PHP_IDE_CONFIG=serverName=swoole-http
```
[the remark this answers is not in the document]
abnegate: This is used to allow debugging from PHPStorm as the existing setup steps were insufficient, see [here](https://github.com/appwrite/appwrite/blob/feat-graphql-support/CONTRIBUTING.md#phpstorm-setup). It is outside the scope though so I can remove or move to separate PR if required

src/Appwrite/Utopia/Request.php
```diff
@@ -27,6 +27,11 @@ public function __construct(SwooleRequest $request)
         parent::__construct($request);
     }
 
+    public function getSwoole(): SwooleRequest
```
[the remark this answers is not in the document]
abnegate: This is being used to manipulate the initial request to redirect to the REST API. Most of the uses could be moved into `set` methods on the request base class. 

There is one usage where I'm not sure how to avoid using this; when [creating a new request object](https://github.com/appwrite/appwrite/blob/575b882b306099309a25cac9ebb7a0f1568efc14/src/Appwrite/GraphQL/Resolvers.php#L299), as the constructor requires the swoole request. The new request is needed so the initial GraphQL request is not ended when the REST request completes.

src/Appwrite/Utopia/Response.php
abnegate: This is used [here](https://github.com/appwrite/appwrite/blob/feat-graphql-support/src/Appwrite/Utopia/Response.php#L442) to allow the inner REST response to generate a payload, but not close the GraphQL request by sending the output

src/Appwrite/Utopia/Request.php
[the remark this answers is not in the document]
abnegate: Managed to remove this by adding some `set` methods to the `Request` object (in utopia-php/framework), and adding a `clone` function in this class to avoid having to get the swoole request to create a new one.

app/controllers/general.php
```diff
@@ -382,7 +382,7 @@
         $response
             ->addHeader('Server', 'Appwrite')
             ->addHeader('Access-Control-Allow-Methods', 'GET, POST, PUT, PATCH, DELETE')
-            ->addHeader('Access-Control-Allow-Headers', 'Origin, Cookie, Set-Cookie, X-Requested-With, Content-Type, Access-Control-Allow-Origin, Access-Control-Request-Headers, Accept, X-Appwrite-Project, X-Appwrite-Key, X-Appwrite-Locale, X-Appwrite-Mode, X-Appwrite-JWT, X-Appwrite-Response-Format, X-SDK-Version, X-SDK-Name, X-SDK-Language, X-SDK-Platform, X-Appwrite-ID, Content-Range, Range, Cache-Control, Expires, Pragma, X-Fallback-Cookies')
+            ->addHeader('Access-Control-Allow-Headers', 'Origin, Cookie, Set-Cookie, X-Requested-With, Content-Type, Access-Control-Allow-Origin, Access-Control-Request-Headers, Accept, X-Appwrite-Project, X-Appwrite-Key, X-Appwrite-Locale, X-Appwrite-Mode, X-Appwrite-JWT, X-Appwrite-Response-Format, X-SDK-Version, X-SDK-Name, X-SDK-Language, X-SDK-Platform, X-SDK-GraphQL, X-Appwrite-ID, Content-Range, Range, Cache-Control, Expires, Pragma, X-Fallback-Cookies')
```
eldadfux, line 385: How do we report this header? From which client?

app/controllers/shared/api.php
```diff
@@ -6,12 +6,13 @@
 use Appwrite\Event\Delete;
 use Appwrite\Event\Event;
 use Appwrite\Event\Mail;
+use Appwrite\Extend\Exception;
```
eldadfux, line 9: Would it make sense to have this shutdown method in a dedicated controller or under the graphql controller?

app/tasks/sdks.php
```diff
@@ -1,5 +1,8 @@
 <?php
 
+use Appwrite\SDK\Language\GraphQL;
+use Appwrite\SDK\Language\HTTP;
```
eldadfux, line 4: Do we need both HTTP an REST?

composer.json
```diff
@@ -78,7 +78,7 @@
         }
     ],
     "require-dev": {
-        "appwrite/sdk-generator": "0.28.*",
+        "appwrite/sdk-generator": "dev-feat-graphql as 0.28.1",
```
eldadfux, line 81: Reminder to change the version number.

docker-compose.yml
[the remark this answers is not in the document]
eldadfux: Might make things a bit more organized.

src/Appwrite/Auth/Auth.php
```diff
@@ -29,7 +30,7 @@ class Auth
     ];
 
     public const DEFAULT_ALGO = 'argon2';
-    public const DEFAULT_ALGO_OPTIONS = ['memoryCost' => 2048, 'timeCost' => 4, 'threads' => 3];
+    public const DEFAULT_ALGO_OPTIONS = ['type' => 'argon2', 'memoryCost' => 2048, 'timeCost' => 4, 'threads' => 3];
```
eldadfux, line 33: Was this an issue in the code base? another PR?

src/Appwrite/Utopia/Request.php
```diff
@@ -9,48 +9,20 @@
...
-            case self::METHOD_POST:
-            case self::METHOD_PUT:
-            case self::METHOD_PATCH:
-            case self::METHOD_DELETE:
-                $requestParameters = $this->generateInput();
-                break;
-            default:
-                $requestParameters = (!empty($this->swoole->get)) ? $this->swoole->get : [];
-        }
+        $requestParameters = parent::getParams();
```
eldadfux, line 25: Seems valid, but would love a bit of explanation.

eldadfux, line 25:
```suggestion
        $parameters = parent::getParams();
```

Request is already part of the namespace, we shouldn't reuse it.

app/config/errors.php
```diff
@@ -549,4 +549,14 @@
         'description' => 'Domain verification for the requested domain has failed.',
         'code' => 401,
     ],
+    Exception::GRAPHQL_NO_QUERY => [
+        'name' => Exception::GRAPHQL_NO_QUERY,
+        'description' => 'Param "query" is not optional.',
+        'code' => 400,
+    ],
+    Exception::GRAPHQL_TOO_MANY_QUERIES => [
```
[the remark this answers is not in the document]
abnegate: This is controlled by the env var `_APP_GRAPHQL_MAX_BATCH_SIZE`, the default is currently set to 10.

Each of the individual queries is also limited by the var `_APP_GRAPHQL_MAX_COMPLEXITY` which defaults to 250. One field, `createdAt` for example, counts as 1 complexity point. For list endpoints it is multiplied by the number of items.

250 is the default due to the idea of a max of 10 list queries with the default `limit` of 25 items, per request.

app/controllers/api/databases.php
```diff
@@ -1753,15 +1752,14 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
         $indexes = $collection->getAttribute('indexes');
 
         // Search for index
-        $indexIndex = array_search($key, array_column($indexes, 'key'));
+        $indexIndex = array_search($key, array_map(fn($idx) => $idx['key'], $indexes));
```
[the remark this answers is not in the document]
abnegate: This can be reverted once https://github.com/appwrite/appwrite/pull/3712 is merged

app/controllers/api/databases.php
```diff
@@ -1753,15 +1752,14 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
...
 
         if ($indexIndex === false) {
             throw new Exception(Exception::INDEX_NOT_FOUND);
         }
 
-        $index = new Document([\array_merge($indexes[$indexIndex], [
-            'collectionId' => $database->getInternalId() . '_' . $collectionId,
-        ])]);
+        $index = $indexes[$indexIndex];
```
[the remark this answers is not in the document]
abnegate: Just seemed redundant to wrap the index in a new document instead of just setting the `collectionId`, but can be reverted once https://github.com/appwrite/appwrite/pull/3712 is merged

app/controllers/general.php
abnegate: The header is added to all SDKs here: https://github.com/appwrite/sdk-generator/blob/feat-graphql/src/Spec/Swagger2.php#L191-L193

This is to handle the requirement that SDK's need to send named parameters, but the controller gets the params from the raw body, to handle object or array requests. 

The controller reads the header and unpacks an SDK request here: https://github.com/appwrite/appwrite/blob/feat-graphql-support/app/controllers/api/graphql.php#L82-L84

app/controllers/shared/api.php
abnegate: Yep makes sense in the graphql controller to me

src/Appwrite/Auth/Auth.php
abnegate: This is to be able to resolve which GraphQL type the options represent when given only the options themselves, necessary for any union response type (currently only attributes and hashoptions): https://github.com/appwrite/appwrite/blob/feat-graphql-support/src/Appwrite/GraphQL/Types/Mapper.php#L354-L356

src/Appwrite/Utopia/Request.php
abnegate: This block was doing what the parent method was doing so seems cleaner to call that instead of repeating

composer.json
eldadfux: Another reminder.

composer.json
```diff
@@ -50,26 +50,27 @@
         "utopia-php/cli": "0.13.*",
         "utopia-php/config": "0.2.*",
         "utopia-php/database": "0.28.*",
-        "utopia-php/preloader": "0.2.*",
         "utopia-php/domains": "1.1.*",
         "utopia-php/framework": "0.25.*",
         "utopia-php/image": "0.5.*",
         "utopia-php/locale": "0.4.*",
-        "utopia-php/logger": "0.3.*",
+		"utopia-php/logger": "0.3.*",
```
eldadfux, line 57: Indentations is wrong.

app/init.php
```diff
@@ -1047,3 +1053,93 @@ function getDevice($root): Device
 
     return $languages;
 });
+
+App::setResource('promiseAdapter', function ($register) {
```
eldadfux, line 1057: I think this is very generic. Let's talk about it in our next sync and mark it as technical debt for now.

app/init.php
```diff
@@ -30,49 +28,53 @@
...
 use Utopia\Messaging\Adapters\SMS\Telesign;
 use Utopia\Messaging\Adapters\SMS\TextMagic;
 use Utopia\Messaging\Adapters\SMS\Twilio;
 use Utopia\Messaging\Adapters\SMS\Vonage;
 use Utopia\Registry\Registry;
-use MaxMind\Db\Reader;
-use PHPMailer\PHPMailer\PHPMailer;
-use Utopia\Cache\Adapter\Redis as RedisCache;
-use Utopia\Cache\Cache;
-use Utopia\Database\Adapter\MariaDB;
```
eldadfux, line 55: All of these were not used?!

app/controllers/api/databases.php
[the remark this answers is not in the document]
eldadfux: Please add a task to use the built-in function.

src/Appwrite/Promises/Promise.php
```diff
@@ -0,0 +1,194 @@
+<?php
+
+namespace Appwrite\Promises;
```
eldadfux, line 3: This should not belong in Appwrite. We should consider whether it should be a utopia library and avoid it for now. How much do we rely on this class?

src/Appwrite/Promises/Swoole.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+namespace Appwrite\Promises;
+
+use Swoole\Coroutine\Channel;
+
+class Swoole extends Promise
+{
```
eldadfux, line 8: What is this class responsible for? why is it part of the promise namespace?

CHANGES.md
```diff
@@ -27,14 +27,12 @@
 
 ## Bugs
 - Fix license detection for Flutter and Dart SDKs [#4435](https://github.com/appwrite/appwrite/pull/4435)
+- Fix missing realtime event for create function deployment [#4574](https://github.com/appwrite/appwrite/pull/4574)
```
eldadfux, line 30: Reminder to update the full change log.

eldadfux (commented):
- I'm not sure we need the GraphQL logo anymore since Console 2.0.
- Do we support disabling the GraphQL API? If yes, we should update the console settings.
- If one of the "regular" services is disabled, do we still allow queries through GraphQL, should we disable them?

app/init.php
abnegate: These were moved elsewhere by alphabetical formatting

src/Appwrite/Promises/Promise.php
abnegate: We need this to allow async query resolution. I've followed the pattern wherein the whole `Promises` directory could be it's own utopia library when we want to abstract it from Appwrite

src/Appwrite/Promises/Swoole.php
abnegate: This extends the promise base, using a swoole coroutine to execute the promise. Akin to extending the base adapter for a vendor specific implementation

abnegate:
> * I'm not sure we need the GraphQL logo anymore since Console 2.0.
> * Do we support disabling the GraphQL API? If yes, we should update the console settings.
>  * If one of the "regular" services is disabled, do we still allow queries through GraphQL, should we disable them?

- Removed
- We do support disabling it, will make an update in console
- If a service is disabled and called through GraphQL, the request will be blocked

src/Appwrite/Promises/Promise.php
eldadfux: Let's add it as a task in the backlog

eldadfux (approved):

app/config/platforms.php
```diff
@@ -135,21 +135,44 @@
...
-            //     'beta' => false,
-            //     'dev' => false,
-            //     'hidden' => false,
-            //     'family' => APP_PLATFORM_CLIENT,
-            //     'prism' => 'java',
-            //     'source' => false,
-            //     'gitUrl' => 'git@github.com:appwrite/sdk-for-java.git',
-            //     'gitRepoName' => 'sdk-for-java',
-            //     'gitUserName' => 'appwrite',
-            // ],
```
Meldiron, line 152: Was java commented out on purpose?

app/controllers/api/users.php
```diff
@@ -63,7 +63,7 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
             'status' => true,
             'password' => (!empty($password)) ? ($hash === 'plaintext' ? Auth::passwordHash($password, $hash, $hashOptionsObject) : $password) : null,
             'hash' => $hash === 'plaintext' ? Auth::DEFAULT_ALGO : $hash,
-            'hashOptions' => $hash === 'plaintext' ? Auth::DEFAULT_ALGO_OPTIONS : $hashOptions,
+            'hashOptions' => $hash === 'plaintext' ? Auth::DEFAULT_ALGO_OPTIONS : $hashOptionsObject + ['type' => $hash],
```
Meldiron, line 66: Why do we need type in hashOptions? the `hash` is the type

Meldiron: Also I think we need to apply this to account controller, and teams controller.

docker-compose.yml
```diff
@@ -174,6 +173,8 @@ services:
       - _APP_MAINTENANCE_RETENTION_USAGE_HOURLY
       - _APP_SMS_PROVIDER
       - _APP_SMS_FROM
+      - _APP_GRAPHQL_MAX_COMPLEXITY
+      - _APP_GRAPHQL_MAX_DEPTH
```
Meldiron, line 177: Should we also include `_APP_GRAPHQL_MAX_BATCH_SIZE`?

app/controllers/api/users.php
Meldiron: After reading more of the PR I see why we do this. But this now requires migration.

docker-compose.yml
eldadfux: I think we should, worst case we can revert later.

app/controllers/api/users.php
Meldiron: I am making myself note to double check migrations between 1.2.x and master

PR #1021, opened 2021-03-23
Move validators from utopia

kodumbeats:
## What does this PR do?

This PR involves a refactor of validators, decoupling Appwrite concerns from Utopia as a standalone framework.

This can be merged alongside its sister PR in Utopia repository.

## Test Plan

`phpunit` tests under `test/unit/`

## Related PRs and Issues

https://github.com/utopia-php/framework/pull/6

yes.


kodumbeats:
🩹 in https://github.com/appwrite/appwrite/pull/1021/commits/448276fcd485f3e58a65e4bca8a8bf5fb37be7db until https://github.com/utopia-php/framework/pull/6/commits/dea859bc5cab7df4b09edcc1d9a940945ade8e71 and https://github.com/utopia-php/framework/pull/6/commits/0b7934ab3a1e081f612f6385fa9100fe9ba191c1 ara available. Might be best to leave the patch in until https://github.com/utopia-php/framework/pull/6 is finished, and updates are pulled into this repo.

src/Appwrite/Network/Validator/Host.php
```diff
@@ -54,7 +54,7 @@ public function getDescription()
      */
     public function getType()
     {
-        return self::TYPE_STRING;
+        return 'string';
```
eldadfux, line 57: I think self should still work here. If not, you can try `static::`

src/Appwrite/Network/Validator/Email.php
```diff
@@ -10,7 +10,7 @@
  * @license The MIT License (MIT) <http://www.opensource.org/licenses/mit-license.php>
  */
```
eldadfux, line 11: We should remove the utopia license here.

src/Appwrite/Network/Validator/Host.php
kodumbeats: There's nothing to inherit in `appwrite/appwrite` right now because the type definitions are still in the utopia branch: https://github.com/utopia-php/framework/commit/0b7934ab3a1e081f612f6385fa9100fe9ba191c1

Am I understanding your feedback correctly?

viv718918 (approved):

viv718918 (approved):

src/Appwrite/Network/Validator/Host.php
eldadfux: Yep, sorry. Once we release the Utopia changes we can inherit the constants properly.

kodumbeats:
One unresolved issue - as a part of the refactor, we wanted to remove the `Domain` validator from Utopia to Appwrite. However, the validator exists in both repositories:

https://github.com/utopia-php/framework/blob/master/src/Validator/Domain.php
https://github.com/appwrite/appwrite/blob/master/src/Appwrite/Network/Validator/Domain.php

I left both untouched - which version should we stick with?

eldadfux:
@kodumbeats They are both basically **almost** the same. I would just pick the one used the most and make sure manually the other places are working as expected.

kodumbeats:
Utopia's Domain validator is completely unused in `appwrite/appwrite`, and Appwrite has its own validator/tests for domains. Removed from Utopia in:
https://github.com/utopia-php/framework/pull/6/commits/9147b070a2ab7ac57f83bcc4d7906ad67a233622

PR #1055, opened 2021-03-31
Refactor function docker exec

kodumbeats:
## What does this PR do?

PHP introduces overhead when calling shell functions, as it creates a shell as a part of the process. To accelerate function calls, we can communicate with the Docker API directly, circumventing all overhead caused by PHP `exec` or the Docker CLI. 

As of PHP 7.0.7, `cURL` can communicate with the Docker API through the Docker socket via `CURLOPT_UNIX_SOCKET_PATH`. 

## TODO

- [x] Confirm execution time is reduced
- [ ] Handle stdout/stderr from [container logs stream](https://docs.docker.com/engine/api/v1.41/#operation/ContainerLogs)
- [x] Obtain exit code from response
- [x] Create listener for event `exec_die` to time execution [not needed]
- [ ] Swoole coroutines?

## Test Plan

Existing test coverage

Yes.

### Is it faster?
 Right now, I'm passing the execution microtime timestamp into a function as an ENV var, doing math in the function, and echoing the execution time:

```php
$executionTime = [ // in seconds
  0.094936847686768,
  0.13563680648804,
  0.10708594322205,
  0.10107183456421,
  0.090137958526611,
  0.093377113342285,
  0.12929201126099,
  0.077444076538086,
  0.10612082481384,
]
```
Compared to the same script running on the old approach:
samescripttimes


TorstenDittmann:
Bildschirmfoto 2021-04-01 um 16 30 15

Speed on a Apple M1 👍🏻

eldadfux:
Please sync 0.8/dev for tests to pass

kodumbeats:
@eldadfux Could use your advice to move forward.

Now that we no longer use a shell to invoke `docker exec`, we also lose the shell formatting that we got as a side effect. Let's use this PHP script as an example (modified from our testing suite):

```php
<?php

echo $_ENV['APPWRITE_FUNCTION_ID']."\n";
echo $_ENV['APPWRITE_FUNCTION_NAME']."\n";
echo $_ENV['APPWRITE_FUNCTION_TAG']."\n";
echo $_ENV['APPWRITE_FUNCTION_TRIGGER']."\n";
echo $_ENV['APPWRITE_FUNCTION_ENV_NAME']."\n";
echo $_ENV['APPWRITE_FUNCTION_ENV_VERSION']."\n";
echo $_ENV['APPWRITE_FUNCTION_EVENT']."\n";
echo $_ENV['APPWRITE_FUNCTION_EVENT_PAYLOAD']."\n";
```

When executed on master, we get the following STDOUT:
```bash
606dffab03f82
testfunction
60748ca502422
http
PHP
8.0
null
null
```

However, on this branch, we no longer have a shell to parse the string, so the STDOUT from the Docker API stream differs:
```
'60747d3c9f6ad'
'testfunction'
'60748cb100474'
'http'
'PHP'
'8.0'
'null'
'null'
```

This is causing [tests to fail](https://github.com/appwrite/appwrite/blob/master/tests/e2e/Services/Functions/FunctionsCustomServerTest.php#L670-L677) with the following kinds of messages:
```
1) Tests\E2E\Services\Functions\FunctionsCustomServerTest::testENVS
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-''607434ad3246c'' 
+'607434ad3246c'
/usr/src/code/tests/e2e/Services/Functions/FunctionsCustomServerTest.php:688
```

### Unanswered questions
**Where's the best/most appropriate place to make this change for ✅ ?**
1) Strip quotes **in code** when we pack the raw stream data into STDOUT [here](https://github.com/appwrite/appwrite/pull/1055/files#diff-16d93434c8aa21e9cfcdcffdf3228784e5d4801476c10df941dbdc2383a5b9d6R527)
2) Strip quotes **in tests** when we explode STDOUT into an array [here](https://github.com/appwrite/appwrite/blob/master/tests/e2e/Services/Functions/FunctionsCustomServerTest.php#L670)
3) Handle extra quotes **in assertions** [here](https://github.com/appwrite/appwrite/blob/master/tests/e2e/Services/Functions/FunctionsCustomServerTest.php#L672)

eldadfux:
Is this the expected behavior when using the Docker socket?

I think we should first make sure this is not happening because of the way we parse the TCP call to Docker. This is not very user friendly, and I don't think we should expect users to remove these quotes.

Again, I would make an effort to double check this is not coming from the new way we use to trigger the func, if not, I think we should remove them before the user gets them in the code package.

Issue #1052, opened 2021-03-31
Set default locale on .env


simultsop:
## 🚀 Feature

To be able to set default locale for Appwrite itself.

## Have you spent some time to check if this issue has been raised before?

Asked maintainers on discord.

### Have you read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/master/CODE_OF_CONDUCT.md)?

👀

## Pitch

Was trying to launch Appwrite for some local fellas to bring more devs using appwrite initially on rapid prototyping.

TorstenDittmann:
This got fixed with https://github.com/appwrite/appwrite/releases/tag/0.8.0 👏🏻

PR #1056, opened 2021-04-01
feat-locale-env-support

lohanidamodar:
## What does this PR do?

Allows to set default Appwrite Locale using .env

## Test Plan

Existing tests should cover

## Related PRs and Issues

#1052 

(Write your answer here.)

### DIFF
--- a/.env
+++ b/.env
@@ -1,5 +1,6 @@
 _APP_ENV=production
 _APP_ENV=development
+_APP_LOCALE=en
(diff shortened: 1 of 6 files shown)


eldadfux:
@lohanidamodar we should also have this default value as part of the Dockerfile

eldadfux:
and the variables config otherwise it won't be included as part of the setup process or docs.

Issue #1025, opened 2021-03-24
Certbot exception when renewal not necessary


kodumbeats:
## 🐛 Bug Report

The certificates worker throws an exception when certbot returns a good response: `Exception: Renew isn't required`. So, when a cert has no problems, the docker logs have an exception with stack trace.

## Have you spent some time to check if this issue has been raised before?

Yes.

## To Reproduce

- Install Appwrite 0.7.2 with the domain configured in advance
- `docker-compose restart`
- `docker-compose logs appwrite-worker-certificates`

## Expected behavior

Message is logged without exception or stack trace.

## Actual Behavior

```bash
$ docker logs appwrite-worker-certificates

Appwrite certificates worker v1 has started
[notice] Starting worker 47da59cba020:7:v1-certificates
[notice] Starting work on (Job{v1-certificates} | ID: 81d8aa96638b454eb84d9dce6a2ea418 | CertificatesV1 | [{"document":[],"domain":"testing.swivi.cloud","validateTarget":false,"validateCNAME":false}])
[notice] (Job{v1-certificates} | ID: 81d8aa96638b454eb84d9dce6a2ea418 | CertificatesV1 | [{"document":[],"domain":"testing.swivi.cloud","validateTarget":false,"validateCNAME":false}]) has finished

$ docker-compose restart; docker logs appwrite-worker-certificates

Appwrite certificates worker v1 has started
[notice] Starting worker 47da59cba020:8:v1-certificates
[notice] Starting work on (Job{v1-certificates} | ID: 6843f26478964bc77f3495aee2c7ce71 | CertificatesV1 | [{"document":[],"domain":"testing.swivi.cloud","validateTarget":false,"validateCNAME":false}])
[critical] (Job{v1-certificates} | ID: 6843f26478964bc77f3495aee2c7ce71 | CertificatesV1 | [{"document":[],"domain":"testing.swivi.cloud","validateTarget":false,"validateCNAME":false}]) has failed Exception: Renew isn't required in /usr/src/code/app/workers/certificates.php:109
Stack trace:
#0 /usr/src/code/vendor/resque/php-resque/lib/Resque/Job.php(201): CertificatesV1->perform()
#1 /usr/src/code/vendor/resque/php-resque/lib/Resque/Worker.php(303): Resque_Job->perform()
#2 /usr/src/code/vendor/resque/php-resque/lib/Resque/Worker.php(242): Resque_Worker->perform(Object(Resque_Job))
[...]

PR #1060, opened 2021-04-01
Set certbot config directory

kodumbeats:
## What does this PR do?

Sets the `certbot` config directory via a command flag, so we can more usefully leverage `certbot` to handle certificate renewals more intelligently.

## Test Plan

- Deploy to a public server, with a domain configured
- Obtain SSL certificate
- Restart Appwrite stack
- Check `docker logs appwrite-worker-certificates` for `Exception: Renew isn't required` 

## Related PRs and Issues

#1025 

Yes.


eldadfux:
@kodumbeats looks good, I'm not sure I get the scope of the problem before the fix. I would love it if you could walk me through it so I could effectively help think if this change has any side-effects 🙏🏻

kodumbeats:
#### To summarize the problem:
Currently, we check the renewal date before running `certbot`. We throw [this exception](https://github.com/appwrite/appwrite/blob/master/app/workers/certificates.php#L106-L110) when renewal isn't required, adding noise to the `appwrite-worker-certificates` logs.

#### My thinking:
Rather than checking/throwing renewal exceptions by hand, we can let `certbot` handle this with the `--keep-until-expiring` flag. We can do this if we set the `--config-dir` to `/storage/certificates`, and we also enable users to work with certbot themselves. For example, this is now possible:

```bash
$ docker-compose exec appwrite-worker-certificates certbot --config-dir=/storage/certificates certificates
Saving debug log to /var/log/letsencrypt/letsencrypt.log

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
Found the following certs:
  Certificate Name: example.com
    Serial Number: [redacted]
    Key Type: RSA
    Domains: example.com
    Expiry Date: 2021-07-04 14:27:51+00:00 (VALID: 89 days)
    Certificate Path: /storage/certificates/live/example.com/fullchain.pem
    Private Key Path: /storage/certificates/live/example.com/privkey.pem
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
```

#### Unanswered questions:
- Should we change the frequency of `appwrite-worker-certificates` from [60 days + safety hour](https://github.com/appwrite/appwrite/blob/master/app/workers/certificates.php#L197) to a regular daily/weekly check? 
- We can make working with `certbot` even easier if we set the `certbot` flags in a [config file](https://certbot.eff.org/docs/using.html#configuration-file). Then, we get:
```
$ docker-compose exec appwrite-worker-certificates certbot [command]
```

eldadfux:
@kodumbeats yes, I think using this method we should change the 60+ days, because falling on the wrong hour will make the next renewal too late and result in ~30 days of invalid certificate.

Exposing the certbot as an easy to use docker command can help with manual intervention.

kodumbeats:
As is, users can use `certbot` directly, they just need to also include `--config-dir=/storage/certificates` flag. Going to keep the scope of this PR small and open another issue for a better developer experience.

eldadfux:
Please sync 0.8/dev for tests to pass

app/workers/certificates.php
```diff
@@ -57,9 +57,6 @@ public function perform()
```
eldadfux, line 57: Update the comment on line 45.

app/workers/certificates.php
```diff
@@ -160,7 +129,6 @@ public function perform()
             ],
             'domain' => $domain->get(),
             'issueDate' => \time(),
```
eldadfux, line 131: It will be a good option to have the expiration date in the DB, might be good for displaying in the UI. It might also be a good idea to store the schedule check. This will potentially allow us to find domains that have missed that check for any reason (redis failure / power shutdown - using `checkDate>1 day`)

eldadfux (changes requested):

app/workers/certificates.php
kodumbeats: To clarify, you'd like to store the `60day+1` renewal check, as well as the daily check that is performed as a part of this PR?

eldadfux: expirationDate = current time + 90 days after a successful creation 
checkDate = current time + 1 day

Rethinking about it, if we have `issueDate` we can calculate `expirationDate` without needing to store it.

kodumbeats: > Rethinking about it, if we have issueDate we can calculate expirationDate without needing to store it.
Works for me - certbot doesn't have a ready-to-parse command for this.

> checkDate = current time + 1 day
Ah, I understand better - thanks for clarifying.

kodumbeats:
After testing my code, I believe it's close to 🚢 when considering `_APP_ENV=production`. I'm not confident on what's being stored and how the code is working when `APP_ENV=development` and could use a sanity check on what's being saved to the database.

TorstenDittmann (approved):
Beside whats already said from @eldadfux 

I tested this PR with a few scenarios and it just works 🙂 Awesome idea 👍🏻

eldadfux:
Closing in favor of #3034

PR #1064, opened 2021-04-04
Feat Disable Clamav

lohanidamodar:
## What does this PR do?

Disables ClamAV for newly installing users

## Test Plan

## Related PRs and Issues

(Write your answer here.)


app/config/variables.php
```diff
@@ -311,7 +311,7 @@
                 'name' => '_APP_STORAGE_ANTIVIRUS',
                 'description' => 'This variable allows you to disable the internal anti-virus scans. This value is set to \'enabled\' by default, to cancel the scans set the value to \'disabled\'. When disabled, it\'s recommended to turn off the ClamAV container for better resource usage.',
```
eldadfux, line 312: We need to update the default value in the description

eldadfux (changes requested):
We also need to use the flag that disables the AV test on the Appwrite main container.

app/config/variables.php
lohanidamodar: Updated!

lohanidamodar:
> We also need to use the flag that disables the AV test on the Appwrite main container.

Not sure I understand what you mean!

PR #1107, opened 2021-04-24
feat: Change theme based on system preference

cadenkriese:
## What does this PR do?

Adds some very basic functionality to change the color theme based on the user's system preference.
I considered making this PR a draft and adding a theme dropdown where you choose between light, dark and system, but I'm already way out of my comfort zone here, so I figured I'd send what I have.

I definitely won't be offended if you decline this PR and implement this some other way. I just figured this was a rather nitpicky thing to make an issue for, so I went the route of making a simple PR instead.

## Test Plan

I've tested the changes in Brave & Safari on macOS 11.2.3 on Apple Silicon.

Yes ✅


TorstenDittmann (changes requested):
Good idea!

But I think we should also respect the users choice when the `localStorage` entry is already set. Meaning we should only check when `user-theme` is not set 👍🏻

eldadfux:
@cadenkriese I agree with @TorstenDittmann. One technical issue: you should move the new logic to inside the callback starting at line 17 to keep with our current conventions.

cadenkriese:
Sorry, I've been inactive on this for two weeks. I'm done with finals now, so I'd like to finish this up.

I have the js part working (just now realizing I forgot to remove my debug messages, oops), but I am very lost with the phtml / CSS bit. The icon doesn't update to a desktop for the system theme, and the current background of the button probably shouldn't stay.

Here's what it looks like currently; I'd appreciate some tips on how you'd guys would like this to look & what to change to implement that.
screenshot-2021-05-06-at-14 16 32

cadenkriese:
Unfortunately I don't have the time or interest to continue working on this, but if anyone finds the code of use that's great.

PR #1120, opened 2021-05-04
Allow external users to update password without an old password

christyjacob4:
## What does this PR do?

Users created using 
* Team Invite
* OAuth Login

Do not have an old password. So there is no way for them to set their password other than performing a password recovery. 
This PR allows these users to **create a new password without passing an old password as a one time exception** 

## Test Plan
Wrote e2e tests to check the scenarios

## Related PRs and Issues
NA

Yes


app/controllers/api/account.php
```diff
@@ -682,7 +682,7 @@
                 'emailVerification' => false,
                 'status' => Auth::USER_STATUS_UNACTIVATED,
                 'password' => null,
-                'passwordUpdate' => \time(),
+                'passwordUpdate' => 0,
```
TorstenDittmann, line 685: I don't think we should allow anonymous users to change their password. Since they don't have an E-Mail address yet.

When Updating the E-Mail address, the user needs to pass an E-Mail & their current Password. In the case of an anonymous user, the passed password will become the new password and the account is converted to a normal account.

christyjacob4: Okay makes sense 👍

app/controllers/api/teams.php
```diff
@@ -341,7 +341,17 @@
...
+                     * Set the password update time to 0 for users created 
+                     * with a team invite. This will allow us to distinguish b/w 
+                     * new and old appwrite users. 
+                     * if (passwordUpdate == 0) {
+                     *      // new appwrite user created with oauth/anonymous/team invite
+                     * } else {
+                     *      // existing appwrite user
+                     * } 
+                     */
+                    'passwordUpdate' => 0,
```
TorstenDittmann, line 354: I think we can make this comment a bit smaller and less verbose. A one liner might be enough :-)

TorstenDittmann (approved):

app/controllers/api/teams.php
christyjacob4: I have updated this.. I hope it's better. But we should be strict about this. Let me know if this needs to be changed

lohanidamodar (approved):
LGTM

eldadfux:
@TorstenDittmann waiting on your review/confirmation

TorstenDittmann (approved):

eldadfux (commented):
@christyjacob4 we should also update the route description here: docs/references/account/update-password.md

PR #1142, opened 2021-05-12
feat(teams): endpoint to update membership roles

christyjacob4:
## What does this PR do?
Adds a new endpoint PATCH '/v1/teams/:teamId/memberships/:membershipId' to allow users to modify roles previously granted to team members.

Also created a new event to trigger in this case `teams.memberships.update`

## Test Plan
Added new tests

## Related PRs and Issues
NA

Yes


app/controllers/api/teams.php
```diff
@@ -471,6 +471,83 @@
...
+
+        $isPrivilegedUser = Auth::isPrivilegedUser(Authorization::$roles);
+        $isAppUser = Auth::isAppUser(Authorization::$roles);
+        $isOwner = false;
+
+        foreach ($memberships as $member) {
+            if ($member->getAttribute('userId') == $user->getId() && \in_array('owner', $member->getAttribute('roles', []))) {
+                $isOwner = true;
+            }
+        }
```
eldadfux, line 528:
```suggestion
        $isOwner = Authorization::isRole('team:'.$team->getId().'/owner');;
```

This should work better, and helps us avoid fetching the memberships list.

app/controllers/api/teams.php
```diff
@@ -471,6 +471,83 @@
...
+
+        $memberships = $projectDB->getCollection([
+            'limit' => 2000,
+            'offset' => 0,
+            'filters' => [
+                '$collection='.Database::SYSTEM_COLLECTION_MEMBERSHIPS,
+                'teamId='.$team->getId(),
+            ],
+        ]);
```
eldadfux, line 518:
```suggestion
```

eldadfux (changes requested):

TorstenDittmann (commented):
We should make sure a team doesn't become ownerless.

Meaning, it shouldn't be possible to be the last `owner` and remove the owner role from yourself.

eldadfux:
> We should make sure a team doesn't become ownerless.
> 
> Meaning, it shouldn't be possible to be the last `owner` and remove the owner role from yourself.

I think I would leave this to be handled on client and not force a specific flow. Even when we create teams from the console they start without any members. Using the servers SDKs you can manipulate the team membership or even use cloud functions to delete orphaned teams.

TorstenDittmann:
> > We should make sure a team doesn't become ownerless.
> > Meaning, it shouldn't be possible to be the last `owner` and remove the owner role from yourself.
> 
> I think I would leave this to be handled on client and not force a specific flow. Even when we create teams from the console they start without any members. Using the servers SDKs you can manipulate the team membership or even use cloud functions to delete orphaned teams.

Fair point 👍🏻

app/controllers/api/teams.php
```diff
@@ -471,6 +471,83 @@
         ;
     });
 
+App::patch('/v1/teams/:teamId/memberships/:membershipId')
+    ->desc('Update Membership Roles')
+    ->groups(['api', 'teams'])
+    ->label('event', 'teams.memberships.update')
+    ->label('scope', 'public')
+    ->label('sdk.auth', [APP_AUTH_TYPE_SESSION, APP_AUTH_TYPE_KEY])
```
lohanidamodar, line 479: I believe `APP_AUTH_TYPE_JWT` is also required?

lohanidamodar (changes requested):
Looks good, just a minor change.

app/controllers/api/teams.php
eldadfux: Good catch!

app/controllers/api/teams.php
christyjacob4: Oh I wasn't sure this was ready!

app/controllers/api/teams.php
christyjacob4: 👍

PR #1150, opened 2021-05-17
Feat stateless webhooks worker

eldadfux:
## What does this PR do?

Drops all DB related stuff from the webhooks worker and just pass the relevant data from the API point where we trigger the task for the worker?

The worker is only accessing  the db to fetch the project's webhooks list which is already available to us in the relevant API scope. This change will increase the size of the task payload but will reduce database queries, improve speed, and make the webhooks worker completely independent.

## Test Plan

Leverage e2e tests

## Related PRs and Issues

N/A

Yes


christyjacob4 (approved):
Looks good @eldadfux

TorstenDittmann (approved):

PR #1260, opened 2021-06-08
feat-image-crop-gravity

lohanidamodar:
## What does this PR do?

Adds Gravity support during cropping

## Test Plan

The gravity feature is tetsted in the utopia/image library

## Related PRs and Issues

YES


app/controllers/api/storage.php
```diff
@@ -242,6 +242,7 @@
     ->param('fileId', '', new UID(), 'File unique ID')
     ->param('width', 0, new Range(0, 4000), 'Resize preview image width, Pass an integer between 0 to 4000.', true)
     ->param('height', 0, new Range(0, 4000), 'Resize preview image height, Pass an integer between 0 to 4000.', true)
+    ->param('gravity', Image::GRAVITY_CENTER, new Range(0, 8), 'Image crop gravity', true)
```
eldadfux, line 245: I think we should use string based names here, 0-8 is very unfriendly. I would use the Whitelist validator instead. I don't mind changing the utopia/image constant values to ones that have better naming conventions. We can translate them internally to match Imagick conventions. We don't need to follow or expose their conventions.

eldadfux (changes requested):

app/controllers/api/storage.php
lohanidamodar: Got it!

PineappleIOnic (approved):

eldadfux:
Syncing for master should resolve the tests after the #1274 fix.

eldadfux (approved):

PR #1296, opened 2021-06-17
Feat Storage bucket update and delete

lohanidamodar:
## What does this PR do?

Adds endpoint and tests to update and delete storage buckets

## Test Plan

Added test on StorageCustomServerTest and WebhookBase

## Related PRs and Issues

YES


app/controllers/api/storage.php
```diff
@@ -138,6 +135,110 @@
...
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_BUCKET)
+    ->param('bucketId', '', new UID(), 'Bucket unique ID.')
+    ->param('name', null, new Text(128), 'Bucket name', false)
+    ->param('read', null, new ArrayList(new Text(64)), 'An array of strings with read permissions. By default inherits the existing read permissions. [learn more about permissions](/docs/permissions) and get a full list of available permissions.', true)
+    ->param('write', null, new ArrayList(new Text(64)), 'An array of strings with write permissions. By default inherits the existing write permissions. [learn more about permissions](/docs/permissions) and get a full list of available permissions.', true)
+    ->param('maximumFileSize', 0, new Integer(), 'Maximum file size allowed.', true)
+    ->param('allowedFileExtensions', ['*'], new ArrayList(new Text(64)), 'Allowed file extensions', true)
+    ->param('enabled', true, new Boolean(), 'Is bucket enabled?', true)
+    ->param('adapter', 'local', new WhiteList(['local']), 'Storage adapter.', true)
```
eldadfux, line 156: We shouldn't allow switching adapters, will create too many problems for existing files. Migrating the files to another bucket and deleting existing one should be the solution.

app/controllers/api/storage.php
```diff
@@ -138,6 +135,110 @@
...
+    ->inject('deletes')
+    ->action(function ($bucketId, $response, $dbForInternal, $audits, $deletes) {
+        /** @var Appwrite\Utopia\Response $response */
+        /** @var Utopia\Database\Database $dbForInternal */
+        /** @var Appwrite\Event\Event $audits */
+        /** @var Appwrite\Event\Event $deletes */
+
+        $bucket = $dbForInternal->getDocument('buckets', $bucketId);
+
+        if (empty($bucket->getId())) {
```
eldadfux, line 220:
```suggestion
        if ($bucket->isEmpty()) {
```

app/controllers/api/storage.php
```diff
@@ -138,6 +135,110 @@
...
+        $write = (is_null($write)) ? $bucket->getAttribute('$write',[]) : $write; // By default inherit write permissions
+
+        $bucket = $dbForInternal->updateDocument('buckets', $bucket->getId(), $bucket
+                ->setAttribute('name',$name)
+                ->setAttribute('$read',$read)
+                ->setAttribute('$write',$write)
+                ->setAttribute('maximumFileSize',$maximumFileSize)
+                ->setAttribute('allowedFileExtensions',$allowedFileExtensions)
+                ->setAttribute('enabled',$enabled)
+                ->setAttribute('adapter',$adapter)
```
eldadfux, line 183:
```suggestion
```

app/controllers/api/storage.php
```diff
@@ -138,6 +135,110 @@
...
+        $deletes
+            ->setParam('type', DELETE_TYPE_DOCUMENT)
+            ->setParam('document', $bucket)
+        ;
+
+        if(!$dbForInternal->deleteDocument('buckets', $bucketId)) {
+            throw new Exception('Failed to remove project from DB', 500);
+        }
+
+        $audits
```
eldadfux, line 233: We should see what is the most performant way to delete a bucket for each adapter. We obviously need to delete all the files from the internalDB, but how do we delete the files? a single call to s3? the full parent directory?

tests/e2e/Scopes/ProjectCustom.php
```diff
@@ -127,6 +127,7 @@ public function getProject(): array
                 'storage.files.update',
                 'storage.files.delete',
                 'storage.buckets.create',
+                'storage.buckets.update',
```
eldadfux, line 130: We should also add a test for delete event.

eldadfux (changes requested):

app/controllers/api/storage.php
lohanidamodar: Whatever we decide, the delete for bucket should be handled by the delete worker right?

eldadfux: Yes. Let's discuss again when we get there.

app/controllers/api/storage.php
```diff
@@ -138,6 +135,116 @@
...
+    ->inject('dbForInternal')
+    ->inject('audits')
+    ->action(function ($bucketId, $name, $read, $write, $maximumFileSize, $allowedFileExtensions, $enabled, $encryption, $antiVirus, $response, $dbForInternal, $audits) {
+        /** @var Appwrite\Utopia\Response $response */
+        /** @var Utopia\Database\Database $dbForInternal */
+        /** @var Appwrite\Event\Event $audits */
+
+        $bucket = $dbForInternal->getDocument('buckets', $bucketId);
+
+        if (empty($bucket->getId())) {
```
eldadfux, line 169:
```suggestion
        if ($bucket->isEmpty()) {
```

app/controllers/api/storage.php
```diff
@@ -138,6 +135,116 @@
...
+        /** @var Appwrite\Event\Event $audits */
+
+        $bucket = $dbForInternal->getDocument('buckets', $bucketId);
+
+        if ($bucket->isEmpty()) {
+            throw new Exception('Bucket not found', 404);
+        }
+
+        $read = (is_null($read)) ? $bucket->getAttribute('$read', []) : $read; // By default inherit read permissions
+        $write = (is_null($write)) ? $bucket->getAttribute('$write',[]) : $write; // By default inherit write permissions
```
TorstenDittmann, line 174:
```suggestion
        $read ??= $bucket->getAttribute('$read', []); // By default inherit read permissions
        $write ??= $bucket->getAttribute('$write',[]); // By default inherit write permissions
```

eldadfux (approved):

PR #1299, opened 2021-06-18
Feat Files with Storage Buckets

lohanidamodar:
## What does this PR do?

Storage File management endpoints with Bucket Support

## Test Plan

New test in StorageBase with storage bucket support

## Related PRs and Issues

YES


app/controllers/api/storage.php
```diff
@@ -245,6 +246,155 @@
...
+        if($bucket->isEmpty()) {
+            throw new Exception("Unable to find the bucket", 404);
+        }
+
+        $file = $request->getFiles('file');
+
+        /*
+         * Validators
+         */
+        // $fileType = new FileType($bucket->getAttribute('allowedFileExtensions', ['*']));
```
TorstenDittmann, line 292: This is waiting for a fix on the `storage` library right?

app/controllers/api/storage.php
```diff
@@ -245,6 +246,155 @@
...
+        ];
+
+        if($bucket->getAttribute('encryption', true)) {
+            $data['openSSLVersion'] = '1';
+            $data['openSSLCipher'] = OpenSSL::CIPHER_AES_128_GCM;
+            $data['openSSLTag'] = \bin2hex($tag);
+            $data['openSSLIV'] = \bin2hex($iv);
+        }
+
+        $file = $dbForInternal->createDocument('files', new Document($data));
```
TorstenDittmann, line 382: What do we do when a file is uploaded, but couldn't be saved to the database?

app/controllers/api/storage.php
```diff
@@ -245,6 +246,155 @@
...
     });
 
+App::post('/v1/storage/buckets/:bucketId/files')
+    ->desc('Create File')
+    ->groups(['api', 'storage'])
+    ->label('scope', 'files.write')
+    ->label('event', 'storage.files.create')
+    ->label('sdk.auth', [APP_AUTH_TYPE_SESSION, APP_AUTH_TYPE_KEY, APP_AUTH_TYPE_JWT])
+    ->label('sdk.namespace', 'storage')
+    ->label('sdk.method', 'createFile')
```
TorstenDittmann, line 256: How do we ensure backwards compatibility?

app/controllers/api/storage.php
lohanidamodar: Yes, FileType validator. Want to make it similar to what it was before

app/controllers/api/storage.php
lohanidamodar: Let's tackle it later :D

app/init.php
```diff
@@ -47,6 +47,9 @@
 const APP_PAGING_LIMIT = 12;
 const APP_LIMIT_COUNT = 5000;
 const APP_LIMIT_USERS = 10000;
+const APP_LIMIT_ANTIVIRUS = 20971520;
+const APP_LIMIT_ENCRYPTION = 20971520;
+const APP_LIMIT_COMPRESSION = 20971520;
```
eldadfux, line 52: Let's add a comment with the human readable size

app/controllers/api/storage.php
```diff
@@ -151,11 +152,11 @@
...
-    ->param('allowedFileExtensions', ['*'], new ArrayList(new Text(64)), 'Allowed file extensions', true)
+    ->param('maximumFileSize', null, new Integer(), 'Maximum file size allowed in bytes. Maximum allowed value is ' . App::getEnv('_APP_STORAGE_LIMIT', 0) . '. For self hosted version you can change the limit by changing _APP_STORAGE_LIMIT environment variable. [Learn more about storage environment variables](docs/environment-variables#storage)', true)
+    ->param('allowedFileExtensions', [], new ArrayList(new Text(64)), 'Allowed file extensions', true)
     ->param('enabled', true, new Boolean(), 'Is bucket enabled?', true)
-    ->param('encryption', true, new Boolean(), 'Is encryption enabled?', true)
-    ->param('antiVirus', true, new Boolean(), 'Is virus scanning enabled?', true)
+    ->param('encryption', true, new Boolean(), 'Is encryption enabled? For file size above ' . Storage::human(APP_LIMIT_ENCRYPTION) . ' encryption is skipped even if it\'s enabled', true)
+    ->param('antiVirus', true, new Boolean(), 'Is virus scanning enabled? For file size above ' . Storage::human(APP_LIMIT_ENCRYPTION) . ' AntiVirus scanning is skipped even if it\'s enabled', true)
```
eldadfux, line 159:
```suggestion
    ->param('antiVirus', true, new Boolean(), 'Is virus scanning enabled? For file size above ' . Storage::human(APP_LIMIT_ANTIVIRUS) . ' AntiVirus scanning is skipped even if it\'s enabled', true)
```

app/controllers/api/storage.php
```diff
@@ -40,12 +41,12 @@
...
+    ->param('allowedFileExtensions', [], new ArrayList(new Text(64)), 'Allowed file extensions', true)
     ->param('enabled', true, new Boolean(), 'Is bucket enabled?', true)
     ->param('adapter', 'local', new WhiteList(['local']), 'Storage adapter.', true)
-    ->param('encryption', true, new Boolean(), 'Is encryption enabled?', true)
-    ->param('antiVirus', true, new Boolean(), 'Is virus scanning enabled?', true)
+    ->param('encryption', true, new Boolean(), 'Is encryption enabled? For file size above ' . Storage::human(APP_LIMIT_ENCRYPTION) . ' encryption is skipped even if it\'s enabled', true)
+    ->param('antiVirus', true, new Boolean(), 'Is virus scanning enabled? For file size above ' . Storage::human(APP_LIMIT_ENCRYPTION) . ' AntiVirus scanning is skipped even if it\'s enabled', true)
```
eldadfux, line 49:
```suggestion
    ->param('antiVirus', true, new Boolean(), 'Is virus scanning enabled? For file size above ' . Storage::human(APP_LIMIT_ANTIVIRUS) . ' AntiVirus scanning is skipped even if it\'s enabled', true)
```

app/controllers/api/storage.php
```diff
@@ -245,6 +246,218 @@
...
+        /** @var Appwrite\Utopia\Response $response */
+        /** @var Utopia\Database\Database $dbForInternal */
+        /** @var Appwrite\Database\Document $user */
+        /** @var Appwrite\Event\Event $audits */
+        /** @var Appwrite\Event\Event $usage */
+
+        $bucket = $dbForInternal->getDocument('buckets',$bucketId);
+
+        if($bucket->isEmpty()) {
+            throw new Exception("Unable to find the bucket", 404);
```
eldadfux, line 284: Let's keep the same format as we use for other 404 errors of this type.

app/controllers/api/storage.php
```diff
@@ -245,6 +246,218 @@
...
+    ->inject('usage')
+    ->action(function ($bucketId, $file, $read, $write, $request, $response, $dbForInternal, $user, $audits, $usage) {
+        /** @var Utopia\Swoole\Request $request */
+        /** @var Appwrite\Utopia\Response $response */
+        /** @var Utopia\Database\Database $dbForInternal */
+        /** @var Appwrite\Database\Document $user */
+        /** @var Appwrite\Event\Event $audits */
+        /** @var Appwrite\Event\Event $usage */
+
+        $bucket = $dbForInternal->getDocument('buckets',$bucketId);
```
eldadfux, line 281:
```suggestion
        $bucket = $dbForInternal->getDocument('buckets', $bucketId);
```

app/controllers/api/storage.php
```diff
@@ -245,6 +246,218 @@
...
+            throw new Exception("Unable to find the bucket", 404);
+        }
+
+        $file = $request->getFiles('file');
+
+        /*
+         * Validators
+         */
+        // $fileType = new FileType($bucket->getAttribute('allowedFileExtensions', ['*']));
+        $fileSize = new FileSize($bucket->getAttribute('maximumFileSize', 0));
```
eldadfux, line 293: Should we throw an error here if bucket's `maximumFileSize` is bigger than `_APP_STORAGE_LIMIT`? Just in case `_APP_STORAGE_LIMIT` was changed after the bucket created, what do you say?

app/controllers/api/storage.php
eldadfux: My idea is to keep old endpoints for X time, use migration to move all the old file to a bucket called `X`, and make sure the old endpoint treat any request if it was for bucket `X`. What do you say?

This might require us to move file locations during migration, so the delete worker could delete 1 parent dir instead of each file.

app/controllers/api/storage.php
```diff
@@ -245,6 +246,218 @@
...
+    ->label('sdk.response.model', Response::MODEL_FILE)
+    ->param('bucketId', null, new UID(), 'Storage bucket unique ID. You can create a new storage bucket using the Storage service [server integration](/docs/server/storage#createBucket).')
+    ->param('fileId', '', new UID(), 'File unique ID.')
+    ->inject('response')
+    ->inject('dbForInternal')
+    ->action(function ($bucketId, $fileId, $response, $dbForInternal) {
+        /** @var Appwrite\Utopia\Response $response */
+        /** @var Utopia\Database\Database $dbForInternal */
+
+        $bucket = $dbForInternal->getDocument('buckets', $bucketId);
```
eldadfux, line 451: We should add `404 check` for the bucket before we read the file, same as we do for collections. This is true for all `/files` endpoints

app/controllers/api/storage.php
eldadfux: Lets keep this comment unresolved until we implement this validation.

app/controllers/api/storage.php
```diff
@@ -245,6 +246,218 @@
...
+
+            if (!$antiVirus->fileScan($path)) {
+                $device->delete($path);
+                throw new Exception('Invalid file', 403);
+            }
+        }
+
+        // Compression
+        $data = $device->read($path);
+        if($size <= APP_LIMIT_COMPRESSION) {
```
eldadfux, line 343: Would be important to add a test that validates a bigger file is uploaded properly and read properly (compression + decompression both happen or not)

app/controllers/api/storage.php
eldadfux: Maybe we can create some sort of a periodic cleanup?

eldadfux: Anyway, we can tackle it later :)

app/controllers/api/storage.php
```diff
@@ -245,6 +246,218 @@
...
+    ->param('bucketId', null, new UID(), 'Storage bucket unique ID. You can create a new storage bucket using the Storage service [server integration](/docs/server/storage#createBucket).')
+    ->param('search', '', new Text(256), 'Search term to filter your list results. Max length: 256 chars.', true)
+    ->param('limit', 25, new Range(0, 100), 'Results limit value. By default will return maximum 25 results. Maximum of 100 results allowed per request.', true)
+    ->param('offset', 0, new Range(0, 2000), 'Results offset. The default value is 0. Use this param to manage pagination.', true)
+    ->param('orderType', 'ASC', new WhiteList(['ASC', 'DESC'], true), 'Order result by ASC or DESC order.', true)
+    ->inject('response')
+    ->inject('dbForInternal')
+    ->action(function ($bucketId, $search, $limit, $offset, $orderType, $response, $dbForInternal) {
+        /** @var Appwrite\Utopia\Response $response */
+        /** @var Utopia\Database\Database $dbForInternal */
```
eldadfux, line 418: We should add 404 check for the bucket before we read the files, same as we do for collections. This is true for all /files endpoints

app/controllers/api/storage.php
```diff
@@ -245,6 +246,218 @@
...
+            ->setParam('resource', 'storage/files/'.$file->getId())
+        ;
+
+        $usage
+            ->setParam('storage', $sizeActual)
+        ;
+
+        $response->setStatusCode(Response::STATUS_CODE_CREATED);
+        $response->dynamic2($file, Response::MODEL_FILE);
+        ;
```
eldadfux, line 395:
```suggestion
```

eldadfux (changes requested):

app/controllers/api/storage.php
eldadfux: OK, let's add it, I think we can also consider it as a server error (500), WDYT?

app/controllers/api/storage.php
```diff
@@ -461,13 +461,424 @@
...
+
+        $storage = 'files';
+
+        if (!\extension_loaded('imagick')) {
+            throw new Exception('Imagick extension is missing', 500);
+        }
+
+        if (!Storage::exists($storage)) {
+            throw new Exception('No such storage device', 400);
+        }
```
eldadfux, line 519: These 2 checks can be first before the bucket check

app/controllers/api/storage.php
```diff
@@ -461,13 +461,424 @@
...
+        $image->crop((int) $width, (int) $height, $gravity);
+        
+        if (!empty($opacity) || $opacity==0) {
+            $image->setOpacity($opacity);
+        }
+
+        if (!empty($background)) {
+            $image->setBackground('#'.$background);
+        }
+
```
eldadfux, line 602: Extra space

app/controllers/api/storage.php
lohanidamodar: done

app/controllers/api/storage.php
lohanidamodar: done!

.env
```diff
@@ -32,7 +32,7 @@ _APP_SMTP_PORT=1025
 _APP_SMTP_SECURE=
 _APP_SMTP_USERNAME=
 _APP_SMTP_PASSWORD=
-_APP_STORAGE_LIMIT=10000000
+_APP_STORAGE_LIMIT=30000000
```
eldadfux, line 35: Do we want to keep this value? was it changed just for tests?

app/controllers/api/storage.php
```diff
@@ -245,6 +246,638 @@
...
+    ->inject('response')
+    ->inject('dbForInternal')
+    ->action(function ($bucketId, $fileId, $response, $dbForInternal) {
+        /** @var Appwrite\Utopia\Response $response */
+        /** @var Utopia\Database\Database $dbForInternal */
+
+        $bucket = $dbForInternal->getDocument('buckets', $bucketId);
+
+        if($bucket->isEmpty()) {
+            throw new Exception("Bucket not found", 404);
```
eldadfux, line 460:
```suggestion
            throw new Exception('Bucket not found', 404);
```

Single quotes are faster 😁

app/controllers/api/storage.php
```diff
@@ -245,6 +246,638 @@
...
```
eldadfux, line 653:
```suggestion
            throw new Exception('Bucket not found', 404);
```

app/controllers/api/storage.php
```diff
@@ -245,6 +246,638 @@
...
```
eldadfux, line 718:
```suggestion
            throw new Exception('Bucket not found', 404);
```

app/controllers/api/storage.php
```diff
@@ -245,6 +246,638 @@
...
+    ->inject('audits')
+    ->action(function ($bucketId, $fileId, $read, $write, $response, $dbForInternal, $audits) {
+        /** @var Appwrite\Utopia\Response $response */
+        /** @var Utopia\Database\Database $dbForInternal */
+        /** @var Appwrite\Event\Event $audits */
+
+        $bucket = $dbForInternal->getDocument('buckets', $bucketId);
+
+        if($bucket->isEmpty()) {
+            throw new Exception("Bucket not found", 404);
```
eldadfux, line 798:
```suggestion
            throw new Exception('Bucket not found', 404);
```

app/controllers/api/storage.php
```diff
@@ -245,6 +246,638 @@
...
+        /** @var Appwrite\Utopia\Response $response */
+        /** @var Utopia\Database\Database $dbForInternal */
+        /** @var Appwrite\Event\Event $events */
+        /** @var Appwrite\Event\Event $audits */
+        /** @var Appwrite\Event\Event $usage */
+        
+        $bucket = $dbForInternal->getDocument('buckets', $bucketId);
+
+        if($bucket->isEmpty()) {
+            throw new Exception("Bucket not found", 404);
```
eldadfux, line 848:
```suggestion
            throw new Exception('Bucket not found', 404);
```

app/controllers/api/storage.php
```diff
@@ -245,6 +246,638 @@
...
+        /** @var Appwrite\Utopia\Response $response */
+        /** @var Utopia\Database\Database $dbForInternal */
+        /** @var Appwrite\Database\Document $user */
+        /** @var Appwrite\Event\Event $audits */
+        /** @var Appwrite\Event\Event $usage */
+
+        $bucket = $dbForInternal->getDocument('buckets', $bucketId);
+
+        if($bucket->isEmpty()) {
+            throw new Exception("Bucket not found", 404);
```
eldadfux, line 284:
```suggestion
            throw new Exception('Bucket not found', 404);
```

app/controllers/api/storage.php
```diff
@@ -245,6 +246,638 @@
...
+    ->inject('response')
+    ->inject('dbForInternal')
+    ->action(function ($bucketId, $search, $limit, $offset, $orderType, $response, $dbForInternal) {
+        /** @var Appwrite\Utopia\Response $response */
+        /** @var Utopia\Database\Database $dbForInternal */
+
+        $bucket = $dbForInternal->getDocument('buckets', $bucketId);
+
+        if($bucket->isEmpty()) {
+            throw new Exception("Bucket not found", 404);
```
eldadfux, line 423:
```suggestion
            throw new Exception('Bucket not found', 404);
```

app/controllers/api/storage.php
```diff
@@ -245,6 +246,638 @@
...
+            throw new Exception('Imagick extension is missing', 500);
+        }
+
+        if (!Storage::exists($storage)) {
+            throw new Exception('No such storage device', 400);
+        }
+        $bucket = $dbForInternal->getDocument('buckets', $bucketId);
+
+        if($bucket->isEmpty()) {
+            throw new Exception("Bucket not found", 404);
```
eldadfux, line 518:
```suggestion
            throw new Exception('Bucket not found', 404);
```

eldadfux (changes requested):

app/controllers/api/storage.php
lohanidamodar: I think this can be resolved.

.env
lohanidamodar: We need this as we added test for large file, without this large file test will fail.

app/controllers/api/storage.php
lohanidamodar: Good enough!

TorstenDittmann (approved):

eldadfux (approved):

PR #1308, opened 2021-06-19
Execute functions up to 5 seconds faster

Meldiron:
## What does this PR do?
Updated INTERVALS of php-resque

In general, this PR speeds up functions, webhooks and tasks by up to 5 seconds. Also, it speeds up the overall speed of the appwrite.

## Test Plan

I run a whole project on Appwrite using functions and real-time feature.
Before using this fork, the time between running `createExecuton` and receiving the update event using real-time took 4-5 seconds.
Using this fork, I reduced the time to 1s.

### Addational research info

I tested everything on a function that takes ~1s to execute (load libraries into NodeJS, connect to MySQL server, get some data, process them and console log result). If a request takes 1s, it means 0ms delay due to Appwrite.

Currently, Appwrite uses the default INTERVAL configuration which is 5 seconds:
https://github.com/resque/php-resque/blob/master/lib/Resque.php#L13
https://github.com/resque/php-resque/blob/master/lib/Resque/Worker.php#L160

By changing this number to 0.1s, the MAXIMUM possible delay was reduced from 5000ms to 100ms. This resulted in really fast function executions but also highly increased CPU usage:
image

I switched to INTERVAL=1 which fixed the high CPU and still increased the speed of every request by up to 4 seconds:
image

Then I wanted to see what happens when I set INTERVAL=1 on everything but set INTERVAL=0.1 on the functions container. This was the result:
image
Also, of course, MAXIMUM of 100ms delay between creating and starting execution.

Finally, I decreased INTERVAL on functions, webhook and tasks real-time to 0.1 which resulted in these stats:
image

---

Current PR sets INTERVAL=1 on everything except functions, tasks and webhooks. These have even faster INTERVAL=0.1

I am not fully sure what to think about these stats, so it will need some more benchmarking but these screenshots should provide solid ground.

### DIFF
--- a/bin/worker-usage
+++ b/bin/worker-usage
@@ -7,4 +7,4 @@ else
     REDIS_BACKEND="redis://${_APP_REDIS_USER}:${_APP_REDIS_PASS}@${_APP_REDIS_HOST}:${_APP_REDIS_PORT}"
 fi
 
-INTERVAL=0.1 QUEUE='v1-usage' APP_INCLUDE='/usr/src/code/app/workers/usage.php' php /usr/src/code/vendor/bin/resque -dopcache.preload=opcache.preload=/usr/src/code/app/preload.php
+INTERVAL=1 QUEUE='v1-usage' APP_INCLUDE='/usr/src/code/app/workers/usage.php' php /usr/src/code/vendor/bin/resque -dopcache.preload=opcache.preload=/usr/src/code/app/preload.php
--- a/bin/worker-usage
+++ b/bin/worker-usage
@@ -7,4 +7,4 @@ else
     REDIS_BACKEND="redis://${_APP_REDIS_USER}:${_APP_REDIS_PASS}@${_APP_REDIS_HOST}:${_APP_REDIS_PORT}"
 fi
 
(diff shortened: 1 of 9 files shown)


bin/worker-usage
```diff
@@ -7,4 +7,4 @@ else
     REDIS_BACKEND="redis://${_APP_REDIS_USER}:${_APP_REDIS_PASS}@${_APP_REDIS_HOST}:${_APP_REDIS_PORT}"
 fi
 
-QUEUE='v1-usage' APP_INCLUDE='/usr/src/code/app/workers/usage.php' php /usr/src/code/vendor/bin/resque -dopcache.preload=opcache.preload=/usr/src/code/app/preload.php
\ No newline at end of file
+INTERVAL=0.1 QUEUE='v1-usage' APP_INCLUDE='/usr/src/code/app/workers/usage.php' php /usr/src/code/vendor/bin/resque -dopcache.preload=opcache.preload=/usr/src/code/app/preload.php
```
eldadfux, line 10: Usage can stay at 1 seconds

eldadfux (changes requested):

bin/worker-usage
Meldiron: My bad, kept it there from testing. Fixed.

eldadfux (approved):

PR #1322, opened 2021-06-23
Fix functions worker

christyjacob4:
## What does this PR do?

We are passing the string `null` as the value of an env variable if it's empty. 
Instead we should pass an empty string. 

## Test Plan
Use existing tests

## Related PRs and Issues
NA

Yes


eldadfux:
@christyjacob4 Would be great to push it with 0.9 if you feel comfortable with the changes.

PR #1323, opened 2021-06-24
Update OAuth2 session doc

lohanidamodar:
## What does this PR do?

Improve OAuth2 create session doc mentioning how the session gets attached to existing user

## Test Plan

N/A

## Related PRs and Issues

N/A

YA


docs/references/account/create-session-oauth2.md
```diff
@@ -1 +1,3 @@
-Allow the user to login to their account using the OAuth2 provider of their choice. Each OAuth2 provider should be enabled from the Appwrite console first. Use the success and failure arguments to provide a redirect URL's back to your app when login is completed.
\ No newline at end of file
+Allow the user to login to their account using the OAuth2 provider of their choice. Each OAuth2 provider should be enabled from the Appwrite console first. Use the success and failure arguments to provide a redirect URL's back to your app when login is completed.
+
+If there is already an active session, the new account details is attached to the same account that has the active session. If there is no active sessions, it will look for the user with the same email address as the email received from OAuth2 provider and attach new session to the existing user if found or create a new user if the email doesn't exist in the database.
```
PineappleIOnic, line 3:
```suggestion
If there is already an active session, the new account details is attached to the same account that already has the active session. If there are no active sessions, it will look for the user with the same email address as the email received from OAuth2 provider and attach the new session to the existing user if found however it will create a new user if the email doesn't exist in the database.
```

PineappleIOnic (changes requested):
Overall this looks really good! Here are a few of my ideas to make the text even better :+1:

docs/references/account/create-session-oauth2.md
```diff
@@ -1 +1,3 @@
-Allow the user to login to their account using the OAuth2 provider of their choice. Each OAuth2 provider should be enabled from the Appwrite console first. Use the success and failure arguments to provide a redirect URL's back to your app when login is completed.
\ No newline at end of file
+Allow the user to login to their account using the OAuth2 provider of their choice. Each OAuth2 provider should be enabled from the Appwrite console first. Use the success and failure arguments to provide a redirect URL's back to your app when login is completed.
+
+If there is already an active session, the new account details is attached to the same account that already has the active session. If there are no active sessions, it will look for the user with the same email address as the email received from OAuth2 provider and attach the new session to the existing user if found however it will create a new user if the email doesn't exist in the database.
```
eldadfux, line 3:
```suggestion
If there is already an active session, the new session will be attached to the logged-in account. If there are no active sessions, the server will attempt to look for a user with the same email address as the email received from the OAuth2 provider and attach the new session to the existing user. If no matching user is found - the server will create a new user..
```

Issue #1334, opened 2021-06-30
Execution Objects: truncated field length is difficult to configure


mlucas-NU:
## 🚀 Feature

Add a way for developers to configure the truncation length for text fields in the Execution Object (specifically `stdout` and `stderr`).

## Have you spent some time to check if this issue has been raised before?

Yes, and I discussed this briefly with Eldad on Discord.

### Have you read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/master/CODE_OF_CONDUCT.md)?

Yes.

## Pitch

Function endpoints respond with an [Execution Object](https://appwrite.io/docs/models/execution). The object's `stdout` and `stderr` fields are truncated to 4000 characters by default, but these values can only be adjusted by modifying the appwrite source.

Is it possible to provide a way for developers to configure this value?

Relevant Code Locations:
* https://github.com/appwrite/appwrite/blob/0d2a4f0cbfc17f9b1e4c431b078a19ac31bb1eaa/app/workers/functions.php#L471-L472

PineappleIOnic:
Update: We are now looking at adding a new environment variable to allow you to configure this variable :+1:

mlucas-NU:
A colleague of mine reminded me of an additional solution that might help functions return huge payloads. When sharing huge files between AWS compute instances, he had the sender upload the file to S3 and only send a signed URL to the receiver. This way, he didn't have to scale up his services to handle all of the transfer bandwidth and maintain long-running connections (these files were generally many gigabytes in size).

In Appwrite terms, a could Function save a huge payload as a file in the Storage service and then share a URL to the file to the calling client. The client could then retrieve the file from Storage when they're ready to use it.

Given that this could fill the Storage service with files that are only used once, I wonder if the function response could include an expiration datetime along with the Storage URL and then schedule a Function to delete the file at that expiration datetime.

TorstenDittmann:
This should be solved now with the new Functions service 👍🏻

arcanechimp:
Is this env variable mentioned in PR #1336 still going to be added?

I would like to use this as a way to access data from a seperate cache and database whilst using Appwrite's authentication, and being able to change the max value would be very helpful to achieving this.

PR #1336, opened 2021-06-30
Add a environment variable to configure max log characters for functions

PineappleIOnic:
## What does this PR do?

It adds a new environment variable `_APP_FUNCTIONS_MAX_OUTPUT` which will configure the max length of logs stored by appwrite from function executions. The default has also been raised from 4000 to 65535 which is the limit we can store with a TEXT Type in our MariaDB Database since that's the type that is storing our logs.

## Related PRs and Issues
https://github.com/appwrite/appwrite/issues/1334

Yes, I have.


app/config/variables.php
```diff
@@ -444,6 +444,15 @@
                 'question' => '',
                 'filter' => ''
             ],
+            [
+                'name' => '_APP_FUNCTIONS_OUTPUT_MAX',
+                'description' => 'The maximum length of characters a function execution is allowed to output. The default value is 65535',
+                'introduction' => '0.9.0',
+                'default' => '65535',
```
TorstenDittmann, line 451: This is the maximum anyway right?

TorstenDittmann (approved):

app/config/variables.php
PineappleIOnic: Yeah it is the max currently using the TEXT Type in MariaDB. Not sure about the rewrite though.

TorstenDittmann: Lets have this in the description - otherwise people might get confused when they raise this value 👍🏻

PineappleIOnic: Done :+1:

eldadfux:
Keeping this PR pending for the new DB changes which will allow more flexibility around the output length

PR #1350, opened 2021-07-05
Feat Custom ID

lohanidamodar:
## What does this PR do?

Adds a required ID parameter in create endpoints, so that user can assign custom ID for resource instead of the auto generated one. Passing the string `unique()` as ID should work as before by generating ID automatically.

## Test Plan

- CustomID validator and relevant Unit test
- Relevant updates to E2E tests for each service updated to have custom id support

## Related PRs and Issues

YES


app/controllers/api/users.php
```diff
@@ -30,19 +31,20 @@
     ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
     ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
     ->label('sdk.response.model', Response::MODEL_USER)
+    ->param('id', 'uid', new Key(), 'Unique Id. Passing `uid()` will auto generate the id.')
```
eldadfux, line 34:
```suggestion
    ->param('userId', 'unique()', new Key(), 'Unique Id. Passing the string `unique()` will auto generate your ID.')
```
- `userId` is more consistent
- `unique()` feels more verbose, we need to adjust Key to accept it.

lohanidamodar: I cannot actually use `unique()` the `()` will cause the validator to fail?

TorstenDittmann: Why `userId` instead of just `id`? 🤔

src/Appwrite/Database/Validator/CustomId.php
```diff
@@ -0,0 +1,21 @@
+<?php
+namespace Appwrite\Database\Validator;
+
+use Utopia\Database\Validator\Key;
+
+class CustomId extends Key {
```
eldadfux, line 6: Need some tests here

lohanidamodar: Sure!

tests/unit/Database/Validator/CustomIdTest.php
```diff
@@ -0,0 +1,41 @@
...
+        $this->assertEquals($this->object->isValid('dasda_asdasd'), true);
+        $this->assertEquals($this->object->isValid('asdasdasdas'), true);
+        $this->assertEquals($this->object->isValid('_asdasdasdas'), false);
+        $this->assertEquals($this->object->isValid('as$$5dasdasdas'), false);
+        $this->assertEquals($this->object->isValid(false), false);
+        $this->assertEquals($this->object->isValid(null), false);
+        $this->assertEquals($this->object->isValid('socialAccountForYoutubeSubscribers'), false);
+        $this->assertEquals($this->object->isValid('socialAccountForYoutubeSubscriber'), false);
+        $this->assertEquals($this->object->isValid('socialAccountForYoutubeSubscribe'), true);
+        $this->assertEquals($this->object->isValid('socialAccountForYoutubeSubscrib'), true);
```
TorstenDittmann, line 39: Maybe a bit more self explanatory strings to test. Seeing all the `asdasd` might be a bit random and doesn't tell me right away what each of those test for 🤔

TorstenDittmann: Something like `with space` instead of `dasda asdasd` 😄

lohanidamodar:
The Empty message was looking weird, so I added `Automatically generated` as message
image

image

here is how it looks when custom id is enabled

image

public/scripts/views/forms/custom-id.js
```diff
@@ -0,0 +1,164 @@
...
+          setIdType(idType);
+        } else {
+          idType = "custom";
+          setIdType(idType);
+        }
+      }
+
+      var validate = function (event) {
+        var service = element.dataset["validator"];
+        service = service.split('.');
```
TorstenDittmann, line 50:
```suggestion
        const [service, method] = service.split('.');
```
Can we do something like this to prevent accessing by the index?

public/scripts/views/forms/custom-id.js
```diff
@@ -0,0 +1,164 @@
...
+        const CAPITAL_Z = 90;
+        const UNDERSCORE = 95;
+
+        const isNotValidDigit = key < ZERO || key > NINE;
+        const isNotValidSmallAlphabet = key < SMALL_A || key > SMALL_Z;
+        const isNotValidCapitalAlphabet = key < CAPITAL_A || key > CAPITAL_Z;
+
+        //Leading underscore is prevented
+        if (key == UNDERSCORE && e.target.value.length == 0) {
+          e.preventDefault();
```
TorstenDittmann, line 149: Not sure if this is the best UX.

So we don't show any hint at all to why the the user can't type `_`?

public/scripts/views/forms/custom-id.js
lohanidamodar: Nice!

public/scripts/views/forms/custom-id.js
lohanidamodar: it's the same for all other disabled keys, we have a helper message below text. It's also same for native maxLength attribute in HTML, it doesn't provide any feedback?

public/scripts/views/forms/custom-id.js
```diff
@@ -0,0 +1,164 @@
...
+          event.target.setCustomValidity("ID is required");
+        } else {
+          if (service[0] == 'projects') {
+            if (service[1] == 'getPlatform') {
+              var projectId = element.form.elements.namedItem("projectId").value;
+              console[service[0]][service[1]](projectId, value).then(function (res) {
+                if (res.$id == value) {
+                  event.target.setCustomValidity("ID already exists");
+                } else {
+                  event.target.setCustomValidity("");
```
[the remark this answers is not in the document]
lohanidamodar: as far as I know it shouldn't be right?

public/scripts/views/forms/custom-id.js
TorstenDittmann: Fair enough 👍🏻

public/scripts/views/forms/custom-id.js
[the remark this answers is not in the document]
TorstenDittmann: Can't think of a case as well 😄 

Let's double check if we don't miss anything.

lohanidamodar:
Custom ID field Icon update, and validator
image

TorstenDittmann (approved):

app/views/console/settings/index.phtml
```diff
@@ -51,9 +51,6 @@ $smtpEnabled = $this->getParam('smtpEnabled', false);
                                 data-failure-param-alert-classname="error">
 
                                 <input name="$id" type="hidden" data-ls-bind="{{console-project.$id}}" />
-                                
-                                <label for="name">Name</label>
-                                <input name="name" id="name" type="text" autocomplete="off" data-ls-bind="{{console-project.name}}" data-forms-text-direction required maxlength="128" />
```
eldadfux, line 54: We decided to keep the project name.

tests/unit/Database/Validator/CustomIdTest.php
eldadfux: Agree.

eldadfux (changes requested):

eldadfux:
Some blockers before we can merge:

1. we need to remove this tooltip:
image

2. We need to make tests pass
3. There is an error in the custom-id.js file - this `keypress` variable is missing

PR #1357, opened 2021-07-06
Feat kotlin java docs

abnegate:
## What does this PR do?

Allows setting SDK's `docDirectories` which will generate docs based on the declared languages and their templates.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -62,7 +62,7 @@
         "slickdeals/statsd": "3.1.0"
     },
     "require-dev": {
-        "appwrite/sdk-generator": "0.11.1",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -62,7 +62,7 @@
         "slickdeals/statsd": "3.1.0"
     },
     "require-dev": {
-        "appwrite/sdk-generator": "0.11.1",
+        "appwrite/sdk-generator": "dev-feat-kotlin-java-docs",
```
eldadfux, line 65:
```suggestion
        "appwrite/sdk-generator": "0.12.0",
```

PR #1375, opened 2021-07-10
Added " to wrap environment name in a log

Meldiron:
## What does this PR do?

Fixes a tiny mistake, adding an `"` to wrap the environment name properly in the log.

## Test Plan

Did not test, should not cause any bugs.

I noticed a bug when someone reported their error on Discord and the environment name was missing.
image


eldadfux:
Thank you ❤️

PR #1380, opened 2021-07-11
Feat storage buckets

lohanidamodar:
## What does this PR do?

Storage Buckets Features as described in [RFC](https://github.com/appwrite/rfc/blob/main/017-storage-buckets.md)

## Test Plan

Relevant E2E tests under Storage service

## Related PRs and Issues

YES


app/controllers/api/storage.php
```diff
@@ -24,9 +25,449 @@
...
+
+        $read ??= $bucket->getAttribute('$read', []); // By default inherit read permissions
+        $write ??= $bucket->getAttribute('$write',[]); // By default inherit write permissions
+        $read??=$bucket->getAttribute('$read', []); // By default inherit read permissions
+        $write??=$bucket->getAttribute('$write', []); // By default inherit write permissions
+        $maximumFileSize??=$bucket->getAttribute('maximumFileSize', (int)App::getEnv('_APP_STORAGE_LIMIT', 0));
+        $allowedFileExtensions??=$bucket->getAttribute('allowedFileExtensions', []);
+        $enabled??=$bucket->getAttribute('enabled', true);
+        $encryption??=$bucket->getAttribute('encryption', true);
+        $antiVirus??=$bucket->getAttribute('antiVirus', true);
```
TorstenDittmann, line 389:
```suggestion
        $read ??= $bucket->getAttribute('$read', []); // By default inherit read permissions
        $write ??= $bucket->getAttribute('$write', []); // By default inherit write permissions
        $maximumFileSize ??= $bucket->getAttribute('maximumFileSize', (int)App::getEnv('_APP_STORAGE_LIMIT', 0));
        $allowedFileExtensions ??= $bucket->getAttribute('allowedFileExtensions', []);
        $enabled ??= $bucket->getAttribute('enabled', true);
        $encryption ??= $bucket->getAttribute('encryption', true);
        $antiVirus ??= $bucket->getAttribute('antiVirus', true);
```

whitespace

app/controllers/api/storage.php
```diff
@@ -24,9 +25,449 @@
...
+        }
+
+        $deletes
+            ->setParam('type', DELETE_TYPE_DOCUMENT)
+            ->setParam('document', $bucket)
+        ;
+
+        if(!$dbForInternal->deleteDocument('buckets', $bucketId)) {
+            throw new Exception('Failed to remove project from DB', 500);
+        }
```
TorstenDittmann, line 452: We should set the event for the deletes worker after a successful deletion of the bucket. Otherwise we might not remove the bucket document - but trigger the delete worker 👍🏻

app/controllers/api/storage.php
```diff
@@ -24,9 +25,449 @@
...
+    ->inject('usage')
+    ->action(function ($bucketId, $name, $read, $write, $maximumFileSize, $allowedFileExtensions, $enabled, $adapter, $encryption, $antiVirus, $response, $dbForInternal, $audits, $usage) {
+        /** @var Appwrite\Utopia\Response $response */
+        /** @var Utopia\Database\Database $dbForInternal */
+        /** @var Appwrite\Event\Event $audits */
+        /** @var Appwrite\Stats\Stats $usage */
+
+        $bucketId = $bucketId == 'unique()' ? $dbForInternal->getId() : $bucketId;
+        try {
+            $dbForInternal->createCollection('bucket_' . $bucketId, [
```
kodumbeats, line 67: Can we keep this schema definition somewhere else in `app/config` and load when needed? I think this controller would look a lot cleaner without this big block.

kodumbeats, line 67: If we're creating a collection per bucket, would it make sense to keep this in `$dbForExternal`? That's where user-created collections are stored, and I think it makes some sense to keep them together.

app/controllers/api/storage.php
```diff
@@ -24,9 +25,449 @@
...
+    ->label('sdk.namespace', 'storage')
+    ->label('sdk.method', 'createBucket')
+    ->label('sdk.description', '/docs/references/storage/create-bucket.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_BUCKET)
+    ->param('bucketId', '', new CustomId(), 'Unique Id. Choose your own unique ID or pass the string `unique()` to auto generate it. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.')
+    ->param('name', '', new Text(128), 'Bucket name', false)
+    ->param('read', [], new ArrayList(new Text(64)), 'An array of strings with read permissions. By default no user is granted with any read permissions. [learn more about permissions](/docs/permissions) and get a full list of available permissions.', true)
+    ->param('write', [], new ArrayList(new Text(64)), 'An array of strings with write permissions. By default no user is granted with any write permissions. [learn more about permissions](/docs/permissions) and get a full list of available permissions.', true)
```
kodumbeats, line 48: How do the bucket permissions interact with file permissions?

kodumbeats, line 48: Could we use the Permissions validator from `Utopia\Database` here?
https://github.com/utopia-php/database/blob/main/src/Database/Validator/Permissions.php

kodumbeats (changes requested):
Only got through the first few files tonight, but found a few things.

app/controllers/api/storage.php
lohanidamodar: Where would be a good place to place this @eldadfux @TorstenDittmann ?

lohanidamodar: @kodumbeats I don't that makes sense, this is not user created collection, this collection is managed by Appwrite, user cannot change schema and stuff. User can only change the document i.e. files.

app/controllers/api/storage.php
lohanidamodar: This has to be updated to match how we handle **Collections** permission. If that's already handled in the DB i will update it accordingly. As we have decided buckets and collections should work same way.

app/controllers/api/storage.php
TorstenDittmann: I do think the external database will be the better fit for this. Since then we handle it the same way as collections - which also cannot be changed by clients.

TorstenDittmann: For better maintainability, we can just grab this informations here from `collections2.php` 👍🏻 

That way we don't have to maintain basically 2 structures for a bucket 🙂

app/controllers/api/storage.php
```diff
@@ -244,6 +245,7 @@
                 'dateCreated' => \time(),
                 'dateUpdated' => \time(),
                 'name' => $name,
+                'permission' => $permission,
```
kodumbeats, line 248: I had to make changes to the Database routes when implementing collection/document permissions. Does anything else need to be added to the Storage controller? 
ref: https://github.com/appwrite/appwrite/pull/1480

kodumbeats (changes requested):

app/controllers/api/storage.php
lohanidamodar: Do we keep it in a separate array? Otherwise another Files collection will also be created which will not be used?

lohanidamodar: The attributes, names, for collections can be changed by the developers, however for files, the collection structure cannot be updated in any way as it's pre-defined by the system. 
I believe any collection pre-defined by the system, we should keep in the internal db, and any collection that users define should be kept in the external db.

app/controllers/api/storage.php
lohanidamodar: @kodumbeats I don't see the changes you made to the routes, can you point me in the right direction?

app/controllers/api/storage.php
```diff
@@ -112,40 +584,49 @@
         }
 
         // Compression
-        $compressor = new GZIP();
         $data = $device->read($path);
-        $data = $compressor->compress($data);
-        $key = App::getEnv('_APP_OPENSSL_KEY_V1');
-        $iv = OpenSSL::randomPseudoBytes(OpenSSL::cipherIVLength(OpenSSL::CIPHER_AES_128_GCM));
-        $data = OpenSSL::encrypt($data, OpenSSL::CIPHER_AES_128_GCM, $key, 0, $iv, $tag);
+        if($size <= APP_LIMIT_COMPRESSION) {
```
TorstenDittmann, line 588: Lets use strict comparison `===`

app/controllers/api/storage.php
```diff
@@ -112,40 +584,49 @@
...
+            $compressor = new GZIP();
+            $data = $compressor->compress($data);
+        }
+        
+        if($bucket->getAttribute('encryption', true) && $size <= APP_LIMIT_ENCRYPTION) {
```
TorstenDittmann, line 593: `<==`

app/controllers/api/storage.php
```diff
@@ -289,6 +794,11 @@
         if (!Storage::exists($storage)) {
             throw new Exception('No such storage device', 400);
         }
+        $bucket = $dbForInternal->getDocument('buckets', $bucketId);
+
+        if($bucket->isEmpty()) {
+            throw new Exception('Bucket not found', 404);
+        }
 
         if ((\strpos($request->getAccept(), 'image/webp') === false) && ('webp' == $output)) { // Fallback webp to jpeg when no browser support
```
TorstenDittmann, line 803: Let's add strict comparison here 👍🏻

app/controllers/api/storage.php
```diff
@@ -301,9 +811,9 @@
         $date = \date('D, d M Y H:i:s', \time() + (60 * 60 * 24 * 45)).' GMT';  // 45 days cache
         $key = \md5($fileId.$width.$height.$gravity.$quality.$borderWidth.$borderColor.$borderRadius.$opacity.$rotation.$background.$storage.$output);
 
-        $file = $dbForInternal->getDocument('files', $fileId);
+        $file = $dbForInternal->getDocument('bucket_' . $bucketId, $fileId);
 
-        if (empty($file->getId())) {
+        if ($file->isEmpty() || $file->getAttribute('bucketId') != $bucketId) {
```
TorstenDittmann, line 816: `!==`

app/controllers/api/storage.php
```diff
@@ -417,18 +927,25 @@
...
+        $bucket = $dbForInternal->getDocument('buckets', $bucketId);
+
+        if($bucket->isEmpty()) {
+            throw new Exception('Bucket not found', 404);
+        }
+
+        $file = $dbForInternal->getDocument('bucket_' . $bucketId, $fileId);
 
-        if (empty($file->getId())) {
+        if ($file->isEmpty() || $file->getAttribute('bucketId') != $bucketId) {
```
TorstenDittmann, line 948: `!==`

app/controllers/api/storage.php
```diff
@@ -482,19 +1000,26 @@
...
+
+        if($bucket->isEmpty()) {
+            throw new Exception('Bucket not found', 404);
+        }
+
+        $file  = $dbForInternal->getDocument('bucket_' . $bucketId, $fileId);
         $mimes = Config::getParam('storage-mimes');
 
-        if (empty($file->getId())) {
+        if ($file->isEmpty() || $file->getAttribute('bucketId') != $bucketId) {
```
TorstenDittmann, line 1022: `!==`

app/controllers/api/storage.php
```diff
@@ -531,7 +1056,7 @@
 
         $usage
             ->setParam('storage.files.read', 1)
-            ->setParam('bucketId', 'default')
+            ->setParam('bucketId', $bucketId)
         ;
 
         // Response
```
TorstenDittmann, line 1062: Can we remove that comment? 🙏🏻

app/controllers/api/storage.php
```diff
@@ -742,7 +1283,11 @@
...
 
-        // TODO: Check if the storage bucket exists else throw 404 
+        $bucket = $dbForInternal->getDocument('buckets', $bucketId);
+
+        if($bucket->isEmpty()) {
+            throw new Exception('Bucket not found', 404);
+        } 
         
         $usage = [];
         if (App::getEnv('_APP_USAGE_STATS', 'enabled') == 'enabled') {
```
TorstenDittmann, line 1293: `===`

app/workers/deletes.php
```diff
@@ -53,6 +53,9 @@ public function run(): void
                     case DELETE_TYPE_TEAMS:
                         $this->deleteMemberships($document, $projectId);
                         break;
+                    case 'buckets':
```
TorstenDittmann, line 56: Can we use a constant here?

TorstenDittmann (changes requested):

app/controllers/api/storage.php
```diff
@@ -24,9 +25,465 @@
...
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'storage')
+    ->label('sdk.method', 'listBuckets')
+    ->label('sdk.description', '/docs/references/storage/list-buckets.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_BUCKET_LIST)
+    ->param('search', '', new Text(256), 'Search term to filter your list results. Max length: 256 chars.', true)
+    ->param('limit', 25, new Range(0, 100), 'Results limit value. By default will return maximum 25 results. Maximum of 100 results allowed per request.', true)
+    ->param('offset', 0, new Range(0, 2000), 'Results offset. The default value is 0. Use this param to manage pagination.', true)
```
kodumbeats, line 299: What do you think about setting this offset maximum to the count limit?
https://github.com/appwrite/appwrite/blob/252720dd2874e42988fff8f8c3619b0eeeebb0fa/app/init.php#L63

app/controllers/api/storage.php
```diff
@@ -24,9 +25,465 @@
...
+
+            if ($cursorBucket->isEmpty()) {
+                throw new Exception("Bucket '{$cursor}' for the 'cursor' value not found.", 400);
+            }
+        }
+
+        $usage->setParam('storage.buckets.read', 1);
+
+        $response->dynamic(new Document([
+            'buckets' => $dbForInternal->find('buckets', $queries, $limit, $offset, [], [$orderType], $cursorBucket ?? null, $cursorDirection),
```
kodumbeats, line 324: What attribute are we ordering by?

kodumbeats (changes requested):
Still need to finish reviewing the storage controller, but I think there's enough to start :)

app/controllers/api/storage.php
lohanidamodar: Don't think there is something like <== ?

src/Appwrite/Utopia/Response/Model/Bucket.php
```diff
@@ -0,0 +1,122 @@
...
+        ->addRule('encryption', [
+            'type' => self::TYPE_BOOLEAN,
+            'description' => 'Bucket is encrypted.',
+            'default' => true,
+            'example' => false,
+        ])
+        ->addRule('antiVirus', [
+            'type' => self::TYPE_BOOLEAN,
+            'description' => 'Virus scanning is enabled.',
+            'default' => true,
```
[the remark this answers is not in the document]
lohanidamodar: I think default has to either true  or false? can't be empty string, and don't think it's okay to be null? Please suggest

src/Appwrite/Messaging/Adapter/Realtime.php
```diff
@@ -273,12 +273,6 @@ public static function fromPayload(string $event, Document $payload): array
                 $channels[] = 'teams.' . $payload->getId();
                 $roles = ['team:' . $payload->getId()];
 
-                break;
-            case strpos($event, 'database.collections.') === 0:
```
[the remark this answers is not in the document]
lohanidamodar: @TorstenDittmann suggested that we are not supporting realtime on server side side and collection events only makes sense on server side. And we will not be adding buckets events makes sense to remove collection events as well as no clients will ever be able to use them. cc @TorstenDittmann

tests/e2e/Services/Realtime/RealtimeBase.php
```diff
@@ -623,19 +623,6 @@ public function testChannelDatabase()
...
-        $this->assertArrayHasKey('data', $response);
-        $this->assertEquals('event', $response['type']);
-        $this->assertNotEmpty($response['data']);
-        $this->assertArrayHasKey('timestamp', $response['data']);
-        $this->assertCount(2, $response['data']['channels']);
-        $this->assertContains('collections', $response['data']['channels']);
-        $this->assertContains('collections.' . $actors['body']['$id'], $response['data']['channels']);
-        $this->assertEquals('database.collections.create', $response['data']['event']);
-        $this->assertNotEmpty($response['data']['payload']);
-
```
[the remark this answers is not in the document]
lohanidamodar: @kodumbeats same as above. Removing realtime events related to collections.

app/controllers/api/storage.php
lohanidamodar: I see different value in collections - for collections it seems to be 40000 and for documents it seems to be 9000000 ? What should I do for buckets and files? and how are we deciding this?

app/controllers/api/storage.php
lohanidamodar: ya right? Thinking the same I just copied the same behavior from collection here https://github.com/appwrite/appwrite/blob/252720dd2874e42988fff8f8c3619b0eeeebb0fa/app/controllers/api/database.php#L234
?

app/controllers/api/storage.php
lohanidamodar: Don't think there is something like <==?

tests/e2e/Services/Realtime/RealtimeBase.php
```diff
@@ -783,8 +770,25 @@ public function testChannelFiles()
 
         /**
          * Test File Create
+         * TODO Buckets Channels
```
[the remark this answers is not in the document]
lohanidamodar: No not.

Meldiron:
With `Buckets -> Files` I can see same structure as we use in `Collections -> Documents`... More specific, I am seeing the same permission system, although, on the collection, you can pick collection/document-level permission. Why don't we have it here? @kodumbeats any thoughts? We could ask the same questions as we ask in database:

- How do permissions nest?
- How do I allow users to edit, but not write new ones?

Meldiron:
We are limiting size, but I can't see a limit for the amount of files uploaded. Should we consider that? I could blow up your appwrite cloud budgets in a few hours overnight.

lohanidamodar:
> With `Buckets -> Files` I can see same structure as we use in `Collections -> Documents`... More specific, I am seeing the same permission system, although, on the collection, you can pick collection/document-level permission. Why don't we have it here? @kodumbeats any thoughts? We could ask the same questions as we ask in database:
> 
> * How do permissions nest?
> * How do I allow users to edit, but not write new ones?

We are using exactly same as collection here we have `bucket` and `file` as `permission`	which corresponds to `collection` and `document` permission as you can see [here](https://github.com/appwrite/appwrite/blob/0dbcf3cf236a21fe9a4bb7af26f7a44ff8ca078c/app/controllers/api/storage.php#L50)

lohanidamodar:
> We are limiting size, but I can't see a limit for the amount of files uploaded. Should we consider that? I could blow up your appwrite cloud budgets in a few hours overnight.

That we are not adding as for that we will have to do full table scan on each file create api call to get the total number of files in each bucket which is a huge performance hit.

app/controllers/api/storage.php
eldadfux: Same, I don't think this operator exists.

app/controllers/api/storage.php
eldadfux: Same, I don't think this operator exists.

tests/e2e/Services/Storage/StorageConsoleClientTest.php
```diff
@@ -61,19 +73,19 @@ public function testGetStorageBucketUsage()
 
         // TODO: Uncomment once we implement check for missing bucketId in the usage endpoint.
```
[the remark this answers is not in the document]
eldadfux: +1

src/Appwrite/Utopia/Response/Model/Bucket.php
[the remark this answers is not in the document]
Meldiron: Is this model visible to a client at any point? Showing the attacker `antiVirus=false` is like saying `I dare you, try me!`.

If this is for server and console-only use, I would vote for `false` default. If I ever used this attribute, it's a security check. Security checks should not pass by default.

tests/e2e/Services/Storage/StorageConsoleClientTest.php
```diff
@@ -61,32 +73,33 @@ public function testGetStorageBucketUsage()
...
+        $response = $this->client->call(Client::METHOD_GET, '/storage/' . $bucketId .  '/usage', array_merge([
             'content-type' => 'application/json',
             'x-appwrite-project' => $this->getProject()['$id']
         ], $this->getHeaders()), [
             'range' => '24h'
         ]);
 
         $this->assertEquals($response['headers']['status-code'], 200);
-        $this->assertEquals(count($response['body']), 6);
+        $this->assertEquals(count($response['body']), 7);
```
Meldiron, line 96: Expected first, IIRC. We need to start fixing it at some point, you can be the first hero 😇

app/controllers/api/storage.php
```diff
@@ -24,9 +25,467 @@
...
+                    'lengths' => [Database::LENGTH_KEY],
+                    'orders' => [Database::ORDER_ASC],
+                ]),
+                new Document([
+                    '$id' => '_key_search',
+                    'type' => Database::INDEX_FULLTEXT,
+                    'attributes' => ['search'],
+                    'lengths' => [2048],
+                    'orders' => [Database::ORDER_ASC],
+                ],),
```
Meldiron, line 254: Isn't this a syntax error? It scares me. Can we get rid of it? 😅

app/controllers/api/storage.php
```diff
@@ -224,28 +776,54 @@
...
     ->inject('dbForInternal')
+    ->inject('dbForExternal')
     ->inject('usage')
-    ->action(function ($fileId, $response, $dbForInternal, $usage) {
+    ->inject('mode')
+    ->action(function ($bucketId, $fileId, $response, $dbForInternal, $dbForExternal, $usage, $mode) {
         /** @var Appwrite\Utopia\Response $response */
         /** @var Utopia\Database\Database $dbForInternal */
+        /** @var Utopia\Database\Database $dbForExternal */
         /** @var Appwrite\Stats\Stats $usage */
```
Meldiron, line 791: Can we add type for 'mode'? I saw it missing in multiple places.

Meldiron (commented):
Can a large file be a bit smaller? Isn't 22MB overkill? Keep in mind, anyone cloning Appwrite will have to download it.

app/controllers/api/storage.php
lohanidamodar: What are we talking about?

lohanidamodar:
@Meldiron with regards to [this](https://github.com/appwrite/appwrite/pull/1380#discussion_r766516697) comment, 
Buckets is only accessible from Server side APIs using API Key, so we should be safe?

lohanidamodar:
> Can a large file be a bit smaller? Isn't 22MB overkill? Keep in mind, anyone cloning Appwrite will have to download it.

It was before we had chunk file support, we wanted to make sure it worked for all file sizes, as we were having some issues reported regarding large file. With chunked upload support we might not need 22MB, anything above 5MB should be fine, but is it too much?
@eldadfux ?

tests/e2e/Services/Storage/StorageConsoleClientTest.php
lohanidamodar: Still not sure I wanna keep it in this PR, this PR is already huge?

Meldiron: Agree, let's refactor in separate PR and let's make sure everyone in the team is aware, so all active PRs can be reviewed with this in mind.

Meldiron:
> ],),

Yes, safe in that case. I would still go with `false` as default. We shouldn't say we scanned the file unless we are 100% sure we did.

lohanidamodar:
> > @Meldiron with regards to this comment,
> > Buckets is only accessible from Server side APIs using API Key, so we should be safe?
> 
> Yes, safe in that case. I would still go with `false` as default. We shouldn't say we scanned the file unless we are 100% sure we did.

This is not whether or not we scanned it, this is just whether or not the antiVirus scanning was enabled for the bucket, and the default value for this param is true in `createBucket` endpoint, so makes sense to keep it true as default value?

Meldiron:
> > > @Meldiron with regards to this comment,
> > > Buckets is only accessible from Server side APIs using API Key, so we should be safe?
> > 
> > 
> > Yes, safe in that case. I would still go with `false` as default. We shouldn't say we scanned the file unless we are 100% sure we did.
> 
> This is not whether or not we scanned it, this is just whether or not the antiVirus scanning was enabled for the bucket, and the default value for this param is true in `createBucket` endpoint, so makes sense to keep it true as default value?

I see. That makes sense, yes.

eldadfux:
> @Meldiron with regards to [this](https://github.com/appwrite/appwrite/pull/1380#discussion_r766516697) comment, Buckets is only accessible from Server side APIs using API Key, so we should be safe?

It should be safe.

eldadfux:
> > Can a large file be a bit smaller? Isn't 22MB overkill? Keep in mind, anyone cloning Appwrite will have to download it.
> 
> It was before we had chunk file support, we wanted to make sure it worked for all file sizes, as we were having some issues reported regarding large file. With chunked upload support we might not need 22MB, anything above 5MB should be fine, but is it too much? @eldadfux ?

I'm pretty sure the tests directory isn't getting packaged with the Docker image, would be great to double check by accessing the container image.

eldadfux:
> We are limiting size, but I can't see a limit for the amount of files uploaded. Should we consider that? I could blow up your appwrite cloud budgets in a few hours overnight.

Implementing a budget limit will be part of the cloud solution scope. It would be easier to implement it as part of usage calculation in the background, having count limit could be very expensive on large setups (count queries are a killer).

lohanidamodar:
> > > Can a large file be a bit smaller? Isn't 22MB overkill? Keep in mind, anyone cloning Appwrite will have to download it.
> > 
> > 
> > It was before we had chunk file support, we wanted to make sure it worked for all file sizes, as we were having some issues reported regarding large file. With chunked upload support we might not need 22MB, anything above 5MB should be fine, but is it too much? @eldadfux ?
> 
> I'm pretty sure the tests directory isn't getting packaged with the Docker image, would be great to double check by accessing the container image.

@eldadfux what @Meldiron meant was anyone cloning the Appwrite repository not the image. Tests is not packaged into our Appwrite docker image.

lohanidamodar:
> Can a large file be a bit smaller? Isn't 22MB overkill? Keep in mind, anyone cloning Appwrite will have to download it.

It seems I was worng, we need this file size to be > than 20 MB, because internally for now we are skipping any file size above 20 MB to be loaded at once in memory so skipping encryption, compression and antivirus for any file size above 20 MB, and want to make sure everything works as well above 22 MB

Meldiron (changes requested):

app/controllers/api/storage.php
```diff
@@ -25,9 +26,465 @@
...
+    ->inject('usage')
+    ->action(function ($search, $limit, $offset, $cursor, $cursorDirection, $orderType, $response, $dbForProject, $usage) {
+        /** @var Appwrite\Utopia\Response $response */
+        /** @var Utopia\Database\Database $dbForProject */
+        /** @var Appwrite\Stats\Stats $usage */
+
+        $queries = ($search) ? [new Query('name', Query::TYPE_SEARCH, $search)] : [];
+        
+        if (!empty($cursor)) {
+            $cursorBucket = $dbForProject->getDocument('buckets', $cursor);
```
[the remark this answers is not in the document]
lohanidamodar: We don't as this is bucket listing not file listing

src/Appwrite/Utopia/Response/Model/Bucket.php
```diff
@@ -0,0 +1,122 @@
...
+            'type' => self::TYPE_STRING,
+            'description' => 'Storage adapter.',
+            'default' => '',
+            'example' => 'local',
+        ])
+        ->addRule('adapterCredentials', [
+            'type' => self::TYPE_STRING,
+            'description' => 'Storage adapter credentials.',
+            'default' => '',
+            'example' => ['key' => 'value'],
```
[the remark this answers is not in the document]
lohanidamodar: fixed

src/Appwrite/Utopia/Response/Model/Bucket.php
```diff
@@ -0,0 +1,122 @@
...
+        ])
+        ->addRule('maximumFileSize', [
+            'type' => self::TYPE_INTEGER,
+            'description' => 'Maximum file size supported.',
+            'default' => 0,
+            'example' => 100,
+        ])
+        ->addRule('allowedFileExtensions', [
```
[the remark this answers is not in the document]
lohanidamodar: fixed

public/scripts/filters.js
```diff
@@ -183,6 +183,13 @@ window.ls.filter
 
     return $value[$value.length - 1].value;
   })
+  .add("statsGetLast", function ($value) {
```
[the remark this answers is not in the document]
lohanidamodar: fixed

app/workers/deletes.php
```diff
@@ -435,4 +438,16 @@ protected function deleteCertificates(Document $document): void
...
+
+    protected function deleteBucket(Document $document, string $projectId)
+    {
+        $bucketId = $document->getId();
+        
+        $this->deleteByGroup($bucketId . '_files',[
+            new Query('bucketId', Query::TYPE_EQUAL, [$bucketId])
+        ], $this->getExternalDB($projectId));
+
+        $device = new Local(APP_STORAGE_UPLOADS.'/app-'.$projectId);
```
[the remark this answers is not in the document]
lohanidamodar: fixed

lohanidamodar:
@eldadfux docs
image

lohanidamodar:
@eldadfux Updated screenshot
image

app/config/roles.php
```diff
@@ -13,6 +13,8 @@
     'documents.write',
     'files.read',
     'files.write',
+    'buckets.read',
+    'buckets.write',
```
TorstenDittmann, line 17: note to myself what this is doing, because of missing collection but existing projects permissions.

src/Appwrite/Messaging/Adapter/Realtime.php
```diff
@@ -295,8 +295,15 @@ public static function fromPayload(string $event, Document $payload, Document $p
                 $roles = ($collection->getAttribute('permission') === 'collection') ? $collection->getRead() : $payload->getRead();
 
                 break;
-            case strpos($event, 'storage.') === 0:
+            case strpos($event, 'storage.buckets.') === 0:
```
TorstenDittmann: reminder to remove this

lohanidamodar: Fixed https://github.com/appwrite/appwrite/pull/2786

TorstenDittmann (approved):
awesome work 👍🏻

PR #1383, opened 2021-07-12
Large File Support

lohanidamodar:
## What does this PR do?

Modifies the storage api so that it's suitable to handle large files.

## Test Plan

Relevant E2E tests on Storage service

## Related PRs and Issues

YES


app/controllers/api/storage.php
```diff
@@ -752,32 +873,50 @@
...
-            ->addHeader('Content-Security-Policy', 'script-src none;')
-            ->addHeader('X-Content-Type-Options', 'nosniff')
-            ->addHeader('Content-Disposition', 'inline; filename="'.$fileName.'"')
-            ->addHeader('Expires', \date('D, d M Y H:i:s', \time() + (60 * 60 * 24 * 45)).' GMT') // 45 days cache
-            ->addHeader('X-Peak', \memory_get_peak_usage())
-            ->send($output)
-        ;
+        if ($device->getFileSize($path) > APP_LIMIT_COMPRESSION) {          
+            $response
+                ->setContentType($contentType)
```
eldadfux, line 880: We have too many code duplications here. We can set all the common headers above this condition.

app/controllers/api/storage.php
```diff
@@ -752,32 +873,50 @@
...
+                ->addHeader('Content-Disposition', 'inline; filename="' . $fileName . '"')
+                ->addHeader('Expires', \date('D, d M Y H:i:s', \time() + (60 * 60 * 24 * 45)) . ' GMT') // 45 days cache
+                ->addHeader('X-Peak', \memory_get_peak_usage())
+                ->addHeader('Content-Length',$device->getFileSize($path))
+            ;
+            $handle = fopen($path, 'rb');
+            while(!feof($handle)) {
+                $response->chunk(@fread($handle,APP_CHUNK_SIZE), feof($handle));
+            }
+            fclose($handle);
```
eldadfux, line 892: You can return void here, (`return;`) and skip the big if<>else wrapper that makes the code a bit harder to read.

app/controllers/api/storage.php
```diff
@@ -752,32 +873,50 @@
...
+        } else {
+            $source = $device->read($path);
+            
+            if (!empty($file->getAttribute('openSSLCipher'))) { // Decrypt
```
eldadfux, line 896: I'm not sure why the `APP_LIMIT_COMPRESSION` also affect encryption<>decryption. can we de-couple them conditions?

app/controllers/api/storage.php
```diff
@@ -652,15 +756,17 @@
     ->label('sdk.methodType', 'location')
     ->param('bucketId', null, new UID(), 'Storage bucket unique ID. You can create a new storage bucket using the Storage service [server integration](/docs/server/storage#createBucket).')
     ->param('fileId', '', new UID(), 'File unique ID.')
+    ->inject('request')
     ->inject('response')
     ->inject('dbForInternal')
-    ->action(function ($bucketId, $fileId, $response, $dbForInternal) {
+    ->action(function ($bucketId, $fileId, $request, $response, $dbForInternal) {
```
eldadfux, line 762: same comments that are relevant for `/view` are also relevant here in `/download`

app/controllers/api/storage.php
```diff
@@ -306,97 +307,200 @@
...
+
+            if ($bucket->getAttribute('encryption', true) && $size <= APP_LIMIT_ENCRYPTION) {
+                $openSSLVersion = '1';
+                $openSSLCipher = OpenSSL::CIPHER_AES_128_GCM;
+                $openSSLTag = \bin2hex($tag);
+                $openSSLIV = \bin2hex($iv);
+            }
+
+            if ($file->isEmpty()) {
+                $data = [
```
eldadfux, line 432: You don't really need this var. You can just create the document instance right here.

app/controllers/api/storage.php
```diff
@@ -306,97 +307,200 @@
...
+                        ->setAttribute('algorithm', $algorithm)
+                        ->setAttribute('openSSLVersion', $openSSLVersion)
+                        ->setAttribute('openSSLCipher', $openSSLCipher)
+                        ->setAttribute('openSSLTag', $openSSLTag)
+                        ->setAttribute('openSSLIV', $openSSLIV)
+                );
+            }
+        } else {
+            if ($file->isEmpty()) {
+                $data = [
```
eldadfux, line 469: You don't really need this var. You can just create the document instance right here.

eldadfux (changes requested):

app/init.php
```diff
@@ -58,6 +58,7 @@
 const APP_LIMIT_ANTIVIRUS = 20971520; //20MB
 const APP_LIMIT_ENCRYPTION = 20971520; //20MB
 const APP_LIMIT_COMPRESSION = 20971520; //20MB
+const APP_CHUNK_SIZE = 2 * (1024 * 1024);
```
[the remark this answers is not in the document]
lohanidamodar: I think we need two size here,
1. chunk size is the return chunk size which i set 2MB
2. chunking limit, the file size above which the download/upload endpoints will chunk and this right now I'm using the compression limit, because once we compress, chunking will not work as expected

app/controllers/api/storage.php
```diff
@@ -306,97 +307,200 @@
...
+            if ($start > $end || $end > $size) {
+                throw new Exception('Invalid content-range header', 400);
+            }
+
+            if ($end == $size) {
+                $chunks = $chunk = -1;
+            } else {
+                $chunks = (int) ceil($size / ($end + 1 - $start));
+                $chunk = (int) ($start / ($end + 1 - $start));
+            }
```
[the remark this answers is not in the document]
lohanidamodar: Sure!

app/controllers/api/storage.php
lohanidamodar: This was a quick fix, but the `$device->read($path);` is required for both cases, and we should skip the read for large files that we want to chunk. And I believe if the files will be chunked it shouldn't be compressed, as we cannot compress and decompress in chunks

app/controllers/api/storage.php
```diff
@@ -306,97 +307,200 @@
...
 
         // Make sure we handle a single file and multiple files the same way
-        $file['name'] = (\is_array($file['name']) && isset($file['name'][0])) ? $file['name'][0] : $file['name'];
-        $file['tmp_name'] = (\is_array($file['tmp_name']) && isset($file['tmp_name'][0])) ? $file['tmp_name'][0] : $file['tmp_name'];
-        $file['size'] = (\is_array($file['size']) && isset($file['size'][0])) ? $file['size'][0] : $file['size'];
+        $fileName = (\is_array($file['name']) && isset($file['name'][0])) ? $file['name'][0] : $file['name'];
+        $fileTmpName = (\is_array($file['tmp_name']) && isset($file['tmp_name'][0])) ? $file['tmp_name'][0] : $file['tmp_name'];
+        $size = (\is_array($file['size']) && isset($file['size'][0])) ? $file['size'][0] : $file['size'];
+
+        $contentRange = $request->getHeader('content-range');
```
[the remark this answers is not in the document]
lohanidamodar: We want to move it to Framework Request or Swoole library request?

app/controllers/api/storage.php
lohanidamodar: done!

app/controllers/api/storage.php
lohanidamodar: updated!

app/controllers/api/storage.php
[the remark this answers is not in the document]
eldadfux: Probably both, the parsing logic can be a protected method in utopia/fw and reused in utopia/swoole

app/controllers/api/storage.php
```diff
@@ -306,97 +307,200 @@
...
+                    (int) App::getEnv('_APP_STORAGE_ANTIVIRUS_PORT', 3310));
+
+                if (!$antiVirus->fileScan($path)) {
+                    $device->delete($path);
+                    throw new Exception('Invalid file', 403);
+                }
+            }
+
+            // Compression
+            if($size <= APP_LIMIT_COMPRESSION || $size <= APP_LIMIT_ENCRYPTION) {
```
eldadfux, line 401: If one of this conditions will be true both actions will happen. We should probably decouple them to separate blocks

app/controllers/api/storage.php
```diff
@@ -672,36 +777,47 @@
...
-        $device = Storage::getDevice('files');
+        $response
+            ->setContentType($file->getAttribute('mimeType'))
+            ->addHeader('Expires', \date('D, d M Y H:i:s', \time() + (60 * 60 * 24 * 45)) . ' GMT') // 45 days cache
+            ->addHeader('X-Peak', \memory_get_peak_usage())
+            ->addHeader('Content-Disposition', 'attachment; filename="' . $file->getAttribute('name', '') . '"')
+        ;
+
+        if ($device->getFileSize($path) > APP_LIMIT_COMPRESSION) {          
+            $response->addHeader('Content-Length',$device->getFileSize($path));
```
eldadfux, line 794:
```suggestion
            $response->addHeader('Content-Length', $device->getFileSize($path));
```

eldadfux (changes requested):

app/controllers/api/storage.php
lohanidamodar: yeh, I just forgot to push the fix.

app/controllers/api/storage.php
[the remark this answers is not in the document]
lohanidamodar: There are two range headers both has can have `rangeStart` and `rangeEnd`
1. Content-range - we are using for upload
2. Range - can be used for download to return only partial part of file, I am planning to use this support resume-able downloads where users can ask only a part of the file specifying the Range start and end.
So I need a way to name these things in a way that I can support parsing both range headers in the Framework/Swoole library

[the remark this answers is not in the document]
eldadfux: `$request->getContentRangeStart()`
`$request->getContentRangeEnd()`
`$request->getContentRangeSize()`
`$request->getRangeStart()`
`$request->getRangeEnd()`
`$request->getRangeSize()`

app/init.php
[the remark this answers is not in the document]
eldadfux: Is this the same limit we use for the Swoole `Response::send()` method?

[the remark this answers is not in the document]
lohanidamodar: don't get what you mean. Response::send() does similar thing, any data above 2MB is chunked, so even if we read a 20 MB  file from storage and send using `Response::send()` as we do now, the response will still be in 2MB chunks.

But we need another limit for files when not to read them whole in memory, we can use that same limit for Compression as well, because we cannot compress/decompress without reading as a whole in the memory at the moment. Don't know yet if there might be other way.

TorstenDittmann (commented):
Nothing found, that @eldadfux hasn't addressed 👍🏻

app/http.php
```diff
@@ -20,7 +20,7 @@
 
 $http = new Server("0.0.0.0", App::getEnv('PORT', 80));
 
-$payloadSize = max(4000000 /* 4mb */, App::getEnv('_APP_STORAGE_LIMIT', 10000000 /* 10mb */));
+$payloadSize = 5 * (1024 * 1024); // 5MB
```
lohanidamodar, line 23: @eldadfux If we set this to 5 MB, Any files above 5MB will have to chunked, this will be the limit for chunking during upload from SDKs and REST API

eldadfux: Sounds like the right logic, we can re-think later if we want to increase this limit.

app/controllers/api/storage.php
```diff
@@ -171,18 +172,18 @@
             throw new Exception('Bucket not found', 404);
         }
 
-        $read ??= $bucket->getAttribute('$read', []); // By default inherit read permissions
-        $write ??= $bucket->getAttribute('$write',[]); // By default inherit write permissions
+        $read??=$bucket->getAttribute('$read', []); // By default inherit read permissions
+        $write??=$bucket->getAttribute('$write', []); // By default inherit write permissions
```
eldadfux, line 176: What is the `??` syntax used for in this way?

lohanidamodar: It's one of @TorstenDittmann 's suggestion and a good one
https://stackoverflow.com/questions/59102708/what-is-null-coalescing-assignment-operator-in-php-7-4

app/controllers/api/storage.php
```diff
@@ -171,6 +172,8 @@
             throw new Exception('Bucket not found', 404);
         }
 
+        $read??=$bucket->getAttribute('$read', []); // By default inherit read permissions
```
TorstenDittmann, line 175:
```suggestion
        $read ??= $bucket->getAttribute('$read', []); // By default inherit read permissions
```

I would leave some space for readabilty.

docs/references/storage/create-file.md
```diff
@@ -1 +1,7 @@
-Create a new file. The user who creates the file will automatically be assigned to read and write access unless he has passed custom values for read and write arguments.
\ No newline at end of file
+Create a new file. The user who creates the file will automatically be assigned to read and write access unless they have passed custom values for read and write arguments.
+
+Larger files should be uploaded using multiple requests with the [content-range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range) header to send a partial request with a maximum supported chunk of `2MB`. The `content-range` header values should always be in bytes.
```
TorstenDittmann, line 3: Was it 5 or 2 MB?

app/controllers/api/storage.php
```diff
@@ -527,93 +553,182 @@
...
         }
 
         // Save to storage
-        $size = $device->getFileSize($file['tmp_name']);
-        $path = $device->getPath(\uniqid().'.'.\pathinfo($file['name'], PATHINFO_EXTENSION));
-        $path = $bucket->getId() . '/' . $path;
-        
-        if (!$device->upload($file['tmp_name'], $path)) { // TODO deprecate 'upload' and replace with 'move'
-            throw new Exception('Failed moving file', 500);
+        $size = $size ?? $device->getFileSize($fileTmpName);
```
TorstenDittmann, line 600:
```suggestion
        $size ??= $device->getFileSize($fileTmpName);
```

app/controllers/api/storage.php
```diff
@@ -378,6 +402,8 @@
             throw new Exception('Bucket not found', 404);
         }
 
+        $read??=$bucket->getAttribute('$read', []); // By default inherit read permissions
+        $write??=$bucket->getAttribute('$write', []); // By default inherit write permissions
         $read ??= $bucket->getAttribute('$read', []); // By default inherit read permissions
         $write ??= $bucket->getAttribute('$write',[]); // By default inherit write permissions
         $read??=$bucket->getAttribute('$read', []); // By default inherit read permissions
```
TorstenDittmann, line 409:
```suggestion
        $read ??= $bucket->getAttribute('$read', []); // By default inherit read permissions
```

app/controllers/api/storage.php
```diff
@@ -389,14 +415,14 @@
         $antiVirus??=$bucket->getAttribute('antiVirus', true);
```
TorstenDittmann, line 415:
```suggestion
        $antiVirus ??= $bucket->getAttribute('antiVirus', true);
```

app/controllers/api/storage.php
```diff
@@ -527,93 +553,182 @@
...
         // Make sure we handle a single file and multiple files the same way
-        $file['name'] = (\is_array($file['name']) && isset($file['name'][0])) ? $file['name'][0] : $file['name'];
-        $file['tmp_name'] = (\is_array($file['tmp_name']) && isset($file['tmp_name'][0])) ? $file['tmp_name'][0] : $file['tmp_name'];
-        $file['size'] = (\is_array($file['size']) && isset($file['size'][0])) ? $file['size'][0] : $file['size'];
+        $fileName = (\is_array($file['name']) && isset($file['name'][0])) ? $file['name'][0] : $file['name'];
+        $fileTmpName = (\is_array($file['tmp_name']) && isset($file['tmp_name'][0])) ? $file['tmp_name'][0] : $file['tmp_name'];
+        $size = (\is_array($file['size']) && isset($file['size'][0])) ? $file['size'][0] : $file['size'];
+
+        $contentRange = $request->getHeader('content-range');
+        $fileId = $fileId == 'unique()' ? $dbForInternal->getId() : $fileId;
```
TorstenDittmann, line 561:
```suggestion
        if ($fileId == 'unique()') {
            $fileId = $dbForInternal->getId();
        }
```

Ternary operator here is a bit too much if we assign in any case.

app/controllers/api/storage.php
```diff
@@ -927,41 +1044,95 @@
...
+            } else {
+                throw new Exception('Invalid range', 416);
+            }
+        }
+
+        $source = '';
         if (!empty($file->getAttribute('openSSLCipher'))) { // Decrypt
+            $source = $device->read($path);
             $source = OpenSSL::decrypt(
                 $source,
```
TorstenDittmann, line 1099:
```suggestion
            $source = OpenSSL::decrypt(
                $device->read($path),
```

We can probably skip this assignment 👍🏻

app/controllers/api/storage.php
```diff
@@ -927,41 +1044,95 @@
...
-            ->addHeader('X-Peak', \memory_get_peak_usage())
-            ->send($source)
-        ;
+        if(!empty($rangeHeader)) {
+            $response->send($device->read($path, $rangeStart, ($rangeEnd - $rangeStart + 1)));
+        }
+
+        if ($size > APP_STORAGE_READ_BUFFER) {          
+            $response->addHeader('Content-Length', $device->getFileSize($path));
+            $chunk = 2000000; // Max chunk of 2 mb
```
TorstenDittmann, line 1129: Can we use a constant here? 🙂

app/controllers/api/storage.php
```diff
@@ -1014,37 +1187,92 @@
...
+            } else {
+                throw new Exception('Invalid range', 416);
+            }
+        }
 
+        $source = '';
         if (!empty($file->getAttribute('openSSLCipher'))) { // Decrypt
+            $source = $device->read($path);
             $source = OpenSSL::decrypt(
                 $source,
```
TorstenDittmann, line 1233:
```suggestion
            $source = OpenSSL::decrypt(
                $device->read($path),
```

app/controllers/api/storage.php
```diff
@@ -1014,37 +1187,92 @@
...
+        }
+
+        if(!empty($rangeHeader)) {
+            $response->send($device->read($path, $rangeStart, ($rangeEnd - $rangeStart + 1)));
+        }
+
+        $size = $device->getFileSize($path);
+        if ($size > APP_STORAGE_READ_BUFFER) {          
+            $response->addHeader('Content-Length', $device->getFileSize($path));
+            $chunk = 2000000; // Max chunk of 2 mb
```
TorstenDittmann, line 1269: Can we use a constant here?

app/controllers/api/storage.php
lohanidamodar: What would i name it, it's required in 2 places, 
1, when range header doesn't pass the rangeEnd and when we are just sending chunk of file as stream

app/http.php
lohanidamodar: Based on S3 multipart upload, we might want to make this at lease 6MB, as S3 explicitly requires each chunk to be at lease 5MB during multipart upload. If we want to support 5MB file chunk , this needs to be above 5 MB as request contains 5MB chunk + other data

docs/references/storage/create-file.md
lohanidamodar: Ongoing, as S3 only supports at least 5 MB chunks, we might rethink this. Right now the payload size is 5MB, but chunk has to be smaller to accomodate other data in the request

app/controllers/api/storage.php
lohanidamodar: We can't as it's required in next step, when encryption is disabled, next step will fail.

app/controllers/api/storage.php
TorstenDittmann:
```suggestion
        $write ??= $bucket->getAttribute('$write', []); // By default inherit write permissions
```

app/controllers/api/storage.php
```diff
@@ -928,41 +1043,95 @@
...
+                }
+                $rangeStart = (int) $rangeStart;
+                if(strlen($rangeEnd) == 0) {
+                    $rangeEnd =  min(($rangeStart + 2000000-1), ($size - 1));
+                } else {
+                    $rangeEnd = (int) $rangeEnd;
+                }
+                if(($rangeStart >= $rangeEnd) || $rangeEnd >= $size) {
+                    throw new Exception('Invalid range', 416);
+                }
```
TorstenDittmann, line 1082: Is it possible to make part of this in the `utopia-php` library?

I feel like we would need to do this in every endpoint with chunked downloads 🤔

app/controllers/api/storage.php
TorstenDittmann: Ah okay 👍🏻

app/controllers/api/storage.php
TorstenDittmann: Something like `MAX_OUTPUT_CHUNK_SIZE`?

app/http.php
```diff
@@ -19,7 +19,7 @@
 
 $http = new Server("0.0.0.0", App::getEnv('PORT', 80));
 
-$payloadSize = max(4000000 /* 4mb */, App::getEnv('_APP_STORAGE_LIMIT', 10000000 /* 10mb */));
+$payloadSize = 6 * (1024 * 1024); // 5MB
```
TorstenDittmann, line 22: Can we mention the 1 MB buffer here?

app/controllers/api/storage.php
lohanidamodar: may be add to framework, get range header or something?

app/http.php
lohanidamodar: We don't need to, The whole payload size is 6 MB (i see i have different in comment). So for files we keep the chunk size to 5 MB so that total request doesn't exceed the payload size.

TorstenDittmann: I know, just explain it there so people know why the comment says 5 MB and is actually 6 MB 👍🏻

app/controllers/api/storage.php
```diff
@@ -381,6 +403,8 @@
             throw new Exception('Bucket not found', 404);
         }
 
+        $read ??= $bucket->getAttribute('$read', []); // By default inherit read permissions
+        $write ??= $bucket->getAttribute('$write', []); // By default inherit write permissions
         $read ??= $bucket->getAttribute('$read', []); // By default inherit read permissions
         $write ??= $bucket->getAttribute('$write',[]); // By default inherit write permissions
```
kodumbeats, line 409: This looks...off to me. Can we get rid of the duplicate lines here?

kodumbeats (changes requested):

app/controllers/api/storage.php
lohanidamodar: Oh wow, how did this happen.

app/controllers/api/functions.php
```diff
@@ -431,42 +431,102 @@
         }
 
         // Make sure we handle a single file and multiple files the same way
-        $file['name'] = (\is_array($file['name']) && isset($file['name'][0])) ? $file['name'][0] : $file['name'];
-        $file['tmp_name'] = (\is_array($file['tmp_name']) && isset($file['tmp_name'][0])) ? $file['tmp_name'][0] : $file['tmp_name'];
-        $file['size'] = (\is_array($file['size']) && isset($file['size'][0])) ? $file['size'][0] : $file['size'];
+        $fileName = (\is_array($file['name']) && isset($file['name'][0])) ? $file['name'][0] : $file['name'];
+        $fileTmpName = (\is_array($file['tmp_name']) && isset($file['tmp_name'][0])) ? $file['tmp_name'][0] : $file['tmp_name'];
+        $size = (\is_array($file['size']) && isset($file['size'][0])) ? $file['size'][0] : $file['size'];
```
TorstenDittmann, line 436: Even though it was like this before, what do you think about breaking these 1 line ternary condition down to more verbose ifs?

I feel like they are very hard to read.

app/controllers/api/functions.php
```diff
@@ -431,42 +431,102 @@
...
-            '$read' => [],
-            '$write' => [],
-            'functionId' => $function->getId(),
-            'dateCreated' => time(),
-            'command' => $command,
-            'path' => $path,
-            'size' => $size,
-            'search' => implode(' ', [$tagId, $command]),
-        ]));
+        if($chunksUploaded == $chunks) {
```
TorstenDittmann, line 493: Lets use strict comparison `===`

app/controllers/api/storage.php
```diff
@@ -543,98 +563,188 @@
...
+        if (!empty($contentRange)) {
+            $start = $request->getContentRangeStart();
+            $end = $request->getContentRangeEnd();
+            $size = $request->getContentRangeSize();
+            $fileId = $request->getHeader('x-appwrite-id', $fileId);
+            if(is_null($start) || is_null($end) || is_null($size)) {
+                throw new Exception('Invalid content-range header', 400);
+            }
+
+            if ($end == $size) {
```
TorstenDittmann, line 584: Lets use strict comparison wherever possible `===`

app/controllers/api/storage.php
```diff
@@ -543,98 +563,188 @@
         }
 
         // Make sure we handle a single file and multiple files the same way
-        $file['name'] = (\is_array($file['name']) && isset($file['name'][0])) ? $file['name'][0] : $file['name'];
-        $file['tmp_name'] = (\is_array($file['tmp_name']) && isset($file['tmp_name'][0])) ? $file['tmp_name'][0] : $file['tmp_name'];
-        $file['size'] = (\is_array($file['size']) && isset($file['size'][0])) ? $file['size'][0] : $file['size'];
+        $fileName = (\is_array($file['name']) && isset($file['name'][0])) ? $file['name'][0] : $file['name'];
+        $fileTmpName = (\is_array($file['tmp_name']) && isset($file['tmp_name'][0])) ? $file['tmp_name'][0] : $file['tmp_name'];
+        $size = (\is_array($file['size']) && isset($file['size'][0])) ? $file['size'][0] : $file['size'];
```
TorstenDittmann, line 568: Same with the previous command, this was hard to read in the first place. This could either be solved with more verbose ifs or a function imo 👍🏻

app/controllers/api/storage.php
```diff
@@ -543,98 +563,188 @@
...
-            throw new Exception('Failed moving file', 500);
+        $size ??= $device->getFileSize($fileTmpName);
+        $path = $device->getPath($fileId . '.' . \pathinfo($fileName, PATHINFO_EXTENSION));
+        $path = str_ireplace($device->getRoot(), $device->getRoot() . DIRECTORY_SEPARATOR . $bucket->getId(), $path);
+
+        $file = $dbForInternal->getDocument('bucket_' . $bucketId, $fileId);
+
+        if (!$file->isEmpty()) {
+            $chunks = $file->getAttribute('chunksTotal', 1);
+            if ($chunk == -1) {
```
TorstenDittmann, line 618: `===`

app/controllers/api/storage.php
```diff
@@ -543,98 +563,188 @@
...
+        $chunksUploaded = $device->upload($fileTmpName, $path, $chunk, $chunks);
+        if (empty($chunksUploaded)) {
+            throw new Exception('Failed uploading file', 500);
+        }
 
-        if (App::getEnv('_APP_STORAGE_ANTIVIRUS') === 'enabled' && $bucket->getAttribute('antiVirus', true) && $size <= APP_LIMIT_ANTIVIRUS) {
-            $antiVirus = new Network(App::getEnv('_APP_STORAGE_ANTIVIRUS_HOST', 'clamav'),
+        $read = (is_null($read) && !$user->isEmpty()) ? ['user:' . $user->getId()] : $read ?? [];
+        $write = (is_null($write) && !$user->isEmpty()) ? ['user:' . $user->getId()] : $write ?? [];
+        if ($chunksUploaded == $chunks) {
```
TorstenDittmann, line 630: `===`

app/controllers/api/storage.php
```diff
@@ -543,98 +563,188 @@
...
-            $compressor = new GZIP();
-            $data = $compressor->compress($data);
-        }
-        
-        if($bucket->getAttribute('encryption', true) && $size <= APP_LIMIT_ENCRYPTION) {
-            $key = App::getEnv('_APP_OPENSSL_KEY_V1');
-            $iv = OpenSSL::randomPseudoBytes(OpenSSL::cipherIVLength(OpenSSL::CIPHER_AES_128_GCM));
-            $data = OpenSSL::encrypt($data, OpenSSL::CIPHER_AES_128_GCM, $key, 0, $iv, $tag);
-        }
+            if ($bucket->getAttribute('encryption', true) && $size <= APP_STORAGE_READ_BUFFER) {
```
TorstenDittmann, line 650: Strict comparison `<==`

app/controllers/api/storage.php
```diff
@@ -873,17 +983,17 @@
         $image = new Image($source);
 
         $image->crop((int) $width, (int) $height, $gravity);
-        
-        if (!empty($opacity) || $opacity==0) {
+
+        if (!empty($opacity) || $opacity == 0) {
```
TorstenDittmann, line 987: `===`

app/controllers/api/storage.php
```diff
@@ -951,41 +1063,87 @@
...
+            ->setContentType($file->getAttribute('mimeType'))
+            ->addHeader('Expires', \date('D, d M Y H:i:s', \time() + (60 * 60 * 24 * 45)) . ' GMT') // 45 days cache
+            ->addHeader('X-Peak', \memory_get_peak_usage())
+            ->addHeader('Content-Disposition', 'attachment; filename="' . $file->getAttribute('name', '') . '"')
+        ;
+
+        $size = $file->getAttribute('sizeOriginal', 0);
+
+        $rangeHeader = $request->getHeader('range');
+        if(!empty($rangeHeader)) {
```
TorstenDittmann, line 1087: missing whitespace after the `if`

app/controllers/api/storage.php
```diff
@@ -951,41 +1063,87 @@
...
+        $size = $file->getAttribute('sizeOriginal', 0);
+
+        $rangeHeader = $request->getHeader('range');
+        if(!empty($rangeHeader)) { 
+            $start = $request->getRangeStart();
+            $end = $request->getRangeEnd();
+            $unit = $request->getRangeUnit();
+
+            if($end == null) {
+                $end =  min(($start + 2000000-1), ($size - 1));
```
TorstenDittmann, line 1093:
```suggestion
                $end = min(($start + 2000000-1), ($size - 1));
```

app/controllers/api/storage.php
```diff
@@ -951,41 +1063,87 @@
...
+            }
+
+            if($unit != 'bytes' || $start >= $end || $end >= $size) {
```
TorstenDittmann, line 1096: Strict comparison 👍🏻

app/controllers/api/storage.php
```diff
@@ -951,41 +1063,87 @@
...
-            ->send($source)
-        ;
+        if(!empty($rangeHeader)) {
+            $response->send($device->read($path, $start, ($end - $start + 1)));
+        }
+
+        if ($size > APP_STORAGE_READ_BUFFER) {          
+            $response->addHeader('Content-Length', $device->getFileSize($path));
+            for ($i=0; $i < ceil($size / MAX_OUTPUT_CHUNK_SIZE); $i++) {
+                $response->chunk($device->read($path, ($i * MAX_OUTPUT_CHUNK_SIZE), min(MAX_OUTPUT_CHUNK_SIZE, $size - ($i * MAX_OUTPUT_CHUNK_SIZE))), (($i + 1) * MAX_OUTPUT_CHUNK_SIZE) >= $size);
```
TorstenDittmann, line 1142: Can we break this down to multiple lines for readabilty?

app/controllers/api/storage.php
```diff
@@ -1038,37 +1198,85 @@
...
+
+        $size = $file->getAttribute('sizeOriginal', 0);
+
+        $rangeHeader = $request->getHeader('range');
+        if(!empty($rangeHeader)) { 
+            $start = $request->getRangeStart();
+            $end = $request->getRangeEnd();
+            $unit = $request->getRangeUnit();
+
+            if($end == null) {
```
TorstenDittmann, line 1218: Let's user either `is_null()` or `===`

app/controllers/api/storage.php
```diff
@@ -1038,37 +1198,85 @@
...
+                $end =  min(($start + 2000000-1), ($size - 1));
+            }
 
+            if($unit != 'bytes' || $start >= $end || $end >= $size) {
```
TorstenDittmann, line 1222: Strict comparison.

app/controllers/api/storage.php
```diff
@@ -1038,37 +1198,85 @@
...
+                throw new Exception('Invalid range', 416);
+            }
+
+            $response
+                ->addHeader('Accept-Ranges', 'bytes')
+                ->addHeader('Content-Range', 'bytes ' . $start . '-' . $end . '/' . $size)
```
TorstenDittmann, line 1228: This would be more readable with template strings 👍🏻

app/controllers/api/storage.php
```diff
@@ -1038,37 +1198,85 @@
...
+        if(!empty($rangeHeader)) {
+            $response->send($device->read($path, $start, ($end - $start + 1)));
+        }
+
+        $size = $device->getFileSize($path);
+        if ($size > APP_STORAGE_READ_BUFFER) {          
+            $response->addHeader('Content-Length', $device->getFileSize($path));
+            $chunk = 2000000; // Max chunk of 2 mb
+            for ($i=0; $i < ceil($size / $chunk); $i++) {
+                $response->chunk($device->read($path, ($i * $chunk), min($chunk, $size - ($i * $chunk))), (($i + 1) * $chunk) >= $size);
```
TorstenDittmann, line 1275: Can we break this down to more verbose multiple lines of code for readability?

app/controllers/api/storage.php
```diff
@@ -1038,37 +1198,85 @@
...
-            ->send($output)
-        ;
+        if(!empty($source)) {
+            if(!empty($rangeHeader)) {
+                $response->send(substr($source, $start, ($end - $start + 1)));
+            }
+            $response->send($source);
+        }
+
+        if(!empty($rangeHeader)) {
```
TorstenDittmann, line 1266: Missing whitespace after `if`'s

TorstenDittmann (changes requested):

Meldiron:
What would SDK code look like? Do I have any control over uploading files as one or in chunks? Can I create progress bar - do I get callbacks with each chunk uploaded?

lohanidamodar:
> What would SDK code look like? Do I have any control over uploading files as one or in chunks? Can I create progress bar - do I get callbacks with each chunk uploaded?

Hi, right now we plan to handle the chunking and everything inside the SDK automatically, however if you use the REST API you will be able to do it any way you like
1. Right now in SDKs we have a basic implementation but we can have a discussion on progress bars, callbacks on each chunk etc.

docs/references/storage/create-file.md
```diff
@@ -1 +1,7 @@
-Create a new file. The user who creates the file will automatically be assigned to read and write access unless he has passed custom values for read and write arguments.
\ No newline at end of file
+Create a new file. The user who creates the file will automatically be assigned to read and write access unless they have passed custom values for read and write arguments.
```
christyjacob4, line 1:
```suggestion
Create a new file. The user that creates the file will automatically be granted read and write permissions unless they have passed custom values for read and write permissions.
```

docs/references/storage/create-file.md
```diff
@@ -1 +1,7 @@
...
+
+Larger files should be uploaded using multiple requests with the [content-range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range) header to send a partial request with a maximum supported chunk of `5MB`. The `content-range` header values should always be in bytes.
+
+When the first request is sent, the server will return the **File** object, and the subsequent part request must include the file's **id** in `x-appwrite-upload-id` header to allow the server to know the partial upload is for the existing file and not for a new one.
+
+If you're creating a new file using one of the Appwrite SDKs, the entire chunking logic will be managed by the SDK internally.
```
christyjacob4, line 7:
```suggestion
If you're creating a new file using one the Appwrite SDKs, all the chunking logic will be managed by the SDK internally.
```

docs/references/storage/create-file.md
```diff
@@ -1 +1,7 @@
-Create a new file. The user who creates the file will automatically be assigned to read and write access unless he has passed custom values for read and write arguments.
\ No newline at end of file
+Create a new file. The user who creates the file will automatically be assigned to read and write access unless they have passed custom values for read and write arguments.
+
+Larger files should be uploaded using multiple requests with the [content-range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range) header to send a partial request with a maximum supported chunk of `5MB`. The `content-range` header values should always be in bytes.
+
+When the first request is sent, the server will return the **File** object, and the subsequent part request must include the file's **id** in `x-appwrite-upload-id` header to allow the server to know the partial upload is for the existing file and not for a new one.
```
christyjacob4, line 5:
```suggestion
When the first request is sent, the server will return the **File** object, and the subsequent part request must include the file's **id** in `x-appwrite-upload-id` header to allow the server to know that the partial upload is for the existing file and not for a new one.
```

app/init.php
```diff
@@ -79,6 +79,7 @@
 const APP_STORAGE_CACHE = '/storage/cache';
 const APP_STORAGE_CERTIFICATES = '/storage/certificates';
 const APP_STORAGE_CONFIG = '/storage/config';
+const APP_STORAGE_READ_BUFFER = 20 * (1024 * 1024); //20MB other names `APP_STORAGE_MEMORY_LIMIT`, `APP_STORAGE_MEMORY_BUFFER`, `APP_STORAGE_READ_LIMIT`, `APP_STORAGE_BUFFER_LIMIT`
```
christyjacob4, line 82: Where is this used ? / What do we use it for ? Might give some ideas on which name applies better

lohanidamodar: This is used when we want to read a file into memory, any file larger than 20 mb will be streamed however any file less than 20mb in size can be loaded at once into memory

composer.json
```diff
@@ -51,8 +51,8 @@
         "utopia-php/registry": "0.5.*",
         "utopia-php/preloader": "0.2.*",
         "utopia-php/domains": "1.1.*",
-        "utopia-php/swoole": "0.2.*",
-        "utopia-php/storage": "0.5.*",
+        "utopia-php/swoole": "0.3.*",
+        "utopia-php/storage": "dev-feat-large-file-support",
```
Meldiron, line 55: Note to update version when released.

Meldiron (changes requested):
Great job, few tiny comments

app/controllers/api/storage.php
lohanidamodar: Was unable to write <== ?

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -357,9 +392,9 @@ public function testListTags(array $data):array
         ]));
 
         $this->assertEquals($function['headers']['status-code'], 200);
-        $this->assertEquals($function['body']['sum'], 1);
+        $this->assertEquals($function['body']['sum'], 2);
```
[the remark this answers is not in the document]
lohanidamodar: Not sure I want to update it in this PR.

composer.json
```diff
@@ -65,7 +65,20 @@
         "adhocore/jwt": "1.1.2",
         "slickdeals/statsd": "3.1.0"
     },
-    "repositories": [],
+    "repositories": [
```
Meldiron, line 68: It was empty, I guess this is temp config. Note to remove when possible

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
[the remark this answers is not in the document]
Meldiron: Agree, let's refactor in separate PR and let's make sure everyone in the team is aware, so all active PRs can be reviewed with this in mind.

Meldiron (approved):
LGTM

Issue #1376, opened 2021-07-10
All runtimes displayed on web UI despite default runtimes not including all runtimes


GryPr:
## 🐛 Bug Report

The web UI will display all runtimes if the _APP_FUNCTIONS_RUNTIMES variable is blank. However, the default value of _APP_FUNCTIONS_RUNTIMES does not include all the runtimes (only a subset). This leads to unexpected behavior when a runtime that isn't part of the default values is used.

## To Reproduce

1.  Set the environment variable _APP_FUNCTIONS_RUNTIMES as blank in the docker-compose environment variables
2. Access the web UI, and create a new function
3. All the runtimes will be displayed instead of just the default runtimes listed here: [https://github.com/appwrite/appwrite/blob/48272e8e5000f40096c873961d06290ae55c2af0/app/config/variables.php#L451](link)

## Expected behavior

To only show the available runtimes when creating a cloud function in the web UI

## Actual Behavior

Currently, it shows every possible runtimes.

![image](https://user-images.githubusercontent.com/51273366/125175449-3710e780-e19a-11eb-9739-8f967f18122a.png)

If the "wrong" runtime is selected and is run, then the function will get stuck, and the docker logs will display the following:

![image](https://user-images.githubusercontent.com/51273366/125175465-6162a500-e19a-11eb-99ec-b5b899bef01e.png)

## Your Environment

v:0.9.0.149 run through docker-compose.yml

PineappleIOnic:
I'm going to start looking into this now. :+1:

eldadfux:
This is now fixed in Appwrite 0.13.

PR #1386, opened 2021-07-12
[FIX] Correctly use runtime defaults with no _APP_FUNCTIONS_RUNTIMES variable

PineappleIOnic:
## What does this PR do?

This PR is a small fix so if _APP_FUNCTIONS_RUNTIMES is blank then default runtimes are loaded correctly instead of just showing all possible runtimes

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/1376

Yes, I have

### DIFF
--- a/app/config/runtimes.php
+++ b/app/config/runtimes.php
@@ -8,7 +8,13 @@ use Appwrite\Runtimes\Runtimes;
  */
 $runtimes = new Runtimes();
 
-$allowList = \explode(',', App::getEnv('_APP_FUNCTIONS_RUNTIMES', 'node-16.0,php-8.0,python-3.9,ruby-3.0,java-16.0'));
+$allowString = App::getEnv('_APP_FUNCTIONS_RUNTIMES', 'node-16.0,php-8.0,python-3.9,ruby-3.0,java-16.0');
+
+$allowList = \explode(',', $allowString);;
+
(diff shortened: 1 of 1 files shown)


app/config/runtimes.php
```diff
@@ -8,7 +8,13 @@
  */
 $runtimes = new Runtimes();
 
-$allowList = empty(App::getEnv('_APP_FUNCTIONS_RUNTIMES')) ? [] : \explode(',', App::getEnv('_APP_FUNCTIONS_RUNTIMES'));
+$allowString = App::getEnv('_APP_FUNCTIONS_RUNTIMES', 'node-16.0,php-8.0,python-3.9,ruby-3.0,java-16.0');
```
christyjacob4, line 11: I dont think we should hard code these environments here 
They should be fetched from https://github.com/appwrite/appwrite/blob/bac529a270df9c15ac471090722af7ffaad11275/app/init.php#L95

christyjacob4: You can check it's usage here
https://github.com/appwrite/appwrite/blob/a0daea0bb97ab4bbc483acb374d7c816983da54a/app/tasks/install.php#L39

christyjacob4 (changes requested):
Some suggestions

PR #1400, opened 2021-07-16
Feat sync session collection

eldadfux:
## What does this PR do?

- Adds a dedicated collection for storing use sessions.
- Fix a bug where we injected `consoleDB` instead of the new db instance
- Removed some old TODO comments
- Added trace log to the HTML error template on dev mode for easy debugging
- Fixed OAuth and Anonymous tests - @TorstenDittmann please verify

## Test Plan

Fixed some e2e tests

## Related PRs and Issues

#1281 

Yes


app/config/collections2.php
```diff
@@ -360,6 +360,254 @@
...
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => 'ip',
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 15,
```
TorstenDittmann, line 449: We should make this future proof and compatible with IP v6 🤔

app/config/collections2.php
```diff
@@ -360,6 +360,254 @@
...
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => 'provider',
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 1000,
```
TorstenDittmann, line 383: Can also be shorter.

app/controllers/api/account.php
```diff
@@ -530,6 +532,11 @@
 
         Authorization::setRole('user:'.$user->getId());
 
+        $session = $dbForInternal->createDocument('sessions', $session
+            ->setAttribute('$read', ['user:'.$user->getId()])
+            ->setAttribute('$write', ['user:'.$user->getId()])
+        );
+
         $user = $dbForInternal->updateDocument('users', $user->getId(), $user);
```
TorstenDittmann, line 540: The new session is not merged with the user yet right?

app/config/collections2.php
eldadfux: Yeh, this is not even the most optimized way to store IP in a database, just wanted to avoid any complex conversions. @kodumbeats please let me know what you think.

app/config/collections2.php
eldadfux: Yep, 128 chars sounds good?

app/controllers/api/account.php
eldadfux: We first insert the session to its own collection and then merge it to the user object on line 540

app/config/collections2.php
kodumbeats: We should set ourselves up to store both IPv4 and IPv6 in one column, so I think a string is okay for now. Looks like the largest string representation of an IPv6 address is **45 characters** ([ref](https://stackoverflow.com/questions/166132/maximum-length-of-the-textual-representation-of-an-ipv6-address))

Since each DB will handle native searching/storing IPs differently, let's do that in a separate PR.

app/config/collections2.php
eldadfux: Done

kodumbeats (approved):
:)

app/controllers/api/account.php
TorstenDittmann: Perfect 👍🏻

TorstenDittmann (approved):

PR #1413, opened 2021-07-19
fix(refactor-db): retry initial database connection on failure

kodumbeats:
## What does this PR do?

This PR adds an additional attempt to connect to the database on startup. A 5 second sleep has been insufficient in ensuring that the database is ready for connection when Appwrite starts. The `depends_on` stanza in `docker-compose` only waits for the containers to be running, _not ready_, when staggering its deployment.

~~This is the crudest approach I could think of - unsure of what an elegant solution should look like in this case.~~
Settled on a do-while loop after consulting with the team.

## Test Plan

Tests all fail if the initial database connection is not established, so the success of this PR should be measured in the consistency it provides.

## Related PRs and Issues

N/A

Yes.


app/http.php
```diff
@@ -54,8 +54,26 @@
...
+                $db = $register->get('dbPool')->get();
+                $redis = $register->get('redisPool')->get();
+                break; // leave the do-while if successful
+            } catch(\Exception $e) {
+                Console::warning("Database not ready. Retrying connection ({$attempts})...");
+                if ($attempts >= $max) {
+                    throw new \Exception('Failed to connect to database: '. $e->getMessage());
+                }
+                sleep($sleep);
+                continue;
```
christyjacob4, line 74: I think the continue here is not necessary :thinking:

app/workers/deletes.php
```diff
@@ -389,9 +389,27 @@ protected function getConsoleDB(): Database
...
+                $dbForConsole = new Database(new MariaDB($register->get('db')), $cache);
+                $dbForConsole->setNamespace('project_console_internal'); // Main DB
+                break; // leave the do-while if successful
+            } catch(\Exception $e) {
+                Console::warning("Database not ready. Retrying connection ({$attempts})...");
+                if ($attempts >= $max) {
+                    throw new \Exception('Failed to connect to database: '. $e->getMessage());
+                }
+                sleep($sleep);
+                continue;
```
christyjacob4, line 410: Same here

christyjacob4 (changes requested):
Just some minor comments

app/http.php
kodumbeats: Ah, t'was left over from a previous attempt at logic structure. Removed in 4242ba7

app/workers/deletes.php
kodumbeats: Removed in 4242ba7

christyjacob4 (approved):
Awesome :rocket:

app/workers/deletes.php
```diff
@@ -389,9 +389,26 @@ protected function getConsoleDB(): Database
     {
         global $register;
 
-        $cache = new Cache(new RedisCache($register->get('cache')));
-        $dbForConsole = new Database(new MariaDB($register->get('db')), $cache);
-        $dbForConsole->setNamespace('project_console_internal'); // Main DB
+        // wait for database to be ready
+        $attempts = 0;
```
eldadfux, line 393: A few questions:

Why do we have this logic only on this worker?
Should we apply to other workers? 
Should we apply this logic on the parent class?
Should we also apply it on `getExternalDB` and `getInternalDB`?

kodumbeats: I included this logic in the deletes worker because it's the only worker that requires a database connection near instantly on startup.
These are the logs of the deletes worker without this logic - you can see that it receives jobs to clean up executions, audit logs, and abuse logs by timestamp almost instantly, before the database is ready:

https://gist.github.com/kodumbeats/a8b80f22f09ecb733802702c28c10628

These three delete jobs need the consoleDB to search for projects here:
https://github.com/appwrite/appwrite/blob/8dab66d455a7d986847bec8edb39029ae73b4d1a/app/workers/deletes.php#L275

So, we only need the try-catch around the consoleDB for the deletes worker to confirm zero projects.

eldadfux: I think that now that we have a share [Worker class](https://github.com/appwrite/appwrite/blob/master/src/Appwrite/Resque/Worker.php) that all the worker extend we can move all the DB getters to there. It's also an Appwrite class so I think coupling it with the new DB classes is not as major issue as usually treat it to be.

If we have the retry logic on one DB adapter I don't see any reason why we shouldn't keep consistency and avoid future issues by implementing this logic on all the db getters (getConsoleDB, getInternalDB, getExternalDB)

src/Appwrite/Resque/Worker.php
```diff
@@ -26,4 +33,107 @@ public function tearDown(): void
...
+     * @param string $projectId
+     * @return Database
+     */
+    protected function getInternalDB(string $projectId): Database
+    {
+        global $register;
+
+        $attempts = 0;
+        $max = 10;
+        $sleep = 2;
```
eldadfux, line 47: I think we can move this vars to become class constants. We might also be able to create more code reuse here in the future, not urgent for this scope.

PR #1417, opened 2021-07-21
feat(refactor-db): new attribute routes for each primitive type

kodumbeats:
## What does this PR do?

This PR splits the `createAttribute` route into individual routes for primitives: strings, ints, floats, and bools. With this approach, we can:
- More easily support custom validation on attributes
- Provide proper validation for default values
- 

## Unanswered Questions:

1. SDK method names: `db.createStringAttribute` vs `db.createAttributeString` vs `db.createString`

## Test Plan

New tests for all attribute types

## Related PRs and Issues

N/A

Yes


app/controllers/api/database.php
```diff
@@ -211,6 +212,20 @@
...
+            $format = $attribute->getAttribute('format', '');
+
+            // Remove the range filter from ints/floats
+            if ($type === Database::VAR_INTEGER || $type === Database::VAR_FLOAT) {
+                if ($format) {
+                    Structure::removeFormat($format);
+                }
+            }
+        });
+
```
eldadfux, line 228: Not sure I understand what is the problem we're trying to solve here.

app/controllers/api/database.php
```diff
@@ -224,57 +239,324 @@
...
+        $collection = $dbForExternal->getCollection($collectionId);
+
+        if ($collection->isEmpty()) {
+            throw new Exception('Collection not found', 404);
+        }
+
+        /** @var string $format Name for custom range validation */
+        $format = null;
+
+        if (!is_null($min) || !is_null($max)) {
```
eldadfux, line 368: Not sure I understand what is the problem we're trying to solve here.

eldadfux (changes requested):
Have some questions added, I think comments on the Utopia PR for custom formats might help us make this PR a bit simpler.

app/controllers/api/database.php
```diff
@@ -224,79 +320,289 @@
...
 
-App::post('/v1/database/collections/:collectionId/attributes')
-    ->desc('Create Attribute')
+App::post('/v1/database/collections/:collectionId/attributes/string')
+    ->desc('Create String Attribute')
     ->groups(['api', 'database'])
     ->label('event', 'database.attributes.create')
     ->label('scope', 'attributes.write')
     ->label('sdk.namespace', 'database')
     ->label('sdk.platform', [APP_PLATFORM_SERVER])
```
eldadfux, line 329: We have stopped using the `sdk.platform` label on Appwrite 0.8. Please use `sdk.auth` instead, some relevant examples can be found in the users service controller.

eldadfux (changes requested):
Please sync with all the latest changes in the target branch

app/controllers/api/database.php
kodumbeats: Addressed in 3846d66

eldadfux (approved):

PR #1420, opened 2021-07-22
Implement Utopia Orchestration library into Appwrite

PineappleIOnic:
## What does this PR do?
This PR will replace our current docker CLI implementation with the Orchestration Library from Utopia

## Test Plan

I didn't change the tests and used them as a way to make sure the refactor was correctly working

Yes, I have


app/workers/functions.php
```diff
@@ -74,40 +63,17 @@
...
-    \parse_str($value, $container);
 
-    if(isset($container['name'])) {
-        $container = [
-            'name' => $container['name'],
-            'online' => (\substr($container['status'], 0, 2) === 'Up'),
-            'status' => $container['status'],
-            'labels' => $container['labels'],
-        ];
+array_map(function($value) use (&$list) {
```
TorstenDittmann, line 70: I wouldn't use array map if you are not returning any value in the callback. Either switch to `foreach` or return in the callback and assign the new value.

app/workers/functions.php
```diff
@@ -412,48 +371,61 @@ public function execute(string $trigger, string $projectId, string $executionId,
...
+
+            $id = $orchestration->run(
+                $runtime['image'],
+                $container,
+                "",
+                array(
+                    'tail',
+                    '-f',
+                    '/dev/null'
+                ),
```
TorstenDittmann, line 391: I would stick to the `[]` syntax instead of `array()`

app/workers/functions.php
```diff
@@ -412,48 +371,61 @@ public function execute(string $trigger, string $projectId, string $executionId,
...
+                "/usr/local/src",
+                array(),
```
TorstenDittmann, line 393: Same here with `[]` less noisy and still readable. I would suggest replacing every occasion 👍🏻

app/workers/functions.php
```diff
@@ -412,48 +371,61 @@ public function execute(string $trigger, string $projectId, string $executionId,
...
-                'labels' => [
+            $list[$container] = new Container($container, 
+                $id, 
+                "Up", 
+                array(
                     'appwrite-type' => 'function',
-                    'appwrite-created' => $executionTime,
-                ],
-            ];
+                    'appwrite-created' => "{$executionTime}"
```
TorstenDittmann, line 425:
```suggestion
                    'appwrite-created' => $executionTime
```

No need for the template literal.

app/workers/functions.php
```diff
@@ -412,48 +371,61 @@ public function execute(string $trigger, string $projectId, string $executionId,
...
+                ));
 
-            Console::info("Function created in " . ($executionEnd - $executionStart) . " seconds with exit code {$exitCode}");
+            Console::info("Function created in " . ($executionEnd - $executionStart) . " seconds");
```
TorstenDittmann, line 428: Try to not mix up `"` and `'`. I would suggest only using double quotes when template string is used and single quotes everywhere else.

TorstenDittmann (changes requested):

TorstenDittmann (approved):

app/workers/functions.php
```diff
@@ -412,48 +371,61 @@ public function execute(string $trigger, string $projectId, string $executionId,
...
 
-            $exitCodeUntar = Console::execute("docker exec ".
-                $container.
-                " sh -c 'mv /tmp/code.tar.gz /usr/local/src/code.tar.gz && tar -zxf /usr/local/src/code.tar.gz --strip 1 && rm /usr/local/src/code.tar.gz'"
-                , '', $stdout, $stderr, 60);
+            $orchestration->setCpus($cpus);
+            $orchestration->setMemory($memory);
+            $orchestration->setSwap($swap);
+
+            \array_walk($vars, function (string &$value) {
```
TorstenDittmann, line 379: Lets switch to a `foreach` here, looks like its more performant over an `array_walk`.

ref: https://steveclifton-12558.medium.com/array-walk-performance-test-b8cfb11511a8

app/workers/functions.php
```diff
@@ -412,48 +371,61 @@ public function execute(string $trigger, string $projectId, string $executionId,
...
+                $value = strval($value);
+            });
+
+            $id = $orchestration->run(
+                $runtime['image'],
+                $container,
+                '',
```
TorstenDittmann, line 386: I know this might become a little weird, but would improve readabilty by 200%.

Let's used named arguments for method such as this - where there are many arguments that doesn't really tell you what they do.

ref: https://stitcher.io/blog/php-8-named-arguments

app/workers/functions.php
TorstenDittmann: I think you missed this 👍🏻

app/workers/functions.php
```diff
@@ -412,48 +371,62 @@ public function execute(string $trigger, string $projectId, string $executionId,
...
-            if($exitCode !== 0) {
-                throw new Exception('Failed to create function environment: '.$stderr);
+
+            $orchestration->setCpus($cpus);
+            $orchestration->setMemory($memory);
+            $orchestration->setSwap($swap);
+
+            foreach($vars as $key => $value) {
+                $value = strval($value);
             }
```
kodumbeats, line 381: Do we need to pass `$value` by reference to mutate the original array items? Not sure how PHP handles this.

app/workers/functions.php
```diff
@@ -412,48 +371,62 @@ public function execute(string $trigger, string $projectId, string $executionId,
...
+                ],
+                stdout: $tarStdout, 
+                stderr: $tarStderr,
+                vars: $vars,
+                timeout: 60);
 
-            if($exitCodeUntar !== 0) {
-                throw new Exception('Failed to extract tar: '.$stderr);
+            if (!$untarSuccess) {
+                throw new Exception('Error untarring code');
```
kodumbeats, line 416: Let's not use `tar` as a verb 👍🏻

app/workers/functions.php
```diff
@@ -463,25 +436,36 @@ public function execute(string $trigger, string $projectId, string $executionId,
...
+                vars: $vars, 
+                timeout: $function->getAttribute('timeout', (int) App::getEnv('_APP_FUNCTIONS_TIMEOUT', 900)));
+        } catch (TimeoutException $e) {
+            $exitCode = 0;
+        }
 
         $executionEnd = \microtime(true);
         $executionTime = ($executionEnd - $executionStart);
-        $functionStatus = ($exitCode === 0) ? 'completed' : 'failed';
+        $functionStatus = $exitCode ? 'completed' : 'failed';
```
kodumbeats, line 456: Exit code 0 indicates a **successful** operation in UNIX, and failures are `0 < $exitCode < 256`. but this logic uses `$exitCode = 0` to indicate failure.

Let's update the logic: `$exitCode = 0` for successful operations, `$exitCode = 124` for timeouts

app/workers/functions.php
```diff
@@ -463,25 +436,36 @@ public function execute(string $trigger, string $projectId, string $executionId,
...
-        Console::info("Function executed in " . ($executionEnd - $executionStart) . " seconds with exit code {$exitCode}");
+        Console::info('Function executed in ' . ($executionEnd - $executionStart) . ' seconds, status: ' . $functionStatus);
 
         Authorization::disable();
         
         $execution = $database->updateDocument(array_merge($execution->getArrayCopy(), [
             'tagId' => $tag->getId(),
             'status' => $functionStatus,
-            'exitCode' => $exitCode,
+            'exitCode' => ($exitCode ? 0 : 1),
```
kodumbeats, line 465: Let's use `$exitCode = 0` for successful operations, `$exitCode = 124` for timeouts,

kodumbeats (changes requested):

app/workers/functions.php
```diff
@@ -463,25 +436,36 @@ public function execute(string $trigger, string $projectId, string $executionId,
...
-        Console::info("Function executed in " . ($executionEnd - $executionStart) . " seconds with exit code {$exitCode}");
+        Console::info('Function executed in ' . ($executionEnd - $executionStart) . ' seconds, status: ' . $functionStatus);
 
         Authorization::disable();
         
         $execution = $database->updateDocument(array_merge($execution->getArrayCopy(), [
             'tagId' => $tag->getId(),
             'status' => $functionStatus,
-            'exitCode' => $exitCode,
+            'exitCode' => ($exitCode === 0 ? 0 : 1),
```
kodumbeats, line 465: Why not just throw the exit code as-is? As written, we lose the context as to why it may fail (124 timeout)

kodumbeats (changes requested):

app/workers/functions.php
```diff
@@ -383,13 +344,11 @@ public function execute(string $trigger, string $projectId, string $executionId,
...
-        if(isset($list[$container]) && !$list[$container]['online']) { // Remove conatiner if not online
+        if(isset($list[$container]) &&  !(\substr($list[$container]->getStatus(), 0, 2) === 'Up')) { // Remove conatiner if not online
             $stdout = '';
             $stderr = '';
-            
-            if(Console::execute("docker rm {$container}", '', $stdout, $stderr, 30) !== 0) {
-                throw new Exception('Failed to remove offline container: '.$stderr);
-            }
+
+            $orchestration->remove($container);
```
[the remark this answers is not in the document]
kodumbeats: We're keeping the exception in `$orchestration->remove()` - should this then be in a try-catch?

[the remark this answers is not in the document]
PineappleIOnic: Yeah, should be. Let me update that now

app/workers/functions.php
```diff
@@ -29,7 +29,8 @@
 
 $dockerUser = App::getEnv('DOCKERHUB_PULL_USERNAME', null);
 $dockerPass = App::getEnv('DOCKERHUB_PULL_PASSWORD', null);
-$orchestration = new Orchestration(new DockerAPI($dockerUser, $dockerPass));
+$dockerEmail = App::getEnv('DOCKERHUB_PULL_EMAIL', null);
```
kodumbeats, line 32: If we're adding this var, it should be in the variables config as well as the changelog (Not sure where else)
ref: https://github.com/appwrite/appwrite/blob/master/app/config/variables.php

kodumbeats (changes requested):

app/workers/functions.php
```diff
@@ -412,48 +371,62 @@ public function execute(string $trigger, string $projectId, string $executionId,
...
-                " {$runtime['image']}".
-                " tail -f /dev/null"
-            , '', $stdout, $stderr, 30);
-
-            if($exitCode !== 0) {
-                throw new Exception('Failed to create function environment: '.$stderr);
+
+            $orchestration->setCpus($cpus);
+            $orchestration->setMemory($memory);
+            $orchestration->setSwap($swap);
```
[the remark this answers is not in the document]
PineappleIOnic: Yeah, Might be a good idea. I'll look into the defaults

eldadfux (approved):

kodumbeats (approved):
Looks good @PineappleIOnic , but let's make sure it gets merged to the right branch? Is `master` the target?
> PineappleIOnic wants to merge 23 commits into `appwrite:master` from `PineappleIOnic:orchestration-api-test`

PR #1422, opened 2021-07-22
Feat locale refactor

christyjacob4:
## What does this PR do?
- [x] Refactor all the locale templates
- [x] Use the new email templates instead of passing a path to `Template()` in  

https://github.com/appwrite/appwrite/blob/16b3c5c3a46e4ad5a5220898e90575a3bd95715f/app/controllers/api/account.php#L1723

## Test Plan
Use Existing tests

## Related PRs and Issues
NA

YES


app/config/locale/terms.json
```diff
@@ -0,0 +1,1274 @@
+[
+    {
+        "term": "settings.inspire",
+        "context": "",
```
eldadfux, line 4: The PR looks great, I would only think how can we make the metadata here much more useful for translator. Can we include a description? Maybe a highlighted screenshot with the relevant area in the UI once we translate the console?

christyjacob4: We can add information about a key in the "context" property like so 
 
Screenshot from 2021-07-23 14-21-53


We can attach images in the comments section like so 

Screenshot from 2021-07-23 14-23-14

eldadfux: Sounds good

app/controllers/api/account.php
```diff
@@ -967,8 +958,8 @@
             $record = $geodb->get($log['ip']);
 
             if ($record) {
-                $output[$i]['countryCode'] = (isset($countries[$record['country']['iso_code']])) ? \strtolower($record['country']['iso_code']) : '--';
-                $output[$i]['countryName'] = (isset($countries[$record['country']['iso_code']])) ? $countries[$record['country']['iso_code']] : $locale->getText('locale.country.unknown');
+                $output[$i]['countryCode'] = $locale->getText('countries.'.strtolower($record['country']['iso_code']), '--');
```
eldadfux, line 961: I think here we just need to get the country code not the translated name.

src/Appwrite/Template/Template.php
```diff
@@ -5,8 +5,57 @@
...
+     * fromHtmlString
+     *
+     * Creates a new Template() using raw HTML.
+     * 
+     * @param string $html 
+     *
+     * @return self
+     * 
+     */
+    public static function fromHtmlString(string $html): self
```
eldadfux, line 49:
```suggestion
    public static function fromString(string $html): self
```

HTML should be completely optional.

src/Appwrite/Template/Template.php
```diff
@@ -5,8 +5,57 @@
...
+    {
+        if (empty($html)) {
+            throw new Exception('Empty HTML string');
```
eldadfux, line 52:
```suggestion
            throw new Exception('Empty string');
```

src/Appwrite/Template/Template.php
```diff
@@ -125,4 +176,36 @@ public static function fromCamelCaseToDash($input): string
...
+
+    /**
+     * setHtml
+     *
+     * Set the Raw html used by this template  
+     *
+     * @return string
+     * 
+     */
+    public function setHtml(string $html): self
```
eldadfux, line 188: Don't see a reason to have a dedicated method for this. We can just set the content directly in the  `fromString` method

src/Appwrite/Template/Template.php
```diff
@@ -125,4 +176,36 @@ public static function fromCamelCaseToDash($input): string
...
+
+     /**
+     * getHtml
+     *
+     * Get the Raw html that this template holds.  
+     *
+     * @return string
+     * 
+     */
+    public function getHtml(): string
```
eldadfux, line 202: Again as the above comment, not sure why we need dedicated setters as getters, seems like there is no much code repetitiveness this is saving us. Probably better avoid coupling with HTML, this template can be used with any view format.

src/Appwrite/Template/Template.php
```diff
@@ -5,8 +5,57 @@
 use Exception;
 use Utopia\View;
 
+use function PHPUnit\Framework\isReadable;
```
eldadfux, line 8:
```suggestion
```

Leftovers

composer.json
```diff
@@ -45,7 +45,7 @@
         "utopia-php/cache": "0.2.*",
         "utopia-php/cli": "0.11.*",
         "utopia-php/config": "0.2.*",
-        "utopia-php/locale": "0.3.*",
+        "utopia-php/locale": "dev-feat-accept-file-path",
```
eldadfux, line 48: We can set it to v0.4.0

app/controllers/api/users.php
```diff
@@ -322,8 +318,8 @@
             $record = $geodb->get($log['ip']);
 
             if ($record) {
-                $output[$i]['countryCode'] = (isset($countries[$record['country']['iso_code']])) ? \strtolower($record['country']['iso_code']) : '--';
-                $output[$i]['countryName'] = (isset($countries[$record['country']['iso_code']])) ? $countries[$record['country']['iso_code']] : $locale->getText('locale.country.unknown');
+                $output[$i]['countryCode'] = $locale->getText('countries.'.strtolower($record['country']['iso_code']), '--');
```
eldadfux, line 321: Again, this portion of code just needs the country code.

app/controllers/api/locale.php
```diff
@@ -42,7 +40,8 @@
 
         if ($record) {
             $output['countryCode'] = $record['country']['iso_code'];
-            $output['country'] = (isset($countries[$record['country']['iso_code']])) ? $countries[$record['country']['iso_code']] : $locale->getText('locale.country.unknown');
+            $output['country'] = $locale->getText('countries.'.strtolower($record['country']['iso_code']), $locale->getText('locale.country.unknown'));
```
eldadfux, line 43: Why did we gave up the isset method check? We better be safe than sorry :)

eldadfux: This has multiple instances across the diff, I'll avoid commenting on each one :)

tests/e2e/Services/Locale/LocaleBase.php
```diff
@@ -159,8 +159,8 @@ public function testGetContinents():array
         $this->assertEquals($response['headers']['status-code'], 200);
         $this->assertEquals(7, $response['body']['sum']);
         $this->assertIsArray($response['body']['continents']);
-        $this->assertEquals($response['body']['continents'][0]['code'], 'NA');
-        $this->assertEquals($response['body']['continents'][0]['name'], 'América del Norte');
+        $this->assertEquals($response['body']['continents'][4]['code'], 'NA');
```
eldadfux, line 162: We should make sure we're sorting the list using the locale ABC

eldadfux (changes requested):

app/controllers/api/account.php
christyjacob4: Done :+1:

src/Appwrite/Template/Template.php
christyjacob4: Done :+1:

src/Appwrite/Template/Template.php
christyjacob4: done :+1:

src/Appwrite/Template/Template.php
christyjacob4: Yes but I've defined `fromString` to be a static method. `setHtml()` would act on an instance. Let me know if this is not required.

src/Appwrite/Template/Template.php
christyjacob4: done :+1:

composer.json
christyjacob4: Done :+1:

app/controllers/api/users.php
christyjacob4: Screenshot from 2021-07-26 10-47-03


I dont think so... 
`$countries = $locale->getText('countries'); `

which returns 
```php
[
   "IN" => "India"
]
```

christyjacob4: Done :+1:

app/controllers/api/locale.php
christyjacob4: `$locale->getText()` takes a default param. So I thought we can use that instead :+1:

tests/e2e/Services/Locale/LocaleBase.php
christyjacob4: Done :+1:

PR #1463, opened 2021-08-04
feat-swagger-update

lohanidamodar:
## What does this PR do?

Modify Swagger spec to provide x-type attribute for Queries object for SDK generator

## Test Plan

N/A

## Related PRs and Issues

https://github.com/appwrite/sdk-generator/pull/234

YES

### DIFF
--- a/src/Appwrite/Specification/Format/Swagger2.php
+++ b/src/Appwrite/Specification/Format/Swagger2.php
@@ -225,6 +225,10 @@ class Swagger2 extends Format
                     'required' => !$param['optional'],
                 ];
 
+                if($name == 'queries') {
+                    $node['x-type'] = 'query';
+                }
+
                 switch ((!empty($validator)) ? \get_class($validator) : '') {
                     case 'Utopia\Validator\Text':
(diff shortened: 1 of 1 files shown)


TorstenDittmann (approved):

src/Appwrite/Specification/Format/Swagger2.php
```diff
@@ -225,6 +225,10 @@ public function parse(): array
                     'required' => !$param['optional'],
                 ];
 
+                if($name == 'queries') {
```
eldadfux, line 228: We should never trust a variable/parameter name for business logic. What we can do is leverage the param instance type for differentiation, just like we do below this condition. 👇

lohanidamodar: Instance type? the param instance type I think is the validator type right? Which in this case is ArrayList, how will that help me determine if it's a query?

lohanidamodar:
Closing this, as this is not required at the moment. We will not be using this for docs at the moment.

PR #1480, opened 2021-08-12
feat(refactor-db) enforce either document or collection permissions

kodumbeats:
## What does this PR do?

This PR adds support for separate enforcement of collection and document permissions. To do so, this PR:
- Makes the changes to Utopia to be able to skip permission validation where unneeded ([on branch](https://github.com/kodumbeats/database/tree/dev))
- Uses the `Authorization::skip()` method to skip auth when satisfying collection permissions. (https://github.com/utopia-php/database/pull/55)

The document that we pass to `Authorization::__construct()` isn't used at all in the Authorization validator. We could use this to instead pass the collection and automatically skip when needed.

## Test Plan

Tests have been added.
- [x] test listDocuments
- [ ] add more permission tests

## Related PRs and Issues

N/A

Yes.


app/controllers/api/database.php
```diff
@@ -1206,6 +1230,7 @@
             throw new Exception($validator->getDescription(), 400);
         }
 
+        $afterDocument = null;
```
TorstenDittmann, line 1233: Why this addition?

If `$afterDocument` is never assigned, `$afterDocument ?? null` will result in `null` anyway 👍🏻

tests/e2e/Services/Database/DatabaseBase.php
```diff
@@ -1090,6 +1102,144 @@ public function testDefaultPermissions(array $data):array
...
+    {
+        $user = 'user:' . $this->getUser()['$id'];
+        $collection = $this->client->call(Client::METHOD_POST, '/database/collections', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'x-appwrite-key' => $this->getProject()['apiKey']
+        ]), [
+            'collectionId' => 'unique()',
+            'name' => 'enforceCollectionPermissions',
+            'permission' => 'collection',
```
TorstenDittmann, line 1116: Reminder for myself to address this in the migration 👍🏻

app/controllers/api/database.php
kodumbeats: Ah, you're right - I removed it in e1c7e49

lohanidamodar (approved):
Looks good to me!

TorstenDittmann (approved):

PR #1481, opened 2021-08-12
feat-usage-daemon

lohanidamodar:
## What does this PR do?

Introduces a Usage daemon that syncs the data from InfluxDB to Appwrite's stats collection for each project

## Open Question
- Should we count the number of operations from DB workers (for example, when a collection is deleted, should we count the document delete operations that happen for that collection as well?

## Test Plan

## Related PRs and Issues

YES


app/workers.php
```diff
@@ -32,3 +32,16 @@
...
+
+    if (empty($host) || empty($port)) {
+        return;
+    }
+    $driver = new InfluxDB\Driver\Curl("http://{$host}:{$port}");
+    $client = new InfluxDB\Client($host, $port, '', '', false, false, 5);
+    $client->setDriver($driver);
+
+    return $client;
+});
```
TorstenDittmann, line 47: We might wanna move this to the `usage.php`, since its only used there and in no other worker.

app/tasks/usage.php
```diff
@@ -0,0 +1,191 @@
...
+        foreach ($projects as $project) {
+            $projectIds[$project['$id']] = true;
+        }
+        $projects = null;
+
+        $latestData = [];
+        $dbForProject = new Database(new MariaDB($register->get('db')), $cacheAdapter);
+        foreach ($projectIds as $id => $value) {
+            $dbForProject->setNamespace("project_{$id}_internal");
+            $doc = $dbForProject->find('stats', [new Query("period", Query::TYPE_EQUAL, ["1d"])], 1, 0, ['time'], [Database::ORDER_DESC]);
```
TorstenDittmann, line 42: You can try to use `findOne`. Then you don't need the `reset` anymore.

docker-compose.yml
```diff
@@ -344,6 +344,37 @@ services:
       - _APP_MAINTENANCE_RETENTION_ABUSE
       - _APP_MAINTENANCE_RETENTION_AUDIT
 
+  appwrite-usage:
```
eldadfux, line 347: This change should also be synced to compose.phtml file

app/tasks/usage.php
```diff
@@ -0,0 +1,187 @@
+<?php
+
+global $cli, $register;
+
+require_once __DIR__ . '/../workers.php';
```
eldadfux, line 5: As this is not a worker, I think we should avoid using the workers init file.

eldadfux (changes requested):

app/views/install/compose.phtml
```diff
@@ -288,6 +288,26 @@ services:
...
+    environment:
+      - _APP_ENV
+      - _APP_DB_HOST
+      - _APP_DB_PORT
+      - _APP_DB_SCHEMA
+      - _APP_DB_USER
+      - _APP_DB_PASS
+      - _APP_INFLUXDB_HOST
+      - _APP_INFLUXDB_PORT
+      - _APP_USAGE_SYNC_INTERVAL
```
eldadfux, line 310:
```suggestion
      - _APP_USAGE_AGGREGATION_INTERVAL
```

eldadfux: This new var should also be documented

PR #1484, opened 2021-08-13
feat(refactor-db): response models for new attribute types

kodumbeats:
## What does this PR do?

This PR splits the default `Attribute` response model into the different formats/primitive routes we now support. A couple of interesting things to get there:

- Refactored `attributesCallback` into proper `createAttribute` function to clean up the story around what happens to each attribute:
    https://github.com/appwrite/appwrite/blob/c3534b43835f7d4ec37c2911dc6ffa99f7cd6ff8/app/controllers/api/database.php#L47
- We have to pass a clone of the attribute document to the workers - since we manipulate the `$attribute` document after the function, we end up in a race condition when trying to write to internal `attributes` table
    https://github.com/appwrite/appwrite/blob/c3534b43835f7d4ec37c2911dc6ffa99f7cd6ff8/app/controllers/api/database.php#L100-L108
- Leveraged `Database::addFilter()` in `app/init.php` to manipulate what was going in and coming out of Utopia\Database layer. `defaultValue` normalizes and denormalizes the value in a way to preserve typecasting. `range` removes `min` and `max` from the document to avoid a `StructureException` when writing to internal attributes table.
- We weren't making use of the rule "require" field, so I put it to work for default values:
    https://github.com/appwrite/appwrite/blob/c3534b43835f7d4ec37c2911dc6ffa99f7cd6ff8/src/Appwrite/Utopia/Response.php#L329
- Added a callback param when creating model rules to know how to select the correct nested response model:
    https://github.com/appwrite/appwrite/blob/c3534b43835f7d4ec37c2911dc6ffa99f7cd6ff8/src/Appwrite/Utopia/Response/Model.php#L69-L78

## Test Plan

- [x] How should we implement tests for this?

- E2E tests to ensure the response model from `createAttribute`, `getAttribute`, `listAttributes`, and `getCollection` all have proper response models.

## Related PRs and Issues

https://github.com/utopia-php/database/pull/69

Yes.


eldadfux:
> How should we implement tests for this?

I guess we just need to have more assertions for different response structures.

src/Appwrite/Utopia/Response.php
```diff
@@ -11,19 +11,24 @@
...
 use Appwrite\Utopia\Response\Model\Document as ModelDocument;
 use Appwrite\Utopia\Response\Model\Domain;
+use Appwrite\Utopia\Response\Model\Email;
 use Appwrite\Utopia\Response\Model\Error;
 use Appwrite\Utopia\Response\Model\ErrorDev;
 use Appwrite\Utopia\Response\Model\Execution;
 use Appwrite\Utopia\Response\Model\File;
 use Appwrite\Utopia\Response\Model\Func;
 use Appwrite\Utopia\Response\Model\Index;
+use Appwrite\Utopia\Response\Model\IP;
```
eldadfux, line 31: I would go with AttributeIP, let's add a prefix to all attributes.

src/Appwrite/Utopia/Response.php
```diff
@@ -62,13 +68,27 @@ class Response extends SwooleResponse
...
+    const MODEL_ATTRIBUTE = 'attribute';
+    const MODEL_ATTRIBUTE_LIST = 'attributeList';
+    const MODEL_ATTRIBUTE_STRING = 'attributeString';
+    const MODEL_ATTRIBUTE_STRING_LIST = 'attributeStringList';
+    const MODEL_INTEGER= 'integer';
+    const MODEL_INTEGER_LIST= 'integerList';
+    const MODEL_FLOAT= 'float';
+    const MODEL_FLOAT_LIST= 'floatList';
+    const MODEL_EMAIL= 'email';
+    const MODEL_EMAIL_LIST= 'emailList';
```
eldadfux, line 86: Same again, let's add a prefix. Email, IP, URL are way too generic.

src/Appwrite/Utopia/Response/Model/AttributeFloat.php
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    /**
+     * Get Name
+     * 
+     * @return string
+     */
+    public function getName():string
+    {
+        return 'Float';
```
eldadfux, line 39:
```suggestion
        return 'AttributeFloat';
```

Will help avoid any future conflicts

eldadfux: Same for other "child" attributes

eldadfux (changes requested):

src/Appwrite/Utopia/Response/Model/AttributeEmail.php
```diff
@@ -0,0 +1,51 @@
...
+
+use Appwrite\Utopia\Response;
+use Appwrite\Utopia\Response\Model\AttributeString;
+
+class AttributeEmail extends AttributeString
+{
+    public function __construct()
+    {
+        $this
+            ->addRule('format', [
```
eldadfux, line 13: We need to normalize the structure and abstract the complexity of extending custom attributes. I don't think this should be an exact representation of the DB lib or the internal data structure.

I would use `type = email` instead of `type=string` and drop the format attribute.

src/Appwrite/Utopia/Response/Model/AttributeFloat.php
```diff
@@ -0,0 +1,55 @@
...
+
+use Appwrite\Utopia\Response;
+use Appwrite\Utopia\Response\Model\Attribute;
+
+class AttributeFloat extends Attribute
+{
+    public function __construct()
+    {
+        $this
+            ->addRule('format', [
```
eldadfux, line 13: Same comment about normalization. Since 0.9 we also try to stay as close as possible to primitives (and successfully so far 😉). Instead of `format` attribute let's add `min`, and `max`. Attribute type can stay `float`. We can adjust the structure in the action after saving to the DB.

src/Appwrite/Utopia/Response/Model/AttributeIP.php
```diff
@@ -0,0 +1,51 @@
...
+use Appwrite\Utopia\Response;
+use Appwrite\Utopia\Response\Model\AttributeString;
+
+class AttributeIP extends AttributeString
+{
+    public function __construct()
+    {
+        $this
+            ->addRule('format', [
+                'type' => self::TYPE_STRING,
```
eldadfux, line 14: Same as above. Type can be `ip` instead of `string`

src/Appwrite/Utopia/Response/Model/AttributeInteger.php
```diff
@@ -0,0 +1,55 @@
...
+use Appwrite\Utopia\Response;
+use Appwrite\Utopia\Response\Model\Attribute;
+
+class AttributeInteger extends Attribute
+{
+    public function __construct()
+    {
+        $this
+            ->addRule('format', [
+                'type' => self::TYPE_STRING,
```
eldadfux, line 14: Should probably be very similar to the float model.

src/Appwrite/Utopia/Response/Model/AttributeURL.php
```diff
@@ -0,0 +1,51 @@
...
+use Appwrite\Utopia\Response\Model\AttributeString;
+
+class AttributeURL extends AttributeString
+{
+    public function __construct()
+    {
+        $this
+            ->addRule('format', [
+                'type' => self::TYPE_STRING,
+                'description' => 'String format.',
```
eldadfux, line 15: Same as above. Type can be `url` instead of `string`.

eldadfux (changes requested):

src/Appwrite/Utopia/Response/Model/AttributeEmail.php
kodumbeats: Now that the code is more complete, would you still like to make this change? Now that `format` is a string, it fits a lot better with `type`, highlighting how we are sticking to primitives in the DB.

eldadfux (changes requested):

app/config/collections2.php
```diff
@@ -174,7 +174,7 @@
                 'required' => false,
                 'default' => null,
                 'array' => false,
-                'filters' => [],
+                'filters' => ['defaultValue'],
```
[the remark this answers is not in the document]
kodumbeats: 53b8a1c

src/Appwrite/Utopia/Response/Model/AttributeBoolean.php
```diff
@@ -0,0 +1,45 @@
...
+     * 
+     * @return string
+     */
+    public function getName():string
+    {
+        return 'AttributeBoolean';
+    }
+
+    /**
+     * Get Collection
```
kodumbeats, line 37:
```suggestion
     * Get Type
```

src/Appwrite/Utopia/Response/Model/AttributeBoolean.php
```diff
@@ -0,0 +1,45 @@
...
+        parent::__construct();
+
+        $this
+            ->addRule('default', [
+                'type' => self::TYPE_BOOLEAN,
+                'description' => 'Default value for attribute when not provided. Cannot be set when attribute is required.',
+                'default' => null,
+                'example' => false,
+                'array' => false,
+                'require' => false,
```
kodumbeats, line 21: Check

src/Appwrite/Utopia/Response/Model/AttributeEmail.php
```diff
@@ -0,0 +1,53 @@
...
+     * 
+     * @return string
+     */
+    public function getName():string
+    {
+        return 'AttributeEmail';
+    }
+
+    /**
+     * Get Collection
```
kodumbeats, line 45:
```suggestion
     * Get Type
```

src/Appwrite/Utopia/Response/Model/AttributeIP.php
```diff
@@ -0,0 +1,53 @@
...
+     * 
+     * @return string
+     */
+    public function getName():string
+    {
+        return 'AttributeIP';
+    }
+
+    /**
+     * Get Collection
```
kodumbeats, line 45:
```suggestion
     * Get Type
```

src/Appwrite/Utopia/Response/Model/AttributeString.php
```diff
@@ -0,0 +1,51 @@
...
+     * 
+     * @return string
+     */
+    public function getName():string
+    {
+        return 'AttributeString';
+    }
+
+    /**
+     * Get Collection
```
kodumbeats, line 43:
```suggestion
     * Get Type
```

src/Appwrite/Utopia/Response/Model/AttributeInteger.php
```diff
@@ -0,0 +1,60 @@
...
+     * Get Name * 
+     * @return string
+     */
+    public function getName():string
+    {
+        return 'AttributeInteger';
+    }
+
+    /**
+     * Get Collection
```
kodumbeats, line 52:
```suggestion
     * Get Type
```

src/Appwrite/Utopia/Response/Model/AttributeFloat.php
```diff
@@ -0,0 +1,61 @@
...
+     * 
+     * @return string
+     */
+    public function getName():string
+    {
+        return 'AttributeFloat';
+    }
+
+    /**
+     * Get Collection
```
kodumbeats, line 53:
```suggestion
     * Get Type
```

src/Appwrite/Utopia/Response/Model/AttributeList.php
```diff
@@ -0,0 +1,62 @@
...
+     *
+     * @return string
+     */
+    public function getName():string
+    {
+        return 'Attributes List';
+    }
+
+    /**
+     * Get Collection
```
kodumbeats, line 54:
```suggestion
     * Get Type
```

src/Appwrite/Utopia/Response/Model/AttributeList.php
```diff
@@ -0,0 +1,62 @@
...
+                        self::TYPE_STRING => match($attribute->getAttribute('format')) {
+                            APP_DATABASE_ATTRIBUTE_EMAIL => Response::MODEL_ATTRIBUTE_EMAIL,
+                            APP_DATABASE_ATTRIBUTE_IP => Response::MODEL_ATTRIBUTE_IP,
+                            APP_DATABASE_ATTRIBUTE_URL => Response::MODEL_ATTRIBUTE_URL,
+                            default => Response::MODEL_ATTRIBUTE_STRING,
+                        },
+                        default => Response::MODEL_ATTRIBUTE,
+                    };
+                },
+            ])
```
TorstenDittmann, line 39: In terms of syntax, it would be awesome to solve it with something like this:

```php
->addRule('attributes', [
    'type' => [
        Response::MODEL_ATTRIBUTE_BOOLEAN,
        Response::MODEL_ATTRIBUTE_INTEGER,
        Response::MODEL_ATTRIBUTE_FLOAT,
        Response::MODEL_ATTRIBUTE_STRING,
        Response::MODEL_ATTRIBUTE_EMAIL,
        Response::MODEL_ATTRIBUTE_URL,
        Response::MODEL_ATTRIBUTE_IP
    ],
    'description' => 'List of attributes.',
    'default' => [],
    'array' => true,
])
```

Since we need to reflect it later on the Swagger Specifications as well. There we need all possible models as well.

But for this to work, we need to have an identifier present in each Mode.

lohanidamodar: Agreed, we need some attribute to parse the response models correctly later in SDKs

kodumbeats: Fixed via https://github.com/appwrite/appwrite/pull/1624

src/Appwrite/Utopia/Response/Model.php
```diff
@@ -8,7 +8,7 @@ abstract class Model
 {
     const TYPE_STRING = 'string';
     const TYPE_INTEGER = 'integer';
-    const TYPE_FLOAT = 'float';
+    const TYPE_FLOAT = 'double';
```
eldadfux, line 11: I still think we should use float here to avoid coupling ourself with legacy PHP behavior. We can probably cast our own `gettype` method that behaves correctly. Let me know what you think.

app/config/collections2.php
```diff
@@ -214,7 +214,7 @@
                 'required' => false,
                 'default' => new stdClass,
                 'array' => false,
-                'filters' => ['json'],
+                'filters' => ['json', 'range'],
```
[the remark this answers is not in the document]
kodumbeats: We needed some way to know if/when to extract data from formatOptions to modify documents to fit their response models. Additionally, this filter gives us a way to modify the document when saving to the DB, so we can adjust for the Structure validator.
The range filter prepares min and max for ints and floats:
https://github.com/appwrite/appwrite/blob/53b8a1c245800b9df14c4dfa70327f93a692bcf3/app/init.php#L199-L219

src/Appwrite/Utopia/Response/Model.php
kodumbeats: I don't think it's a good idea to wrap a method in the standard library - I think it'd be hard to maintain and onboard.

PR #1498, opened 2021-08-18
fix(refactor-db): catch duplicate exception when creating and updating documents

kodumbeats:
## What does this PR do?

This PR catches `\Utopia\Database\Exception\Duplicate` when creating or updating documents. This catches the enforcement of unique indexes.

## Test Plan

Tests included in the PR.

## Related PRs and Issues

https://github.com/utopia-php/database/pull/63 must be released for these tests to pass.

Yes.

### DIFF
--- a/app/controllers/api/database.php
+++ b/app/controllers/api/database.php
@@ -1307,6 +1307,9 @@ App::patch('/v1/database/collections/:collectionId/documents/:documentId')
         catch (AuthorizationException $exception) {
             throw new Exception('Unauthorized permissions', 401);
         }
+        catch (DuplicateException $exception) {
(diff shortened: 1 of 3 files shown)


app/controllers/api/database.php
```diff
@@ -1145,6 +1145,9 @@
         catch (StructureException $exception) {
             throw new Exception($exception->getMessage(), 400);
         }
+        catch (DuplicateException $exception) {
+            throw new Exception($exception->getMessage(), 400);
```
eldadfux, line 1149: Should we use 400 or 409 for this error?

kodumbeats: 409 makes sense 👍🏻 will update

PR #1547, opened 2021-08-29
feat-users-update

lohanidamodar:
## What does this PR do?

Adds new endpoint and related E2E tests to do the following in Users service
- Update user name
- Update user email
- Update user password

## Test Plan

Relevant E2E tests on UsersBase.php

## Related PRs and Issues

(If this PR is related to any other PR or resolves any issue or related to any issue link all related PR and issues here.)

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

YES

PR #1548, opened 2021-08-29
feat-users-update-new

lohanidamodar:
## What does this PR do?

Adds new endpoint and related E2E tests to do the following in Users service
- Update user name
- Update user email
- Update user password

Also update audit resource by removing the `service` prefix.

## Test Plan

Relevant E2E tests on UsersBase.php

## Related PRs and Issues

https://github.com/appwrite/appwrite/pull/1547

YES


eldadfux (approved):

app/controllers/api/users.php
```diff
@@ -376,6 +377,131 @@
...
+        }
+
+        $email = \strtolower($email);
+        $profile = $dbForInternal->findOne('users', [new Query('email', Query::TYPE_EQUAL, [\strtolower($email)])]); // Get user by email address
+
+        if ($profile) {
+            throw new Exception('Email already exists', 409);
+        }
+
+        $user = $dbForInternal->updateDocument('users', $user->getId(), $user->setAttribute('email', $email));
```
eldadfux, line 494: We can try - catch for `Duplicate` exception instead of checking it before hand. Some good examples can be found in the DB controller.

lohanidamodar: Should I update the `account` service in similar way as well?

eldadfux: Yeh, sure

PR #1554, opened 2021-08-30
Moved project json attributes to their own collection

Meldiron:
## What does this PR do?

Defines new collections for:

- [X] projectWebhooks
- [X] projectKeys
- [X] projectProviders
- [X] projectDomains
- [X] projectsPlatforms
- [X] projectsServices

These attributes in `collection` table are no longer used and set to `null`.

This PR also defines filters that tell database how to "join" these tables into project request. These filters are:

- [X] subQueryProjectDomains
- [X] subQueryProjectPlatforms
- [X] subQueryProjectServices
- [X] subQueryProjectWebhooks
- [X] subQueryProjectKeys
- [X] subQueryProjectProviders

## Test Plan

Current tests should cover almost all scenarios. While writing the code, I found these possible problems:

- There isn't test for failure (404) when updating project platform, I think
- There are no tests for pagination described as first point below in technical questions.. Do we implement them?
- This was somehow okay, tests didn't see it as a problem: https://cln.sh/kUds4S
- With the error `Data too long for column 'secret'` it returns status code 200 on `Project create key`. Might this be some deep problem?
- Tests don't really test for filters.. I had bug in filters and all tests were green. I think we need to add one more test at the end to list all changes we did (to webhooks, services, providers...) using the main project API route and check if everything is present in there

## Technical questions

- If I want  do `.list()` for sub-collection.. I will check if the `$projectId` exists.. But do I also need to add this filter to the `.list` queries array?
- This refactor introduced a lot of new `.find()` queries which requires `limit` parameter. You can find these in filters or REST API itself. What should we do about them? Should we introduce pagination to all HTTP routes where we need it? What about those limits in database filters (joins)?
- The responses are typed in it's own file, for example `Platform.php`. Do I need to do anything with them regarding this refactor?
- I copied `write` and `read` permissions from `createProject` as a code... Is that correct or should I fetch them from document itself?
- If the value of an attribute that is joined using the filter is set to `[]`, it ignores the filter.. It simply returns `[]`.. So all attributes needs to be set to `null` in order for filter to work. Do we need to write this somewhere for future reference?
- With auths and services, when do we delete the document? Currently, never
- How should the `auths` collection look like? Currently, it is this if I understand it:

```
[
  limit => 5,
  google => true,
  facebook => true,
  discord => false,
  linkedin => true
]
```

We need collection like:

```
key: string;
value: boolean/number
```

Or we need to move `limit` somewhere else, for example directly as a parameter of a project.


app/config/collections2.php
```diff
@@ -558,6 +558,471 @@
         ],
     ],
 
+    'projectsPlatforms' => [
```
kodumbeats, line 561: I think to fit with current patterns, we should omit the `projects` prefix on these tables.

kodumbeats (commented):
Quite a solid draft! Left just a few comments directly in the code, and I'll answer your specific questions here:

> There isn't test for failure (404) when updating project platform, I think

That could very well be true. If you find a missing test, let's add it! :)

> There are no tests for pagination described as first point below in technical questions.. Do we implement them?

Answered in related question.

> This was somehow okay, tests didn't see it as a problem

Do we have tests around domains?

> With the error Data too long for column 'secret' it returns status code 200 on Project create key

I've never seen this error - can we maybe document it in a separate issue? 

> Tests don't really test for filters.. I had bug in filters and all tests were green. I think we need to add one more test at the end to list all changes we did (to webhooks, services, providers...) using the main project API route and check if everything is present in there

Maybe we could consider some way to unit test them? I'm not sure I like the idea of adding a final test to check the state after a test suite, as each test should _ideally_ only test for one behavior. Might be hard to maintain 🤔 

> If I want do .list() for sub-collection.. I will check if the $projectId exists.. But do I also need to add this filter to the .list queries array?

The chance of collision is low, but non-zero. Let's include a query for `$projectId`

> This refactor introduced a lot of new .find() queries which requires limit parameter. You can find these in filters or REST API itself. What should we do about them? Should we introduce pagination to all HTTP routes where we need it? What about those limits in database filters (joins)?

I don't think we need to change the API signature with these changes, but happy to consider further. I believe we'll be fine to just set the limit high enough so one project won't hit the internal query limit (5000 documents). @eldadfux could answer this more definitively.

> The responses are typed in it's own file, for example Platform.php. Do I need to do anything with them regarding this refactor?

If we wanted to change the response signature, we would update these files. Since we don't plan to change the response model, you shouldn't need to modify them.

> I copied write and read permissions from createProject as a code... Is that correct or should I fetch them from document itself?

In each of these routes, the first call we make to the database is to get the project:
`$project = $db>getDocument('projects', $projectId);`
The `getDocument()` method checks for permissions in the Database adapter, and will throw a 404 if the query returns no results. I think we can/should rely on the read/write permissions of the project, and store the "subquery" tables without permissions.

> If the value of an attribute that is joined using the filter is set to [], it ignores the filter.. It simply returns [].. So all attributes needs to be set to null in order for filter to work. Do we need to write this somewhere for future reference?

Maybe as a comment either when defining the schema `app/config/collections.php` or when defining the filters `app/init.php`

> With auths and services, when do we delete the document? Currently, never

I'm not sure we need to delete them, since it's basically just storing project state.

> Or we need to move limit somewhere else, for example directly as a parameter of a project.

I think we'll probably have to keep this limit on the project document. 🤔

eldadfux:
@Meldiron `projectsServices` and `projectsAuths` should stay the same, they are just custom configurations and not children documents of the project document.

app/controllers/api/projects.php
```diff
@@ -468,24 +472,51 @@
...
 
-        $services = $project->getAttribute('services', []);
-        $services[$service] = $status;
+        $service = $dbForConsole->findOne("projectsServices", [
+            new Query('projectId', Query::TYPE_EQUAL, [$project->getId()]),
+            new Query('key', Query::TYPE_EQUAL, [$serviceKey]),
+        ]);
+
+        // TODO: Implement delete method. Do we delete for "true" or for "false"? Same for auth!!!!
+        if($service == false || $service->isEmpty()) {
```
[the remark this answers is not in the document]
Meldiron: I renamed the parameter value to `$serviceKey` so I can use `$service` for the actual service object from database. Did that mess-up with what Appwrite is used to?

```php
    ->action(function ($projectId, $serviceKey, $status, $response, $dbForConsole) {

        $service = $dbForConsole->findOne("projectsServices", [
            new Query('projectId', Query::TYPE_EQUAL, [$project->getId()]),
            new Query('key', Query::TYPE_EQUAL, [$serviceKey]),
        ]);
    
    ```

**Comment by Meldiron** [comment]
> This was somehow okay, tests didn't see it as a problem
> **Do we have tests around domains?**

We do.. Maaaaybe I didn't restart Appwrite? Maaaybe it ignores the value after collections are created and I didn't remove volumes before tests? Let's not focus on that.

---

Thanks for the rest of the comments, I addressed them all in the upcoming commit.

File: app/controllers/api/projects.php (hunk 2 above)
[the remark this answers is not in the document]
Comment (Meldiron) [reply app/controllers/api/projects.php:497]: Oh, I just found an example in the code so I renamed it back to `$service` and user `$document` for the service document from database.

File: app/config/collections2.php (hunk 1 above)
Comment (Meldiron) [reply app/config/collections2.php:561]: Will do. Same for filter names.

File: app/controllers/api/projects.php [hunk 3]
```diff
@@ -468,24 +472,51 @@
...
+        } else {
+            if($service->getAttribute("status") != $status) {
+                $service->setAttribute("status", $status);
+                $dbForConsole->updateDocument("projectsServices", $service->getId(), $service);
 
+                $dbForConsole->purgeDocument('projects', $project->getId());
+            }
+        }
+
+        $project = $dbForConsole->getDocument('projects', $projectId);
```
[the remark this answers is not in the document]
Meldiron: Great catch, fixed on 2 places.

eldadfux:
> > @Meldiron `projectsServices` and `projectsAuths` should stay the same, they are just custom configurations and not children documents of the project document.
> 
> I haven't touched `projectsAuths` so I won't do that, but what about `projectsServices`? I have already migrated it and it looks fine, should I revert it? Currently it stores `key` and `status`, the rest is read from the config.
> 
> What about `projectProviders`? It is pretty similar to `projectsServices` in a way that it is not really accessible using the API and is only used in one request.

We should keep project services and project providers the same as we don’t require them as child entities at this stage.

tests/e2e/Services/Projects/ProjectsConsoleClientTest.php
```diff
@@ -7,6 +7,7 @@
 use Tests\E2E\Scopes\SideClient;
 use Tests\E2E\Services\Projects\ProjectsBase;
 use Tests\E2E\Client;
+use function array_merge;
```
eldadfux, line 10:
```suggestion
```

Not needed, we can use '\' before the function call for a tiny performance increase.

eldadfux (changes requested):

app/init.php
```diff
@@ -196,7 +199,55 @@ function($value, Document $document, Database $database) {
...
+
+Database::addFilter('subQueryPlatforms',
+    function($value) {
+        return null;
+    },
+    function($value, Document $document, Database $database) {
+        return $database
+            ->find('platforms', [
+                new Query('projectId', Query::TYPE_EQUAL, [$document->getId()])
+            ], 5000, 0, []);
```
kodumbeats, line 214: @eldadfux what's an appropriate default limit for these queries? Is `APP_LIMIT_COUNT` seems too high for me.

app/config/collections2.php
```diff
@@ -558,6 +558,471 @@
...
+                '$id' => '_key_project',
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['projectId'],
+                'lengths' => [Database::LENGTH_KEY],
+                'orders' => [Database::ORDER_ASC],
+            ],
+        ],
+    ],
+
+    'services' => [
```
kodumbeats, line 666: Do we still need this table if we don't need to subquery it?

app/config/collections2.php
```diff
@@ -558,6 +558,471 @@
...
+                '$id' => '_key_project',
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['projectId'],
+                'lengths' => [Database::LENGTH_KEY],
+                'orders' => [Database::ORDER_ASC],
+            ],
+        ],
+    ],
+
+    'providers' => [
```
kodumbeats, line 716: Do we still need this table if we don't need to subquery it?

app/controllers/api/projects.php
```diff
@@ -747,9 +755,12 @@
...
         }
 
-        $webhook = $project->find('$id', $webhookId, 'webhooks');
+        $webhook = $dbForConsole->findOne('webhooks', [
+            new Query('_uid', Query::TYPE_EQUAL, [$webhookId]),
+            new Query('projectId', Query::TYPE_EQUAL, [$project->getId()])
+        ]);
 
-        if (empty($webhook) || !$webhook instanceof Document) {
+        if ($webhook == false || $webhook->isEmpty()) {
```
kodumbeats, line 763: I think we should use strict comparisons wherever possible
```suggestion
        if ($webhook === false || $webhook->isEmpty()) {
```

app/controllers/api/projects.php
```diff
@@ -788,22 +799,26 @@
...
         $security = ($security === '1' || $security === 'true' || $security === 1 || $security === true);
 
-        $webhook = $project->find('$id', $webhookId, 'webhooks');
+        $webhook = $dbForConsole->findOne('webhooks', [
+            new Query('_uid', Query::TYPE_EQUAL, [$webhookId]),
+            new Query('projectId', Query::TYPE_EQUAL, [$project->getId()])
+        ]);
 
-        if (empty($webhook) || !$webhook instanceof Document) {
+        if ($webhook == false || $webhook->isEmpty()) {
```
kodumbeats, line 807:
```suggestion
        if ($webhook === false || $webhook->isEmpty()) {
```

app/controllers/api/projects.php
```diff
@@ -468,12 +470,13 @@
     ->label('sdk.response.model', Response::MODEL_PROJECT)
     ->param('projectId', '', new UID(), 'Project unique ID.')
     ->param('service', '', new WhiteList(array_keys(array_filter(Config::getParam('services'), function($element) {return $element['optional'];})), true), 'Service name.')
-    ->param('status', null, new Boolean(), 'Service status.')
+    ->param('status', true, new Boolean(), 'Service status.')
```
kodumbeats, line 473: What's the reasoning behind changing this default value? We should be careful when changing default behavior 🤔

app/controllers/api/projects.php
```diff
@@ -788,22 +799,26 @@
...
-                ->setAttribute('httpUser', $httpUser)
-                ->setAttribute('httpPass', $httpPass)
-            , 'webhooks');
+        $webhook
+            ->setAttribute('name', $name)
+            ->setAttribute('events', $events)
+            ->setAttribute('url', $url)
+            ->setAttribute('security', $security)
+            ->setAttribute('httpUser', $httpUser)
+            ->setAttribute('httpPass', $httpPass);
```
kodumbeats, line 817: nitpick on style:
```suggestion
            ->setAttribute('httpPass', $httpPass)
        ;
```

app/controllers/api/projects.php
```diff
@@ -831,11 +846,18 @@
             throw new Exception('Project not found', 404);
         }
 
-        if (!$project->findAndRemove('$id', $webhookId, 'webhooks')) {
+        $webhook = $dbForConsole->findOne('webhooks', [
+            new Query('_uid', Query::TYPE_EQUAL, [$webhookId]),
+            new Query('projectId', Query::TYPE_EQUAL, [$project->getId()])
+        ]);
+
+        if($webhook == false || $webhook->isEmpty()) {
```
kodumbeats, line 854:
```suggestion
        if($webhook === false || $webhook->isEmpty()) {
```

app/controllers/api/projects.php
```diff
@@ -905,7 +930,9 @@
             throw new Exception('Project not found', 404);
         }
 
-        $keys = $project->getAttribute('keys', []);
+        $keys = $dbForConsole->find('keys', [
+            new Query('projectId', Query::TYPE_EQUAL, [$project->getId()]),
+        ], 5000);
```
kodumbeats, line 935: Note to double-check this query limit

app/controllers/api/projects.php
```diff
@@ -715,7 +721,9 @@
             throw new Exception('Project not found', 404);
         }
 
-        $webhooks = $project->getAttribute('webhooks', []);
+        $webhooks = $dbForConsole->find('webhooks', [
+            new Query('projectId', Query::TYPE_EQUAL, [$project->getId()])
+        ]);
```
kodumbeats, line 726: Note to check these query limits

app/controllers/api/projects.php
```diff
@@ -928,15 +955,21 @@
...
         }
 
-        $key = $project->find('$id', $keyId, 'keys');
+        $key = $dbForConsole->findOne('keys', [
+            new Query('_uid', Query::TYPE_EQUAL, [$keyId]),
+            new Query('projectId', Query::TYPE_EQUAL, [$project->getId()])
+        ]);
 
-        if (empty($key) || !$key instanceof Document) {
+        if ($key == false || $key->isEmpty()) {
```
kodumbeats, line 972:
```suggestion
        if ($key === false || $key->isEmpty()) {
```

app/controllers/api/projects.php
```diff
@@ -969,18 +1002,21 @@
...
```
kodumbeats, line 1010:
```suggestion
        if ($key === false || $key->isEmpty()) {
```

app/controllers/api/projects.php
```diff
@@ -969,18 +1002,21 @@
...
             throw new Exception('Key not found', 404);
         }
 
-        $project->findAndReplace('$id', $key->getId(), $key
-                ->setAttribute('name', $name)
-                ->setAttribute('scopes', $scopes)
-            , 'keys');
+        $key->setAttribute('name', $name)
+            ->setAttribute('scopes', $scopes);
```
kodumbeats, line 1015: nitpick on style:
```suggestion
        $key
            ->setAttribute('name', $name)
            ->setAttribute('scopes', $scopes)
        ;
```

app/controllers/api/projects.php
```diff
@@ -1008,11 +1044,18 @@
             throw new Exception('Project not found', 404);
         }
 
-        if (!$project->findAndRemove('$id', $keyId, 'keys')) {
+        $key = $dbForConsole->findOne('keys', [
+            new Query('_uid', Query::TYPE_EQUAL, [$keyId]),
+            new Query('projectId', Query::TYPE_EQUAL, [$project->getId()])
+        ]);
+
+        if($key == false || $key->isEmpty()) {
```
kodumbeats, line 1052:
```suggestion
        if($key === false || $key->isEmpty()) {
```

app/controllers/api/projects.php
```diff
@@ -1089,7 +1135,9 @@
             throw new Exception('Project not found', 404);
         }
 
-        $platforms = $project->getAttribute('platforms', []);
+        $platforms = $dbForConsole->find('platforms', [
+            new Query('projectId', Query::TYPE_EQUAL, [$project->getId()])
+        ], 5000);
```
kodumbeats, line 1140: Note to check subquery limits

app/controllers/api/projects.php
```diff
@@ -1121,9 +1169,12 @@
...
         }
 
-        $platform = $project->find('$id', $platformId, 'platforms');
+        $platform = $dbForConsole->findOne('platforms', [
+            new Query('_uid', Query::TYPE_EQUAL, [$platformId]),
+            new Query('projectId', Query::TYPE_EQUAL, [$project->getId()])
+        ]);
 
-        if (empty($platform) || !$platform instanceof Document) {
+        if ($platform == false || $platform->isEmpty()) {
```
kodumbeats, line 1177:
```suggestion
        if ($platform === false || $platform->isEmpty()) {
```

app/controllers/api/projects.php
```diff
@@ -1158,9 +1209,12 @@
...
```
kodumbeats, line 1217:
```suggestion
        if ($platform === false || $platform->isEmpty()) {
```

app/controllers/api/projects.php
```diff
@@ -1208,11 +1256,18 @@
             throw new Exception('Project not found', 404);
         }
 
-        if (!$project->findAndRemove('$id', $platformId, 'platforms')) {
+        $platform = $dbForConsole->findOne('platforms', [
+            new Query('_uid', Query::TYPE_EQUAL, [$platformId]),
+            new Query('projectId', Query::TYPE_EQUAL, [$project->getId()])
+        ]);
+
+        if ($platform == false || $platform->isEmpty()) {
```
kodumbeats, line 1264:
```suggestion
        if ($platform === false || $platform->isEmpty()) {
```

app/controllers/api/projects.php
```diff
@@ -1298,7 +1359,9 @@
             throw new Exception('Project not found', 404);
         }
 
-        $domains = $project->getAttribute('domains', []);
+        $domains = $dbForConsole->find('domains', [
+            new Query('projectId', Query::TYPE_EQUAL, [$project->getId()])
+        ], 5000);
```
kodumbeats, line 1364: Note to check subquery limits

app/controllers/api/projects.php
```diff
@@ -1330,9 +1393,12 @@
...
         }
 
-        $domain = $project->find('$id', $domainId, 'domains');
+        $domain = $dbForConsole->findOne('domains', [
+            new Query('_uid', Query::TYPE_EQUAL, [$domainId]),
+            new Query('projectId', Query::TYPE_EQUAL, [$project->getId()])
+        ]);
 
-        if (empty($domain) || !$domain instanceof Document) {
+        if ($domain == false || $domain->isEmpty()) {
```
kodumbeats, line 1401:
```suggestion
        if ($domain === false || $domain->isEmpty()) {
```

app/controllers/api/projects.php
```diff
@@ -1363,9 +1429,12 @@
...
```
kodumbeats, line 1437:
```suggestion
        if ($domain === false || $domain->isEmpty()) {
```

app/controllers/api/projects.php
```diff
@@ -1385,11 +1454,10 @@
             throw new Exception('Failed to verify domain', 401);
         }
 
-        $project->findAndReplace('$id', $domain->getId(), $domain
-                ->setAttribute('verification', true)
-            , 'domains');
+        $domain->setAttribute('verification', true);
 
-        $dbForConsole->updateDocument('projects', $project->getId(), $project);
+        $dbForConsole->updateDocument('domains', $domain->getId(), $domain);
```
kodumbeats, line 1459: Can condense this down a bit since `$document->setAttribute()` returns `self`
```suggestion
        $dbForConsole->updateDocument('domains', $domain->getId(), $domain->setAttribute('verification', true));
```

app/controllers/api/projects.php
```diff
@@ -1385,11 +1454,10 @@
             throw new Exception('Failed to verify domain', 401);
         }
 
-        $project->findAndReplace('$id', $domain->getId(), $domain
-                ->setAttribute('verification', true)
-            , 'domains');
+        $domain->setAttribute('verification', true);
```
kodumbeats, line 1457: Can condense this down a bit since `$document->setAttribute()` returns `self`
```suggestion
```

app/controllers/api/projects.php
```diff
@@ -1424,13 +1492,18 @@
...
         }
 
-        $domain = $project->find('$id', $domainId, 'domains');
+        $domain = $dbForConsole->findOne('domains', [
+            new Query('_uid', Query::TYPE_EQUAL, [$domainId]),
+            new Query('projectId', Query::TYPE_EQUAL, [$project->getId()])
+        ]);
 
-        if (!$project->findAndRemove('$id', $domainId, 'domains')) {
+        if ($domain == false || $domain->isEmpty()) {
```
kodumbeats, line 1500:
```suggestion
        if ($domain === false || $domain->isEmpty()) {
```

kodumbeats (changes requested):

app/controllers/api/projects.php
Meldiron: Great catch. It is left-over from putting services into sub-collection which was later cancelled. Will remove.

app/config/collections2.php
Meldiron: We dont, forgot to remove that 😶 Will remove that

app/config/collections2.php
Meldiron: No we dont

app/init.php
Meldiron: `param('limit', 25, new Range(0, 100),`
Put 25 and 100 into variables, use everywhere..
Then use 100 as a limit in all subqueries.

If you need more than 100, create API endpoint for this sub-collection.

Issue #714, opened 2020-10-27
Speed Up our Docker Travis CI builds


eldadfux:
We want to use caching to speed up Appwrite's Travis CI build process, and we can use the community help here.

Currently our build process time is around ~10 minutes which is OK, but as faster it can be less time maintainers needs to wait for confirmation that there changes are running as expected.

This change should be focused on our Travis CI YAML file. Any suggestions for improving the code base should be handled on an independent issue.

SoftCreatR:
1. Step: Get rid of Travis and use GitHub Actions ;)

enggal64:
I

Sj1218305:
I want to claim this issue.
@eldadfux , is it still open?

eldadfux:
@Sj1218305 Yes it is, please reach out on Discord to share your plan of action on this issue.

eldadfux:
@Sj1218305 please work against branch `0.7.x` which is where the upcoming Appwrite version located

Glide7:
Hey @Sj1218305 :)
Are you still working on this issue? Do you perhaps need help from the team?

Sj1218305:
@Glide7 , I am working on this. Actually , I am new to Docker so I need some more time. I will surely reach out if any kind of help is required.

Glide7:
Awesome, thanks for the update!

Glide7:
@Sj1218305 also, feel free to join the #mainteners channel on our discord! https://discord.gg/yVQ7RBv4

Glide7:
Hey @Sj1218305 - how are things going? Do you need assistance from us?

hack3r-0m:
hi, I would like to take it.
@SoftCreatR suggested I will use this approach https://github.com/marketplace/actions/build-docker-images-using-cache#from-a-compose-file. I am thinking of creating a new branch from 0.7.x and make PR to that so it does not disturb the current development flow. I have a few doubts:
- if I am not wrong, there is no image of 0.7.x on the docker hub
-  in the `compose_file: docker-compose.yml` line, is the path to docker-compose.yml relative considering .github/workflows working dir?
- that action has some info regarding [testing](https://github.com/marketplace/actions/build-docker-images-using-cache#tests), is it for that particular repo, or is it relevant in this case?

eldadfux:
[...]

PR #1562, opened 2021-09-01
Feat - Parellel builds to reduce build time

kodumbeats:
## What does this PR do?

This PR makes a few changes to the structure of the Dockerfile and Travis config to allow for much faster builds:

- Uses `buildkit` as the build engine, which is much better at running builds in parallel
- Breaks the `compile` step into multiple stages, since we can use [a previous stage as a new stage](https://docs.docker.com/develop/develop-images/multistage-build/#use-a-previous-stage-as-a-new-stage)
- Renames stages to be descriptive, instead of serial.
- Calls `docker-compose pull` before bringing the stack up, which pulls images in parallel

## Test Plan

Test suite will demonstrate the build change is successful.

## TODO
- [x] investigate why `composer:2.0` fails to pull correctly on `arm64`:
https://app.travis-ci.com/github/appwrite/appwrite/jobs/535129292

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/714

Yes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,12 @@ notifications:
   email:
   - team@appwrite.io
 
+env:
+  - COMPOSE_INTERACTIVE_NO_CLI=1
+  - DOCKER_BUILDKIT=1
+  - COMPOSE_DOCKER_CLI_BUILD=1
+  - BUILDKIT_PROGRESS=plain
+
 before_install:
 - curl -fsSL https://get.docker.com | sh
 - echo '{"experimental":"enabled"}' | sudo tee /etc/docker/daemon.json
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -23,14 +23,24 @@ before_install:
     echo "${DOCKERHUB_PULL_PASSWORD}" | docker login --username "${DOCKERHUB_PULL_USERNAME}" --password-stdin
   fi
 - docker --version
+- sudo rm /usr/local/bin/docker-compose
+- curl -L https://github.com/docker/compose/releases/download/1.29.2/docker-compose-`uname -s`-`uname -m` > docker-compose
+- chmod +x docker-compose
+- sudo mv docker-compose /usr/local/bin
+- docker-compose --version
```
TorstenDittmann, line 30: I think it might be wise to add comments to these commands at this point - otherwise we will forget the reasoning of their existence 👍🏻

TorstenDittmann (changes requested):

.travis.yml
kodumbeats: 2d622db

kodumbeats:
Upstream issue: https://github.com/moby/moby/issues/41864

kodumbeats:
Is `master` the correct target branch for this?

PR #1567, opened 2021-09-02
feat(refactor-db): delete internal attributes and indexes on deleteCollection

kodumbeats:
## What does this PR do?

This PR leverages the Deletes worker to remove rows from internal `attributes` and `indexes` table when a collection is deleted.

## Test Plan

- [ ] Manually verified attributes/indexes are missing with 8c3ca68, but I don't know how to write an e2e test for this.

## Related PRs and Issues

N/A

Yes.


TorstenDittmann (approved):

app/workers/deletes.php
```diff
@@ -92,6 +95,25 @@ public function shutdown(): void
...
 
+    /**
+     * @param Document $document teams document
+     * @param string $projectId
+     */
+    protected function deleteCollection(Document $document, string $projectId): void
+    {
+        $collectionId = $document->getId();
+
+        $dbForInternal = $this->getInternalDB($projectId);
```
lohanidamodar, line 106: Don't we need to cleanup anything from external DB? or it's cleaned automatically?

kodumbeats: The `$dbForExternal->deleteCollection($collectionId)` call in the Database controller handles two responsibilities:

- [Drops the whole table from MariaDB](https://github.com/utopia-php/database/blob/fc6304cae24d344a3277812c89b6923c9f12582c/src/Database/Adapter/MariaDB.php#L178-L185)
- [Cleans up the document from the `metadata` collection](https://github.com/utopia-php/database/blob/fc6304cae24d344a3277812c89b6923c9f12582c/src/Database/Database.php#L350-L355)

There shouldn't be anything more to clean up from `$dbForExternal` after that, so we can use the Deletes worker to remove the associated documents from `$dbForInternal` in tables: `collections`, `attributes`, and `indexes`.

Thank you for the review @lohanidamodar - while answering your question, I realized that we didn't call `$dbForInternal->deleteDocument('collections', $collectionId)`. Added in bfdf6de

app/workers/deletes.php
```diff
@@ -92,6 +95,27 @@ public function shutdown(): void
...
+
+        $this->deleteByGroup('attributes', [
+            new Query('collectionId', Query::TYPE_EQUAL, [$collectionId])
+        ], $dbForInternal);
+
+        $this->deleteByGroup('indexes', [
+            new Query('collectionId', Query::TYPE_EQUAL, [$collectionId])
+        ], $dbForInternal);
+
+        $dbForInternal->deleteDocument('collections', $collectionId);
```
eldadfux, line 116: I was pretty sure this delete was the only one that should have actually be called in the HTTP call, and the drop table call was the one to be called here. Did I miss anything?

kodumbeats: Good catch - bfdf6de was the incorrect solution and was technically duplicated code (`deleteDocument` is indeed called in the deleteCollection route). Updated in cb7c810

eldadfux (approved):

PR #1569, opened 2021-09-02
feat(refactor-db): clean up internal indexes collection on delete attribute

kodumbeats:
## What does this PR do?

This PR updates our internal collection for `indexes` when an attribute is deleted. MariaDB, at least, cleans up indexes automatically, so this PR leverages the `Deletes` worker to clean up indexes on the `deleteAttribute` queue trigger.

## Unanswered Questions

 - [x] What is the behavior of MongoDB in this circumstance? Need to research.

## Test Plan

E2E tests added for this behavior

## Related PRs and Issues

N/A

Yes.


eldadfux:
Should we test this with MongoDB before merging?

kodumbeats:
> Should we test this with MongoDB before merging?

Did some research, and Mongo won't exhibit this behavior. Since Mongo collections are unstructured and lack our concept of `attributes`, we'll have to have to have custom logic when MongoDB is used.

We'll need to have this custom logic inside Appwrite and not the adapters, since this is all about keeping our new `$dbForInternal->getCollection('attributes')` table in sync with the underlying DB.

lohanidamodar (approved):
Looks good to me!

PineappleIOnic (approved):

app/workers/database.php
```diff
@@ -120,6 +120,28 @@ protected function deleteAttribute(Document $collection, Document $attribute, st
             $dbForInternal->updateDocument('attributes', $attribute->getId(), $attribute->setAttribute('status', 'failed'));
         }
 
+        // the underlying database removes/rebuilds indexes when attribute is removed
+        // update indexes table with changes
+        /** @var Document[] $indexes */
+        $indexes = $collection->getAttribute('indexes', []);
+
+        foreach ($indexes as $index) {
```
eldadfux, line 128: Why are we updating or deleting the index, shouldn't we always delete any relevant index when at least on attribute is missing?

eldadfux: I'm pretty sure that at least for mariadb the index never gets rebuild this way, but completely removed.🤔

kodumbeats: According to my research and testing, MariaDB doesn't drop the whole index. From the docs:
> If the column is part of any index, the column will be dropped from them, except if you add a new column with identical name at the same time. The index will be dropped if all columns from the index were dropped.
[ref](https://mariadb.com/kb/en/alter-table/#drop-column)

I tested this manually with the following schema:
mariadbdropindex

I then dropped column `one`, and the indexes were rebuilt accordingly:
droppedonecolumn

kodumbeats: Going to add logic for the edge case where `three` is deleted and we end up with duplicate indexes:
indexedgecase

kodumbeats: ☝🏻 edge case handled in c960116 and tested in fdd4383

PR #1616, opened 2021-09-10
Feat enum attributes

kodumbeats:
## What does this PR do?

This PR adds support for enum attributes, stored as strings. The column size is chosen from the largest element provided, since attributes can only be created or deleted.

## Test Plan

E2E tests have been added for the response model and validator enforcement.

## Related PRs and Issues

https://github.com/appwrite/appwrite/pull/1484

Yes.

### DIFF
--- a/app/config/collections2.php
+++ b/app/config/collections2.php
@@ -214,7 +214,7 @@ $collections = [
                 'required' => false,
                 'default' => new stdClass,
                 'array' => false,
-                'filters' => ['json', 'range'],
+                'filters' => ['json', 'range', 'enum'],
             ],
             [
                 '$id' => 'filters',
--- a/app/init.php
+++ b/app/init.php
@@ -198,6 +198,22 @@ Database::addFilter('casting',
     }
 );
 
+Database::addFilter('enum',
+    function($value, Document $attribute) {
+        if ($attribute->isSet('elements')) {
+            $attribute->removeAttribute('elements');
+        }
+        return $value;
+    },
+    function($value, Document $attribute) {
+        $formatOptions = json_decode($attribute->getAttribute('formatOptions', []), true);
+        if (isset($formatOptions['elements'])) {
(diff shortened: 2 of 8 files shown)


app/config/collections2.php
```diff
@@ -214,7 +214,7 @@
                 'required' => false,
                 'default' => new stdClass,
                 'array' => false,
-                'filters' => ['json', 'range'],
+                'filters' => ['json', 'range', 'enum'],
```
eldadfux, line 217: Shouldn't this value hold only `json`, why do we need `range` and `enum`?

app/init.php
```diff
@@ -196,6 +198,22 @@ function($value) {
     }
 );
 
+Database::addFilter('enum',
```
eldadfux, line 201: Not sure why we need a filter? We can just store the element options inside formatOptions, and leave validation to our standard validation flow.

eldadfux (changes requested):

app/init.php
kodumbeats: We need a way to extract values from `formatOptions` and add them into the response model. This `enum` filter ensures:
- we don't clash with the Structure validator when writing the attribute document to the internal table
- Extracts the necessary values from `formatOptions` for the proper response. In this case, we want to present the `enum` elements as a top-level `elements` key, and not simply returning them inside `formatOptions`

app/config/collections2.php
kodumbeats: We needed some way to know if/when to extract data from `formatOptions` to modify documents to fit their response models. Additionally, this filter gives us a way to modify the document when saving to the DB, so we can adjust for the Structure validator.

`enum` filter is explained below - the `range` filter prepares `min` and `max` for ints and floats:
https://github.com/appwrite/appwrite/blob/53b8a1c245800b9df14c4dfa70327f93a692bcf3/app/init.php#L199-L219

kodumbeats:
Moving to draft until response model PR settles.

PR #1626, opened 2021-09-15
Feat-support-oauth-without-email

lohanidamodar:
## What does this PR do?

- Add support for OAuth2 providers that don't provide access to user's email
- Add Reddit OAuth2 provider, that doesn't return email

## Test Plan

N/A

## Related PRs and Issues

YES

### DIFF
--- a/app/config/providers.php
+++ b/app/config/providers.php
@@ -151,6 +151,16 @@ return [ // Ordered by ABC.
         'beta' => false,
         'mock' => false
     ],
+    'reddit' => [
+        'name' => 'Reddit',
+        'developers' => 'https://www.reddit.com/dev/api/',
+        'icon' => 'icon-reddit',
+        'enabled' => true,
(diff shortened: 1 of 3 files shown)


TorstenDittmann (commented):
What happens to a user that deletes his session?

His account will be forever lost, since the e-mail is the only identifier which indicates a connection to an OAuth2 login. The same goes for trying to log in on a different device - it will just create a new account for you.

This is expected behavior for an anonymous account, but for OAuth it is not.

I have addressed this issue on the Phone Auth RFC => https://github.com/appwrite/rfc/blob/rfc-phone-auth/020-phone-auth.md#design-proposal-step-2

stnguyen90:
@lohanidamodar is this PR still worth keeping?

lohanidamodar:
@stnguyen90 this should not be relevant for now.

PR #1637, opened 2021-09-17
Add live reloader

Emiliaaah:
## What does this PR do?

Adds hot reloading to the Appwrite container.
This can be extended to reload any other container with the label `appwrite-dev-reload` label

## Test Plan

I edited a file while appwrite was running and it restarted by itself.

## Related PRs and Issues

#516

Yes

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -60,6 +60,8 @@ services:
       - traefik.http.routers.appwrite_api_https.rule=PathPrefix(`/`)
       - traefik.http.routers.appwrite_api_https.service=appwrite_api
       - traefik.http.routers.appwrite_api_https.tls=true
+      # live reloader
+      - appwrite-dev-reload
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
       - appwrite-uploads:/storage/uploads:rw
@@ -582,6 +584,19 @@ services:
   #     - './debug:/tmp'
   #   ports:
   #     - '3001:80'
+
+  live-reloader:
+    image: apogiatzis/livereloading
+    container_name: livereloader
+    privileged: true
+    environment:
+      - RELOAD_LABEL=appwrite-dev-reload
+      - RELOAD_DIR=/usr/src/code
+    #  - OBSERVER_TYPE=1 # Uncomment this if you're mounting a windows code directory and it's having issues reloading
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+      - ./app:/usr/src/code/app
+      - ./src:/usr/src/code/src
   
   # Dev Tools End   ------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -582,6 +584,19 @@ services:
   #     - './debug:/tmp'
   #   ports:
   #     - '3001:80'
+
+  live-reloader:
+    image: apogiatzis/livereloading
+    container_name: livereloader
+    privileged: true
+    environment:
```
kodumbeats, line 592: I'd like to use some additional vars from this project:
`RELOAD_DELAY` = 1 sec (Default 1.5, but it's nice to see the number)
`RESTART_TIMEOUT` = 1 sec (Not all our containers respond to `SIGINT` and only die when `SIGKILL` is sent after the timeout of 10 sec)

docker-compose.yml
```diff
@@ -60,6 +60,8 @@ services:
       - traefik.http.routers.appwrite_api_https.rule=PathPrefix(`/`)
       - traefik.http.routers.appwrite_api_https.service=appwrite_api
       - traefik.http.routers.appwrite_api_https.tls=true
+      # live reloader
+      - appwrite-dev-reload
```
kodumbeats, line 64: Any reason why we shouldn't have this on all Appwrite containers (that mount `app/` and `src/`)?

kodumbeats (changes requested):
⭐

docker-compose.yml
Emiliaaah: Ehm I kind of forgot them since I haven't personally had the need to reload them, but I'll add it to the container that have those mounted.

docker-compose.yml
Emiliaaah: Sure, I'll add the number so you can see it :)
Ah alright I probably would've noticed this if I had added it to all the containers, but will do!

kodumbeats (commented):
Tested locally after live reloading was added to the workers, and it works well when the change doesn't throw errors. However, if you save a typo or something, all the containers crash, and the live reloader doesn't restart exited containers, even if you fix it and resave. It requires a full restart  

Can we remedy this? 🤔

Emiliaaah:
I made a [PR](https://github.com/apogiatzis/docker-compose-livereloader/pull/11) to their repo adding an option to also restart exited containers.

kodumbeats:
I'm happy to consider this work `hacktoberfest-accepted`, but I don't think we should merge until the upstream changes have been accepted. It adds quite nice functionality so long as you don't accidentally save with a typo. If you do, the container won't automatically recover from the crash, and you need to restart the containers manually anyway - the problem this tool should address. 

Definitely appreciate your offer to build and upload an image ❤️ , but we avoid using custom-built Docker images in the stack to avoid the maintenance cost. 🙇🏻

Emiliaaah:
That would be great, I'll keep an eye on the upstream changes and post a comment here once it has been merged.

christyjacob4:
@Emiliaaah thanks a lot for the PR. I'm not sure if this is updated to work with the latest version of Appwrite. I'll close this PR for now. Please feel free to re-open the PR when it's ready.

Issue #1390, opened 2021-07-13
Improved SMTP error logging in development


kodumbeats:
## 🚀 Feature

Improve error logging in the Mails worker by conditionally setting `$mail->SMTPDebug = true;`

## Have you spent some time to check if this issue has been raised before?

Yes.

### Have you read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/master/CODE_OF_CONDUCT.md)?

Yes.

## Pitch

Appwrite uses `phpmailer` in the Mails worker to send SMTP messages. If the SMTP server rejects the email for whatever reason, it responds with an unhelpful error message:

`SMTP Error: Data not accepted.`

I propose we conditionally set `$email->SMTPDebug = true;` when `_APP_ENV=development` to give some control when troubleshooting SMTP.

eldadfux:
I think we can have this setting enabled by default using the doctor command

loftwah:
Is this in the documentation? I'm poking around with Appwrite and this is the kind of thing I'd hopefully find there. Is this a setting that could be controlled from the UI?

manoelfilho:
Hello. I am testing the appwrite as an alternative here but I am having problems to use the email processor. Every moment I see **Error sending mail: SMTP Error: data not accepted**. Can somebody help-me with it?

My Configs of Mailgun are:

_APP_SYSTEM_EMAIL_NAME=NAME
_APP_SYSTEM_EMAIL_ADDRESS= email Confirmed in Mailgun

_APP_SMTP_HOST=smtp.mailgun.org
_APP_SMTP_PORT=587
_APP_SMTP_SECURE=tls
_APP_SMTP_USERNAME=postmaster@sandbox***
_APP_SMTP_PASSWORD=9****0

What is wrong here?

stnguyen90:
We will no longer enable SMTP debug in the doctor command because it's a bit messy. See https://github.com/appwrite/appwrite/pull/5075#discussion_r1105358745

stnguyen90:
We also don't want to enable SMTP debug in the worker logs because of how messy it is. If we move forward with this, we would add a separate `docker compose exec appwrite debug smtp` command that would send a test email with the SMTP debug enabled.

mustansirgodhrawala:
@stnguyen90 Could you please assign this issue to me, since I just understood how we implement the doctor smtp implementation?

stnguyen90:
[...]

PR #1644, opened 2021-09-21
Feat: enable SMTP debugging when not in production

kodumbeats:
## What does this PR do?

PHPMailer doesn't have very useful error messages by default, and in fact provides a [URL to their wiki](https://github.com/PHPMailer/PHPMailer/wiki/Troubleshooting). This PR increases the debugging level of PHPMailer when not in production.

## Test Plan

Check logs on failed SMTP delivery.

## Unanswered questions

- [ ] This will `echo` the logs by default, but I can overwrite the logging function. Will `echo`ing logs work, or should we use the Utopia CLI lib?

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/1390

Yes.

### DIFF
--- a/app/workers/mails.php
+++ b/app/workers/mails.php
@@ -76,6 +76,11 @@ class MailsV1 extends Worker
         /** @var \PHPMailer\PHPMailer\PHPMailer $mail */
         $mail = $register->get('smtp');
 
+        // https://github.com/PHPMailer/PHPMailer/wiki/Troubleshooting#enabling-debug-output
+        if (App::getEnv('_APP_ENV') !== App::MODE_TYPE_PRODUCTION) {
+            $mail->SMTPDebug = \PHPMailer\PHPMailer\SMTP::DEBUG_SERVER;
+        }
+
         // Set project mail
         /*$register->get('smtp')
             ->setFrom(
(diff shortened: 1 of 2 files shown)


TorstenDittmann (approved):

app/workers/mails.php
```diff
@@ -76,6 +76,11 @@ public function run(): void
         /** @var \PHPMailer\PHPMailer\PHPMailer $mail */
         $mail = $register->get('smtp');
 
+        // https://github.com/PHPMailer/PHPMailer/wiki/Troubleshooting#enabling-debug-output
+        if (App::getEnv('_APP_ENV') !== App::MODE_TYPE_PRODUCTION) {
```
eldadfux, line 80: This method makes more sense to me: https://github.com/utopia-php/framework/blob/master/src/App.php#L411
I would just double check we initialize the env setting in the worker scope the same way we do for the REST API.

eldadfux (changes requested):

stnguyen90:
@kodumbeats this is quite stale now and we've decided on a [different approach](https://github.com/appwrite/appwrite/issues/1390#issuecomment-1430273185) for debugging. As such, can this be closed?

PR #1646, opened 2021-09-21
Feat: Deprecate user for security reasons

Meldiron:
## Problem

When you delete a user, his record is deleted. This means a new user can register and use his ID. This becomes a way bigger problem if we introduce custom IDs or custom ID generation strategies.

## What does this PR do?

When you attempt to remove an account, instead of deleting a document, we delete all crucial attributes and set its status to `deleted`. This kind of account is ignored **everywhere** in the application, except API for creating a new user. If an ID is already used, the system does not allow the creation of a new account.

## Test Plan

- Tested manually, worked.
- Current tests pass
- We might need new tests for this behaviour

✅


app/config/collections2.php
```diff
@@ -695,6 +695,17 @@
                 'array' => true,
                 'filters' => ['json'],
             ],
+            [
+                '$id' => 'deleted',
```
kodumbeats, line 699: We need to create a compound index for `deleted` and `email` since we'll be searching against both of those columns. That way, we get much better query performance. 

Normally, the order of orders matters in compound indexes (example: `['ASC', 'ASC']` is a different index from `['ASC','DESC']`), but since one column is a boolean, I don't expect a large impact to performance (small impact at worst, none at best).

kodumbeats (changes requested):
> We might need new tests for this behaviour

We should add a test using customIDs to be 💯 sure the customID can't be reused, as that's the permission gap we need to close.

app/config/collections2.php
Meldiron: Added index 😇

Meldiron:
> > We might need new tests for this behaviour
> 
> We should add a test using customIDs to be 💯 sure the customID can't be reused, as that's the permission gap we need to close.

Tests added 😊

tests/e2e/Services/Users/UsersCustomServerTest.php
```diff
@@ -11,4 +12,51 @@ class UsersCustomServerTest extends Scope
...
+
+        $this->assertEquals(201, $response['headers']['status-code']);
+
+        // Delete user with custom ID 'meldiron'
+        $response = $this->client->call(Client::METHOD_DELETE, '/users/meldiron', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()), [
+
+        ]);
```
kodumbeats, line 41:
```suggestion
        ], $this->getHeaders()));
```

tests/e2e/Services/Users/UsersCustomServerTest.php
```diff
@@ -11,4 +12,51 @@ class UsersCustomServerTest extends Scope
...
+
+        $this->assertEquals(204, $response['headers']['status-code']);
+
+        // Try to create user with custom ID 'meldiron' again, but now it should fail
+        $response1 = $this->client->call(Client::METHOD_POST, '/users', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()), [
+            'userId' => 'meldiron',
+            'email' => 'matej@appwrite.io',
```
kodumbeats, line 51: I want to be sure that this call fails because of a uniqueID conflict and nothing else. Can we use a different name/email for these assertions?

kodumbeats (changes requested):

docker-compose.yml
```diff
@@ -63,7 +63,7 @@ services:
       - ./psalm.xml:/usr/src/code/psalm.xml
       - ./tests:/usr/src/code/tests
       - ./app:/usr/src/code/app
-      # - ./vendor/utopia-php/database:/usr/src/code/vendor/utopia-php/database
+      - ./vendor:/usr/src/code/vendor
```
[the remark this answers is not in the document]
Meldiron: 👍 Fixed

tests/e2e/Services/Users/UsersCustomServerTest.php
Meldiron: Great catch! Updated

TorstenDittmann (approved):

kodumbeats (approved):

PR #1791, opened 2021-09-30
Fix typo in feature request issue template

SoftCreatR:
n/t

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature.yaml
+++ b/.github/ISSUE_TEMPLATE/feature.yaml
@@ -1,4 +1,4 @@
-name: 🚀 Feature"
+name: "🚀 Feature"
 description: "Submit a proposal for a new feature"
 title: "🚀 Feature: "
 labels: [feature]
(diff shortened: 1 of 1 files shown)


thinkverse (commented):
Could this PR also include a fix for the extra `"` in the issue template name? Which is caused by the missing `"` after `name: ` on line 1. This can be fixed by either adding a starting double-quote or removing the leftover one.

Screen Shot 2021-10-01 at 1 18 54 AM

.github/ISSUE_TEMPLATE/feature.yaml
```diff
@@ -1,12 +1,12 @@
...
+name: "🚀 Feature"
 description: "Submit a proposal for a new feature"
 title: "🚀 Feature: "
 labels: [feature]
 body:
   - type: markdown
     attributes:
       value: |
-        Thanks for taking the time to fill out our bug report form 🙏
+        Thanks for taking the time to fill out feature request report form 🙏
```
thinkverse, line 9:
```suggestion
        Thanks for taking the time to fill out the feature request form 🙏
```

This makes more sense than "fill out feature request report form", IMHO. Up to you though. 👍

SoftCreatR: That's what happens, when in hurry 😄

PR #1793, opened 2021-10-01
(Security) Added Dependabot configuration.

Mou-Ikkai:
## What does this PR do?

This adds the automated dependency updater that got recently directly implemented into github. This adds the configuration for it, plus some dependency updates I did during testing.

Yep, this shouldn't be too breaking, this should actually help on security or flaws in dependent software.

## Note:
You might have to go to Insights, go to Dependency Graph, click on the Dependabot tab and enable dependabot. This is because the enabling of this feature obviously won't cross over from a PR.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,7 @@ updates:
       interval: "daily"
     # Try to remove the PR limit
     open-pull-requests-limit: 999
-updates:
+
   - package-ecosystem: "composer"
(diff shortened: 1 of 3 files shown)


lohanidamodar:
@Mou-Ikkai Thank you for your hard work. However we have some specific version dependencies between utopia frameworks and don't want them automatically upgrading. We are not accepting this PR at the moment.

Issue #1837, opened 2021-10-02
📚 Documentation: Missing Instructions to Stop the Appwrite Docker Instance


shadowshot-x:
### 💭 Description

It is always a good practice to add the instructions or commands to stop the instance of running software. I could not find the instructions to do a clean kill for appwrite on the documentation page and also by googling.

As Appwrite only uses Docker, so, I killed all the running containers to stop the instance. As docker containers may keep running in the background, appwrite will be up even when a developer is not using it. Adding stopping instructions will help beginners who don't have knowledge of docker to gain more control of the resources of their system.

I used this to stop the running containers on Ubuntu:-
`$ docker kill $(docker ps -q)`

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I read the Code of Conduct

shadowshot-x:
Also, I will be happy to contribute to this issue if feasible

eldadfux:
Feel free to add the new section on our installation page: https://appwrite.io/docs/installation. You can add it just before the `Learn More` section. Please use the proper syntax for the code examples as used in the page, and proper English.

The relevant file to update is:
https://github.com/appwrite/docs/blob/main/app/views/docs/installation.phtml

Good luck, we're available on Discord for any questions: https://appwrite.io/discord

shadowshot-x:
Thank You @eldadfux . I will get started on it.

shadowshot-x:
I noticed all the containers started by this process begin with appwrite. So it would be sensible to stop the containers using 
`docker container stop $(docker container ls -q --filter name=appwrite*)`
This stops all the containers cleanly. I will be adding this in the documentation.

eldadfux:
This section as recently added to the docs: https://appwrite.io/docs/installation

gustawdaniel:
I was trying

```
docker container stop `docker container ls -q --filter name=appwrite`
```

but does not work. Finally I uninstalled by

```
docker rmi `docker images appwrite/appwrite -q` -f
```

rayanfer32:
```
sudo docker stop $(sudo docker ps -aqf "name=^appwrite" | tr '\n' ' ')
```

PR #1844, opened 2021-10-02
[add] Command to stop the running docker containers

malikakarsh:
## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)


README.md
```diff
@@ -82,6 +82,12 @@ docker run -it --rm ,
     appwrite/appwrite:0.10.4
 ```
 
+### Stop The Running Docker Containers (Windows/Unix)
+
+```bash
+docker-compose -f docker-compose.yml down
+```
```
shadowshot-x, line 89: I don't think everyone using appwrite will have installed it on their system. So, to make this work they will have to download the yml file and then run docker-compose. Is there some workaround for this?

malikakarsh: Hey but I simply run the docker command and it automatically started the docker-compose. So I thought why not add a script to put it down. Cause a lot many containers were running simultaneously, and it would be a lot of pain to stop them individually.

shadowshot-x: Right I understand that. Docker-compose is run during the installation. But the user might not have that yml file in that directory. Maybe we can point some additional comments about the location of the yml file to give a better understanding. I have proposed a similar solution here and stopping like this works for me. https://github.com/appwrite/appwrite/issues/1837

malikakarsh: Okay thank you, I will look into this.

malikakarsh: I checked your command however in my terminal it says `zsh: no matches found: name=appwrite*` . So I modified my previous command and now it works fine. Please do check it.

shadowshot-x: I think this is because it could not find any running containers. Did you confirm the appwrite containers running when you ran this command?
However, I your command also works on bash.
`docker container stop $(docker ps | grep "appwrite*" | cut -d " " -f 1)`

malikakarsh: Yes at the moment when I ran your script, the containers were running.

PR #2159, opened 2021-10-08
feat(db-refactor): remove old database

TorstenDittmann:
## What does this PR do?

- removes old Database
- moves Database Validators to Utopia
- removes migration tests (not needed anymore)
- remove response filter tests (not needed anymore)

## Test Plan

🤞🏻


eldadfux:
@TorstenDittmann is this PR still relevant? I think we need the old DB lib for both migration and some backward compatibility.

PR #2217, opened 2021-10-14
Session duration implementation

Meldiron:
## What does this PR do?

This PR adds a new optional parameter `duration` to all (4) of our API endpoints for creating a session.
The default value is 1 year and the allowed range is from 1 day to 1 year. All of these 3 variables can be configured using `_APP_AUTH_DEFAULT_DURATION, _APP_AUTH_MAX_DURATION, _APP_AUTH_MIN_DURATION` variables in `.env`.

Possible problem: **Session renewing not implemented**

- I log in at 15:20 with a 5-minute session
- I do some action (list documents) at 15:23
- When do I get logged out? 15:25 or 15:28? Currently, at 15:25

Possible problem: **OAuth way of storing duration**

We store duration in the `state` of the OAuth2 process. We validate the duration every time it's provided by the client, but there might be some attacks that change the value without the client knowing about it. There might be some policies to protect that and we might need to store the duration inside the user object.

Possible problem: **Team invite email session expiry**

Currently, if you invite a user that is not part of Appwrite into your team, it sends him a link that logs him in. This session uses default duration and is not configurable.

Possible problem: **MagicURL expiry**

We currently use one `expiry` parameter for 2 expiration dates - when the MagicURL link expires and when the session expires. Let's say you set the expiry to 15 minutes. You wait for 5 minutes for the email to arrive and you click it. You only had 10 minutes left, otherwise, the link would be invalid. After clicking it, you are logged in, but the session will explore in 10 minutes - because the total duration is 15 minutes and 5 minutes was already taken by waiting for the email.

## Test Plan

Wrote new tests that passed 👍

- OAuth tests
- MagicURL tests
- AnonymousLogin tests
- CreateSession tests (email+password)

✅


Meldiron:
No idea why tests are failing...

From GitHub, I can see some Swagger errors..
For me locally, it also does not make sense. It was:

```
Time: 10:06.957, Memory: 26.00 MB

There was 1 failure:

1) Tests\E2E\Services\Functions\FunctionsCustomServerTest::testTimeout
Failed asserting that 1.9997320175171 is greater than 2.

/usr/src/code/tests/e2e/Services/Functions/FunctionsCustomServerTest.php:523

FAILURES!
Tests: 405, Assertions: 48158, Failures: 1, Skipped: 1.
```

Then, on second run, all passed:
```
Time: 10:13.161, Memory: 28.00 MB

OK (405 tests, 48186 assertions)
```

PR #2238, opened 2021-10-16
added 'u' in regular expression

Biki-das:
adding 'u' in the regex looks good here, open to feedback anyways!


Biki-das:
@TorstenDittmann could you review this PR?

Biki-das:
@eldadfux could you review?

PR #2332, opened 2021-10-22
fix function model permission

lohanidamodar:
## What does this PR do?

Fixes the response model for Function permission, previously it was using the permission but as function only has execute permission and not read-write so it was not working as expected

## Test Plan

N/A

## Related PRs and Issues

YES

### DIFF
--- a/src/Appwrite/Utopia/Response.php
+++ b/src/Appwrite/Utopia/Response.php
@@ -22,6 +22,7 @@ use Appwrite\Utopia\Response\Model\ErrorDev;
 use Appwrite\Utopia\Response\Model\Execution;
 use Appwrite\Utopia\Response\Model\File;
 use Appwrite\Utopia\Response\Model\Func;
+use Appwrite\Utopia\Response\Model\FuncPermissions;
 use Appwrite\Utopia\Response\Model\JWT;
 use Appwrite\Utopia\Response\Model\Key;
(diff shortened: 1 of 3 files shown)


kodumbeats (approved):

kodumbeats (changes requested):
After some more thinking, I don't think we should abbreviate `FuncPermissions` - can we use the entire words? (Could be wrong, but I don't think we use abbreviations a lot in the codebase)

lohanidamodar:
@kodumbeats as the original func model is abbreviated to func, this feels right doesn't it? Do we even update the original function model to Function instead?

TorstenDittmann (approved):

eldadfux:
After we flattened the permissions model for documents and files, shouldn't we do the same for function?

Maybe $execute? WDYT?

lohanidamodar:
> After we flattened the permissions model for documents and files, shouldn't we do the same for function?
> 
> Maybe $execute? WDYT?

Agreed, but this is for the current master, having issue on Dart SDK response models because of this.

eldadfux:
> > After we flattened the permissions model for documents and files, shouldn't we do the same for function?
> > Maybe $execute? WDYT?
> 
> Agreed, but this is for the current master, having issue on Dart SDK response models because of this.

OK, but let's make sure this is tackled in newer branches.

PR #2379, opened 2021-11-01
Fix-swagger-spec-for-permissions

lohanidamodar:
## What does this PR do?

Fixes swagger spec for new Permission validator (before considered as String, now Array as it should be)

## Test Plan

## Related PRs and Issues

YES

### DIFF
--- a/src/Appwrite/Specification/Format/Swagger2.php
+++ b/src/Appwrite/Specification/Format/Swagger2.php
@@ -329,7 +329,7 @@ class Swagger2 extends Format
                         $node['items'] = [
                             'type' => 'string',
                         ];
-                        $node['x-example'] = '["role:all"]'
+                        $node['x-example'] = '["role:all"]';
                         break;
                     case 'Appwrite\Auth\Validator\Password':
                         $node['type'] = $validator->getType();
--- a/src/Appwrite/Specification/Format/Swagger2.php
+++ b/src/Appwrite/Specification/Format/Swagger2.php
@@ -323,6 +323,13 @@ class Swagger2 extends Format
                             'type' => 'string',
                         ];
                         break;
+                    case 'Utopia\Database\Validator\Permissions':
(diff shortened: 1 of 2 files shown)


ekol-tech:
Ekrem Yılmaz
iPhone’umdan gönderildi

kodumbeats (approved):
Looks good, just one question:

When does a validator get `x-example` set in the spec format?

lohanidamodar:
@kodumbeats added `x-example`, seems skipped it before as it was not there in the ArrayList, not sure if this works.

kodumbeats:
I'm not convinced any `x-example` is working, but I think this should be solved in a separate PR. 

To peek at this Permissions spec:



Even looking at something that we set `x-example` - for the JSON validator, but it is null as well

kodumbeats (approved):

TorstenDittmann (approved):

src/Appwrite/Specification/Format/Swagger2.php
```diff
@@ -323,6 +323,14 @@ public function parse(): array
                             'type' => 'string',
                         ];
                         break;
+                    case 'Utopia\Database\Validator\Permissions':
+                        $node['type'] = $validator->getType();
+                        $node['collectionFormat'] = 'multi';
```
eldadfux, line 328: what is multi used for?

eldadfux (commented):
We miss a similar change for OpenAPI3

src/Appwrite/Specification/Format/Swagger2.php
lohanidamodar: No idea, I just took this from ArrayList https://github.com/appwrite/appwrite/blob/404a3c5754c21a3a56f793ac26d7d7545337f398/src/Appwrite/Specification/Format/Swagger2.php#L321
As permissions is essentially a ArrayList.

Issue #2372, opened 2021-10-30
🚀 Feature: Prevent email address harvesting


ralienpp:
### 🔖 Feature description

Appwrite should be more _discrete_ during the password recovery workflow. Currently, it says "password recovery failed" (or that it was successful) making it possible to harvest email addresses.

![image](https://user-images.githubusercontent.com/295338/139534577-fedcc132-b394-472c-82ee-8db2ecbfa831.png)

Though one could argue it is not a major security issue, it is often the first step in a more elaborate attack.

A proposed alternative is to say "We have tried to send the instructions to the address you provided, please check your email". This text neither confirms nor denies the existence of the given address.

### 🎤 Pitch

This is a good security practice, and it will propagate to all the systems built on top of Appwrite.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

eldadfux:
Hey Alex, thanks for using Appwrite and sending this feedback. I would argue and say this is not a small security issue.
We must have a true/false indication for success or failure on this API to allow proper usage by devs and to allow UIs to have some representation to the state of the operations.

We did limit this endpoint for 10 attempts per hour for each email, which was originally designed to block abuse of a specific user. Obviously this is not bullet proof, and will not stop an harvesting attack. To add extra protection can add an an IP limit alongside the email one. This will make an harvesting attack VERY expensive and not worthwhile.

In this line we can add support for array of abuse keys to check multiple factors for abuse: https://github.com/appwrite/appwrite/blob/master/app/controllers/api/account.php#L1736

We should probably also make sure the documentation site knows to explain multiple abuse rules vs. a single key rule.

eldadfux:
This was fixed and released as part of Appwrite 0.12!
@ralienpp thanks for raising the issue!

PR #2401, opened 2021-11-08
Update abuse-key for password recovery to add IP

PineappleIOnic:
## What does this PR do?
This PR adds IP Checking to the abuse-key for password recovery to help combat the ability to email harvest on Appwrite

## Related PRs and Issues

#2372 

Yes,  I have


eldadfux:
We need to have this as a separate check instead of adding the IP to the unique string, the string only defines what values would be considered as unique values for each request and will increase the abuse count. We need to apply a mechanism to support multiple abuse checks, each by a different unique key.

app/controllers/shared/api.php
```diff
@@ -37,41 +37,60 @@
...
+    } else {
+        $timeLimit = new TimeLimit($abuseKeyLabel, $route->getLabel('abuse-limit', 0), $route->getLabel('abuse-time', 3600), $db);
+        $timeLimit->setNamespace('app_'.$project->getId());
+        $timeLimit
+            ->setParam('{userId}', $user->getId())
+            ->setParam('{userAgent}', $request->getUserAgent(''))
+            ->setParam('{ip}', $request->getIP())
+            ->setParam('{url}', $request->getHostname().$route->getPath())
+        ;
+        $timeLimitArray[] = $timeLimit;
```
TorstenDittmann, line 63: To be more dry, you can just check if `$abuseKeyLabel` is not an array and push it to `$timeLimitArray` - that way you will only have one occurence of `new TimeLimit()`.

Something like:
```php
if (!is_array($abuseKeyLabel)) {
    $timeLimitArray[] = $abuseKeyLabel;
}
```

That way you can iterate over the array everytime 👍🏻

app/controllers/shared/api.php
```diff
@@ -37,41 +37,60 @@
...
 
-    $isPrivilegedUser = Auth::isPrivilegedUser(Authorization::$roles);
-    $isAppUser = Auth::isAppUser(Authorization::$roles);
+        if ($timeLimit->limit()) {
+            $response
+                ->addHeader('X-RateLimit-Limit', $timeLimit->limit())
+                ->addHeader('X-RateLimit-Remaining', $timeLimit->remaining())
+                ->addHeader('X-RateLimit-Reset', $timeLimit->time() + $route->getLabel('abuse-time', 3600))
+            ;
+        }
```
TorstenDittmann, line 83: Looks like these headers will be set for the last item in `$timeLimit` 🤔

app/controllers/shared/api.php
```diff
@@ -37,41 +37,60 @@
...
 
-    if (($abuse->check() // Route is rate-limited
-        && App::getEnv('_APP_OPTIONS_ABUSE', 'enabled') !== 'disabled') // Abuse is not diabled
-        && (!$isAppUser && !$isPrivilegedUser)) // User is not an admin or API key
-        {
-        throw new Exception('Too many requests', 429);
+        $isPrivilegedUser = Auth::isPrivilegedUser(Authorization::$roles);
+        $isAppUser = Auth::isAppUser(Authorization::$roles);
```
TorstenDittmann, line 86: This part is not necessary to be part of the loop.

TorstenDittmann (commented):
Looking at this implementation we might need to be more smart.

Since the headers sent need to be the closest to the abuse. Meaning if the IP has only 2 requests left, but the email has 3 and comes last in the iteration - the headers are wrong.

We need to keep track what headers are the closest to be trigger the limit and set them after the loop.

app/controllers/shared/api.php
PineappleIOnic: Isn't this used to check if the user can bypass the restrictions?

PineappleIOnic: Isn't this used to check if the user can bypass the restrictions?

app/controllers/shared/api.php
```diff
@@ -37,41 +37,54 @@
...
+            ->setParam('{userId}', $user->getId())
+            ->setParam('{userAgent}', $request->getUserAgent(''))
+            ->setParam('{ip}', $request->getIP())
+            ->setParam('{url}', $request->getHostname().$route->getPath());
+        $timeLimitArray[] = $timeLimit;
     }
 
+    //TODO make sure we get array here
+
+    $closestLimit = 999;
```
TorstenDittmann, line 60: Not really like the hardcoded `999`, lets make it `null` and add a new condition to the next lines:

```php
($timeLimit->remaining() < $closestLimit || is_null($closestLimit))
```

app/controllers/shared/api.php
```diff
@@ -37,41 +37,54 @@
...
+    $abuseKeyLabel = $route->getLabel('abuse-key', 'url:{url},ip:{ip}');
+    $timeLimitArray = [];
 
-    foreach ($request->getParams() as $key => $value) { // Set request params as potential abuse keys
-        if(!empty($value)) {
-            $timeLimit->setParam('{param-'.$key.'}', (\is_array($value)) ? \json_encode($value) : $value);
-        }
+    if (!is_array($abuseKeyLabel)) {
+        $abuseKeyLabel = [$abuseKeyLabel];
     }
```
eldadfux, line 45:
```suggestion
    $abuseKeyLabel =  (!is_array($abuseKeyLabel)) ? [$abuseKeyLabel] : $abuseKeyLabel;
```

TorstenDittmann: This would be an assignment even if it's already an array 🤔

eldadfux (approved):

Issue #2406, opened 2021-11-11
🐛 Bug Report: accounts.create does not get trigger function on account created with OAuth provider


matfire:
### 👟 Reproduction steps

- Create a function that should get triggered on accounts.create
- Deploy tag (my use case was creating a document in a collection)
- Create an account with email and password from a web frontend (React in my specific case)
- Function should get triggered
- Create another account using an Oauth provider (Google in my case)
- Function should not get triggered

### 👍 Expected behavior

Function should get triggered in both cases, when an account is created using email + password and when using a third party provider (Google, Github etc)

### 👎 Actual Behavior

Function gets triggered only when creating an account using email and password

### 🎲 Appwrite version

Different version (specify in environment)

### 💻 Operating system

Linux

### 🧱 Your Environment

Appwrite v:0.11.0.170

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

JaviBonilla:
I’m also having the same issue.

PineappleIOnic:
I'm on the case 👍

garvitomer:
Is this issue solved in 13.4 or it is still there? Last I checked it was there in 12.x versions

stnguyen90:
@PineappleIOnic, looking at the code, it looks like the user is created, but no new user event is triggered:

https://github.com/appwrite/appwrite/blob/648b6f2038fb59b2beb78d81cea08aa730a106c3/app/controllers/api/account.php#L466-L501

karniv00l:
I think this issue also exists when creating a user using Magic URL

ddenev:
@stnguyen90 , can this be prioritized higher?

I see this has been lagging since 0.16 milestone but currently this bug prevents me from using Google OAuth2 in my app.

Thank you!

garvitomer:
Use 'session.create' for time being, I am also using the same.

t0mm4rx:
Still an issue with Magic URLs. Checking for session creation with some sort of state is possible but not ideal.

moshOntong-IT:
What is the status for this issue?

moshOntong-IT:
We should consider introducing a new method named createOAuth2.

[...]

PR #2421, opened 2021-11-16
Fix Event not being triggered when a new account is created through OAuth

PineappleIOnic:
## What does this PR do?
This PR fixes a bug that event triggers are not being correctly fired when a new account is created with the OAuth endpoints.

## Related PRs and Issues

#2406 

Yes, I have.

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -20,6 +20,7 @@ use Appwrite\Database\Exception\Duplicate;
 use Appwrite\Database\Validator\UID;
 use Appwrite\Database\Validator\Authorization;
 use Appwrite\Detector\Detector;
+use Appwrite\Event\Event;
 use Appwrite\Template\Template;
 use Appwrite\OpenSSL\OpenSSL;
 use Appwrite\URL\URL as URLParser;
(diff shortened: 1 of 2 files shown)


TorstenDittmann:
Can you add tests? 🙏🏻

app/controllers/api/account.php
```diff
@@ -524,7 +525,40 @@
                         'registration' => \time(),
                         'reset' => false,
                         'name' => $name,
+                        'prefs' => [],
                     ], ['email' => $email]);
+
+                    $createUserEvent = clone $audits;
+
+                    $createUserEvent
+                        ->setParam('eventData', $response->output($user, Response::MODEL_USER))
```
TorstenDittmann, line 534: Since this can be quite the heavy task with all the things going on - we might want to avoid calling `$response->output` three time to spare some compute 👍🏻

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -391,6 +393,58 @@ public function testGetExecution(array $data):array
...
+
+        // Get latest execution logs
+        $executions = $this->client->call(Client::METHOD_GET, '/functions/'.$data['functionId'].'/executions', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()));
+
+        $latestExecution = $executions['body']['executions'][1];
+
+        $this->assertEquals('event', $latestExecution['trigger']);
```
TorstenDittmann, line 443: Can we even check for the specific event name here and also the payload?

TorstenDittmann (approved):

eldadfux:
@PineappleIOnic @TorstenDittmann could you explain what is the core issue? is the API lifecycle cut mid request? why can't we rely on the original trigger. Adding and managing each event per request is not the best design choice if we have other options.

TorstenDittmann:
> @PineappleIOnic @TorstenDittmann could you explain what is the core issue? is the API lifecycle cut mid request? why can't we rely on the original trigger. Adding and managing each event per request is not the best design choice if we have other options.

First of, we can't rely on the original trigger since the events have different payloads. Also triggering resets the Event class, which would require more work as well.

Since we have 2 different payloads, we have to find another way than just relying on the response payload.

That's why I suggested doing this for now and have a discussion around a possible implementation.

eldadfux:
@TorstenDittmann we do have an option to overwrite the default event data without completely replicating the functionality  like implemented here: https://github.com/appwrite/appwrite/blob/a585a9090ad995dab4097164be9a1a4194e29d44/app/controllers/shared/api.php#L190-L192 - why isn't the original event being triggered?

TorstenDittmann:
> @TorstenDittmann we do have an option to overwrite the default event data without completely replicating the functionality  like implemented here: https://github.com/appwrite/appwrite/blob/a585a9090ad995dab4097164be9a1a4194e29d44/app/controllers/shared/api.php#L190-L192 - why isn't the original event being triggered? 

That was my first suggestion. Then we would still had to clone it, since the event object is reset after the trigger.

eldadfux (commented):
I'm not sure why the lifecycle is stopped after redirect, I feel like this is a patch and I'm not 100% sure what is the root cause for the event to not trigger.

karniv00l:
Any progress on that? 👀

PineappleIOnic:
PR is heavily outdated, creating a new one using the new event system...

Issue #2412, opened 2021-11-12
🐛 Bug Report: Update Account Email from the Account API does not send the new confirmation mail 


JaviBonilla:
### 👟 Reproduction steps

With the `Account API`:

- Update Account Email

### 👍 Expected behavior

According to the Account API [documentation](https://appwrite.io/docs/client/account?sdk=flutter-default#accountUpdateEmail):

- [x] Update email address
- [x] Reset user confirmation status
- [x] Send a new confirmation mail

### 👎 Actual Behavior

- [x] Update email address
- [x] Reset user confirmation status
- [ ] Send a new confirmation mail **(This is not done)**

### 🎲 Appwrite version

Version 0.10.x

### 💻 Operating system

Linux

### 🧱 Your Environment

- Flutter Appwrite 2.0.3

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

eldadfux:
The docs here are misleading and should be fixed. The confirmation mail is decoupled from the actual update endpoint to allow each app its own behavior and authentication flows.

eldadfux:
This was fixed and will be updated as part of the Appwrite 0.12 release.

eldadfux:
This fix is now available as part of Appwrite 0.12 which is now officially released. You can learn more at: https://dev.to/appwrite/its-here-announcing-the-release-of-appwrite-012-5c8b

PR #2422, opened 2021-11-16
Fix description for the update email endpoint

PineappleIOnic:
## What does this PR do?

This PR will fix the documentation of the update email endpoint to clarify that a new confirmation email is not automatically send and to use the send confirmation email endpoint to send one.

## Related PRs and Issues

#2412 

Yes, I have


docs/references/account/update-email.md
```diff
@@ -1,2 +1,2 @@
-Update currently logged in user account email address. After changing user address, user confirmation status is being reset and a new confirmation mail is sent. For security measures, user password is required to complete this request.
+Update currently logged in user account email address. After changing user address, the user confirmation status will get reset. A new confirmation email is not sent automatically however and you will need to use the send confirmation email endpoint again. For security measures, user password is required to complete this request.
```
TorstenDittmann, line 1: Does the `however` make sense here?

PineappleIOnic: It would make a bit more sense to use `however you can` instead of `however you will` as however is normally used to explain contridiction of the previous point

PR #2427, opened 2021-11-19
fix(deletes): add realtime logs and improve code

TorstenDittmann:
## What does this PR do?

- adds realtime logs deletion
- improve codebase to be more dry

(Write your answer here.)

### DIFF
--- a/app/workers/deletes.php
+++ b/app/workers/deletes.php
@@ -11,7 +11,7 @@ use Utopia\Abuse\Adapters\TimeLimit;
 use Utopia\CLI\Console;
 use Utopia\Audit\Audit;
 
-require_once __DIR__.'/../init.php';
+require_once __DIR__ . '/../init.php';
 
 Console::title('Deletes V1 Worker');
 Console::success(APP_NAME . ' deletes worker v1 has started' . "\n");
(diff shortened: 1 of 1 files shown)


app/workers/deletes.php
```diff
@@ -319,22 +324,28 @@ protected function deleteForProjectIds(callable $callback): void
...
             foreach ($projectIds as $projectId) {
-                $callback($projectId);
+                if (!($dbForInternal = $this->getInternalDB($projectId))) {
+                    throw new Exception('Failed to get projectDB for project ' . $projectId);
+                }
+
+                if (!($dbForExternal = $this->getExternalDB($projectId))) {
+                    throw new Exception('Failed to get projectDB for project ' . $projectId);
+                }
+                $callback($projectId, $dbForInternal, $dbForExternal);
```
kodumbeats, line 348: I can understand that this change makes the code a bit more DRY, but I got totally lost when trying to follow the database connections through the logic. Maybe we should repeat ourselves a bit to make the story easier to follow?

kodumbeats (changes requested):

app/workers/deletes.php
TorstenDittmann: Actually, the underlying method already throws an exception - so I can remove that completely from the Deletes class 👍🏻

kodumbeats (approved):
Reads a lot better <3

PR #2431, opened 2021-11-22
refactor(db): prepend collections table name

TorstenDittmann:
## What does this PR do?

- prepends `collection_` to all database service collection tables

## Test Plan

*n/a*

## Related PRs and Issues

#1281 

✅


kodumbeats (changes requested):
@TorstenDittmann I'm also getting the realtime failures locally, so your system may be an exception 🤔

TorstenDittmann:
> @TorstenDittmann I'm also getting the realtime failures locally, so your system may be an exception 🤔

Well, we depend on the table name for collection ID I just realized :-/ That means we need more discussion here.

PS: I was only running Database tests 😛

PR #2432, opened 2021-11-23
Implement utopia-php/logger

Meldiron:
## What does this PR do?

Implements error logging to external providers such as Sentry, Raygun or AppSignal.

## Test Plan

Tests for logs are covered in utopia library. Current tests passing is validation we didn't break anything.

## Related PRs and Issues

https://github.com/utopia-php/logger/pull/3

✅


app/controllers/general.php
```diff
@@ -289,19 +291,66 @@
...
+            $log->setUser(new User($user->getId()));
+        }
+
+        $log->setNamespace("http");
+        $log->setServer(App::getEnv("_APP_LOGGING_SERVERNAME", "selfhosted-001"));
+        $log->setVersion($version);
+        $log->setType(Log::TYPE_ERROR);
+        $log->setMessage($error->getMessage());
+
+        $log->setTags([
```
eldadfux, line 320: This is more related to the library implementation. I would like us to avoid signatures that expect assoc array as much as possible. This usually creates a loose interface that is not strictly typed.

This implementation should be closer to our usual pattern:
```php
$log
    ->addTag('string1', 'stringX')
    ->addTag('string2', 'stringY')
;
```

app/controllers/general.php
```diff
@@ -289,19 +291,66 @@
...
+            'method' => $route->getMethod(),
+            'url' => $route->getPath(),
+            'code' => $error->getCode(),
+            'verbose_type' => get_class($error),
+            'projectId' => $project->getId(),
+            'hostname' => $request->getHostname(),
+            'locale' => (string)$request->getParam('locale', $request->getHeader('x-appwrite-locale', '')),
+        ]);
+
+        $log->setExtra([
```
eldadfux, line 330: Same as above. If this options are only for the Sentry adapter, we can add them only for Sentry. We can also add a method that checks whether the current adapter support a specific feature. Would love to share more if this is still not clear enough.

app/init.php
```diff
@@ -163,6 +164,23 @@ function($value) {
...
+    $providerName = App::getEnv('_APP_LOGGING_PROVIDER', '');
+    $providerConfig = App::getEnv('_APP_LOGGING_CONFIG', '');
+
+    if(empty($providerName) || empty($providerConfig)) {
+        return;
+    }
+
+    if(!Logger::hasProvider($providerName)) {
+        throw new Exception("Logging provider not supported. Logging disabled.");
+        return;
```
eldadfux, line 177:
```suggestion
```
No need in return after an exception.

app/realtime.php
```diff
@@ -27,6 +28,40 @@
 
 Runtime::enableCoroutine(SWOOLE_HOOK_ALL);
```
eldadfux, line 29: @TorstenDittmann @Meldiron I think it would be great to have the ability to register a error callback on the WebSocket abstraction library like we have in utopia/fw

app/workers/audits.php
```diff
@@ -12,6 +12,10 @@
 
 class AuditsV1 extends Worker
 {
+    public function getWorkerName(): string {
```
eldadfux, line 15:
```suggestion
    public function getName(): string {
```
The word `worker` is already obvious from the context. Not need to mention it again.

app/workers/certificates.php
```diff
@@ -18,6 +18,10 @@
 
 class CertificatesV1 extends Worker
 {
+    public function getWorkerName(): string {
```
eldadfux, line 21:
```suggestion
    public function getName(): string {
```
The word `worker` is already obvious from the context. Not need to mention it again.

app/workers/deletes.php
```diff
@@ -23,6 +23,10 @@ class DeletesV1 extends Worker
 {
     protected Database $consoleDB;
 
+    public function getWorkerName(): string {
```
eldadfux, line 26:
```suggestion
    public function getName(): string {
```
The word `worker` is already obvious from the context. Not need to mention it again.

app/workers/functions.php
```diff
@@ -101,6 +101,10 @@ class FunctionsV1 extends Worker
 
     public array $allowed = [];
 
+    public function getWorkerName(): string {
```
eldadfux, line 104:
```suggestion
    public function getName(): string {
```
The word `worker` is already obvious from the context. Not need to mention it again.

app/workers/mails.php
```diff
@@ -13,6 +13,10 @@
 
 class MailsV1 extends Worker
 {
+    public function getWorkerName(): string {
```
eldadfux, line 16:
```suggestion
    public function getName(): string {
```

The word `worker` is already obvious from the context. Not need to mention it again.

app/workers/tasks.php
```diff
@@ -17,6 +17,10 @@
 
 class TasksV1 extends Worker
 {
+    public function getWorkerName(): string {
```
eldadfux, line 20:
```suggestion
    public function getName(): string {
```

The word `worker` is already obvious from the context. Not need to mention it again.

app/workers/usage.php
```diff
@@ -11,6 +11,10 @@
 
 class UsageV1 extends Worker
 {
+    public function getWorkerName(): string {
```
eldadfux, line 14:
```suggestion
    public function getName(): string {
```

The word `worker` is already obvious from the context. Not need to mention it again.

app/workers/webhooks.php
```diff
@@ -11,6 +11,10 @@
 
 class WebhooksV1 extends Worker
 {
+    public function getWorkerName(): string {
```
eldadfux, line 14:
```suggestion
    public function getName(): string {
```

The word `worker` is already obvious from the context. Not need to mention it again.

composer.json
```diff
@@ -20,6 +20,12 @@
             "Appwrite\\Tests\\": "tests/extensions"
         }
     },
+    "repositories": [
```
eldadfux, line 23: Reminder to remove this part after we register the lib on Packagist.

docker-compose.yml
```diff
@@ -71,7 +71,7 @@ services:
       - ./psalm.xml:/usr/src/code/psalm.xml
       - ./tests:/usr/src/code/tests
       - ./app:/usr/src/code/app
-      # - ./vendor:/usr/src/code/vendor
+      - ./vendor:/usr/src/code/vendor
```
eldadfux, line 74: Better keep this as a comment by default.

docker-compose.yml
```diff
@@ -158,6 +161,7 @@ services:
     volumes:
       - ./app:/usr/src/code/app
       - ./src:/usr/src/code/src
+      - ./vendor:/usr/src/code/vendor
```
eldadfux, line 164: Better to keep this as comment by default.

src/Appwrite/Resque/Worker.php
```diff
@@ -2,10 +2,17 @@
...
+use Utopia\App;
+use Utopia\CLI\Console;
+use Utopia\Logger\Log;
+use function get_class;
+
 abstract class Worker
 {
     public array $args = [];
 
+    abstract public function getWorkerName(): string;
```
eldadfux, line 14:
```suggestion
    abstract public function getName(): string;
```

The word `worker` is already obvious from the context. Not need to mention it again.

src/Appwrite/Resque/Worker.php
```diff
@@ -2,10 +2,17 @@
 
 namespace Appwrite\Resque;
 
+use Utopia\App;
+use Utopia\CLI\Console;
+use Utopia\Logger\Log;
+use function get_class;
+
 abstract class Worker
```
eldadfux, line 10: Would be great to add doc blocks for all the abstract classes for future devs.

src/Appwrite/Resque/Worker.php
```diff
@@ -14,16 +21,138 @@ abstract public function shutdown(): void;
 
     public function setUp(): void
     {
```
eldadfux, line 23: I think we should have here an error handler callback that we can register on the init file. This implementation couples the Worker and Logger libs in a way we can avoid. 

I think we can try to have a similar implementation to what we have in in utopia/fw, where you pass an error callback and the library triggers it in correct places. This will also allow us to dramatically reduce the code duplication and complexity the error handler has added to this class. 

@Meldiron I would love to chat about this specific pattern later, let me know when you have time.

eldadfux (changes requested):
Great work, left a few comments and suggestions - would love to elaborate more if needed.

app/realtime.php
TorstenDittmann: No argument against it, I am all for it 👍🏻

app/controllers/api/storage.php
```diff
@@ -115,7 +115,7 @@
         $iv = OpenSSL::randomPseudoBytes(OpenSSL::cipherIVLength(OpenSSL::CIPHER_AES_128_GCM));
         $data = OpenSSL::encrypt($data, OpenSSL::CIPHER_AES_128_GCM, $key, 0, $iv, $tag);
 
-        if (!$device->write($path, $data, $mimeType)) {
+        if (!$device->write($path . "/rewerwerwerw.asdas", $data, $mimeType)) {
```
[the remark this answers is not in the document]
Meldiron: Removed.

.env
```diff
@@ -43,3 +43,6 @@ _APP_MAINTENANCE_RETENTION_EXECUTION=1209600
 _APP_MAINTENANCE_RETENTION_ABUSE=86400
 _APP_MAINTENANCE_RETENTION_AUDIT=1209600
 _APP_USAGE_STATS=enabled
+_APP_LOGGING_PROVIDER=sentry
+_APP_LOGGING_CONFIG=9a4ebfef5ab14c8483366039b9de2835;6076895
+_APP_LOGGING_SERVERNAME=selfhosted-001
```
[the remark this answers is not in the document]
Meldiron: Removed `_APP_LOGGING_SERVERNAME` variable and used `\gethostname()` instead.

app/controllers/general.php
Meldiron: Added to utopia-php/logger: https://github.com/utopia-php/logger/commit/5d8cc1bbbfc738ae28ede0d438a1d5318011cf6a

Applied this to: `tags` named array, `breadcrumbs` array, `extra` named array.

app/controllers/general.php
Meldiron: Updated to `addExtra`

app/init.php
Meldiron: Removed return.

app/workers/audits.php
Meldiron: Renamed all `getWorkerName` to `getName` - in abstract class, in every worker and in log where I used this method.

app/workers/certificates.php
Meldiron: Renamed all `getWorkerName` to `getName` - in abstract class, in every worker and in log where I used this method.

app/workers/mails.php
Meldiron: Renamed all `getWorkerName` to `getName` - in abstract class, in every worker and in log where I used this method.

app/workers/deletes.php
Meldiron: Renamed all `getWorkerName` to `getName` - in abstract class, in every worker and in log where I used this method.

app/workers/tasks.php
Meldiron: Renamed all `getWorkerName` to `getName` - in abstract class, in every worker and in log where I used this method.

app/workers/webhooks.php
Meldiron: Renamed all `getWorkerName` to `getName` - in abstract class, in every worker and in log where I used this method.

app/workers/functions.php
Meldiron: Renamed all `getWorkerName` to `getName` - in abstract class, in every worker and in log where I used this method.

app/workers/usage.php
Meldiron: Renamed all `getWorkerName` to `getName` - in abstract class, in every worker and in log where I used this method.

src/Appwrite/Resque/Worker.php
Meldiron: Renamed all `getWorkerName` to `getName` - in abstract class, in every worker and in log where I used this method.

composer.json
Meldiron: Removed.

docker-compose.yml
Meldiron: Removed.

docker-compose.yml
Meldiron: Removed all vendor volumes I added.

src/Appwrite/Resque/Worker.php
Meldiron: I documented Worker class with PHPDoc, hopefully, that's what you requested. Is that okay? https://github.com/appwrite/appwrite/commit/fd1dcae24709251695243d2a3bc31fb2fae288d6

composer.json
Meldiron: Put it back in order to fix Travis tests. Reopened this issue, will remove after publishing logger library.

src/Appwrite/Resque/Worker.php
eldadfux: Perfect 👍

app/realtime.php
Meldiron: Implemented `error` method on utopia-php/websocket to allow registering erro handlers.

app/realtime.php
```diff
@@ -90,6 +125,9 @@ function () use ($register, $db, $cache) {
             Authorization::enable();
             $documentId = $document->getId();
         } catch (\Throwable $th) {
+            var_dump("Error starting");
```
christyjacob4, line 128: Left overs 🙃

Meldiron: Removed, thanks.

app/controllers/general.php
```diff
@@ -289,19 +291,64 @@
...
+App::error(function ($error, $utopia, $request, $response, $layout, $project, $logger, $user, $loggerBreadcrumbs) {
     /** @var Exception $error */
     /** @var Utopia\App $utopia */
     /** @var Utopia\Swoole\Request $request */
     /** @var Appwrite\Utopia\Response $response */
     /** @var Utopia\View $layout */
     /** @var Appwrite\Database\Document $project */
+    /** @var Utopia\Logger\Logger $logger */
+    /** @var Appwrite\Database\Document $user */
+
```
christyjacob4, line 303: Let's add the phpdoc for `$loggerBreadcrumbs` as well

app/init.php
```diff
@@ -344,6 +361,13 @@ function($value) {
 ]);
 
 // Runtime Execution
+App::setResource('logger', function($register) {
+    return $register->get('logger');
```
christyjacob4, line 365: I think this can throw a runtime exception if logger has not been set. Can you confirm this ? We might need to wrap it in a try/ catch in that case

app/workers/tasks.php
```diff
@@ -17,6 +17,10 @@
 
 class TasksV1 extends Worker
 {
+    public function getName(): string {
+        return "tasks";
```
christyjacob4, line 21: I think names of all the workers should be in uppercase

src/Appwrite/Resque/Worker.php
```diff
@@ -2,28 +2,199 @@
...
 namespace Appwrite\Resque;
 
+use Utopia\App;
+use Utopia\CLI\Console;
+use Utopia\Logger\Log;
+
 abstract class Worker
 {
+    /**
+     * Named array holding all information passed into worker alongside a new task.
```
christyjacob4, line 12:
```suggestion
     * Associative array holding all information passed into the worker
```

app/controllers/general.php
Meldiron: Added

app/workers/tasks.php
Meldiron: It will be visible to users in Sentry/RayGun. Do we still want it uppercased? Or lowercase it before pushing? What exactly is the reason behind wanting this uppercase?

app/init.php
Meldiron: From what I can see, it would only throw an error if `$register->set('logger',` did not run, but we run that all the time so it should be good.

It can return `null` tho, so we need to make sure the logger is set. We do that in all API, real-time and worker - on all 3 places.

The only moment we throw an error is when the user defines a provider that is not supported, but that makes sense.

Meldiron: Just tested it to double-check and all seems to be working fine.

composer.json
eldadfux: Package is now available:
https://packagist.org/packages/utopia-php/logger

app/workers/tasks.php
christyjacob4: Okay I was under the impression that other appwrite workers use Title case names. 
But upon checking, they use lower case as well. 

So you can ignore my comment 🙂

composer.json
Meldiron: Removed repositories array

composer.json
```diff
@@ -52,7 +53,7 @@
         "utopia-php/domains": "1.1.*",
         "utopia-php/swoole": "0.2.*",
         "utopia-php/storage": "0.5.*",
-        "utopia-php/websocket": "0.0.*",
+        "utopia-php/websocket": "dev-feat-onerror-callback",
```
eldadfux, line 56: We have released a new tag with the new error handler: 0.1.0

Meldiron: Resolved in next commit

app/controllers/general.php
```diff
@@ -289,19 +291,74 @@
...
     /** @var Exception $error */
     /** @var Utopia\App $utopia */
     /** @var Utopia\Swoole\Request $request */
     /** @var Appwrite\Utopia\Response $response */
     /** @var Utopia\View $layout */
     /** @var Appwrite\Database\Document $project */
+    /** @var Utopia\Logger\Logger $logger */
+    /** @var Utopia\Logger\Log\Breadcrumb[] $loggerBreadcrumbs */
+
+    $user = null;
```
TorstenDittmann, line 304: No need to assign `null` here, If you use `isset()` later on to check if a user is set 👍🏻

app/controllers/general.php
```diff
@@ -289,19 +291,74 @@
...
+    /** @var Appwrite\Database\Document $user */
+
+    try {
+        $user = $utopia->getResource('user');
+    } catch(\Throwable $th) {
+        // All good, user is optional information for logger
+    }
```
TorstenDittmann, line 311: This should only be done if logger is used right?

app/realtime.php
```diff
@@ -90,6 +125,8 @@ function () use ($register, $db, $cache) {
             Authorization::enable();
             $documentId = $document->getId();
         } catch (\Throwable $th) {
+            call_user_func($logError, $th, "createWorkerDocument");
+
             Console::error('[Error] Type: ' . get_class($th));
```
eldadfux, line 130: Can we move all the Console::error logs to the error callback?

app/realtime.php
```diff
@@ -153,6 +190,8 @@ function () use ($register, $db, $cache) {
                 'value' => json_encode($payload)
             ]);
         } catch (\Throwable $th) {
+            call_user_func($logError, $th, "updateWorkerDocument");
+
```
eldadfux, line 194: Can we move all the Console::error logs to the error callback?

eldadfux (changes requested):
Minor changes suggested.

app/controllers/general.php
Meldiron: Applied, thanks.

app/controllers/general.php
Meldiron: Correct, moved it.

app/realtime.php
Meldiron: We can. Moved it.

app/realtime.php
Meldiron: We can. Moved it.

app/controllers/general.php
```diff
@@ -301,23 +301,21 @@
...
-    } catch(\Throwable $th) {
-        // All good, user is optional information for logger
-    }
-
     $version = App::getEnv('_APP_VERSION', 'UNKNOWN');
     $route = $utopia->match($request);
 
     if($logger) {
         if($error->getCode() >= 500 || $error->getCode() === 0) {
+            try {
```
eldadfux, line 309: @Meldiron please remind me why do we use try-catch here? If the user is not available should we get null or an empty user object?

Meldiron: getResurce can throw error if JWT is invalid.

eldadfux: In that case shouldn't we also throw 401 error? Can we implement the error handler on resource creation as part of utopia/fw?

eldadfux: @christyjacob4 explained to me the issue here, as long as this is only and error handler thing, we ca definitely wait and find a more elegant solution later.

eldadfux (approved):

eldadfux:
Amazing work on this @Meldiron !! ❤️

PR #2477, opened 2021-12-10
fix(database/storage): permissions using an admin user

TorstenDittmann:
## What does this PR do?

- checks for an Admin User to skip permissions check on File and Document create/update
- fixes error message
- use `Authorization::getRoles()` over accessing the static property
- make `isAppUser()` and `isPrivilegedUser()` work with non assoc arrays

## Related PRs and Issues

https://github.com/appwrite/appwrite/discussions/2451#discussioncomment-1771016

✅

### DIFF
--- a/app/controllers/api/database.php
+++ b/app/controllers/api/database.php
@@ -1641,17 +1641,17 @@ App::post('/v1/database/collections/:collectionId/documents')
         $data['$write'] = (is_null($write) && !$user->isEmpty()) ? ['user:'.$user->getId()] : $write ?? []; //  By default set write permissions for user
 
         // Users can only add their roles to documents, API keys and Admin users can add any
-        $roles = \array_fill_keys(Authorization::getRoles(), true); // Auth::isAppUser expects roles to be keys, not values of assoc array
+        $roles = Authorization::getRoles();
 
         if (!Auth::isAppUser($roles) && !Auth::isPrivilegedUser($roles)) {
             foreach ($data['$read'] as $read) {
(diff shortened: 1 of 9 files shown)


app/controllers/api/database.php
```diff
@@ -1640,16 +1640,19 @@ function createAttribute($collectionId, $attribute, $response, $dbForInternal, $
...
+                    throw new Exception('Read permissions must be one of: ('.\implode(', ', array_keys($roles)).')', 400);
+                }
             }
-        }
-        foreach ($data['$write'] as $write) {
-            if (!Auth::isAppUser($roles) && !Authorization::isRole($write)) {
-                throw new Exception('Write permissions must be one of: ('.\implode(', ', $roles).')', 400);
+            foreach ($data['$write'] as $write) {
+                if (!Authorization::isRole($write)) {
+                    throw new Exception('Write permissions must be one of: ('.\implode(', ', array_keys($roles)).')', 400);
```
kodumbeats, line 1654: We should probably fix this instead of hacking around the data structure mismatch 😆 
The output of `Authorization::getRoles()` should match the input of `Auth::isAppUser()` - an indexed array is probably the easiest instead of associative.

kodumbeats (changes requested):

app/controllers/api/database.php
TorstenDittmann: Done 🥳

kodumbeats (approved):
LGTM 👍🏻

PR #2485, opened 2021-12-13
Feat-refactor-storage-resource

lohanidamodar:
## What does this PR do?

Refactor to set and get storage from device from resource

## Test Plan

Existing tests

## Related PRs and Issues

YES


app/controllers/api/storage.php
```diff
@@ -1067,7 +878,7 @@
         $fileLogos = Config::getParam('storage-logos');
 
         $date = \date('D, d M Y H:i:s', \time() + (60 * 60 * 24 * 45)).' GMT';  // 45 days cache
-        $key = \md5($fileId.$width.$height.$gravity.$quality.$borderWidth.$borderColor.$borderRadius.$opacity.$rotation.$background.$storage.$output);
```
Meldiron, line 1070: A piece of functionality was removed here, just double-checking if that is okay.

Meldiron (approved):
Looks good to me

app/init.php
```diff
@@ -135,6 +139,7 @@
 Config::load('providers', __DIR__.'/config/providers.php');
 Config::load('platforms', __DIR__.'/config/platforms.php');
 Config::load('collections', __DIR__.'/config/collections.php');
+Config::load('subCollections', __DIR__.'/config/sub_collections.php');
```
eldadfux, line 142: What is a sub_collection? Why do we use it? This is a design change that require more discussion.

lohanidamodar: Hey, yes 
1. We need to create pre-defined files collection for each bucket
2. Previously the file collection config was directly kept in the storage controller create bucket which needed a refactor
3. So sub-collection is configs for such collection that we create per parent collection
4. We can use this later for - function tags collection if we create separate tags collections per function, and so on on other related case
5. Only similar thing we previously do is the documents collection, however, those don't have pre-defined config, as those collections are created via API by users

eldadfux: It wasn't intuitive for me. I was expecting to just fetch the files collections and create it when a bucket is created similar to how we set collection when a project is created. Would it make more sense to add a `sub => true` attribute and skip this collection when initing a project?

We should also take under consideration the `console` project that is initialized on the first setup of the server, it needs a "default" bucket to upload project logos and other future assets.

lohanidamodar: adding `sub=>true` might make more sense. Regarding console project, the files collection will live in external db wouldn't that be a problem?

lohanidamodar: how about `project=>true` so that collections are created for each project and false would imply will not be created for the project. Those that are false, we can use anywhere by accessing directly using the key?

lohanidamodar: Ok, I've an issue, the project logo tries to upload project logo in the project scope so it's not console project we require the default bucket in, its with every project we will require default bucket and files collection for that bucket not the console project? I just setup console project with default bucket and default files collection and then realized this while testing. Also, if we create default bucket for each project we should prevent using this (though it will throw duplicate exception) we should throw 400 error and mark default bucket as reserved?

lucalves (approved):

PR #2486, opened 2021-12-13
Feat prefs improvement

Meldiron:
## What does this PR do?

- Updates docs to reflect current behaviour
- Add API level size validation (so DB does not have to throw an error on too long prefs)

The maximum allowed size is `65535`, which is 64kB.

## Opened question

- I decided **not** to work on Teams prefs API, because it would mean a lot of copy&paste. If we agree on new endpoints as suggested by Eldad, we can define a new API group (personalizations) that will be able to handle both users&teams in one endpoint

## Test Plan

Added new test

## Related PRs and Issues

https://github.com/utopia-php/framework/pull/40

✅


tests/e2e/Services/Account/AccountBase.php
```diff
@@ -3,6 +3,7 @@
 namespace Tests\E2E\Services\Account;
 
 use Tests\E2E\Client;
+use function array_merge;
```
TorstenDittmann, line 6: Lets not important root level functions like that 👍🏻

composer.json
```diff
@@ -38,7 +38,7 @@
         "appwrite/php-clamav": "1.1.*",
         "appwrite/php-runtimes": "0.6.*",
 
-        "utopia-php/framework": "0.19.*",
+        "utopia-php/framework": "0.19.1",
```
TorstenDittmann, line 41: This should be fixed now, you can use `0.19.*` again 👍🏻

Meldiron:
Reverted version limitation

composer.json
Meldiron: Perfect 👌

kodumbeats (changes requested):
I think the root of this problem is the size of this field in `app/config/collections.php` and can't be solved in the route:

- By the time the request gets through middleware, `$prefs` is already an assoc array, not JSON string, so I don't exactly know how/why `strlen()` even works here 🤔 
- By setting this field to `16384`, we're actually creating a MEDIUMTEXT field (max 64Kb) when we actually want a TEXT field for this limit: https://github.com/utopia-php/database/blob/main/src/Database/Adapter/MariaDB.php#L989-L991
- In theory, the database should throw an exception regardless of what we do in the code at this point, if the field is actually TEXT.

app/controllers/api/users.php
```diff
@@ -591,6 +591,10 @@
         /** @var Utopia\Database\Database $dbForInternal */
         /** @var Appwrite\Stats\Stats $usage */
 
+        if(\strlen($prefs) > 16384) {
```
eldadfux, line 594: This limit should be a part of a validator for better consistency. We also discussed mentioning the size limit as part of the param description. I think the extension can be located at `src/Appwrite/Auth/Validator/Prefs` unless someone has a better idea. 

> Note:  when adding a new param we need to make sure it is translated to the proper format for the Swagger / OpenAPI exports.

eldadfux (changes requested):

Meldiron:
To get tests to pass, we need this merged and released: https://github.com/appwrite/appwrite/pull/2486

I could not use the repositories trick, some other libraries did not like the version as branch number instead of a number.

```
utopia-php/database[0.12.0, ..., 0.12.1] require utopia-php/framework 0.*.* -> found utopia-php/framework[0.1.0, ..., 0.19.3] but it conflicts with your root composer.json require (dev-fix-improve-validators).
```

PR #2505, opened 2021-12-16
feat: replace attributeId and indexId with key

TorstenDittmann:
## What does this PR do?

- replaces `attributeId` and `indexId` with `key` to avoid confusion

### Why don't we replace it with `$id`?

We already use `$id` internally and concatenate the collection id and key. This is needed, since otherwise keys would need to be unique across collections.

## Test Plan

- adapted existing tests

✅


christyjacob4 (approved):

christyjacob4:
Looks Good 🚀

wess (approved):

PR #2522, opened 2021-12-22
feat: cursor metadata

TorstenDittmann:
## What does this PR do?

- very much not happy with the possible implementations, since it will be a lot of repeated logic across all endpoints that offer cursor pagination

## Test Plan

- extended


eldadfux:
Could we abstract the replicated logic by having it as part of something like: `$database->findCursorOf`?

PR #2526, opened 2021-12-28
fix: database query value limits

TorstenDittmann:
## What does this PR do?

- check if there are more than 100 values for a query used and throw an error

## Test Plan

- added tests

## Related PRs and Issues

(Write your answer here.)

### DIFF
--- a/app/controllers/api/database.php
+++ b/app/controllers/api/database.php
@@ -1717,7 +1717,15 @@ App::get('/v1/database/collections/:collectionId/documents')
             }
         }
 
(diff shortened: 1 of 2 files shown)


Meldiron (approved):
Looks good to me

eldadfux (approved):
Looks good, we should also have a limit on the array of queries size.

TorstenDittmann:
> Looks good, we should also have a limit on the array of queries size.

This needs to be part of `utopia-php/framework` itself as part of the `ArrayList` validator 👍🏻

PR #2562, opened 2022-01-05
fix: migration for leftover documents on deleted projects

TorstenDittmann:
## What does this PR do?

- some documents can be left for previously deleted projects, this PR skips them


Meldiron (approved):
Looks good

PR #2563, opened 2022-01-05
fix: ui for functions variables

TorstenDittmann:
## What does this PR do?

- `data-forms-key-value` was executed too early


Meldiron (approved):
LGTM

Issue #2593, opened 2022-01-12
🐛 Bug Report: Membership verification for the same id possible several times


Malte2036:
### 👟 Reproduction steps

1. Create a team
2. send a [membership invitation to a user](https://appwrite.io/docs/client/teams#teamsCreateMembership)
3. user recieves email with an membership verification link
4. user clicks on the link multiple times and triggers the [verification endpoint](https://appwrite.io/docs/client/teams#teamsUpdateMembershipStatus) multiple times with the same parameters and the same membershipId.

### 👍 Expected behavior

The first time you click on the link, the user will be added to the team and the sum variable will be incremented.

If you click on the link again and start the verification process, the endpoint should return an error because the membershipId is already confirmed. And nothing should change for the team itself.

### 👎 Actual Behavior

The first time you click on the link, the user will be added to the team and the sum variable will be incremented.

If you click on the link again and start the verification process, *no* error occurs and the sum variable of the team is incremented again. 

This means that if, for example, the verification endpoint is triggered 10 times with the same parameters, the sum variable is increased by 10.

### 🎲 Appwrite version

Different version (specify in environment)

### 💻 Operating system

Linux

### 🧱 Your Environment

I use appwrite v0.12.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

TorstenDittmann:
This got fixed and is released with `0.13.1` 👏🏻

PR #2594, opened 2022-01-12
fix: check whether the membershipId has been confirmed to avoid recon…

Malte2036:
## What does this PR do?

The changes check when calling the ["Update Team Membership Status" endpoint](https://appwrite.io/docs/client/teams#teamsUpdateMembershipStatus) whether the membershipId parameter has already been confirmed to prevent the same membershipId from being confirmed more than once. 

If the membershipId has already been confirmed, an error is thrown and the process is aborted.

## Related PRs and Issues

Fix for this bug  #2593. :bug:

yes


app/controllers/api/teams.php
```diff
@@ -627,6 +627,10 @@
             throw new Exception('Team IDs don\'t match', 404);
         }
 
+        if ($membership->getAttribute('confirm') !== false) {
+            throw new Exception('Membership already confirmed', 400);
```
eldadfux, line 631:
```suggestion
            throw new Exception('Membership already confirmed', 409);
```

I think 409 will be more descriptive here.

eldadfux: Reference: https://appwrite.io/docs/error-codes

app/controllers/api/teams.php
```diff
@@ -627,6 +627,10 @@
             throw new Exception('Team IDs don\'t match', 404);
         }
 
+        if ($membership->getAttribute('confirm') !== false) {
```
eldadfux, line 630:
```suggestion
        if ($membership->getAttribute('confirm') === true) {
```

Checking for acceptance feels a bit safer.

eldadfux (changes requested):
Good catch! Left a few suggestions.

eldadfux:
Waiting for tests to pass, the docker rate limit seems to make problems.

TorstenDittmann:
Restarted the CI 👍🏻

Malte2036:
Tests should be fixed and added new e2e-test :robot:

Malte2036:
The `e2e/Services/Functions` tests fail even without my changes.
All `e2e/Services/Teams` tests work, so I think the pull request is ready to merge.

eldadfux (approved):

eldadfux:
@Malte2036 I think you need to sync with master in order for tests to pass.

Malte2036:
Ready to merge @eldadfux? 🤔

TorstenDittmann:
Looks like we missed that one, will take care of this 👍🏻

TorstenDittmann:
Merged it and will be released soon! Thanks alot 🥳 

Sorry for the delay - the team was in release stress 🙇🏻

PR #2599, opened 2022-01-12
Fix: 409 for createCollection

Meldiron:
## What does this PR do?

In current master, `createCollection` first tries to create table, then document in `collections`.

That means, 409 never exists, because it first ends with error500 because of attempt to create duplicate key.

This PR solves this issue by first trying to create document, then create the table itself. I also added test to test for this in future.

## Test Plan

Added 1 new test

✅


kodumbeats (changes requested):
First off, great find! I've got a few tweaks, and then this'll be G2G:

- Let's keep the `createCollection()` call inside the try-catch, right after `createDocument()`
- It looks like `createCollection()` also throws `Utopia\Database\Exception\Limit`, so I think we should explicitly catch those, similar to what we do in the `createDocument` route:
https://github.com/appwrite/appwrite/blob/babd77f663e17491214cf4302e2d3c0c76f35a7e/app/controllers/api/database.php#L1648-L1653

ref in database: https://github.com/utopia-php/database/blob/main/src/Database/Database.php#L322

eldadfux:
Synced with master to try and pass tests.

kodumbeats (approved):

eldadfux (approved):

PR #2600, opened 2022-01-12
Wrap unique() in double quotes to better indicate that it's a literal…

gewenyu99:
… string

## What does this PR do?

In the documentation, many endpoints allow the string `"unique()"` to be used to indicate that a random uuid should be generated. Currently, the documentation displays `"unique()"` as plain text. 

Screen Shot 2022-01-12 at 3 56 55 PM

This can mislead readers (like myself), to think that there is a function `unique()` to generate a uuid on the client side. The current approach in the documentation uses single \` signs around `unique` to indicate a code block, but it does not get rendered in the actual docs.

The proposed fix, wrapping `unique()` in double quotes, is a temporary fix to clearly indicate that it's a string. A more permanent fix would be to fix the rendering of inline code snippets in the docs.

https://appwrite.io/docs/server/functions

## Related PRs and Issues

None that I know of

### Have you read the [Contributing Guidelines on issues]

Yes :D


eldadfux (approved):
LGTM

I think we might want to consider a built in class like we have for `Query`, but it will be a bit harder to explain this across different languages as syntax can be 100% the same.

cc @lohanidamodar

eldadfux:
@gewenyu99 congrats on your first appwrite/appwrite PR!! 🥳

Issue #1494, opened 2021-08-18
Project gets abandoned when last member left


iamdineshkumar:
## 🐛 Bug Report

Project owner or last member have leave option in Members. If project owner leaves then particular project gets abandoned. 

## Have you spent some time to check if this issue has been raised before?

Yes

## To Reproduce

Just create project and leave from members but still project works through sdk but not able to access it.

## Expected behavior
It should either not allowed last user or owner to leave and ask to delete project instead or leaving last person also delete the project.
## Actual Behavior
It just remove user from access but project exists without anyone to access further.
## Your Environment
Docker container created in [justcloudify.com](https://justcloudify.com/)

TorstenDittmann:
Yeah, we should probably just not show the button to leave the team when you are the only member of that project left.

eldadfux:
This issue was pending for too long! Raised a quick PR! #2615

TorstenDittmann:
We just released 0.13 and this should be fixed 👍🏻 

You can find the announcement [here](https://dev.to/appwrite/announcing-appwrite-013-with-major-upgrades-to-storage-and-functions-3hpf).

Thank you a lot for reporting this 🙏🏻

PR #2615, opened 2022-01-15
Fix for console memberships

eldadfux:
## What does this PR do?

Deny last user from leaving a project by disabling the leave button, and adds a redirect to home console page if current user is the one leaving the project instead of getting permission error.

Because the project membership is using the teams API under the hood, I left this fix only on the UI level, not sure if we want to enforce this behavior on all Appwrite projects.

## Test Plan

N/A

## Related PRs and Issues

#1494

Many times.

### DIFF
--- a/app/views/console/settings/index.phtml
+++ b/app/views/console/settings/index.phtml
@@ -453,7 +453,7 @@ $smtpEnabled = $this->getParam('smtpEnabled', false);
                     <div class="box margin-bottom">
                         <ul data-ls-loop="members.memberships" data-ls-as="member" class="list">
                             <li class="clear">
-                                <form class="pull-end"
+                                <form class="pull-end" data-ls-if="{{account.$id}} !== {{member.userId}}"
                                     data-analytics
                                     data-analytics-activity
                                     data-analytics-event="submit"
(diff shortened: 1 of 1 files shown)


app/views/console/settings/index.phtml
```diff
@@ -473,7 +473,33 @@ $smtpEnabled = $this->getParam('smtpEnabled', false);
                                     <input name="teamId" data-ls-attrs="id=leave-teamId-{{member.$id}}" type="hidden" data-ls-bind="{{console-project.teamId}}">
                                     <input name="membershipId" data-ls-attrs="id=leave-membershipId-{{member.$id}}" type="hidden" data-ls-bind="{{member.$id}}">
 
-                                    <button class="danger">Leave</button>
+                                    <button data-ls-if="1 === {{members.memberships.length}}" class="danger" disabled>Leave</button>
+                                    <button data-ls-if="1 < {{members.memberships.length}}" class="danger">Leave</button>
```
Meldiron, line 477: This ls-if shouldnt be required. If there is member ID that is not yours, there are at least 2 members already

Meldiron, line 477: This is for non-me users. Should the text be `Kick` or `Remove`?

app/views/console/settings/index.phtml
```diff
@@ -473,7 +473,33 @@ $smtpEnabled = $this->getParam('smtpEnabled', false);
...
+                                    data-success-param-trigger-events="teams.deleteMembership"
+                                    data-success-param-redirect-url="/console"
+                                    data-failure="alert"
+                                    data-failure-param-alert-text="Failed to Remove Member"
+                                    data-failure-param-alert-classname="error">
+
+                                    <input name="teamId" data-ls-attrs="id=leave-teamId-{{member.$id}}" type="hidden" data-ls-bind="{{console-project.teamId}}">
+                                    <input name="membershipId" data-ls-attrs="id=leave-membershipId-{{member.$id}}" type="hidden" data-ls-bind="{{member.$id}}">
+
+                                    <button data-ls-if="1 === {{members.memberships.length}}" class="danger" disabled>Leave</button>
```
Meldiron, line 501: Are we okay by doing `disabled` check only client-side? In other words, is there cleanup process if the only member leaves project?

Meldiron (changes requested):
Left few concerns.

app/views/console/settings/index.phtml
eldadfux: We don't care about the member userId to block the `leave` button, we just want to make sure the project is not abandoned no matter who is the last user. You can leave the project if you want to. The user ID check is for redirecting the user to the console home after leaving the project instead of getting an errored project list since you don't have access to the project anymore.

eldadfux: Good call. I changed both to `Remove` for consistency

app/views/console/settings/index.phtml
eldadfux: Mentioned it in the beginning of the PR. Not sure if we want to force this behavior on all team, maybe we can add this as a feature in the future. Wouldn't block this PR for that, we can tackle it later.

Meldiron (approved):
lgtm

PR #2629, opened 2022-01-17
fix: hide internal id from documents

TorstenDittmann:
## What does this PR do?

- hides internalId using response filter

## Test Plan

- added tests


src/Appwrite/Utopia/Response/Model/DocumentList.php
```diff
@@ -0,0 +1,17 @@
...
+namespace Appwrite\Utopia\Response\Model;
+
+use Utopia\Database\Document;
+
+class DocumentList extends BaseList
+{
+    public function filter(Document $document): Document
+    {
+        foreach ($document->getAttribute('documents', []) as $node) {
+            $node->removeAttribute('$internalId');
```
eldadfux, line 12: Are we sure we need to loop here? Isn't the nested model taking care of it?

TorstenDittmann: Good catch! For some reason tests were failing locally - but shouldn't have.

Reverted that part.

PR #2631, opened 2022-01-18
Updated password length min requirment in the console

eldadfux:
## What does this PR do?

Updated the console validations to match #2439 

## Test Plan

N/A

## Related PRs and Issues

#2439 

Yes.

### DIFF
--- a/app/views/console/users/index.phtml
+++ b/app/views/console/users/index.phtml
@@ -174,7 +174,7 @@ $smtpEnabled = $this->getParam('smtpEnabled', false);
                         <input type="email" class="full-width" id="user-email" name="email" required autocomplete="off" />
 
                         <label for="user-password">Password</label>
-                        <input type="password" class="full-width" id="user-password" name="password" required pattern=".{6,}" title="Six or more characters" autocomplete="off" />
+                        <input type="password" class="full-width" id="user-password" name="password" required pattern=".{8,}" title="Six or more characters" autocomplete="off" />
 
                         <hr />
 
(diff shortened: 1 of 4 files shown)


TorstenDittmann:
We might wanna use the `minlength` attribute here instead of `pattern`. This will also trigger the proper event to the browser => https://developer.mozilla.org/en-US/docs/Web/API/ValidityState/tooShort

Which then will give native a11y notes from the browser to the user 👍🏻

kodumbeats (approved):
I'm not up-to-speed on the difference between `pattern` and `minlength`, so I defer to y'all there. If we do change that, let's remember to change the other form from https://github.com/appwrite/appwrite/pull/2439:
https://github.com/appwrite/appwrite/blob/61ffe18d2b229b788e4194abf8ba47768f83d63b/app/views/home/auth/signup.phtml#L48

Otherwise, LTGM 👍🏻

TorstenDittmann:
> I'm not up-to-speed on the difference between `pattern` and `minlength`, so I defer to y'all there. If we do change that, let's remember to change the other form from #2439:
> 
> https://github.com/appwrite/appwrite/blob/61ffe18d2b229b788e4194abf8ba47768f83d63b/app/views/home/auth/signup.phtml#L48
> 
> Otherwise, LTGM 👍🏻

https://svelte.dev/repl/3d6730a47fd643299cbd74b9bde9e0f9?version=3.46.2

Try entering less than 8 chars and you will get a proper note whats wrong in terms of length 👍🏻

kodumbeats:
> https://svelte.dev/repl/3d6730a47fd643299cbd74b9bde9e0f9?version=3.46.2
> 
> Try entering less than 8 chars and you will get a proper note whats wrong in terms of length 👍🏻

Ah, I see. I guess the browser doesn't have a friendly way to decode the pattern. Looks clean

PR #2633, opened 2022-01-18
Fix function authorization exception

abnegate:
## What does this PR do?

Fixes an exception thrown when a function is executed without any execute permissions:

```
appwrite  | [Error] Method: POST
appwrite  | [Error] URL: /v1/functions/:functionId/executions
appwrite  | [Error] Type: Error
appwrite  | [Error] Message: Object of class Utopia\Database\Document could not be converted to string
appwrite  | [Error] File: /usr/src/code/vendor/utopia-php/database/src/Database/Validator/Authorization.php
appwrite  | [Error] Line: 61
```

Due to the `Authorization` implementation being updated to utopia-php/database from appwrite, where the appwrite implementation used to take a `Document` as the first parameter.

## Test Plan

Test added in `e2e/Services/Functions/FunctionsCustomClientTest.php` that creates a function with no execution permissions then checks that attempting to execute said function response with a 401.

### DIFF
--- a/app/controllers/api/functions.php
+++ b/app/controllers/api/functions.php
@@ -714,7 +714,7 @@ App::post('/v1/functions/:functionId/executions')
             throw new Exception('Tag not found. Deploy tag before trying to execute a function', 404);
         }
 
-        $validator = new Authorization($function, 'execute');
+        $validator = new Authorization('execute');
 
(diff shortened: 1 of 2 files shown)


TorstenDittmann (approved):

eldadfux:
Great work @abnegate !

PR #2636, opened 2022-01-18
Add float validator case to spec to prevent fallthrough

abnegate:
## What does this PR do?

SDK method's required string parameters for `min`, `max`, `default` when calling `createFloatAttribute` or `createIntegerAttribute`, setting `$loose = true` when constructing the validators will allow those calls to succeed.

## Test Plan

## Related PRs and Issues

Closes https://github.com/appwrite/sdk-for-kotlin/issues/17

Yes


TorstenDittmann:
Shouldn't the SDK send the actual float value instead of a String? 🤔

christyjacob4:
@TorstenDittmann doesn't the POST body automatically get converted to a string nevertheless ??  👀

eldadfux:
> @TorstenDittmann doesn't the POST body automatically get converted to a string nevertheless ?? 👀

The JSON payload on POST should return correct types.

eldadfux:
We probably need to have the same for the open-api 3 class.

eldadfux (approved):

eldadfux:
Thanks @abnegate!

PR #2638, opened 2022-01-18
Fix list collections when passing `search` parameter

abnegate:
## What does this PR do?

Add fulltext index for collections collection `name` column to fix `listCollections` when using the `search` parameter.

Currently throws a PDOException:

```
[Error] File: /usr/src/code/vendor/utopia-php/framework/src/App.php
[Error] Line: 756
[Error] Type: PDOException
[Error] Message: Can't find FULLTEXT index matching the column list
[Error] File: @swoole-src/library/core/Database/PDOStatementProxy.php
[Error] Line: 64
```

## Test Plan

Used the `listCollections` function from SDK's passing a value for the `search` parameter

## Related PRs and Issues

Closes https://github.com/appwrite/sdk-for-kotlin/issues/18

### DIFF
--- a/app/config/collections.php
+++ b/app/config/collections.php
@@ -102,6 +102,13 @@ $collections = [
                 'lengths' => [1024],
                 'orders' => [Database::ORDER_ASC],
             ],
+            [
+                '$id' => '_fulltext_name',
+                'type' => Database::INDEX_FULLTEXT,
+                'attributes' => ['name'],
+                'lengths' => [256],
+                'orders' => [Database::ORDER_ASC],
+            ],
         ],
     ],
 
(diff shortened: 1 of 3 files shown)


eldadfux:
@abnegate you're on 🔥 today!

TorstenDittmann:
We are actually using the `search` parameter which contains the collection id and name => https://github.com/appwrite/appwrite/blob/master/app/controllers/api/database.php#L180

We only need to change this https://github.com/appwrite/appwrite/blob/master/app/controllers/api/database.php#L233 to the `search` attribute 👍🏻

TorstenDittmann (approved):

PR #2641, opened 2022-01-18
Functions Refactor

PineappleIOnic:
## What does this PR do?

This PR is for the upcoming functions refactor introducing the new execution model to allow for significantly faster execution of functions and syncronous executions.

## Test Plan

New tests have been written for syncronous execution and current tests have been updated for the new model.

Yes


app/controllers/api/functions.php
```diff
@@ -446,17 +494,22 @@
     ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
     ->label('sdk.response.model', Response::MODEL_TAG)
     ->param('functionId', '', new UID(), 'Function ID.')
-    ->param('command', '', new Text('1028'), 'Code execution command.')
+    ->param('entrypoint', '', new Text('1028'), 'Entrypoint File.')
     ->param('code', [], new File(), 'Gzip file with your code package. When used with the Appwrite CLI, pass the path to your code directory, and the CLI will automatically package your code. Use a path that is within the current directory.', false)
+    ->param('automaticDeploy', false, new Boolean(true), 'Automatically deploy the function when it is finished building.', false)
```
eldadfux, line 499:
```suggestion
    ->param('deploy', false, new Boolean(true), 'Automatically deploy the function when it is finished building.', false)
```

Seems more consistent

app/controllers/api/functions.php
```diff
@@ -498,24 +551,77 @@
...
+                $dbForProject->updateDocument('tags', $tag->getId(), $tag);
+            }
+        }
         
         $tagId = $dbForProject->getId();
         $tag = $dbForProject->createDocument('tags', new Document([
             '$id' => $tagId,
-            '$read' => [],
-            '$write' => [],
+            '$read' => ['role:all'],
```
eldadfux, line 571: Why are we giving here role:all?

app/controllers/api/functions.php
```diff
@@ -725,13 +874,13 @@
         $execution = Authorization::skip(fn() => $dbForProject->createDocument('executions', new Document([
             '$id' => $executionId,
             '$read' => (!$user->isEmpty()) ? ['user:' . $user->getId()] : [],
-            '$write' => [],
+            '$write' => ['role:all'],
```
eldadfux, line 877: Could be very risky, any reason for the change?

app/controllers/api/functions.php
```diff
@@ -865,3 +1053,141 @@
 
         $response->dynamic($execution, Response::MODEL_EXECUTION);
     });
+
+App::get('/v1/builds')
+    ->groups(['api', 'functions'])
+    ->desc('Get Builds')
```
eldadfux, line 1059:
```suggestion
    ->desc('List Builds')
```

app/controllers/api/functions.php
```diff
@@ -865,3 +1053,141 @@
...
+
+            if ($cursorExecution->isEmpty()) {
+                throw new Exception("Execution '{$cursor}' for the 'cursor' value not found.", 400);
+            }
+        }
+
+        $queries = [];
+
+        if (!empty($search)) {
+            $queries[] = new Query('search', Query::TYPE_SEARCH, [$search]);
```
eldadfux, line 1090: There is no search field in the data structure, do we have tests for this?

app/controllers/api/functions.php
```diff
@@ -650,6 +767,38 @@
             throw new Exception('Tag not found', 404);
         }
 
+        // Request executor to delete tag containers
```
eldadfux, line 770: Should be moved to a worker, no need to wait for response. We might have to introduce a `deleting` status to prevent conflicts while being deleted.

app/controllers/api/functions.php
```diff
@@ -569,6 +675,15 @@
         $results = $dbForProject->find('tags', $queries, $limit, $offset, [], [$orderType], $cursorTag ?? null, $cursorDirection);
         $sum = $dbForProject->count('tags', $queries, APP_LIMIT_COUNT);
 
+        // Get Current Build Data
+        foreach ($results as &$tag) {
```
eldadfux, line 679: Again, I think we can skip this output and include it in the console from a second API call.

app/controllers/api/functions.php
```diff
@@ -865,3 +1053,141 @@
 
         $response->dynamic($execution, Response::MODEL_EXECUTION);
     });
+
+App::get('/v1/builds')
```
eldadfux, line 1057: Shouldn't build be nested under their tag? They should probably also use the function read and write permissions and not the execution ones. (`execution.read` -> `functions.read` and `execution.write` -> `functions.write`)

app/controllers/api/functions.php
```diff
@@ -629,10 +744,12 @@
     ->inject('response')
     ->inject('dbForProject')
     ->inject('usage')
-    ->action(function ($functionId, $tagId, $response, $dbForProject, $usage) {
+    ->inject('project')
+    ->action(function ($functionId, $tagId, $response, $dbForProject, $usage, $project) {
```
eldadfux, line 748: When we delete a tag we should delete all builds

eldadfux (changes requested):
Gave an initial review, added some more comments and fixes in this PR: https://github.com/appwrite/appwrite/pull/2673

docker-compose.yml
```diff
@@ -345,11 +347,66 @@ services:
...
+      - DOCKERHUB_PULL_USERNAME
+      - DOCKERHUB_PULL_PASSWORD
+
+  appwrite-executor:
+    container_name: appwrite-executor
+    entrypoint:
+      - php
+      - -e
+      - /usr/src/code/app/executor.php
+      - -dopcache.preload=opcache.preload=/usr/src/code/app/preload.php
```
eldadfux, line 367: We should create an executable file on our Dockerfile, like we do for other services, same for the `compose.phtml` file.

app/config/collections.php
```diff
@@ -1989,6 +2044,133 @@
...
+                'format' => '',
+                'size' => 16384,
+                'signed' => true,
+                'required' => false,
+                'default' => '',
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => 'buildTime',
```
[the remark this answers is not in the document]
lohanidamodar: @PineappleIOnic can we add functionId to builds collection, we would want to aggregate build time per function, and I'll need functionId here to do that?

app/config/collections.php
```diff
@@ -1925,9 +1925,20 @@
...
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
             [
                 'array' => false,
-                '$id' => 'command',
+                '$id' => 'entrypoint',
```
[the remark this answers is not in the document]
PineappleIOnic: This change was not required however without it the request may be confusing as we are no longer using full commands to launch servers. Instead we are directly importing the user's function file.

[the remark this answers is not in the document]
PineappleIOnic: This change is breaking however most of the functions refactor itself is breaking as old functions will cease to work due to the new import based functions system.

CONTRIBUTING.md
```diff
@@ -282,7 +282,7 @@ docker buildx build --platform linux/amd64,linux/arm64,linux/arm/v6,linux/arm/v7
 ```
 **Build Functions Runtimes**
 
-The Runtimes for all supported cloud functions (multicore builds) can be found at the [appwrite/php-runtimes](https://github.com/appwrite/php-runtimes) repository.
+The Runtimes for all supported cloud functions (multicore builds) can be found at the [appwrite/runtimes](https://github.com/appwrite/runtimes) repository.
```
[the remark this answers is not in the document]
PineappleIOnic: We do have a contribution guide which explains this currently in a PR. https://github.com/appwrite/runtimes/pull/27

app/config/collections.php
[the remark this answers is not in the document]
PineappleIOnic: I'm not sure if we can add functionId to builds collection as it's supposed to stay detached from functions so it can be reused for static hosting.

docs/references/functions/list-builds.md
```diff
@@ -0,0 +1 @@
+Get a list of all the current user build logs. You can use the query params to filter your results. On admin mode, this endpoint will return a list of all of the project's executions. [Learn more about different API modes](/docs/admin).
```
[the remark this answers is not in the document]
PineappleIOnic: I'm currently 50/50 whether to keep it included, I originally added it for debugging but at the same time it could help with adding build status to the dashboard without me needing to clone the latest build information on the tag.

app/controllers/api/functions.php
PineappleIOnic: Builds are not currently nested in their tag since builds need to be seperate from functions so static websites can use them aswell.

docker-compose.yml
PineappleIOnic: Done 👍

app/controllers/api/functions.php
PineappleIOnic: not yet but I'll write some

PineappleIOnic: not currently, but I can write some

src/Appwrite/Utopia/Response/Model/Tag.php
```diff
@@ -28,17 +28,52 @@ public function __construct()
...
+                'example' => 'ready',
+            ])
+            ->addRule('buildId', [
+                'type' => self::TYPE_STRING,
+                'description' => 'The current build ID.',
                 'default' => '',
-                'example' => 'python-3.8',
+                'example' => '5e5ea5c16897e',
+            ])
+            ->addRule('buildStdout', [
```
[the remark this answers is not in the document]
christyjacob4: Also remove the `buildStatus` attribute from the `tags` collection

app/controllers/api/functions.php
PineappleIOnic: Tests have been written.

app/config/collections.php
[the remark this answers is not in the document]
lohanidamodar: @PineappleIOnic Then  can we sync the buildTime to the tags collection, like we are syncing other attributes? That way I can use the tags collection which already has the functionId right?

app/executor.php
```diff
@@ -0,0 +1,1402 @@
...
+                hostname: $container,
+                mountFolder: $tagPathTargetDir,
+            );
+
+            if (empty($id)) {
+                throw new Exception('Failed to create container');
+            }
+
+            // Add to network
+            $orchestration->networkConnect($container, 'appwrite_runtimes');
```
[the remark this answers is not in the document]
PineappleIOnic: I'll start looking into this now.

app/config/collections.old.php
```diff
@@ -1322,8 +1322,8 @@
             ],
             [
                 '$collection' => Database::SYSTEM_COLLECTION_RULES,
-                'label' => 'Tag',
-                'key' => 'tag',
+                'label' => 'Deployment',
```
lohanidamodar, line 1325: Do we need these changes in the old collections?

app/config/collections.php
```diff
@@ -1894,14 +1894,14 @@
                 'attributes' => ['search'],
                 'lengths' => [2048],
                 'orders' => [Database::ORDER_ASC],
-            ],
+            ]
         ],
     ],
 
-    'tags' => [
+    'deployments' => [
```
lohanidamodar, line 1901: Make sure to update usage worker that calculates the storage taken by tags

app/config/collections.php
```diff
@@ -1989,6 +2044,133 @@
...
+                'format' => '',
+                'size' => 16384,
+                'signed' => true,
+                'required' => false,
+                'default' => new stdClass(),
+                'array' => false,
+                'filters' => ['json'],
+            ],
+            [
+                '$id' => 'sourceType',
```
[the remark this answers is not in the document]
lohanidamodar: Not sure what sourceType is for?

app/views/install/compose.phtml
```diff
@@ -153,6 +155,50 @@ services:
...
+    entrypoint: executor
+    container_name: appwrite-executor
+    stop_signal: SIGINT
+    networks:
+      appwrite:
+      runtimes:
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+      - appwrite-functions:/storage/functions:rw
+      - /tmp:/tmp:rw
```
lohanidamodar, line 169: do we need this?

app/views/install/compose.phtml
```diff
@@ -278,10 +324,40 @@ services:
...
+      - _APP_DB_HOST
+      - _APP_DB_PORT
+      - _APP_DB_SCHEMA
+      - _APP_DB_USER
+      - _APP_DB_PASS
+      - _APP_FUNCTIONS_TIMEOUT
+      - _APP_EXECUTOR_SECRET
+      - _APP_USAGE_STATS
+
+  appwrite-executor:
```
lohanidamodar, line 343: why do I see two `appwrite-executor`?

app/tasks/usage.php
```diff
@@ -370,7 +370,7 @@
 
                         // Get total storage
                         $dbForProject->setNamespace('_project_' . $projectId);
-                        $storageTotal = $dbForProject->sum('files', 'sizeOriginal') + $dbForProject->sum('tags', 'size');
+                        $storageTotal = $dbForProject->sum('files', 'sizeOriginal') + $dbForProject->sum('deployments', 'size');
```
lohanidamodar, line 373: We also need to add build sizes as we are also keeping the built code

app/workers/deletes.php
```diff
@@ -309,15 +309,15 @@ protected function deleteFunction(Document $document, string $projectId): void
         $dbForProject = $this->getProjectDB($projectId);
         $device = new Local(APP_STORAGE_FUNCTIONS . '/app-' . $projectId);
 
-        // Delete Tags
-        $this->deleteByGroup('tags', [
+        // Delete Deployments
```
lohanidamodar, line 312: We need to handle delete build

docs/references/functions/list-builds.md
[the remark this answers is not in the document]
lohanidamodar: I say we keep it

lohanidamodar (commented):
Another round of brief review

docs/references/functions/list-builds.md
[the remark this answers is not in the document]
eldadfux: Agree, but we have to document them.

Dockerfile
```diff
@@ -177,11 +177,14 @@ ENV _APP_SERVER=swoole \
...
     _APP_SMTP_USERNAME= \
     _APP_SMTP_PASSWORD= \
+    _APP_FUNCTIONS_SIZE_LIMIT=30000000 \
     _APP_FUNCTIONS_TIMEOUT=900 \
     _APP_FUNCTIONS_CONTAINERS=10 \
     _APP_FUNCTIONS_CPUS=1 \
     _APP_FUNCTIONS_MEMORY=128 \
     _APP_FUNCTIONS_MEMORY_SWAP=128 \
+    _APP_EXECUTOR_SECRET=a-random-secret \
+    _APP_EXECUTOR_RUNTIME_NETWORK=appwrite_runtimes \
```
eldadfux, line 187: In the `docker-compose.yml` file we already have a network called `runtimes`, what is the difference between them? Can we user a term that will help differentiate between the two?

Dockerfile
```diff
@@ -277,8 +280,10 @@ RUN chmod +x /usr/local/bin/doctor && \
     chmod +x /usr/local/bin/worker-database && \
     chmod +x /usr/local/bin/worker-deletes && \
     chmod +x /usr/local/bin/worker-functions && \
+    chmod +x /usr/local/bin/worker-builds && \
     chmod +x /usr/local/bin/worker-mails && \
-    chmod +x /usr/local/bin/worker-webhooks
+    chmod +x /usr/local/bin/worker-webhooks && \
+    chmod +x /usr/local/bin/executor
```
eldadfux, line 286: Let's move `executor` to the upper part of the list so it feels more organized. Also does it mean we need executable for `realtime` and `http`?

app/config/collections.php
eldadfux: +1

app/config/collections.php
```diff
@@ -1818,6 +1840,17 @@
                 'array' => false,
                 'filters' => [],
             ],
+            [
+                '$id' => 'metadata',
```
eldadfux, line 1844: What is metadata and what is it used for?

app/config/collections.php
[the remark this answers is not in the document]
eldadfux: cc @christyjacob4

src/Appwrite/Utopia/Response/Model/Func.php
```diff
@@ -118,4 +120,22 @@ public function getType():string
     {
         return Response::MODEL_FUNCTION;
     }
+
+    /**
+     * Get Collection
```
eldadfux, line 125: Docblock is wrong here.

eldadfux: It would also be great to document what this filter is trying to do. (Fixing PHP default value to be {} instead of [])

src/Appwrite/Utopia/Response/Model/SyncExecution.php
```diff
@@ -0,0 +1,53 @@
+<?php
```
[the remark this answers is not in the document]
eldadfux: Ping

src/Executor/Executor.php
```diff
@@ -0,0 +1,314 @@
...
+        
+        $status = $response['headers']['status-code'];
+        if ($status >= 400) {
+            throw new \Exception($response['body']['message'], $status);
+        }
+
+        return $response['body'];
+    }
+
+    public function createExecution(
```
eldadfux, line 99: Docblock is missing. Extra description will important here.

src/Executor/Executor.php
```diff
@@ -0,0 +1,314 @@
...
+
+        $status = $response['headers']['status-code'];
+        if ($status >= 400) {
+            throw new \Exception($response['body']['message'], $status);
+        } 
+
+        return $response['body'];
+    }
+
+    public function deleteRuntime(string $projectId, string $functionId, string $deploymentId)
```
eldadfux, line 78: A lot of docblocks are missing in this class. Let's keep consistency.

tests/resources/docker/docker-compose.yml
```diff
@@ -266,6 +266,34 @@ services:
       - _APP_REDIS_PORT
       - _APP_SMTP_HOST
       - _APP_SMTP_PORT
+  
+  appwrite-worker-builds:
```
eldadfux, line 270: We don't really need this here. This file is only used to test our docker compose parser class.

tests/resources/functions/node/index.js
```diff
@@ -0,0 +1,29 @@
...
+        'payload' - object with request body data
+        'env' - object with environment variables
+    'res' variable has:
+        'send(text, status)' - function to return text response. Status code defaults to 200
+        'json(obj, status)' - function to return JSON response. Status code defaults to 200
+    
+    If an error is thrown, a response with code 500 will be returned.
+*/
+
+module.exports = async (req, res) => {
```
eldadfux, line 13: Would we like to test functions that also have dependencies?

src/Appwrite/Utopia/Response/Model/Build.php
```diff
@@ -0,0 +1,88 @@
...
+                'default' => 0,
+                'example' => 1592981250,
+            ])
+            ->addRule('deploymentId', [
+                'type' => self::TYPE_STRING,
+                'description' => 'The deployment that created this build',
+                'default' => '',
+                'example' => '5e5ea5c16897e',
+            ])
+            // Build Status
```
eldadfux, line 31: If we need this docblock for maintainers, we should definitely have it for devs. `description` is automatically used in the generated model docs.

docker-compose.yml
```diff
@@ -364,18 +392,51 @@ services:
...
       - _APP_FUNCTIONS_CPUS
       - _APP_FUNCTIONS_MEMORY
       - _APP_FUNCTIONS_MEMORY_SWAP
-      - _APP_USAGE_STATS
-      - _APP_STATSD_HOST
-      - _APP_STATSD_PORT
-      - DOCKERHUB_PULL_USERNAME
-      - DOCKERHUB_PULL_PASSWORD
+      - _APP_EXECUTOR_SECRET
+      - _APP_EXECUTOR_RUNTIME_NETWORK
```
eldadfux, line 434: Wan't most of this settings need to use `OPENRUNTIMES` as a prefix at some point?

app/controllers/api/projects.php
```diff
@@ -601,7 +601,7 @@
             throw new Exception('Project not found', 404, Exception::PROJECT_NOT_FOUND);
         }
 
-        $security = ($security === '1' || $security === 'true' || $security === 1 || $security === true);
+        $security = (bool) filter_var($security, FILTER_VALIDATE_BOOLEAN);
```
eldadfux, line 604: This is not the only place we need to update. Search for `$search =` on all files.

app/controllers/shared/api.php
```diff
@@ -76,8 +76,8 @@
             ;
         }
 
-        if (($abuse->check() // Route is rate-limited
-        && App::getEnv('_APP_OPTIONS_ABUSE', 'enabled') !== 'disabled') // Abuse is not disabled
+        if ((App::getEnv('_APP_OPTIONS_ABUSE', 'enabled') !== 'disabled' // Route is rate-limited
```
eldadfux, line 79: Why did we change this condition?

app/views/install/compose.phtml
eldadfux: Yes, but we have a future task for refactoring all the functions mounts names.

app/views/install/compose.phtml
```diff
@@ -153,6 +156,51 @@ services:
...
+      - _APP_OPENSSL_KEY_V1
+      - _APP_REDIS_HOST
+      - _APP_REDIS_PORT
+      - _APP_REDIS_USER
+      - _APP_REDIS_PASS
+      - _APP_DB_HOST
+      - _APP_DB_PORT
+      - _APP_DB_SCHEMA
+      - _APP_DB_USER
+      - _APP_DB_PASS
```
eldadfux, line 186: This block is not synced with the main compose file and uses a lot of un-required env vars.

eldadfux (changes requested):
@lohanidamodar it would be great if you could confirm important build, functions, execution stats are being reported as expected.

app/workers/builds.php
```diff
@@ -0,0 +1,169 @@
...
+                vars: $vars, 
+                runtime: $key, 
+                baseImage: $baseImage,
+                workdir: '/usr/code',
+                remove: true,
+                commands: [
+                    'sh', '-c',
+                    'tar -zxf /tmp/code.tar.gz -C /usr/code && \
+                    cd /usr/local/src/ && ./build.sh'
+                ]
```
lohanidamodar, line 130: Why are we not sending entrypoint here? createRuntime requires entrypoint which is used by runtimes.

src/Executor/Executor.php
```diff
@@ -0,0 +1,314 @@
...
+        'content-type' => '',
+    ];
+
+    public function __construct(string $endpoint = 'http://appwrite-executor/v1')
+    { 
+        $this->endpoint = $endpoint;
+    }
+
+    public function createRuntime(
+        string $functionId,
```
lohanidamodar, line 35: function id is never used?

lohanidamodar (commented):
Initial review. Also missing entrypoint is causing errors in dart runtime

Dockerfile
PineappleIOnic: They are the same network, the `appwrite_` prefix is because it's apart of the appwrite docker project

app/config/collections.php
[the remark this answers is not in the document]
PineappleIOnic: sourceType is supposed to remember what storage device was used when storing the file where as source remembers the path. You can see what it will contain here:
https://github.com/appwrite/appwrite/blob/5d9a9f53d82d464b4e1730d9883af97b562764d9/app/workers/builds.php#L93-L94

src/Appwrite/Utopia/Response/Model/SyncExecution.php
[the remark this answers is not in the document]
PineappleIOnic: We could merge this into one, without much problem. This was a holdover from when I wasn't sure just how different the synchronous execution model was going to be.

app/controllers/shared/api.php
PineappleIOnic: The abuse check was still technically running the DB calls even if the _APP_OPTIONS_ABUSE environment variable was set by placing $abuse->check() after the env check we prevent that from happening

app/executor.php
```diff
@@ -0,0 +1,722 @@
...
+            $secret = \bin2hex(\random_bytes(16));
+            $vars = \array_merge($vars, [
+                'INTERNAL_RUNTIME_KEY' => $secret,
+                'INTERNAL_RUNTIME_ENTRYPOINT' => $entrypoint,
+            ]);
+            $vars = array_map(fn ($v) => strval($v), $vars);
+            $orchestration
+                ->setCpus(App::getEnv('_APP_FUNCTIONS_CPUS', 1))
+                ->setMemory(App::getEnv('_APP_FUNCTIONS_MEMORY', 256))
+                ->setSwap(App::getEnv('_APP_FUNCTIONS_MEMORY_SWAP', 256));
```
lohanidamodar, line 205: Some linux setups have issues while providing `CPU` and `SWAP` option, haven't encountered similar thing for Memory but it might as well cause issue, so there must be a way to disable this options - easiest was , if the environment variable is empty the we don't provide default limit. Recently got this on a digitalocean droplet

When a swap limit is set to something like 2000
```
Docker Error: WARNING: Your kernel does not support swap limit capabilities or the cgroup is not mounted. Memory limited without swap.
```
(this seems to be only warning but seems to cause build to fail)

When I set env for swap empty
```
Utopia\Orchestration\Orchestration::setSwap(): Argument #1 ($mb) must be of type int, string given, called in /usr/src/code/app/executor.php on line 205
```

.env
```diff
@@ -34,15 +34,20 @@ _APP_SMTP_SECURE=
...
+_APP_EXECUTOR_RUNTIME_NETWORK=appwrite_runtimes
+_APP_EXECUTOR_SECRET=a-random-secret
 _APP_MAINTENANCE_INTERVAL=86400
 _APP_MAINTENANCE_RETENTION_EXECUTION=1209600
 _APP_MAINTENANCE_RETENTION_ABUSE=86400
 _APP_MAINTENANCE_RETENTION_AUDIT=1209600
 _APP_USAGE_STATS=enabled
 _APP_LOGGING_PROVIDER=
 _APP_LOGGING_CONFIG=
+OPENRUNTIMES_INACTIVE_THRESHOLD=60
```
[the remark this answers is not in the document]
eldadfux: I agree, we can actually do something very similar to what we do here, and keep everything consistent with the Appwrite standard. Very nice decoupling.

https://github.com/appwrite/appwrite/blob/master/docker-compose.yml#L475

app/controllers/shared/api.php
TorstenDittmann: For performance, basically it will not write to the abuse table if abuse is disabled.

TorstenDittmann (approved):

PR #2663, opened 2022-01-21
Fix: pin PHP to patch version

kodumbeats:
## What does this PR do?

Pins to specific PHP and Alpine versions to avoid per-patch regressions.

## Test Plan

Needed to add a small sleep to realtime tests which is unrelated to pinned versions.

Successful compilation and existing test suite are sufficient.

## Pending questions

- [x] Should we be specific in pinning node? We can do this, and may want to since we're running build scripts, but I didn't yet because 👇🏻 
- [x] Should we be specific in pinning composer? We can only pin to composer versions, not PHP, but we aren't actually running scripts, just fetching code.

## Related PRs and Issues

https://github.com/php/php-src/issues/7978

Yes.

### DIFF
--- a/tests/e2e/Services/Realtime/RealtimeBase.php
+++ b/tests/e2e/Services/Realtime/RealtimeBase.php
@@ -66,6 +66,7 @@ trait RealtimeBase
         $this->assertEquals('error', $payload['type']);
         $this->assertEquals(1008, $payload['data']['code']);
         $this->assertEquals('Missing or unknown project ID', $payload['data']['message']);
+        \usleep(250000); // 250ms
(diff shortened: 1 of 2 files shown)


tests/e2e/Services/Realtime/RealtimeBase.php
```diff
@@ -66,6 +66,7 @@ public function testConnectionFailureUnknownProject(): void
         $this->assertEquals('error', $payload['type']);
         $this->assertEquals(1008, $payload['data']['code']);
         $this->assertEquals('Missing or unknown project ID', $payload['data']['message']);
+        \usleep(250000); // 250ms
```
TorstenDittmann, line 69: this is actually 250000 ms right?

PR #2673, opened 2022-01-23
Feat fx review

eldadfux:
## What does this PR do?

@PineappleIOnic I added a few minor fixes on top of the PR review.

* Reorganized namespaces
* Reorganized executor functions
* Closed port 8080
* Removed all function volumes (docker.sock, tmp, and functions storage)
* Removed mount: /var/run/docker.sock from dev mode compose file on the `appwrite` container 
* Removed unused env vars from the functions worker

TODO:
* Move build stage to a dedicated worker that could be reused later
* Fix executor HTTP API to be REST compatible with proper usage of HTTP CRUD methods and entities as endpoint names.
* Double check new indexes and if we miss any
* When we delete a tag we should delete all builds
* Should all v1/functions/builds be nested under /v1/functions/:id/tags/:id/builds or just under /v1/functions/:id?
* Clean up functions that have not been executed for X INTERVAL
* Have a maximum amount of runtimes that can co-exist. 

## Test Plan

N/A

## Related PRs and Issues

N/A

Yes


christyjacob4:
> When we delete a tag we should delete all builds

@eldadfux  in the current structure, a build does not store the tagID it's associated with. 

We need to add a new `tagId` attribute for the `builds` collection so we're able to find all the associated builds for a tag easier.

TorstenDittmann:
Fixed all the conflicts with the target branch, tests pass locally 👍🏻

PR #2675, opened 2022-01-23
Feat add builds worker

christyjacob4:
## What does this PR do?

In response to https://github.com/appwrite/appwrite/pull/2641
I've added a new worker called `builds` and started migrating the API calls from the `functions` controller  to the `builds` worker.  

## Test Plan

## Related PRs and Issues

https://github.com/appwrite/appwrite/pull/2641
https://github.com/appwrite/appwrite/pull/2673

(Write your answer here.)


app/workers/builds.php
```diff
@@ -0,0 +1,262 @@
...
+
+            default:
+                throw new \Exception('Invalid build type');
+                break;
+        }
+    }
+
+    protected function triggerBuildStage(string $projectId, string $buildId)
+    {
+        // TODO: What is a reasonable time to wait for a build to complete?
```
christyjacob4, line 56: What should be a reasonable timeout for this ?

We're in the builds worker here. Does it matter if this request takes 900 seconds to complete? The rest of the builds will be queued in the meantime

app/workers/builds.php
```diff
@@ -0,0 +1,262 @@
...
+            throw new \Exception("Build failed");
+        }
+    }
+
+    protected function buildTag(string $projectId, string $functionId, string $tagId) 
+    {
+        $dbForProject = $this->getProjectDB($projectId);
+        
+        // TODO: Why does it need to skip authorization?
+        $function = Authorization::skip(fn() => $dbForProject->getDocument('functions', $functionId));
```
christyjacob4, line 144: For some reason, `getDocument()` returns an empty object when I make the call without the `::skip` method. 
I need some eyes here 👀

app/workers/builds.php
```diff
@@ -0,0 +1,262 @@
...
+        // If build ID is empty, create a new build
+        if (empty($buildId)) {
+            try {
+                $buildId = $dbForProject->getId();
+                // TODO : There is no way to associate a build with a tag. So we need to add a tagId attribute to the build document
+                // TODO : What should be the read and write permissions for a build ? 
+                $dbForProject->createDocument('builds', new Document([
+                    '$id' => $buildId,
+                    '$read' => ['role:all'],
+                    '$write' => ['role:all'],
```
christyjacob4, line 174: What should the read and write permissions be for a build ?

app/workers/builds.php
PineappleIOnic: It really varies with builds from rust taking up to 20 minutes on slow hardware. Could we make it a environment variable?

app/controllers/api/functions.php
```diff
@@ -433,11 +433,8 @@
 
         // Request executor to delete deployment containers
         $ch = \curl_init();
```
TorstenDittmann, line 435: Just an idea, we have a CURL client in the e2e tests - maybe we can abastract it for this. We are doing alot of duplicate curl stuff all over the place.

app/controllers/api/functions.php
```diff
@@ -582,45 +579,18 @@
...
-
-        // Send start build reqeust to executor using /v1/deployment
-        $function = $dbForProject->getDocument('functions', $functionId);
-
-        $ch = \curl_init();
-        \curl_setopt($ch, CURLOPT_URL, "http://appwrite-executor/v1/deployment");
-        \curl_setopt($ch, CURLOPT_POST, true);
-        \curl_setopt($ch, CURLOPT_POSTFIELDS, json_encode([
+        // Enqueue a message to start the build
+        Resque::enqueue('v1-builds', 'BuildsV1', [
```
TorstenDittmann, line 583: We have constants for the in the event class right?

Meldiron (approved):

docker-compose.yml
```diff
@@ -285,6 +285,34 @@ services:
...
+      - _APP_REDIS_USER
+      - _APP_REDIS_PASS
+      - _APP_DB_HOST
+      - _APP_DB_PORT
+      - _APP_DB_SCHEMA
+      - _APP_DB_USER
+      - _APP_DB_PASS
+      - _APP_LOGGING_PROVIDER
+      - _APP_LOGGING_CONFIG
+      - _APP_EXECUTOR_SECRET
```
TorstenDittmann, line 315: This also needs to be added to the docker-compose template for installation.

app/workers/builds.php
christyjacob4: After a discussion, it was decided to have these permissions empty. Which means we the builds documents cannot be read through an API

app/controllers/api/functions.php
christyjacob4: done 👍

docker-compose.yml
christyjacob4: done 👍

app/controllers/api/functions.php
christyjacob4: Yes we need to since there are a couple of places where we're making these calls. We should probably move the Client class to a different namespace for this.

PR #2690, opened 2022-01-25
Feat cleanup collections

PineappleIOnic:
## What does this PR do?

A PR with various changes for collections:
 - Rename functionId to resourceId on Deployments
 - Add a resourceType to Deployments
 - Remove status, buildStdout and buildStderr from deployments
 - Rename dateCreated on Builds to startTime
 - Rename time on Builds to endTime
 - Added a deploymentId to Builds
 - Added duration to builds
 - Remove Vars from Build

TODO: UI Changes

## Test Plan

Tests stay the same as before

Yes, I have


app/controllers/api/functions.php
```diff
@@ -556,7 +556,8 @@
             // Remove deploy for all other deployments.
             $deployments = $dbForProject->find('deployments', [
                 new Query('deploy', Query::TYPE_EQUAL, [true]),
-                new Query('functionId', Query::TYPE_EQUAL, [$functionId])
+                new Query('resourceId', Query::TYPE_EQUAL, [$functionId]),
+                new Query('resourceType', Query::TYPE_EQUAL, ['functions'])
```
christyjacob4, line 560: Since we're also querying by `resource` type, I think we need an index for this field as well.

app/executor.php
```diff
@@ -412,23 +410,18 @@ function execute(string $trigger, string $projectId, string $executionId, string
...
-                    'ENTRYPOINT_NAME' => $deployment->getAttribute('entrypoint'),
-                    'APPWRITE_FUNCTION_ID' => $function->getId(),
-                    'APPWRITE_FUNCTION_NAME' => $function->getAttribute('name', ''),
-                    'APPWRITE_FUNCTION_RUNTIME_NAME' => $runtime['name'],
-                    'APPWRITE_FUNCTION_RUNTIME_VERSION' => $runtime['version'],
-                    'APPWRITE_FUNCTION_PROJECT_ID' => $projectId,
-                ]
+                'endTime' => 0,
+                'duration' => 0,
+                'search' => implode(' ', [$deployment->getId(), $buildId]),
```
christyjacob4, line 424: I don't think we need the `search` field since we don't use it anywhere. 

Let's remove it from the collection as well.

app/executor.php
```diff
@@ -1088,23 +1103,18 @@ function (string $functionId, Response $response, Database $dbForProject) use ($
...
-                        'ENTRYPOINT_NAME' => $deployment->getAttribute('entrypoint'),
-                        'APPWRITE_FUNCTION_ID' => $function->getId(),
-                        'APPWRITE_FUNCTION_NAME' => $function->getAttribute('name', ''),
-                        'APPWRITE_FUNCTION_RUNTIME_NAME' => $runtime['name'],
-                        'APPWRITE_FUNCTION_RUNTIME_VERSION' => $runtime['version'],
-                        'APPWRITE_FUNCTION_PROJECT_ID' => $projectID,
-                    ]
+                    'endTime' => 0,
+                    'duration' => 0,
+                    'search' => implode(' ', [$deployment->getId(), $buildId])
```
christyjacob4, line 1117: Let's remove the `search` field from here and `collections.php` as well

PR #2733, opened 2022-01-31
Feat oauth refresh tokens

Meldiron:
## What does this PR do?

🚨 BREAKING CHANGES 🚨
`providerToken` in database is now `providerAccessToken`. Same in response model
🚨🚨🚨🚨🚨🚨

This PR stores the expiry token and expiryDate on a session if OAuth authentication is used. This is really powerful for developers since access tokens are short-lived and would require users to re-login every 4 hours.

TODO: This also creates `PATCH /v1/account/sessions/{sessionId}/oauth-tokens` endpoint to refresh token after expiration.

## Test Plan

I added a test for Mock OAuth2 covering new`updateSession` endpoint.

✅


app/config/collections.php
```diff
@@ -1158,7 +1158,18 @@
...
+                'format' => '',
+                'size' => 16384,
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => ['encrypt'],
+            ],
+            [
+                '$id' => 'providerRefreshToken',
```
eldadfux, line 1172: Let's add `providerAccessTokenExpiry` with a unix timestamp as the value.

src/Appwrite/Auth/OAuth2.php
```diff
@@ -62,9 +62,9 @@ abstract public function getLoginURL():string;
     /**
      * @param string $code
      *
-     * @return string
+     * @return array
      */
-    abstract public function getAccessToken(string $code):string;
+    abstract public function getTokens(string $code):array;
```
eldadfux, line 67: A better developer friendly signature might be to have both `getAccessToken` and `getRefreshToken` methods where each returns a string. We can store the result of the first API call and cache it in `private $tokens` if `$this->tokens` is not empty we can skip a second API call and just return the relevant string according to the method needs.

We follow a similar pattern for `getUserEmail`, `getUserID`, and `getUserName`.

eldadfux (commented):
TODO:

- [ ] Add access token expiry
- [ ] Add endpoint for refreshing the endpoint, what should it look like? Maybe PATCH request for the given session?

src/Appwrite/Auth/OAuth2.php
Meldiron: User's actually have `getUser` that gets the data, and functions `getUserEmail` `getUserName`to return specific ones.

To follow that, we would have `getTokens`, store the response in variable, and then return specific parts with functions `getAccessToken` and `getRefreshToken`.

Hope that is what you meant. I have implemented in Sporify, please let me know what you think, and I can add it elsewhere: https://github.com/appwrite/appwrite/pull/2733/files#diff-00ac4df20657d38a00b1d590988c2daee3ba5476f65b9b915ef3851cbc453deb

(getRefreshToken, getTokenExpiry and getAccessToken will be in abstract class to prevent repetition)

eldadfux: Yes, this example look good, but `getTokens` can be a private method.

src/Appwrite/Utopia/Response/Model/Session.php
```diff
@@ -40,9 +40,21 @@ public function __construct()
...
+            ->addRule('providerAccessToken', [
                 'type' => self::TYPE_STRING,
-                'description' => 'Session Provider Token.',
+                'description' => 'Session Provider AccessToken.',
+                'default' => '',
+                'example' => 'MTQ0NjJkZmQ5OTM2NDE1ZTZjNGZmZjI3',
+            ])
+            ->addRule('providerAccessTokenExpiry', [
+                'type' => self::TYPE_INTEGER,
+                'description' => 'Date, the user has accepted the invitation to join the team in Unix timestamp.',
```
brandonroberts, line 51:
```suggestion
                'description' => 'Date, the Unix timestamp of when the access token expires',
```

src/Appwrite/Auth/OAuth2/Amazon.php
```diff
@@ -59,31 +64,27 @@ public function getLoginURL(): string
...
-            ])
-        );
-        
-        $accessToken = \json_decode($accessToken, true);
-
-        if (isset($accessToken['access_token'])) {
-            return $accessToken['access_token'];
+        if(empty($this->tokens)) {
+            $headers = ['Content-Type: application/x-www-form-urlencoded;charset=UTF-8'];
+            $this->tokens = \json_decode($this->request(
```
brandonroberts, line 73: Just so I understand correctly, this stores the whole response from the access token request right? But we only provide ways to get the access token and refresh token publicly?

How does this handle nested JSON objects?

src/Appwrite/Auth/OAuth2/Notion.php
```diff
@@ -51,32 +56,25 @@ public function getLoginURL():string
...
-                'code' => $code
-            ])
-        );
-
-        $response = \json_decode($response, true);
-
-        if (isset($response['access_token'])) {
-            return $response['access_token'];
+        if(empty($this->tokens)) {
+            $headers = ['Authorization: Basic ' . \base64_encode($this->appID . ':' . $this->appSecret)];
```
brandonroberts, line 64: Should building the `Authorization Basic` header become a function for reuse?

brandonroberts (changes requested):

Meldiron:
- Added tokenExpiry information

CleanShot 2022-02-01 at 16 34 28

- Implemented `sdk.account.updateOAuth2Tokens("current", true)`

Code:

```
   async function flow() {
    var { providerAccessTokenExpiry: e, providerAccessToken: t } =
     await sdk.account.getSession("current");
    console.log("Current token ", t, " , expires at ", e);

    var { providerAccessTokenExpiry: e, providerAccessToken: t } =
     await sdk.account.updateOAuth2Tokens("current");
    console.log("Updated token to ", t, " , expires at ", e);

    var { providerAccessTokenExpiry: e, providerAccessToken: t } =
     await sdk.account.getSession("current");
    console.log("Current token ", t, " , expires at ", e);

    var { providerAccessTokenExpiry: e, providerAccessToken: t } =
     await sdk.account.updateOAuth2Tokens("current", true);
    console.log("FORCE Updated token to ", t, " , expires at ", e);

    var { providerAccessTokenExpiry: e, providerAccessToken: t } =
     await sdk.account.getSession("current");
    console.log("Current token ", t, " , expires at ", e);
   }
```

Output:
CleanShot 2022-02-01 at 17 36 29

(first-time token was not re-generated and responded with empty 204, because it did not expire yet. Second time I forced it with `true`, so new token was provided)

app/config/collections.php
Meldiron: Done.

src/Appwrite/Auth/OAuth2.php
Meldiron: Changed it to 'protected'. I added some universal functions to OAuth class, and it now needs to access this `getTokens` method.

src/Appwrite/Auth/OAuth2/Notion.php
Meldiron: It could, for this PR I re-used what was already there. We could make new PR with these improvements, I also saw a lo of potential for re-use. 90% of providers follow protocol strictly and only differs in path.

src/Appwrite/Auth/OAuth2/Amazon.php
Meldiron: Yes, methods `getRefreshToken` and `getAccessToken` are on parent - on OAuth2.php. I decided to put it there because all of our current providers follow the same response structure. If there is some provider that does not, the function could be overwritten.

Meldiron:
> TODO:
> 
> * [ ]  Add access token expiry
> * [ ]  Add endpoint for refreshing the endpoint, what should it look like? Maybe PATCH request for the given session?

Both addressed

brandonroberts (approved):

app/controllers/api/account.php
```diff
@@ -1647,6 +1654,99 @@
...
+        /** @var Appwrite\Event\Event $events */
+        /** @var Appwrite\Stats\Stats $usage */
+
+        $sessionId = ($sessionId === 'current')
+            ? Auth::sessionVerify($user->getAttribute('sessions'), Auth::$secret)
+            : $sessionId;
+
+        $sessions = $user->getAttribute('sessions', []);
+
+        foreach ($sessions as $key => $session) {/** @var Document $session */
```
eldadfux, line 1699: I can definitely see your problem with this. We could try and optimize the code, but could also consider deprecating the sessions attribute and just use a subquery like we do for the project sub-entities.

app/controllers/api/account.php
```diff
@@ -1647,6 +1654,99 @@
         throw new Exception('Session not found', 404);
     });
 
+App::patch('/v1/account/sessions/:sessionId/oauth2-tokens')
```
eldadfux, line 1657: I'm not sure this path makes total sense. I'd love us to come with other REST complaint ideas.

src/Appwrite/Auth/OAuth2.php
```diff
@@ -62,9 +62,16 @@ abstract public function getLoginURL():string;
...
+     */
+    abstract protected function getTokens(string $code):array;
+
+    /**
+     * @param string $refreshToken
+     *
+     * @return array
      */
-    abstract public function getAccessToken(string $code):string;
+    abstract public function refreshTokens(string $refreshToken):array;
```
eldadfux, line 74: `refreshTokens` or `refreshToken`?

src/Appwrite/Auth/OAuth2.php
```diff
@@ -109,6 +116,38 @@ protected function getScopes():array
...
 
+    /**
+     * @param string $code
+     *
+     * @return string
+     */
+    public function getAccessToken(string $code):string
+    {
+        $tokens = $this->getTokens($code);
+        return $tokens['access_token'];
```
eldadfux, line 127: We could use null safety here to make the code a bit more solid in case the response hasn't came in as expected.

src/Appwrite/Auth/OAuth2.php
```diff
@@ -109,6 +116,38 @@ protected function getScopes():array
...
+
+    /**
+     * @param string $code
+     *
+     * @return string
+     */
+    public function getRefreshToken(string $code):string
+    {
+        $tokens = $this->getTokens($code);
+        return $tokens['refresh_token'];
```
eldadfux, line 138: We could use null safety here to make the code a bit more solid in case the response hasn't came in as expected.

src/Appwrite/Auth/OAuth2.php
```diff
@@ -109,6 +116,38 @@ protected function getScopes():array
...
+
+    /**
+     * @param string $code
+     *
+     * @return string
+     */
+    public function getAccessTokenExpiry(string $code):string
+    {
+        $tokens = $this->getTokens($code);
+        return $tokens['expires_in'];
```
eldadfux, line 149: We could use null safety here to make the code a bit more solid in case the response hasn't came in as expected.

src/Appwrite/Utopia/Response/Model/Session.php
```diff
@@ -40,9 +40,21 @@ public function __construct()
...
+            ->addRule('providerAccessToken', [
                 'type' => self::TYPE_STRING,
-                'description' => 'Session Provider Token.',
+                'description' => 'Session Provider Access Token.',
+                'default' => '',
+                'example' => 'MTQ0NjJkZmQ5OTM2NDE1ZTZjNGZmZjI3',
+            ])
+            ->addRule('providerAccessTokenExpiry', [
+                'type' => self::TYPE_INTEGER,
+                'description' => 'Date, the Unix timestamp of when the access token expires',
```
eldadfux, line 51:
```suggestion
                'description' => 'Date, the Unix timestamp of when the access token expires.',
```

app/controllers/api/account.php
```diff
@@ -1647,6 +1654,99 @@
...
+            ? Auth::sessionVerify($user->getAttribute('sessions'), Auth::$secret)
+            : $sessionId;
+
+        $sessions = $user->getAttribute('sessions', []);
+
+        foreach ($sessions as $key => $session) {/** @var Document $session */
+            if ($sessionId == $session->getId()) {
+                $expireAt = (int) $session->getAttribute('providerAccessTokenExpiry');
+                if(\time() < $expireAt - 5 && !$force) { // 5 seconds time-sync and networking gap, to be safe
+                    return $response->noContent();
```
eldadfux, line 1703: Not sure this is the right response here. We usually use no-content only on delete operations.

app/controllers/api/account.php
```diff
@@ -1647,6 +1654,99 @@
...
+                }
+
+                $provider = $session->getAttribute('provider');
+                $refreshToken = $session->getAttribute('providerRefreshToken');
+
+                $appId = $project->getAttribute('providers', [])[$provider.'Appid'] ?? '';
+                $appSecret = $project->getAttribute('providers', [])[$provider.'Secret'] ?? '{}';
+
+                $className = 'Appwrite\\Auth\\OAuth2\\'.\ucfirst($provider);
+                $oauth2 = new $className($appId, $appSecret, '', [], []);
```
eldadfux, line 1713: We're missing a check to see if provider is actually supported.

```php
        $className = 'Appwrite\\Auth\\OAuth2\\' . \ucfirst($provider);
        if (!\class_exists($className)) {
            throw new Exception('Provider is not supported', 501);
        }
```

eldadfux (changes requested):

app/controllers/api/account.php
```diff
@@ -1647,6 +1654,103 @@
         throw new Exception('Session not found', 404);
     });
 
+App::patch('/v1/account/sessions/:sessionId')
+    ->desc('Update session - refresh OAuth2 access token')
```
eldadfux, line 1658:
```suggestion
    ->desc('Update Session (Refresh Tokens)')
```

eldadfux (changes requested):

app/controllers/api/account.php
Meldiron: Changed to `PATCH /v1/account/sessions/:sessionId`, and we can use same endpoint for session renewing in the future.

src/Appwrite/Auth/OAuth2.php
Meldiron: Tokens. Auth0 says `With refresh token rotation enabled, every time an application exchanges a refresh token to get a new access token, a new refresh token is also returned `.

This means refresh token can also change when the access token is renewed, so one or two tokens could be changing, depending on specific OAuth provider.

src/Appwrite/Auth/OAuth2.php
Meldiron: Adde null safety

src/Appwrite/Auth/OAuth2.php
Meldiron: Adde null safety

app/controllers/api/account.php
Meldiron: Added null safety

src/Appwrite/Auth/OAuth2.php
Meldiron: Added null safety

app/controllers/api/account.php
Meldiron: I updated this part by removing the force parameter. If a developer decides to hit this endpoint, we always renew the token. They can check expiryDate from session to prevent firing it too often. Worst-case scenario, their app ID/secret hits the API limit. Not a concern for Appwrite Cloud.

app/controllers/api/account.php
Meldiron: I checked the rest of the code and I believe we currently do a lot of duplicate work. Every write is done on both 'sessions' and 'user'. We 100% should have subQuery on sessions, but that would need to be a separate refactor.

With that said, I would stick with this solution that can get slower with many active sessions, and fix it alongside other endpoints once we introduce subQuery for sessions.

app/controllers/api/account.php
Meldiron: Great catch, thanks.

app/controllers/api/account.php
eldadfux: Agree, and it will be better to address it in a new PR.

eldadfux (approved):

Issue #2710, opened 2022-01-28
🚀 Feature: Adding Hash Option to Server API


tannermeade:
### 🔖 Feature description

Add an option in the server API to choose a hashing format of the password. For example plain or bcrypt. Plain is default (string converted to bcrypt). If its already in bcrypt we can skip the hashing and just store the string.

Comes from a discord discussion between Eldadfux and Tanner: https://discord.com/channels/564160730845151244/564161855862603789/936504708976234547

![api](https://user-images.githubusercontent.com/46303912/151499817-650dc48c-f4f9-457d-9d0b-e8903f9ab2bf.png)

### 🎤 Pitch

The use case here is migrating data into Appwrite. If someone has user information that they want to import, there currently is no way to store the hashed passwords in the database without circumventing Appwrite.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

tannermeade:
This plays into the purposes of this feature as well: https://github.com/appwrite/appwrite/issues/2709

eldadfux:
Rethinking this issue, instead of `format`, I think `hash` might be more descriptive. We should also research what popular hashing algos we should support. @tannermeade have any suggestions?

Meldiron:
> Rethinking this issue, instead of `format`, I think `hash` might be more descriptive. We should also research what popular hashing algos we should support. @tannermeade have any suggestions?

MD5 is really common for older applications. It is not considered secure, but if the point is to support importing from existing services, we could consider supporting MD5 too.

eldadfux:
Yeh, I think MD5 should be good as long as we don't offer it as a choice for new users, but only for importing. Maybe we can also warn in the UI, that the user hashing algo is outdated.

tannermeade:
@Meldiron Firebase uses a modified version of scrypt that AES encrypts the hash. This pull request will be implementing the unmodified version of scrypt (Colin Percival) correct?

If so, that'll work well. When importing hashes, the firebase hashes will need to be unencrypted outside of AES before importing, however that is doable.

[...]

Issue #2709, opened 2022-01-28
🚀 Feature: Adding scrypt algo


tannermeade:
### 🔖 Feature description

Add support for the scrypt algo.

Comes from a discord conversation between Tanner and Eldadfux: https://discord.com/channels/564160730845151244/564161855862603789/936507537736491091

### 🎤 Pitch

Appwrite currently uses bcrypt because it was the "best" algo at the time. There is another algo that is incrementally better called script. While there can be a discussion about if scrypt is needed or not, what is important is that Firebase and other systems use scrypt. So if anyone from Firebase wants to start using Appwrite, the scrypt hashes are not supported right now.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

tannermeade:
@Meldiron Firebase uses a modified version of scrypt that AES encrypts the hash. This pull request will be implementing the unmodified version of scrypt (Colin Percival) correct?

If so, that'll work well. When importing hashes, the firebase hashes will need to be unencrypted outside of AES before importing, however that is doable.

Meldiron:
I can proudly share that I have tests passing in my PR regarding MD5 (older systems), BCrypt (many systems), PHPass (WordPress), SCrypt (a bit safer one, not sure who uses it), SCryptModified (Google's fork of scrypt)

TorstenDittmann:
This got implemented and released with Appwrite 1.0.0 🥳 

Thank you for your contribution 🙏🏻  You can read more [here](https://appwrite.io/1.0).

PR #2747, opened 2022-02-04
Password Hashing algos

Meldiron:
## What does this PR do?

Appwrite uses bcrypt encryption. This PR adds support for more algorithms. This PR will make it possible to use multiple algorithms under the hood, making user migration into Appwrite way more accessible.

Hashing algo switched to Argon2. Looks to be more secure than BCrypt, so let's be the safest baas 😎 Thanks to this PR, projects using old algo will still work.

## Test Plan

All current tests should pass. New tests implemented.

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/2709
- https://github.com/appwrite/appwrite/issues/2710

✅


app/controllers/api/account.php
```diff
@@ -49,20 +49,26 @@
     ->param('email', '', new Email(), 'User email.')
     ->param('password', '', new Password(), 'User password. Must be at least 8 chars.')
     ->param('name', '', new Text(128), 'User name. Max length: 128 chars.', true)
+    ->param('hash', 'bcrypt', new WhiteList(['bcrypt', 'scrypt', 'md5']), 'Hashing algorithm for password. The default value is bcrypt.', true)
+    ->param('import', false, new Boolean(), 'Are you importing hashed password?', true)
```
eldadfux, line 53:
```suggestion
    ->param('hash', 'bcrypt', new WhiteList(['bcrypt', 'scrypt', 'md5']), 'Hashing algorithm for password. The default value is bcrypt.', true)
    ->param('import', false, new Boolean(), 'Are you importing hashed password?', true)
```

My original idea was for the `hash` param to indicate what kind of string was passed as the password. `plain` should have been the default type, meaning the password was passed with no hashing. In this case we should hash it with our default algorithm.

Using this implementation we don't need the `import` param, and we don't need to change the default behavior of the API.

We can also use Auth::HASH_BCRYPT and other constants for supported algorithms.

app/controllers/api/account.php
```diff
@@ -499,7 +506,7 @@
                         'email' => $email,
                         'emailVerification' => true,
                         'status' => true, // Email should already be authenticated by OAuth2 provider
-                        'password' => Auth::passwordHash(Auth::passwordGenerator()),
+                        'password' => Auth::passwordHash(Auth::passwordGenerator(), 'bcrypt'),
```
eldadfux, line 509: Shouldn't we use `scrypt` as the new default?

app/controllers/api/account.php
eldadfux: Also, changes should apply only to the users API, for server side import of users.

eldadfux (changes requested):

app/controllers/api/account.php
Meldiron: We have been using `bcrypt` until now, so to prevent hell during migration, I would stick with current as default.

tannermeade:
@Meldiron Firebase uses a modified version of scrypt that AES encrypts the hash. This pull request will be implementing the unmodified version of scrypt (Colin Percival) correct?

If so, that'll work well. When importing hashes, the firebase hashes will need to be unencrypted outside of AES before importing, however that is doable.

tannermeade:
@Meldiron Firebase uses a modified version of scrypt that AES encrypts the hash. This pull request will be implementing the unmodified version of scrypt (Colin Percival) correct?

If so, that'll work well. When importing hashes, the firebase hashes will need to be unencrypted outside of AES before importing, however that is doable.

app/controllers/api/account.php
eldadfux: I think we should support the latest algorithm, we should add a hash attribute to each user with bcrypt for every user. It's not much different from things we have done before on migration. After migration each new user created will use the best and latest algorithm.

app/controllers/api/account.php
christyjacob4: We also need to add plaintext option

Meldiron:
I have done research about hashing to full understand the problem.

[BCrypt](https://en.wikipedia.org/wiki/Bcrypt)

- Cost. Appwrite uses 8, others seems to use 10
- Version (A or B, X, Y), B used widely
- Salt can be provided, instead of random generation

[MD5](https://en.wikipedia.org/wiki/MD5)

- No configuration. Not secure, should not be allowed for registration (only for import)

[SCrypt](https://en.wikipedia.org/wiki/Scrypt)

- Complex as hell, more research needed

app/controllers/api/account.php
Meldiron: From the feedback I got, the safest algo at the moment seems to be `Argon2`, I wrote an implementation for it (we should support it anyway for importing purposes).

Please let me know and we can start using it as the default hashing algo.

app/config/collections.php
```diff
@@ -990,8 +991,30 @@
                 'required' => false,
                 'default' => null,
                 'array' => false,
+                'filters' => ['encrypt'],
```
Meldiron, line 994: Is this useful addition? Does it serve any purpose, or just make things slower?

Meldiron: It is useful, yet. If people upload MD5 hashed password, we should at least encrypt it to protect it as much as possible.

Regarding MD5, there is also the logic to re-hash the password as soon as the user logs in. That makes the timeframe when we have MD5 passwords as small as possible.

src/Appwrite/Auth/Auth.php
```diff
@@ -2,11 +2,31 @@
...
+use Appwrite\Auth\Hash\SCrypt;
+use Appwrite\Auth\Hash\SCryptModified;
 use Utopia\Database\Document;
 use Utopia\Database\Validator\Authorization;
 
 class Auth
 {
+
+    const SUPPORTED_ALGOS = [
+        'argon2' => 'Argon2',
```
eldadfux, line 18: Can we use camelCase or all lower?

Meldiron: We use all lower-case for user-facing stuff, and we follow the same casing as algo regarding class names internally.

Do you want me to switch to camelCase for internal classes?

eldadfux (changes requested):
Added a few comments. 

One more concern that I have is about the hash options in the API signature. I feel it is hard to explain how to use it. We have a similar problem on OAuth options in the projects controller. I had a plan to solve it by adding a dedicated HTTP endpoint for each provider with an API signature that tries to stick to primitive types and with that approach creates a simpler documentation. Would love your feedback.

src/Appwrite/Auth/Auth.php
eldadfux: iirc correctly we stick with camelCase for everything, if not, please point me to the right direction so we can review if it had any reasoning behind it.

src/Appwrite/Auth/Hash/SHA.php
```diff
@@ -0,0 +1,47 @@
...
+        return $this->hash($password) === $hash;
+    }
+    
+    /**
+     * Get default options for specific hashing algo
+     * 
+     * @return mixed options named array
+     */
+    public function getDefaultOptions(): mixed {
+        return [ 'type' => 'sha3-512' ];
```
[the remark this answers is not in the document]
Meldiron: These are defaults. Yes, it is configurable by `hashOptions` API parameter.

src/Appwrite/Auth/Auth.php
Meldiron: I was referring to `/app/config/provieres.php`. I can see API-facing keys are lower-cased, and `name` attribute is the same case as the company's official name. But then I realized that PHP class is actually just uppercased first letter.

I updated this PR to follow sane conventions - lowercase API-facing, first-upper-case class names.

Please review again, does this look good now?

src/Appwrite/Utopia/Response/Model/Account.php
```diff
@@ -0,0 +1,38 @@
+<?php
```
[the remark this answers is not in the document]
Meldiron: Regarding tokens, it is done on controller level: https://github.com/appwrite/appwrite/blob/master/app/controllers/api/account.php#L741-L744

I don't think I like this approach. If we ever introduce new endpoints, it's our duty to remember to hide it from the response. With separate models (extending each other, and hiding stuff), the security layer for NOT exposing some keys is done for us. To me, this is as important as doing validation of input before doing an action.

Do you agree? If not, I can update PR to do the same as we do regarding tokens.

src/Appwrite/Utopia/Response/Model.php
```diff
@@ -87,6 +87,22 @@ protected function addRule(string $key, array $options): self
...
     }
 
+    /**
+     * Delete an existing Rule
+     * If rule exists, it will be removed
+     *
+     * @param string $key
+     * @param array $options
+     */
+    protected function removeRule(string $key): self
```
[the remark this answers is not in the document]
Meldiron: It's an abstract class, how would we implement tests? Do we test each response model separately? Or do we create some `Test.php` model?

src/Appwrite/Auth/Hash/BCrypt.php
```diff
@@ -0,0 +1,43 @@
...
+        return \password_verify($password, $hash);
+    }
+
+    /**
+     * Get default options for specific hashing algo
+     * 
+     * @return mixed options named array
+     */
+    public function getDefaultOptions(): mixed {
+        return [ 'cost' => 8 ];
```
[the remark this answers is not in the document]
Meldiron: These are defaults. Yes, it is configurable by `hashOptions` API parameter.a

src/Appwrite/Utopia/Response/Model/User.php
```diff
@@ -23,6 +23,24 @@ public function __construct()
...
+                'default' => '',
+                'example' => '$argon2id$v=19$m=2048,t=4,p=3$aUZjLnliVWRINmFNTWMudg$5S+x+7uA31xFnrHFT47yFwcJeaP0w92L/4LdgrVRXxE',
+            ])
+            ->addRule('hash', [
+                'type' => self::TYPE_STRING,
+                'description' => 'Password hashing algorithm.',
+                'default' => '',
+                'example' => 'bcrypt',
+            ])
+            ->addRule('hashOptions', [
```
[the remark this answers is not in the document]
Meldiron: > In the demo, the empty default was [] instead of {}.
I think this is a bug of utopia/database. Should I look at it as part of this feature?

> I think it would make more sense to have relevant hash models added
Agree, great suggestion.

> I guess we only need one for argo and modified script?
All algos except PHPass and MD5 has some options. But we usually don't need them. We only need them for SCryptModified. For the others, even tho we don't need them, they will be valuable when exporting users from Appwrite, so their new backend can use exact same options to keep exact same level of security.

I will implement Hash models. Regarding PHPass and MD5, should we have a model too? That would just be empty - no rules? IMO it would be a great indicator (in docs) that no options go there. Otherwise, people would simply not find a model for their algo and would be confused.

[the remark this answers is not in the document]
Meldiron: > I think this is a bug of utopia/database. Should I look at it as part of this feature?

Note: Look at how we do it in preferences, should be a solution. Maybe even vars attribute in webhook/function.

src/Appwrite/Utopia/Response/Model.php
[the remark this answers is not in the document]
Meldiron: Note: List all places where we miss unit tests (inside Appwrite)

src/Appwrite/Utopia/Response/Model/User.php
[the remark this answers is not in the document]
Meldiron: This should not be resolved due to more refactoring.

src/Appwrite/Utopia/Response/Model.php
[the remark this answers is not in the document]
Meldiron: Can we make it a separate task? I think it is out of the scope of this PR.

Dockerfile
```diff
@@ -33,6 +33,8 @@ ENV PHP_REDIS_VERSION=5.3.7 \
     PHP_MONGODB_VERSION=1.13.0 \
     PHP_SWOOLE_VERSION=v4.8.9 \
     PHP_IMAGICK_VERSION=3.7.0 \
+    # For SCRYPT never go v1.4.2 or lower. That doesnt work on PHP 8 + Alpine3.15/16
+    PHP_SCRYPT_VERSION=master \
```
Meldiron, line 37: Let's lock it to commit number

Dockerfile
```diff
@@ -94,6 +96,15 @@ RUN \
   ./configure && \
   make && make install
 
+## Scrypt Extension
+FROM compile AS scrypt
+RUN \
+  git clone --depth 1 --branch $PHP_SCRYPT_VERSION https://github.com/DomBlack/php-scrypt.git && \
```
eldadfux, line 102: As mentioned earlier, better use a specific commit than master.

src/Appwrite/Auth/Auth.php
```diff
@@ -130,26 +144,56 @@ public static function hash(string $string)
...
+            $options = Auth::DEFAULT_ALGO_OPTIONS;
+        }
+
+        if (!\array_key_exists($algo, Auth::SUPPORTED_ALGOS)) {
+            throw new \Exception('Hashing algorithm \'' . $algo . '\' is not supported.');
+        }
+
+        $className = Auth::SUPPORTED_ALGOS[$algo];
+        $classPath = '\\Appwrite\\Auth\\Hash\\' . $className;
+        $hasher = new $classPath($options);
```
eldadfux, line 167: Not a huge fan of this. I know we do it when we create an OAuth adapter. Would a switch case work here?
Also, having a custom, well defined signature for initing each hash would be much friendlier.

src/Appwrite/Utopia/Response/Model/Account.php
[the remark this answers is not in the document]
eldadfux: Agree.

src/Appwrite/Utopia/Response/Model/User.php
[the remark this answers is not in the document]
eldadfux: I still have some issues with this attribute. It's very hard to describe. What kind of structure can it have?

tests/unit/Auth/AuthTest.php
```diff
@@ -48,12 +48,121 @@ public function testHash()
...
     public function testPassword()
     {
-        $secret = 'secret';
-        $static = '$2y$08$PDbMtV18J1KOBI9tIYabBuyUwBrtXPGhLxCy9pWP6xkldVOKLrLKy';
-        $dynamic = Auth::passwordHash($secret);
+        /*
+        General tests, using pre-defined hashes generated by online tools
+        */
+
+        // Bcrypt - Version Y
```
eldadfux, line 55: Does this test give us good coverage for all the new "Hash" classes?

Meldiron: A combination of this and UsersAPI tests (to create&signIn) should give good enough coverage. We try defaults, as well as play around with options. There could be something platform-specific we would need to adjust to, but hashing algo, in general, should be well tested.

src/Appwrite/Utopia/Response/Model/User.php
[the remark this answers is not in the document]
Meldiron: I took what we did for attributes, and implemented it here. Please revisit and let me know if you like it.
Commit: https://github.com/appwrite/appwrite/pull/2747/commits/df365806576467bce9292b31c963b8815d86f4d4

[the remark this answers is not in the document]
Meldiron: Oh, right. I took the snake case from the PHP functions I use.

Updated in this PR: https://github.com/appwrite/appwrite/pull/2747/commits/0aaee54a63cd9ed18c27f8906050802980fcb158
We now use camelCase, just like all other places.

app/controllers/api/account.php
```diff
@@ -33,6 +33,7 @@
 use Utopia\Validator\Assoc;
 use Utopia\Validator\Range;
 use Utopia\Validator\Text;
+use Utopia\Validator\JSON;
```
lohanidamodar, line 36: Not sure this validator is used? I don't see it used in the changes.

src/Appwrite/Specification/Format/Swagger2.php
```diff
@@ -401,11 +401,20 @@ public function parse(): array
 
         foreach ($this->models as $model) {
             foreach ($model->getRules() as $rule) {
-                if (
-                    in_array($model->getType(), $usedModels)
-                    && !in_array($rule['type'], ['string', 'integer', 'boolean', 'json', 'float'])
-                ) {
-                    $usedModels[] = $rule['type'];
+                if (!in_array($model->getType(), $usedModels)) {
```
lohanidamodar, line 404: Why this change came in this PR?

lohanidamodar (changes requested):
Looks good.  Some comments and also Lints are failing. And needs a sync due to internal ID change

src/Appwrite/Auth/Hash.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace Appwrite\Auth;
+
+abstract class Hash
+{
+    /**
+     * @var mixed $options Hashing-algo specific options
+    */
+    protected mixed $options = [];
```
TorstenDittmann, line 10: is it `mixed` or and `array`?

src/Appwrite/Auth/Hash.php
```diff
@@ -0,0 +1,62 @@
...
+    {
+        $this->setOptions($options);
+    }
+
+    /**
+     * Set hashing algo options
+     *
+     * @param mixed $options Hashing-algo specific options
+    */
+    public function setOptions(mixed $options): self
```
TorstenDittmann, line 25: same here, this should be `array` 👍🏻

src/Appwrite/Auth/Hash/Md5.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Appwrite\Auth\Hash;
+
+use Appwrite\Auth\Hash;
+
+/*
+ * MD5 does not accept any options.
+ *
+ * Refference: https://www.php.net/manual/en/function.md5.php
```
TorstenDittmann, line 10:
```suggestion
 * Reference: https://www.php.net/manual/en/function.md5.php
```

src/Appwrite/Auth/Hash/Phpass.php
```diff
@@ -0,0 +1,290 @@
...
+*/
+class Phpass extends Hash
+{
+    /**
+     * Alphabet used in itoa64 conversions.
+     *
+     * @var    string
+     * @since  0.1.0
+     */
+    protected $itoa64 = './0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz';
```
TorstenDittmann, line 44: missing type

src/Appwrite/Auth/Hash/Phpass.php
```diff
@@ -0,0 +1,290 @@
...
+
+    /**
+     * Get default options for specific hashing algo
+     *
+     * @return mixed options named array
+     */
+    public function getDefaultOptions(): mixed
```
TorstenDittmann, line 51: should also be `array`

src/Appwrite/Auth/Hash/Phpass.php
```diff
@@ -0,0 +1,290 @@
...
+            $hash = crypt($password, $hash);
+        }
+
+        /**
+         * This is not constant-time.  In order to keep the code simple,
+         * for timing safety we currently rely on the salts being
+         * unpredictable, which they are at least in the non-fallback
+         * cases (that is, when we use /dev/urandom and bcrypt).
+         */
+        return $hash === $hash;
```
TorstenDittmann, line 113: Shouldnt this be `true`?

src/Appwrite/Specification/Format/Swagger2.php
Meldiron: We now introduced new (yet unsupported) concept to Swagger.
It's a definition that is only referenced in a definition, not in a path.

These changes make it work, screenshots in this PR: https://github.com/appwrite/docs/pull/202

TorstenDittmann (changes requested):
I guess once all the typings are added and formatting is done - this looks good 👍🏻

app/controllers/api/account.php
Meldiron: Leftover.

src/Appwrite/Auth/Hash/Phpass.php
Meldiron: Woah. Let me fix.. and.. write tests to prevent such a bug.

src/Appwrite/Auth/Auth.php
```diff
@@ -133,26 +154,112 @@ public static function hash(string $string)
...
+
+        if (!\in_array($algo, Auth::SUPPORTED_ALGOS)) {
+            throw new \Exception('Hashing algorithm \'' . $algo . '\' is not supported.');
+        }
+
+        switch ($algo) {
+            case 'argon2':
+                $hasher = new Argon2($options);
+                return $hasher->verify($plain, $hash);
+                break;
```
TorstenDittmann, line 235: dont need the breaks here

TorstenDittmann (approved):

Dockerfile
```diff
@@ -123,6 +123,33 @@ RUN \
...
+ENV PATH=/tmp/zig:$PATH
+RUN cargo install cargo-zigbuild
+ENV RUSTFLAGS="-C target-feature=-crt-static"
+
+FROM rust_compile as scrypt
+
+WORKDIR /usr/local/lib/php/extensions/
+
+RUN \
+  git clone --depth 1 https://github.com/PineappleIOnic/scrypt-php.git && \
```
TorstenDittmann: can we use the appwrite one?

Meldiron: Updated to Appwrite one.

Dockerfile
```diff
@@ -237,6 +264,7 @@ RUN \
   libmaxminddb-dev \
   certbot \
   docker-cli \
+  docker-compose \
```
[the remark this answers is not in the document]
Meldiron: No idea. Merge conflict resolved incorrectly, probably. Removed it.

src/Appwrite/Utopia/Response/Model/AlgoArgon2.php
```diff
@@ -0,0 +1,54 @@
...
+            ])
+            ->addRule('timeCost', [
+                'type' => self::TYPE_INTEGER,
+                'description' => 'Amount of time it may take to compute hash.',
+                'default' => '',
+                'example' => 4,
+            ])
+            ->addRule('threads', [
+                'type' => self::TYPE_INTEGER,
+                'description' => 'Number of threads to use to compute hash.',
```
eldadfux, line 28:
```suggestion
                'description' => 'Number of threads used to compute hash.',
```

src/Appwrite/Utopia/Response/Model/AlgoArgon2.php
```diff
@@ -0,0 +1,54 @@
...
+        $this
+            ->addRule('memoryCost', [
+                'type' => self::TYPE_INTEGER,
+                'description' => 'Memory that may be used to compute hash.',
+                'default' => '',
+                'example' => 65536,
+            ])
+            ->addRule('timeCost', [
+                'type' => self::TYPE_INTEGER,
+                'description' => 'Amount of time it may take to compute hash.',
```
eldadfux, line 22:
```suggestion
                'description' => 'Amount of time it took to compute hash.',
```

src/Appwrite/Utopia/Response/Model/AlgoArgon2.php
```diff
@@ -0,0 +1,54 @@
...
+
+class AlgoArgon2 extends Model
+{
+    public function __construct()
+    {
+        // No options if imported. If hashed by Appwrite, following configuration is available:
+        $this
+            ->addRule('memoryCost', [
+                'type' => self::TYPE_INTEGER,
+                'description' => 'Memory that may be used to compute hash.',
```
eldadfux, line 16:
```suggestion
                'description' => 'Memory used to compute hash.',
```

src/Appwrite/Utopia/Response/Model/AlgoScryptModified.php
```diff
@@ -0,0 +1,53 @@
...
+                'default' => '',
+                'example' => 'UxLMreBr6tYyjQ==',
+            ])
+            ->addRule('saltSeparator', [
+                'type' => self::TYPE_STRING,
+                'description' => 'Separator used to compute hash.',
+                'default' => '',
+                'example' => 'Bw==',
+            ])
+            ->addRule('signerKey', [
```
eldadfux, line 25: Should this be public?

src/Appwrite/Specification/Format/Swagger2.php
```diff
@@ -17,15 +17,32 @@ public function getName(): string
...
-            if (
-                in_array($model->getType(), $usedModels)
-                && !in_array($rule['type'], ['string', 'integer', 'boolean', 'json', 'float', 'double'])
-            ) {
-                $usedModels[] = $rule['type'];
-                foreach ($this->models as $m) {
-                    if ($m->getType() === $rule['type']) {
-                        $this->getNestedModels($m, $usedModels);
-                        return;
+            if (!in_array($model->getType(), $usedModels)) {
```
eldadfux, line 20: Don't we already have support for multiple types? why do we need this change?
Also, if required, we should probably have the same change for OpenAPI3

app/controllers/api/users.php
```diff
@@ -43,47 +86,226 @@
...
-                'name' => $name,
-                'prefs' => new \stdClass(),
-                'sessions' => null,
-                'tokens' => null,
-                'memberships' => null,
-                'search' => implode(' ', [$userId, $email, $name])
-            ]));
-        } catch (Duplicate $th) {
-            throw new Exception('Account already exists', 409, Exception::USER_ALREADY_EXISTS);
+App::post('/v1/users/import/bcrypt')
```
eldadfux, line 102: We don't need the import part. Completely fine to just have the `/algorithm`

```suggestion
App::post('/v1/users/bcrypt')
```

app/controllers/api/users.php
```diff
@@ -28,6 +28,49 @@
 use Utopia\Validator\Range;
 use Utopia\Validator\Boolean;
 use MaxMind\Db\Reader;
+use Utopia\Validator\Integer;
+
+function createUser(string $hash, mixed $hashOptions, string $userId, string $email, string $password, string $name, Database $dbForProject, Stats $usage, Event $events): Document
```
eldadfux, line 33: We should add a todo to remove this function when we move to using utopia/platform.

eldadfux (changes requested):

docs/references/users/create-argon2-user.md
```diff
@@ -0,0 +1 @@
+Create a new user. Password entered must be hashed with the the [Argon2](https://en.wikipedia.org/wiki/Argon2) algorithm. Use the [POST /users](/docs/server/users#usersCreate) endpoint to import plain text password.
```
gewenyu99, line 1: I would say `password provided` and not `entered`; we cannot assume that this is only used for client auth.

I also personally dislike the use of `import` here, because the endpoint is called create.

(Also you have `the the` here, reminder to run it through Grammarly XD)

Suggested format for all docs:
> Create a new user. Password provided must be hashed with the [Argon2](https://en.wikipedia.org/wiki/Argon2) algorithm. Use the [POST /users](/docs/server/users#usersCreate) endpoint to create users with a plain text password.

src/Appwrite/Utopia/Response/Model/AlgoArgon2.php
gewenyu99: +1

src/Appwrite/Utopia/Response/Model/AlgoArgon2.php
gewenyu99: I dislike the verb `take` here, it's used mostly informally to describe the passage of time.

> Amount of time consumed to compute hash

gewenyu99 (changes requested):
I left a few comments related to grammar :)

Super excited to see this feature merged 💯

src/Appwrite/Utopia/Response/Model/AlgoArgon2.php
eldadfux: Updated my suggestion.

app/controllers/api/users.php
```diff
@@ -43,47 +87,226 @@
     ->label('sdk.response.model', Response::MODEL_USER)
     ->param('userId', '', new CustomId(), 'User ID. Choose your own unique ID or pass the string "unique()" to auto generate it. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.')
     ->param('email', '', new Email(), 'User email.')
-    ->param('password', '', new Password(), 'User password. Must be at least 8 chars.')
+    ->param('password', '', new Password(), 'Plaintext user password. Must be at least 8 chars.')
```
eldadfux, line 90:
```suggestion
    ->param('password', '', new Password(), 'Plain text user password. Must be at least 8 chars.')
```

src/Appwrite/Specification/Format/Swagger2.php
Meldiron: Honestly, hard to tell.. I remember sitting with Torsten trying to merge my changes with his (nested models) together. This is the outcome.

I tried to remove these changes and use what is there in 0.16.x, but then specs were missing some hash types:
CleanShot 2022-08-16 at 13 34 35

The thing is, this PR creates a new edge-case when a model (definition) is not used directly in any endpoint, instead, is used by a model that is being used in an endpoint.
To be specific, there is "Get User" that returns "User" model. and on "User" there is "hashOptions" that is oneOf AlgoArgon, AlgoScrypt, AlgoBCrypt...

This change to Swagger2 adds support for oneOf/anyOf models that are not directly used in an endpoint, instead, they are mentioned in a different model.

Meldiron: You are right we need the same change in Swagger3. Added. I copy&pasted with the only difference being support for `double` in Swagger3.

eldadfux: OK, but we should probably replicate this to OPENAPI3

Meldiron: My bad, that's what I meant. I just added it to OpenAPI3.

eldadfux (approved):

PR #2755, opened 2022-02-07
Added unrestricted permissions feature

acalatrava:
## What does this PR do?

This PR will add a new "unrestricted mode" permissions level to the Documents so any user can set the permission to another user without the need to be in the same team. 

## Test Plan

The feature is disabled by default, so if you need to access this "unrestricted mode" feature you need to enable an environment variable called `_APP_FEATURE_DOCUMENTS_UNRESTRICTED`. This will enable a new menu on the Collection settings page to enable the new permission model.

For Storage, since there is not Collections attached to them, you can enable the "unrestricted mode" by specifying a prefix on your fileID. So you need to setup the prefix on the `_APP_FEATURE_STORAGE_UNRESTRICTED_PREFIX` environment variable.

## Related PRs and Issues

https://github.com/appwrite/appwrite/discussions/2689

Yep!


reyesmfabian:
Hi @eldadfux, I think this should be added to the next version, it is super necessary, also the possibility to auto validate invitations to teams without sending the invitation email, or otherwise allow access to information even if the user is not validated.

eldadfux:
We're still not 100% confident about the required changes, and it seems like the community also has some conflicting opinions. Let's keep the discussion going on #2689

PR #2756, opened 2022-02-07
fix: permissions and table names

TorstenDittmann:
## What does this PR do?

- uses internal ID for collection table names
- uses the new permissions offered by Utopia PHP
  - each collection (internal/external) has its own table for permissions
  - this allows us to have more reliability
- removes the `_project` prepend for more total length (63 chars max for table names, 36 of them are reserved for a custom id)
- migration 🥳 

Also:
- some style changes to the code
- removed most of the foreach using a reference
- added some missing typings and phpdocs

## Related PRs and Issues

https://github.com/utopia-php/database/pull/108

✅


eldadfux:
Restarted tests.

src/Appwrite/Migration/Migration.php
```diff
@@ -114,6 +131,7 @@ public function forEachDocument(callable $callback): void
                             try {
                                 $new = $this->projectDB->updateDocument($document->getCollection(), $document->getId(), $document);
                             } catch (\Throwable $th) {
+                                var_dump($th->getTraceAsString());
```
christyjacob4, line 134: I believe this is expected ? or do we remove this ?

christyjacob4 (approved):
Looks good. Are there more places where we're using `setNamespace()` ?? 🤔

TorstenDittmann:
> Looks good. Are there more places where we're using `setNamespace()` ?? 🤔

Removed the leftover var_dump 👍🏻 

There shouldn't be any places where this PR doesn't cover it. I am gonna rebase it again with 0.13.x and see if there are stuff left.

Issue #2772, opened 2022-02-10
🚀 Feature: Allow turning-off analytics


tzdev0:
### 🔖 Feature description

Adding a new environment variable / configuration option to disable all externally send analytics data.

### 🎤 Pitch

The default behavior of all Appwrite instances is to send usage data to Google Analytics to help the Appwrite team with future decisions, roadmap planning, etc. IMHO, sending this data is fine as long as no private or user-identifying data is sent.

Nevertheless, an option that allows the Appwrite user to turn off any analytics would be preferred, so that at least the possibility of an autonomous decision is available. Additionally, the current negative press surrounding Google Analytics ([Link](https://noyb.eu/en/austrian-dsb-eu-us-data-transfers-google-analytics-illegal) / [Link](https://www.wired.co.uk/article/google-analytics-europe-austria-privacy-shield)) in the EU may be holding developers back from using Appwrite, especially since there is currently no option to turn it off.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

tzdev0:
I have already prepared a pull request ( #2773 ) which would implement this feature.

Happy for any feedback!

thilojaeggi:
I would also like to have this option.

hialvaro:
Please, could we have this option in our instances?

linus1703:
I would also appreciate an option to disable all external analytics to AppWrite (https://appwrite.io/v1/analytics) and to Google. Gives me a bad taste when I change my username on my local installation and my actions beeing tracked on Google and AppWrite servers.

LickABrick:
What is the current status on this? I would also like to disable the external Google Analytics sharing.

Tilmanfs95:
I would also appreciate this because I can not use Appwrite with Google Analytics in an completely legal way in Germany.

eldadfux:
Appwrite collects data from the self-hosted console for planing, and product improvements. We already respect DO NOT TRACK headers in your browser. Following the feedback in this thread, we will also introduce a new environment variable to disable any usage tracking features in the console.

DH-555:
If I'm not wrong Appwrite now uses privacy-friendly tracking/plausible only

thyngster:
[...]

PR #2773, opened 2022-02-10
Feat 2772 allow turning off analytics

tzdev0:
## What does this PR do?
Adds a new environment variable named `_APP_OPTIONS_ANALYTICS` which allows to disable any externally sent analytics data. 

While setting the env var to the default value `enabled`, there will be no new behaviour regarding analytics. Setting the env var to `disabled`, these changes will prevent the loading of the `ga`object and will prevent adding the event handlers on the different `data-analytics` html-elements.

## Test Plan

Manual testing, by monitoring outgoing 3rd party requests from dev tools while turning the env var on and off.

## Related PRs and Issues

#2772 

Yes!


app/tasks/install.php
```diff
@@ -42,11 +43,14 @@
...
-         * This allows us to collect information without invading the privacy of our users.
-         */
-        $analytics = new GoogleAnalytics('UA-26264668-9', uniqid('server.', true));
+        $analyticsOption = App::getEnv('_APP_OPTIONS_ANALYTICS', 'enabled');
+        if($analyticsOption !== 'disabled') {
+            /**
+             * We are using a random value every execution for identification.
+             * This allows us to collect information without invading the privacy of our users.
+             */
+            $analytics = new GoogleAnalytics('UA-26264668-9', uniqid('server.', true));
```
stnguyen90, line 52: `$analytics` has an [`enabled` flag](https://github.com/utopia-php/analytics/blob/7f4f7688090d1e4e05af95c693f9a8f2d898c87f/src/Analytics/Adapter/GoogleAnalytics.php#L76) that can be used to disable sending events. That is probably a better approach than adding disabled. checks for every `$analytics->createEvent()` call.

app/views/layouts/default.phtml
```diff
@@ -1,18 +1,19 @@
...
-$locale         = $this->getParam('locale', null);
-$runtimes       = $this->getParam('runtimes', null);
+$protocol        = $this->getParam('protocol', '');
+$domain          = $this->getParam('domain', '');
+$endpoint        = $this->getParam('endpoint', '');
+$platforms       = $this->getParam('platforms', []);
+$version         = $this->getParam('version', '0.0.0');
+$isDev           = $this->getParam('isDev', false);
+$litespeed       = $this->getParam('litespeed', true);
+$analytics       = $this->getParam('analytics', 'UA-26264668-9');
```
stnguyen90, line 10: Do you happen to know what happens if the tracking ID isn't set? I wonder if it's a better approach to let this be configurable, but default to the Appwrite Core tracking ID. Ideally, if it's blank, there won't be any errors and there won't be any tracking. People who self host appwrite can set the tracking ID to their own tracking ID so they can get analytics themselves.

tzdev0: I've looked into blanking the tracking ID and it _kinda_ worked as no [pageview](https://github.com/appwrite/appwrite/blob/dcd731875cf1ddc0835e6a8ddcc421e66cbf00d4/public/scripts/views/analytics/pageview.js#L25) and no [event](https://github.com/appwrite/appwrite/blob/dcd731875cf1ddc0835e6a8ddcc421e66cbf00d4/public/scripts/views/analytics/event.js#L23) analytics got send to Google. Still, two concerns remain with this approach:
1. Google Server will be hit to download the `analytics.js` script even if no tracking ID is provided.
2. [activity analytics](https://github.com/appwrite/appwrite/blob/dcd731875cf1ddc0835e6a8ddcc421e66cbf00d4/public/scripts/views/analytics/activity.js#L19) will be send which while not being send to Google still are analytics.

To resolve the first point, we would need to exclude the GA function in the [`layouts/default.phtml`](https://github.com/appwrite/appwrite/blob/dcd731875cf1ddc0835e6a8ddcc421e66cbf00d4/app/views/layouts/default.phtml#L69) file, which in turn would force us to change the `pageview.js` and `event.js` scripts to handle the missing GA function correctly (these script would currently throw a Reference error in Appwrite's released version if no GA function is in global scope!). To resolve the second point we would need to change the `activity.js` script to prevent sending analytics to appwrite.io. Both of these solutions are kind of what I'm already doing in my implementation.

Nevertheless, I like the idea of providing a custom tracking ID and using the Appwrite ID as the default one and would implement it, if it's wanted.

> Ideally, if it's blank, there won't be any errors...

Currently, the only error thrown is due to the missing GA function. This could easily be resolved by simply not checking for the GA object if analytics is manually disabled.

app/tasks/install.php
tzdev0: To be honest, I'm quite not sure what the install task is for and therefore am not sure how to test it properly. Just adding `$analytics->enabled = false;` after creation of the `$analytics` object currently throws an error as the `enabled` property is protected.

Other than that, I am too not really happy with my implementation and would love to change that to a simple boolean solution. You got any pointers for me?

tzdev0: Just tested `$analytics->disabled()` and it works as expected. I've already pushed a new version to the PR. Thanks @stnguyen90!

Towerful:
I would really appreciate this feature

eldadfux:
The team will review it as soon as we get some spare time from the 0.13 release.

app/controllers/web/console.php
```diff
@@ -14,6 +14,7 @@
     $layout
         ->setParam('description', 'Appwrite Console allows you to easily manage, monitor, and control your entire backend API and tools.')
         ->setParam('analytics', 'UA-26264668-5')
+        ->setParam('analytics-option', App::getEnv('_APP_OPTIONS_ANALYTICS', 'enabled'))
```
lohanidamodar, line 17: `analytics-enabled` with value `true` or `false` makes more sense.

lohanidamodar (commented):
Simple suggestion, rest looks good to me.

app/controllers/web/console.php
tzdev0: Thanks for the recommendation! I will change that in the course of further suggestions.

linus1703:
Any updates on this PR?

gorootde:
Any updates on when this will be merged?

gabriellemon:
So the analytics are only for the console, not for every microservice?

DHsoftw:
That's fundamental for GDPR compliance

eric-sison:
Any updates on this PR?

christyjacob4:
Closing this PR since the console has now been moved to https://github.com/appwrite/console and we no longer use Google Analytics within the Appwrite server.

Issue #2789, opened 2022-02-14
🐛 Bug Report: Can no longer search through memberships


JakeAi:
### 👟 Reproduction steps

In 0.11, you used to be able to 
```TS
getMemberships(teamId, 'userId='+account.$id)
```

In 0.12, searching is omitted from [teams.php](https://github.com/appwrite/appwrite/blob/0.12.x/app/controllers/api/teams.php#L422)

### 👍 Expected behavior

Return memberships filtered by search

### 👎 Actual Behavior

No search filtering happened

### 🎲 Appwrite version

Version 0.12.x

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

TorstenDittmann:
Hey 👋🏻 

As of right now you can just search for the memberships using the `$id`, `email` and `name` 👍🏻 
```
getMemberships(teamId, account.$id)
getMemberships(teamId, account.email)
getMemberships(teamId, account.name)
```

JakeAi:
> 

This returns all memberships for the team. No filtering is done.

TorstenDittmann:
Oh, good catch!

`$search` is actually not used 👍🏻  gonna work on a fix.

JakeAi:
> Oh, good catch!
> 
> `$search` is actually not used 👍🏻 gonna work on a fix.

I see 0.12.3 is released, I pulled the new docker images. How does the search work now?

eldadfux:
I think this would get a fix with the 0.13 data. migration

JakeAi:
> I think this would get a fix with the 0.13 data. migration

Is there a sql command I can run to fix it so I don’t have to update my android and iOS apps?

TorstenDittmann:
We just released 0.13 and this should be fixed 👍🏻 

You can find the announcement [here](https://dev.to/appwrite/announcing-appwrite-013-with-major-upgrades-to-storage-and-functions-3hpf).

Thank you a lot for reporting this 🙏🏻 

@JakeAi sadly this wasn't a Database Issue - but also an API issue which cannot be fixed with only SQL.

JakeAi:
> We just released 0.13 and this should be fixed 👍🏻
> 
[...]

PR #2802, opened 2022-02-16
fix: team membership roles & search

TorstenDittmann:
## What does this PR do?

- adds `search` attribute to memberships
- adds fulltext index on `search` to memberships
- fixes team read permissions on migration
- fixes team update membership roles

## Test Plan

E2E tests addded

## Related PRs and Issues

#2789 

✅

### DIFF
--- a/app/controllers/api/teams.php
+++ b/app/controllers/api/teams.php
@@ -63,7 +63,9 @@ App::post('/v1/teams')
         ])));
 
         if (!$isPrivilegedUser && !$isAppUser) { // Don't add user on server mode
+            $membershipId = $dbForProject->getId();
             $membership = new Document([
+                '$id' => $membershipId,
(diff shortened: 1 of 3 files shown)


app/controllers/api/teams.php
```diff
@@ -73,6 +75,7 @@
                 'joined' => \time(),
                 'confirm' => true,
                 'secret' => '',
+                'search' => implode(' ', [$membershipId, $user->getId()])
```
eldadfux, line 78: Do we want to add the user name and email here?

TorstenDittmann: thought about it first, but we would need to maintain it when a name or email is updated in both account and user endpoints.

eldadfux: Got it, let's leave it for now

wess (approved):

PR #2807, opened 2022-02-17
feat: retry execution logic

christyjacob4:
## What does this PR do?

Improvements to the executor

Implement other endpoints - Done
Handle shutdown - Done
Get list of supported runtimes on startup - Done
Pull runtimes on startup -- Done
Move executor start logic into the "onStart" Swoole hook - Done
Add `updated` property to swoole table - Done
Clean up deployments older than X seconds - Done
Remove orphans on startup - done
Remove multiple request attempt to the runtime logic in executor - done
Remove builds param from delete endpoint - done
Shutdown callback isn't working as expected - done
Fix error handling - done
Decide on logic for build and runtime containers names ( runtime-ID and build-ID) - done
Add size validators for the runtime IDs - done
Fix logging - done
Incorporate Matej's changes in the build stage ( moving of the tar file will be performed by the runtime and not the build stage ) - done

## Test Plan

Added some more tests for the node runtime.

## Related PRs and Issues

NA

YES


eldadfux (approved):

app/executor.php
```diff
@@ -251,16 +257,13 @@ function logError(Throwable $error, string $action, Utopia\Route $route = null)
 
                 $device = new Local($destination);
                 $outputPath = $device->getPath(\uniqid() . '.' . \pathinfo('code.tar.gz', PATHINFO_EXTENSION));
+                var_dump($outputPath);
```
eldadfux, line 260: leftover

app/workers/builds.php
```diff
@@ -143,8 +140,14 @@ functionId: $functionId,
 
             Console::success("Build id: $buildId created");
 
+
+            var_dump("Going to set auto deploy");
+            var_dump($deployment);
+            var_dump($deployment->getAttribute('activate'));
+            var_dump(gettype($deployment->getAttribute('activate')));
```
lohanidamodar, line 147: do we need to remove this?

app/workers/builds.php
```diff
@@ -143,8 +140,14 @@ functionId: $functionId,
...
             /** Set auto deploy */
             if ($deployment->getAttribute('activate') === true) {
+                var_dump("Setting auto deploy");
```
lohanidamodar, line 150: remove dump?

composer.json
```diff
@@ -42,7 +42,7 @@
         "ext-zlib": "*",
         "ext-sockets": "*",
         "appwrite/php-clamav": "1.1.*",
-        "appwrite/php-runtimes": "dev-refactor",
+        "appwrite/php-runtimes": "0.7.2",
```
lohanidamodar, line 45: shouldn't it already be `apwrite/runtimes`?

src/Executor/Executor.php
```diff
@@ -34,29 +35,37 @@ public function createRuntime(
...
             'runtimeId' => "$projectId-$deploymentId",
             'source' => $source,
-            'destination' => APP_STORAGE_BUILDS . "/app-$projectId",
-            'vars' => $vars,
+            'destination' => $destination,
             'runtime' => $runtime,
             'baseImage' => $baseImage,
+            'entrypoint' => $entrypoint,
+            'workdir' => $workdir,
+            'network' => empty($network) ? App::getEnv('_APP_EXECUTOR_RUNTIME_NETWORK', 'openruntimes') : $network,
```
lohanidamodar, line 62: is default `openruntimes` fine? This being Appwrite, may be we can keep `appwriteruntimes`?

src/Executor/Executor.php
```diff
@@ -119,11 +126,46 @@ public function createExecution(
...
+                        );
+                        $response = $this->call(self::METHOD_POST, $route, $headers, $params, true, 30);
+                        $status = $response['headers']['status-code'];
+                        break;
+                    case 406:
+                        $response = $this->call(self::METHOD_POST, $route, $headers, $params, true, 30);
+                        $status = $response['headers']['status-code'];
+                        break;
+                    default:
+                        throw new \Exception($response['body']['message'], $status);
```
lohanidamodar, line 154: This seems problematic to me. Doesn't this will throw error  even if status is say `200` (ie status is not 404 or 406) before reaching the line 157, because default behavior is to throw an exception?

lohanidamodar (commented):
Few initial comments.

app/controllers/api/functions.php
```diff
@@ -519,7 +519,9 @@
             throw new Exception('Failed moving file', 500);
         }
 
-        if ((bool) $activate) {
+        $activate = (bool) filter_var($activate, FILTER_VALIDATE_BOOLEAN);
```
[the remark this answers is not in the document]
christyjacob4: Screenshot 2022-02-21 at 3 00 02 PM

Unfortunately not.

app/workers/builds.php
christyjacob4: Yes 👍  Removed it Thanks

app/workers/builds.php
christyjacob4: Removed 👍

composer.json
christyjacob4: I did try it, but for some reason my composer couldn't find it . I didn't spend much time trying to debug it.

src/Executor/Executor.php
christyjacob4: Good point. Let me do that 👍

app/executor.php
```diff
@@ -38,11 +40,11 @@
 // Fix error handling - done
 // Decide on logic for build and runtime containers names ( runtime-ID and build-ID) - done
 // Add size validators for the runtime IDs - done
+// Fix logging - done
+// Incorporate Matej's changes in the build stage ( moving of the tar file will be performed by the runtime and not the build stage ) - done
```
lohanidamodar, line 44: should we remove this if this is done? May be if we need reference keep it as a PR description instead?

app/executor.php
```diff
@@ -116,230 +130,184 @@ function logError(Throwable $error, string $action, Utopia\Route $route = null)
...
                 ->setCpus(App::getEnv('_APP_FUNCTIONS_CPUS', 1))
                 ->setMemory(App::getEnv('_APP_FUNCTIONS_MEMORY', 256))
                 ->setSwap(App::getEnv('_APP_FUNCTIONS_MEMORY_SWAP', 256));
             
-            $buildId = $orchestration->run(
+            $entrypoint = empty($commands) ? [] : [
+                'tail',
+                '-f',
+                '/dev/null'
+            ];
```
lohanidamodar, line 205: don't understand what we are doing here?

lohanidamodar (commented):
few more comments

src/Executor/Executor.php
christyjacob4: https://github.com/appwrite/appwrite/pull/2807/files#diff-23222bb861952e27ab08be6b8bbb449d43186436a2a110b5a24f3bbc61c287ffR157 

Since each switch case has a break statement, control comes out of the switch case. and this condition on like 157 ensures that we return the response body for any status code less than 400. 

I can explain further if its confusing. If you have pointers on how we could make it better, Id love to do that

app/executor.php
christyjacob4: done 👍

app/executor.php
christyjacob4: Okay so ..

In each runtime, we have an entrypoint command that runs `start.sh`

This script expects the build output to be present in /tmp/code.tar.gs
https://github.com/open-runtimes/open-runtimes/blob/main/runtimes/ruby-3.1/start.sh#L2

If the built code files are not present, the entry point command fails and the container shuts down. 

So what were doing here is essentially this

If there are build commands to be executed, keep the container active using `tail -f ...` else the built files are already preset so we dont need it.

PR #2810, opened 2022-02-18
Function Stroage Sync

lohanidamodar:
## What does this PR do?

Update storage changes and introduce deployment limit

## Test Plan

Existing tests

## Related PRs and Issues

YES


app/controllers/api/functions.php
```diff
@@ -375,7 +375,7 @@
         }
 
         if ($deployment->isEmpty()) {
-            throw new Exception('Deployment not found', 404);
+            throw new Exception('Deployment not found', Exception::DEPLOYMENT_NOT_FOUND);
```
christyjacob4, line 378:
```suggestion
            throw new Exception('Deployment not found', 404, Exception::DEPLOYMENT_NOT_FOUND);
```

At the moment, the `Exception` class still takes in the error code as the second parameter and the error type as the third parameter.  

In the next iteration, we'll derive the error codes from the error type as well. But for now, we still need 3 params

app/executor.php
```diff
@@ -238,11 +263,12 @@ function logError(Throwable $error, string $action, Utopia\Route $route = null)
                     throw new Exception('Something went wrong during the build process', 500);
                 }
 
-                $device = new Local($destination);
-                $outputPath = $device->getPath(\uniqid() . '.' . \pathinfo('code.tar.gz', PATHINFO_EXTENSION));
+                $destinationDevice =  getStorageDevice($destination);
+                $localDevice = new Local();
```
christyjacob4, line 267:
```suggestion
```
I believe we can reuse the `$localDevice` variable from above

app/workers/builds.php
```diff
@@ -39,7 +39,7 @@ public function run(): void
     {
         $type = $this->args['type'] ?? '';
         $projectId = $this->args['projectId'] ?? '';
-        $functionId = $this->args['functionId'] ?? '';
+        $functionId = $this->args['resourceId'] ?? '';
```
christyjacob4, line 42:
```suggestion
        $functionId = $this->args['functionId'] ?? '';
```
At the moment, the controller is passing `functionId` as key and not `resourceId` yet. So let's keep it that way .

christyjacob4 (changes requested):
Please address the comments

PR #2818, opened 2022-02-22
Functions refactor synced with 0.13.x

lohanidamodar:
## What does this PR do?

Sync 0.13.x with Functions refactor

## Test Plan

Existing tests

## Related PRs and Issues

https://github.com/appwrite/appwrite/pull/2810

YES


app/config/collections.php
```diff
@@ -1932,9 +1751,9 @@
                 '$id' => '_key_search',
                 'type' => Database::INDEX_FULLTEXT,
                 'attributes' => ['search'],
-                'lengths' => [2048],
-                'orders' => [Database::ORDER_ASC],
-            ]
+                'lengths' => [],
+                'orders' => [],
+            ],
```
christyjacob4, line 1756: Can you revert the change to the functions collection. This change shouldn't be a part of this PR

app/config/variables.php
```diff
@@ -420,12 +420,93 @@
...
+            [
+                'name' => '_APP_STORAGE_DEVICE',
+                'description' => 'Select default storage device. The default value is \'Local\'. List of supported adapters are \'Local\', \'S3\' and \'DOSpaces\'.',
+                'introduction' => '0.13.0',
+                'default' => 'Local',
+                'required' => false,
+                'question' => '',
+            ],
+            [
+                'name' => '_APP_STORAGE_DEVICE_S3_ACCESS_KEY',
```
christyjacob4, line 432:
```suggestion
                'name' => '_APP_STORAGE_S3_ACCESS_KEY',
```
The DEVICE word seems unnecessary as it does not add any additional meaning to the already existing word `_APP_STORAGE_S3_ACCESS_KEY` . I recommend discarding the DEVICE word and keeping the key name shorter.

app/config/variables.php
```diff
@@ -420,12 +420,93 @@
...
+                'description' => 'AWS S3 storage access key. Required when the storage adapter is set to S3. You can get your access key from your AWS console',
+                'introduction' => '0.13.0',
+                'default' => '',
+                'required' => false,
+                'question' => '',
+            ],
+            [
+                'name' => '_APP_STORAGE_DEVICE_S3_SECRET',
```
christyjacob4, line 440:
```suggestion
                'name' => '_APP_STORAGE_S3_SECRET',
```
Similar comment as above

app/config/variables.php
```diff
@@ -420,12 +420,93 @@
...
+                'description' => 'AWS S3 storage secret key. Required when the storage adapter is set to S3. You can get your secret key from your AWS console.',
+                'introduction' => '0.13.0',
+                'default' => '',
+                'required' => false,
+                'question' => '',
+            ],
+            [
+                'name' => '_APP_STORAGE_DEVICE_S3_REGION',
```
christyjacob4, line 448:
```suggestion
                'name' => '_APP_STORAGE_S3_REGION',
```

app/config/variables.php
```diff
@@ -420,12 +420,93 @@
...
+                'description' => 'AWS S3 storage region. Required when storage adapter is set to S3. You can find your region info for your bucket from AWS console.',
+                'introduction' => '0.13.0',
+                'default' => 'us-eas-1',
+                'required' => false,
+                'question' => '',
+            ],
+            [
+                'name' => '_APP_STORAGE_DEVICE_S3_BUCKET',
```
christyjacob4, line 456:
```suggestion
                'name' => '_APP_STORAGE_S3_BUCKET',
```

app/config/variables.php
```diff
@@ -420,12 +420,93 @@
...
+                'description' => 'AWS S3 storage bucket. Required when storage adapter is set to S3. You can create buckets in your AWS console.',
+                'introduction' => '0.13.0',
+                'default' => '',
+                'required' => false,
+                'question' => '',
+            ],
+            [
+                'name' => '_APP_STORAGE_DEVICE_DO_SPACES_ACCESS_KEY',
```
christyjacob4, line 464:
```suggestion
                'name' => '_APP_STORAGE_DO_SPACES_ACCESS_KEY',
```

app/config/variables.php
```diff
@@ -420,12 +420,93 @@
...
+                'description' => 'DigitalOcean spaces access key. Required when the storage adapter is set to DOSpaces. You can get your access key from your DigitalOcean console.',
+                'introduction' => '0.13.0',
+                'default' => '',
+                'required' => false,
+                'question' => '',
+            ],
+            [
+                'name' => '_APP_STORAGE_DEVICE_DO_SPACES_SECRET',
```
christyjacob4, line 472:
```suggestion
                'name' => '_APP_STORAGE_DO_SPACES_SECRET',
```

app/config/variables.php
```diff
@@ -420,12 +420,93 @@
...
+                'description' => 'DigitalOcean spaces secret key. Required when the storage adapter is set to DOSpaces. You can get your secret key from your DigitalOcean console.',
+                'introduction' => '0.13.0',
+                'default' => '',
+                'required' => false,
+                'question' => '',
+            ],
+            [
+                'name' => '_APP_STORAGE_DEVICE_DO_SPACES_REGION',
```
christyjacob4, line 480:
```suggestion
                'name' => '_APP_STORAGE_DO_SPACES_REGION',
```

app/config/variables.php
```diff
@@ -420,12 +420,93 @@
...
+                'description' => 'DigitalOcean spaces region. Required when storage adapter is set to DOSpaces. You can find your region info for your space from DigitalOcean console.',
+                'introduction' => '0.13.0',
+                'default' => 'us-eas-1',
+                'required' => false,
+                'question' => '',
+            ],
+            [
+                'name' => '_APP_STORAGE_DEVICE_DO_SPACES_BUCKET',
```
christyjacob4, line 488:
```suggestion
                'name' => '_APP_STORAGE_DO_SPACES_BUCKET',
```

app/config/variables.php
```diff
@@ -420,12 +420,93 @@
...
+                'question' => '',
+            ],
         ],
     ],
     [
         'category' => 'Functions',
         'description' => '',
         'variables' => [
+            [
+                'name' => '_APP_FUNCTIONS_DEPLOYMENT_LIMIT',
```
christyjacob4, line 502:
```suggestion
                'name' => '_APP_FUNCTIONS_SIZE_LIMIT',
```

I feel `_APP_FUNCTIONS_SIZE_LIMIT` or `_APP_DEPLOYMENT_SIZE_LIMIT` conveys the meaning better.  LMK your thoughts

app/config/collections.php
lohanidamodar: what do you mean?

app/config/collections.php
```diff
@@ -2021,6 +1840,50 @@
...
+                '$id' => 'metadata',
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 16384, // https://tools.ietf.org/html/rfc4288#section-4.2
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => ['json'],
+            ],
```
christyjacob4, line 1886:
```suggestion
```
Looks like metadata is declared twice

app/config/variables.php
lohanidamodar: _APP_FUNCTIONS because it's related to functions and in functions category, it's not a function's size limit but limit for DEPLOYMENT_SIZE. WDYT?

app/controllers/api/functions.php
```diff
@@ -535,33 +574,62 @@
...
-            'type' => BUILD_TYPE_DEPLOYMENT
-        ]);
+        if($chunksUploaded === $chunks) {
+            $fileSize = $deviceFunctions->getFileSize($path);
+
+            if ($deployment->isEmpty()) {
+                $deployment = $dbForProject->createDocument('deployments', new Document([
+                    '$id' => $deploymentId,
+                    '$read' => ['role:all'],
+                    '$write' => ['role:all'],
```
christyjacob4, line 584:
```suggestion
                    '$read' => [],
                    '$write' => [],
```
The `role:all` part does not seem secure.. @TorstenDittmann  @eldadfux  should we just leave the permissions empty ?

app/controllers/api/functions.php
```diff
@@ -535,33 +574,62 @@
...
+                    'resourceId' => $function->getId(),
+                    'resourceType' => 'functions',
+                    'dateCreated' => time(),
+                    'entrypoint' => $entrypoint,
+                    'path' => $path,
+                    'size' => $fileSize,
+                    'search' => implode(' ', [$deploymentId, $entrypoint]),
+                    'activate' => ((bool) $activate === true),
```
christyjacob4, line 592:
```suggestion
                    'activate' => $activate,
```

app/controllers/api/functions.php
```diff
@@ -535,33 +574,62 @@
...
+                    'resourceId' => $function->getId(),
+                    'resourceType' => 'functions',
+                    'dateCreated' => time(),
+                    'entrypoint' => $entrypoint,
+                    'path' => $path,
+                    'size' => 0,
+                    'chunksTotal' => $chunks,
+                    'chunksUploaded' => $chunksUploaded,
+                    'search' => implode(' ', [$deploymentId, $entrypoint]),
+                    'activate' => ((bool) $activate === true),
```
christyjacob4, line 624:
```suggestion
                    'activate' => $activate,
```

app/controllers/api/functions.php
```diff
@@ -698,15 +769,17 @@
...
         }
 
         if ($deployment->getAttribute('resourceId') !== $function->getId()) {
-            throw new Exception('Deployment not found', 404);
+            throw new Exception('Deployment not found', 404, Exception::DEPLOYMENT_NOT_FOUND);
         }
 
-        if (!$dbForProject->deleteDocument('deployments', $deployment->getId())) {
-            throw new Exception('Failed to remove deployment from DB', 500);
+        if ($deviceFunctions->delete($deployment->getAttribute('path', ''))) {
```
christyjacob4, line 779: We shouldn't be doing any delete operations here. The deletes worker should be doing it in the background

app/controllers/api/functions.php
```diff
@@ -535,33 +574,62 @@
...
-        ;
+            $usage
+                ->setParam('storage', $deployment->getAttribute('size', 0))
+            ;
+        } else {
+            if($deployment->isEmpty()) {
+                $deployment = $dbForProject->createDocument('deployments', new Document([
+                    '$id' => $deploymentId,
+                    '$read' => ['role:all'],
+                    '$write' => ['role:all'],
```
christyjacob4, line 614: similar comments about the permissions

app/controllers/api/functions.php
lohanidamodar: @christyjacob4 took from your branch?
https://github.com/appwrite/appwrite/blob/e9abe0f943b9e932bdb990fa594a9d555f5ebc4a/app/controllers/api/functions.php#L541

app/controllers/api/functions.php
lohanidamodar: is this handled in the background?

app/controllers/api/functions.php
```diff
@@ -535,33 +574,62 @@
...
+                    'resourceId' => $function->getId(),
+                    'resourceType' => 'functions',
+                    'dateCreated' => time(),
+                    'entrypoint' => $entrypoint,
+                    'path' => $path,
+                    'size' => 0,
```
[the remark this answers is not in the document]
lohanidamodar: fixed

app/controllers/api/functions.php
lohanidamodar: you have the same in your branch?

app/controllers/api/storage.php
```diff
@@ -2,32 +2,338 @@
...
+    ->label('event', 'storage.buckets.create')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'storage')
+    ->label('sdk.method', 'createBucket')
+    ->label('sdk.description', '/docs/references/storage/create-bucket.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_BUCKET)
+    ->param('bucketId', '', new CustomId(), 'Unique Id. Choose your own unique ID or pass the string `unique()` to auto generate it. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.')
+    ->param('name', '', new Text(128), 'Bucket name', false)
```
christyjacob4, line 51:
```suggestion
    ->param('name', '', new Text(128), 'Bucket name')
```
the `optional` param is `false` by default. might be a good idea to skip it

app/controllers/api/storage.php
```diff
@@ -2,32 +2,338 @@
...
+    ->param('enabled', true, new Boolean(), 'Is bucket enabled?', true)
+    ->param('maximumFileSize', (int) App::getEnv('_APP_STORAGE_LIMIT', 0), new Integer(), 'Maximum file size allowed in bytes. Maximum allowed value is ' . Storage::human(App::getEnv('_APP_STORAGE_LIMIT', 0), 0) . '. For self-hosted setups you can change the max limit by changing the `_APP_STORAGE_LIMIT` environment variable. [Learn more about storage environment variables](docs/environment-variables#storage)', true)
+    ->param('allowedFileExtensions', [], new ArrayList(new Text(64)), 'Allowed file extensions', true)
+    ->param('encryption', true, new Boolean(), 'Is encryption enabled? For file size above ' . Storage::human(APP_STORAGE_READ_BUFFER, 0) . ' encryption is skipped even if it\'s enabled', true)
+    ->param('antivirus', true, new Boolean(), 'Is virus scanning enabled? For file size above ' . Storage::human(APP_LIMIT_ANTIVIRUS, 0) . ' AntiVirus scanning is skipped even if it\'s enabled', true)
```
christyjacob4, line 59: When using the Boolean validator, I think it's important to use the `loose` argument to true.
Since some SDKs ( like JS ) send booleans as strings. and without the loose argument, it would fail for a valid string like "true" or "false". 

Also when dealing with booleans represented by strings, it's a good idea to use PHPs `filter_var` function to cast it to a boolean like we do in functions 
https://github.com/appwrite/appwrite/blob/feat-functions-refactor/app/controllers/api/functions.php#L522

app/controllers/api/storage.php
```diff
@@ -2,32 +2,338 @@
...
+    ->param('name', '', new Text(128), 'Bucket name', false)
+    ->param('permission', null, new WhiteList(['file', 'bucket']), 'Permissions type model to use for reading files in this bucket. You can use bucket-level permission set once on the bucket using the `read` and `write` params, or you can set file-level permission where each file read and write params will decide who has access to read and write to each file individually. [learn more about permissions](/docs/permissions) and get a full list of available permissions.')
+    ->param('read', null, new Permissions(), 'An array of strings with read permissions. By default no user is granted with any read permissions. [learn more about permissions](/docs/permissions) and get a full list of available permissions.', true)
+    ->param('write', null, new Permissions(), 'An array of strings with write permissions. By default no user is granted with any write permissions. [learn more about permissions](/docs/permissions) and get a full list of available permissions.', true)
+    ->param('enabled', true, new Boolean(), 'Is bucket enabled?', true)
```
christyjacob4, line 55:
```suggestion
    ->param('enabled', true, new Boolean(true), 'Is bucket enabled?', true)
```

app/controllers/api/storage.php
```diff
@@ -2,32 +2,338 @@
...
+    ->inject('usage')
+    ->action(function ($bucketId, $name, $permission, $read, $write, $enabled, $maximumFileSize, $allowedFileExtensions, $encryption, $antivirus, $response, $dbForProject, $audits, $usage) {
+        /** @var Appwrite\Utopia\Response $response */
+        /** @var Utopia\Database\Database $dbForProject */
+        /** @var Appwrite\Event\Event $audits */
+        /** @var Appwrite\Stats\Stats $usage */
+
+        $bucketId = $bucketId === 'unique()' ? $dbForProject->getId() : $bucketId;
+        try {
+            $files = Config::getParam('collections', [])['files'] ?? [];
```
christyjacob4, line 72: I think trying to access `files` without using `isset()` to check it first, causes a warning ( or throws an error ) 🤔

public/scripts/views/forms/upload.js
```diff
@@ -108,7 +108,7 @@
           expression.parse(element.dataset["write"] || "[]")
         );
 
-        sdk.storage.createFile('unique()', files[0], read, write, 1).then(
+        sdk.storage.createFile('default', 'unique()', files[0], read, write, 1).then(
           function(response) {
```
christyjacob4: Any reason we chose to have the bucket as the first param in create file  ? 🤔 
Every other  Appwrite endpoint that creates something, expects the ID first 🤔 .
Even the `createDocument` endpoint. What do you think ?

app/init.php
```diff
@@ -60,6 +66,10 @@
 const APP_PAGING_LIMIT = 12;
 const APP_LIMIT_COUNT = 5000;
 const APP_LIMIT_USERS = 10000;
+const APP_LIMIT_ANTIVIRUS = 20000000; //20MB
+const APP_LIMIT_ENCRYPTION = 20000000; //20MB
+const APP_LIMIT_COMPRESSION = 20000000; //20MB
+const APP_LIMIT_PREVIEW = 10000000; //10MB file size limit for preview endpoint
```
christyjacob4, line 72:
```suggestion
const APP_LIMIT_PREVIEW = 10 * (1000 * 1000); //10MB file size limit for preview endpoint
```

Might be nice to split it like this as you have done for `APP_STORAGE_READ_BUFFER `

app/init.php
```diff
@@ -60,6 +66,10 @@
 const APP_PAGING_LIMIT = 12;
 const APP_LIMIT_COUNT = 5000;
 const APP_LIMIT_USERS = 10000;
+const APP_LIMIT_ANTIVIRUS = 20000000; //20MB
+const APP_LIMIT_ENCRYPTION = 20000000; //20MB
+const APP_LIMIT_COMPRESSION = 20000000; //20MB
```
christyjacob4, line 71: Might be nice to split the number like you did for APP_STORAGE_READ_BUFFER

app/controllers/api/storage.php
lohanidamodar: the null check operator `?? []` prevents that.

public/scripts/views/forms/upload.js
lohanidamodar: I was not aware of this change.

app/init.php
lohanidamodar: I'm doing it for APP_STORAGE_READ_BUFFER because it's using MiB (1024), but this is using metric system and with comment //10 MB it's quite obvious, same for other similar vars?

app/init.php
lohanidamodar: answered above

src/Appwrite/Utopia/Response/Model/Bucket.php
```diff
@@ -0,0 +1,111 @@
...
+use Appwrite\Utopia\Response;
+use Appwrite\Utopia\Response\Model;
+
+class Bucket extends Model
+{
+
+    public function __construct()
+    {
+        $this
+        ->addRule('$id', [
```
christyjacob4: Very minor, looks like we use 1 level of indent here . Would be nice to maintain consistency 

https://github.com/appwrite/appwrite/blob/feat-functions-refactor/src/Appwrite/Utopia/Response/Model/Log.php#L13

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -303,6 +299,41 @@ public function testCreateDeployment($data):array
 
         // Wait for deployment to build.
         sleep(15);
+
+        /**
+         * Test for Large Code File SUCCESS
+         */
+        $source = realpath(__DIR__ . '/../../../resources/functions/php-large.tar.gz');
```
christyjacob4: We have a easier way to do this
https://github.com/appwrite/appwrite/blob/feat-functions-refactor/tests/e2e/Services/Functions/FunctionsCustomServerTest.php#L285-L295

We don't keep tar files in the repo anymore. The test suite creates the tar files from the source code during runtime. 

So you can remove this file as well
tests/resources/functions/php-large.tar.gz

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -303,6 +299,41 @@ public function testCreateDeployment($data):array
...
+            $counter++;
+            $id = $largeTag['body']['$id'];
+        }
+        @fclose($handle);
+
+        $this->assertEquals(201, $largeTag['headers']['status-code']);
+        $this->assertNotEmpty($largeTag['body']['$id']);
+        $this->assertIsInt($largeTag['body']['dateCreated']);
+        $this->assertEquals('index.php', $largeTag['body']['entrypoint']);
+        $this->assertGreaterThan(10000, $largeTag['body']['size']);
```
christyjacob4: Also, let's keep the large file upload test, as an independent test and not couple it with the existing `createDeployment` test. 

Let's do something like the `testCreateCustomNodeExecution()` method . This creates a deployment, an execution and even cleans up the container after the test. 

https://github.com/appwrite/appwrite/blob/feat-functions-refactor/tests/e2e/Services/Functions/FunctionsCustomServerTest.php#L851

app/http.php
```diff
@@ -141,8 +144,61 @@
...
+                'dateUpdated' => \time(),
+                'name' => 'Default',
+                'permission' => 'file',
+                'maximumFileSize' => (int) App::getEnv('_APP_STORAGE_LIMIT', 0), // 10MB
+                'allowedFileExtensions' => [],
+                'enabled' => true,
+                'encryption' => true,
+                'antivirus' => true,
+                '$read' => ['role:all'],
+                '$write' => ['role:all'],
```
christyjacob4, line 164: Are we sure about these permissions ?

app/controllers/api/storage.php
```diff
@@ -2,32 +2,338 @@
...
+    ->param('enabled', true, new Boolean(), 'Is bucket enabled?', true)
+    ->param('maximumFileSize', null, new Integer(), 'Maximum file size allowed in bytes. Maximum allowed value is ' . Storage::human((int)App::getEnv('_APP_STORAGE_LIMIT', 0), 0) . '. For self hosted version you can change the limit by changing _APP_STORAGE_LIMIT environment variable. [Learn more about storage environment variables](docs/environment-variables#storage)', true)
+    ->param('allowedFileExtensions', [], new ArrayList(new Text(64)), 'Allowed file extensions', true)
+    ->param('encryption', true, new Boolean(), 'Is encryption enabled? For file size above ' . Storage::human(APP_STORAGE_READ_BUFFER, 0) . ' encryption is skipped even if it\'s enabled', true)
+    ->param('antivirus', true, new Boolean(), 'Is virus scanning enabled? For file size above ' . Storage::human(APP_LIMIT_ANTIVIRUS, 0) . ' AntiVirus scanning is skipped even if it\'s enabled', true)
```
christyjacob4, line 231: Please refer to the comment above about booleans

app/controllers/api/storage.php
```diff
@@ -41,150 +347,345 @@
...
 
-        /*
-         * Validators
-         */
-        //$fileType = new FileType(array(FileType::FILE_TYPE_PNG, FileType::FILE_TYPE_GIF, FileType::FILE_TYPE_JPEG));
-        $fileSize = new FileSize(App::getEnv('_APP_STORAGE_LIMIT', 0));
-        $upload = new Upload();
+        /**
+     * Validators
+     */
```
christyjacob4, line 416: Can we get these aligned ?

app/controllers/api/storage.php
```diff
@@ -245,28 +785,53 @@
     ->label('sdk.response.code', Response::STATUS_CODE_OK)
     ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
     ->label('sdk.response.model', Response::MODEL_FILE)
+    ->param('bucketId', null, new UID(), 'Storage bucket unique ID. You can create a new storage bucket using the Storage service [server integration](/docs/server/storage#createBucket).')
     ->param('fileId', '', new UID(), 'File ID.')
```
christyjacob4, line 789: Don't we have an upper limit on the number of characters for an ID ? Shouldn't we specify it as part of the UID() constructor ?

christyjacob4 (changes requested):
Please address the comments

app/controllers/api/storage.php
christyjacob4: You can refer to the comment above 
https://github.com/appwrite/appwrite/pull/2818#discussion_r811911899

public/scripts/views/forms/upload.js
lohanidamodar: I don't think that's the case,  as collectionId, bucketId being the path param it comes before body param
https://appwrite.io/docs/client/database?sdk=web-default#databaseCreateDocument

app/http.php
lohanidamodar: yes, as permission is set to file level, this shouldn't matter

app/controllers/api/storage.php
lohanidamodar: It's already limited by the validator to 36 chars, we don't need to do anything

lohanidamodar:
@christyjacob4 all the comments should be addressed.

app/controllers/api/storage.php
christyjacob4: Okay perfect 👍

app/http.php
christyjacob4: okay 👍

app/init.php
christyjacob4: Alright makes sense 👍

PR #2819, opened 2022-02-22
Update response filters ready for 0.13

PineappleIOnic:
## What does this PR do?

Updates the Response filters to handle 0.12 SDK's contacting a 0.13 server

Yes, I have


app/controllers/general.php
```diff
@@ -157,6 +158,9 @@
     $responseFormat = $request->getHeader('x-appwrite-response-format', App::getEnv('_APP_SYSTEM_RESPONSE_FORMAT', ''));
     if ($responseFormat) {
         switch($responseFormat) {
+            case version_compare ($responseFormat , '0.12.3', '<='):
```
Meldiron, line 161: Why this specific version? Cant we do `0.12.0` or `0.13.0`?

src/Appwrite/Utopia/Response/Filters/V12.php
```diff
@@ -0,0 +1,101 @@
...
+        unset($content['bucketId']);
+        unset($content['chunksTotal']);
+        unset($content['chunksUploaded']);
+
+        return $content;
+    }
+
+    protected function parseFunction(array $content)
+    {
+        $content['execute'] = implode(' ', $content['execute']);
```
Meldiron, line 70: It was aray the whole time, wasnt it?

Meldiron (changes requested):

app/controllers/general.php
PineappleIOnic: the version_compare should compare those versions automatically

Meldiron: We released 0.12.4. Will that work fine? That's the only concern I have.

christyjacob4: we should give the number of our latest version here. 
so @PineappleIOnic , we need to replace it with `0.12.4`

app/controllers/general.php
```diff
@@ -157,6 +158,9 @@
...
+                Response::setFilter(new FiltersV12());
```
christyjacob4, line 162: This naming is inconsistent with the filter we already have. 

Since we also have request and response filter, lets use `ResponseV12` and `ResponseV11` here and `RequestV12` for the request filters

src/Appwrite/Utopia/Response/Filters/V12.php
```diff
@@ -0,0 +1,110 @@
...
+
+            case Response::MODEL_EXECUTION:
+                $parsedResponse = $this->parseExecution($content);
+
+            case Response::MODEL_USAGE_BUCKETS:
+                $parsedResponse = $this->parseUsageBuckets($content);
+
+            case Response::MODEL_USAGE_STORAGE:
+                $parsedResponse = $this->parseUsageStorage($content);
+
```
christyjacob4, line 38: We're missing a default case here

christyjacob4, line 38: Are we sure these are the only models that changed? I think a lot more has changed around the storage api, the ErrorDev response model, deployments . Please double check them

christyjacob4 (changes requested):
Please double check these . I find it hard to believe only this has changed 😅  😆

app/controllers/general.php
TorstenDittmann: We should do `0.12.*` - so we don't have to care about this.

TorstenDittmann: nvm, not working

Meldiron: I will either sound really smart or really dumb here, but why not `'0.13.0', '<'`?

Meldiron:
I don't think we have tests for this. Did you manually try to send HTTP requests with version header, trying (at least some of) these filters?

PR #2829, opened 2022-02-24
Update Request Filters for 0.13

PineappleIOnic:
## What does this PR do?

Update the Request Filters to provide backwards compatibility with 0.12 SDK's

Yes, I have


src/Appwrite/Utopia/Request/Filters/V13.php
```diff
@@ -0,0 +1,33 @@
...
+namespace Appwrite\Utopia\Request\Filters;
+
+use Appwrite\Utopia\Request\Filter;
+
+class V13 extends Filter
+{
+    // Convert 0.12 params format to 0.13 format
+    public function parse(array $content, string $model): array
+    {
+        switch ($model) {
```
Meldiron, line 12: I think we need one for function entrypoint, right? It was 'code', now it's 'entrypoint'. Wasnt it renamed?

Even if name wasn't changed, we should probably split by space and use the last thing.

src/Appwrite/Utopia/Request/Filters/V13.php
```diff
@@ -0,0 +1,33 @@
...
+            // Replaced Types
+            case "database.createFloatAttribute":
+                $content = $this->convertStringToNum($content, "min");
+                $content = $this->convertStringToNum($content, "max");
+                $content = $this->convertStringToNum($content, "default");
+                break;
+            case "database.createIntegerAttribute":
+                $content = $this->convertStringToNum($content, "min");
```
Meldiron, line 20: I am not sure if these were changed, please double-check

Meldiron (changes requested):

src/Appwrite/Utopia/Request/Filters/V13.php
PineappleIOnic: They were changes however double-checking it appears they changed the opposite way. From string to number. I've just fixed this.

app/controllers/general.php
```diff
@@ -45,6 +46,9 @@
     $requestFormat = $request->getHeader('x-appwrite-response-format', App::getEnv('_APP_SYSTEM_RESPONSE_FORMAT', ''));
     if ($requestFormat) {
         switch($requestFormat) {
+            case version_compare ($requestFormat , '0.13.0', '<') :
+                Request::setFilter(new V13());
```
christyjacob4, line 50: Let's use a prefix like `RequestV12` `RequestV13`

PR #2847, opened 2022-02-27
update create magic url session doc

lohanidamodar:
## What does this PR do?

Update documentation for magic url session because

Currently magic URL session docs (https://appwrite.io/docs/client/account?sdk=flutter-default#accountCreateMagicURLSession)
 specify that If you are on a mobile device you can leave the URL parameter empty, so that the login completion will be handled by your Appwrite instance by default. 

However, this has a problem as the login completion is handled by appwrite instance, the mobile doesn't receive the cookie and doesn't have the session data and related cookie, so therefore doesn't have a valid session and user is not authenticated. I think we should update to make it, on mobile device, the device should handle the login url and call updateMagicURLSession explicitly to get the cookies and session.

## Test Plan

N/A

## Related PRs and Issues

YES


docs/references/account/create-magic-url-session.md
```diff
@@ -1 +1,3 @@
...
\ No newline at end of file
+Sends the user an email with a secret key for creating a session. When the user clicks the link in the email, the user is redirected back to the URL you provided with the secret key and userId values attached to the URL query string. Use the query string parameters to submit a request to the [PUT /account/sessions/magic-url](/docs/client/account#accountUpdateMagicURLSession) endpoint to complete the login process. The link sent to the user's email address is valid for 1 hour.
+
+If you are on a mobile device, you must explicitly provide the URL parameter and handle the redirection of that URL to your app so that users clicking on the email will be redirected to your app with userId and secret. You can call `updateMagicURLSession` using the obtained information to create a valid session for the user in your application.
```
TorstenDittmann, line 3: `If you are on a mobile device` might be a bit misleading. Since also WebApps can be on a mobile device.

lohanidamodar: What do you suggest?

stnguyen90: How about:

> If you are building a mobile app, you must explicitly...

or:

> If you are not building a web app, you must explicitly...

lohanidamodar:
@TorstenDittmann updated.

docs/references/account/create-magic-url-session.md
```diff
@@ -1 +1,3 @@
...
\ No newline at end of file
+Sends the user an email with a secret key for creating a session. When the user clicks the link in the email, the user is redirected back to the URL you provided with the secret key and userId values attached to the URL query string. Use the query string parameters to submit a request to the [PUT /account/sessions/magic-url](/docs/client/account#accountUpdateMagicURLSession) endpoint to complete the login process. The link sent to the user's email address is valid for 1 hour.
+
+If you not building a web application, you must explicitly provide the URL parameter and handle the redirection of that URL to your app so that users clicking on the email will be redirected to your app with userId and secret. You can call `updateMagicURLSession` using the obtained information to create a valid session for the user in your application.
```
gewenyu99, line 3:
```suggestion
If you are not building a web application, you must explicitly provide the URL parameter and handle the URL redirect to your app so users will be redirected to your app with their userId and secret. You can call `updateMagicURLSession` using the obtained information to create a valid session for the user in your application.
```

gewenyu99 (changes requested):

eldadfux (approved):

PR #2848, opened 2022-02-27
Fix: rename 'sum' to 'total'

Meldiron:
## What does this PR do?

We noticed `sum` is not a commonly used standard, so we migrated to `total` instead.

- Response models, responses, variable names, and documentation were all updated
- Specs and SDKs (appwrite.js) were re-generated
- Response filter has been added
- appwrite/docs and appwrite-ee were checked, and no actions are required there

⚠️ Response filters not created yet ⚠️

## Test Plan

- Current tests must pass
- Manual QA (clicking through Appwrite Console, creating bunch of stuff)

## Related PRs and Issues

x

✅


src/Appwrite/Utopia/Response/Model/AttributeList.php
```diff
@@ -11,9 +11,9 @@ class AttributeList extends Model
     public function __construct()
     {
         $this
-            ->addRule('sum', [
+            ->addRule('total', [
                 'type' => self::TYPE_INTEGER,
-                'description' => 'Total sum of items in the list.',
+                'description' => 'Total amount of attributes in a collection.',
```
eldadfux, line 16:
```suggestion
                'description' => 'Total amount of attributes in the given collection.',
```

eldadfux: Maybe this will make it more consistent with memberships: `Total number of collection attributes.`

app/tasks/usage.php
```diff
@@ -487,7 +487,7 @@
                                 'files' => [
                                     'namespace' => '',
                                     'collectionPrefix' => 'bucket_',
-                                    'sum' => [
+                                    'total' => [
```
eldadfux, line 490: This is most likely forcing changes in the UI phtml files.

eldadfux (changes requested):
Looks good. Added a few comments. We're missing changes in the UI

src/Appwrite/Utopia/Response/Model/AttributeList.php
gewenyu99: Yes, please also use `total number` instead of `total amount`. In English, we use "the number of" for countable things, and "the amount of" for non-discrete stuff like weight or volume. It's a weird grammar thing.

This is one of those weird things that I still mess up all the time.

app/tasks/usage.php
Meldiron: Any specific place? I just updated UI, I think I found all sums.

gewenyu99 (changes requested):

eldadfux (approved):
Looks good, have you got a chance to verify all UI changes work properly?

Meldiron:
> Looks good, have you got a chance to verify all UI changes work properly?

I just did manual QA, and all seems to be fine.

docker-compose.yml
```diff
@@ -21,7 +21,7 @@ services:
       - --entrypoints.appwrite_websecure.address=:443
       - --accesslog=true
     ports:
-      - 80:80
+      - 8080:80
```
christyjacob4, line 24: This is for gitpod right ? Is this the recommended way to do it ?

Meldiron: Oh, i forgot to remove this. Yeah, its for GitPod. We cant use ports below 1024 (system ports)

PR #2852, opened 2022-02-28
feat: add option to configure worker numbers

TorstenDittmann:
## What does this PR do?

- adds `_WORKER_PER_CORE` env variable to configure it when needed
- using this variable to set the amount of workers per core on HTTP and Realtime

✅


christyjacob4 (changes requested):
Shouldn't we add this to the executor as well ? @TorstenDittmann

TorstenDittmann:
> Shouldn't we add this to the executor as well ? @TorstenDittmann

Yes, but for our sanity I wanted to not put more stuff into the functions refactor branch. So this needs to be adapted once functions refactor is up2date with `0.13.x`.

If you want, I can also prepare a PR to target functions refactor :-)

eldadfux (approved):

PR #2857, opened 2022-03-01
Handle JSON rules as objects instead of strings

abnegate:
## What does this PR do?

Handles JSON rules as objects instead of strings in specs, and sets a nested `additionalProperties` attribute. SDKs can handle these from the spec as maps instead of raw JSON strings


Meldiron (approved):
LGTM

christyjacob4 (approved):

PR #2895, opened 2022-03-03
feat: fix timeout issues

christyjacob4:
## What does this PR do?

1. Respect timeout environment variable in the API validators when set . 
2. Fix an uncaught exception
3. Increase Maintenance interval duration.
4. Set default value for execution `data` field

## Test Plan

## Related PRs and Issues

(Write your answer here.)

### DIFF
--- a/app/executor.php
+++ b/app/executor.php
@@ -423,7 +423,7 @@ App::post('/v1/execution')
             $errNo = -1;
             $executorResponse = '';
 
-            $timeout = $timeout ?? (int) App::getEnv('_APP_FUNCTIONS_TIMEOUT', 900);
+            $timeout ??= (int) App::getEnv('_APP_FUNCTIONS_TIMEOUT', 900);
 
             $ch = \curl_init();
             $body = \json_encode([
--- a/app/executor.php
+++ b/app/executor.php
@@ -25,14 +25,14 @@ use Utopia\Swoole\Response;
 use Utopia\Validator\ArrayList;
 use Utopia\Validator\Assoc;
 use Utopia\Validator\Boolean;
-use Utopia\Validator\Range as ValidatorRange;
+use Utopia\Validator\Range;
 use Utopia\Validator\Text;
 
 
 Runtime::enableCoroutine(true, SWOOLE_HOOK_ALL);
 
 /** Constants */
-const MAINTENANCE_INTERVAL = 1200; // 20 minutes
+const MAINTENANCE_INTERVAL = 3600; // 3600 seconds = 1 hour
 
 /**
 * Create a Swoole table to store runtime information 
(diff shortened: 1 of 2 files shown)


app/executor.php
```diff
@@ -395,9 +395,9 @@ function getStorageDevice($root): Device {
 App::post('/v1/execution')
     ->desc('Create an execution')
     ->param('runtimeId', '', new Text(64), 'The runtimeID to execute')
-    ->param('vars', [], new Assoc(), 'Environment variables required for the build', false)
-    ->param('data', '', new Text(8192), 'Data to be forwarded to the function, this is user specified.', true)
-    ->param('timeout', 15, new ValidatorRange(1, 900), 'Function maximum execution time in seconds.', true)
+    ->param('vars', [], new Assoc(), 'Environment variables required for the build')
+    ->param('data', '{}', new Text(8192), 'Data to be forwarded to the function, this is user specified.', true)
+    ->param('timeout', 15, new Range(1, (int) App::getEnv('_APP_FUNCTIONS_TIMEOUT', 900)), 'Function maximum execution time in seconds.')
```
TorstenDittmann, line 400: Timeout is now a required parameter?

That would be a breaking change for all SDK's.

app/executor.php
```diff
@@ -423,17 +423,19 @@ function (string $runtimeId, array $vars, string $data, $timeout, $activeRuntime
             $errNo = -1;
             $executorResponse = '';
 
+            $timeout = $timeout ?? (int) App::getEnv('_APP_FUNCTIONS_TIMEOUT', 900);
```
TorstenDittmann, line 426: You can use a short syntax here:

```suggestion
            $timeout ??= (int) App::getEnv('_APP_FUNCTIONS_TIMEOUT', 900);
```

src/Executor/Executor.php
```diff
@@ -159,13 +161,18 @@ public function createExecution(
...
-        if ($status >= 400) {
-            for ($attempts = 0; $attempts < 10; $attempts++) {
-                switch ($status) {
-                    case 404:
+        for ($attempts = 0; $attempts < 10; $attempts++) {
+            try {
+                switch (true) {
+                    case $status < 400:
+                        return $response['body'];
+                    case $status == 404:
```
TorstenDittmann, line 175: I would suggest using strict comparison here with `===` to prevent type juggling.

src/Executor/Executor.php
```diff
@@ -176,31 +183,23 @@ public function createExecution(
...
-                if ($status != 406) {
-                    throw new \Exception($response['body']['message'], $status);
-                }
-
-                sleep(2);
+            } catch (\Exception $e) {
+                throw new \Exception($e->getMessage(), $e->getCode());
             }
-            throw new Exception($response['body']['message'], 503);
+            sleep(2);
```
TorstenDittmann, line 199: What is this timeout for?

app/executor.php
christyjacob4: No don't worry about this one. This is the internal API route used for communication b/w appwrite and the executor. The functions controller itself is unchanged .

TorstenDittmann: Ah yeah! just realized 😆

src/Executor/Executor.php
christyjacob4: addressed 👍

app/executor.php
christyjacob4: done 👍

src/Executor/Executor.php
christyjacob4: This timeout is used when we receive a `Runtime Not Found` error from the executor.

So we need to wait and make a couple of execution attempts to wait for the runtime to be created

TorstenDittmann (approved):

arturgoms:
I can confirm that this solve the problem that I was having, Thank you @christyjacob4 !

Issue #2770, opened 2022-02-09
bug: when uploaded image is missing the extension, now showing the preview and thumbnail


YahalomsGuy:
### 👟 Reproduction steps

Following discussion with @eldadfux 
Upload an image file to the server without the file extension.

go to the console, 
The thumbnail and the preview of the image are not shown, although it recognizes the file and "image" 
press the "New Window" link, it opens.

try to "storage.getFilePreview" through the API in the front end , does not render the image

### 👍 Expected behavior

since the console recognizes the file as image although it does not have extension, should be able to render and show the preview and the thumbnail

### 👎 Actual Behavior

Does not show.

### 🎲 Appwrite version

Version 0.12.x

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

eldadfux:
This was fixed, merged and will be released in Appwrite 0.14

PR #2942, opened 2022-03-13
fix preview output when no output param and type doesn't exist

lohanidamodar:
## What does this PR do?

- When a file extension was not provided, the preview endpoint without `output` param would fail as output type would be empty (default was taken from the file extension). This PR fixes by making default output to `jpg` if both output and file type are missing.
- Adds `_APP_STORAGE_PREVIEW_LIMIT` environment variable to configure the image preview file size limit

## Test Plan

## Related PRs and Issues

#2770 

YES


app/controllers/api/storage.php
```diff
@@ -950,9 +950,8 @@
         $cache = new Cache(new Filesystem(APP_STORAGE_CACHE . DIRECTORY_SEPARATOR . 'app-' . $project->getId() . DIRECTORY_SEPARATOR . $bucketId . DIRECTORY_SEPARATOR . $fileId)); // Limit file number or size
         $data = $cache->load($key, 60 * 60 * 24 * 30 * 3/* 3 months */);
 
+        $output = (empty($output)) ? ( empty($type) ? 'jpg' : $type ) : $output;
```
eldadfux, line 953: Why are we assuming jpeg? What happens if the file is not an image? Maybe using the file mime-type to determine the default would be better.

lohanidamodar: If the file is not an image, that part is already calculated above to get a default image preview.

lohanidamodar: Using mime/type here to determine will again cause issue if the mime-type is not of an image.

eldadfux: Agree, but if it is an image, we can tell if it's jpg, gif, or png.

lohanidamodar: If it's an image, `$type` already has the type and it takes `$type` instead of default `jpg` ?

eldadfux (approved):

app/controllers/api/storage.php
```diff
@@ -944,9 +944,11 @@
         $cache = new Cache(new Filesystem(APP_STORAGE_CACHE . DIRECTORY_SEPARATOR . 'app-' . $project->getId() . DIRECTORY_SEPARATOR . $bucketId . DIRECTORY_SEPARATOR . $fileId)); // Limit file number or size
         $data = $cache->load($key, 60 * 60 * 24 * 30 * 3/* 3 months */);
 
-        if ($data) {
-            $output = (empty($output)) ? $type : $output;
+        if(empty($output)) {
+            $output = empty($type) ? (array_search($mime, $outputs) ?? 'jpg') : $type;
```
wess, line 948: If reading correctly, we are falling back to jpg for file type?

Meldiron (approved):

TorstenDittmann:
> This PR is for `App::get('/v1/storage/buckets/:bucketId/files/:fileId/preview')` and we are assuming images only for this, perhaps it should be `:bucketId/images/...` otherwise we can't assume everything is an image, yes?

That would imply that an image is not a file? 😉

wess:
> This PR is for App::get('/v1/storage/buckets/:bucketId/files/:fileId/preview') and we are > assuming images only for this, perhaps it should be :bucketId/images/... otherwise we can't assume everything is an image, yes?


> That would imply that an image is not a file? 

No but `jpg` as the fallback implies it's an image, do we offer preview for, say, a text file? :)

lohanidamodar:
> > This PR is for App::get('/v1/storage/buckets/:bucketId/files/:fileId/preview') and we are > assuming images only for this, perhaps it should be :bucketId/images/... otherwise we can't assume everything is an image, yes?
> 
> > That would imply that an image is not a file?
> 
> No but `jpg` as the fallback implies it's an image, do we offer preview for, say, a text file? :)

Yes for non image files we have a fallback icons used for preview. So everything coming to this point is a file with mime type for some kind of image.

app/controllers/api/storage.php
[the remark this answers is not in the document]
lohanidamodar: Sure we can,

lohanidamodar: yes at this point, what we receive will be image. Non image files will use default logo images for preview.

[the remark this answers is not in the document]
lohanidamodar: added comment

wess (approved):

lohanidamodar:
@TorstenDittmann if you don't have any other concerns let's merge this.

TorstenDittmann (approved):

Issue #2935, opened 2022-03-10
🐛 Bug Report: All function files are deleted when the container is stopped


Asored-D:
### 👟 Reproduction steps

1. Deploy a cloud function for a project (Just a simple Hello World output)
2. Run `docker-compose down`
3. Run `docker-compose up -d`
4. Try to execute the function. There will be an error with the string: `File Not Found`
5. Deploy the same function again
6. Now executions are working 

### 👍 Expected behavior

I would expect that after rebooting the containers, you don't have to redeploy all the functions every time. For large projects, that could be a lot. 

### 👎 Actual Behavior

Currently, you have to redeploy the functions after each container restart. Otherwise they do not work and return "File not found". The programming language of the function does not seem to matter. I tested it with DART and PHP functions.

### 🎲 Appwrite version

Version 0.13.x

### 💻 Operating system

MacOS

### 🧱 Your Environment

No special customizations.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

stnguyen90:
My theory is the code is mounted into a volume, but the volume is in the appwrite-executor container. When the container is removed, the volume is also removed.

Source: https://github.com/appwrite/appwrite/blob/406f914585f74bef574f1bcf6cda36c39eb8696b/app/executor.php#L229

Meldiron:
> My theory is the code is mounted into a volume, but the volume is in the appwrite-executor container. When the container is removed, the volume is also removed.
> 
> Source:
> 
> https://github.com/appwrite/appwrite/blob/406f914585f74bef574f1bcf6cda36c39eb8696b/app/executor.php#L214-L229

This sounds very correct, great feedback!

s3ppo:
> My theory is the code is mounted into a volume, but the volume is in the appwrite-executor container. When the container is removed, the volume is also removed.
> 
> Source:
> 
> https://github.com/appwrite/appwrite/blob/406f914585f74bef574f1bcf6cda36c39eb8696b/app/executor.php#L214-L229

i think you are 100% right
i testet it and redeployed the executor docker.. after that all my functions where red with the error ```File not found```

Meldiron:
I am not able to reproduce the problem 😬

Meldiron:
[...]

PR #2954, opened 2022-03-16
Fix: Merged docker composes

Meldiron:
## What does this PR do?

There were many differences between development and production docker-compose, such as:
- Version missmatches
- Missing/additional environment variables
- Missing volumes

This PR made a diff between those files and made sure all changes are applied between them.

Same has been done for .env.

## Test Plan

1. Manual QA, Appwrite should still work
2. Compare new differences to make sure nothing is left: https://www.diffchecker.com/QqPCJgQM
3. Compare .env with json-diff.com

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/2935

✅


app/views/install/compose.phtml
```diff
@@ -512,25 +516,25 @@ services:
     command: 'mysqld --innodb-flush-method=fsync'
 
   redis:
-    image: redis:6.0-alpine3.12
+    image: redis:6.2-alpine
```
TorstenDittmann, line 519: We decided back then to specify the alpine version - since the base image is re-used across all images. So lets try to stick with it 👍🏻 

Not sure if we are consistent, if not or there is no 3.12 for redis 6.2 you can leave it as it is 🙂

TorstenDittmann (approved):

app/views/install/compose.phtml
Meldiron: Leaving as is because there isnt 3.12 version, only 3.13.

christyjacob4 (approved):

PR #2979, opened 2022-03-21
Fix: Default payload to empty string

Meldiron:
## What does this PR do?

Currently default payload is `{}`. Open Runtimes were updated to default to `''`, and so should be Appwrite. This PR makes sure `req.payload` in user function is empty string if not provided, instead of string `'{}'`.

## Test Plan

Manual QA went well.

## Related PRs and Issues

x

✅


Meldiron:
With this change, example code now properly works: `const payload = req.payload || 'No payload.'`

christyjacob4 (approved):

PR #2980, opened 2022-03-21
Feat increase execution output limit to 1MB and Fix Unicode

PineappleIOnic:
## What does this PR do?
This PR increases the output limit

## Test Plan

3 test functions were written, one that outputs one more than the limit, one that outputs at the limit and another that is below the limit. The first test correctly truncated the data while the two others did not truncate the data.

Yes, I have


Meldiron:
Was this change on purpose?

CleanShot 2022-03-21 at 15 34 53

Meldiron (commented):
- It was 16kb before, or 16k characters? Is it now 1MB, or 1M characters? Please double-check
- If it's 1MB, we use multi-byte, and only allow up to 250k characters. Is that enough?
- Is there any indicator that a response was cut? If not, can we add one?
- Let's add a test to make sure 500kB is possible, and 2MB is not

app/config/collections.php
```diff
@@ -2135,7 +2135,7 @@
                 '$id' => 'stderr',
                 'type' => Database::VAR_STRING,
                 'format' => '',
-                'size' => 16384,
+                'size' => 1000000,
```
christyjacob4, line 2138: Im not sure of the size parameter. 

Is this the number of bytes or the number of characters ?

PineappleIOnic: Good question, I'll go into the Database library and see what it's setting

PineappleIOnic: it appears to be chars which can vary because of unicode

PineappleIOnic: There could be potentially 1 million characters but can also potentially only be 250k if they are using more unique characters that takes 4 bytes in UTF-8

Meldiron:
Reminder we should make sure this limit is properly documented.

Meldiron:
We will separate `stdout`, `stderr` and `response` in other PRs. Should `response` be limit to 1MB too? Should we throw error if response is bigger than 1MB? People usually respond with JSONs anyway, and they would fail to parse.

lohanidamodar (approved):
LGTM.

abnegate (approved):

app/executor.php
```diff
@@ -283,8 +283,8 @@ function getStorageDevice($root): Device {
             $endTime = \time();
             $container = array_merge($container, [
                 'status' => 'ready',
-                'stdout' => \utf8_encode($stdout),
-                'stderr' => \utf8_encode($stderr),
+                'stdout' => \mb_strcut(\utf8_encode($stdout), 0, 1000000), // Limit to 1MB
+                'stderr' => \mb_strcut(\utf8_encode($stderr), 0, 1000000), // Limit to 1MB
```
Meldiron, line 287: We keep utf8_encode here, but we remove it from elsewhere. Is that fine?

Meldiron (approved):
LGTM, one concern

app/executor.php
PineappleIOnic: Thank you for bringing this up, It is now fixed. 👍

app/executor.php
```diff
@@ -508,12 +508,12 @@ function (string $runtimeId, array $vars, string $data, $timeout, $activeRuntime
...
         
             Console::success('Function executed in ' . $executionTime . ' seconds, status: ' . $functionStatus);
-        
+
             $execution = [
                 'status' => $functionStatus,
                 'statusCode' => $statusCode,
-                'stdout' => \utf8_encode(\mb_substr($stdout, -16384)),
-                'stderr' => \utf8_encode(\mb_substr($stderr, -16384)),
+                'stdout' => \mb_strcut($stdout, 0, 1000000), // Limit to 1MB
```
Meldiron, line 515: Can we get this number from config? Its something that jake did in his PR: https://github.com/appwrite/appwrite/pull/3160/files

christyjacob4: +1 ^

christyjacob4: It's fine here since this is the executor, and we will move it out of Appwrite eventually. So we cannot use the attributes from Appwrite in here.

christyjacob4 (approved):

christyjacob4:
@PineappleIOnic  the tests are failing, please check.

christyjacob4 (changes requested):
@PineappleIOnic  the tests are failing, please check.

PineappleIOnic:
> @PineappleIOnic the tests are failing, please check.

It was a typo, just fixed it.

PR #2981, opened 2022-03-21
Feat: HTTPs Redirection Exception

Meldiron:
## What does this PR do?

If you have `_APP_OPTIONS_FORCE_HTTPS` enabled, and use `http` protocol, all requests are behaving as `GET` posts, resulting in really weird unexpected results.

As I could not find any good solution to redirect while keeping the method, Appwrite will now throw error if this happens. A solution for developer would be to start using `HTTPS` protocol with instances that has HTTPS force option enabled.

## Test Plan

- [x] Manual QA:
image

## Related PRs and Issues

x (problem found during 0.13 QA)

✅


TorstenDittmann (approved):

app/config/errors.php
```diff
@@ -78,6 +78,11 @@
         'description' => 'An internal server error occurred.',
         'code' => 500,
     ],
+    Exception::GENERAL_FORCED_HTTPS_IGNORED => [
+        'name' => Exception::GENERAL_FORCED_HTTPS_IGNORED,
+        'description' => 'Appwrite instance forces secure communication. Switch to HTTPS protocol in your endpoint.',
```
christyjacob4, line 83: PR looks good but im not satisfied with the naming of the KEY and Description

app/controllers/general.php
```diff
@@ -173,6 +173,10 @@
      */
     if (App::getEnv('_APP_OPTIONS_FORCE_HTTPS', 'disabled') === 'enabled') { // Force HTTPS
         if ($request->getProtocol() !== 'https') {
+            if($request->getMethod() !== Request::METHOD_GET) {
+                throw new Exception('HTTPS communication required.', 500, Exception::GENERAL_FORCED_HTTPS_IGNORED);
```
christyjacob4, line 177: We can use a better description here

christyjacob4 (changes requested):
Please address the comments

app/controllers/general.php
christyjacob4:
```suggestion
                throw new Exception('Method unsupported over HTTP',405, Exception::GENERAL_METHOD_UNSUPPORTED);
```

src/Appwrite/Extend/Exception.php
```diff
@@ -46,6 +46,7 @@ class Exception extends \Exception
     const GENERAL_ROUTE_NOT_FOUND           = 'general_route_not_found';
     const GENERAL_CURSOR_NOT_FOUND          = 'general_cursor_not_found';
     const GENERAL_SERVER_ERROR              = 'general_server_error';
+    const GENERAL_FORCED_HTTPS_IGNORED      = 'general_forced_https_ignored';
```
christyjacob4, line 49:
```suggestion
    const GENERAL_METHOD_UNSUPPORTED      = 'general_method unsupported';
```

app/config/errors.php
christyjacob4, line 83:
```suggestion
    Exception::GENERAL_METHOD_UNSUPPORTED => [
        'name' => Exception::GENERAL_METHOD_UNSUPPORTED,
        'description' => 'The HTTP method is unsupported by the current protocol or resource.',
```

app/controllers/general.php
```diff
@@ -173,6 +173,10 @@
      */
     if (App::getEnv('_APP_OPTIONS_FORCE_HTTPS', 'disabled') === 'enabled') { // Force HTTPS
         if ($request->getProtocol() !== 'https') {
+            if($request->getMethod() !== Request::METHOD_GET) {
+                throw new Exception('Method unsupported over HTTP.', 405, Exception::GENERAL_METHOD_UNSUPPORTED);
```
TorstenDittmann, line 177: Here it says `405` - but in the errors its a `500`?

Meldiron: Changed to 405 everywhere.

app/config/errors.php
```diff
@@ -78,6 +78,11 @@
         'description' => 'An internal server error occurred.',
         'code' => 500,
     ],
+    Exception::GENERAL_METHOD_UNSUPPORTED => [
+        'name' => Exception::GENERAL_METHOD_UNSUPPORTED,
+        'description' => 'The HTTP method is unsupported by the current protocol or resource.',
+        'code' => 405, 
+    ],
```
christyjacob4, line 85:
```suggestion
    Exception::GENERAL_PROTOCOL_UNSUPPORTED => [
        'name' => Exception::GENERAL_PROTOCOL_UNSUPPORTED,
        'description' => 'The current protocol is unsupported for the current request. ',
        'code' => 500, 
    ],
```

app/controllers/general.php
christyjacob4, line 177:
```suggestion
                throw new Exception('Method unsupported over HTTP.', 500, Exception:: GENERAL_PROTOCOL_UNSUPPORTED);
```

src/Appwrite/Extend/Exception.php
```diff
@@ -46,6 +46,7 @@ class Exception extends \Exception
     const GENERAL_ROUTE_NOT_FOUND           = 'general_route_not_found';
     const GENERAL_CURSOR_NOT_FOUND          = 'general_cursor_not_found';
     const GENERAL_SERVER_ERROR              = 'general_server_error';
+    const GENERAL_METHOD_UNSUPPORTED        = 'general_method_unsupported';
```
christyjacob4, line 49:
```suggestion
    const GENERAL_PROTOCOL_UNSUPPORTED        = 'general_protocol_unsupported';
```

christyjacob4 (changes requested):
Sorry for the change ..
But  405 - Method Not Allowed does not apply here since we're talking about protocols ( https, http ) and not methods ( GET,  POST, ... )

src/Appwrite/Extend/Exception.php
```diff
@@ -46,6 +46,7 @@ class Exception extends \Exception
     const GENERAL_ROUTE_NOT_FOUND           = 'general_route_not_found';
     const GENERAL_CURSOR_NOT_FOUND          = 'general_cursor_not_found';
     const GENERAL_SERVER_ERROR              = 'general_server_error';
+    const GENERAL_PROTOCOL_UNSUPPORTED        = 'general_protocol_unsupported';
```
christyjacob4, line 49:
```suggestion
    const GENERAL_PROTOCOL_UNSUPPORTED       = 'general_protocol_unsupported';
```

abnegate:
Could we use a [307](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307) or [308](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308) status code override [here](https://github.com/appwrite/appwrite/blob/1e0fe177a8d859a5490674927af0a41742b9dfd0/app/controllers/general.php#L180) to redirect while keeping the method intact?

christyjacob4:
@abnegate, That's an interesting idea. 
This was our first approach to the problem as well but I believe @Meldiron tried it and it was unsuccessful ? 

@Meldiron  Can you confirm ?

TorstenDittmann:
> @abnegate, That's an interesting idea. 
> This was our first approach to the problem as well but I believe @Meldiron tried it and it was unsuccessful ? 
> 
> @Meldiron  Can you confirm ?  

Yes, it didn't work. That was my initial idea to solve this.

PR #2982, opened 2022-03-21
Fix:  Validation regex for custom ID

Meldiron:
## What does this PR do?

image

The current frontend regex we use does not allow 1 letter attribute IDs, even tho backend allows it.

This PR fixed the regex to allow 1 letter IDs.

## Test Plan

- [x] Manual QA:
CleanShot 2022-03-21 at 15 54 56

## Related PRs and Issues

x

✅

### DIFF
--- a/app/views/console/comps/header.phtml
+++ b/app/views/console/comps/header.phtml
@@ -211,7 +211,7 @@
             required
             maxlength="36"
             class=""
-            pattern="^[a-zA-Z0-9][a-zA-Z0-9_.-]{1,36}$"
+            pattern="^[a-zA-Z0-9][a-zA-Z0-9._-]{0,36}$"
             name="projectId" />
 
         <label>Name</label>
--- a/app/views/console/comps/header.phtml
+++ b/app/views/console/comps/header.phtml
@@ -211,7 +211,7 @@
             required
             maxlength="36"
             class=""
-            pattern="^[a-zA-Z0-9][a-zA-Z0-9._-]{0,36}$"
+            pattern="^[a-zA-Z0-9][a-zA-Z0-9._-]{0,35}$"
             name="projectId" />
 
         <label>Name</label>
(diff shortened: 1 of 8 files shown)


app/views/console/comps/header.phtml
```diff
@@ -211,7 +211,7 @@
             required
             maxlength="36"
             class=""
-            pattern="^[a-zA-Z0-9][a-zA-Z0-9_.-]{1,36}$"
+            pattern="^[a-zA-Z0-9][a-zA-Z0-9._-]{0,36}$"
```
TorstenDittmann, line 214:
```suggestion
            pattern="^[a-zA-Z0-9][a-zA-Z0-9._-]{0,35}$"
```

It needs to be 0,35. Just to make sure - even though we have a maxlength.

Meldiron: Applied.

app/views/console/comps/header.phtml
```diff
@@ -211,7 +211,7 @@
             required
             maxlength="36"
             class=""
-            pattern="^[a-zA-Z0-9][a-zA-Z0-9_.-]{1,36}$"
+            pattern="^[a-zA-Z0-9][a-zA-Z0-9._-]{0,35}$"
```
christyjacob4, line 214: Is there a difference between 
`[a-zA-Z0-9_.-]` and `[a-zA-Z0-9._-]`

christyjacob4 (changes requested):
Just one comment

app/views/console/comps/header.phtml
Meldiron: There isn't, just for consistency. It was reversed only in one place, so I fixed it.

CleanShot 2022-03-23 at 08 42 56

christyjacob4 (approved):

PR #2983, opened 2022-03-21
Show Stdout with build logs when build fails

PineappleIOnic:
## What does this PR do?

This PR will make it so the logs popup box for function builds will display stdout aswell as stderr when a build fails

Yes, I have


Meldiron (changes requested):
__Manual QA__

1. I can see `logs` when its building. It indicates I would see realtine data, which I don't (its empty). Can we hide it?
2. When finished and output is empty, language on right side is missplaced. Can we solve this? Maybe a task for Elad? If not, can we say `No logs` so it looks better?

CleanShot 2022-03-22 at 09 12 41@2x
CleanShot 2022-03-22 at 09 13 14@2x

TorstenDittmann:
> **Manual QA**
> 
> 1. I can see `logs` when its building. It indicates I would see realtine data, which I don't (its empty). Can we hide it?
> 2. When finished and output is empty, language on right side is missplaced. Can we solve this? Maybe a task for Elad? If not, can we say `No logs` so it looks better?
> 
> CleanShot 2022-03-22 at 09 12 41@2x
> 
> CleanShot 2022-03-22 at 09 13 14@2x

If no logs, we should show `No Logs` 👍🏻

christyjacob4 (approved):

christyjacob4:
@PineappleIOnic Can you attach a screenshot of what it looks like now ?

PineappleIOnic:
image

eldadfux:
> image

Would it make more sense to change the labels to stdout and stderr?

eldadfux:
Also, how does the UI look when both logs have a lot of data?

PineappleIOnic:
It currently doesn't look the best, we may need to add a max-height with a auto scroll y
image

PineappleIOnic:
I've implemented your suggestions, logs now look like this:
image

eldadfux (approved):
@christyjacob4 I added this to manual QA so we get more hands testing the new components when having long logs.

eldadfux:
Not relevant due to the new console.

Issue #2933, opened 2022-03-10
🐛 Bug Report: Custom domain certificates fail after re-adding


Meldiron:
### 👟 Reproduction steps

1. Add custom domain
2. Verify custom domain
(HTTPS works)
4. Delete custom domain
5. Add exact same custom domain
6. Verify exact same domain
(HTTPS does not work)

### 👍 Expected behavior

HTTPS working after re-adding exact same domain.

### 👎 Actual Behavior

The certificate worker does not renew, as it thinks the certificate is still valid. (Renew not required).

As far as I am concerned, document in the certificates collection is not being removed, although the certificate is removed from sotrage.

### 🎲 Appwrite version

Version 0.13.x

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

Meldiron:
Fixed, released in 0.14 😇

PR #2997, opened 2022-03-23
Fix: Certificates in general

Meldiron:
## What does this PR do?

I noticed multiple problems with certificates on custom domains, and addressed them all in this PR.

#### ✅ 1. Allow using same domain on multiple projects

Currently, your second project would be stuck at `in progress` of cert creation, because `Renew isn't required`, which is true. This PR makes sure to set certificate ID on the domain, even tho renew isn't required.

https://user-images.githubusercontent.com/19310830/159718472-8c29373a-d6bb-4a90-a45e-fbd276f5b7bc.mp4

#### ✅ 2. Delete certificate document when deleting certificate file

We currently only remove files, which results in Appwrite still thinking certificate is there (and renewal isn't required), even tho file is already missing and Traefik cannot communicate over HTTPS.

https://user-images.githubusercontent.com/19310830/159723802-5ae4466d-8302-404a-beb4-4c5405a24b34.mp4

#### ✅ 3. Only delete certificate if no other domain is using it

With step 2 fixed, now if you delete a domain, it deletes the certificate too. Problem is, another project could still be using it. Let's not remove it! In deletes worker, we check if no domain is using certificate before deleting it.

https://user-images.githubusercontent.com/19310830/159730642-4f9560d4-6239-4f88-a048-ef218964a925.mp4

## Test Plan

The PR was created from DO droplet, where I was manually testing the solution is working before pushing to git.

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/2933

✅

### DIFF
--- a/app/workers/deletes.php
+++ b/app/workers/deletes.php
@@ -18,6 +18,9 @@ use Utopia\Audit\Audit;
 
 require_once __DIR__ . '/../init.php';
 
+Authorization::disable();
+Authorization::setDefaultStatus(false);
+
 Console::title('Deletes V1 Worker');
 Console::success(APP_NAME . ' deletes worker v1 has started' . "\n");
 
@@ -38,7 +41,6 @@ class DeletesV1 extends Worker
 
     public function run(): void
     {
-        Authorization::disable();
 
         $projectId = $this->args['projectId'] ?? '';
         $type = $this->args['type'] ?? '';
@@ -113,8 +115,6 @@ class DeletesV1 extends Worker
                 Console::error('No delete operation for type: ' . $type);
                 break;
         }
-
-        Authorization::reset();
     }
 
     public function shutdown(): void
--- a/app/workers/deletes.php
+++ b/app/workers/deletes.php
@@ -18,8 +18,6 @@ use Utopia\Audit\Audit;
 
 require_once __DIR__ . '/../init.php';
 
-Authorization::disable();
-
 Console::title('Deletes V1 Worker');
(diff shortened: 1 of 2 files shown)


app/workers/deletes.php
```diff
@@ -113,6 +113,8 @@ public function run(): void
                 Console::error('No delete operation for type: ' . $type);
                 break;
         }
+
+        Authorization::reset();
```
eldadfux, line 117: Why did we change the Authorization behavior? usually in worker context we want to act as admin and have no permission check.

Meldiron: Initially, I wanted to only disable it at the top of the file, but that didn't work for some reason - some other files must have reseted it somewhere.

What I did is I copied the solution from another worker (certificates), to disable at the beginning of the run, and reset at the end.

TorstenDittmann: Maybe there is some calls in the database library itself?

TorstenDittmann: Maybe the db library itself has calls to reset it?

TorstenDittmann: Maybe the db library itself has calls to reset it?

arbile26:
I'm facing the same problem....

app/workers/deletes.php
Meldiron: A problem was caused when library called reset, yes. It is now fixed, please revisit.

TorstenDittmann (approved):

PR #3006, opened 2022-03-24
Fix: Network ENV var

Meldiron:
## What does this PR do?

We currently allowed `network` parameter in `Executor.php`, which is a HTTP client to talk to executor. This means, since no one is providing any custom network anyway, it always uses `getEnv` to as default value.

The problem was this ENV var would need to be proivede to `appwrite` and `appwrite-worker-functions` and `appwrite-worker-builds` containers. Christy mentioned this would not be a good solution, instead, we decided to remove option `network` from Executor HTTP server in `executor.php` (and HTTP client in `Executor.php`). Then we used `getEnv` in `appwrite-executor`, so we only need to provide variable there.

## Test Plan

QA must pass.

## Related PRs and Issues

x

Yes


TorstenDittmann:
Why are we removing the network? 👀

Meldiron:
Whoops, forgot to add a description, I made a PR in rush.

Please check description. With that said, we probably can keep the parameter. Should add it back, just move the `getEnv` to proper file?

TorstenDittmann (approved):

app/executor.php
```diff
@@ -235,9 +233,7 @@ function getStorageDevice($root): Device {
                 throw new Exception('Failed to create build container', 500);
             }
 
-            if (!empty($network)) {
-                $orchestration->networkConnect($runtimeId, $network);
-            }
+            $orchestration->networkConnect($runtimeId, App::getEnv('_APP_EXECUTOR_RUNTIME_NETWORK', 'appwrite_runtimes'));
```
eldadfux, line 236: in this case, would it make more sense to use an "Open Runtime" prefix? This part is planned to be an independent docker image under the OPR org.

Meldiron: Renamed to `OPEN_RUNTIMES_NETWORK`. WDYT?

eldadfux (approved):

PR #3024, opened 2022-03-27
project level secrets data structure

lohanidamodar:
## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)


app/config/collections.php
```diff
@@ -2668,15 +2690,15 @@
                 'filters' => [],
             ],
             [
-                '$id' => 'openSSLVersion',
+                '$id' => 'fileSecret',
```
eldadfux, line 2693: Same, we should store a list of keys not a single one.

app/controllers/api/account.php
```diff
@@ -1060,7 +1062,7 @@
             throw new Exception('No valid session found', 404, Exception::USER_SESSION_NOT_FOUND);
         }
 
-        $jwt = new JWT(App::getEnv('_APP_OPENSSL_KEY_V1'), 'HS256', 900, 10); // Instantiate with key, algo, maxAge and leeway.
+        $jwt = new JWT($project->getId() == 'console' ? App::getEnv('_APP_OPENSSL_KEY_V1') : $project->getAttribute('jwtSecret'), 'HS256', 900, 10); // Instantiate with key, algo, maxAge and leeway.
```
eldadfux, line 1065: Let's add a small explanation above this line.

eldadfux: Should we also keep older version to make sure we avoid any race conditions and failed validations during the switch? Not sure if we can ignore this scenario, wdyt?

app/controllers/api/functions.php
```diff
@@ -899,7 +899,7 @@
             }
 
             if(!$current->isEmpty()) {
-                $jwtObj = new JWT(App::getEnv('_APP_OPENSSL_KEY_V1'), 'HS256', 900, 10); // Instantiate with key, algo, maxAge and leeway.
+                $jwtObj = new JWT($project->getId() == 'console' ? App::getEnv('_APP_OPENSSL_KEY_V1') : $project->getAttribute('jwtSecret'), 'HS256', 900, 10); // Instantiate with key, algo, maxAge and leeway.
```
eldadfux, line 902: Should we add JWT as a resource we initialize in one place?

app/controllers/api/projects.php
```diff
@@ -98,6 +99,8 @@
             'keys' => null,
             'domains' => null,
             'auths' => $auths,
+            'databaseSecrets' => [\uniqid() => \bin2hex(OpenSSL::randomPseudoBytes(128))],
```
eldadfux, line 102: Can we abstract this logic in the Auth library?

app/controllers/api/account.php
eldadfux, line 1065: Can we set/cast these attributes (jwtSecret, databaseSecrets) globally on the `console` project object? Might be easier than forcing dev to remember adding this condition.

app/controllers/api/functions.php
eldadfux: as mentioned above, we can also try to cast the console attributes in a higher level to avoid code repetition.

app/controllers/api/storage.php
```diff
@@ -508,9 +508,16 @@
                 if (empty($data)) {
                     $data = $deviceFiles->read($path);
                 }
-                $key = App::getEnv('_APP_OPENSSL_KEY_V1');
+
+                $fileSecret = \bin2hex(OpenSSL::randomPseudoBytes(128));
```
eldadfux, line 512: Same, can we abstract the secret generation?

composer.json
```diff
@@ -45,7 +45,7 @@
         "utopia-php/cache": "0.6.*",
         "utopia-php/cli": "0.12.*",
         "utopia-php/config": "0.2.*",
-        "utopia-php/database": "0.15.*",
+        "utopia-php/database": "dev-feat-filters-upgrade as 0.15.4",
```
eldadfux, line 48: Reminder to release a version.

src/Appwrite/OpenSSL/OpenSSL.php
```diff
@@ -20,6 +20,7 @@ class OpenSSL
      */
     public static function encrypt($data, $method, $key, $options = 0, $iv = '', &$tag = null, $aad = '', $tag_length = 16)
     {
+        var_dump($data);
```
eldadfux, line 23: Leftover.

eldadfux (changes requested):

app/config/collections.php
lohanidamodar: we are just using keys per file, so I don't think we need arrays here? Because if/when we change the key the file should already be encrypted using the new key.

app/controllers/api/account.php
lohanidamodar: I'm not sure? If we don't keep the old version, only thing is that all the existing JWT keys will be invalid right? Is that such a big problem? or am I missing something here?

app/controllers/api/projects.php
lohanidamodar: getting a new secret?

app/controllers/api/account.php
lohanidamodar: may be adding a JWT resource will be a good idea, I'll try that.

app/controllers/api/storage.php
lohanidamodar: will do.

app/controllers/api/account.php
lohanidamodar: As we already have console resource, I updated the console project to also include `jwtSecrets` attribute so that we don't need to check anything else here

app/controllers/api/projects.php
lohanidamodar: Shouldn't this be a part of OpenSSL instead of Auth library? I've put it on OpenSSL library. Do let me know WYT.

PR #3031, opened 2022-03-28
Fix function ENV setup

Meldiron:
## What does this PR do?

By setting variables to `0`, somehow the SWAP warning during function build still comes up. Setting it to nothing resolves the problem.

## Test Plan

Current tests should pass.

- [x] Manual QA with DO 1-click droplet

## Related PRs and Issues

x (reported on Discord)

Yes


christyjacob4 (changes requested):
From our discussion, we need to verify this is actually the fix.
At the moment there is no evidence this is causing the issue.

PR #3039, opened 2022-03-30
Feat: Improved SSL issue&renew

Meldiron:
🚨 CHANGE BASE TO MASTER BEFORE MERGING 🚨

## What does this PR do?

- We now use scheduled queue as non-persistent, with way shorter durations
- We now only have one certificate document per domain. Previously we had more and created inconsistencies in IDs (that only worked because they both pointed to the same SSL path because path only has a domain)
- We now retry to issue certificates that failed (caused by 5XX from letsEnrypt, or API limits) with sane attempts limit
- We now to prioritize domains closer to expiration, making it more ready for scaling
- We now try to properly respect API limit as much as possible, to prevent unnecessary failures. We can still hit the API limit if too many new domains request certificates. No worries tho, future renewal task will fix it
- We continue to renew certs gradually, instead of a bunch at once. This prevents hitting API limits

## Test Plan

TODO: Test properly, as much as I can

## Related PRs and Issues

x

Yes

### DIFF
--- a/app/tasks/ssl.php
+++ b/app/tasks/ssl.php
@@ -15,6 +15,7 @@ $cli
             Make sure your domain points to your server or restart to try again.');
 
         ResqueScheduler::enqueueAt(\time() + 30, 'v1-certificates', 'CertificatesV1', [
+            'type' => 'issue',
             'document' => [],
             'domain' => $domain,
             'validateTarget' => false,
--- a/app/workers/certificates.php
+++ b/app/workers/certificates.php
@@ -57,7 +57,8 @@ class CertificatesV1 extends Worker
         // Refference: https://letsencrypt.org/docs/faq/
         // "Our certificates are valid for 90 days. We recommend automatically renewing your certificates every 60 days."
         $certificateForRenewal = $dbForConsole->findOne("certificates", [
-            new Query('result', Query::TYPE_EQUAL, ['done', 'apiLimit']),
+            new Query('attempts', Query::TYPE_LESSER, [540]), // 3 API limit reset cycles
+            new Query('result', Query::TYPE_EQUAL, ['done', 'letsEncryptError']),
             new Query('updated', Query::TYPE_LESSEREQUAL, [ \time() - 5184000 ]), // - 60 days
         ]);
 
@@ -243,7 +244,9 @@ class CertificatesV1 extends Worker
         } catch(Exception $err) {
             $errorType = 'error';
 
-            // TODO: set errorType to apiLimit, if it is apiLimit
+            if(\str_starts_with($err->getMessage(), 'Failed to issue a certificate with message')) {
+                $errorType = 'letsEncryptError';
+            }
 
             if(empty($certificate)) {
                 $certificate = new Document([
--- a/app/workers/certificates.php
+++ b/app/workers/certificates.php
@@ -18,6 +18,8 @@ Console::success(APP_NAME . ' certificates worker v1 has started');
 
 class CertificatesV1 extends Worker
 {
+    protected static $lastRenewCycle = null;
+
     // Refference: https://letsencrypt.org/docs/rate-limits/
     const LETSENCRYPT_API_LIMIT = 300; // 30 orders
     const LETSENCRYPT_API_LIMIT_INTERVAL = 10800; // 3 hours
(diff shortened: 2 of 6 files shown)


app/tasks/ssl.php
```diff
@@ -15,6 +15,7 @@
             Make sure your domain points to your server or restart to try again.');
 
         ResqueScheduler::enqueueAt(\time() + 30, 'v1-certificates', 'CertificatesV1', [
+            'type' => 'issue',
```
eldadfux, line 18: You can use the new constant

app/workers/certificates.php
```diff
@@ -1,7 +1,9 @@
 <?php
 
+use Appwrite\Event\Event;
 use Appwrite\Network\Validator\CNAME;
 use Appwrite\Resque\Worker;
+use PHPUnit\Framework\Constraint\Operator;
```
eldadfux, line 6: This seems wrong.

app/workers/certificates.php
```diff
@@ -108,90 +171,140 @@ public function run(): void
...
-        if ($exit !== 0) {
-            throw new Exception('Failed to issue a certificate with message: ' . $stderr);
-        }
-
-        $path = APP_STORAGE_CERTIFICATES . '/' . $domain->get();
-
-        if (!\is_readable($path)) {
-            if (!\mkdir($path, 0755, true)) {
-                throw new Exception('Failed to create path...');
+        try {
```
eldadfux, line 174: Big try-catch block are not healthy. We have an error handler for workers, can we use it?

app/workers/certificates.php
```diff
@@ -26,8 +34,63 @@ public function init(): void
...
+
+        if($certificatesCreated >= 250) { // Limit is 300, but we keep 50 to stay safe with new cert generation
+            Console::log('Renew cycle skipped due to potentialy exceeded API limits.');
+        } else {
+            // We will renew all certs older than 60 days
+            // Refference: https://letsencrypt.org/docs/faq/
+            // "Our certificates are valid for 90 days. We recommend automatically renewing your certificates every 60 days."
+            $certificateForRenewal = $dbForConsole->findOne("certificates", [
+                new Query('attempts', Query::TYPE_LESSER, [540]), // 3 API limit reset cycles
+                new Query('result', Query::TYPE_EQUAL, ['done', 'letsEncryptError']),
```
eldadfux, line 76:
```suggestion
                new Query('result', Query::TYPE_EQUAL, ['done', 'letsEncryptError']),
```

This is a bit messy, it's not clear what type of data we expect in this result param, or how do we use this information later.

app/workers/certificates.php
```diff
@@ -26,8 +34,63 @@ public function init(): void
...
+
+            if($certificateForRenewal) {
+                $this->issueCertificate([
+                    'domain' => $certificateForRenewal->getAttribute('domain'),
+                ]);
+            }
+        }
+
+        // When done, schedule again in 1 minute. This creates makes the function recursive.
+        ResqueScheduler::enqueueIn(60, Event::CERTIFICATES_QUEUE_NAME, Event::CERTIFICATES_CLASS_NAME, [
```
eldadfux, line 88: The main idea of this task was to decouple us from the `ResqueScheduler` to avoid trusting Redis for persistency. This could also potentially lead to to multiple infinite loops in some edge cases. Also why should we trigger renewal every 1 minute? The idea was to check for renewals from the maintenance container every day for certificates that were last issues +60 days ago.

app/workers/certificates.php
```diff
@@ -108,90 +171,140 @@ public function run(): void
...
-        Authorization::reset();
     }
 
     public function shutdown(): void
     {
     }
 }
+
+// We trigger once, to start recursive renewal loop. 5 second delay to stay safe with startup
+ResqueScheduler::enqueueIn(5, Event::CERTIFICATES_QUEUE_NAME, Event::CERTIFICATES_CLASS_NAME, [
```
eldadfux, line 308: Not sure why are we re-scheduling renewals again? Also, it's a bit hard to track any changes done inside the try-catch block, I would love to revalidate them with you.

eldadfux (changes requested):

Meldiron:
Closing PR due to a decision to refactor the renewal and issuing logic in a totally different way. New PR can be found here: https://github.com/appwrite/appwrite/pull/3074

Issue #2959, opened 2022-03-17
🐛 Bug Report: Old cached document data is getting returned instead of latest data when document ID is uppercased and original ID was in lowercase


ameerfayiz:
### 👟 Reproduction steps

Created a document from console with any lowercased ID
Called getDocument from dart server sdk using uppercased ID , got the correct data.
Then, i updated same document from console.
Again, called getDocument from dart server sdk using uppercased ID , i got old unmodified data.

### 👍 Expected behavior

document id should be either case sensitive or not case sensitive. 

### 👎 Actual Behavior

old cached data is returning from server when document id is given as uppercased and the original document id was lower cased instead of NO document found or with latest data
but, the updated data is returning from server if lowercased document id is given.

### 🎲 Appwrite version

Version 0.13.x

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

Meldiron:
The problem is easily reproducible. Ill discuss case sensitivity internally and create a PR for this problem.

stnguyen90:
Related PR: https://github.com/utopia-php/cache/pull/17

Meldiron:
This issue came from Redis not being case-insensitive. Once PR above is marked and released, Appwrite can be upgraded to a new version and mark cache as case-insensitive. That will solve the problem.

I'll keep PR posted about the progress.

eldadfux:
This feature is now released and available as part of Appwrite 0.14. More on the new release here: https://github.com/appwrite/appwrite/discussions/3240

PR #3041, opened 2022-03-30
Feat: Cache library upgrade

Meldiron:
## What does this PR do?

Upgrades cache and database libraries, both related to ID case sensitivity issues.

## Test Plan

Unit tests on library-level + current tests must pass.

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/2959

Yes


composer.json
```diff
@@ -42,10 +42,10 @@
         "utopia-php/abuse": "0.7.*",
         "utopia-php/analytics": "0.2.*",
         "utopia-php/audit": "0.8.*",
-        "utopia-php/cache": "0.4.*",
+        "utopia-php/cache": "0.5.*",
         "utopia-php/cli": "0.12.*",
         "utopia-php/config": "0.2.*",
-        "utopia-php/database": "0.15.*",
+        "utopia-php/database": "dev-feat-cache-upgrate as 0.15.1",
```
christyjacob4, line 48: Reminder to use the correct library version 👍

app/init.php
```diff
@@ -135,6 +135,8 @@
 // Response related
 const MAX_OUTPUT_CHUNK_SIZE = 2*1024*1024; // 2MB
 
+Cache::setCaseSensitivity(false);
```
christyjacob4, line 138: I had a similar comment in here 
https://github.com/utopia-php/database/pull/117

If majority of use cases require the caseSensitivity to be turned off, lets keep false as the default in 
https://github.com/utopia-php/cache/blob/02ffbe165f9632e05519eb75a2ce32deb19638a0/src/Cache/Cache.php#L15

christyjacob4 (changes requested):
PR looks good 👍 
Just one comment

composer.json
Meldiron: Updated

eldadfux (approved):

Issue #3037, opened 2022-03-30
🐛 Bug Report: Error when uploading small mp4 video file


istornz:
### 👟 Reproduction steps

- Create a new bucket with settings:
  - Maximum File Size: **90000000**
  - Enabled: **Yes**
  - Encryption: **Yes**
  - Anti Virus: **Yes**
  - Allowed File Extensions: **mp4**
  - Read Access: **role:all**
- By using web console, or API, just upload a **.mp4 video** of 2.8Mo

Appwrite logs show:
```
[Error] Message: Error bucket maximum file size is larger than _APP_STORAGE_LIMIT
```

FYI, .env file was not modified & show for _APP_STORAGE_LIMIT
```
30000000
```

I tried to disable AV but nothing works...

### 👍 Expected behavior

Upload the file & show success message

### 👎 Actual Behavior

File was not uploaded & show failed

### 🎲 Appwrite version

Version 0.13.x

### 💻 Operating system

Linux

### 🧱 Your Environment

Ubuntu

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

Meldiron:
The issue here is that on the bucket you tried to configure a 90MB limit, while the server only allows 30MB. That resulted in an unexpected error, as the controller was checking 90MB max size on the 30MB instance limit.

A quick solution is to set the size on the bucket to less or equal to the number set as a limit in `.env`.

I will work on a PR to address this, not allowing to configure such a size limit on bucket that would be bigger than instance limit.

istornz:
@Meldiron thanks for your answer, I can confirm that changing in ```.env``` file the limit to 90Mb works 🤗

Issue #3036, opened 2022-03-30
🐛 Bug Report: Storage Size Limit Error 


rlee1990:
### 👟 Reproduction steps

Add a file larger than the storage limit size.
I am using flutter.

### 👍 Expected behavior

An error should be sent back to the client with a proper error.

### 👎 Actual Behavior

I just receive the below error:
`AppwriteException: general_server_error, Server Error (500)`
The on appwriteexception does not even get called.

### 🎲 Appwrite version

Version 0.13.x

### 💻 Operating system

MacOS

### 🧱 Your Environment

None

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

Meldiron:
@lohanidamodar Can you take this one? It's flutter-related.

lohanidamodar:
@rlee1990 there might be other errors that causing the 500 error. Can you check docker console logs for appwrite?
by running `docker-compose logs appwrite` right after you make a request to upload a large file,  from your terminal where appwrite is installed. You will be able to see the actual error on the console and that will help us figure out the issue

rlee1990:
@lohanidamodar I did this was the issue. I even have the information in the storage discord channel. Was raising this issue so it could be looked into.

lohanidamodar:
@rlee1990 Thanks, that does make sense, when your `_APP_STORAGE_LIMIT` is less than your **Bucket File Size limit**, you do get 500. We will explore more what error code and status might make sense here, without exposing much information to the client. Thank you for the issue.

rlee1990:
@lohanidamodar Let me test that later today and I will let you know.

stnguyen90:
@rlee1990, is this still a problem or can it be closed now?

PR #3047, opened 2022-03-31
Fix: Bucket Size Update

Meldiron:
## What does this PR do?

If you set `_APP_STORAGE_LIMIT` to 30MB and try to set a 100MB limit on the bucket, it will successfully update. Problem is, later you will get a 500 error when trying to upload a file.

This PR will now properly check the validity of the maximum size on the bucket, and throw an error if the instance limit is exceeded.

## Test Plan

Added a new test for this scenario.

- [x] Plus manual QA:

https://user-images.githubusercontent.com/19310830/161063007-84f17b5f-7b5d-47e5-80d2-a1bbddd371d5.mp4

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/3037
https://github.com/appwrite/appwrite/issues/3036

Yes


app/controllers/api/storage.php
```diff
@@ -241,6 +241,11 @@
         /** @var Appwrite\Event\Event $audits */
         /** @var Appwrite\Stats\Stats $usage */
 
+        // Maximum allowed size validation
+        if ($maximumFileSize > (int) App::getEnv('_APP_STORAGE_LIMIT', 0)) {
+            throw new Exception('Bucket maximum file size is larger than _APP_STORAGE_LIMIT.', 400, Exception::ATTRIBUTE_VALUE_INVALID);
```
lohanidamodar, line 246:
```suggestion
            throw new Exception('Bucket maximum file size allowed is larger than _APP_STORAGE_LIMIT.', 400, Exception::ATTRIBUTE_VALUE_INVALID);
```
may be this makes more sense?

app/controllers/api/storage.php
```diff
@@ -421,7 +431,7 @@
 
         $maximumFileSize = $bucket->getAttribute('maximumFileSize', 0);
         if ($maximumFileSize > (int) App::getEnv('_APP_STORAGE_LIMIT', 0)) {
-            throw new Exception('Error bucket maximum file size is larger than _APP_STORAGE_LIMIT', 500, Exception::GENERAL_SERVER_ERROR);
+            throw new Exception('Error bucket maximum file size is larger than _APP_STORAGE_LIMIT', 400, Exception::STORAGE_INVALID_FILE_SIZE);
```
christyjacob4, line 434: Why are we using a different error code here ? Compared to `ATTRIBUTE_VALUE_INVALID ` in the previous cases ?

christyjacob4 (changes requested):
Just one comment 👍  rest looks good

app/controllers/api/storage.php
Meldiron: One is client-side, and one is server-side endpoint. The client will get INVALID_FILE_SIZE, just like he would get if his file was bigger than the bucket setting.

The server (a console user) gets invalid input because his action is to update the value.

Does it make sense?

app/controllers/api/storage.php
```diff
@@ -67,6 +67,11 @@
         /** @var Appwrite\Event\Event $audits */
         /** @var Appwrite\Stats\Stats $usage */
 
+        // Maximum allowed size validation
+        if ($maximumFileSize > (int) App::getEnv('_APP_STORAGE_LIMIT', 0)) {
+            throw new Exception('Bucket maximum file size is larger than _APP_STORAGE_LIMIT.', 400, Exception::ATTRIBUTE_VALUE_INVALID);
+        }
+
```
christyjacob4, line 74: Why aren't we using the range validator to do this for us like we do in functions ? 
https://github.com/appwrite/appwrite/blob/5ab38c0f9889e005b012ea267488d030f7002396/app/controllers/api/functions.php#L52

app/controllers/api/storage.php
```diff
@@ -241,6 +246,11 @@
         /** @var Appwrite\Event\Event $audits */
         /** @var Appwrite\Stats\Stats $usage */
 
+        // Maximum allowed size validation
+        if ($maximumFileSize > (int) App::getEnv('_APP_STORAGE_LIMIT', 0)) {
+            throw new Exception('Bucket maximum file size allowed is larger than _APP_STORAGE_LIMIT.', 400, Exception::ATTRIBUTE_VALUE_INVALID);
+        }
+
```
christyjacob4, line 253: Why aren't we using the range validator to do this for us like we do in functions ? In that case, Utopia automatically throws a 400 error with type `GENERAL_ARGUMENT_INVALID` 

https://github.com/appwrite/appwrite/blob/5ab38c0f9889e005b012ea267488d030f7002396/app/controllers/api/functions.php#L52

app/controllers/api/storage.php
christyjacob4: not really, at the end of the day, $maximumFileSize is a parameter passed through the API and needs to use the `GENERAL_ARGUMENT_INVALID` error and not even  `ATTRIBUTE_VALUE_INVALID `

`ATTRIBUTE_VALUE_INVALID ` is meant to be used for database attributes . 
We're dealing with Utopia parameters here 😄 

Also refer to my comments above about using the `Range` Validator to do this check

christyjacob4: Okay. In the previous case, the parameter is passed through the API, here we fetch it from the Database 👍

christyjacob4: Error code makes sense, I would rewrite the error message a bit.

christyjacob4: No still, 400 series errors are for user errors and it may not be appropriate to use it in the context of the `createFile` endpoint where `$maximumFileSize` is **not passed** to it.   

I still think this is a server error to be treated with 5xx error codes. 

In the previous instances, where `maximumFileSize` is passed as a parameter to the endpoint, it makes sense to use a 4xx error code. 

So, I would stick to a 5xx code but the description looks like it could use some edits.

christyjacob4, line 434:
```suggestion
            throw new Exception('Maximum bucket file size is larger than _APP_STORAGE_LIMIT', 500, Exception:: GENERAL_SERVER_ERROR);
```

christyjacob4 (changes requested):
Please address the comments

app/controllers/api/storage.php
Meldiron: This is now 500 server error, as suggested below.

TorstenDittmann (approved):

PR #3074, opened 2022-04-10
Feat: Refactored SSL genration

Meldiron:
## What does this PR do?

a LOT!

- Renew is now properly controlled by maintenance cycle instead of relying on Redis
- We now double-check renew date with certificate file
- We now properly retry certificate generation
- We now properly respect 'attempts' limit

- [ ] SSL command now runs without 30 seconds delay
- [ ] SSL command now accepts optional attribute for domain that needs to be renewed
- [ ] SSL command skips renew date validation
- [ ] If issue fails, we schedule a job to send email to administrator

## Test Plan

- [ ] Manual QA

## Related PRs and Issues

x

Yes

### DIFF
--- a/app/cli.php
+++ b/app/cli.php
@@ -5,6 +5,9 @@ require_once __DIR__.'/controllers/general.php';
 use Utopia\App;
 use Utopia\CLI\CLI;
 use Utopia\CLI\Console;
+use Utopia\Database\Validator\Authorization;
+
+Authorization::disable();
 
 $cli = new CLI();
 
--- a/app/tasks/maintenance.php
+++ b/app/tasks/maintenance.php
@@ -1,15 +1,58 @@
 <?php
 
 global $cli;
+global $register;
 
 use Appwrite\Event\Event;
 use Utopia\App;
+use Utopia\Cache\Cache;
 use Utopia\CLI\Console;
+use Utopia\Database\Adapter\MariaDB;
+use Utopia\Database\Database;
+use Utopia\Registry\Registry;
+use Utopia\Cache\Adapter\Redis as RedisCache;
+use Utopia\Database\Query;
+
+function getDatabase(Registry &$register)
+{
+    $attempts = 0;
+
+    do {
+        try {
+            $attempts++;
+
+            $db = $register->get('dbPool')->get();
+            $redis = $register->get('redisPool')->get();
+
+            $cache = new Cache(new RedisCache($redis));
+            $database = new Database(new MariaDB($db), $cache);
+            $database->setDefaultDatabase(App::getEnv('_APP_DB_SCHEMA', 'appwrite'));
+            $database->setNamespace('_console');
+
+            break; // leave loop if successful
+        } catch(\Exception $e) {
+            Console::warning("Database not ready. Retrying connection ({$attempts})...");
+            if ($attempts >= DATABASE_RECONNECT_MAX_ATTEMPTS) {
+                throw new \Exception('Failed to connect to database: '. $e->getMessage());
+            }
+            sleep(DATABASE_RECONNECT_SLEEP);
+        }
+    } while ($attempts < DATABASE_RECONNECT_MAX_ATTEMPTS);
+
+    return [
+        $database,
+        function () use ($register, $db, $redis) {
+            $register->get('dbPool')->put($db);
+            $register->get('redisPool')->put($redis);
+        }
+    ];
+
+};
 
 $cli
     ->task('maintenance')
     ->desc('Schedules maintenance tasks and publishes them to resque')
-    ->action(function () {
+    ->action(function () use ($register) {
         Console::title('Maintenance V1');
         Console::success(APP_NAME.' maintenance process v1 has started');
 
@@ -54,6 +97,29 @@ $cli
             ]);
         }
 
+        function renewCertificates($dbForConsole)
+        {
+            $time = date('d-m-Y H:i:s', time());
+            /** @var Utopia\Database\Database $dbForConsole */
+
+            $certificates = $dbForConsole->find('certificates', [
+                new Query('attempts', Query::TYPE_LESSER, [5]), // Maximum 5 attempts
+                new Query('renewDate', Query::TYPE_LESSEREQUAL, [\time()]) // includes 60 days cooldown (we have 30 days to renew)
+            ], 300); // Limit 300 comes from LetsEncrypt (orders per 3 hours)
+
+            if(\count($certificates) > 0) {
+                Console::info("[{$time}] Found " . \count($certificates) . " certificates for renewal, scheduling jobs.");
+
+                foreach ($certificates as $certificate) {
+                    Resque::enqueue(Event::CERTIFICATES_QUEUE_NAME, Event::CERTIFICATES_CLASS_NAME, [
+                        'domain' => $certificate->getAttribute('domain'),
+                    ]);
+                }
+            } else {
+                Console::info("[{$time}] No certificates for renewal.");
+            }
+        }
+
         // # of days in seconds (1 day = 86400s)
         $interval = (int) App::getEnv('_APP_MAINTENANCE_INTERVAL', '86400');
         $executionLogsRetention = (int) App::getEnv('_APP_MAINTENANCE_RETENTION_EXECUTION', '1209600');
@@ -62,13 +128,25 @@ $cli
         $usageStatsRetention30m = (int) App::getEnv('_APP_MAINTENANCE_RETENTION_USAGE_30M', '129600');//36 hours
         $usageStatsRetention1d = (int) App::getEnv('_APP_MAINTENANCE_RETENTION_USAGE_1D', '8640000'); // 100 days
 
-        Console::loop(function() use ($interval, $executionLogsRetention, $abuseLogsRetention, $auditLogRetention, $usageStatsRetention30m, $usageStatsRetention1d) {
-            $time = date('d-m-Y H:i:s', time());
-            Console::info("[{$time}] Notifying deletes workers every {$interval} seconds");
-            notifyDeleteExecutionLogs($executionLogsRetention);
-            notifyDeleteAbuseLogs($abuseLogsRetention);
-            notifyDeleteAuditLogs($auditLogRetention);
-            notifyDeleteUsageStats($usageStatsRetention30m, $usageStatsRetention1d);
-            notifyDeleteConnections();
+        Console::loop(function() use ($register, $interval, $executionLogsRetention, $abuseLogsRetention, $auditLogRetention, $usageStatsRetention30m, $usageStatsRetention1d) { 
+            go(function () use ($register, $interval, $executionLogsRetention, $abuseLogsRetention, $auditLogRetention, $usageStatsRetention30m, $usageStatsRetention1d) {
+                try {
+                    [$database, $returnDatabase] = getDatabase($register, '_console');
+    
+                    $time = date('d-m-Y H:i:s', time());
+                    Console::info("[{$time}] Notifying deletes workers every {$interval} seconds");
+                    notifyDeleteExecutionLogs($executionLogsRetention);
+                    notifyDeleteAbuseLogs($abuseLogsRetention);
+                    notifyDeleteAuditLogs($auditLogRetention);
+                    notifyDeleteUsageStats($usageStatsRetention30m, $usageStatsRetention1d);
+                    notifyDeleteConnections();
+    
+                    renewCertificates($database);
+                } catch (\Throwable $th) {
+                    throw $th;
+                } finally {
+                    call_user_func($returnDatabase);
+                }
+            });
         }, $interval);
     });
--- a/app/tasks/ssl.php
+++ b/app/tasks/ssl.php
@@ -11,13 +11,15 @@ $cli
     ->action(function () {
         $domain = App::getEnv('_APP_DOMAIN', '');
 
-        Console::log('Issue a TLS certificate for master domain ('.$domain.') in 30 seconds.
+        // TODO: Instead of waiting, let's ping Traefik. If responds, we can schedule instantly
+        // TODO: Add support for argument (domain)
+
+        Console::log('Issue a TLS certificate for master domain ('.$domain.') in 2 seconds.
             Make sure your domain points to your server or restart to try again.');
 
-        ResqueScheduler::enqueueAt(\time() + 30, 'v1-certificates', 'CertificatesV1', [
-            'document' => [],
+        // Const for types not available here
+        ResqueScheduler::enqueueAt(\time() + 2, 'v1-certificates', 'CertificatesV1', [
             'domain' => $domain,
-            'validateTarget' => false,
-            'validateCNAME' => false,
+            'skipRenewCheck' => true // TODO: Discuss this behabiour. true? false? parameter? How do we document it?
         ]);
     });
--- a/app/workers/certificates.php
+++ b/app/workers/certificates.php
@@ -1,7 +1,9 @@
 <?php
 
+use Appwrite\Event\Event;
 use Appwrite\Network\Validator\CNAME;
 use Appwrite\Resque\Worker;
+use Appwrite\Exception\Certificate as ExceptionCertificate;
 use Utopia\App;
 use Utopia\CLI\Console;
 use Utopia\Database\Document;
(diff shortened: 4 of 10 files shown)


app/tasks/maintenance.php
```diff
@@ -54,6 +97,29 @@ function notifyDeleteConnections()
...
 
+        function renewCertificates($dbForConsole)
+        {
+            $time = date('d-m-Y H:i:s', time());
+            /** @var Utopia\Database\Database $dbForConsole */
+
+            $certificates = $dbForConsole->find('certificates', [
+                new Query('attempts', Query::TYPE_LESSER, [5]), // Maximum 5 attempts
+                new Query('renewDate', Query::TYPE_LESSEREQUAL, [\time()]) // includes 60 days cooldown (we have 30 days to renew)
+            ], 300); // Limit 300 comes from LetsEncrypt (orders per 3 hours)
```
eldadfux, line 108: What do you mean by `orders per 3 hours`?

app/tasks/maintenance.php
```diff
@@ -62,13 +128,25 @@ function notifyDeleteConnections()
...
-            notifyDeleteAuditLogs($auditLogRetention);
-            notifyDeleteUsageStats($usageStatsRetention30m, $usageStatsRetention1d);
-            notifyDeleteConnections();
+        Console::loop(function() use ($register, $interval, $executionLogsRetention, $abuseLogsRetention, $auditLogRetention, $usageStatsRetention30m, $usageStatsRetention1d) { 
+            go(function () use ($register, $interval, $executionLogsRetention, $abuseLogsRetention, $auditLogRetention, $usageStatsRetention30m, $usageStatsRetention1d) {
+                try {
+                    [$database, $returnDatabase] = getDatabase($register, '_console');
+    
+                    $time = date('d-m-Y H:i:s', time());
+                    Console::info("[{$time}] Notifying deletes workers every {$interval} seconds");
```
eldadfux, line 137: We should change this log, not only delete workers are notified.

src/Appwrite/Exception/Certificate.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace Appwrite\Exception;
+
+class Certificate extends \Exception
+{
+    private $incrementAttempts = false;
+
+    // Only set incrementAttempts to TRUE if exception occured AFTER certbot command execution
+    public function __construct(string $message, bool $incrementAttempts = false)
```
eldadfux, line 10: This is not a very common way for us to use an exception.

app/workers/certificates.php
```diff
@@ -26,169 +28,224 @@ public function init(): void
...
-            $target = new Domain(App::getEnv('_APP_DOMAIN_TARGET', ''));
+        $dbForConsole = $this->getConsoleDB();
 
-            if(!$target->isKnown() || $target->isTest()) {
-                throw new Exception('Unreachable CNAME target ('.$target->get().'), please use a domain with a public suffix.');
+        $certificate = new Document();
+
+        try {
+            /**
+             * TODO: Update
```
eldadfux, line 39: I think this steps are outdated now.

app/workers/certificates.php
```diff
@@ -26,169 +28,224 @@ public function init(): void
...
+    
+            // Get attributes
+            $domain = $this->args['domain']; // String of domain (hostname)
+            $domain = new Domain((!empty($domain)) ? $domain : '');
+
+            $certificate->setAttribute('domain', $domain->get());
+    
+            $skipRenewCheck = $this->args['skipRenewCheck'] ?? false; // If true, we won't double-check expiry from cert file
+    
+            $mainDomain = null; // ENV or first ever visited domain
```
eldadfux, line 60: Why do we have this main domain logic here? Can we simplify the worker and treat all domains equally?

app/workers/certificates.php
```diff
@@ -26,169 +28,224 @@ public function init(): void
...
-        }
-
-        $path = APP_STORAGE_CERTIFICATES . '/' . $domain->get();
-
-        if (!\is_readable($path)) {
-            if (!\mkdir($path, 0755, true)) {
-                throw new Exception('Failed to create path...');
+    
+            if (empty($domain->get())) {
+                throw new ExceptionCertificate('Missing certificate domain.');
```
eldadfux, line 75: We should use a regular exception, and not trust the exception to handle the count for us. Resque is restarting the PHP script every execution so you can't trust the script memory.

Even with the new library and multiple threads the count might get mixed between different domains.

You should rely on the certificate document for counting the attempts and reseting them on every successful generation.

app/workers/certificates.php
```diff
@@ -26,169 +28,224 @@ public function init(): void
...
+            // If certificate exists already, double-check expiry date
+            // If asked to skip, we won't
+            $certPath = APP_STORAGE_CERTIFICATES . '/' . $domain->get() . '/cert.pem';
+            if (!$skipRenewCheck && \file_exists($certPath)) {
+                $validTo = null;
+
+                try {
+                    $certData = openssl_x509_parse(file_get_contents($certPath));
+    
+                    $validTo = $certData['validTo_time_t'];
```
eldadfux, line 111: Missing null safety

app/workers/certificates.php
```diff
@@ -26,169 +28,224 @@ public function init(): void
...
+            $certificate->setAttribute('renewDate', $validTo - $expiryInAdvance);
+
+            // All went well at this point 🥳
+            
+            // Reset attempts count for next renwal
+            $certificate->setAttribute('attempts', 0);
+
+            // Mark issue date
+            $certificate->setAttribute('issueDate', \time());
+        } catch(ExceptionCertificate $e) {
```
eldadfux, line 212: We have introduced a proper error handler for all workers to avoid big try-catch blocks, let's use it.
https://github.com/appwrite/appwrite/blob/37c6fd75d8c72c385ed4f719466325beaf1a4d29/src/Appwrite/Resque/Worker.php#L146

app/workers/certificates.php
```diff
@@ -26,169 +28,224 @@ public function init(): void
...
+            // These exceptions are expected if renew shouldn't or can't happen
+
+            // Add exception as log into certificate
+            $certificate->setAttribute('log', $e->getMessage());
+
+            $attempt = $certificate->getAttribute('attempts', 0);
+            $attempt++;
+            
+            // Save increased attempts count if requested by exception
+            if($e->getIncrementAttempts()) {
```
eldadfux, line 222: As mentioned in the comment above, we shouldn't rely on the exception to track # of failed attempts.

app/workers/certificates.php
```diff
@@ -26,169 +28,224 @@ public function init(): void
...
+                $certificate = $dbForConsole->createDocument('certificates', $certificate);
+            }
 
-        ResqueScheduler::enqueueAt($renew + $safety, 'v1-certificates', 'CertificatesV1', [
-            'document' => [],
-            'domain' => $domain->get(),
-            'validateTarget' => $validateTarget,
-            'validateCNAME' => $validateCNAME,
-        ]);  // Async task rescheduale
+            // Update domains with new certificate ID
```
eldadfux, line 243: Are we missing something here? why do we need `$certificateId`?

eldadfux (changes requested):

app/cli.php
```diff
@@ -5,6 +5,9 @@
 use Utopia\App;
 use Utopia\CLI\CLI;
 use Utopia\CLI\Console;
+use Utopia\Database\Validator\Authorization;
+
+Authorization::disable();
```
christyjacob4, line 10: leftover?

app/tasks/ssl.php
```diff
@@ -11,13 +11,15 @@
...
+        // TODO: Instead of waiting, let's ping Traefik. If responds, we can schedule instantly
+        // TODO: Add support for argument (domain)
+
+        Console::log('Issue a TLS certificate for master domain ('.$domain.') in 2 seconds.
             Make sure your domain points to your server or restart to try again.');
 
-        ResqueScheduler::enqueueAt(\time() + 30, 'v1-certificates', 'CertificatesV1', [
-            'document' => [],
+        // Const for types not available here
+        ResqueScheduler::enqueueAt(\time() + 2, 'v1-certificates', 'CertificatesV1', [
```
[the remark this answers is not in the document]
Meldiron: Yes, plan is to add HTTP ping to check if server is up. If so, we don't need to delay anymore.

app/tasks/ssl.php
```diff
@@ -11,13 +11,15 @@
...
             'domain' => $domain,
-            'validateTarget' => false,
-            'validateCNAME' => false,
+            'skipRenewCheck' => true // TODO: Discuss this behabiour. true? false? parameter? How do we document it?
```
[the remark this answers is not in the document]
Meldiron: This is mainly a developer experience concern. If I directly run command to renew certificate.. Should we forcefully renew it, or tell him "Renew not required"?

The idea is if anything goes wrong, this command would ALWAYS renew a certificate, no matter what... It would be a quick fix for devs, and they can, later on, check logs to prevent issues in future.

(that was my thinking when implementing it)

app/tasks/maintenance.php
Meldiron: LetsEncrypt has a limit of 300 orders per 3 hours. The order I believe means successful certificate generation for a domain that does need a regeneration. The limit is different for the domain that doesn't need it, but that's none of our concerns.

If we attempt to renew more than 300 domains (that we know actually need renewal), we would be hitting API limit anyway. Limit 300 makes sense to me, or even, maybe set it to just 200. That way other work can be also done within the maintenance hour. If we renew use up all 300 orders, if user adds new custom domain, it would always fail due to API limit.

src/Appwrite/Exception/Certificate.php
Meldiron: What's the better way? I need to differentiate between expected error, and any unexpected one. Unexpected should be logged into container, while expected should be logged into database.

app/workers/certificates.php
Meldiron: We cant be looking for CNAME on the main domain because there you have A record.

app/workers/certificates.php
Meldiron: I dont think we can use it. Inside error(), we only have exception, not certificateID or other stuff. Should we pass these inside exception?

app/workers/certificates.php
Meldiron: Missing TODO. We need to update all domains to have `certificateId`. Required for proper UI functionality of custom domains.

app/cli.php
Meldiron: This script includes script from `maintenance.php`, which needs to read from a database. Should this be done in maintanance.php? I thought it would be okay to disable it for all CLI stuff, as CLI does not need authorization.

app/tasks/maintenance.php
Meldiron: Updated.

app/tasks/ssl.php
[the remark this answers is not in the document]
Meldiron: Ping implemented

app/workers/certificates.php
Meldiron: Updated

Meldiron: Updated

src/Appwrite/Exception/Certificate.php
Meldiron: I removed increment attempts parameter from exception.

app/workers/certificates.php
Meldiron: Added in 2 places

app/workers/certificates.php
Meldiron: We no longer take this from exception, we now increment count with all failed attempts.

app/workers/certificates.php
Meldiron: Implemented.

app/init.php
```diff
@@ -127,6 +127,7 @@
 const MAIL_TYPE_MAGIC_SESSION = 'magicSession';
 const MAIL_TYPE_RECOVERY = 'recovery';
 const MAIL_TYPE_INVITATION = 'invitation';
+const MAIL_TYPE_CERTIFICATE = 'certificat';
```
eldadfux, line 130:
```suggestion
const MAIL_TYPE_CERTIFICATE = 'certificate';
```

app/tasks/ssl.php
```diff
@@ -2,22 +2,35 @@
...
 $cli
     ->task('ssl')
     ->desc('Validate server certificates')
-    ->action(function () {
-        $domain = App::getEnv('_APP_DOMAIN', '');
+    ->param('domain', App::getEnv('_APP_DOMAIN', ''), new Hostname(), 'Domain to generate certificate for. If empty, main domain will be used.', true)
+    ->action(function ($domain) {
+        // HTTTP ping to check if domain is Appwrite server
+        $ch = \curl_init();
+        \curl_setopt($ch, CURLOPT_URL, 'http://appwrite/manifest.json');
```
TorstenDittmann, line 17: For what reason we do this?

app/tasks/maintenance.php
```diff
@@ -1,15 +1,58 @@
...
 use Utopia\App;
+use Utopia\Cache\Cache;
 use Utopia\CLI\Console;
+use Utopia\Database\Adapter\MariaDB;
+use Utopia\Database\Database;
+use Utopia\Registry\Registry;
+use Utopia\Cache\Adapter\Redis as RedisCache;
+use Utopia\Database\Query;
+
+function getDatabase(Registry &$register)
```
eldadfux, line 16: Let's be more verbose and call this function `getConsoleDB`

app/tasks/maintenance.php
```diff
@@ -54,6 +97,29 @@ function notifyDeleteConnections()
             ]);
         }
 
+        function renewCertificates($dbForConsole)
+        {
+            $time = date('d-m-Y H:i:s', time());
+            /** @var Utopia\Database\Database $dbForConsole */
+
+            $certificates = $dbForConsole->find('certificates', [
+                new Query('attempts', Query::TYPE_LESSER, [5]), // Maximum 5 attempts
```
TorstenDittmann, line 106: Should this be `Query::TYPE_LESSEREQUAL`?

app/tasks/ssl.php
```diff
@@ -2,22 +2,35 @@
...
 use Utopia\App;
 use Utopia\CLI\Console;
+use Utopia\Validator\Hostname;
 
 $cli
     ->task('ssl')
     ->desc('Validate server certificates')
-    ->action(function () {
-        $domain = App::getEnv('_APP_DOMAIN', '');
+    ->param('domain', App::getEnv('_APP_DOMAIN', ''), new Hostname(), 'Domain to generate certificate for. If empty, main domain will be used.', true)
```
eldadfux, line 13: @Meldiron are we documenting this command anywhere? It actually quite useful.

app/workers/certificates.php
eldadfux: Would it make sense to have it stored in a worker `$this->certificate` attribute?

app/tasks/ssl.php
Meldiron: There was 30s delay previously, to ensure the HTTP server started properly (the first start can take some time).

This pings the Appwrite container HTTP server to ensure it can properly respond. This means it should be also able to properly respond to the HTTP LetsEncrypt challenge in a few seconds, if we schedule a task.

app/tasks/ssl.php
Meldiron: I created task to update these docs to include this tip: https://appwrite.io/docs/certificates

app/tasks/ssl.php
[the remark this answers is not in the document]
Meldiron: We decided to skip the check from CLI. This is an administration tool and should do exactly what it does, no unnecessary checks. We risk someone hit limit of 5 requests per week per successful domain, but it would be i responsible usage of the command. This will be all documented in docs.

app/tasks/ssl.php
Meldiron: Addressed in PR: https://github.com/appwrite/docs/pull/167
Mentioned in PR description, closing conversation.

app/realtime.php
```diff
@@ -194,7 +194,7 @@ function () use ($register, $db, $redis) {
         }
 
         try {
-            [$database, $returnDatabase] = getDatabase($register, '_console');
+            [$database, $returnDatabase] = getConsoleDB($register, '_console');
```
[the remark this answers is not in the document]
Meldiron: Eldad suggested renaming it in my PR. Since I copied it from realtime, I also renamed it there for consistancy.

app/workers/certificates.php
Meldiron: Count has been refactored.

app/tasks/ssl.php
Meldiron: Removed check.

app/workers/certificates.php
Meldiron: I don't think this is possible due to following reasons:

- Error handlers are registered "forever".. And there isn't a function in worker that would run only once.. So we need to implement unregister logic, and store error handlers under some key
- Error callbacks are stored on static variable. If we run 2 actions at the same time, 1 action will trigger errorCallback for second too. We would need to make Worker aware of some action ID and make `error()` not static
- Finally block would not be possible. We would need duplicate code, or one function at the top

A refactor of error handling in worker would be required before making this change.

app/realtime.php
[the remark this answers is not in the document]
TorstenDittmann: Sure in the worker, but here in realtime it will not work anymore like this 👍🏻

TorstenDittmann (changes requested):

Issue #3082, opened 2022-04-11
🐛 Bug Report: getFlag aspect ratio / wrong rendering


XeroxDev:
### 👟 Reproduction steps

I used the `/v1/avatars/flags/` endpoint and got not expected results. See results below:

### 👍 Expected behavior

Return the right aspect ratio of the flag like in the files itself (https://github.com/appwrite/appwrite/blob/master/app/config/avatars/flags/de.png)

How to fix it: calculate the right (missing) width / height, especially because they're optional.

### 👎 Actual Behavior

`/v1/avatars/flags/de`
![image](https://user-images.githubusercontent.com/10864620/162791065-32a0a670-5d9d-43eb-b4ba-00f5f710c169.png)

`/v1/avatars/flags/de?height=25`
![image](https://user-images.githubusercontent.com/10864620/162791166-f19c2ab5-4d52-4942-8913-cc2fb3e015a4.png)

`/v1/avatars/flags/de?width=50`
![image](https://user-images.githubusercontent.com/10864620/162791246-d4109052-9d90-4295-99d6-110196c9045e.png)

`/v1/avatars/flags/de?width=200`
![image](https://user-images.githubusercontent.com/10864620/162791315-28492402-7588-4193-9e16-83d4744e8cf8.png)

### 🎲 Appwrite version

Version 0.13.x

### 💻 Operating system

Linux

### 🧱 Your Environment

I use nginx as a reverse proxy.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

christyjacob4:
@XeroxDev  thanks for raising this issue. 
@Meldiron Please take a look at the image library

Meldiron:
Issue resolved with PR mentioned above.

eldadfux:
Actually this is the expected behavior. By default, the API crop the image and both default values for width and height are set to 100. To get the automatically calculate relative value, you just need to set `0` for one of the params.

XeroxDev:
> Actually this is the expected behavior. By default, the API crop the image and both default values for width and height are set to 100. To get the automatically calculate relative value, you just need to set `0` for one of the params.

Interesting, thanks. That works! But I'm curious, shouldn't this be the default behavior?

eldadfux:
[...]

PR #3087, opened 2022-04-12
Properly crop images in avatars API

Meldiron:
## What does this PR do?

Solves issue when resizing was not done properly if only one parameter was provided.

- [x] Fix Initials
- [x] Fix Flags
- [x] Fix Credit Cards
- [x] Fix Browsers
- [x] Fix URL Image

Not only we support proper resizing now, we also return full resolution unless asked to downscale.

## Test Plan

- [x] Manual QA

CleanShot 2022-04-12 at 13 36 30
CleanShot 2022-04-12 at 13 37 48
CleanShot 2022-04-12 at 13 37 31
CleanShot 2022-04-12 at 13 37 08

(Yes, I tested all affected endpoints)

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/3082

Yes

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202115796713529


eldadfux:
Not sure about this.

1. this is a breaking change
2. default image will have 0x0 size
3. If this is the case, should we make both params required?

Could this be solved by just mentioning that providing the value 0 will keep the original width or height?

Meldiron:
@eldadfux 

1. Kind-of. Same code will still work, but instead of getting default size of 100x100, you will get source image at maximum resolution
2. Correct and that's fine, that means original resolutions will be provided

Documenting endpoints better would also be a valid solution, yes.

eldadfux:
I think I would prefer to force the 100x100 resolution for preview images, to make devs aware of this instead of loading images in full resolution by default where most of the time it's far from optimized.

Meldiron:
@eldadfux PR updated to only address documentation changes. It also includes fox when QR code size 0x0 was allowed, resulting in a server error.

docs/references/avatars/get-browser.md
```diff
@@ -1 +1,3 @@
-You can use this endpoint to show different browser icons to your users. The code argument receives the browser code as it appears in your user /account/sessions endpoint. Use width, height and quality arguments to change the output settings.
\ No newline at end of file
+You can use this endpoint to show different browser icons to your users. The code argument receives the browser code as it appears in your user /account/sessions endpoint. Use width, height and quality arguments to change the output settings.
```
eldadfux, line 1:
```suggestion
You can use this endpoint to show different browser icons to your users. The code argument receives the browser code as it appears in your user /account/sessions endpoint. Use width, height and quality arguments to change the output settings.
```
We should / could use a link for `/account/sessions`?

eldadfux (commented):
One minor comment.

eldadfux (approved):

eldadfux:
Approved, once conflict is resolved we can merge.

PR #3092, opened 2022-04-13
Fix 'magic URL' accounts from not being able to create a new password.

PineappleIOnic:
## What does this PR do?

This PR fixes magic URL generated accounts from not being able to set a new password by setting the 'passwordUpdate' field on account creation to 0.

This PR also updates Anonymous session's `passwordUpdate` field to be 0 on user creation to keep better inline with the other non-password methods.

## Test Plan

A new test has been added to `AccountBase.php` which will attempt to update the password after creating a new account using magicURL.

Yes, I have

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -675,7 +675,7 @@ App::post('/v1/account/sessions/magic-url')
                 'emailVerification' => false,
                 'status' => true,
                 'password' => null,
-                'passwordUpdate' => \time(),
+                'passwordUpdate' => 0,
                 'registration' => \time(),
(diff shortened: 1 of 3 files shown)


eldadfux (changes requested):
How did this behavior change? I'm pretty sure it was already doing that.
We should also update the updatePassword docs to reflect this behavior.

eldadfux (approved):

eldadfux (commented):
Thinking about it, we should probably have a test for this. cc @TorstenDittmann

TorstenDittmann (approved):

PR #3105, opened 2022-04-17
CustomId console, reserved project

lohanidamodar:
## What does this PR do?

- Prevent creating a project with `console` as ID. making `console` a reserved project ID

## Test Plan

## Related PRs and Issues

YES

### DIFF
--- a/app/controllers/api/projects.php
+++ b/app/controllers/api/projects.php
@@ -77,6 +77,9 @@ App::post('/v1/projects')
         }
 
         $projectId = ($projectId == 'unique()') ? $dbForConsole->getId() : $projectId;
+        if($projectId === 'console') {
+            throw new Exception("'console' is a reserved project.", 400, Exception::PROJECT_RESERVED_PROJECT);
+        }
         $project = $dbForConsole->createDocument('projects', new Document([
             '$id' => $projectId == 'unique()' ? $dbForConsole->getId() : $projectId,
             '$read' => ['team:' . $teamId],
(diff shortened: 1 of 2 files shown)


app/controllers/api/projects.php
```diff
@@ -77,6 +77,9 @@
         }
 
         $projectId = ($projectId == 'unique()') ? $dbForConsole->getId() : $projectId;
+        if($projectId === 'console') {
```
Meldiron, line 80: Could you double-check if this is as string elsewhere, or do we have const for it?

lohanidamodar: we are using similar thing in other places, so I don't think we have a constant for this
https://github.com/appwrite/appwrite/blob/9e5436f80d88f62b48c6c2c9d87dd31169a8058a/app/init.php#L742

Meldiron: Awesome, thanks.

Meldiron (approved):
LGTM

eldadfux (approved):

TorstenDittmann:
It's not really a reserved keyword, but the project in this case actually already exists right?

lohanidamodar:
> It's not really a reserved keyword, but the project in this case actually already exists right?

it does, but I don't get back duplicate project error, I get error on developer tools, and when I refresh, console project is listed along with my other projects. I can't create collections, buckets but I can create keys, webhooks, add platforms etc.

PR #2721, opened 2022-01-30
Wildcard for hostname

Meldiron:
🚨 Utopia/framework release required before merge 🚨

## What does this PR do?

We currently don't support hostname syntax `*.myweb.com` to allow all subdomains. With this PR, you can start using the wildcard symbol `*` for sections where you want to allow any value.

I also added a proper validator, so people are warned when trying to add hostname like `http://myapp.com` or `myapp:8000`.

## Test Plan

The validator is properly tested at the library level. Current tests should pass.

- [x] Manual QA:
image

Before:

image

After: 

CleanShot 2022-03-28 at 11 18 15

(using `https://8080-tringug-authmodal-8l5p3rlrnxe.ws-eu38xl.gitpod.io/examples/appwrite.html` URL)

- [x] Validator QA:

CleanShot 2022-03-28 at 11 58 57

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/2939

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

✅

Issue #2939, opened 2022-03-11
🚀 Feature: Setting wildcard hosts for a platform


Asored-D:
### 🔖 Feature description

Adding platforms with static hosts is a very good and secure thing in general. But what if we use Appwrite for applications for which we don't know the host or the hosts are dynamic?

For example: it is amazing to use the Appwrite cloud functions from a WordPress plugin. But we can't add a web platform with a static domain because we don't know who is using the plugin and from where. Each plugin user call the cloud function from another host.

At this point we would need the possibility to set wildcard hosts for a project as optional feature. 

What do you think about it?

### 🎤 Pitch

Without the functionality to set wildcard hosts / domains for a platform we cannot use the Appwrite functionality in applications like WordPress plugins for example or for other applications where the host varies. The result will be a CORS error.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

Meldiron:
I recently wrote hostname validator, and the next step for me is to implement it into Appwrite 😇 You can find my implementation of wildcard validator here: https://github.com/utopia-php/framework/pull/47

Would such a validator work for you? It pretty-much allows you to do `*.vercel.app`.

Asored-D:
> I recently wrote hostname validator, and the next step for me is to implement it into Appwrite 😇 You can find my implementation of wildcard validator here: [utopia-php/framework#47](https://github.com/utopia-php/framework/pull/47)
> 
> Would such a validator work for you? It pretty-much allows you to do `*.vercel.app`.
> 
> EDIT: You can look at tests to see what is and isn't expected: https://github.com/utopia-php/framework/pull/47/files#diff-a84acb8776212ddc789b519f0ae8df5119fbf45f8f23a19268754c49604df2b7R60

When we use Appwrite not for standalone entire applications which are always running on the same domain, but for addons or plugins for existing projects, for examples content management systems like WordPress, this also will not work. Because the users are running our plugin on their own website and we don't know who is using the plugin. Here the host is unknown in general. We don't know it before. For such use cases only a `*` wildcard will work I think.

Another use case would be if we build an own API for others. This also won't work if the hosts are limited and must be defined before.

Maybe @eldadfux has a good idea on this point? :-)

Meldiron:
[...]

PR #3124, opened 2022-04-20
Feat: Wildcard hostname

Meldiron:
## What does this PR do?

- We now support `*.vercel.all` or `*.*.gitpod.io` or even simply `*` hostnames
- Hostname is now properly validated when creating/updating. You now cannot enter an invalid hostname
- In UI, when platform update fails, refresh list so we don't have incorrect data in list

## Test Plan

- [x] Manual QA:

1. Don't allow port and path:

https://user-images.githubusercontent.com/19310830/164199120-6b01218d-36ba-4a74-b2d3-6fc90aeaa5a4.mp4

2. Ensure wildcard can be inserted properly:

CleanShot 2022-04-20 at 11 37 39

3. Ensure wildcards are actually supported:

CleanShot 2022-04-20 at 11 43 25

(I was able to create and read an anonymous session from `https://8080-meldiron-almostredditpla-1cocjbj09gp.ws-eu34xl.gitpod.io/index2.html` with the hostname was `*.*.gitpod.io` or `*` or `*.*.gitpod.*`)

## Related PRs and Issues

- Outdated closed PR: https://github.com/appwrite/appwrite/pull/2721
- Issue: https://github.com/appwrite/appwrite/issues/2939

Yes


src/Appwrite/Network/Validator/Host.php
```diff
@@ -45,17 +46,16 @@ public function getDescription(): string
...
             return false;
         }
 
-        if (\in_array(\parse_url($value, PHP_URL_HOST), $this->whitelist)) {
-            return true;
-        }
-
-        return false;
+        $hostname = \parse_url($value, PHP_URL_HOST);
+        $hostnameValudator = new Hostname($this->whitelist);
```
TorstenDittmann, line 57:
```suggestion
        $hostnameValidator = new Hostname($this->whitelist);
```

src/Appwrite/Network/Validator/Origin.php
```diff
@@ -122,11 +123,8 @@ public function isValid($origin): bool
             return true;
         }
 
-        if (\in_array($host, $this->clients)) {
-            return true;
-        }
-
-        return false;
+        $validator = new Hostname($this->clients);
+        return $validator->isValid($host);
```
TorstenDittmann, line 127: As a rule of thumb, I would always have a empty line before the return for more clarity.

```suggestion
        $validator = new Hostname($this->clients);

        return $validator->isValid($host);
```

app/views/console/home/index.phtml
```diff
@@ -299,8 +299,9 @@ $usageStatsEnabled = $this->getParam('usageStatsEnabled', true);
...
         <input type="text" class="full-width" name="name" required autocomplete="off" placeholder="My Web App" maxlength="128" />
 
-        <label for="hostname">Hostname <span class="tooltip large" data-tooltip="The hostname that your website will use to interact with the <?php echo APP_NAME; ?> APIs in production or development environments. No port number required."><i class="icon-question"></i></span></label>
-        <input name="hostname" type="text" class="margin-bottom" autocomplete="off" placeholder="localhost" required>
+        <label for="hostname">Hostname <span class="tooltip large" data-tooltip="The hostname that your website will use to interact with the <?php echo APP_NAME; ?> APIs in production or development environments. No protocol or port number required."><i class="icon-question"></i></span></label>
+        <input name="hostname" type="text" class="margin-bottom" autocomplete="off" placeholder="yourapp.com" required>
+        <div class="text-fade text-size-xs margin-top-negative-small margin-bottom">You can use * to allow wildcard hostnames or subdomains.</div>
```
eldadfux, line 304: In the future, it might be cool to add small buttons to allow to quickly add `*.vercel.app` or `*.netlify.app`

app/controllers/api/projects.php
```diff
@@ -1016,6 +1016,14 @@
         /** @var Appwrite\Utopia\Response $response */
         /** @var Utopia\Database\Database $dbForConsole */
 
+        // Ensure hostname has proper structure (no port, protocol..)
```
eldadfux, line 1019: Why don't we initialize this param in the route signature instead of the Text validator?

src/Appwrite/Network/Validator/Host.php
```diff
@@ -1,6 +1,7 @@
 <?php
 
 namespace Appwrite\Network\Validator;
+use Utopia\Validator\Hostname;
```
eldadfux, line 4: Let's fix line spacing

src/Appwrite/Network/Validator/Origin.php
```diff
@@ -1,6 +1,7 @@
 <?php
 
 namespace Appwrite\Network\Validator;
+use Utopia\Validator\Hostname;
```
eldadfux, line 4: Line spacing.

app/views/console/home/index.phtml
Meldiron: Aaand `*.*.gitpod.io` probably

app/controllers/api/projects.php
Meldiron: We nest it to get both checks. FIrst, check for size (max length 256), then format (port, protocol,...)

eldadfux: We used to have this validator just for cases like this: https://github.com/utopia-php/framework/blob/fix-for-0.4/src/Validator/Multiple.php - not sure why we deprecate it. It could be very useful and help us stay much more consistent in this case and take advantage of the framework proper lifecycle: `input -> validation -> execution`

eldadfux: Thinking about it, we might have deprecated it because it was hard to generate a proper type for it in the APIs specs. Another option could be to add a hostname default length.

Some resources:
* https://en.wikipedia.org/wiki/Hostname#:~:text=The%20entire%20hostname%2C%20including%20the,maximum%20of%20253%20ASCII%20characters.
* https://stackoverflow.com/questions/8724954/what-is-the-maximum-number-of-characters-for-a-host-name-in-unix
image

Meldiron: Addressed in a PR: https://github.com/utopia-php/framework/pull/59/files

TorstenDittmann (approved):

PR #3144, opened 2022-04-25
Feat: Array API limits

Meldiron:
## What does this PR do?

- If there is an limit for array param, I document it properly
- If there is param without limit (unlimited size, or unlimited element length), I add limit and document it

## Test Plan

- Current tests must pass
- Manual QA (team membership, API keys, webhooks, storage extentions, function events&perms, db enums&indexes, listDocuments)

## Related PRs and Issues

x

Yes


TorstenDittmann:
Can we have a `const` for the `100`?

TorstenDittmann:
> @TorstenDittmann Added. Should I also use it for a description of parameters, instead of a hard-written number `100`? Also, do you think `APP_LIMIT_ATTRIBUTE_COUNT` sounds good?

Should be used wherever needed. So at some point if we wanna increase it - we only need to change a single number.

Meldiron:
Updated descriptions to use varuable, renamed variable to `APP_LIMIT_ARRAY_PARAMS_SIZE`

PR #3158, opened 2022-04-27
Feat: User membership subquery

Meldiron:
## What does this PR do?

Implement subquery for memberships (users) so we no longer need to keep duplicate data.

## Test Plan

- Current tests must pass

## Related PRs and Issues

x

Yes

### DIFF
--- a/app/config/collections.php
+++ b/app/config/collections.php
@@ -1077,8 +1077,8 @@ $collections = [
                 'signed' => true,
                 'required' => false,
                 'default' => [],
-                'array' => true,
-                'filters' => ['json'],
+                'array' => false,
+                'filters' => ['subQueryMemberships'],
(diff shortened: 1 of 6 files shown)


TorstenDittmann (commented):
General question, so the sub query attribute is still being created in the Database right?

Do we need that?

Meldiron:
I believe we need attributes for the database library to map stuff properly. The value of this attribute is always null. The real value is added during `getDocument()` when the subquery filter is applied.

eldadfux:
> General question, so the sub query attribute is still being created in the Database right?
> 
> Do we need that?

I think it's fine to keep it for now, we can refactor in the future if we have more insights for or against it.

eldadfux (approved):

PR #3210, opened 2022-05-12
Fi: SSL rework

Meldiron:
## What does this PR do?

Improve many aspects of SSL rework after manual QA.

## Test Plan

- [x] Manual QA

## Related PRs and Issues

x

Yes


app/controllers/general.php
```diff
@@ -99,13 +99,13 @@
...
-    
-                Resque::enqueue(Event::CERTIFICATES_QUEUE_NAME, Event::CERTIFICATES_CLASS_NAME, [
-                    'domain' => $domain->get()
-                ]);
+                    Console::info('Issuing a TLS certificate for the main domain (' . $domain->get() . ') in a few seconds...');
+        
+                    Resque::enqueue(Event::CERTIFICATES_QUEUE_NAME, Event::CERTIFICATES_CLASS_NAME, [
+                        'domain' => $domain->get()
+                    ]);
+                }
```
TorstenDittmann, line 108: Shouldn't we always enqueue here?

Meldiron: It was like this originally, but then I decided to move it before merging because at first sight, I had the same thought. But now during QA, I noticed it is generating certificates even for domains that have valid certificates.

This means "If we have domain document, we no longer generate certificate". Which is correct. The only way to get it there is by this exact script.

The worst-case scenario is that you visit main domain, generation fails (because your DNS is aware of record, but LetsEncrypt isn't yet)  and new you are screwed. For that case, there is this administrative command `docker-compose exec appwrite ssl`

TorstenDittmann (approved):

Issue #3046, opened 2022-03-31
🚀 Feature: API user endpoint to get list of memberships


outrelouxhe:
### 🔖 Feature description

An API endpoint to get a list of memberships for a user. 

### 🎤 Pitch

 At the moment, I need to get all teams and check for each one of them if the user is present in the memberships. This becomes an issue when the number of teams becomes large.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

Meldiron:
As far as I am concerned, this is not possible 🤔 @christyjacob4 @TorstenDittmann @eldadfux should we implement this? Sounds reasonable to me.

rafaelgazani:
@outrelouxhe I'm thinking of triggering a function to feed a collection that has the user id, team id or association id. How did you resolve?

TorstenDittmann:
I am not a 100% sure it makes sense to get all memberships, without the context of the team itself 🤔

outrelouxhe:
It makes sense for use cases where users, teams and team memberships are managed by a remote server through server api. At the moment I deal with a few hundreds of users but with thousands of teams which are used to filter the data within collections based on different categories : accessible customers, accessible geographic zones...   In order to avoid scanning all teams to get the memberships of a given user, it would be a lot easier to just get a list of team memberships for one user. This is more a user centric approach than a team centric approach like you're referring to, because the starting point is the user, not the team. Does it make sense to you ? Thank you.

outrelouxhe:
Thank you @rafaelgazani for sharing your solution which could also work for my use case.
However, I really hope that this feature request will be accepted because I would like to avoid to add and maintain a collection when everything is already present in the internal database and could be retrieved with a new API endpoint.

rafaelgazani:
I also hope @outrelouxhe . If this endpoint comes out in a future release, it will be easy to replace my solution.

christyjacob4:
@outrelouxhe , this looks reasonable and looks like you have a valid use case. But I'm going to have to wait with the decision of implementing it until we have more consensus

eldadfux:
[...]

PR #3211, opened 2022-05-12
feat(api): user memberships

TorstenDittmann:
## What does this PR do?

- adds an endpoint to list all memberships of a user
- adds new route to UI under a user
- changes response model for memberships:

```diff
userId
- name
- email
+ userName
+ userEmail
teamId
+ teamName
```
Bildschirmfoto 2022-05-12 um 15 23 13

## Test Plan

- adapted and added tests

## Related PRs and Issues

- #3046 

✅


app/views/console/users/team.phtml
```diff
@@ -111,14 +111,14 @@
                                                 <button class="danger">Remove</button>
                                             </form>
 
-                                            <img src="" data-ls-attrs="src={{member|avatar}}" data-size="200" alt="User Avatar" class="avatar pull-start margin-end" loading="lazy" width="60" height="60" />
```
Meldiron, line 114: Not sure about this change. Why from `member` to `member.userName` Was `member` previously object, ID, email or name?

TorstenDittmann: https://github.com/appwrite/appwrite/blob/70efbf1fb9c3277043ef0255ba320b02beee3bbe/public/scripts/filters.js#L8

We have a fallback to the name property 👍🏻

app/controllers/api/teams.php
```diff
@@ -423,8 +423,9 @@
 
         $response->setStatusCode(Response::STATUS_CODE_CREATED);
         $response->dynamic($membership
-            ->setAttribute('email', $email)
-            ->setAttribute('name', $name)
+            ->setAttribute('teamName', $team->getAttribute('name'))
+            ->setAttribute('userName', $user->getAttribute('name'))
+            ->setAttribute('userEmail', $user->getAttribute('email'))
```
Meldiron, line 428: Previous:
```
->setAttribute('email', $email)
->setAttribute('name', $name)
```
            
Again just making sure, was this on purpose - not using variable?

docs/references/users/get-user-memberships.md
```diff
@@ -0,0 +1 @@
+Get the user membership list by its unique ID.
```
Meldiron, line 1: Can we mention the limit? I can't see pagination, so there is some.

src/Appwrite/Utopia/Response/Model/Membership.php
```diff
@@ -22,24 +22,30 @@ public function __construct()
...
+                'example' => 'john@appwrite.io',
+            ])
             ->addRule('teamId', [
                 'type' => self::TYPE_STRING,
                 'description' => 'Team ID.',
                 'default' => '',
                 'example' => '5e5ea5c16897e',
             ])
-            ->addRule('name', [
+            ->addRule('teamName', [
```
Meldiron, line 43: Any reason why not to add teamID too? Could not be useful for us in UI, but might be really useful in server-SDK. Overall, if we mention some resources (like a team) and provide some specific data (name), we should also provide ID so they can get all other data of the team if they need to.

Meldiron (commented):
Last comments, the rest looks good.

app/controllers/api/teams.php
TorstenDittmann: Yeah it was, there is so much happening before I would use `$name` again, that I wanted to be a 100% sure and use the object from the database 👍🏻

Meldiron (approved):

app/controllers/api/teams.php
Cowtail1: Response

PR #3214, opened 2022-05-13
Refactor `functions.php`

everly-gif:
## What does this PR do?

Refactors `functions.php`

## Test Plan

Manual

- checked if the IDE is still able to pickup params + builds locally

## Related PRs and Issues

NIL

yes


everly-gif:
Need a **_careful review_** on the optional parameteres.

eldadfux (approved):

app/controllers/api/functions.php
```diff
@@ -191,11 +187,7 @@
     ->param('range', '30d', new WhiteList(['24h', '7d', '30d', '90d']), 'Date range.', true)
     ->inject('response')
     ->inject('dbForProject')
-    ->action(function ($functionId, $range, $response, $dbForProject) {
-        /** @var Appwrite\Utopia\Response $response */
-        /** @var Utopia\Database\Document $project */
-        /** @var Utopia\Database\Database $dbForProject */
-        /** @var Utopia\Registry\Registry $register */
+    ->action(function (string $functionId, ?string $range, Response $response, Database $dbForProject) {
```
TorstenDittmann, line 190: Not sure if `?string` makes sense here?

Since it will never be `null` and is always set to `30d`.

eldadfux: I think as long as it's set as optional we should be strict and keep it as is.

TorstenDittmann: But it's never optional here at this point, it will always be what is set as default in the param.

TorstenDittmann (approved):

PR #3219, opened 2022-05-15
Executor version environment to use proper version of Runtimes

lohanidamodar:
## What does this PR do?

Adds executor version environment variable (_APP_EXECUTOR_VERSION), currently using only internally that will allow us to use the correct version of runtimes depending on the executor version.

## Test Plan

Existing tests

## Related PRs and Issues

YES


app/config/runtimes.php
```diff
@@ -6,7 +6,7 @@
 /**
  * List of Appwrite Cloud Functions supported runtimes
  */
-$runtimes = new Runtimes();
+$runtimes = new Runtimes(App::getEnv('_APP_EXECUTOR_VERSION'));
```
Meldiron, line 9: Can we have a default here?

```suggestion
$runtimes = new Runtimes(App::getEnv('_APP_EXECUTOR_VERSION', 'v2'));
```

app/config/variables.php
```diff
@@ -615,6 +614,15 @@
                 'question' => '',
                 'filter' => ''
             ],
+            [
+                'name' => '_APP_EXECUTOR_VERSION',
+                'description' => 'Executor version',
+                'introduction' => '0.14.0',
+                'default' => 'v1',
```
Meldiron, line 621:
```suggestion
                'default' => 'v2',
```
In 0.14 we will use V2, right?

app/executor.php
```diff
@@ -584,7 +584,7 @@ function (string $runtimeId, array $vars, string $data, $timeout, $activeRuntime
     /** 
      * Warmup: make sure images are ready to run fast 🚀
      */
-    $runtimes = new Runtimes();
+    $runtimes = new Runtimes(App::getEnv('_APP_EXECUTOR_VERSION'));
```
Meldiron, line 587:
```suggestion
    $runtimes = new Runtimes(App::getEnv('_APP_EXECUTOR_VERSION', 'v2'));
```

Same as before

Dockerfile
```diff
@@ -187,6 +187,7 @@ ENV _APP_SERVER=swoole \
     _APP_EXECUTOR_SECRET=a-random-secret \
     _APP_EXECUTOR_HOST=http://appwrite-executor/v1 \
     _APP_EXECUTOR_RUNTIME_NETWORK=appwrite_runtimes \
```
Meldiron, line 189: I think we are missing:
-  value in `.env`
- binding in `docker-compose.yml`
- Both of the above for released files (they are in PHTMLs)

Meldiron (changes requested):

app/config/variables.php
lohanidamodar: There was never V1, so probably we will use V1 still. As executor is still part of Appwrite, we don't have versioning here yet.

Dockerfile
lohanidamodar: As you can see my comment in the description, we are not yet documenting this Environment variable publicly, for now we will be using it internally.

Meldiron (approved):

TorstenDittmann:
Why does this need a ENV variable?

Shouldn't this be part of the composer dependency?

PR #3223, opened 2022-05-16
feat: account update status

TorstenDittmann:
## What does this PR do?

Currently, the client method says `account.delete()`, indicating you are deleting the account. Thing is, you are only blocking the account because to properly delete the account and all its data (storage, database..), you will need a cloud function.

This PR renames the method to `account.block()`.

This PR also deletes users fully if deleted using UsersAPI. This means, we no longer reserve ID.

## Test Plan

Tests were missing, I added tests for this scenario.

✅


app/controllers/api/users.php
```diff
@@ -716,6 +715,11 @@
             ->setParam('sessionId', $sessionId)
         ;
 
+        $deletes
+            ->setType(DELETE_TYPE_USERS)
+            ->setDocument($user)
+        ;
+
```
Meldiron, line 722: Not sure about this addition. I understand it as deleting user when we delete session.

Meldiron (approved):

PR #3236, opened 2022-05-18
Fix: Google OAuth (userID)

Meldiron:
## What does this PR do?

Code quality changes in 0.14.x discovered wrong usage of Google API. User ID was missing, as the response does not give `id`, but `sub` instead.

> The response will be a JSON object with several properties about the user. The response will
> 
> **always include the sub key, which is the unique identifier for the user**.
> 
> Google also returns the user’s profile information such as name (first and last), profile photo URL, gender, locale, profile URL, and email. The server can also add its own claims, such as Google’s hd showing the “hosted domain” of the account when using a G Suite account.

Reference: https://www.oauth.com/oauth2-servers/signing-in-with-google/verifying-the-user-info/

## Test Plan

- [x] Manual QA on Gitpod

## Related PRs and Issues

Reported on Discord

Yes

### DIFF
--- a/src/Appwrite/Auth/OAuth2/Google.php
+++ b/src/Appwrite/Auth/OAuth2/Google.php
@@ -114,7 +114,7 @@ class Google extends OAuth2
     {
         $user = $this->getUser($accessToken);
 
-        return $user['id'] ?? '';
+        return $user['sub'] ?? '';
     }
 
     /**


saswatmoo:
@eldadfux Please look into it fix this as soon as possible.

Issue #3250, opened 2022-05-19
🐛 Bug Report: Collection attribute with name 'order' causes exception on listDocuments()


ddenev:
### 👟 Reproduction steps

I have a collection that contains an integer attribute named 'order'.
When I call `listDocuments()` like this (Web API):
```js
appwrite.database.listDocuments('budget-categories', [], 100, 0, '', 'after', ['order'], ['ASC'])
```
I get the following exception:
```
Access to XMLHttpRequest at 'http://192.168.88.11/v1/database/collections/budget-categories/documents?limit=100&offset=0&cursor=&cursorDirection=after&orderAttributes%5B0%5D=order&orderTypes%5B0%5D=ASC' from origin 'http://localhost:9000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
AppwriteException: Network request failed
    at Appwrite.<anonymous> (http://localhost:9000/node_modules/.q-cache/vite/spa/deps/appwrite.js?v=af4ab88e:1985:15)
    at Generator.throw (<anonymous>)
    at rejected (http://localhost:9000/node_modules/.q-cache/vite/spa/deps/appwrite.js?v=af4ab88e:495:32)
GET http://192.168.88.11/v1/database/collections/budget-categories/documents?limit=100&offset=0&cursor=&cursorDirection=after&orderAttributes%5B0%5D=order&orderTypes%5B0%5D=ASC net::ERR_FAILED 500
```
This only happens with the 'order' attribute. If I change to any other attribute from the collection, it works.

### 👍 Expected behavior

No exception thrown when ordering by the 'order' attribute

### 👎 Actual Behavior

See reproduction above

### 🎲 Appwrite version

Version 0.14.x

### 💻 Operating system

Windows

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

stnguyen90:
For the core team, I tested this on 0.13.4 and got the follow in my appwrite logs:

```
appwrite  | [Error] Type: PDOException
appwrite  | [Error] Message: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'order DESC, table_main._id ASC
appwrite  |             LIMIT 0, 12' at line 21
[...]

PR #3254, opened 2022-05-20
fix: escape order attribute names

jkklapp:
## What does this PR do?

Escapes `orderAttributes` order attributes to handle potential collision of order attributes names with SQL keywords.

## Test Plan

No tests should break

## Related PRs and Issues

#3250 

Yes

### DIFF
--- a/public/scripts/dependencies/appwrite.js
+++ b/public/scripts/dependencies/appwrite.js
@@ -1831,7 +1831,7 @@
                         payload['cursorDirection'] = cursorDirection;
                     }
                     if (typeof orderAttributes !== 'undefined') {
-                        payload['orderAttributes'] = orderAttributes;
+                        payload['orderAttributes'] = orderAttributes.map((orderAttribute) => '`' + orderAttribute + '`');
                     }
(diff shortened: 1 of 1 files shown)


public/scripts/dependencies/appwrite.js
```diff
@@ -1831,7 +1831,7 @@
                         payload['cursorDirection'] = cursorDirection;
                     }
                     if (typeof orderAttributes !== 'undefined') {
-                        payload['orderAttributes'] = orderAttributes;
+                        payload['orderAttributes'] = orderAttributes.map((orderAttribute) => '`' + orderAttribute + '`');
```
stnguyen90, line 1834: Hmm. I'm not sure if this is the best place to fix this problem because the API can still throw an error if it's made outside of the Appwrite console (in other SDKs). It would probably be best to fix this server side in the the utopia database package.

TorstenDittmann:
Hey 👋🏻 

This is the wrong file you changed, but we are going to solve this on the server-side as @stnguyen90 suggested.

Thank you very much for the effort 👍🏻

PR #3286, opened 2022-05-26
Refactor `mock.php`

everly-gif:
## What does this PR do?

Refactors `mock.php`

Yes


app/controllers/mock.php
```diff
@@ -237,9 +237,7 @@
     ->param('file', [], new File(), 'Sample file param', false)
     ->inject('request')
     ->inject('response')
-    ->action(function ($x, $y, $z, $file, $request, $response) {
-        /** @var Appwrite\Utopia\Request $request */
-        /** @var Utopia\Swoole\Response $response */
+    ->action(function (string $x, int $y, array $z, array $file, Request $request, Response $response) {
```
christyjacob4, line 240: Hmm, not sure if `$file` is an array.


Can you validate it using the SDK generator tests??  
By default the SDK generator tests make a request to the mock endpoints on `appwrite.io`

In our case, you need to get a local/Gitpod instance of Appwrite running and change the endpoint accordingly in the language files here
https://github.com/appwrite/sdk-generator/tree/master/tests/languages

You can test for any one language like PHP and confirm that the tests are passing in two cases
1. When file is not sent
2. When file is sent

christyjacob4 (changes requested):
Just one comment to validate the type of $files. Please check

app/controllers/mock.php
TorstenDittmann: For context, just checked that File is an array 👍🏻

TorstenDittmann (approved):

PR #3330, opened 2022-06-03
Feat : better initialization of storage libs

everly-gif:
## What does this PR do?

Removes redundant storage init logic

## Test Plan

Manual

## Related PRs and Issues

NIL

Yes


eldadfux:
I would wait with this PR until we find a better way to share resources across components in Appwrite. Coupling the HTTP server and executor with the Workers layer will introduce flow in our decoupled, single purpose components architecture.

Issue #2850, opened 2022-02-27
🚀 Feature: An easier way to just update document permissions


rdmchr:
### 🔖 Feature description

Currently, to update the permission of a document, you also have to update some kind of data, or at least pass some kind of data.
This does not work:
```TS
await database.updateDocument(listCollection, list, {}, [`user:${user.$id}`]);
```
So to update permissions I have to pass some data, even if the data is the same as the current one:
```TS
await database.updateDocument(listCollection, list, {name: "test"}, [`user:${user.$id}`]);
```

### 🎤 Pitch

It would be simple to remove the requirement of data, by removing [this check](https://github.com/appwrite/appwrite/blob/619ff84f016c72d43da19fa1a9622188ac79ddd1/app/controllers/api/database.php#L1995) or you could check if some permissions are passed and then do the check. 
This would make it much more simple to update permissions.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

gepd:
Today I'm facing this, would be good to have a solution for this.

Would be acceptable for you remove the check? or maybe only throw the error when no permission are updated, what do you think? @eldadfux 

I'll you accept a PR?

PR #3346, opened 2022-06-06
Allows to update permissions without payload

gepd:
## What does this PR do?

Currently, to update the permission of a document, you also have to update some kind of data, or at least pass some kind of data.

This does not work:

```js
await database.updateDocument(listCollection, list, {}, [`user:${user.$id}`]);
```

So to update permissions we have to pass some data, even if the data is the same as the current one:

```js
await database.updateDocument(listCollection, list, {name: "test"}, [`user:${user.$id}`]);
```

This change will allow to send an empty payload, but empty payload and read/write permission will keep throwing an error

```js
await database.updateDocument(listCollection, list, {}, [], []);
```

## Test Plan

No new test

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/2850

yes


gepd:
The failing test doesn't seems to be related with my changes

TorstenDittmann:
Restarted the tests 👍🏻 

I think it would make sense to make the `data` param optional. Otherwise it would be required but optional 🙂

gepd:
@TorstenDittmann good catch I totally forgot about that, I'm not good at types in PHP, but reading the manual, that should be the way to make it optional

let me know if that is enough 😀

TorstenDittmann:
Utopia has all those `->param` before the callback. If the last argument is true, the parameter of the rest API is marked as optional 😉

gepd:
@TorstenDittmann I think I got it, Is everyting good now?

app/controllers/api/database.php
```diff
@@ -1868,7 +1868,7 @@ function createAttribute(string $collectionId, Document $attribute, Response $re
     ->inject('usage')
     ->inject('events')
     ->inject('mode')
-    ->action(function (string $collectionId, string $documentId, string|array $data, ?array $read, ?array $write, Response $response, Database $dbForProject, EventAudit $audits, Stats $usage, Event $events, string $mode) {
+    ->action(function (string $collectionId, string $documentId, string|array|null $data, ?array $read, ?array $write, Response $response, Database $dbForProject, EventAudit $audits, Stats $usage, Event $events, string $mode) {
```
everly-gif, line 1871:
```suggestion
    ->action(function (string $collectionId, string $documentId, string|array $data, ?array $read, ?array $write, Response $response, Database $dbForProject, EventAudit $audits, Stats $usage, Event $events, string $mode) {
```
you can remove `null` from the types here , now that you have set the the last argument to true for the param.

Meldiron (changes requested):

gepd:
@Meldiron can you reset the tests?

gepd:
I need help
After checking the test I realized there is something else looking for the document structure

I'm getting error `400` which I believe comes from here: https://github.com/gepd/appwrite/blob/feat-empty-payload/app/controllers/api/database.php#L1943

I'm not sure but, is this related to utopia php?

Seems like here is looking for a specific structure: https://github.com/utopia-php/database/blob/main/src/Database/Database.php#L1169-L1173

@Meldiron @TorstenDittmann

app/controllers/api/database.php
```diff
@@ -1900,8 +1900,8 @@ function createAttribute(string $collectionId, Document $attribute, Response $re
 
         $data = (\is_string($data)) ? \json_decode($data, true) : $data; // Cast to JSON array
 
-        if (empty($data)) {
-            throw new Exception('Missing payload', 400, Exception::DOCUMENT_MISSING_PAYLOAD);
+        if (empty($data) && empty($read) && empty($write)) {
```
Meldiron, line 1903: I am unsure about the `empty` check here. Would this allow me to set permissions to an empty array - remove all current permissions?

If not, we might need check with `isset` or `=== null` or `=== ''` or change default value when defining `->param()` to something else. Not sure, please play around to see what works the best.

gepd:
I have finally fixed, I have updated the test, I think not deleting the collection was causing problems

@Meldiron @TorstenDittmann

Meldiron (approved):
Looks good to me

TorstenDittmann (approved):

Issue #3203, opened 2022-05-10
🚀 Feature: add verification code to log in?


paintingStyle:
### 🔖 Feature description

Can you add verification code to log in? The product needs this function

### 🎤 Pitch

The user can use the mobile phone number and verification code to complete the login, which is faster than email verification

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

everly-gif:
Closing this issue as Appwrite supports phone authentication 🥳 from the latest relase[ 0.15](https://dev.to/appwrite/announcing-appwrite-015-with-phone-authentication-more-5cjj). 
Here are some resources to get started with : 
[Appwrite Phone Authentication with Flutter](https://dev.to/appwrite/appwrite-phone-authentication-with-flutter-3jii)
[Phone Authentication with Appwrite and Twilio](https://dev.to/appwrite/phone-authentication-with-appwrite-and-twilio-26ek)

everly-gif:
#3357

Issue #1068, opened 2021-04-06
Phone Auth


gfyre:
## 🚀 Feature

Use a third-party API such as Twilio Verify to authenticate a user with their phone number.

## Have you spent some time to check if this issue has been raised before?

I've checked for this and this issue/feature request hasn't been raised previously.

(Have you googled for a similar issue or checked our older issues for a similar bug)

Yes.

### Have you read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/master/CODE_OF_CONDUCT.md)?

Yes.

## Pitch

Considering all major authentication methods, phone auth seems to be one of the most convenient one for users accessing an app through their phones. At this time, it's almost a necessity.

TorstenDittmann:
Hey @gfyre 👋🏻 

This feature is actually in our backlog for some time now, but wasn't a priority yet.
Also, we want to avoid coupling phone authentication to explicitly to a 3rd party provider like Twilio. Meaning, we still haven't completely decided on how to solve that problem. An open discussion can be found here: #907 

Right now, verifying a registered user with an e-mail address, could actually be achieved with WebHooks and Cloud functions.

gfyre:
Hey @TorstenDittmann,

So if I wanted to use Phone Auth in a project right now with Appwrite, is there a work-around?

TorstenDittmann:
Depending on the use case, yes.

But you would need to write all the logic for it, which might not work perfectly for all the cases. That would be no small amount of work.

Therefore, I would rather advise to have some patience until an official solution from Appwrite is ready.

gfyre:
Sounds good! Thanks @TorstenDittmann

zainab81:
اي

zainab81:
Hi

hatem-u:
Please also consider adding an interface for a GSM modem alongside 3rd party services. This could be very beneficial in many cases and cut costs as well.

ismaelAbogabal:
please guys update us about the news of this feature.
this feature is crucial to us.
Thanks.

TorstenDittmann:
This hasn't been a priority for some time now. But we are aware of this feature and that it's very much in demand.

markosole:
I am looking how to enable SMS OTP and I am stuck with: 
 - There is no documentation how to set it up (only blok post above)
 - Appwrite Admin does not have any settings for it. Where exactly do I set Twillio for example?

Can someone point me in right direciton please. thx

gewenyu99:
[...]

PR #3357, opened 2022-06-08
feat: phone authentication

TorstenDittmann:
## Todo
- [x] fix naming conventions for the classes (they are all over the place right now)
- [x] considering the approach for env variables we have for storage
- [x] add missing tests for the users service
- [x] add all the documentation
- [x] update UI

## What does this PR do?

Adds the option to use your phone as a method of authentication.

The initial implementation will allow using the following 3 providers:
- [Twilio](https://twilio.com/)
- [Telesign](https://www.telesign.com/)
- [TextMagic](https://www.textmagic.com/)

A Messaging Worker has been added to the Stack.

Following attributes have been added to the user:
- phone
- phoneVerification

Following env variables have been added:
- _APP_PHONE_PROVIDER
- _APP_PHONE_USER
- _APP_PHONE_SECRET
- _APP_PHONE_FROM

Following endpoints have been added:

**Create Phone Session**
`post('/v1/account/sessions/phone')`
- userId
- number

Will initialize the phone login and sends 6 random digits to the phone.

**Create Phone session (confirmation)**
`put('/v1/account/sessions/phone')`
- userId
- secret

Will confirm the the session and create a user account if not found. The 6 digits will be passed as the secret.

**Update Account Phone**
`patch('/v1/account/phone')`
- number
- password

Updates the users phone number.

**Create Phone Verification**
`post('/v1/account/verification/phone')`
- *no params*

Creates a verification token and sends 6 digits to the user's phone.

**Create Phone Verification (confirmation)**
`put('/v1/account/verification/phone')`
- userId
- secret

Confirms the phone number by passing the 6 digits as the secret.

**Update Phone**
`patch('/v1/users/:userId/email')`
- userId
- number

Updates the user's phone number.

**Update Phone Verification**
`patch('/v1/users/:userId/verification')`
- userId
- phoneVerification

Updates the phoneVerification attribute on the user.

## Test Plan

- added tests for all the new endpoints using a mock adapter for the
- manually tested all 3 adapters

## Related PRs and Issues

- #3203 
- #1068 

✅


.env
```diff
@@ -56,6 +56,10 @@ _APP_SMTP_PORT=1025
 _APP_SMTP_SECURE=
 _APP_SMTP_USERNAME=
 _APP_SMTP_PASSWORD=
+_APP_PHONE_PROVIDER=mock
```
eldadfux, line 59: The DSN approach that @christyjacob4 is implementing for databases might work better here and possibly for storage as well.

app/controllers/api/users.php
```diff
@@ -438,6 +439,45 @@
...
+    ->label('scope', 'users.write')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'users')
+    ->label('sdk.method', 'updatePhoneVerification')
+    ->label('sdk.description', '/docs/references/users/update-user-phone-verification.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_USER)
+    ->param('userId', '', new UID(), 'User ID.')
+    ->param('phoneVerification', false, new Boolean(), 'User phone verification status.')
```
eldadfux, line 455: We can drop the prefix here:`phoneVerification`?

app/controllers/api/users.php
```diff
@@ -438,6 +439,45 @@
         $response->dynamic($user, Response::MODEL_USER);
     });
 
+App::patch('/v1/users/:userId/verification')
```
eldadfux, line 442: Isn't this conflicting with email verification? 
```suggestion
App::patch('/v1/users/:userId/verification/phone')
```

app/controllers/api/users.php
eldadfux: OK, I understand why you did it like this.

app/init.php
```diff
@@ -64,6 +69,8 @@
 use Utopia\Storage\Device\Linode;
 use Utopia\Storage\Device\Wasabi;
 
+use function PHPUnit\Framework\matches;
```
eldadfux, line 72: Leftover

app/workers/messaging.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+use Appwrite\Auth\Phone;
+use Appwrite\Auth\Phone\{
```
eldadfux, line 4: Let's use a consistent syntax, one way or another.

src/Appwrite/Auth/Phone.php
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * @param string $method
+     * @param string $url
+     * @param array  $headers
+     * @param string $payload
+     *
+     * @return string
+     */
+    protected function request(string $method, string $url, array $headers = [], ?string $payload = null, ?string $userpwd = null): string
```
eldadfux, line 45: We usually call this method `call`

src/Appwrite/Utopia/Response/Model/User.php
```diff
@@ -47,12 +47,24 @@ public function __construct()
                 'default' => '',
                 'example' => 'john@appwrite.io',
             ])
+            ->addRule('phone', [
+                'type' => self::TYPE_STRING,
+                'description' => 'User phone number.',
+                'default' => '',
+                'example' => '+49 30 901820',
```
eldadfux, line 54: Should we keep spaces?

src/Appwrite/Utopia/Response/Model/User.php
```diff
@@ -47,12 +47,24 @@ public function __construct()
                 'default' => '',
                 'example' => 'john@appwrite.io',
             ])
+            ->addRule('phone', [
+                'type' => self::TYPE_STRING,
+                'description' => 'User phone number.',
```
eldadfux, line 52: We should mention the format to expect with link to relevant information about it.

app/controllers/api/account.php
```diff
@@ -519,7 +522,7 @@
             'countryCode' => ($record) ? \strtolower($record['country']['iso_code']) : '--',
```
eldadfux, line 522: I still think we should change the default create session endpoint to specifically create a session with email. Currently it seems like the API has a bias and suggest email is *the* way to go.

eldadfux (changes requested):

src/Appwrite/Auth/Phone.php
TorstenDittmann: I got this from OAuth, I can change both either way.

src/Appwrite/Utopia/Response/Model/User.php
TorstenDittmann: Should definitely be removed here.

app/controllers/api/account.php
TorstenDittmann: We can do that 👍 I initially did this for less breaking change and less work for me 😉

PR #3359, opened 2022-06-08
WIP: Refactor listUsers

Meldiron:
(WIP PR. I will update description later)

Concerns I have:

- Query length limit is 128. This might not be enough. For example emails. Let's make it const, and do something like 1024?

## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)


app/controllers/general.php
```diff
@@ -313,6 +313,27 @@
...
+        appwrite  |   string(25) "team:62a0ceb229130190e2bb"
+        appwrite  |   [4]=>
+        appwrite  |   string(31) "team:62a0ceb229130190e2bb/owner"
+        appwrite  | }
+        appwrite  | string(6) "member"
+
+        appwrite  | [Error] Message: matej@appwrite.io (role: ) missing scope (users.read)
+        */
+        \var_dump($roles);
+        \var_dump($role);
```
TorstenDittmann, line 335: https://github.com/appwrite/appwrite/blob/9cb315d272d12ca3d6c8bf1db91611ab11298d7b/app/controllers/general.php#L301

this breaks it - just rename the variable or not use it at all and it works again 👍🏻

app/views/console/users/index.phtml
```diff
@@ -21,10 +21,12 @@ $smtpEnabled = $this->getParam('smtpEnabled', false);
 
             <h2>Users</h2>
 
+            <!-- TODO: New queries syntax doesnt work. Needs to be fixed -->
+            <!-- TODO: Sorting not imlemented. Do we want to, or keep it for later refactor? -->
```
TorstenDittmann, line 25: We don't do sorting yet in the UI.

app/config/collections.php
```diff
@@ -1092,7 +1092,7 @@
                 'filters' => ['subQueryMemberships'],
             ],
             [
-                '$id' => 'search',
+                '$id' => 'search', // TODO: We should probably rename this, as it will be users-facing. Maybe even remove it? But how would I search for user with mame "Matej" OR email "Matej"?
```
TorstenDittmann, line 1095: we might want to remove it completely if possible 👍🏻

app/config/collections.php
```diff
@@ -1059,7 +1059,7 @@
                 'filters' => [],
             ],
             [
-                '$id' => 'sessions',
+                '$id' => 'sessions', // TODO: Subqueries seems to be impossible to index. We dont have '.includes()' yet
```
TorstenDittmann, line 1062: yeah, that we can skip completely for now 👍🏻  only on root attributes - no nesting yet.

app/config/collections.php
Meldiron: What query would I use to search for matej? (in Appwrite Dashboard UI, to keep current functionality).
```
[
  Query.search("name", "Matej"),
  Query.search("email", "Matej")
]
```

This will not work, this is AND. We would need something like `Query.search(["name", "email"], "Matej")`, but that is not supported. Should we see if we can implement that?

Do we deprecate the search field for now? Or do we search only by one attribute - which one?
I don't think implementing complex searching in the current dashboard UI is a good choice.

TorstenDittmann: I would say lets remove it and we search by one attribute on the old dashboard or add basic functionality to select the attribute to search for 👍🏻

app/views/console/users/index.phtml
```diff
@@ -21,10 +21,11 @@ $smtpEnabled = $this->getParam('smtpEnabled', false);
 
             <h2>Users</h2>
 
+            <!-- TODO: New queries syntax doesnt work. Needs to be fixed -->
             <form class="box padding-small margin-bottom search"
                 data-service="users.list"
                 data-event="submit"
-                data-param-search="{{router.params.search}}"
+                data-param-queries="search.search('{{router.params.search}}')"
```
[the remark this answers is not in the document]
Meldiron: This is still open, I could not yet figure out how to do it in our JS framework. Torsten will investigate it later.

app/controllers/api/users.php
```diff
@@ -99,17 +99,17 @@
     ->label('sdk.response.code', Response::STATUS_CODE_OK)
     ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
     ->label('sdk.response.model', Response::MODEL_USER_LIST)
-    ->param('search', '', new Text(256), 'Search term to filter your list results. Max length: 256 chars.', true)
+    ->param('queries', [], new ArrayList(new Text(128), APP_LIMIT_ARRAY_PARAMS_SIZE), 'Array of query strings generated using the Query class provided by the SDK. [Learn more about queries](https://appwrite.io/docs/database#querying-documents). Maximum of 100 queries are allowed, each 128 characters long.', true)
```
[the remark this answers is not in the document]
Meldiron: We will need database indexes research before making a decision here.
For now, I would suggest going with the safest way. Only allowing attributes that you can read (that are in the response model), and only allowing queries for indexes that we have created internally (no compound queries).

This is limiting, but still much better than the current solution. Later we can improve this (after research) and release even more possibilities.

app/views/console/users/index.phtml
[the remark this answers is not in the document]
Meldiron: We decided to address this in separate UI PR, once we have API changes finished. It will make PR review easier.

I have removed changes to PHTML files.

PR #3361, opened 2022-06-09
Feat Internal Runtime State

PineappleIOnic:
## What does this PR do?

This PR re-implements runtime state into the executor allowing us to delay executions while a runtime is being launched, previously if an execution took place while a runtime was still launching it would fall through and result in an error. This PR remedies that.

## Test Plan

No tests have been changed,

Yes, I have


app/executor.php
```diff
@@ -219,7 +236,6 @@ function getStorageDevice($root): Device
             /**
              * Create container
              */
-            $secret = \bin2hex(\random_bytes(16));
```
Meldiron, line 222: Why are we removing secret?

app/executor.php
```diff
@@ -188,6 +192,19 @@ function getStorageDevice($root): Device
...
 
+        $secret = \bin2hex(\random_bytes(16));
+
+        if (!$remove) {
+            $activeRuntimes->set($runtimeId, [
+                'id' => $containerId,
+                'name' => $runtimeId,
+                'created' => $startTime,
+                'updated' => $endTime,
+                'status' => 'pending',
```
Meldiron, line 203: This might be a bug. Should we set here to pending, or running?

PineappleIOnic: We set it to pending since we haven't launched the runtime yet, we change it to running later on
https://github.com/appwrite/appwrite/blob/515c8c2e3212dd6e93ecccfed4a297a38cd6f035/app/executor.php#L332-L341

PineappleIOnic: This does remind me to handle errors though...

Meldiron (approved):
Looks good. Make sure to do manual QA. Even add some var_dumps, and make sure this edge case occured.

app/executor.php
```diff
@@ -177,6 +177,10 @@ function getStorageDevice($root): Device
     ->inject('response')
     ->action(function (string $runtimeId, string $source, string $destination, array $vars, array $commands, string $runtime, string $baseImage, string $entrypoint, bool $remove, string $workdir, $orchestrationPool, $activeRuntimes, Response $response) {
         if ($activeRuntimes->exists($runtimeId)) {
+            if ($activeRuntimes->get($runtimeId)['key'] == 'pending') {
```
christyjacob4, line 180:
```suggestion
            if ($activeRuntimes->get($runtimeId)['status'] == 'pending') {
```

christyjacob4 (changes requested):
@PineappleIOnic  This looks fine, I have some concerns over how it will behave with the retry logic that we already have in the Executor class... 

Let's discuss tmro

PR #3362, opened 2022-06-09
Feat: Start using docker compose V2

Meldiron:
## What does this PR do?

`docker-compose` seems to not have some features Appwrite started using on development. As a first step before implementing these in production is to make Appwrite ready to use new `docker compose` syntax.

This PR removes `docker-compose` installation from  the Appwrite Docker image, and adds `docker compose`.

**This PR does not include documentation changes. They will be required before releasing.**

## Test Plan

(Manual QA)

- Manually build Appwrite image `docker build . -t appwritest`
- Create Appwrite instance using new image:
```sh
docker run -it --rm \
    --volume /var/run/docker.sock:/var/run/docker.sock \
    --volume "$(pwd)"/appwrite:/usr/src/code/appwrite:rw \
    --entrypoint="install" \
    appwritest:latest
```
- Appwrite was started, and working properly

CleanShot 2022-06-09 at 13 40 50

## Related PRs and Issues

x

Yes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,8 +34,7 @@ ENV PHP_REDIS_VERSION=5.3.7 \
     PHP_SWOOLE_VERSION=v4.8.9 \
     PHP_IMAGICK_VERSION=3.7.0 \
     PHP_YAML_VERSION=2.2.2 \
-    PHP_MAXMINDDB_VERSION=v1.11.0 \
-    DOCKER_COMPOSE_VERSION=v2.5.0
+    PHP_MAXMINDDB_VERSION=v1.11.0
 
 RUN \
   apk add --no-cache --virtual .deps \
@@ -131,7 +130,9 @@ LABEL maintainer="team@appwrite.io"
 ARG VERSION=dev
 ARG DEBUG=false
 ENV DEBUG=$DEBUG
+
 ENV DOCKER_CONFIG=${DOCKER_CONFIG:-$HOME/.docker}
+ENV DOCKER_COMPOSE_VERSION=v2.5.0
 
 ENV _APP_SERVER=swoole \
     _APP_ENV=production \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -232,12 +233,16 @@ RUN \
...
   docker-cli \
-  docker-compose \
   libgomp \
   && docker-php-ext-install sockets opcache pdo_mysql \
   && apk del .deps \
   && rm -rf /var/cache/apk/*
 
+RUN \
+  mkdir -p $DOCKER_CONFIG/cli-plugins \
+  && curl -SL https://github.com/docker/compose/releases/download/v2.5.0/docker-compose-linux-x86_64 -o $DOCKER_CONFIG/cli-plugins/docker-compose \
```
TorstenDittmann, line 243: Can we make the version number here as a variable on top?

TorstenDittmann (approved):

Dockerfile
stnguyen90: Does this also need to consider architecture as well instead of hardcoding `x86_64`?

Meldiron: - Will make version ENV variable
- Don't think we need to consider architecture, right? It is all running in the same environment, inside a docker container. I will add it tho. Won't hurt, may help.

Meldiron: I was not able to use the command suggested by @stnguyen90 , as it is a `bash` script, but Apwrite runs on `sh` (`/bin/sh: /bin/bash: not found`)

I replaced it with just `$(uname -m)` instead, WDYT?

Meldiron: Nevermind, I think I figured it out. I used:

```sh
ARCH=$(uname -m) && if [ $ARCH == "armv7l" ]; then $ARCH="armv7"; fi
```

stnguyen90: > Don't think we need to consider architecture, right? It is all running in the same environment, inside a docker container. I will add it tho. Won't hurt, may help.

I think there are different images for different architectures and the matching image is used automatically. On my rpi:

```shell
$ uname -m
aarch64

$ docker run --rm -it appwrite/appwrite uname -m
aarch64
```

On my (intel) macbook:

```shell
$ uname -m                                      
x86_64

$ docker run --rm -it appwrite/appwrite uname -m
x86_64
```

TorstenDittmann (approved):

Issue #3329, opened 2022-06-03
🚀 Feature: Appwrite Query Character Limit Changeable


Infinil:
### 🔖 Feature description

Query Character Limit should be above 128 characters or there should be an option to remove this limit for URL attributes atleast.

### 🎤 Pitch

Recently in Appwrite 0.14 an update was pushed which limits the size of query to 128 characters for an attribute. Let's say my database has URL attribute and I want to list the documents with that URL [key index] then it can very easily surpass the character limit and can result in error while listing the documents. Hence I think this change is important. 

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

stnguyen90:
FYI, the limit in question here is the `new Text(128)` as part of the `queries` [validation](https://github.com/appwrite/appwrite/blob/f1afc45ae2007bc49a9a0509ed25109c8c11705c/app/controllers/api/database.php#L1605).

Infinil:
> FYI, the limit in question here is the `new Text(128)` as part of the `queries` [validation](https://github.com/appwrite/appwrite/blob/f1afc45ae2007bc49a9a0509ed25109c8c11705c/app/controllers/api/database.php#L1605).

Yes but it wasn't like that before 0.14 update right?

stnguyen90:
@Meldiron, would you be able to comment on this? This looks like the related commit: https://github.com/appwrite/appwrite/commit/d36a2602ed22843dad78ded063ae477956934619.

superseby2:
Just adding to the thread. I am encountering the pb with a simple query which goes like this Query.equal("id",[<an rray of 20 items>]) (so no need to have a query using URLs)
Quite an issue for us at the moment as the whole service works for some and not for others. Thx!

stnguyen90:
@superseby2, what version of Appwrite are you on? This should have actually been released. 🧐

superseby2:
Oh I see. Thx @stnguyen90  I am a version behind (I had just finished migrating our functions when 0.15 went out and could do it again :)) . 
Thx!

superseby2:
Could not find the info in the releases changes but is there still a limit we should be aware of or the limit is gone ?

Thx again.

stnguyen90:
[...]

PR #3381, opened 2022-06-15
Feat: Increase array element length

Meldiron:
## What does this PR do?

Previously, we didn't have any hard limit. In 0.14 we introduced a limit of 100 elements per array, with 128 length limit for each element.

This created a problem for devs who use longer (for instance) queries longer than 128 characters. This can easily happen when searching for an email or URL as both can cross 128 by standard.

This PR increases this limit to 2100 (maximum URL length 2048 + some more for query syntax).

## Test Plan

- [x] Implement tests

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/3329

Yes


TorstenDittmann (approved):

Infinil:
> I noticed that limit 2100 would apply to the whole query, which means, that if you had OR query (multiple options on an attribute), this limit would be shared. So if our rule was to support URL length, we would need 210k, because you can have a maximum of 100 OR options in a query. But that is too high it feels like.
> 
> After some more thinking, in theory, OR queries can be split into multiple requests. If we properly documented the limits, we should be fine with whatever length limit. That way we only need to ensure the limit allows at least 1 option no matter how long it is - 2100 should be enough.
> 
> Not confident with either solution tho.. Would love to see more opinions on this.

How about making a total limit of 128k characters of Queries . Basically you can still only make upto 100 queries but it either can be 64 queries of each 2k characters or 100 queries of each 1280 characters [Keeping 2k characters per query the maximum] 
That's just a suggestion tho 😅

PR #3438, opened 2022-06-27
Feat: Functions proxy

Meldiron:
## What does this PR do?

Appwrite currently scales functions by running one runtime of each function on each executor. This might be handy for some projects, but for others, it creates a lot of resource waste. This PR implements a function proxy that has info about all executors and forwards traffic using smart strategy (round-robin now, hardware-usage-based in future)

- [x] Update production docker-compose.yml
- [x] Do coroutine server on functions proxy (use commit as reference https://github.com/appwrite/appwrite/pull/3438/commits/42302e562351a4c764af73d2442d5f31f2ee24fa  )
- [x] Remove `_APP_OPENSSL_KEY_V1` from functions proxy
- [x] Remove `_APP_WORKER_PER_CORE` from functions proxy
- [ ] Write proper changelog

## Test Plan

 - [x] Existing tests must pass. These has all been underlying changes

## Related PRs and Issues

x

Yes


src/FunctionsProxy/Adapter/RoundRobin.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace FunctionsProxy\Adapter;
+
+use FunctionsProxy\Adapter;
+
+class RoundRobin extends Adapter
```
eldadfux, line 7: While we use the adapter pattern, this is not really and adapter, but an algorithm.

tests/resources/docker/docker-compose.yml
```diff
@@ -27,7 +27,7 @@ services:
     networks:
       - gateway
       - appwrite
-  
+
```
eldadfux, line 30: We're using this compose file, only for testing our docker-compose parser. No need to add extra env vars. Actually, it might be better if use a non-appwrite examples here.

eldadfux (changes requested):
Initial review. I still think we can improve the new classes structure to better tell the story the new "src" libraries are trying to solve.

PR #3473, opened 2022-06-29
feat: nested models in spec

TorstenDittmann:
## What does this PR do?

- adds a recursive function to get all nested models - previously we only went "one level" deep
- this allows us to finally generate Console SDK without issues

## Test Plan

- open api & swagger specs are more valid now

✅


src/Appwrite/Specification/Format.php
```diff
@@ -8,40 +8,22 @@
...
-     */
-    protected $authCount;
-
-    /**
-     * @var array
-     */
-    protected $params = [
+    protected array $services;
+    protected array $keys;
+    protected int $authCount;
```
lohanidamodar, line 25: Doc comment missing

src/Appwrite/Specification/Format.php
```diff
@@ -56,14 +38,6 @@ abstract class Format
...
 
-    /**
-     * @param App $app
-     * @param array $services
-     * @param Route[] $routes
-     * @param Model[] $models
-     * @param array $keys
-     * @param int $authCount
-     */
     public function __construct(App $app, array $services, array $routes, array $models, array $keys, int $authCount)
```
lohanidamodar, line 41: Did we decide to remove doc comments in favor of type definition?

lohanidamodar (approved):
LGTM, just a query

src/Appwrite/Specification/Format.php
TorstenDittmann: I don't see too much value for doc types anymore when not needed.

It just creates a lot of noise in the code and increases the code we have to write, without giving any value.

app/tasks/specs.php
```diff
@@ -123,105 +123,94 @@
...
+        foreach ($platforms as $platform) {
+            $routes = [];
+            $models = [];
+            $services = [];
+
+            foreach ($appRoutes as $key => $method) {
+                foreach ($method as $route) {
+                    /** @var \Utopia\Route $route */
+                    $routeSecurity = $route->getLabel('sdk.auth', []);
+                    $sdkPlatofrms = [];
```
abnegate, line 135: Since you have fixed them elsewhere, could we fix this name typo too

abnegate (approved):
Looks good!

PR #3491, opened 2022-07-03
cache cleanup

shimonewman:
## What does this PR do?

1. Handling cache files logic in a central place (app::init(), app::shutdown()).
2. Purging outdated cache files via deletes worker.


app/config/collections.php
```diff
@@ -2764,6 +2764,45 @@
...
+                'type' => Database::VAR_INTEGER,
+                'format' => '',
+                'size' => 0,
+                'signed' => false,
+                'required' => true,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
```
eldadfux, line 2783: Indentation seems wrong here, we can run the coding standards fixer.

app/controllers/api/avatars.php
```diff
@@ -48,10 +50,21 @@
 
     $cache = new Cache(new Filesystem(APP_STORAGE_CACHE . '/app-0')); // Limit file number or size
     $data = $cache->load($key, 60 * 60 * 24 * 30 * 3/* 3 months */);
-
     if ($data) {
         //$output = (empty($output)) ? $type : $output;
 
+        $cacheRow = $dbForProject->getDocument('cache', $key);
```
eldadfux, line 56: Can we find a better name? We don't ever use `Row`

app/controllers/api/avatars.php
```diff
@@ -48,10 +50,21 @@
...
+        if($cacheRow->isEmpty()){
+            Authorization::skip(fn () => $dbForProject->createDocument('cache',  new Document([
+                '$id' => $key,
+                'dateAccessed' => time(),
```
eldadfux, line 60: Lets sync with @fogelito once we move away from unix timestamps.

app/controllers/api/avatars.php
```diff
@@ -48,10 +50,21 @@
...
+                'path' => 'app-0'
+            ])));
+        } else {
```
eldadfux, line 63:
```suggestion
        }
        else {
```

app/workers/deletes.php
```diff
@@ -279,12 +310,16 @@ protected function deleteExecutionLogs(int $timestamp): void
...
-        $this->deleteForProjectIds(function (string $projectId) use ($timestamp) {
+        $this->deleteForProjectIds(
+
+            function (string $projectId) use ($timestamp) {
             $dbForProject = $this->getProjectDB($projectId);
             // Delete Sessions
             $this->deleteByGroup('sessions', [
                 new Query('expire', Query::TYPE_LESSER, [$timestamp])
             ], $dbForProject);
+
```
eldadfux, line 321: Extra whitespace

app/config/collections.php
```diff
@@ -2764,6 +2764,45 @@
             ],
         ]
     ],
+    'cache' => [
+        '$collection' => Database::METADATA,
+        '$id' => 'cache',
+        'name' => 'Cache',
+        'attributes' => [
+            [
+                '$id' => 'dateAccessed',
```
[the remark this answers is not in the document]
shimonewman: accessedAt .timestamp is a reserved word

app/config/collections.php
shimonewman: Ok

app/controllers/api/avatars.php
shimonewman: Ok

app/controllers/api/avatars.php
shimonewman: Ok

app/workers/deletes.php
```diff
@@ -122,6 +127,35 @@ public function shutdown(): void
...
+            $cache = new Local(APP_STORAGE_CACHE);
+
+            $this->deleteByGroup(
+                'cache',
+                [
+                new Query('accessedAt', Query::TYPE_LESSER, [$timestamp])
+                ],
+                $dbForProject,
+                function (Document $document) use ($cache) {
+                    $path = $cache->getRoot() . '/' . $document->getAttribute('path') . '/' .  $document->getId();
```
eldadfux, line 148: Shouldn't we also delete the database log?

docker-compose.yml
```diff
@@ -107,6 +107,7 @@ services:
       - ./docs:/usr/src/code/docs
       - ./public:/usr/src/code/public
       - ./src:/usr/src/code/src
+      - ./vendor:/usr/src/code/vendor
```
eldadfux, line 110: We don't need to commit this file.

app/controllers/api/avatars.php
```diff
@@ -48,9 +48,10 @@
 
     $cache = new Cache(new Filesystem(APP_STORAGE_CACHE . '/app-0')); // Limit file number or size
     $data = $cache->load($key, 60 * 60 * 24 * 30 * 3/* 3 months */);
-
     if ($data) {
         //$output = (empty($output)) ? $type : $output;
+        App::setResource('cacheKey', fn () => $key);
+        App::setResource('cachePath', fn () => 'app-0');
```
eldadfux, line 54: This is very uncommon to how we handle resources. We usually don't manipulate them from within the action. For simple primitive strings, it might work better to use labels with special params patterns, very similar to how we use labels for abuse. We can use the cache shutdown hook to parse the pattern and populate the proper cache keys based on the request object.

eldadfux (changes requested):

app/config/collections.php
```diff
@@ -2764,6 +2764,33 @@
...
+                'signed' => false,
+                'required' => true,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+         ],
+        'indexes' => [
+            [
+                '$id' => '_key_accessed',
```
eldadfux, line 2786: Is this key name matching our other conventions?

.gitignore
```diff
@@ -10,3 +10,6 @@
 debug/
 app/sdks
 dev/yasd_init.php
+test/tmp/
```
eldadfux, line 13: Leftovers

src/Appwrite/Utopia/Response.php
```diff
@@ -445,6 +445,20 @@ public function output(Document $document, string $model): array
         return $this->payload;
     }
 
+    public function file($data, $contentType, $date, $state): void
```
eldadfux, line 448: - This could potentially go to the core framework.
- Indentation is off.
- No need to couple a core method with Appwrite specific behavior `X-Appwrite-Cache` can be at controller level. No need for `state` param. 
- Not sure this method works with bigger files.

src/Appwrite/Utopia/Response.php
```diff
@@ -472,7 +486,7 @@ public function yaml(array $data): void
     /**
      * @return array
      */
-    public function getPayload(): array
+    public function getPayload(): string | array
```
eldadfux, line 489: This is usually a bad practice, why do we need it to return a string?

app/init.php
```diff
@@ -694,6 +695,8 @@ function (mixed $value) {
     return $layout;
 }, ['locale']);
 
+App::setResource('cacheKey', fn () => '');
```
eldadfux, line 698: No need for this resources anymore.

app/controllers/shared/api.php
```diff
@@ -114,6 +115,27 @@
 
     $deletes->setProject($project);
     $database->setProject($project);
+
```
eldadfux, line 118: Why do we add this method twice in two different init hooks?

app/controllers/shared/api.php
```diff
@@ -114,6 +115,27 @@
...
+                Authorization::skip(fn () => $dbForProject->createDocument('cache', new Document([
+                        '$id' => $key,
+                        'accessedAt' => time(),
+                    ])));
+            } else {
+                $cacheLog->setAttribute('accessedAt', time());
+                Authorization::skip(fn () => $dbForProject->updateDocument('cache', $cacheLog->getId(), $cacheLog));
+            }
+            $data = json_decode($data, true);
+            $response->file(base64_decode($data['payload']), $data['content-type'], $data['date'], 'hit');
```
eldadfux, line 136: Here we can just set the "state" header with `$response->header()`.

app/controllers/general.php
```diff
@@ -612,6 +612,7 @@
     });
 
 include_once __DIR__ . '/shared/api.php';
+include_once __DIR__ . '/shared/cache.php';
```
eldadfux, line 615: Seems like we're not using this controller anymore. We can also remove all the `cache` groups from the different routes.

app/controllers/api/storage.php
```diff
@@ -800,7 +800,8 @@
 App::get('/v1/storage/buckets/:bucketId/files/:fileId/preview')
     ->alias('/v1/storage/files/:fileId/preview', ['bucketId' => 'default'])
     ->desc('Get File Preview')
-    ->groups(['api', 'storage'])
+    ->groups(['api', 'storage', 'cache'])
+    ->label('useCache', true)
```
eldadfux, line 804: We usually use a namespace convention for labels.
```suggestion
    ->label('cache', true)
```

eldadfux (changes requested):

src/Appwrite/Utopia/Response.php
shimonewman: The first ver was a string then i changed to [].
Fixed it!

app/controllers/shared/api.php
```diff
@@ -114,6 +114,32 @@
...
     $deletes->setProject($project);
     $database->setProject($project);
+
+    $useCache = $route->getLabel('cache', false);
+    if ($useCache) {
+            $key = md5($request->getURI() . $request->getServer('query_string'));
+            $cache = new Cache(new Filesystem(APP_STORAGE_CACHE . DIRECTORY_SEPARATOR . 'app-' . $project->getId()));
+            $data = $cache->load($key, 60 * 60 * 24 * 30 * 1);
+        if (!empty($data)) {
+            $cacheLog = $dbForProject->getDocument('cache', $key);
```
lohanidamodar, line 124: Don't we need to skip Authorization here?

app/controllers/shared/api.php
```diff
@@ -238,6 +264,28 @@
...
 
     $route = $utopia->match($request);
+
+    $useCache = $route->getLabel('cache', false);
+    if ($useCache) {
+            $key = md5($request->getURI() . $request->getServer('query_string'));
+            $cache = new Cache(new Filesystem(APP_STORAGE_CACHE . DIRECTORY_SEPARATOR . 'app-' . $project->getId()));
+            $data = $response->getPayload();
+        if (!empty($data)) {
+            $cacheLog = $dbForProject->getDocument('cache', $key);
```
lohanidamodar, line 274: authorization skip might be needed here?

lohanidamodar (commented):
Looks good, few comments.

.env
```diff
@@ -71,7 +71,8 @@ _APP_FUNCTIONS_INACTIVE_THRESHOLD=60
 OPEN_RUNTIMES_NETWORK=appwrite_runtimes
 _APP_EXECUTOR_SECRET=your-secret-key
 _APP_EXECUTOR_HOST=http://appwrite-executor/v1
-_APP_MAINTENANCE_INTERVAL=86400
+_APP_MAINTENANCE_INTERVAL=2592000
```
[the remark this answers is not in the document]
shimonewman: fixed

app/config/collections.php
shimonewman: fixed

app/controllers/api/avatars.php
shimonewman: changed

app/controllers/api/avatars.php
shimonewman: Ok

app/workers/deletes.php
```diff
@@ -279,13 +313,15 @@ protected function deleteExecutionLogs(int $timestamp): void
...
     protected function deleteExpiredSessions(int $timestamp): void
     {
-        $this->deleteForProjectIds(function (string $projectId) use ($timestamp) {
-            $dbForProject = $this->getProjectDB($projectId);
+        $this->deleteForProjectIds(
+            function (string $projectId) use ($timestamp) {
+                $dbForProject = $this->getProjectDB($projectId);
             // Delete Sessions
-            $this->deleteByGroup('sessions', [
+                $this->deleteByGroup('sessions', [
```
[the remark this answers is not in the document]
shimonewman: fixed

app/config/collections.php
```diff
@@ -2764,6 +2764,45 @@
...
+                'format' => '',
+                'size' => 0,
+                'signed' => false,
+                'required' => true,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                    'array' => false,
```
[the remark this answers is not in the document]
shimonewman: fixed

app/workers/deletes.php
```diff
@@ -122,6 +127,35 @@ public function shutdown(): void
...
+     * @param int $timestamp
+     */
+    protected function deleteCache(int $timestamp): void
+    {
+        $this->deleteForProjectIds(function (string $projectId) use ($timestamp) {
+
+            $dbForProject = $this->getProjectDB($projectId);
+            $cache = new Local(APP_STORAGE_CACHE);
+
+            $this->deleteByGroup(
```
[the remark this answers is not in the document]
shimonewman: fixed

app/workers/deletes.php
shimonewman: We are using deleteByGroup()

app/controllers/api/avatars.php
shimonewman: solution refactored

app/config/collections.php
shimonewman: changed

.gitignore
shimonewman: removed

src/Appwrite/Utopia/Response.php
shimonewman: changed

app/init.php
shimonewman: removed

app/controllers/shared/api.php
shimonewman: fixed

app/controllers/general.php
shimonewman: removed

app/controllers/api/storage.php
shimonewman: changed

app/controllers/shared/api.php
shimonewman: fixed

app/controllers/shared/api.php
shimonewman: fixed

app/controllers/api/avatars.php
```diff
@@ -37,50 +35,30 @@
...
-
     $data = $image->output($output, $quality);
-
-    $cache->save($key, $data);
-
     $response
-        ->setContentType('image/png')
-        ->addHeader('Expires', $date)
+        ->addHeader('Expires', \date('D, d M Y H:i:s', \time() + (60 * 60 * 24 * 30)) . ' GMT')
         ->addHeader('X-Appwrite-Cache', 'miss')
```
eldadfux, line 51: Would be great to if we can add this header from one of the hooks

app/controllers/shared/api.php
```diff
@@ -114,6 +114,33 @@
...
+                Authorization::skip(fn () => $dbForProject->updateDocument('cache', $cacheLog->getId(), $cacheLog));
+            }
+
+            $data = json_decode($data, true);
+
+            $response
+                ->addHeader('Expires', \date('D, d M Y H:i:s', \time() + (60 * 60 * 24 * 30)) . ' GMT')
+                ->addHeader('X-Appwrite-Cache', 'hit')
+                ->setContentType($data['content-type'])
+                ->file(base64_decode($data['payload']), $data['content-type'], $data['date']);
```
eldadfux, line 141: I have second thoughts about this `file()` method. It doesn't feel like it adds much value compared to just using `send()`. Anyway we need to send the content type one line up, so not sure there's a point for it.

app/controllers/shared/api.php
```diff
@@ -238,6 +265,28 @@
...
+            if ($cacheLog->isEmpty()) {
+                Authorization::skip(fn () => $dbForProject->createDocument('cache', new Document([
+                        '$id' => $key,
+                        'accessedAt' => time(),
+                    ])));
+            } else {
+                $cacheLog->setAttribute('accessedAt', time());
+                Authorization::skip(fn () => $dbForProject->updateDocument('cache', $cacheLog->getId(), $cacheLog));
+            }
+            $data['payload'] = base64_encode($data['payload']);
```
eldadfux, line 285: Let's use null safety operators when accessing keys in assoc arrays.

app/controllers/shared/api.php
```diff
@@ -238,6 +265,28 @@
     }
 
     $route = $utopia->match($request);
+
+    $useCache = $route->getLabel('cache', false);
+    if ($useCache) {
+            $key = md5($request->getURI() . $request->getServer('query_string'));
```
eldadfux, line 271: getServer is not the optimal way to access this data, we might even want to deprecate it this method, as it is more useful with our legacy FPM server. `getParams()` should be better

src/Appwrite/Utopia/Response.php
eldadfux: I can still see both types.

src/Appwrite/Utopia/Response.php
```diff
@@ -445,6 +445,27 @@ public function output(Document $document, string $model): array
...
+     *
+     * @param string $data
+     * @param string $contentType
+     * @param string $date
+     *
+     * @return void
+     */
+    public function file(string $data, string $contentType, string $date): void
+    {
+        $this->payload = [
```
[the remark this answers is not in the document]
eldadfux: Still off.

app/controllers/shared/api.php
```diff
@@ -238,6 +265,28 @@
...
+                Authorization::skip(fn () => $dbForProject->createDocument('cache', new Document([
+                        '$id' => $key,
+                        'accessedAt' => time(),
+                    ])));
+            } else {
+                $cacheLog->setAttribute('accessedAt', time());
+                Authorization::skip(fn () => $dbForProject->updateDocument('cache', $cacheLog->getId(), $cacheLog));
+            }
+            $data['payload'] = base64_encode($data['payload']);
+            $cache->save($key, json_encode($data));
```
eldadfux, line 286: We can add the `miss` header here.

eldadfux (changes requested):

app/controllers/shared/api.php
```diff
@@ -114,6 +114,36 @@
...
+        $data = $cache->load($key, $timestamp);
+        if (!empty($data)) {
+            $cacheLog = Authorization::skip(fn () => $dbForProject->getDocument('cache', $key));
+            if ($cacheLog->isEmpty()) {
+                Authorization::skip(fn () => $dbForProject->createDocument('cache', new Document([
+                '$id' => $key,
+                'accessedAt' => time(),
+                ])));
+            } else {
+                $cacheLog->setAttribute('accessedAt', time());
```
eldadfux, line 132: We need to update this attribute only once a day not on every request, this could overload the DB with write requests and significantly slow down our API.

app/controllers/shared/api.php
```diff
@@ -114,6 +114,36 @@
...
+                Authorization::skip(fn () => $dbForProject->updateDocument('cache', $cacheLog->getId(), $cacheLog));
+            }
+
+            $data = json_decode($data, true);
+            $response
+              ->addHeader('Expires', \date('D, d M Y H:i:s', \time() + $timestamp) . ' GMT')
+              ->addHeader('X-Appwrite-Cache', 'hit')
+              ->setContentType($data['content-type'])
+              ->file(base64_decode($data['payload']), $data['content-type'])
```
eldadfux, line 141: Have we made sure that this causes the framework to skip the route execution? If not, the cache will not work.

app/controllers/shared/api.php
```diff
@@ -238,6 +268,32 @@
...
+
+            $cacheLog = $dbForProject->getDocument('cache', $key);
+            if ($cacheLog->isEmpty()) {
+                Authorization::skip(fn () => $dbForProject->createDocument('cache', new Document([
+                '$id' => $key,
+                'accessedAt' => time(),
+                ])));
+            } else {
+                $cacheLog->setAttribute('accessedAt', time());
+                Authorization::skip(fn () => $dbForProject->updateDocument('cache', $cacheLog->getId(), $cacheLog));
```
eldadfux, line 288: Same as the above comments for the init part.

eldadfux (changes requested):

app/controllers/api/avatars.php
```diff
@@ -331,21 +279,15 @@
...
-        $cache->save($key, $data);
-
         $response
+            ->addHeader('Expires', \date('D, d M Y H:i:s', \time() + 60 * 60 * 24 * 30) . ' GMT')
             ->setContentType('image/png')
-            ->addHeader('Expires', $date)
-            ->addHeader('X-Appwrite-Cache', 'miss')
-            ->send($data);
-
+            ->file($data, 'image/png')
```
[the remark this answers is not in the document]
shimonewman: We agreed on moving overriding the payload in send().
The catch is that we can not change Response::send() signature in order to pass the content-type as well

src/Appwrite/Utopia/Response.php
```diff
@@ -445,6 +445,64 @@ public function output(Document $document, string $model): array
...
+    /**
+     * Output response
+     *
+     * Generate HTTP response output including the response header (+cookies) and body and prints them.
+     *
+     * @param string $body
+     *
+     * @return void
+     */
+    public function send(string $body = ''): void
```
eldadfux, line 458: We don't need to replicate all the logic here, we can rely on the parent class to keep all the basic stuff.

app/controllers/shared/api.php
```diff
@@ -238,6 +269,30 @@
...
+        $data = $response->getPayload();
+        if (!empty($data)) {
+            $key = md5($request->getURI() . implode('*', $request->getParams()));
+            $cacheLog = $dbForProject->getDocument('cache', $key);
+            if ($cacheLog->isEmpty()) {
+                Authorization::skip(fn () => $dbForProject->createDocument('cache', new Document([
+                '$id' => $key,
+                'accessedAt' => \time(),
+                ])));
+            } elseif (date('Y/m/d', \time()) > date('Y/m/d', $cacheLog->getAttribute('accessedAt'))) {
```
eldadfux, line 284: Are you sure this condition works? Seems like it compares strings rather than objects or integers.

eldadfux (changes requested):

app/controllers/shared/api.php
eldadfux: From what I see the return value is string so this condition shouldn't work well.
See here: https://www.php.net/manual/en/function.date.php

eldadfux (approved):
We only need to comment the vendor paths in the compose file.

app/controllers/shared/api.php
```diff
@@ -271,15 +304,85 @@
...
+            preg_match_all('/{(.*?)}/', $label, $matches);
+            foreach ($matches[1] ?? [] as $pos => $match) {
+                $find = $matches[0][$pos];
+                $parts = explode('.', $match);
+
+                if (count($parts) !== 2) {
+                    throw new Exception('Too less or too many parts', 400, Exception::GENERAL_ARGUMENT_INVALID);
+                }
+
+                $namespace = $parts[0];
```
eldadfux, line 320: Add null safety

app/controllers/shared/api.php
```diff
@@ -124,11 +121,45 @@
...
+
+        if ($useCache) {
+            $key = md5($request->getURI() . implode('*', $request->getParams()));
+            $cache = new Cache(new Filesystem(APP_STORAGE_CACHE . DIRECTORY_SEPARATOR . 'app-' . $project->getId()));
+            $timestamp = 60 * 60 * 24 * 30;
+            $data = $cache->load($key, $timestamp);
+            if (!empty($data)) {
+                $cacheLog = Authorization::skip(fn() => $dbForProject->getDocument('cache', $key));
+
+                if ($cacheLog->isEmpty()) {
```
eldadfux, line 139: Do we need this block both in init and shutdown?

.env
```diff
@@ -72,6 +72,7 @@ OPEN_RUNTIMES_NETWORK=appwrite_runtimes
 _APP_EXECUTOR_SECRET=your-secret-key
 _APP_EXECUTOR_HOST=http://appwrite-executor/v1
 _APP_MAINTENANCE_INTERVAL=86400
+_APP_MAINTENANCE_RETENTION_CACHE=2592000
```
christyjacob4, line 75: We need to include this variable in variables.php and compose.phtml
https://github.com/appwrite/appwrite/blob/master/app/config/variables.php
https://github.com/appwrite/appwrite/blob/master/app/views/install/compose.phtml

app/config/collections.php
```diff
@@ -2798,6 +2798,62 @@
...
+                'size' => 255,
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => 'accessedAt',
+                'type' => Database::VAR_INTEGER,
```
christyjacob4, line 2819: date time changes are required here... Reminder to update the datatype, add filter after the merge.

app/controllers/api/avatars.php
```diff
@@ -331,28 +285,23 @@
...
+            ->file($data)
+        ;
         unset($image);
     });
 
 App::get('/v1/avatars/qr')
     ->desc('Get QR Code')
     ->groups(['api', 'avatars'])
     ->label('scope', 'avatars.read')
+    ->label('cache.resource', 'avatar/qr')
```
christyjacob4, line 304: dont we need the `cache, true` label here  ?

app/controllers/api/avatars.php
```diff
@@ -381,19 +330,20 @@
...
             ->setContentType('image/png')
-            ->send($image->output('png', 9));
+            ->file($image->output('png', 9))
+        ;
     });
 
 App::get('/v1/avatars/initials')
     ->desc('Get User Initials')
     ->groups(['api', 'avatars'])
     ->label('scope', 'avatars.read')
```
christyjacob4, line 345: label(cache, true) ?

app/controllers/shared/api.php
```diff
@@ -14,10 +14,38 @@
...
+            default => $responsePayload,
+        };
+
+        if (array_key_exists($replace, $params)) {
+            $label = \str_replace($find, $params[$replace], $label);
+        }
+    }
+
+    return $label;
+};
```
christyjacob4, line 48: wait this same change is present in the abuse PR right ? how did it make it to this PR ?

app/controllers/shared/api.php
```diff
@@ -271,15 +321,58 @@
...
+                    Authorization::skip(fn () => $dbForProject->createDocument('cache', new Document([
+                    '$id' => $key,
+                    'resource' => $resource,
+                    'accessedAt' => \time(),
+                    'signature' => $signature,
+                    ])));
+                } elseif (date('Y/m/d', \time()) > date('Y/m/d', $cacheLog->getAttribute('accessedAt'))) {
+                    $cacheLog->setAttribute('accessedAt', \time());
+                    Authorization::skip(fn () => $dbForProject->updateDocument('cache', $cacheLog->getId(), $cacheLog));
+                }
```
christyjacob4, line 356: I thought @eldadfux  mentioned to use a different way to check this?

phpunit.xml
```diff
@@ -6,14 +6,14 @@
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
          processIsolation="false"
-         stopOnFailure="false"
-        >
+         stopOnFailure="true"
+>
```
christyjacob4, line 10: can you revert this change

phpunit.xml
```diff
@@ -6,14 +6,14 @@
...
     <extensions>
         <extension class="Appwrite\Tests\TestHook" />
     </extensions>
     <testsuites>
         <testsuite name="unit">
-            <directory>./tests/unit</directory>
+            <directory>./tests/unit/</directory>
```
christyjacob4, line 16: this as well ?

docker-compose.yml
```diff
@@ -330,7 +327,6 @@ services:
     volumes: 
       - ./app:/usr/src/code/app
       - ./src:/usr/src/code/src
-      # - ./vendor/utopia-php/database:/usr/src/code/vendor/utopia-php/database
```
christyjacob4, line 333: You can leave these commented out instead of removing them

tests/e2e/Services/Storage/StorageBase.php
```diff
@@ -490,7 +490,7 @@ public function testFilePreviewCache(array $data): array
 
         $this->assertEquals(204, $file['headers']['status-code']);
         $this->assertEmpty($file['body']);
-
+        sleep(1);
```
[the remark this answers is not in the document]
shimonewman: just for the meanwhile until we find bulletproof solution for this

app/controllers/api/storage.php
```diff
@@ -904,23 +905,12 @@
             throw new Exception(Exception::STORAGE_FILE_NOT_FOUND);
         }
 
-        $cache = new Cache(new Filesystem(APP_STORAGE_CACHE . DIRECTORY_SEPARATOR . 'app-' . $project->getId() . DIRECTORY_SEPARATOR . $bucketId . DIRECTORY_SEPARATOR . $fileId)); // Limit file number or size
-        $data = $cache->load($key, 60 * 60 * 24 * 30 * 3/* 3 months */);
-
```
christyjacob4, line 909: If we removed this, where are we fetching the cached files from ?

app/controllers/api/avatars.php
shimonewman: fixed

app/controllers/api/storage.php
shimonewman: api shutdown hook

app/controllers/shared/api.php
shimonewman: They both share this function

app/controllers/api/storage.php
christyjacob4: okay 👍

docker-compose.yml
```diff
@@ -207,7 +205,6 @@ services:
     volumes:
       - ./app:/usr/src/code/app
```
christyjacob4, line 206: We need to add the `_APP_MAINTENANCE_RETENTION_CACHE` variable to all the containers that need it. 

It includes the `appwrite` and `appwrite-maintenance` containers

app/views/install/compose.phtml
```diff
@@ -146,6 +146,7 @@ services:
       - _APP_STATSD_HOST
       - _APP_STATSD_PORT
```
christyjacob4, line 147: We need to add the `_APP_MAINTENANCE_RETENTION_CACHE` variable to all the containers that need it. 

It includes the `appwrite` and `appwrite-maintenance` containers

christyjacob4 (commented):
Please address the last set of comments

christyjacob4 (approved):

PR #3492, opened 2022-07-03
Feat: Allow date overriding

Meldiron:
## What does this PR do?

Allows server SDK to do changes to `$createdAt` attribute of all collections. Use cases:

- I had `updated` field. I migrated to 0.15. I want to switch to `$createdAt` so I don't need Appwrite Function anymore. I need to migrate current dates from the custom field into `$createdAt`
- Invoice document was created, and `$createdAt` is used as information for when it was issued. An acconting team needs to update the date for legal reasons. Tho this edge-case, 99% of the time the `createdAt` is correct.

## Test Plan

- [ ] Implement new tests

## Related PRs and Issues

x

Yes

### DIFF
--- a/app/controllers/api/databases.php
+++ b/app/controllers/api/databases.php
@@ -2279,10 +2279,17 @@ App::patch('/v1/databases/:databaseId/collections/:collectionId/documents/:docum
         $data = \array_merge($document->getArrayCopy(), $data);
 
         $data['$collection'] = $collection->getId(); // Make sure user don't switch collectionID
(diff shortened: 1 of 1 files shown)


eldadfux:
I would vote against it. The use case seems very specific, the feature is very hard to explain and I'm not sure it worth the added complexity

Meldiron:
Closing a PR. Since this is only mainly useful for migration reasons, such a feature should be implemented as a CLI command for administrators, not as a feature for developers.

Issue #3120, opened 2022-04-20
🚀 Feature: Add event createMagicURL


gepd:
### 🔖 Feature description

add `account.session.magic.create` (or similar ) event to be use with appwrite webhooks and functions to get the token when we use `createMagicURL`, this will allow us to run a function and generate our own email design with a custom domain if we want

### 🎤 Pitch

Currently Appwrite have two issues with emails; 
the first one, we can't easily customise the design and the second, we can't use a domain per project to send an email.

This feature will complement https://github.com/appwrite/appwrite/issues/3115 and https://github.com/appwrite/appwrite/issues/730 and will give full controll to send email with custom design and use a different domain per project

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

Meldiron:
I like it... Sounds like a smooth solution to custom channels for sharing secrets.

One concern  I have is what if I only want to send secrets using SMS? How would I disable the email that Appwrite sends by default? 🤔

gepd:
Thinking more about the SMTP, disable it with wrong data is not the best option (or any other way to bypass it), currently we use that data for actions with the super admin too, so If I disable the SMTP and I want to recover the account of the super admin it won't be possible. 

a better solution would be to have an env var like `_APP_SMTP_SUPER` when this is true, SMTP will be used only for actions who required an email from the super admin, the rest of the email should be manually handled.

eldadfux:
Stale issue, closing down.

Issue #3115, opened 2022-04-19
🚀 Feature: Add Token to Team Membership Create event.


TheMardy:
### 🔖 Feature description

Like `account.recovery.create` and `account.verification.create`, send the secret / token that is required to accept a team membership when the `teams.memberships.create` event is triggered.
Possibly append it to the membership object that is returned at the moment? 
(With the same security measures as https://appwrite.io/docs/models/token)

### 🎤 Pitch

I found https://github.com/appwrite/appwrite/issues/730, which adds the token response to the password recovery and email verification endpoint. It would be great to be able to get the secret / token for a team invite as well. Adding this feature would enable Appwrite users to use their own Function for mailing.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

Meldiron:
Same comment as here: https://github.com/appwrite/appwrite/issues/3120#issuecomment-1103992105

I like it, but I am not sure how a developer would disable the original Appwrite functionality. If you want to use the custom channel to share it (for example custom email, SMS, or Discord bot), you need to disable the original email so you can ensure they got a secret from your channel.

TheMardy:
I was thinking of the same solution: set any string as SMTP to avoid the error.
The addition of a way to disable SMTP would be a nice-to-have, but we can get custom channels to work without it. As long as we can retrieve the secret.

gepd:
As I commented in my issue, use wrong data is not a good solution as you will keep needing the smtp to recover the super admin account, however the behavior we need is already in appwrite; We should have the same behavior as when we use the server sdk. Even with the right smtp data, the admin sdk doesn't send emails to the user when client sdk does.

I haven't seen the code but if this is some kind of flag detecting the sdk, it could have and extra var setting this value (to use system or dev emails)

What you think about it?

TheMardy:
I agree. It's unpractical to set any string because it would not allow super admin recovery. 
I think the best way would maybe be to store this flag per project, add a toggle to "Services" in Project Settings (and maybe even store SMTP credentials per project).

stnguyen90:
[...]

PR #3493, opened 2022-07-03
WIP Feat token events

gepd:
## What does this PR do?

I'm working in this feature for my fork, and as I don't know if you are interesting in this feature I'll let it here to ask for help and to know you are interested too

Currently we can get a token using `account.recovery.create` and `account.verification.create`. Would be good to have the same behavior with `createMagicURL` and `createMembership`

It's is very useful and will solve the following issues

1. Customize emails sent by the system, there is not an easy way to do it now, with this new hooks we can implement our own way to send email

2. If you have multiples projects and domains, you can send emails from custom SMTP server/email address

3. We can have the ability to notificate a user with an external API like whatsapp, to recover the password or create a session with magic link.

There was one problem doing this, we should have a way to disable the current email. Many of us was adding wrong data to the SMTP server that way, the system do not send the email, but if we want to recover an admin account we will never get any email as the SMTP server have wrong data.

To solve this, I have added a new env var who will disable the notifications in the endpoints but not in the whole system

Finally have this two events/webhooks will add more consistency to the current API as there are two endponts doing that and two that can not

#### Features:

* adds `user.session.magic.create` event/webhook to get the token when `createMagicURL` is triggered
* adds `teams.membership.create` event/webhook to get the token when a membership is added to a team
* adds `_APP_SMTP_USE_EMAIL` env var to disable the emails with the client, but allowing keep using the SMTP to recover the admin account or any other thing that requires to send an email

The env var disable the following:

- Email when you ask for a magic URL session
- Email for password recovery
- Email to verificate your account
- Email to join to a team (add membership)

## Test Plan

I have added a test to get the secret when `createMagicURL` is called and a test to get the secret when create a team membership is call, however in the last one I'm having a problem, I'll comment in the code to see if some can help me

I wanted to test if the email wasn't send, but @Meldiron told me there is no way to enable or disable an env var from a test, so maybe you can help me with this too

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/3115 
https://github.com/appwrite/appwrite/issues/3120

Yes 👀

### DIFF
--- a/app/config/events.php
+++ b/app/config/events.php
@@ -21,6 +21,13 @@ return [
             'delete' => [
                 '$description' => 'This event triggers when a session for a user is deleted.'
             ],
+            'magic' => [
+            '$model' => Response::MODEL_TOKEN,
+            '$resource' => false,
+            'create' => [
+                '$description' => 'This event triggers when a magic link token for a session is created.',
+            ],
+        ],
         ],
         'recovery' => [
             '$model' => Response::MODEL_TOKEN,
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -702,11 +702,7 @@ App::post('/v1/account/sessions/magic-url')
             ;
         }
 
-        $events
-            ->setParam('userId', $user->getId())
-            ->setParam('tokenId', $loginSecret)
-            ->setUser($user)
-            ->setPayload(
+        $events->setPayload(
             $response->output(
                 $token->setAttribute('secret', $loginSecret),
                 Response::MODEL_TOKEN
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -625,6 +625,7 @@ App::post('/v1/account/sessions/magic-url')
         $roles = Authorization::getRoles();
         $isPrivilegedUser = Auth::isPrivilegedUser($roles);
         $isAppUser = Auth::isAppUser($roles);
+        $useEmail = App::getEnv('_APP_SMTP_USE_EMAIL', true);
 
         $user = $dbForProject->findOne('users', [new Query('email', Query::TYPE_EQUAL, [$email])]);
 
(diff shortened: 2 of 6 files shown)


gepd:
Hey @eldadfux @TorstenDittmann @Meldiron

As I mentioned above, I'm using this in my fork, It would be great to have it in appwrite, If you think there some something to improve or change I'll happy to discuss it

Meldiron:
Hey there 👋

I can see this is meant to be a solution for custom authorization flows, tho I have big concerns about `_APP_SMTP_USE_EMAIL`. It's functionality is hard to explain, too specific, and won't work well on Appwrite Cloud (it should be per-project instead).

For this problem, what I would expect is a proper token generation from server, and session creating using token. Something like:

```js
// On server
const users = new Users(client);
const tokenResponse = await users.createSessionToken("SOME_USER_ID");
console.log("Token for user to sigh in with is", tokenResponse.token); // JWT, 15min expiry. Maybe less

// On client
await account.createTokenSession("SOME_JWT_TOKEN");
console.log("Signed in as", await account.get());
```

Ill invite more core members to discuss this.

gepd:
I agree with you @Meldiron  `_APP_SMTP_USE_EMAIL` is not the best option, even for test it.

I like your proposal to generate token, this endpoint shouldn't send an email or notification.

This can be used for custom emails, or notifications and peer project.

gepd:
@Meldiron Did you had any discuss with the team about this?

eldadfux:
Hey @gepd thanks for submitting the PR! A few problems with this implementation that I can think of:
- Submitting custom events just for magic URLs seems inconsistent with other auth methods.
- Firing multiple events could be an option, but it might cause some bigger issues with realtime and webhooks as this might not be expected. 
- We could consider having some metadata to allow you to detect the auth method. This is already part of the session object, we could add it to the token.
- Disabling the email sending this way is unsafe and could easily be forgotten by a maintainer. Why not just disable your SMTP by providing no credentials?

gepd:
Hey @eldadfux 

currently `account.recovery.create` and `account.verification.create` give us the token in an event, that is why I was thinking to complement the API.

> We could consider having some metadata to allow you to detect the auth method. This is already part of the session object, we could add it to the token.

This could be a better solution, my use case is to implement a passwordless login without email, so this will fit my needs. And will allow others to implement a custom template email system

> Disabling the email sending this way is unsafe and could easily be forgotten by a maintainer. Why not just disable your SMTP by providing no credentials?

I'm currently doing this, the problem is if I want to recover an admin account I won't receive the email to recover it.
I have not tested yet, but does the `account.recovery.create` event trigger in the previous case (recover admin account)? If yes, I can use the same method to reset password

eldadfux:
> I'm currently doing this, the problem is if I want to recover an admin account I won't receive the email to recover it.
I have not tested yet, but does the account.recovery.create event trigger in the previous case (recover admin account)? If yes, I can use the same method to reset password

I think it should work the same.

gepd:
Excellent @eldadfux so should I close this PR?

Issue #3504, opened 2022-07-05
🐛 Bug Report: ListDocuments() w/ Older SDK Returns 401


stnguyen90:
### 👟 Reproduction steps

1. Use an SDK version meant for Appwrite version 0.14
2. Call listDocuments()

### 👍 Expected behavior

Appwrite should be backwards compatible and still work.

### 👎 Actual Behavior

Error response:

```json
{
    "message": "User (role: guest) missing scope (collections.read)",
    "code": 401,
    "type": "general_unauthorized_scope",
    "version": "0.15.1"
}
```

Appwrite docker log:

```
appwrite  | [Error] Timestamp: 2022-07-05T21:40:58+00:00
appwrite  | [Error] Method: GET
appwrite  | [Error] URL: /v1/databases/:databaseId/collections/:collectionId/usage
appwrite  | [Error] Type: Appwrite\Extend\Exception
appwrite  | [Error] Message: User (role: guest) missing scope (collections.read)
appwrite  | [Error] File: /usr/src/code/app/controllers/general.php
appwrite  | [Error] Line: 319
```

It looks like there's a problem with the usage alias: https://github.com/appwrite/appwrite/blob/08704e8c9397b511df92e42c2e646658b9535a5e/app/controllers/api/databases.php#L2672.

### 🎲 Appwrite version

Version 0.15.x

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

methodman7556:
Unfortunately, I am still getting a similar error, but for a different route;

```
appwrite_1                      | [Error] Timestamp: 2022-08-07T15:27:32+00:00
appwrite_1                      | [Error] Method: GET
appwrite_1                      | [Error] URL: /v1/databases/:databaseId/collections
appwrite_1                      | [Error] Type: Appwrite\Extend\Exception
appwrite_1                      | [Error] Message: foo@bar.com (role: member) missing scope (collections.read)
appwrite_1                      | [Error] File: /usr/src/code/app/controllers/general.php
appwrite_1                      | [Error] Line: 319
```

and the HTTP response 401;

```
{
    "message": "foo@bar.com (role: member) missing scope (collections.read)",
    "code": 401,
[...]

PR #3505, opened 2022-07-05
Fix get collection usage alias

stnguyen90:
## What does this PR do?

Fix the get collection usage API alias so it doesn't conflict with list documents

## Test Plan

1. Created e2e test
2. Manual test

Before fix:

image

After fix:

image

## Related PRs and Issues

#3504 

Yes

### DIFF
--- a/app/controllers/api/databases.php
+++ b/app/controllers/api/databases.php
@@ -2669,7 +2669,7 @@ App::get('/v1/databases/:databaseId/usage')
     });
 
 App::get('/v1/databases/:databaseId/collections/:collectionId/usage')
-    ->alias('/v1/database/collections/:collectionId/documents', ['databaseId' => 'default'])
+    ->alias('/v1/database/:collectionId/usage', ['databaseId' => 'default'])
     ->desc('Get usage stats for a collection')
     ->groups(['api', 'database'])
     ->label('scope', 'collections.read')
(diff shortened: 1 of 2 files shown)


app/controllers/api/databases.php
```diff
@@ -2669,7 +2669,8 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
     });
 
 App::get('/v1/databases/:databaseId/collections/:collectionId/usage')
-    ->alias('/v1/database/collections/:collectionId/documents', ['databaseId' => 'default'])
+    ->alias('/v1/database/:collectionId/usage', ['databaseId' => 'default']) // endpoint pre-0.15.0
+    ->alias('/v1/database/collections/:collectionId/usage', ['databaseId' => 'default'])
```
stnguyen90, line 2673: I put 2 aliases because `/v1/database/:collectionId/usage` was the endpoint pre-0.15.0, but I feel like it should have been `/v1/database/collections/:collectionId/usage`. Let me know if/how you want it changed.

lohanidamodar: Yes, it should have been as you suggested, however as `alias` is only for backward compatibility I think we are safe with keeping it the way it was before.

stnguyen90: Updated per your recommendation.

christyjacob4 (approved):
Great catch 👌

TorstenDittmann (approved):

Issue #3434, opened 2022-06-25
🚀 Feature: Function to return complete execution response


iamdineshkumar:
### 🔖 Feature description

Hi In present function execution response which received has character limit of 4000 as per document which says as "The script response output string. Logs the last 4,000 characters of the execution response output.".  This character limit can be lifted to give entire response. In certain cases we might require to use cloud function to generate on call based details such as consolidated/computed response through functions. For an example day wise sales report for given range of period time. In present case its not possible to achieve due to character limitation.

### 🎤 Pitch

Provide Environemental Varaiable to control this limit so based on use case users can modify this as well as this complete response is required for synchronously calling method and for logging purpose can be logged after truncketing

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

iamdineshkumar:
In this [line](https://github.com/appwrite/appwrite/blob/f9056d1dcae27563b40b80f5c65908260f17d39e/app/executor.php#L306) its hard coded as 1MB. This feature can be acheived by  modifing as below
Existing code:
`'response' => \mb_strcut($stdout, 0, 1000000), // Limit to 1MB`
modifyied As:

`'response' => \mb_strcut($stdout, 0, App::getEnv('APP_FUNCTIONS_RESPONSE_SIZE_LIMIT', 1000000)), // Limit to 1MB`
so we can introduce APP_FUNCTIONS_RESPONSE_SIZE_LIMIT as new environmental variable. 

@tomer  @eldadfux Kindly provide your view on this. This will give flexibility to user to control it based on requirement.

Meldiron:
Hey there 👋 In my opinion the limit makes sense, as it “protects” developer from returning big payloads and potentially worsing experience of users with slow connection.

With that said, I can see this limit increasing to 5MB. After-all 5MB, is already a limit for Appwrite Storage - files above 5MB are chunked. Actually, Appwrite Storage can be used for responses above 5MB (create file, respond with fileID, download file on client).

iamdineshkumar:
[...]

PR #3541, opened 2022-07-12
Function Response Size Limit Based On Env value

iamdineshkumar:
## What does this PR do?

Made Function Response Size Limit dynamic based on Environment Value. At present its hard coded to limit of 1MB in some case if require this will allow developers to set their limit.

## Test Plan
In Environement Limit is set to default 1MB and which can be changed based on necessity of developer.

## Related PRs and Issues

[Existing Issue](https://github.com/appwrite/appwrite/issues/3434)

### Have you read the [Contributing Guidelines on issues]
Yes


app/executor.php
```diff
@@ -322,8 +322,8 @@ function getStorageDevice($root): Device
             $endTime = \time();
             $container = array_merge($container, [
                 'status' => 'ready',
-                'response' => \mb_strcut($stdout, 0, 1000000), // Limit to 1MB
-                'stderr' => \mb_strcut($stderr, 0, 1000000), // Limit to 1MB
+                'response' => \mb_strcut($stdout, 0, App::getEnv('_APP_FUNCTIONS_RESPONSE_SIZE_LIMIT', 1000000)), // Default Limit to 1MB
+                'stderr' => \mb_strcut($stderr, 0, App::getEnv('_APP_FUNCTIONS_RESPONSE_SIZE_LIMIT', 1000000)), // Default Limit to 1MB
```
stnguyen90, line 326: Thanks for making this PR. I'm not sure if only changing the limit is the best approach to will really solve the problem since the data is also stored in the database which has a finite length as well:

https://github.com/appwrite/appwrite/blob/648b6f2038fb59b2beb78d81cea08aa730a106c3/app/config/collections.php#L2367-L2388

stnguyen90 (changes requested):

stnguyen90:
See https://github.com/appwrite/appwrite/issues/3434#issuecomment-1184661743

Issue #3565, opened 2022-07-20
🐛 Bug Report: users.*.create does not trigger when the user is created using OAuth


rdmchr:
### 👟 Reproduction steps

1. Set up a function that triggers on the event `users.*.create`
2. Create a user using OAuth (I used Discord)
3. The function should not trigger
4. Set the event to `users.*.sessions.*.create`
5. Create another user
6. The function should trigger

### 👍 Expected behavior

I expect the function to trigger even if the user is created using OAuth.

### 👎 Actual Behavior

The function does not get triggered.

### 🎲 Appwrite version

Version 0.15.x

### 💻 Operating system

Linux

### 🧱 Your Environment

I'm using Cloudflare and I think the only thing I changed from the default config is the SMTP env variables.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

PineappleIOnic:
This issue appears to be caused by the same problem as https://github.com/appwrite/appwrite/issues/2406

PR #3567, opened 2022-07-20
Fix OAuth Create User event

PineappleIOnic:
## What does this PR do?

This PR fixes a bug that event triggers are not being correctly fired when a new account is created with the OAuth endpoints.

Both webhooks and Function execution has been tested and confirmed to be working.

## Related PRs and Issues

#2406  #3565 

Yes

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -34,6 +34,7 @@ use Utopia\Locale\Locale;
 use Appwrite\Extend\Exception;
 use Appwrite\Messaging\Adapter\Realtime;
 use Appwrite\Utopia\Response\Model\User;
+use Utopia\Registry\Registry;
 use Utopia\Validator\ArrayList;
 use Utopia\Validator\Assoc;
 use Utopia\Validator\Range;
@@ -373,7 +374,8 @@ App::get('/v1/account/sessions/oauth2/:provider/redirect')
     ->inject('audits')
     ->inject('events')
     ->inject('usage')
-    ->action(function (string $provider, string $code, string $state, Request $request, Response $response, Document $project, Document $user, Database $dbForProject, Reader $geodb, Audit $audits, Event $events, Stats $usage) use ($oauthDefaultSuccess) {
+    ->inject('register')
+    ->action(function (string $provider, string $code, string $state, Request $request, Response $response, Document $project, Document $user, Database $dbForProject, Reader $geodb, Audit $audits, Event $events, Stats $usage, Registry $register) use ($oauthDefaultSuccess) {
 
         $protocol = $request->getProtocol();
         $callback = $protocol . '://' . $request->getHostname() . '/v1/account/sessions/oauth2/callback/' . $provider . '/' . $project->getId();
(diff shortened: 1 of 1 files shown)


stnguyen90:
Would it be worthwhile to wrap this logic in a function or something because there are a few places where we probably need this (creating user during creating phone session, magic url, team membership)?

app/controllers/api/account.php
```diff
@@ -494,6 +498,61 @@
...
+                        payload: $user
+                    );
+
+                    Realtime::send(
+                        projectId: 'console',
+                        payload: $user->getArrayCopy(),
+                        events: $allEvents,
+                        channels: $target['channels'],
+                        roles: $target['roles']
+                    );
```
TorstenDittmann, line 536: we don't need this event for the console 👍🏻

PineappleIOnic: Resolved. 👍

app/controllers/api/account.php
```diff
@@ -494,6 +498,53 @@
...
+                    /** Trigger Webhook */
+                    $userModel = new User();
+
+                    $userCreate = new Event(Event::WEBHOOK_QUEUE_NAME, Event::WEBHOOK_CLASS_NAME);
+                    $userCreate
+                        ->setProject($project)
+                        ->setUser($user)
+                        ->setEvent('users.[userId].create')
+                        ->setParam('userId', $userId)
+                        ->setPayload($user->getArrayCopy(array_keys(($userModel->getRules()))))
```
stnguyen90, line 511: What's the reasoning on this rather than just using `$user`?

stnguyen90 (commented):
Deferring to @TorstenDittmann on https://github.com/appwrite/appwrite/pull/3567#issuecomment-1190644563.

app/controllers/api/account.php
PineappleIOnic: To prevent data such as the password hash from being accidentally exposed

karniv00l:
Sorry to push, but any progress on that? Those events are crucial for my app

eldadfux:
@PineappleIOnic Why are replicating the entire logic for the shutdown method? I'm sure there's a way for us to add support for multiple events. We can try and change some of the terminology we use from `setEvent` to `addEvent`. Then, we can just iterate over all the available events and send them to our workers one by one.

eldadfux (changes requested):
@PineappleIOnic Why are replicating the entire logic for the shutdown method? I'm sure there's a way for us to add support for multiple events. We can try and change some of the terminology we use from setEvent to addEvent. Then, we can just iterate over all the available events and send them to our workers one by one.

stnguyen90:
@PineappleIOnic, what's the blocker on this? Looks like there are merge conflicts now.

PineappleIOnic:
We've been trying to figure out a better way of doing this instead of manually firing the event again, preferably by updating the event middleware to support multiple events on one route. I need to find some time to sync with @TorstenDittmann to talk to him about a possible solution to this.

stnguyen90:
@PineappleIOnic can this be closed since this isn't the approach we will be taking?

PineappleIOnic:
Yeah, we should be able to close this. Closing it now.

PR #3581, opened 2022-07-21
Use co-routines for creating projects

PineappleIOnic:
## What does this PR do?

This PR rewrites some of the project creation logic to perform database operations in parallel to decrease wait times.

## Test Plan

Benchmark 0.15.x branch then compare it to this branch, if it's faster than we have succeeded 🥳 

Yes

### DIFF
--- a/app/controllers/api/projects.php
+++ b/app/controllers/api/projects.php
@@ -65,8 +65,7 @@ App::post('/v1/projects')
     ->inject('response')
     ->inject('dbForConsole')
     ->inject('dbForProject')
-    ->inject('register')
-    ->action(function (string $projectId, string $name, string $teamId, string $description, string $logo, string $url, string $legalName, string $legalCountry, string $legalState, string $legalCity, string $legalAddress, string $legalTaxId, Response $response, Database $dbForConsole, Database $dbForProject, Registry $register) {
+    ->action(function (string $projectId, string $name, string $teamId, string $description, string $logo, string $url, string $legalName, string $legalCountry, string $legalState, string $legalCity, string $legalAddress, string $legalTaxId, Response $response, Database $dbForConsole, Database $dbForProject) {
 
         $team = $dbForConsole->getDocument('teams', $teamId);
 
--- a/app/controllers/api/projects.php
+++ b/app/controllers/api/projects.php
@@ -23,6 +23,10 @@ use Utopia\Database\Validator\UID;
 use Utopia\Domains\Domain;
 use Utopia\Registry\Registry;
 use Appwrite\Extend\Exception;
+use Utopia\Cache\Adapter\Redis;
+use Utopia\Cache\Cache;
+use Utopia\CLI\Console;
+use Utopia\Database\Adapter\MariaDB;
 use Utopia\Validator\ArrayList;
 use Utopia\Validator\Boolean;
 use Utopia\Validator\Hostname;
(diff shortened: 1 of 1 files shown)


app/controllers/api/projects.php
```diff
@@ -121,40 +126,56 @@
...
-                    'array' => $attribute['array'],
-                    'filters' => $attribute['filters'],
-                    'default' => $attribute['default'] ?? null,
-                    'format' => $attribute['format'] ?? ''
-                ]);
-            }
+            go(function () use ($collection, $register, $key, $project) {
+                $db = $register->get('dbPool')->get();
+                $redis = $register->get('redisPool')->get();
+                $cache = new Cache(new Redis($redis));
```
TorstenDittmann, line 135: There is no need to get a new connection here.

app/controllers/api/projects.php
```diff
@@ -121,40 +126,56 @@
...
 
-            $dbForProject->createCollection($key, $attributes, $indexes);
+                $projectDB->createCollection($key, $attributes, $indexes);
+                $register->get('dbPool')->put($db);
+                $register->get('redisPool')->put($redis);
+            });
         }
 
+        Console::info('Project Successfully created, Generated in ' . (\microtime(true) - $time) * 1000 . 'ms');
+
```
TorstenDittmann, line 178: We might have to wait here until the co-routines for the collections are finished here. Otherwise we return a response while it's still creating when the db is slow.

TorstenDittmann (changes requested):

app/controllers/api/projects.php
PineappleIOnic: Is there a way we can do something like promise.all without co\run? Since co\run won't work due to swoole attempting to create a new scheduler. Otherwise we may have to use a sleep

TorstenDittmann: You need to use `WaitGroup` 👍🏻

app/controllers/api/projects.php
```diff
@@ -63,7 +64,8 @@
     ->inject('response')
     ->inject('dbForConsole')
     ->inject('dbForProject')
-    ->action(function (string $projectId, string $name, string $teamId, string $description, string $logo, string $url, string $legalName, string $legalCountry, string $legalState, string $legalCity, string $legalAddress, string $legalTaxId, Response $response, Database $dbForConsole, Database $dbForProject) {
+    ->inject('register')
```
TorstenDittmann, line 67: not needed anymore

app/controllers/api/projects.php
PineappleIOnic: Implemented 👍

app/controllers/api/projects.php
```diff
@@ -121,40 +123,53 @@
...
+                    ]);
+                }
 
-            $dbForProject->createCollection($key, $attributes, $indexes);
+                $dbForProject->createCollection($key, $attributes, $indexes);
+                $wg->done();
+            });
         }
 
+        $wg->wait(5);
```
eldadfux, line 169: Is this value in second or ms?

PineappleIOnic: Seconds, That call doesn't actually wait in the usual way. It waits for all the coroutines to run; that 5 seconds is the maximum time, it will wait.

eldadfux:
Do we already have benchmarks for this PR?

PineappleIOnic:
average project creation with co-routines is: 323.91340732574ms
average project creation before co-routines is: 342.40982749245ms

Issue #3528, opened 2022-07-09
🐛 Bug Report: Cannot add empty values in a array of type not required


aadarshadhakalg:
### 👟 Reproduction steps

From Appwrite Dashboard, try to create a document with a few empty values in an optional array. No message is shown. The document is not created and in the console, this error is thrown.

![image](https://user-images.githubusercontent.com/19362725/178108171-89afc557-2997-4f44-997e-60b05cc183f0.png)
The document

### 👍 Expected behavior

When adding null values in array, error message should be shown.

### 👎 Actual Behavior

But, the process throws an error in the console. 

### 🎲 Appwrite version

Version 0.15.x

### 💻 Operating system

Linux

### 🧱 Your Environment

Local environment. Everything default.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

stnguyen90:
Thanks for raising this! We'll look into it!

stnguyen90:
FYI for Appwrite team: the API call is not made due to the error in the UI.

manishkumar-dsi:
@stnguyen90 could you please assign it to me? I would like to work on this issue.

stnguyen90:
Assigned! Thanks for your help! :pray:

PR #3591, opened 2022-07-23
fix for null issue during document create event

manishkumar-dsi:
- Added null & undefined type check in form.js

## What does this PR do?

Fix for the defect. It does the null check on value for array type field.

## Test Plan
Pre-requisite:
- Go to 'Database' page and create a new 'Collection'
- Create attributes, here at least one attribute should be marked 'Array' as required & 'Required' as optional.
- Once 'Collection' is created.

Test case - 1:
1) Go to Create a new 'Document' page
2) Click on 'Add Attribute' for 'Array' attribute minimum 2 times
3) Do not enter anything in the Value field
4) Create the Document and the document should be created with empty values

Test case - 2:
1) Go to Create a new 'Document' page
2) Click on 'Add Attribute' for 'Array' attribute 1 time
3) Do not enter anything in the Value field
4) Create the Document and the document should be created with empty values

Test case - 3:
1) Select any existing document
2) Update data with different values.
3) document should be updated

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/3528

Yes

### DIFF
--- a/public/scripts/services/form.js
+++ b/public/scripts/services/form.js
@@ -20,7 +20,7 @@
                     value = parseFloat(value);
                     break;
                 case 'string':
-                    value = value.toString();
+                    value = (typeof value === 'undefined' || value === null) ? "": value.toString();
                     if (value.length === 0) {
                         value = null;
                     }


public/scripts/services/form.js
```diff
@@ -20,7 +20,7 @@
                     value = parseFloat(value);
                     break;
                 case 'string':
-                    value = value.toString();
+                    value = (typeof value === 'undefined' || value === null) ? "": value.toString();
```
TorstenDittmann, line 23:
```suggestion
                    value = value?.toString() ?? '';
```

This might be a bit more elegant 🤔

TorstenDittmann (changes requested):
Do you mind running `npm run build` so the `dist` files are build?

stnguyen90:
@manishkumar-dsi sorry, this doesn't seem relevant anymore with the release of [1.1.0](https://github.com/appwrite/appwrite/releases/tag/1.1.0)

PR #3597, opened 2022-07-25
Fix: OAuth duplication error

Meldiron:
## What does this PR do?

When claiming an anonymous account with an OAuth2 provider, if an account with such an email exists, you end up with a 500 error. With this PR, we throw a proper error message.

For developers, a solution is to first delete the anonymous session before signing in, to go through the signing flow instead of the account claiming flow.

## Test Plan

- [ ] Manual QA

## Related PRs and Issues

Reported on Discord as unexpected behavior.

Yes

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -540,7 +540,11 @@ App::get('/v1/account/sessions/oauth2/:provider/redirect')
 
         Authorization::setRole('user:' . $user->getId());
 
-        $dbForProject->updateDocument('users', $user->getId(), $user);
+        try {
+            $dbForProject->updateDocument('users', $user->getId(), $user);
+        } catch (Duplicate $th) {
+            throw new Exception('Account already exists', 409, Exception::USER_ALREADY_EXISTS);
+        }
 
         $session = $dbForProject->createDocument('sessions', $session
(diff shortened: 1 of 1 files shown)


app/controllers/api/account.php
```diff
@@ -540,7 +540,11 @@
 
         Authorization::setRole('user:' . $user->getId());
 
-        $dbForProject->updateDocument('users', $user->getId(), $user);
+        try {
+            $dbForProject->updateDocument('users', $user->getId(), $user);
```
eldadfux, line 544: Shouldn't we expect this to be the case when updating a document? How is this error even possible?

Meldiron: 1. You sign in with OAuth
2. You sign out
3. You create an anonymous session
4. You use OAuth to claim your anonymous account

It will try to update your current anonymous account with a new email from OAuth. This ends up with duplicate error since there already is account with this email.

Hardanish-Singh (commented):
.

Meldiron:
I am almost certain this is a stale problem; If anyone was waiting for this to get merged, please open GitHub issue and ping me, I will look into any bug that is still unresolved regarding this

Issue #3615, opened 2022-07-27
🐛 Bug Report: Cannot search user by phone


XanManZA:
### 👟 Reproduction steps

Given a user in the database has the phone number `'+123456789'`. When I try and find the user by their phone:
```javascript
const { users } = await appwriteUsersSdk.list(
    Query.equal('phone', '+123456789'),
);
console.debug(users);
```

### 👍 Expected behavior

`console.debug(users);` should return an array with 1 user Appwrite user object in it.

### 👎 Actual Behavior

`console.debug(users);` returns an empty array.

### 🎲 Appwrite version

Version 0.15.x

### 💻 Operating system

Linux

### 🧱 Your Environment

AWS EC2 Ubuntu 20.04

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

stnguyen90:
The [List Users](https://appwrite.io/docs/server/users?sdk=nodejs-default#usersList) API takes a search term rather than a Query. For phone based accounts, you should be able to put the phone number into the search and it should work:

https://github.com/appwrite/appwrite/blob/7ea8c3d1732e9f049d682d073e5371e39782a621/app/controllers/api/account.php#L902

Are you making any other modifications to the User after the session is created?

XanManZA:
> The [List Users](https://appwrite.io/docs/server/users?sdk=nodejs-default#usersList) API takes a search term rather than a Query. For phone based accounts, you should be able to put the phone number into the search and it should work:
> 
> https://github.com/appwrite/appwrite/blob/7ea8c3d1732e9f049d682d073e5371e39782a621/app/controllers/api/account.php#L902
> 
> Are you making any other modifications to the User after the session is created?

Hey Steven. Just before I lodged the bug report I made doubly sure and reproduced the error - neither the `Query` nor the normal string search works.

Meldiron:
Hey there @XanManZA 👋 Thanks for the confirmation.

I tried to create a phone session and find it using the `search` parameter, and it worked fine:

![CleanShot 2022-07-28 at 11 15 43](https://user-images.githubusercontent.com/19310830/181469381-dcdde989-6f34-4b9b-9d17-0958694fba1f.png)

[...]

PR #3619, opened 2022-07-28
Fix: User Search attribute (phone)

Meldiron:
## What does this PR do?

Users have a `search` attribute to support simple and performant search operations. This attribute should include all id, name, email, and phone number.

When we implemented phone numbers, we didn't give this enough attention. We now need to include a phone number with each update, and email is no longer required.

This PR makes sure to keep the proper state of this `search` attribute on the user each time a user is updated.

## Test Plan

No tests were added as it would need interception between AccountAPI and UsersAPI. Let me know if I should add a test anyway.

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/3615

Yes

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -1443,7 +1443,7 @@ App::patch('/v1/account/name')
 
         $user = $dbForProject->updateDocument('users', $user->getId(), $user
             ->setAttribute('name', $name)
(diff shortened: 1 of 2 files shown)


stnguyen90:
@Meldiron, would it make more sense to implement this as filter rather than in so many different places? 🧐

eldadfux:
> @Meldiron, would it make more sense to implement this as filter rather than in so many different places? 🧐

We should and can do that in a separate PR.

PR #3634, opened 2022-08-02
Implement Function Variable API

PineappleIOnic:
## What does this PR do?

This PR moves the API over to the function's namespace and implements the UI for it.

image

image

image

Yes


composer.json
```diff
@@ -76,7 +76,7 @@
         }
     ],
     "require-dev": {
-        "appwrite/sdk-generator": "0.19.5",
+        "appwrite/sdk-generator": "0.18.8",
```
PineappleIOnic, line 79: note to self, put sdk-generator back up to normal version once console SDK patch is implemented

app/controllers/api/functions.php
```diff
@@ -1114,3 +1114,201 @@
 
         $response->noContent();
     });
+
+// Variables
+
+App::post('/v1/functions/variables/:functionId')
+    ->desc('Create Variable')
+    ->groups(['api', 'functions'])
```
eldadfux, line 1122: We miss the `sdk.description` label with a link to a markdown file on all new endpoints.

app/controllers/api/functions.php
```diff
@@ -911,8 +908,11 @@
             return $response->dynamic($execution, Response::MODEL_EXECUTION);
         }
 
-        /** Collect environment variables */
-        $vars = \array_merge($function->getAttribute('vars', []), [
+        $variables = $dbForProject->find('variables', [
+            new Query('functionInternalId', Query::TYPE_EQUAL, [$function->getInternalId()]),
+        ], 5000);
```
eldadfux, line 913: We can use the `APP_LIMIT_COUNT` constant here.

eldadfux (changes requested):
Not sure the UI works best here as well, we don't need the web icon. @TorstenDittmann @sara-k-48 could you please help improve this area?

app/controllers/api/functions.php
```diff
@@ -911,8 +915,17 @@
             return $response->dynamic($execution, Response::MODEL_EXECUTION);
         }
 
-        /** Collect environment variables */
-        $vars = \array_merge($function->getAttribute('vars', []), [
+        $vars = [];
+
+        $variables = $dbForProject->find('variables', [
```
eldadfux, line 920: We should be able to get this data from the function object, using find is much more expensive.

app/workers/builds.php
```diff
@@ -143,7 +144,15 @@ protected function buildDeployment(Document $project, Document $function, Docume
         );
 
         $source = $deployment->getAttribute('path');
-        $vars = $function->getAttribute('vars', []);
+        $vars = [];
+        $variables = $dbForProject->find('variables', [
```
eldadfux, line 148: Same, vars should already be available in the function document.

app/workers/functions.php
```diff
@@ -257,8 +258,18 @@ private function execute(
         $execution->setAttribute('status', 'processing');
         $execution = $dbForProject->updateDocument('executions', $executionId, $execution);
 
+        $vars = [];
+
+        $variables = $dbForProject->find('variables', [
```
eldadfux, line 263: Same, variables should be available in the function document with the new subquery filter.

app/workers/builds.php
```diff
@@ -179,6 +188,7 @@ protected function buildDeployment(Document $project, Document $function, Docume
             if ($deployment->getAttribute('activate') === true) {
                 $function->setAttribute('deployment', $deployment->getId());
                 $function = $dbForProject->updateDocument('functions', $function->getId(), $function);
+                $dbForProject->deleteCachedDocument('functions', $function->getId());
```
eldadfux, line 191: I think we should delete the cache only when we update a variable. When updating the function itself the cache should be automatically cleaned.

eldadfux: Please recheck all usage of this method in the PR.

src/Appwrite/Utopia/Response/Model/Func.php
```diff
@@ -62,10 +62,11 @@ public function __construct()
                 'example' => '5e5ea5c16897e',
             ])
             ->addRule('vars', [
-                'type' => self::TYPE_JSON,
+                'type' => Response::MODEL_VARIABLE,
                 'description' => 'Function environment variables.',
-                'default' => new \stdClass(),
-                'example' => ['key' => 'value'],
+                'default' => [],
+                'example' => new \stdClass(),
```
eldadfux, line 68: Example should be array not stdClass.

src/Appwrite/Utopia/Response/Model/Variable.php
```diff
@@ -0,0 +1,72 @@
...
+        $this
+            ->addRule('$id', [
+                'type' => self::TYPE_STRING,
+                'description' => 'Variable ID.',
+                'default' => '',
+                'example' => '5e5ea5c16897e',
+            ])
+            ->addRule('$createdAt', [
+                'type' => self::TYPE_INTEGER,
+                'description' => 'Variable creation date in Unix timestamp.',
```
eldadfux, line 21: We should update the format once merge with our new Dates PR.

src/Appwrite/Utopia/Response/Model/Variable.php
```diff
@@ -0,0 +1,72 @@
...
+                'default' => 0,
+                'example' => 1592981250,
+            ])
+            ->addRule('$updatedAt', [
+                'type' => self::TYPE_INTEGER,
```
eldadfux, line 26: We should update the format once merge with our new Dates PR.

eldadfux (changes requested):

PineappleIOnic:
image

PineappleIOnic:
image

docs/references/functions/create-variable.md
```diff
@@ -0,0 +1 @@
+Create a new function variable. These variables can be accessed within functions using the `env` object recieved through the payload of an execution.
```
gewenyu99, line 1:
```suggestion
Create a new function variable. These variables can be accessed within functions using the `env` object received through the payload of an execution.
```

gewenyu99: Do we also have a section we can link to in functions doc?

src/Appwrite/Utopia/Response/Model/Func.php
```diff
@@ -62,10 +62,11 @@ public function __construct()
                 'example' => '5e5ea5c16897e',
             ])
             ->addRule('vars', [
-                'type' => self::TYPE_JSON,
+                'type' => Response::MODEL_VARIABLE,
                 'description' => 'Function environment variables.',
```
gewenyu99, line 66: Let's decide on a word to describe these and be consistent. I say we call them `function variables` as you do in the endpoint descriptions.

src/Appwrite/Utopia/Response/Model/Variable.php
gewenyu99, line 21: Will this remain a unix timestamp, or will this become datetime?

gewenyu99 (changes requested):

src/Appwrite/Utopia/Response/Model/Variable.php
eldadfux: @PineappleIOnic we need to update this PR to work with the new datetime PR.

docs/references/functions/create-variable.md
eldadfux: @gewenyu99 This is the closest thing we have: https://appwrite.io/docs/functions#environmentVariables
I think we do miss a line or two about adding your own variables.

src/Appwrite/Utopia/Response/Model/Variable.php
PineappleIOnic: Yeah, Currently waiting on it to be stable before merging datetime into this PR

Meldiron: Updated to datetime.

app/controllers/api/functions.php
```diff
@@ -1290,3 +1296,234 @@
...
+    ->desc('Delete Variable')
+    ->groups(['api', 'functions'])
+    ->label('scope', 'functions.write')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'functions')
+    ->label('sdk.method', 'deleteVariable')
+    ->label('sdk.description', '/docs/references/functions/delete-variable.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_NOCONTENT)
+    ->label('sdk.response.model', Response::MODEL_NONE)
+    ->param('functionId', null, new UID(), 'Function unique ID.', false)
```
abnegate, line 1505: All ID params should default to an empty string, otherwise when not passed, a type error will be thrown due to trying to call the action with `null` where a non-nullable `string` is expected. 

Looks like this might be a general issue not just in this controller

tests/e2e/Services/Functions/FunctionsConsoleClientTest.php
```diff
@@ -95,10 +90,184 @@ public function testGetCollectionUsage(array $data)
...
+
+        $response = $this->client->call(Client::METHOD_GET, '/functions/' . $data['functionId'] . '/variables', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()));
+
+        $this->assertEquals(200, $response['headers']['status-code']);
+        $this->assertEquals(0, sizeof($response['body']['variables']));
+
+        return $data;
```
abnegate, line 271: Could we add some more tests for failure cases here? 

- Create variable with key/value too large
- Get variable that doesn't exist
- Update variable that doesn't exist
- Update variable with key/value too large
- Delete variable that doesn't exist

abnegate (changes requested):

app/config/collections.php
```diff
@@ -3381,6 +3381,99 @@
...
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => 'value',
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 16384,
```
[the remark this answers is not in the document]
Meldiron: Size of `vars` does not really matter, it will always be `null`. I set that one to 16k since all other subqueries uses 16k. not sure why.

Good point tho, previously value was 8k, not 16k. So I will decrease this to `8192`.

app/controllers/api/functions.php
```diff
@@ -79,13 +80,13 @@
...
             'runtime' => $runtime,
             'deployment' => '',
-            'vars' => $vars,
             'events' => $events,
             'schedule' => $schedule,
             'schedulePrevious' => null,
             'scheduleNext' => null,
             'timeout' => $timeout,
             'search' => implode(' ', [$functionId, $name, $runtime]),
+            'vars' => null
```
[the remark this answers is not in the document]
Meldiron: We only need to set vars to null in `documentCreate` - this is the only documentCreate for functions.

During update, if we never touch vars, they will remain null - meaning subqueries takes place.

app/controllers/api/functions.php
Meldiron: General issue. Making myself note, will address it in a separate PR in future.

tests/e2e/Services/Functions/FunctionsConsoleClientTest.php
Meldiron: Added failure tests.

tests/e2e/Services/Functions/FunctionsConsoleClientTest.php
```diff
@@ -95,10 +90,283 @@ public function testGetCollectionUsage(array $data)
...
+         */
+
+        $response = $this->client->call(Client::METHOD_GET, '/functions/' . $data['functionId'] . '/variables', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()), [
+            'queries' => [ 'equal("value", "MY_SECRET")' ]
+        ]);
+
+        $this->assertEquals(400, $response['headers']['status-code']);
```
abnegate, line 241: I think this should be a 200 with an empty list :wink:

Meldiron: This is 400 because you cannot query `value` attribute. I added a new test for querying `key` with non-existing key. That results in 200 but 0 results.

app/controllers/api/functions.php
christyjacob4, line 89: we dont need that here if its a subquery 👍

app/controllers/api/functions.php
```diff
@@ -1049,8 +1048,15 @@
...
 
-        /** Collect environment variables */
-        $vars = \array_merge($function->getAttribute('vars', []), [
+        $vars = [];
+
+        $variables = $function['vars'];
+
+        foreach ($variables as $variable) {
+            $vars[$variable['key']] = $variable['value'];
+        }
```
christyjacob4, line 1057:
```suggestion
        $variables = array_reduce($function['vars'] ?? [], function (array $carry, array $var) {
            $carry[$var['key']] = $var['value'];
            
            return $carry;
        }, []);
```

something like that makes it less noisy and not creates too many var/vars/variable variables :-D

app/controllers/api/functions.php
```diff
@@ -1290,3 +1296,231 @@
...
+
+        $queries = Query::parseQueries($queries);
+
+        if (!empty($search)) {
+            $queries[] = Query::search('search', $search);
+        }
+
+        // Get cursor document if there was a cursor query
+        $cursor = Query::getByType($queries, Query::TYPE_CURSORAFTER, Query::TYPE_CURSORBEFORE)[0] ?? null;
+        if ($cursor !== null) {
```
christyjacob4, line 1384:
```suggestion
        $cursor = reset(Query::getByType($queries, Query::TYPE_CURSORAFTER, Query::TYPE_CURSORBEFORE));
        if (!$cursor) {
```

app/controllers/api/functions.php
```diff
@@ -1290,3 +1296,231 @@
...
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_VARIABLE)
+    ->param('functionId', null, new UID(), 'Function unique ID.', false)
+    ->param('variableId', null, new UID(), 'Variable unique ID.', false)
+    ->param('key', null, new Text(255), 'Variable key. Max length: 255 chars.', true)
+    ->param('value', null, new Text(8192), 'Variable value. Max length: 8192 chars.', true)
+    ->inject('response')
+    ->inject('dbForProject')
+    ->action(function (string $functionId, string $variableId, ?string $key, ?string $value, Response $response, Database $dbForProject) {
+        if (empty($key) && empty($value)) {
```
christyjacob4, line 1456:
```suggestion
        if (empty($key) || empty($value)) {
```

should be like this right?

app/controllers/api/functions.php
```diff
@@ -1290,3 +1296,231 @@
...
+        }
+
+        $function = $dbForProject->getDocument('functions', $functionId);
+
+        if ($function->isEmpty()) {
+            throw new Exception(Exception::FUNCTION_NOT_FOUND);
+        }
+
+        $variable = $dbForProject->findOne('variables', [
+            Query::equal('_uid', [$variableId]),
```
christyjacob4, line 1467: we shouldn't use internal column names here, we should use `$id`

app/controllers/api/functions.php
```diff
@@ -1290,3 +1296,231 @@
...
+    ->inject('dbForProject')
+    ->action(function (string $functionId, string $variableId, Response $response, Database $dbForProject) {
+        $function = $dbForProject->getDocument('functions', $functionId);
+
+        if ($function->isEmpty()) {
+            throw new Exception(Exception::FUNCTION_NOT_FOUND);
+        }
+
+        $variable = $dbForProject->findOne('variables', [
+            Query::equal('_uid', [$variableId]),
```
christyjacob4, line 1514: we shouldnt use internal column names here, we should use `_uid`

app/workers/functions.php
```diff
@@ -259,8 +259,16 @@ private function execute(
...
         $execution = $dbForProject->updateDocument('executions', $executionId, $execution);
 
+        $vars = [];
+
+        $variables = $function['vars'];
+
+        foreach ($variables as $variable) {
+            $vars[$variable['key']] = $variable['value'];
+        }
+
```
christyjacob4, line 269: we should do the same as my suggestion in the controller to make this more safe

christyjacob4 (commented):
this is done by @TorstenDittmann at christys macbook

app/controllers/api/functions.php
Meldiron: Addressed in all places

app/controllers/api/functions.php
Meldiron: Doing `OR` would mean you need to provide both. What we check here is that at least one is provided. Notice this is an update endpoint, not create. I think the original is right. Right?

app/controllers/api/functions.php
Meldiron: Updated on multiple places.

app/workers/functions.php
Meldiron: Updated on multiple places.

app/controllers/api/functions.php
christyjacob4: I think this should be something like this 
```suggestion
        if (empty($key) || $value === null) {
```

TorstenDittmann: yeah, should be required 👍🏻

Meldiron: `key` is now required. Possible combinations:
- no input: 400
- value only: 400
- key only: 200, key updated, value the same
- key + value: 200, both updated

All cases got their own tests.

PR #3641, opened 2022-08-05
feat: add new attributes for API Keys

christyjacob4:
## What does this PR do?

Add new metadata to API Keys

## Test Plan

## Related PRs and Issues

(Write your answer here.)


app/controllers/general.php
```diff
@@ -300,6 +301,35 @@
...
+                        'appwrite:deno',
+                        'appwrite:php',
+                        'appwrite:python',
+                        'appwrite:ruby',
+                        'appwrite:go',
+                        'appwrite:java',
+                        'appwrite:dotnet',
+                        'appwrite:dart',
+                        'appwrite:kotlin',
+                        'appwrite:swift'
```
christyjacob4, line 318: @eldadfux What would be a better place to keep these constants ?

eldadfux: Not sure, a config might be too much. Let’s leave them here for now

christyjacob4: Alright 👍

app/controllers/general.php
```diff
@@ -300,6 +301,35 @@
 
                 Authorization::setRole('role:' . Auth::USER_ROLE_APP);
                 Authorization::setDefaultStatus(false);  // Cancel security segmentation for API keys.
+
+                if (time() > ($key->getAttribute('accessedAt', 0) + (24 * 60 * 60))) {
```
eldadfux, line 305: Let's use a constant for the 24 hours range. We might also want to wait for the dates PR.

app/controllers/general.php
```diff
@@ -300,6 +301,35 @@
...
+                    $key->setAttribute('accessedAt', time());
+                    $sdkValidator = new WhiteList([
+                        'appwrite:nodejs',
```
eldadfux, line 308: Do we want this prefix? Anyways, we need to validate this once all SDKs can report the new user-agent headers.

app/controllers/general.php
```diff
@@ -300,6 +301,35 @@
...
+                        'appwrite:ruby',
+                        'appwrite:go',
+                        'appwrite:java',
+                        'appwrite:dotnet',
+                        'appwrite:dart',
+                        'appwrite:kotlin',
+                        'appwrite:swift'
+                    ], true);
+    
+                    $sdk = $request->getHeader('x-sdk-version', 'UNKNOWN');
```
eldadfux, line 321: I think we might be able to add a set of specific headers to the sdks like
* x-sdk-name
* x-sdk-version

Which will make parsing easier. We can use the full combined string for the standard `user-agent` header. WDYT?

app/controllers/general.php
```diff
@@ -300,6 +301,35 @@
...
+                    $sdk = substr($sdk, 0, strrpos($sdk, ':'));
+
+                    if ($sdkValidator->isValid($sdk)) {
+                        $sdks = $key->getAttribute('sdks', []);
+                        array_push($sdks, $sdk);
```
eldadfux, line 326: This method will get the array bigger every time.

Do we also want to separate the 24 hours check and this check? Otherwise, we could potentially miss some SDKs. We could perform writes once every 24 hours for the access time, and every time a new (allowed) SDK is discovered. This could ensure min writes and max data integrity.

app/controllers/general.php
```diff
@@ -300,6 +301,35 @@
...
+                        $key->setAttribute('sdks', $sdks);
+                    }
+
+                    $dbForConsole->updateDocument('keys', $key->getId(), $key);
+                    $dbForConsole->deleteCachedDocument('projects', $project->getId());
```
eldadfux, line 331: Hopefully this line could be removed when we move all cache cleaning logic for documents that are using subqueries into the DB filters.

src/Appwrite/Utopia/Response/Model/Key.php
```diff
@@ -58,6 +58,18 @@ public function __construct()
                 'default' => '',
                 'example' => '919c2d18fb5d4...a2ae413da83346ad2',
             ])
+            ->addRule('accessedAt', [
+                'type' => self::TYPE_INTEGER,
+                'description' => 'Most recent access date in Unix timestamp.',
+                'default' => null,
+                'example' => '1653990687',
+            ])
+            ->addRule('sdks', [
```
eldadfux, line 67: This one should be marked as an array.

app/controllers/general.php
christyjacob4: done 👍

app/controllers/general.php
christyjacob4: yes! that would remove all the parsing logic.

app/controllers/general.php
christyjacob4: Yes agreed 👍

src/Appwrite/Utopia/Response/Model/Key.php
christyjacob4: updated 👍

app/controllers/general.php
christyjacob4: removed the prefix👍

app/controllers/general.php
christyjacob4: > This method will get the array bigger every time.
good catch. Fixed it 👍

christyjacob4: > We could perform writes once every 24 hours for the access time, and every time a new (allowed) SDK is discovered. This could ensure min writes and max data integrity.

If we want to write every time a new SDK is used, it makes sense to update the access time as well. Because its a single write operation after all

eldadfux: We can do that.

app/config/collections.php
```diff
@@ -972,6 +972,28 @@
                 'array' => false,
                 'filters' => [],
             ],
+            [
+                '$id' => 'accessedAt',
+                'type' => Database::VAR_INTEGER,
```
eldadfux, line 977: Reminder to wait for the dates PR and change this.

app/controllers/api/projects.php
```diff
@@ -841,6 +841,8 @@
             'name' => $name,
             'scopes' => $scopes,
             'expire' => $expire,
+            'sdks' => [],
+            'accessedAt' => 0,
```
eldadfux, line 845: We need to change this default value after the dates PR.

app/controllers/general.php
```diff
@@ -300,6 +302,27 @@
 
                 Authorization::setRole('role:' . Auth::USER_ROLE_APP);
                 Authorization::setDefaultStatus(false);  // Cancel security segmentation for API keys.
+
+                if (time() > $key->getAttribute('accessedAt', 0) + APP_KEY_ACCCESS) {
```
eldadfux, line 306: We need to change this condition after the dates PR.

composer.json
```diff
@@ -76,7 +76,7 @@
         }
     ],
     "require-dev": {
-        "appwrite/sdk-generator": "0.19.5",
+        "appwrite/sdk-generator": "dev-feat-new-headers",
```
eldadfux, line 79: Reminder: Update this version.

src/Appwrite/Utopia/Response/Model/Key.php
```diff
@@ -58,6 +58,19 @@ public function __construct()
                 'default' => '',
                 'example' => '919c2d18fb5d4...a2ae413da83346ad2',
             ])
+            ->addRule('accessedAt', [
+                'type' => self::TYPE_INTEGER,
+                'description' => 'Most recent access date in Unix timestamp.',
```
eldadfux, line 63: Update date format after dates PR.

eldadfux (approved):

composer.json
christyjacob4: We need to release a new SDK generator version first 👍

src/Appwrite/Utopia/Response/Model/Key.php
christyjacob4: Updated 👍

Meldiron (changes requested):

src/Appwrite/Utopia/Response/Model/User.php
```diff
@@ -65,7 +65,7 @@ public function __construct()
             ->addRule('registration', [
                 'type' => self::TYPE_DATETIME,
                 'description' => 'User registration date in Datetime.',
-                'default' => null,
+                'default' => '',
```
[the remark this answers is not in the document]
christyjacob4: yes . with the new datetime changes, the default for all datetime attributes is an empty string

app/controllers/mock.php
```diff
@@ -395,6 +396,35 @@
         $response->noContent();
     });
 
+/** Endpoint to test if required headers are sent from the SDK */
```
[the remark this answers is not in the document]
christyjacob4: added them 👍

eldadfux:
I love this instant features, great work @christyjacob4 !

PR #3652, opened 2022-08-09
Feat audits label

shimonewman:
## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)


docker-compose.yml
```diff
@@ -103,6 +103,7 @@ services:
       - ./phpunit.xml:/usr/src/code/phpunit.xml
       - ./tests:/usr/src/code/tests
       - ./app:/usr/src/code/app
+      #- ./vendor:/usr/src/code/vendor #TODO remove when done!!
```
eldadfux, line 106: Reminder

app/controllers/api/teams.php
```diff
@@ -36,6 +36,8 @@
     ->groups(['api', 'teams'])
     ->label('event', 'teams.[teamId].create')
     ->label('scope', 'teams.write')
+    ->label('audits.resource', 'team/{response.$id}')
+    ->label('audits.payload', true)
```
eldadfux, line 40: No need for this, can be enabled by default _for now_.

eldadfux (changes requested):
We should add logs to functions.

christyjacob4:
@shimonewman please sync this branch with 0.16.x and change the base branch to 0.16.x as well 👍

app/controllers/api/account.php
```diff
@@ -1068,6 +1049,8 @@
     ->label('event', 'users.[userId].sessions.[sessionId].create')
     ->label('scope', 'public')
     ->label('auth.type', 'anonymous')
+    ->label('audits.resource', 'user/{response.userId}')
+    ->label('audits.userId', '{userId}')
```
eldadfux, line 1053:
```suggestion
    ->label('audits.userId', '{user.$id}')
```

This is what I had in mind for consistency.

app/controllers/api/account.php
```diff
@@ -2092,6 +2045,8 @@
     ->groups(['api', 'account'])
     ->label('scope', 'account')
     ->label('event', 'users.[userId].verification.[tokenId].create')
+    ->label('audits.resource', 'user/{response.userId}')
+    ->label('audits.userId', '{userId}')
```
eldadfux, line 2049: Same as above.

app/controllers/api/functions.php
```diff
@@ -447,13 +451,14 @@
     ->desc('Create Deployment')
     ->label('scope', 'functions.write')
     ->label('event', 'functions.[functionId].deployments.[deploymentId].create')
+    ->label('audits.resource', 'function/{request.functionId}/deployment/{response.$id}')
```
eldadfux, line 454: I think we can save all deployments events on the main function resource no need to nest down.

app/controllers/api/functions.php
```diff
@@ -1063,6 +1069,7 @@
     ->desc('Retry Build')
     ->label('scope', 'functions.write')
     ->label('event', 'functions.[functionId].deployments.[deploymentId].update')
+    ->label('audits.resource', 'function/{request.functionId}/deployment/{request.deploymentId}/build/{request.buildId}')
```
eldadfux, line 1072: Again, all logs can be saved on the functions scope.
```suggestion
    ->label('audits.resource', 'function/{request.functionId}')
```

eldadfux (changes requested):

app/controllers/shared/api.php
```diff
@@ -255,7 +251,64 @@
             }
         }
 
+        $route = $utopia->match($request);
+        $requestParams = array_combine(
+            array_keys($route->getParams()), array_column($route->getParams(), 'value')
+        );
```
lohanidamodar, line 257: @shimonewman this will not work, we need the previous code, the route matched, doesn't have value field, that's only calculated during `execute` phase. so value here is always `null`

lohanidamodar: The values is in `$utopia->$matches` which is not exposed, may be we should expose the matches? @eldadfux what od you think?

eldadfux: With the latest framework changes the values should be populated. See this line: https://github.com/utopia-php/framework/blob/master/src/App.php#L644

Because we access the Route object in the shutdown every relevant variable should be accessible. Am I missing something?

lohanidamodar: Never mind, I was using the old version of the framework. @eldadfux @shimonewman All good.

shimonewman: Cool

eldadfux (approved):

tests/e2e/Services/Webhooks/WebhooksCustomServerTest.php
```diff
@@ -289,6 +289,7 @@ public function testUpdateUserPrefs(array $data): array
         $this->assertEquals($webhook['headers']['X-Appwrite-Webhook-Id'] ?? '', $this->getProject()['webhookId']);
         $this->assertEquals($webhook['headers']['X-Appwrite-Webhook-Project-Id'] ?? '', $this->getProject()['$id']);
         $this->assertEquals(empty($webhook['headers']['X-Appwrite-Webhook-User-Id'] ?? ''), ('server' === $this->getSide()));
+
```
christyjacob4, line 292:
```suggestion
```

PR #3656, opened 2022-08-10
Functions Console Logs Capture

lohanidamodar:
## What does this PR do?

- Captures and stores console logs from user functions in runtimes

## Test Plan

- Tested in https://github.com/open-runtimes/open-runtimes/pull/74

## Related PRs and Issues
(need this released with v2 runtimes tag so we can run  CI tests)
- https://github.com/open-runtimes/open-runtimes/pull/74

YES


app/executor.php
```diff
@@ -276,7 +276,7 @@ function getStorageDevice($root): Device
                 throw new Exception('Failed to create build container', 500);
             }
 
-            $orchestration->networkConnect($runtimeId, App::getEnv('OPEN_RUNTIMES_NETWORK', 'appwrite_runtimes'));
+            $orchestration->networkConnect($runtimeId, App::getEnv('OPEN_RUNTIMES_NETWORK', 'appwrite-runtimes'));
```
christyjacob4, line 279: Can you test this locally as well ? This was working perfectly fine before ( even in the current Appwrite 0.15 ) and I'm sceptical about changing this.

christyjacob4 (commented):
Looks good @lohanidamodar . 

I would also like this to be tested in a concurrent setting as @stnguyen90 mentioned. You can have a simple script to create a series of simultaneously executions.

Within the cloud function you can console.log the current timestamp. 
Within the logs, you can then check that no two executions have the same timestamp

app/views/console/functions/function.phtml
```diff
@@ -424,21 +424,37 @@ sort($patterns);
...
 
-                                                    <h1>STDOUT</h1>
+                                                    <h1>RESPONSE</h1>
 
                                                     <div class="margin-bottom ide" data-ls-if="({{execution.response.length}})">
                                                         <pre data-ls-bind="{{execution.response}}"></pre>
                                                     </div>
 
                                                     <div class="margin-bottom" data-ls-if="(!{{execution.response.length}})">
+                                                        <p>No Response was logged.</p>
```
christyjacob4, line 444: Can you add some screenshots of how it looks??

lohanidamodar: will do

app/executor.php
lohanidamodar: I can test, but even if it worked, it's wrong and network name is not `appwrite_runtimes` but `appwrite-runtimes`. It might have worked without assigning the proper network. but we need to make sure we assign the proper network

app/views/console/functions/function.phtml
lohanidamodar: image

lohanidamodar: image

lohanidamodar: when clicked on the Output button
image

eldadfux:
Let's give the links a bit more space so they can be in one line.
image

We should also change:
`Output` -> Stdout
`Errors` -> Stderr

eldadfux:
Also in the table `Runtime` should be just `Time`

image

app/controllers/api/functions.php
```diff
@@ -945,6 +945,7 @@
             $execution->setAttribute('status', $executionResponse['status']);
             $execution->setAttribute('statusCode', $executionResponse['statusCode']);
             $execution->setAttribute('response', $executionResponse['response']);
+            $execution->setAttribute('stdout', $executionResponse['stdout']);
```
eldadfux, line 948: Do we want to expose stdout and stderr to the client side API? It might contain sensitive information. What attribute are we using to return logs data?

lohanidamodar:
> Let's give the links a bit more space so they can be in one line. image
> 
> We should also change: `Output` -> Stdout `Errors` -> Stderr

The output is actual response
log is the stdout
and error is the stderr

app/controllers/api/functions.php
lohanidamodar: Whether or not to send the logs to client is a decision needed to be made and how to handle it. 
- stdout = logs
- response = users code's response
- stderr = errors

app/executor.php
lohanidamodar: @christyjacob4 figured out the problem, there was a bug with Docker compose 2.8.0 which was what gitpod was using at the time, which has been reverted and fixed in docker-compose 2.9.0 so I will revert back this change.
https://github.com/docker/compose/releases/tag/v2.8.0

app/controllers/api/functions.php
lohanidamodar: @eldadfux renamed but was unable to fix the column even with long column width, column just overflows but the buttons still not align
image

eldadfux:
> > Let's give the links a bit more space so they can be in one line. image
> > We should also change: `Output` -> Stdout `Errors` -> Stderr
> 
> The output is actual response log is the stdout and error is the stderr

We should fix it, but if that's the case where are the logs?

lohanidamodar:
> > > Let's give the links a bit more space so they can be in one line. image
> > > We should also change: `Output` -> Stdout `Errors` -> Stderr
> > 
> > 
> > The output is actual response log is the stdout and error is the stderr
> 
> We should fix it, but if that's the case where are the logs?

Not sure what you mean, here
1. Output button (renamed to `Response`) => shows the response from user's code (execution.response)
2. Logs (Renamed to `Stdout`)  => shows the user's code's console logs (execution.stdout)
3. error button (Renamed to `Stderr`) => shows the execution errors (execution.stderr)

lohanidamodar:
> Looks good @lohanidamodar .
> 
> I would also like this to be tested in a concurrent setting as @stnguyen90 mentioned. You can have a simple script to create a series of simultaneously executions.
> 
> Within the cloud function you can console.log the current timestamp. Within the logs, you can then check that no two executions have the same timestamp

@christyjacob4 I will try, however as our runtimes are web server, which should handle concurrent requests already in correct sequence and output is buffered inside the execution not from the docker logs so we should be safe as far s I know.

christyjacob4:
@lohanidamodar we can potentially avoid patch releases if we test this scenario well... 
So let's test it instead of giving rise to bugs post release. 👍

app/views/console/functions/function.phtml
christyjacob4: Looks good.. Just the placement of the links on the UI need to be fixed

src/Appwrite/Utopia/Response/Model/Execution.php
```diff
@@ -65,6 +65,12 @@ public function __construct()
                 'default' => '',
                 'example' => '',
             ])
+            ->addRule('stdout', [
```
christyjacob4, line 68: We need to hide this attribute for client side sdks 🤔

christyjacob4 (changes requested):
The PR looks good. Just need to make final changes to the UI and try and hide the `stdout` attribute from the client response

app/controllers/api/functions.php
```diff
@@ -1023,6 +1023,16 @@
         $results = $dbForProject->find('executions', $queries, $limit, $offset, [], [Database::ORDER_DESC], $cursorExecution ?? null, $cursorDirection);
         $total = $dbForProject->count('executions', $queries, APP_LIMIT_COUNT);
 
+        $roles = Authorization::getRoles();
+        $isPrivilegedUser = Auth::isPrivilegedUser($roles);
+        $isAppUser = Auth::isAppUser($roles);
+        if ($isPrivilegedUser || $isAppUser) {
```
eldadfux, line 1029: I think the condition here should be opposite. I think this part is important enough for having its own tests.

christyjacob4:
Related 
1. https://github.com/appwrite/runtimes/releases/tag/0.11.0
2. https://github.com/open-runtimes/open-runtimes/releases/tag/0.1.3

app/controllers/api/functions.php
christyjacob4: Yes.. 👍

app/controllers/api/functions.php
```diff
@@ -965,6 +966,12 @@
             ->setParam('functionStatus', $execution->getAttribute('status', ''))
             ->setParam('functionExecutionTime', $execution->getAttribute('time') * 1000); // ms
 
+        $roles = Authorization::getRoles();
+        $isPrivilegedUser = Auth::isPrivilegedUser($roles);
+        $isAppUser = Auth::isAppUser($roles);
+        $execution->setAttribute('stdout', ($isPrivilegedUser || $isAppUser) ? $executionResponse['stdout'] : '');
+        $execution->setAttribute('stderr', ($isPrivilegedUser || $isAppUser) ? $executionResponse['stderr'] : '');
```
christyjacob4, line 973: Let's be consistent with the way we set these attributes.

In the listExecutions endpoint, we use an if condition. Let's stick to the same here as well for consistency

app/controllers/api/functions.php
```diff
@@ -1055,6 +1073,12 @@
             throw new Exception('Execution not found', 404, Exception::EXECUTION_NOT_FOUND);
         }
 
+        $roles = Authorization::getRoles();
+        $isPrivilegedUser = Auth::isPrivilegedUser($roles);
+        $isAppUser = Auth::isAppUser($roles);
+        $execution->setAttribute('stdout', ($isPrivilegedUser || $isAppUser) ? $execution->getAttribute('stdout') : '');
+        $execution->setAttribute('stderr', ($isPrivilegedUser || $isAppUser) ? $execution->getAttribute('stderr') : '');
```
christyjacob4, line 1080: Same comment as above

christyjacob4 (changes requested):
Looks good. Just some final changes and then we can merge. 

We also need to update the appwrite/runtimes version in this file

app/executor.php
christyjacob4: awesome 👍

christyjacob4 (approved):
Awesome work 👌

src/Appwrite/Utopia/Response/Model/Execution.php
Meldiron: It is set to an empty string on controller-level before returning a response. Should be in all places, I double-checked.

app/executor.php
```diff
@@ -538,13 +540,19 @@ function (string $runtimeId, array $vars, string $data, $timeout, $activeRuntime
...
-                    $stderr = $executorResponse ?? 'Internal Runtime error.';
+                    $stderr = ($executorResponse ?? [])['stderr'] ?? 'Internal Runtime error.';
+                    $stdout = ($executorResponse ?? [])['stdout'] ?? 'Internal Runtime error.';
                     break;
                 case $statusCode >= 100:
-                    $stdout = $executorResponse;
+                    $stdout = $executorResponse['stdout'];
+                    $res = $executorResponse['response'];
+                    if (is_array($res)) {
+                        $res = json_encode($res);
```
Meldiron, line 550: When is $res array? Function must always return string, right?

app/executor.php
```diff
@@ -557,7 +565,8 @@ function (string $runtimeId, array $vars, string $data, $timeout, $activeRuntime
             $execution = [
                 'status' => $functionStatus,
                 'statusCode' => $statusCode,
-                'response' => \mb_strcut($stdout, 0, 1000000), // Limit to 1MB
+                'response' => \mb_strcut($res, 0, 1000000), // Limit to 1MB
+                'stdout' => \mb_strcut($stdout, 0, 1000000), // Limit to 1MB
                 'stderr' => \mb_strcut($stderr, 0, 1000000), // Limit to 1MB
```
Meldiron, line 570: We are losing emojis again.

```
appwrite-executor  | string(46) "{"message":"Hello from Appwrite \ud83d\udc4b"}"
```

Meldiron: Fixed

app/executor.php
Meldiron: I miss-understood. It is correct.

tests/e2e/Services/Functions/FunctionsCustomClientTest.php
```diff
@@ -397,6 +397,9 @@ public function testSynchronousExecution(): array
         $this->assertEquals($this->getUser()['$id'], $output['APPWRITE_FUNCTION_USER_ID']);
         $this->assertNotEmpty($output['APPWRITE_FUNCTION_JWT']);
         $this->assertEquals($projectId, $output['APPWRITE_FUNCTION_PROJECT_ID']);
+        // Client should never see logs and errors
+        $this->assertEmpty($execution['body']['stdout']);
+        $this->assertEmpty($execution['body']['stderr']);
```
christyjacob4, line 402: Can we ensure these keys are not even present in the array ? and not just check for empty ?

christyjacob4 (approved):
Looks good. Just one comment.

PR #3668, opened 2022-08-12
Refactor permissions

abnegate:
## What does this PR do?

- [BREAKING] Introduce new permissions syntax replacing write permission with create, update and delete.
- [BREAKING] Remove $read attribute from Documents
- [BREAKING] Remove $write attribute from Documents
- Add $permissions array attribute to Documents
- WIP Update UI to account for new permissions
- Update all ID usages to use new ID helper

## Test Plan

Existing tests + new tests to cover changed functionality

## Related PRs and Issues

https://github.com/utopia-php/database/pull/161
https://github.com/utopia-php/abuse/pull/22


app/controllers/api/databases.php
```diff
@@ -1869,43 +1865,35 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
...
         }
 
-        $data['$collection'] = $collection->getId(); // Adding this param to make API easier for developers
-        $data['$id'] = $documentId == 'unique()' ? $dbForProject->getId() : $documentId;
-        $data['$read'] = (is_null($read) && !$user->isEmpty()) ? ['user:' . $user->getId()] : $read ?? []; //  By default set read permissions for user
-        $data['$write'] = (is_null($write) && !$user->isEmpty()) ? ['user:' . $user->getId()] : $write ?? []; //  By default set write permissions for user
-
-        // Users can only add their roles to documents, API keys and Admin users can add any
-        $roles = Authorization::getRoles();
+        $permissions = PermissionsProcessor::addDefaultsIfNeeded(
```
eldadfux, line 1873: Not sure I fully understand everything that goes here, seems like a lot of added complexity that I think we could try and reduce with abstraction and interfaces that explain the logic better.

eldadfux: Could this potentially be part of the validation logic. Would love some description in a docblock.

app/controllers/api/databases.php
```diff
@@ -2231,24 +2213,31 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
...
-    ->action(function (string $databaseId, string $collectionId, string $documentId, string|array $data, ?array $read, ?array $write, Response $response, Database $dbForProject, EventAudit $audits, Stats $usage, Event $events, string $mode) {
+    ->action(function (string $databaseId, string $collectionId, string $documentId, string|array $data, ?array $permissions, Response $response, Database $dbForProject, EventAudit $audits, Stats $usage, Event $events, string $mode) {
+
+        $data = (\is_string($data)) ? \json_decode($data, true) : $data; // Cast to JSON array
+
+        if (empty($data) && empty($permissions)) {
+            throw new Exception('Missing payload or permissions', 400, Exception::DOCUMENT_MISSING_PAYLOAD);
+        }
+
+        if (!\is_array($data)) {
```
eldadfux, line 2231: This is contradicting to the logic at the beginning of the action, not sure why we have it from the first place, seems like legacy.

app/controllers/api/databases.php
```diff
@@ -2231,24 +2213,31 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
...
     ->inject('audits')
     ->inject('usage')
     ->inject('events')
     ->inject('mode')
-    ->action(function (string $databaseId, string $collectionId, string $documentId, string|array $data, ?array $read, ?array $write, Response $response, Database $dbForProject, EventAudit $audits, Stats $usage, Event $events, string $mode) {
+    ->action(function (string $databaseId, string $collectionId, string $documentId, string|array $data, ?array $permissions, Response $response, Database $dbForProject, EventAudit $audits, Stats $usage, Event $events, string $mode) {
+
+        $data = (\is_string($data)) ? \json_decode($data, true) : $data; // Cast to JSON array
+
+        if (empty($data) && empty($permissions)) {
```
eldadfux, line 2227: Why are we checking both in the same block with the same error? Can we split it to two checks with different errors? I was also under the assumption we allow no permissions.

app/controllers/api/storage.php
```diff
@@ -5,6 +5,7 @@
 use Appwrite\Event\Audit;
```
eldadfux, line 5: Let's make sure that whatever changes we implement on the database controller we also implement on storage.

public/scripts/services/form.js
```diff
@@ -3,9 +3,9 @@
 
     window.ls.container.set('form', function () {
 
-        function cast(value, to) {
+        function cast(value, from, to,) {
```
eldadfux, line 6: What exactly was the problem here? This is a risky change as it affects most forms on the console, let's validate carefully.

eldadfux (changes requested):
@abnegate great work, left my comments. Please add UI screenshots / videos. We should also change `role:app` to `apps`

src/Appwrite/Utopia/Response/Model/Collection.php
```diff
@@ -60,11 +53,11 @@ public function __construct()
                 'default' => true,
                 'example' => false,
             ])
-            ->addRule('permission', [
-                'type' => self::TYPE_STRING,
-                'description' => 'Collection permission model. Possible values: `document` or `collection`',
+            ->addRule('documentSecurity', [
```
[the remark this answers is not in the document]
abnegate: Just for the attribute name or for the UI as well? `security` by itself sounds like it would be dangerous to disable, even though when disabled collection/bucket permissions would still apply

src/Appwrite/Permissions/PermissionsProcessor.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+namespace Appwrite\Permissions;
+
+use Appwrite\Auth\Auth;
+use Utopia\Database\Database;
+use Utopia\Database\Validator\Authorization;
+
+class PermissionsProcessor
```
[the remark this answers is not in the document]
abnegate: This was to abstract some of the common permissions logic into a reusable helper, should we just replicate the code where needed instead?

app/controllers/api/databases.php
abnegate: Yep looks like legacy, will remove

app/controllers/api/databases.php
abnegate: The expectation was that developers need to supply _either_ data or permissions to update, but the `empty($permissions)` check should be `is_null($permissions)` to allow empty permissions, will update

public/scripts/services/form.js
abnegate: The problem is when binding an array to a hidden input, it is being bound as comma separated string, so when passed to `cast`, it is cast to a single element array instead of splitting into individual elements

E.g given

```javascript
permissions = ['read("any")', 'create("any")'];
```

Would bind as:

`'read("any"),create("any")'`

Which would cast to:

`['read("any"),create("any")']`

Instead of the expected

`['read("any")', 'create("any")']`

The change should only have any effect when both `data-cast-to="array"` and `data-cast-from="csv"` are set on the same element

abnegate:
Current UI:

https://user-images.githubusercontent.com/5857008/184541458-591dd523-3fc9-41c3-98b6-e8ee4f11343c.mov

Screen Shot 2022-08-15 at 2 20 01 AM

app/controllers/api/databases.php
abnegate: This is adding the permissions for the current user for any that are missing on a document (and file) creation

E.g. if a user creates a document with only `"read("any")"` permissions, this will add "update("user:xxx")" and "delete("user:xxx")", this was existing logic, is it expected with the new model?

app/controllers/api/storage.php
```diff
@@ -635,25 +612,18 @@
...
-                        $file = Authorization::skip(fn () => $dbForProject->updateDocument('bucket_' . $bucket->getInternalId(), $fileId, $file));
-                    } else {
-                        $file = $dbForProject->updateDocument('bucket_' . $bucket->getInternalId(), $fileId, $file);
-                    }
+                    $file = $dbForProject->updateDocument('bucket_' . $bucket->getInternalId(), $fileId, $file);
                 }
             } catch (StructureException $exception) {
                 throw new Exception($exception->getMessage(), 400, Exception::DOCUMENT_INVALID_STRUCTURE);
-            } catch (DuplicateException $exception) {
+            } catch (DuplicateException) {
```
eldadfux, line 626: Is this syntax possible?! Why do even need to define the type without the variable?

app/controllers/api/storage.php
```diff
@@ -777,31 +737,30 @@
...
-            $file = $dbForProject->getDocument('bucket_' . $bucket->getInternalId(), $fileId);
-        }
+        $file = $dbForProject->getDocument('bucket_' . $bucket->getInternalId(), $fileId);
 
         if ($file->isEmpty() || $file->getAttribute('bucketId') !== $bucketId) {
             throw new Exception('File not found', 404, Exception::STORAGE_FILE_NOT_FOUND);
         }
 
+        if ($fileSecurity) {
+            $valid |= $validator->isValid($file->getRead());
```
eldadfux, line 758: What does `|=` do?

phpunit.xml
```diff
@@ -13,7 +13,7 @@
     </extensions>
     <testsuites>
         <testsuite name="unit">
-            <directory>./tests/unit/</directory>
```
eldadfux, line 16: I think this was added on purpose by @TorstenDittmann

src/Appwrite/Permissions/PermissionsProcessor.php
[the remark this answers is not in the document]
eldadfux: I don't remember we needed so much logic before hand, let's discuss later.

src/Appwrite/Utopia/Response.php
```diff
@@ -435,9 +435,13 @@ public function output(Document $document, string $model): array
                             throw new Exception('Missing model for rule: ' . $ruleType);
                         }
 
-                        $item = $this->output($item, $ruleType);
+                        $data[$key][$index] = $this->output($item, $ruleType);
                     }
                 }
+            } else {
```
eldadfux, line 441: Why did we make this change? Is it part of this PR context?

tests/unit/Utopia/Nested.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Tests\Unit\Utopia;
+
+use Appwrite\Utopia\Response\Model;
+
+class Nested extends Model
```
eldadfux, line 7: What exactly is this class?

eldadfux (changes requested):

app/controllers/api/databases.php
```diff
@@ -1929,43 +1929,39 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
...
-            foreach ($data['$write'] as $write) {
-                if (!Authorization::isRole($write)) {
-                    throw new Exception('Write permissions must be one of: (' . \implode(', ', $roles) . ')', 400, Exception::USER_UNAUTHORIZED);
-                }
-            }
+        if (!PermissionsProcessor::allowedForResourceType('document', $permissions)) {
+            throw new Exception('Invalid permission', 400, Exception::GENERAL_PERMISSION_INVALID);
+        }
+        if (!PermissionsProcessor::allowedForUserType($permissions)) {
+            throw new Exception('Permissions must be one of: (' . \implode(', ', Authorization::getRoles()) . ')', 400, Exception::USER_UNAUTHORIZED);
```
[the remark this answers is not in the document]
abnegate: - We first handle the aggregates (`write`, in future `admin`)
- Then we check the user did not pass any invalid permission for the resource type, as the validator allows `create/read/update/delete/...` but documents/files should only be allowed `read/update/delete`
- Then we check that the current user is allowed to add those permissions, e.g. auth key can add any users role, but session token can only add their own role

app/controllers/api/storage.php
abnegate: It must be, I got an IDE warning it was unused and could be removed 😅  in this case the type just defines which exception should be handled in that `catch` particular block

app/controllers/api/storage.php
abnegate: It is a bitwise or equals to avoid short-circuiting, in this case, if `$valid` is false, the right-hand side will still be evaluated, and then `OR`d with `$valid`, then the result will be stored in `$valid`

app/controllers/api/databases.php
```diff
@@ -505,32 +507,32 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
...
         $database = Authorization::skip(fn () => $dbForProject->getDocument('databases', $databaseId));
 
         if ($database->isEmpty()) {
             throw new Exception('Database not found', 404, Exception::DATABASE_NOT_FOUND);
         }
 
-        $collectionId = $collectionId == 'unique()' ? $dbForProject->getId() : $collectionId;
+        $collectionId = $collectionId == 'unique()' ? ID::unique() : $collectionId;
+
+        $permissions = PermissionsProcessor::handleAggregates($permissions);
```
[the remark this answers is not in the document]
eldadfux: I think the can and should be abstracted by the database library, what are other methods responsible for?

app/controllers/api/teams.php
```diff
@@ -807,6 +831,14 @@
             throw new Exception('Team not found', 404, Exception::TEAM_NOT_FOUND);
         }
 
+        /**
+         * Force document security
+         */
+        $validator = new Authorization('delete');
```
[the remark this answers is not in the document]
abnegate: Because the DB doesn't know about `documentSecurity`, we can't check at that level in case it is disabled

src/Appwrite/Utopia/Response.php
abnegate: I'm not sure where this change has come from 🤔 Will investigate

tests/unit/Utopia/Nested.php
abnegate: I'm not sure where this change has come from 🤔 Will investigate

src/Appwrite/Utopia/Response.php
abnegate: There are coming from `fix-unit-tests` which is already merged in `0.16.x`, shall I try remove? Or sync `0.16.x`?

tests/unit/Utopia/Nested.php
abnegate: There are coming from `fix-unit-tests` which is already merged in `0.16.x`, shall I try remove? Or sync `0.16.x`?

app/controllers/api/account.php
```diff
@@ -514,9 +525,9 @@
         $expire = DateTime::addSeconds(new \DateTime(), Auth::TOKEN_EXPIRATION_LOGIN_LONG);
 
         $session = new Document(array_merge([
-            '$id' => $dbForProject->getId(),
-            'userId' => $user->getId(),
-            'userInternalId' => $user->getInternalId(),
+            '$id' => ID::unique(),
+            'userId' => ID::custom($user->getId()),
+            'userInternalId' => ID::custom($user->getInternalId()),
```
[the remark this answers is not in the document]
abnegate: Yep I agree it doesn't make sense when passing existing resource IDs, will remove 👍

app/controllers/api/databases.php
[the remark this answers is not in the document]
abnegate: It was my understanding we wanted to avoid handling `write` (or any aggregate) at the database level and only handle it in the API? Not sure if the other methods could be handled at the DB level since they deal with Appwrite-specific concepts:

`addDefaultsIfNeeded`: Add permissions for the current user for any permission types that are missing

`allowedForUserType`: Check that the current user is allowed to add those permissions, e.g. auth key can add any users role, but session token can only add their own role

`allowedForResourceType`: Check the user did not pass any invalid permission for the resource type, as the validator allows create/read/update/delete/... but documents/files should only be allowed read/update/delete

tests/unit/Utopia/Nested.php
eldadfux: Sync with 0.16x

src/Appwrite/Utopia/Response.php
eldadfux: Sync with 0.16x

app/controllers/api/storage.php
```diff
@@ -697,19 +667,13 @@
 
         $bucket = Authorization::skip(fn () => $dbForProject->getDocument('buckets', $bucketId));
 
-        if (
-            $bucket->isEmpty()
-            || (!$bucket->getAttribute('enabled') && $mode !== APP_MODE_ADMIN)
-        ) {
+        if ($bucket->isEmpty() || (!$bucket->getAttribute('enabled') && $mode !== APP_MODE_ADMIN)) {
             throw new Exception('Bucket not found', 404, Exception::STORAGE_BUCKET_NOT_FOUND);
```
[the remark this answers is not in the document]
abnegate: This will be hard to do without merging that branch into here; as it is only merged in `0.16.x` not the base `date time-attributes` it will blow up the diff

src/Appwrite/Permissions/PermissionsProcessor.php
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Appwrite\Permissions;
+
+use Utopia\Database\Database;
+use Utopia\Database\Permission;
+
+class PermissionsProcessor
+{
+    public static function aggregate(?array $permissions, string $resource): ?array
+    {
```
eldadfux, line 11: Can this also be part of the original Permission class in the db library?

app/realtime.php
```diff
@@ -203,7 +203,7 @@ function () use ($register, $db, $redis) {
         /**
          * Sending current connections to project channels on the console project every 5 seconds.
          */
-        if ($realtime->hasSubscriber('console', 'role:member', 'project')) {
+        if ($realtime->hasSubscriber('console', 'users', 'project')) {
```
eldadfux, line 206: Can we use our constant here?

app/realtime.php
```diff
@@ -254,12 +254,12 @@ function () use ($register, $db, $redis) {
         /**
          * Sending test message for SDK E2E tests every 5 seconds.
          */
-        if ($realtime->hasSubscriber('console', 'role:guest', 'tests')) {
+        if ($realtime->hasSubscriber('console', 'guests', 'tests')) {
```
eldadfux, line 257: Can we use our constant here?

app/realtime.php
```diff
@@ -254,12 +254,12 @@ function () use ($register, $db, $redis) {
...
             $payload = ['response' => 'WS:/v1/realtime:passed'];
 
             $event = [
                 'project' => 'console',
-                'roles' => ['role:guest'],
+                'roles' => ['guests'],
```
eldadfux, line 262: Can we use our constant here?

src/Appwrite/Permissions/PermissionsProcessor.php
abnegate: I'm not sure it would fit as `Permission` deals with singular permissions, and this function relies on knowing about Appwrite specific `$resource` types internally

tests/unit/Messaging/MessagingTest.php
```diff
@@ -204,23 +207,28 @@ public function testFromPayloadCollectionLevelPermissions(): void
...
-                '$id' => 'collection',
-                '$read' => ['role:all'],
-                '$write' => ['role:all'],
-                'permission' => 'collection'
+                '$id' => ID::custom('test'),
+                '$collection' => ID::custom('collection'),
+                '$permissions' => [
+                    'read(admin)',
+                    'update(admin)',
+                    'delete(admin)',
```
christyjacob4, line 215: An `admin` permission ? I believe this was something we planned for future versions ?

app/controllers/api/storage.php
```diff
@@ -365,40 +376,66 @@
...
+                    $permissions[] = (new Permission($permission, 'user', $user->getId()))->toString();
                 }
             }
-            foreach ($write as $role) {
-                if (!Authorization::isRole($role)) {
-                    throw new Exception(Exception::USER_UNAUTHORIZED, 'Write permissions must be one of: (' . \implode(', ', $roles) . ')', 400);
+        } else {
+            foreach ($allowedPermissions as $permission) {
+                // If the permission is not set, add it for the current user
+                if (empty(\preg_grep("#^{$permission}\(.+\)$#", $permissions)) && !empty($user->getId())) {
```
christyjacob4, line 416: What exactly are we doing here ? Doesn't the parse function come in handy to do this ?

christyjacob4 (changes requested):
Some minor comments. 
Awesome work 🚀

src/Appwrite/Specification/Format/Swagger2.php
```diff
@@ -317,7 +317,15 @@ public function parse(): array
                         $node['items'] = [
                             'type' => 'string',
                         ];
-                        $node['x-example'] = '["role:all"]';
+                        $node['x-example'] = '["read(any)"]';
```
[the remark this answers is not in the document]
abnegate: For a permission, as we used to have the named permission separately we only needed a role as the example

app/controllers/api/storage.php
abnegate: We are checking if the current allowed permission is anywhere in the request's permissions, and if not, and there is a current user, we add permission for them

`preg_grep` was used to search an entire array for a string instead of nesting more for loops and adding additional if checks. Neither are simple - if we want to use loops for less complexity I can update :+1:

PR #3681, opened 2022-08-14
Changed provider name from phone to sms and added phone param to create endpoints

eldadfux:
## What does this PR do?

- Changed the phone adapter terminology from `phone` to `sms` to better describe the protocol is the focus not the end device
- Added missing _APP_SMS_FROM to the dev and prod compose files
- Change the param name `number` to `phone` to better describe the value and be consistent with the new changes in the users API
- Decouple Auth code from SMS provider adapter
- Improved code generator logic (was limited to 10 chars, and chars couldn't repeat)

TODO:

Add phone a phone attribute to all user create endpoints in the users service and in the main user creation endpoint in the account service.

## Test Plan

Update existing tests.

## Related PRs and Issues

#2747 

Yes.


app/controllers/api/account.php
```diff
@@ -99,8 +100,10 @@
                 '$write' => ['user:' . $userId],
                 'email' => $email,
                 'emailVerification' => false,
+                'phone' => $phone,
+                'phoneVerification' => false,
                 'status' => true,
-                'password' => Auth::passwordHash($password, Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS),
+                'password' => (!empty($password)) ? Auth::passwordHash($password, Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS) : '',
```
christyjacob4, line 106: In cases where we want to pass an empty password, we do it with a `null` rather than an empty string.

https://github.com/appwrite/appwrite/blob/bb4693ddfc154fe01f617a811fbc71af926b74d2/app/controllers/api/account.php#L1015

app/controllers/api/account.php
```diff
@@ -56,8 +56,9 @@
     ->label('sdk.response.model', Response::MODEL_ACCOUNT)
     ->label('abuse-limit', 10)
     ->param('userId', '', new CustomId(), 'Unique Id. Choose your own unique ID or pass the string "unique()" to auto generate it. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.')
-    ->param('email', '', new Email(), 'User email.')
-    ->param('password', '', new Password(), 'User password. Must be at least 8 chars.')
+    ->param('email', null, new Email(), 'User email.', true)
+    ->param('password', '', new Password(), 'User password. Must be at least 8 chars.', true)
+    ->param('phone', null, new Phone(), 'Phone number. Format this number with a leading \'+\' and a country code, e.g., +16175551212.', true)
```
christyjacob4, line 61: I don't think we want to allow clients to use this endpoint without an email right ? 

We have other endpoints for this.... Magic URL and anonymous session

app/views/install/compose.phtml
```diff
@@ -595,7 +595,7 @@ services:
       - _APP_REDIS_PASS
 
   mariadb:
-    image: mariadb:10.8.3 # fix issues when upgrading using: mysql_upgrade -u root -p
+    image: mariadb:10.7 # fix issues when upgrading using: mysql_upgrade -u root -p
```
christyjacob4, line 598: Is this intended ?

docker-compose.yml
```diff
@@ -638,7 +638,7 @@ services:
       - _APP_REDIS_PASS
 
   mariadb:
-    image: mariadb:10.8.3 # fix issues when upgrading using: mysql_upgrade -u root -p
+    image: mariadb:10.7 # fix issues when upgrading using: mysql_upgrade -u root -p
```
christyjacob4, line 641: Is this expected ?

src/Appwrite/Auth/Auth.php
```diff
@@ -280,6 +280,28 @@ public static function tokenGenerator(int $length = 128): string
...
+    /**
+     * Code Generator.
+     *
+     * Generate random code string
+     *
+     * @param int $length
+     *
+     * @return string
+     *
+     * @throws \Exception
```
christyjacob4, line 292:
```suggestion
```

I don't think this throws an exception

app/controllers/api/account.php
eldadfux: The idea to not couple this endpoint with an email. For that why would email be required, and phone optional? The other alternative is to keep it the same and add `/email` to the path.

app/views/install/compose.phtml
eldadfux: The new version broke my setup, same for Damodar. Not sure what to do with it.

src/Appwrite/Auth/Auth.php
eldadfux: Fixed.

app/controllers/api/account.php
```diff
@@ -2268,17 +2270,16 @@
...
     ->inject('events')
     ->inject('usage')
     ->inject('messaging')
-    ->action(function (Request $request, Response $response, Phone $phone, Document $user, Database $dbForProject, Audit $audits, Event $events, Stats $usage, EventPhone $messaging) {
+    ->action(function (Request $request, Response $response, Document $user, Database $dbForProject, Audit $audits, Event $events, Stats $usage, EventPhone $messaging) {
 
-        if (empty(App::getEnv('_APP_PHONE_PROVIDER'))) {
-            throw new Exception(Exception::GENERAL_PHONE_DISABLED, 'Phone provider not configured');
+        if (empty(App::getEnv('_APP_SMS_PROVIDER'))) {
+            throw new Exception('Phone provider not configured', 503, Exception::GENERAL_PHONE_DISABLED);
```
christyjacob4, line 2282:
```suggestion
            throw new Exception(Exception::GENERAL_PHONE_DISABLED);
```
The exception class has the following signature 
`Exception($type, $message, $code)`

app/views/install/compose.phtml
christyjacob4: Okay 👍

app/controllers/api/account.php
christyjacob4: But the way it is now, it would allow someone to create an account without passing email, password or phone.. 
If this is the case, how would they be able to create sessions?

tests/e2e/Services/Account/AccountBase.php
```diff
@@ -51,41 +51,6 @@ public function testCreateAccount(): array
 
         $this->assertEquals($response['headers']['status-code'], 409);
 
-        $response = $this->client->call(Client::METHOD_POST, '/account', array_merge([
```
christyjacob4, line 54: any reason to remove the tests?

app/controllers/api/account.php
eldadfux: Yeh, I think this is fine for the users service, for the account service you have a point. What about making it more obvious that this endpoint is coupled with email creation by adding `/email` at the end of the path and changing the name + description a bit?

tests/e2e/Services/Account/AccountBase.php
eldadfux: Because of the optional params we added, this tests don't need to fail anymore. I'll bring it back if we change the endpoint.

app/controllers/api/account.php
christyjacob4: Yes thats a good idea. We just refactored the Create Email Session endpoint to reflect this better
https://github.com/appwrite/appwrite/blob/7723c633a021509b56c30f110917d51058b84cdb/app/controllers/api/account.php#L499

tests/e2e/Services/Account/AccountBase.php
christyjacob4: Understood 👍

app/controllers/api/account.php
Meldiron: This code has been removed.

tests/e2e/Services/Account/AccountBase.php
Meldiron: Tests re-added

Meldiron:
I reverted changes regarding AccountAPI createUser. Email and passwords are again required, and the phone is not present. With that, I re-added tests to check the required parameters there.

---

I updated UsersAPI createUser:
- email and password are optional
- added phone
- If none is provided, it creates an anonymous user

---

I also implemented new tests for these cases above.

Meldiron (approved):

app/controllers/api/users.php
```diff
@@ -31,21 +31,27 @@
...
             'email' => $email,
             'emailVerification' => false,
+            'phone' => $phone,
+            'phoneVerification' => false,
             'status' => true,
-            'password' => $hash === 'plaintext' ? Auth::passwordHash($password, $hash, $hashOptionsObject) : $password,
+            'password' => (!empty($password)) ? ($hash === 'plaintext' ? Auth::passwordHash($password, $hash, $hashOptionsObject) : $password) : null,
             'hash' => $hash === 'plaintext' ? Auth::DEFAULT_ALGO : $hash,
             'hashOptions' => $hash === 'plaintext' ? Auth::DEFAULT_ALGO_OPTIONS : $hashOptions,
             'passwordUpdate' => \time(),
```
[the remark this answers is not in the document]
Meldiron: Not sure if I get it.

If no password, then `null`. Just like with the anonymous session.
Also, this runs only for createDocument, so no need to worry about updating from this method.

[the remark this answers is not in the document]
Meldiron: I get it now. Updated to

```php
'passwordUpdate' => (!empty($password)) ? \time() : 0,
```

Only set passwordUpdate if we are actually setting a password. Else, 0 - just like we do with magicURL and anonymous accounts.

PR #3689, opened 2022-08-15
Update List Users API to Accept Queries

stnguyen90:
## What does this PR do?

Showcase of how Queries V2 would be handled by the Appwrite API.

## Test Plan

New automated tests

## Related PRs and Issues

Dependent on:

1. https://github.com/utopia-php/database/pull/167
2. https://github.com/appwrite/appwrite/pull/3700

To be merged after this one:
1. https://github.com/appwrite/appwrite/pull/3702

Yes


app/controllers/api/users.php
```diff
@@ -107,43 +110,60 @@
     ->label('sdk.response.code', Response::STATUS_CODE_OK)
     ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
     ->label('sdk.response.model', Response::MODEL_USER_LIST)
+    ->param('queries', [], new ArrayList(new Text(APP_LIMIT_ARRAY_ELEMENT_SIZE), APP_LIMIT_ARRAY_PARAMS_SIZE), 'Array of query strings generated using the Query class provided by the SDK. [Learn more about queries](https://appwrite.io/docs/databases#querying-documents). Maximum of ' . APP_LIMIT_ARRAY_PARAMS_SIZE . ' queries are allowed, each ' . APP_LIMIT_ARRAY_ELEMENT_SIZE . ' characters long. You may filter on the following attributes: ' . implode(', ', ListUsers::ALLOWED_ATTRIBUTES), true)
```
eldadfux, line 113: We should use the new extended queries array in this stage unless absolutely not possible.

src/Appwrite/Utopia/Database/Validator/Queries/Users/ListUsers.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace Appwrite\Utopia\Database\Validator\Queries\Users;
+
+use Appwrite\Utopia\Database\Validator\Queries\Collection;
+
+class ListUsers extends Collection
```
christyjacob4, line 7: I think this file can be under Validators/Queries and can be called `Users`

app/controllers/api/users.php
```diff
@@ -107,43 +110,60 @@
...
-    ->param('orderType', Database::ORDER_ASC, new WhiteList([Database::ORDER_ASC, Database::ORDER_DESC], true), 'Order result by ASC or DESC order.', true)
     ->inject('response')
     ->inject('dbForProject')
     ->inject('usage')
-    ->action(function (string $search, int $limit, int $offset, string $cursor, string $cursorDirection, string $orderType, Response $response, Database $dbForProject, Stats $usage) {
+    ->action(function (array $queries, string $search, Response $response, Database $dbForProject, Stats $usage) {
+
+        // Parse queries input and also extract cursor, if there is one
+        $cursor = null;
+        foreach ($queries as $i => $query) {
```
eldadfux, line 122: I think we can abstract this logic with a helper method. Maybe `Utopia\Database\Query::getByType($queries, Query::TYPE_CURSORAFTER);`

src/Appwrite/Utopia/Database/Validator/Queries/Users/ListUsers.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace Appwrite\Utopia\Database\Validator\Queries\Users;
```
eldadfux, line 3: No need for the extra namespace for now.
```suggestion
namespace Appwrite\Utopia\Database\Validator\Queries;
```

src/Appwrite/Utopia/Database/Validator/Queries/Users/ListUsers.php
eldadfux, line 7: Class name could be just users unless we need support for something else that should be called `users*`
```suggestion
class Users extends Collection
```

eldadfux: @stnguyen90, for general knowledge. When choosing vars, attributes, or classes names we usually avoid including the namespace in the name as we already get the context from the namespace.

eldadfux (changes requested):

src/Appwrite/Utopia/Database/Validator/Queries/Users/ListUsers.php
stnguyen90: I wanted to use `List`, but it seems to be reserved. 

image

Any suggestions?

stnguyen90: How about later when we need multiple Validators per service?

app/controllers/api/users.php
stnguyen90: Sorry, what do you mean?

app/controllers/api/users.php
stnguyen90: I was trying to avoid iterating over the array again.

eldadfux: Sometime we have to prefer code maintenance over performance 🤷‍♂️

src/Appwrite/Utopia/Database/Validator/Queries/Users/ListUsers.php
eldadfux: We can drop the `Users` directory and call the validator class `Users`, should be perfectly aligned with our conventions.

app/controllers/api/users.php
eldadfux:
```suggestion
    ->param('queries', [], new Users(), 'Array of query strings generated using the Query class provided by the SDK. [Learn more about queries](https://appwrite.io/docs/databases#querying-documents). Maximum of ' . APP_LIMIT_ARRAY_PARAMS_SIZE . ' queries are allowed, each ' . APP_LIMIT_ARRAY_ELEMENT_SIZE . ' characters long. You may filter on the following attributes: ' . implode(', ', ListUsers::ALLOWED_ATTRIBUTES), true)
```

eldadfux: Something like this 👆

app/controllers/api/users.php
stnguyen90: In that case, what about:

```php
$grouped = Queries::groupByType($queries);
$cursor = $grouped['cursorDirection'] === Database::CURSOR_AFTER  ? Query::cursorAfter($grouped['cursor']) : Query::cursorBefore($grouped['cursor']);
```

to avoid creating another function?

src/Appwrite/Utopia/Database/Validator/Queries/Users/ListUsers.php
stnguyen90: How about later when we need multiple Validators per service?

app/controllers/api/users.php
stnguyen90: The queries need to be parsed before the data can be passed to the validator. If I put the pre-processing into the validator, wouldn't I need to parse the queries again in the body?

src/Appwrite/Utopia/Database/Validator/Queries/Users/ListUsers.php
eldadfux: Good point. We can always adapt and change, it's better to not pre-optimize to every situation that might happen or might not. Anyways, most chances that this filtering will be per resource and not per service.

For example in storage we filter by files and buckets. Resources might appear in multiple services so they shouldn't necessarily be coupled under a service name. For example we might enable an endpoint to list the user's teams in the account service, and use the exact same filtering as we use for teams in the teams service. Another example might be our vars list that is now being listed under functions, but will also be listed under the project service for supporting global vars.

stnguyen90: > We can always adapt and change, it's better to not pre-optimize to every situation that might happen or might not.

Wouldn't we tackle the other resources/endpoints right after this endpoint? I'm hoping to clarify this stuff now so that whoever works on the next endpoints as a clear understanding of where/how to build the next Validator for the next endpoint.

stnguyen90: But I think looking at it as response models makes the most sense. So in src/Appwrite/Utopia/Database/Validator/Queries, well have a class for each response model (Files, Users, Collections, Databases, etc.). That sound good?

eldadfux: Answered below.

app/controllers/api/users.php
eldadfux: Good point. Let's think of a view variations and see if we can abstract some of this logic from the action.

Right now my main concerns:
1. This is a lot of repeatable code in the action which we'll need to implement in +10 routes?
2. Action complexity is higher, we always try to max their simplicity (not in every cost), for sake of readability and maintenance, performance will be less priority in many of this cases. Especially when impact is small. 
3. Separation of concerns is not followed, action is handling the input validation and not the param validators (we do have some rare cases where we do this, but mostly this is unusual). We also avoid this as future devs could easily start using not 100% validated variables, which is hard to spot in a review and very risky at times. (AKA "Developer proofing")
4. If we do abstract the logic to the DB library, the structure and names have to make sense and be complimentary to each other, not forced just in sake of a quick solution.

Let me know what you think.

src/Appwrite/Utopia/Database/Validator/Queries/Users/ListUsers.php
eldadfux: Yes 👍

app/controllers/api/users.php
stnguyen90: > 1. This is a lot of repeatable code in the action which we'll need to implement in +10 routes?

I definitely agree with trying to reduce duplicated code. I'm just not sure where to put things. Maybe as static methods in Appwrite's queries Validator?

> 2. Action complexity is higher, we always try to max their simplicity (not in every cost), for sake of readability and maintenance, performance will be less priority in many of this cases. Especially when impact is small.

Thanks for the clarification on the design philosophy 🙏 I'll try to keep this in mind

> 3. Separation of concerns is not followed, action is handling the input validation and not the param validators (we do have some rare cases where we do this, but mostly this is unusual). We also avoid this as future devs could easily start using not 100% validated variables, which is hard to spot in a review and very risky at times. (AKA "Developer proofing")

I can try to put the Validator in the params but I think we'll still need to parse the queries in the action body. Is that okay?

> 4. If we do abstract the logic to the DB library, the structure and names have to make sense and be complimentary to each other, not forced just in sake of a quick solution.

I'm not sure I like the idea of having the Database->find() method accept an array of strings. Do really like being able to pass an array of Query.

eldadfux (approved):

src/Appwrite/Utopia/Database/Validator/Queries/Users.php
```diff
@@ -0,0 +1,28 @@
...
+{
+    public const ALLOWED_ATTRIBUTES = [
+        'name',
+        'email',
+        'phone',
+        'status',
+        'passwordUpdate',
+        'registration',
+        'emailVerification',
+        'phoneVerification',
```
Meldiron, line 17: Let's please double-check we have indexes in collections.php for all these attributes

app/controllers/api/users.php
Meldiron: We now have these validation classes. Resolving discussion.

src/Appwrite/Utopia/Database/Validator/Queries/Users/ListUsers.php
Meldiron: We now have these validation classes. Resolving discussion.

Meldiron: The above has been applied.

src/Appwrite/Utopia/Database/Validator/Queries/Users/ListUsers.php
Meldiron: I believe this was resolved, cannot find such occurrence in the codebase.

src/Appwrite/Utopia/Database/Validator/Queries/Users.php
Meldiron: All good, I double-checked.

Meldiron (approved):

app/controllers/api/databases.php
```diff
@@ -1964,15 +1968,42 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
...
+                        'type' => Database::VAR_DATETIME,
+                        'array' => false,
+                    ]),
+                    new Document([
+                        'key' => '$updatedAt',
+                        'type' => Database::VAR_DATETIME,
+                        'array' => false,
+                    ]),
+                ]
+            );
```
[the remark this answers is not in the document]
Meldiron: We do the same in `Base.php` for other query validators. It is only used for the validator, nowhere else. It allows you to use these 3 internal attributes in queries, like `orderDesc('$createdAt')` or `equal('$id', ['ID1', 'ID2', 'ID3'])`.

Similarly, we already expose them to index creation - you can create compound indexes with these internal attributes in them.

tests/unit/Utopia/Database/Validator/Query/OrderTest.php
```diff
@@ -0,0 +1,54 @@
...
+
+        // Test for Failure
+        $this->assertEquals($this->validator->isValid(Query::limit(-1)), false, $this->validator->getDescription());
+        $this->assertEquals($this->validator->isValid(Query::limit(101)), false, $this->validator->getDescription());
+        $this->assertEquals($this->validator->isValid(Query::offset(-1)), false, $this->validator->getDescription());
+        $this->assertEquals($this->validator->isValid(Query::offset(5001)), false, $this->validator->getDescription());
+        $this->assertEquals($this->validator->isValid(Query::equal('attr', ['v'])), false, $this->validator->getDescription());
+        $this->assertEquals($this->validator->isValid(Query::equal('dne', ['v'])), false, $this->validator->getDescription());
+        $this->assertEquals($this->validator->isValid(Query::equal('', ['v'])), false, $this->validator->getDescription());
+        $this->assertEquals($this->validator->isValid(Query::orderDesc('dne')), false, $this->validator->getDescription());
```
christyjacob4, line 52: Lets add another testcase for 
`$this->assertEquals($this->validator->isValid(Query::orderAsc('dne')), false, $this->validator->getDescription());`

tests/unit/Utopia/Database/Validator/Query/OffsetTest.php
```diff
@@ -0,0 +1,41 @@
...
+class OffsetTest extends TestCase
+{
+    /**
+     * @var Base
+     */
+    protected $validator = null;
+
+    public function setUp(): void
+    {
+        $this->validator = new Offset();
```
christyjacob4, line 19: Lets initialise the `Offset()` validator constructor with the offset explicitly for better readability. 

`new Offset(5000);`

tests/unit/Utopia/Database/Validator/Query/LimitTest.php
```diff
@@ -0,0 +1,37 @@
...
+class LimitTest extends TestCase
+{
+    /**
+     * @var Base
+     */
+    protected $validator = null;
+
+    public function setUp(): void
+    {
+        $this->validator = new Limit();
```
christyjacob4, line 19:
```suggestion
        $this->validator = new Limit();
```
```suggestion
        $this->validator = new Limit(100);
```

src/Appwrite/Utopia/Database/Validator/Query/Base.php
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * Is valid.
+     *
+     * @param Query $value
+     *
+     * @return bool
+     */
+    abstract public function isValid($query): bool;
```
christyjacob4, line 64: dont think we need this. Since the `Validator` class already has an abstract `isValid` method IIRC

src/Appwrite/Utopia/Database/Validator/Query/Base.php
```diff
@@ -0,0 +1,70 @@
...
+     * Get Description.
+     *
+     * Returns validator description
+     *
+     * @return string
+     */
+    public function getDescription(): string
+    {
+        return $this->message;
+    }
```
christyjacob4, line 31: We could remove this as well. Since its already present in the base `Validator` class

src/Appwrite/Utopia/Database/Validator/Query/Base.php
```diff
@@ -0,0 +1,70 @@
...
+    public const METHOD_TYPE_LIMIT = 'limit';
+    public const METHOD_TYPE_OFFSET = 'offset';
+    public const METHOD_TYPE_CURSOR = 'cursor';
+    public const METHOD_TYPE_ORDER = 'order';
+    public const METHOD_TYPE_FILTER = 'filter';
+
+    /**
+     * @var string
+     */
+    protected $message = 'Invalid query';
```
christyjacob4, line 19: We could remove this as well. Since its already present in the base Validator class

christyjacob4 (changes requested):
Please address the comments

app/controllers/api/databases.php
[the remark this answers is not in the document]
Meldiron: I see. Looks like this PR is a bit outdated, in a new version there is `Documents.php` doing exactly this check. Then the validator is used here to validate queries: 
- https://github.com/appwrite/appwrite/blob/2b3dc48bf11dfb68e8dc78ec91fbd78fc7d4049e/src/Appwrite/Utopia/Database/Validator/Queries/Documents.php
- https://github.com/appwrite/appwrite/blob/2b3dc48bf11dfb68e8dc78ec91fbd78fc7d4049e/app/controllers/api/databases.php#L1970-L1975

PR #3691, opened 2022-08-16
Datetime attributes merge

abnegate:
## What does this PR do?

- Pull in 0.16 changes

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -75,11 +75,11 @@ App::post('/v1/account')
             $whitelistIPs = $project->getAttribute('authWhitelistIPs');
 
             if (!empty($whitelistEmails) && !\in_array($email, $whitelistEmails)) {
-                throw new Exception('Console registration is restricted to specific emails. Contact your administrator for more information.', 401, Exception::USER_EMAIL_NOT_WHITELISTED);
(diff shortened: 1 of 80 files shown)


app/controllers/api/account.php
```diff
@@ -1484,8 +1484,8 @@
     ->action(function (string $password, string $oldPassword, Response $response, Document $user, Database $dbForProject, Audit $audits, Stats $usage, Event $events) {
 
         // Check old password only if its an existing user.
-        if ($user->getAttribute('passwordUpdate') !== null && !Auth::passwordVerify($oldPassword, $user->getAttribute('password'))) { // Double check user password
-            throw new Exception('Invalid credentials', 401, Exception::USER_INVALID_CREDENTIALS);
+        if ($user->getAttribute('passwordUpdate') !== 0 && !Auth::passwordVerify($oldPassword, $user->getAttribute('password'))) { // Double check user password
```
christyjacob4, line 1487: Do you know when/why this condition was changed?

christyjacob4 (approved):
Just one comment. Looks good 👍

app/controllers/api/account.php
abnegate: I'm not sure 🤔 it is already in `master` though

Issue #3666, opened 2022-08-12
🐛 Bug Report: Event-triggered function with missing deployment prevents other functions from being executed


redarflap:
### 👟 Reproduction steps

- Create a function and subscribe to an event, but don't create a deployment.
- Subscribe other functions to the same event.
- Trigger the event (verified using event buckets.*.files.*.create)

### 👍 Expected behavior

- The function without deployment should fail.
- All other functions should be executed.

### 👎 Actual Behavior

- The function without deployment fails with Exception:
`Deployment not found. Create deployment before trying to execute a function in /usr/src/code/app/workers/functions.php:208`

- No other function is executed afterwards.

### 🎲 Appwrite version

Version 0.15.x

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

PR #3696, opened 2022-08-17
Fix: event multi-execution

Meldiron:
## What does this PR do?

With this PR, if multiple functions execute with an event, they will all attempt to be executed instead of stopping when one execution fails. Use case explained well in the issue linked below.

I believe this is the correct behavior, as we don't have a function chaining feature in Appwrite.

## Test Plan

- [x] Manual QA

CleanShot 2022-08-17 at 15 36 52

CleanShot 2022-08-17 at 15 36 35

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/3666

Yes 😇

### DIFF
--- a/app/workers/functions.php
+++ b/app/workers/functions.php
@@ -88,7 +88,7 @@ class FunctionsV1 extends Worker
 
                         Console::success('Triggered function: ' . $events[0]);
                     } catch (\Throwable $th) {
-                        Console::error("Failed to execute " . $function->getAttribute('id') . " with error: " . $th->getMessage());
+                        Console::error("Failed to execute " . $function->getId() . " with error: " . $th->getMessage());
                     }
                 }
             }
--- a/app/workers/functions.php
+++ b/app/workers/functions.php
@@ -74,18 +74,22 @@ class FunctionsV1 extends Worker
 
                     Console::success('Iterating function: ' . $function->getAttribute('name'));
 
-                    $this->execute(
-                        project: $project,
-                        function: $function,
-                        dbForProject: $database,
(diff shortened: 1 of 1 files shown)


eldadfux:
Why is this even causing the issue? shouldn't it be handled by our default error handler for workers?

Meldiron:
@eldadfux The problem is that if an event comes into a function worker, and there are multiple functions that need to be executed for this event... Well, if the first execution fails, then no other will run. It could fail, for instance, when a function does not have deployment yet.

Sooo yes, the worker will handle the error and work on the next task in the queue. But in one task, it was supposed to execute 5 functions, but it only executed one and stopped due to an exception.

Does it make sense?

eldadfux:
Do we have 5 tasks in a single task because or our new event model? Have we considered pushing 5 tasks for 5 events instead of running 5 events in one task? What is the max amount of events we can have in the same task?

stnguyen90:
@eldadfux, when an event comes in for the function worker to process, it loops through each function to check execute the event. If one function throws an exception, the for loop fails to finish iterating over all functions.

Meldiron:
@eldadfux  Regarding maximum executions under one event, that is limited as far as I am concerned. For instance, document create: `databases.*.collections.*.documents.*`

There are 3 dynamic parts that can either be specific ID or wildcard. That makes a combination of 8 options if I am not mistaken. For 4 dynamic parts it would be 16, for 5 it would be 32.

christyjacob4 (approved):
Let's change the target branch to 1.3.x

PR #3698, opened 2022-08-18
POC: Extending SDK

lohanidamodar:
## What does this PR do?

## Test Plan

## Related PRs and Issues

[(If this PR is related to any other PR or resolves any issue or related to any issue link all related PR and issues here.)]

- https://github.com/appwrite/sdk-generator/pull/519

(Write your answer here.)


app/controllers/api/storage.php
```diff
@@ -362,6 +362,7 @@
     ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
     ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
     ->label('sdk.response.model', Response::MODEL_FILE)
+    ->label('sdk.example.filters', ['permissions' => 'permissionHelperExample'])
```
eldadfux, line 365: I think it would be much easier for us to support this as a global generic feature on the generator. We can just have sdk.example.params that is passed as `x-example-params` on the spec.

On each language example template we can use this value or the list of required params in case this value is not available.

My main concern about extending Appwrite was about the Queries and Permissions helper classes, less about the example param which we can natively support for any product or platform and not just Appwrite.

eldadfux (changes requested):

PR #3700, opened 2022-08-19
Refactor permissions

abnegate:
## What does this PR do?

- [BREAKING] Introduce new permissions syntax replacing write permission with create, update and delete.
- [BREAKING] Remove $read attribute from Documents
- [BREAKING] Remove $write attribute from Documents
- Add $permissions array attribute to Documents
- WIP Update UI to account for new permissions
- Update all ID usages to use new ID helper

## Test Plan

Existing tests + new tests to cover changed functionality

### Original PR

https://github.com/appwrite/appwrite/pull/3668


app/controllers/api/databases.php
```diff
@@ -1879,42 +1889,72 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
...
 
-        if (!Auth::isAppUser($roles) && !Auth::isPrivilegedUser($roles)) {
-            foreach ($data['$read'] as $read) {
-                if (!Authorization::isRole($read)) {
-                    throw new Exception(Exception::USER_UNAUTHORIZED, 'Read permissions must be one of: (' . \implode(', ', $roles) . ')');
+        /**
+         * Add permissions for current the user for any missing types
+         * from the allowed permissions for this resource type.
+         */
+        $allowedPermissions = \array_filter(
```
eldadfux, line 1907: The logic here should be much simpler. If the permissions array is empty and we have document security enabled, we give the current user read+write permissions. Same should apply for files.

We don't try to provide fallbacks for missing permissions.

app/controllers/api/databases.php
```diff
@@ -1879,42 +1889,72 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
...
-                $document = Authorization::skip(fn() => $dbForProject->createDocument('database_' . $database->getInternalId() . '_collection_' . $collection->getInternalId(), new Document($data)));
-            } else {
-                $document = $dbForProject->createDocument('database_' . $database->getInternalId() . '_collection_' . $collection->getInternalId(), new Document($data));
+        // Users can only manage their own roles, API keys and Admin users can manage any
+        // Users can only manage their own roles, API keys and Admin users can manage any
+        $roles = Authorization::getRoles();
+        if (!Auth::isAppUser($roles) && !Auth::isPrivilegedUser($roles)) {
+            foreach (Database::PERMISSIONS as $type) {
+                foreach ($permissions as $permission) {
+                    $permission = Permission::parse($permission);
```
eldadfux, line 1936: IIRC this validation was abstracted by the Authorization validator so maintains don't need to be trusted to remember implementing it over and over again. Also, I don't see a similar check in other endpoints. Might be very risky.

app/controllers/api/storage.php
```diff
@@ -348,41 +359,69 @@
...
+                 * and there is a current user, add it for the current user.
+                 */
+                if (empty(\preg_grep("#^{$permission}\(.+\)$#", $permissions)) && !empty($user->getId())) {
+                    $permissions[] = (new Permission($permission, 'user', $user->getId()))->toString();
+                }
+            }
+        }
+
+        // Users can only manage their own roles, API keys and Admin users can manage any
+        $roles = Authorization::getRoles();
```
eldadfux, line 409: Same as the comments about for the DB controller.

eldadfux (changes requested):

src/Appwrite/Permissions/PermissionsProcessor.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace Appwrite\Permissions;
+
+use Utopia\Database\Database;
+use Utopia\Database\Permission;
+
+class PermissionsProcessor
```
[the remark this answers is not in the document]
abnegate: Where do you suggest we move the `aggregate` functionality?

app/controllers/api/teams.php
```diff
@@ -787,6 +811,14 @@
             throw new Exception(Exception::TEAM_NOT_FOUND);
         }
 
+        /**
+         * Force document security
+         */
+        $validator = new Authorization('delete');
```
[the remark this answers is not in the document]
abnegate: It's tough because this is the one case where we want document-only permissions, contrary to the new permission model in general. I can't think of a good way to do it other than handling it per case

app/controllers/api/databases.php
abnegate: Happy to change as suggested, just to be clear though, that will differ from the existing code, where we do provide fallbacks: 

https://github.com/appwrite/appwrite/blob/master/app/controllers/api/databases.php#L1874-L1875

app/controllers/api/databases.php
abnegate: From what I can see, the `Authorization` validator isn't handling this case, which somewhat makes sense to me, as being an app/privileged user is Appwrite domain specific

https://github.com/utopia-php/database/blob/main/src/Database/Validator/Authorization.php#L57-L79

app/controllers/api/databases.php
```diff
@@ -2032,13 +2046,12 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
             }
         }
 
-        if ($collection->getAttribute('permission') === 'collection') {
-            /** @var Document[] $documents */
-            $documents = Authorization::skip(fn () => $dbForProject->find('database_' . $database->getInternalId() . '_collection_' . $collection->getInternalId(), $allQueries));
-            $total = Authorization::skip(fn () => $dbForProject->count('database_' . $database->getInternalId() . '_collection_' . $collection->getInternalId(), $filterQueries, APP_LIMIT_COUNT));
-        } else {
+        if ($documentSecurity && !$valid) {
```
stnguyen90, line 2049: @abnegate, is this `$valid` check required here? If so, why and is it also needed in the cursorDocument section above?

cc @Meldiron

abnegate: Yep it is needed; if `$valid`, user has collection level permissions and we should skip the document level check.

It is needed in the cursorDocument cursor and I have missed it, thank you :grin:

lohanidamodar (commented):
Reviewed mostly the usage stats, Changes look good.

app/controllers/api/storage.php
```diff
@@ -66,9 +69,13 @@
     ->inject('response')
     ->inject('dbForProject')
     ->inject('events')
-    ->action(function (string $bucketId, string $name, string $permission, ?array $read, ?array $write, bool $enabled, int $maximumFileSize, array $allowedFileExtensions, bool $encryption, bool $antivirus, Response $response, Database $dbForProject, Event $events) {
+    ->action(function (string $bucketId, string $name, ?array $permissions, string $fileSecurity, bool $enabled, int $maximumFileSize, array $allowedFileExtensions, bool $encryption, bool $antivirus, Response $response, Database $dbForProject, Event $events) {
```
lohanidamodar, line 72: isn't $fileSecurity supposed to be bool?
```suggestion
    ->action(function (string $bucketId, string $name, ?array $permissions, bool $fileSecurity, bool $enabled, int $maximumFileSize, array $allowedFileExtensions, bool $encryption, bool $antivirus, Response $response, Database $dbForProject, Event $events) {
```

app/controllers/api/storage.php
```diff
@@ -233,30 +238,35 @@
     ->inject('response')
     ->inject('dbForProject')
     ->inject('events')
-    ->action(function (string $bucketId, string $name, string $permission, ?array $read, ?array $write, bool $enabled, ?int $maximumFileSize, array $allowedFileExtensions, bool $encryption, bool $antivirus, Response $response, Database $dbForProject, Event $events) {
+    ->action(function (string $bucketId, string $name, ?array $permissions, string $fileSecurity, bool $enabled, ?int $maximumFileSize, array $allowedFileExtensions, bool $encryption, bool $antivirus, Response $response, Database $dbForProject, Event $events) {
```
lohanidamodar, line 241: fileSecurity has boolean validator?
```suggestion
    ->action(function (string $bucketId, string $name, ?array $permissions, bool $fileSecurity, bool $enabled, ?int $maximumFileSize, array $allowedFileExtensions, bool $encryption, bool $antivirus, Response $response, Database $dbForProject, Event $events) {
```

app/controllers/api/storage.php
```diff
@@ -339,53 +348,59 @@
...
         }
 
-        $file = $request->getFiles('file');
-
-        /**
-         * Validators
-         */
-        $allowedFileExtensions = $bucket->getAttribute('allowedFileExtensions', []);
-        $fileExt = new FileExt($allowedFileExtensions);
-
```
lohanidamodar, line 388: Why is this removed?

app/controllers/api/storage.php
```diff
@@ -744,25 +734,25 @@
...
-        // Check bucket permissions when enforced
-        if ($bucket->getAttribute('permission') === 'bucket') {
-            $validator = new Authorization('read');
-            if (!$validator->isValid($bucket->getRead())) {
-                throw new Exception(Exception::USER_UNAUTHORIZED);
-            }
+        $fileSecurity = $bucket->getAttribute('fileSecurity', false);
+        $validator = new Authorization(Database::PERMISSION_READ);
+        $valid = $validator->isValid($bucket->getRead());
+        if (!$valid && !$fileSecurity) {
```
lohanidamodar, line 744: Let's keep the order same everywhere
```suggestion
        if (!$fileSecurity && !$valid) {
```

app/controllers/api/storage.php
```diff
@@ -1098,41 +1090,39 @@
...
         } else {
-            $file = $dbForProject->getDocument('bucket_' . $bucket->getInternalId(), $fileId);
+            $file = Authorization::skip(fn() => $dbForProject->getDocument('bucket_' . $bucket->getInternalId(), $fileId));
         }
 
-        $mimes = Config::getParam('storage-mimes');
-
         if ($file->isEmpty() || $file->getAttribute('bucketId') !== $bucketId) {
             throw new Exception(Exception::STORAGE_FILE_NOT_FOUND);
         }
```
lohanidamodar, line 1116: I don't think we need the second part of this test, as file is always loaded from the bucket specific collection. Not sure why I did not remove this

app/controllers/api/storage.php
```diff
@@ -1239,68 +1229,76 @@
...
+        $roles = Authorization::getRoles();
+        if (!Auth::isAppUser($roles) && !Auth::isPrivilegedUser($roles)) {
+            foreach (Database::PERMISSIONS as $type) {
+                foreach ($permissions as $permission) {
+                    $permission = Permission::parse($permission);
+                    if ($permission->getPermission() != $type) {
+                        continue;
+                    }
+                    $role = (new Role(
+                        $permission->getRole(),
```
lohanidamodar, line 1281: Wouldn't this cause issue if the permission already contains a permission from another user as in line 1268 we are doing
`$permissions = $file->getPermissions() ?? [];` that means $permissions may contain the existing permissions of the file?

lohanidamodar (changes requested):
I think there are some small typos as well as I have few queries.

app/controllers/api/storage.php
abnegate: These are duplicate assignments; you can see they are being set later on here:

https://github.com/appwrite/appwrite/blob/f945ab79ec68fd73da6711baf83ac0a2e204e187/app/controllers/api/storage.php#L409
https://github.com/appwrite/appwrite/blob/refactor-permissions-inc-console-fix/app/controllers/api/storage.php#L447
https://github.com/appwrite/appwrite/blob/refactor-permissions-inc-console-fix/app/controllers/api/storage.php#L448

app/controllers/api/storage.php
abnegate: Yep, the line your referenced should be _after_ this block, will fix

src/Appwrite/Usage/Calculators/Database.php
```diff
@@ -59,13 +61,14 @@ protected function createPerPeriodMetric(string $projectId, string $metric, int
 
         // Required for billing
         if ($monthly) {
-            $time = strtotime("first day of the month");
+            $time = DateTime::createFromFormat('Y-m-d\TH:i:s.v', \date('Y-m-01\T00:00:00.000'))->format(DateTime::RFC3339);
```
eldadfux, line 64: Would have been great if we had a single consistent class to handle dates. We could also address later with @fogelito

src/Appwrite/Utopia/Response/Model/Bucket.php
```diff
@@ -28,25 +28,18 @@ public function __construct()
                 'default' => '',
                 'example' => self::TYPE_DATETIME_EXAMPLE,
             ])
-            ->addRule('$read', [
+            ->addRule('$permissions', [
                 'type' => self::TYPE_STRING,
-                'description' => 'File read permissions.',
+                'description' => 'File permissions.',
```
eldadfux, line 33: If we already improve permissions. We could add here a link to the permissions page to allow devs to learn more about them.

src/Appwrite/Utopia/Response/Model/Bucket.php
```diff
@@ -28,25 +28,18 @@ public function __construct()
...
                 'default' => [],
-                'example' => ['role:all'],
+                'example' => ['read("any")'],
                 'array' => true,
             ])
-            ->addRule('$write', [
+            ->addRule('fileSecurity', [
```
eldadfux, line 38: A link to learn more about permissions in Appwrite would be useful.

app/controllers/api/databases.php
eldadfux: This is OK, as its similar to the previous implementation. We could consider to have this logic available on the database library to reduce complexity in this layer. We can handle it individually later.

eldadfux (approved):
Looks great!

app/controllers/api/databases.php
```diff
@@ -1250,6 +1258,8 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
     ->label('event', 'databases.[databaseId].collections.[collectionId].attributes.[attributeId].create')
     ->label('scope', 'collections.write')
     ->label('audits.resource', 'database/{request.databaseId}/collection/{request.collectionId}')
+    ->label('usage.metric', 'collections.{scope}.requests.update')
+    ->label('usage.params', ['databaseId:{request.databaseId}'])
```
lohanidamodar, line 1262: These changes are confusing, on 0.16.x this should already be there, is there issue on the `datetime` branch during sync with 0.16.x?

lohanidamodar (commented):
Just a query. Database controller looks good. Can we change the PR to 0.16.x or `origin/datetime-attributes` have some other changes

app/controllers/api/databases.php
abnegate: Correct, the POST `/attributes/boolean` route was newly added in the datetime branch, before the usage updates were merged into 0.16

lohanidamodar (approved):
Looks good. Focused my review on databases, storage and usage.

tests/unit/Auth/AuthTest.php
```diff
@@ -403,7 +405,7 @@ public function testPrivilegedUserRoles(): void
         $roles = Auth::getRoles($user);
 
         $this->assertCount(5, $roles);
-        $this->assertNotContains('role:member', $roles);
+        $this->assertNotContains('users', $roles);
```
Meldiron, line 408: Let's use Role class.

tests/unit/Auth/AuthTest.php
```diff
@@ -437,7 +439,7 @@ public function testAppUserRoles(): void
...
```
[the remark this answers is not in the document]
Meldiron: Also below in other asserts

tests/unit/Auth/AuthTest.php
```diff
@@ -369,7 +371,7 @@ public function testUserRoles(): void
         $roles = Auth::getRoles($user);
 
         $this->assertCount(7, $roles);
-        $this->assertContains('role:member', $roles);
+        $this->assertContains('users', $roles);
```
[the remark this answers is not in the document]
Meldiron: Also below in other asserts

tests/unit/Auth/AuthTest.php
Meldiron: Also below in other asserts

app/controllers/api/storage.php
```diff
@@ -102,25 +109,24 @@
...
-            $bucket = $dbForProject->createDocument('buckets', new Document([
+            $dbForProject->createDocument('buckets', new Document([
                 '$id' => $bucketId,
                 '$collection' => 'buckets',
+                '$permissions' => $permissions,
                 'name' => $name,
-                'permission' => $permission,
                 'maximumFileSize' => $maximumFileSize,
                 'allowedFileExtensions' => $allowedFileExtensions,
+                'fileSecurity' => (bool) filter_var($fileSecurity, FILTER_VALIDATE_BOOLEAN),
```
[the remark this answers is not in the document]
Meldiron: Applies to multiple changes in this file.

tests/e2e/Services/Databases/DatabasesBase.php
```diff
@@ -2318,96 +2392,308 @@ public function testEnforceCollectionPermissions()
...
+            'collectionId' => ID::unique(),
             'name' => 'enforceCollectionPermissions',
-            'permission' => 'collection',
-            'read' => [$user],
-            'write' => []
+            'permissions' => [
+                Permission::read(Role::user($user)),
+                Permission::create(Role::user($user)),
+                Permission::update(Role::user($user)),
+                Permission::delete(Role::user($user)),
```
Meldiron, line 2522: Write was empty previously.

Meldiron (changes requested):

app/controllers/api/databases.php
```diff
@@ -476,8 +476,8 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
...
     ->param('name', '', new Text(128), 'Collection name. Max length: 128 chars.')
-    ->param('permissions', null, new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE), 'An array of strings with permissions. By default no user is granted with any read permissions. [learn more about permissions](https://appwrite.io/docs/permissions) and get a full list of available permissions.')
-    ->param('documentSecurity', false, new Boolean(true), 'Specifies the permissions model used in this collection, which accepts either \'collection\' or \'document\'. For \'collection\' level permission, the permissions specified in read and write params are applied to all documents in the collection. For \'document\' level permissions, read and write permissions are specified in each document. [learn more about permissions](https://appwrite.io/docs/permissions) and get a full list of available permissions.')
+    ->param('permissions', null, new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE), 'An array of permissions strings. By default no user is granted with any permissions. [Learn more about permissions](/docs/permissions) and get a full list of available permissions.')
```
gewenyu99, line 479: "and get a full list of available permissions" is redundant. I think just linking Learn more about permissions is enough.

Space is valuable here.

app/controllers/api/databases.php
```diff
@@ -476,8 +476,8 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
...
+    ->param('documentSecurity', false, new Boolean(true), 'Whether to enable document-level permissions, where each document\'s permissions will be merged with the collection\'s permissions to determine who has access to each document individually. [Learn more about permissions](/docs/permissions).')
```
gewenyu99, line 480: Enables configuring permissions for individual documents. A user needs one of document or collection level permissions to access a document. [Learn more about permissions](/docs/permissions).

app/controllers/api/databases.php
```diff
@@ -728,8 +728,8 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
     ->param('databaseId', '', new UID(), 'Database ID.')
     ->param('collectionId', '', new UID(), 'Collection ID.')
     ->param('name', null, new Text(128), 'Collection name. Max length: 128 chars.')
-    ->param('permissions', null, new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE), 'An array of strings with permissions. By default no user is granted with any permissions. [learn more about permissions](/docs/permissions) and get a full list of available permissions.', true)
-    ->param('documentSecurity', false, new Boolean(true), 'Whether to enable document-level permission where each document\'s permissions parameter will decide who has access to each file individually. [learn more about permissions](/docs/permissions) and get a full list of available permissions.')
+    ->param('permissions', null, new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE), 'An array of permission strings. By default the current permission are inherited. [Learn more about permissions](/docs/permissions) and get a full list of available permissions.', true)
```
gewenyu99, line 731: I think "and get a full list of available permissions" is redundant here.

app/controllers/api/databases.php
```diff
@@ -728,8 +728,8 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
...
+    ->param('documentSecurity', false, new Boolean(true), 'Whether to enable document-level permissions, where each document\'s permissions will be merged with the collection\'s permissions to determine who has access to each document individually. [Learn more about permissions](/docs/permissions).')
```
gewenyu99, line 732: Consider the same phrasing as above.

app/controllers/api/databases.php
```diff
@@ -1813,7 +1813,7 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
...
     ->param('collectionId', null, new UID(), 'Collection ID. You can create a new collection using the Database service [server integration](https://appwrite.io/docs/server/databases#databasesCreateCollection). Make sure to define attributes before creating documents.')
     ->param('data', [], new JSON(), 'Document data as JSON object.')
-    ->param('permissions', null, new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE, [Database::PERMISSION_READ, Database::PERMISSION_UPDATE, Database::PERMISSION_DELETE]), 'An array of strings with permissions. By default no user is granted with any permissions. [learn more about permissions](/docs/permissions) and get a full list of available permissions.', true)
+    ->param('permissions', null, new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE, [Database::PERMISSION_READ, Database::PERMISSION_UPDATE, Database::PERMISSION_DELETE]), 'An array of permissions strings. By default the current user is granted with all permissions. [Learn more about permissions](/docs/permissions) and get a full list of available permissions.', true)
```
gewenyu99, line 1816: I think "and get a full list of available permissions" is redundant here. "Learn more about permissions" implies getting a full list.

app/controllers/api/databases.php
```diff
@@ -2202,7 +2202,7 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
     ->param('collectionId', null, new UID(), 'Collection ID.')
     ->param('documentId', null, new UID(), 'Document ID.')
     ->param('data', [], new JSON(), 'Document data as JSON object. Include only attribute and value pairs to be updated.', true)
-    ->param('permissions', null, new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE), 'An array of strings with permissions. By default no user is granted with any permissions. [learn more about permissions](/docs/permissions) and get a full list of available permissions.', true)
+    ->param('permissions', null, new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE), 'An array of permissions strings. By default the current permissions are inherited. [Learn more about permissions](/docs/permissions) and get a full list of available permissions.', true)
```
gewenyu99, line 2205: I think "and get a full list of available permissions" is redundant here. "Learn more about permissions" implies getting a full list.

app/controllers/api/storage.php
```diff
@@ -59,8 +59,8 @@
...
     ->param('bucketId', '', new CustomId(), 'Unique Id. Choose your own unique ID or pass the string `unique()` to auto generate it. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.')
     ->param('name', '', new Text(128), 'Bucket name')
-    ->param('permissions', null, new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE), 'An array of strings with permissions. By default no user is granted with any permissions. [learn more about permissions](/docs/permissions) and get a full list of available permissions.', true)
-    ->param('fileSecurity', false, new Boolean(true), 'Whether to enable file-level permission where each files permissions parameter will decide who has access to each file individually. [learn more about permissions](/docs/permissions) and get a full list of available permissions.')
+    ->param('permissions', null, new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE), 'An array of permission strings. By default no user is granted with any permissions. [Learn more about permissions](/docs/permissions) and get a full list of available permissions.', true)
```
gewenyu99, line 62: I think "and get a full list of available permissions" is redundant here. "Learn more about permissions" implies getting a full list.

app/controllers/api/storage.php
```diff
@@ -59,8 +59,8 @@
...
+    ->param('fileSecurity', false, new Boolean(true), 'Whether to enable file-level permissions, where each file\'s permissions will be merged with the bucket\'s permissions to determine who has access to each file individually. [Learn more about permissions](/docs/permissions).')
```
gewenyu99, line 63: Enables permission configuration for individual files. A user needs one of file or bucket level permissions to access a file. [Learn more about permissions](/docs/permissions).

Maybe use `Toggles` instead of enables.

app/controllers/api/storage.php
```diff
@@ -228,8 +228,8 @@
     ->label('sdk.response.model', Response::MODEL_BUCKET)
     ->param('bucketId', '', new UID(), 'Bucket unique ID.')
     ->param('name', null, new Text(128), 'Bucket name', false)
-    ->param('permissions', null, new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE), 'An array of strings with permissions. By default no user is granted with any permissions. [learn more about permissions](/docs/permissions) and get a full list of available permissions.', true)
-    ->param('fileSecurity', false, new Boolean(true), 'Whether to enable file-level permission where each files permissions parameter will decide who has access to each file individually. [learn more about permissions](/docs/permissions) and get a full list of available permissions.')
+    ->param('permissions', null, new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE), 'An array of permission strings. By default the current permissions are inherited. [Learn more about permissions](/docs/permissions) and get a full list of available permissions.', true)
```
gewenyu99, line 231: I think "and get a full list of available permissions" is redundant here. "Learn more about permissions" implies getting a full list.

app/controllers/api/storage.php
```diff
@@ -228,8 +228,8 @@
...
+    ->param('fileSecurity', false, new Boolean(true), 'Whether to enable file-level permissions, where each file\'s permissions will be merged with the bucket\'s permissions to determine who has access to each file individually. [Learn more about permissions](/docs/permissions).')
```
gewenyu99, line 232: Ditto above comments

gewenyu99 (changes requested):
I tried to make these descriptions more concise. See if you like the suggestions and accept the comments you agree with :)

- I think "and get a full list of available permissions" is redundant after "Learn more about permissions".
- I suggested a reword for documentSecurity and fileSecurity.

christyjacob4 (changes requested):
some tiny comments.. Looks great 👍

composer.json
```diff
@@ -74,22 +74,11 @@
...
-            "type": "git"
-        },
-        {
-            "url": "https://github.com/utopia-php/audit.git",
-            "type": "git"
         }
     ],
     "require-dev": {
-        "appwrite/sdk-generator": "0.20.0",
+        "ext-fileinfo": "*",
```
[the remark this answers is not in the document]
abnegate: Yes, we use fileinfo in tests so we should declare it as required for development.

composer.json
```diff
@@ -74,22 +74,11 @@
...
-            "url": "https://github.com/utopia-php/database.git",
-            "type": "git"
-        },
-        {
-            "url": "https://github.com/utopia-php/abuse.git",
-            "type": "git"
-        },
-        {
-            "url": "https://github.com/utopia-php/audit.git",
-            "type": "git"
```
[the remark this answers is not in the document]
abnegate: They are redundant, I was asked to do the same removal for database library so applied the change here too

docs/specs/authentication.drawio.svg
```diff
@@ -342,7 +342,7 @@
                             <div style="display: inline-block; font-size: 12px; font-family: Helvetica; color: rgb(0, 0, 0); line-height: 1.2; pointer-events: all; white-space: normal; overflow-wrap: normal;">
                                 Guset
                                 <br/>
-                                role:guest
+                                guests
```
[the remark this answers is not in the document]
abnegate: I believe this file is generated with specs

tests/e2e/Services/Databases/DatabasesPermissionsGuestTest.php
```diff
@@ -49,45 +57,51 @@ public function createCollection(): array
...
     public function readDocumentsProvider()
     {
         return [
-            [['role:all'], []],
-            [['role:member'], []],
-            [[] ,['role:all']],
-            [['role:all'], ['role:all']],
-            [['role:member'], ['role:member']],
-            [['role:all'], ['role:member']],
+            [[Permission::read(Role::any())]],
```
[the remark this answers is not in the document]
abnegate: That is how the `@dataProvider` annotation works; it passes the outer array items as parameters to the test function, so in the test, we just get the inner array

tests/e2e/Services/Databases/DatabasesPermissionsMemberTest.php
```diff
@@ -23,53 +26,110 @@ public function createUsers(): array
...
+                'users' => 7,
+                'doconly' => 2,
+            ],
+            [
+                'permissions' => [Permission::read(Role::any()), Permission::update(Role::any()), Permission::delete(Role::any())],
+                'any' => 8,
+                'users' => 8,
+                'doconly' => 3,
+            ],
+            [
```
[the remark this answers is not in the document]
abnegate: Have added another test for read users, write users

tests/unit/Messaging/MessagingTest.php
```diff
@@ -196,103 +207,125 @@ public function testConvertChannelsUser(): void
...
+                    Permission::read(Role::team('123abc')),
+                    Permission::update(Role::team('123abc')),
+                    Permission::delete(Role::team('123abc')),
+                ],
             ]),
             database: new Document([
-                '$id' => 'database',
+                '$id' => ID::custom('database'),
+            ]),
+            collection: new Document([
```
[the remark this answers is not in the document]
abnegate: The default is false so no need

tests/e2e/Services/Realtime/RealtimeCustomClientTest.php
```diff
@@ -642,11 +645,15 @@ public function testChannelDatabase()
...
+            'collectionId' => ID::unique(),
             'name' => 'Actors',
-            'read' => [],
-            'write' => [],
-            'permission' => 'document'
+            'permissions' => [
+                Permission::read(Role::users()),
+                Permission::create(Role::users()),
+                Permission::update(Role::users()),
+                Permission::delete(Role::users()),
```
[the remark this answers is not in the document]
abnegate: Yep we need at least create permission for the current user; otherwise, only admins/API key could create documents in the collection. I've updated it to only include create permission for the current user

app/controllers/api/storage.php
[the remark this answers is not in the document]
abnegate: Was just following the existing pattern for booleans here but I've verified it's unnecessary and used them directly

app/controllers/api/databases.php
abnegate: Yep I agree, updated all instances

app/controllers/api/account.php
```diff
@@ -119,9 +125,9 @@
             throw new Exception(Exception::USER_ALREADY_EXISTS);
         }
 
-        Authorization::unsetRole('role:' . Auth::USER_ROLE_GUEST);
-        Authorization::setRole('user:' . $user->getId());
-        Authorization::setRole('role:' . Auth::USER_ROLE_MEMBER);
+        Authorization::unsetRole(Role::guests()->toString());
+        Authorization::setRole(Role::user($user->getId())->toString());
+        Authorization::setRole(Role::users()->toString());
```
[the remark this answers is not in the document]
abnegate: `setRole` appends to a `$roles` array; this is ensuring that the current user has both `user:id` and `users` roles

tests/e2e/Services/Databases/DatabasesBase.php
```diff
@@ -1630,11 +1656,13 @@ public function testInvalidDocumentStructure()
...
-            'collectionId' => 'unique()',
+            'collectionId' => ID::unique(),
             'name' => 'invalidDocumentStructure',
-            'read' => [],
-            'write' => [],
-            'permission' => 'document',
+            'permissions' => [
+                Permission::create(Role::any()),
+                Permission::read(Role::any()),
+            ],
```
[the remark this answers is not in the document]
abnegate: Omitted individual permissions are not granted to anyone. If the `permissions` parameter is not passed at all, all permissions are granted to the current user. Empty array means no one has permission for anything (excluding admin/API keys)

src/Appwrite/Utopia/Response/Model/Execution.php
```diff
@@ -28,11 +29,11 @@ public function __construct()
                 'default' => '',
                 'example' => self::TYPE_DATETIME_EXAMPLE,
             ])
-            ->addRule('$read', [
+            ->addRule('$permissions', [
```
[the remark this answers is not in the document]
abnegate: It would make more sense to call it `$roles` since it is only role strings and not related to any specific permission. What do you think @eldadfux?

app/controllers/api/account.php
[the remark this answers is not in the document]
christyjacob4: okay 👍

tests/e2e/Services/Databases/DatabasesBase.php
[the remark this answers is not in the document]
christyjacob4: > Empty array means no one has permission for anything (excluding admin/API keys)

I assumed so too. 

If you look at the diff here. Previously the permissions were empty which means no one has access to the document. Now its been replaced with `::any` which means anyone can access it. Which is the complete opposite of what it was previously. Is this expected ? 

Screenshot 2022-08-27 at 10 00 36 AM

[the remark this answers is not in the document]
abnegate: We used to skip checking collection permissions if document-level were enabled, so having empty permissions worked in later tests like `createDocument`. In the new model, the collection permissions are always checked, and the document permissions are additionally checked if enabled. 

So for the test referenced, we needed to add _at least_ create permission for the current user, or no one could create any documents. I will update as such.

Issue #3627, opened 2022-07-31
🚀 Feature: Databases - Return attribute or index name on 409 error (document_already_exists)


pilcrowOnPaper:
### 🔖 Feature description

When a user tries to create a document that violates a unique type index, return the name of the index that it violates.

Current: 
```json
{
   "code": "document_already_exists",
   "message": "Document already exists"
}
```

New:
```json
{
   "code": "document_already_exists",
   "message": "Document already exists: document violates type unique for index 'index_name'"
}
```

### 🎤 Pitch

If you have a website where users can store items with multiple unique attributes, it would be nice to tell the user which input should be fixed.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

stnguyen90:
Thank you for raising this issue! 🙏 I'll be sure to pass this feedback to the team!

eldadfux:
This will force us to make additional analyses for every 409 error which is not great for performance and will also introduce more complexity and larger maintenance surface for something that should be pretty straight forward for a developer familiar with their data structure. We do plan to introduce in the future more tools around error observability and analysis, we could keep this idea in mind as part of that.

PR #3711, opened 2022-08-22
Improve duplicate exception messages

Meldiron:
## What does this PR do?

In the current implementation, there are 2 problems:

- It always says ID is duplicated for all duplication errors
- It doesn't tell you any info if your custom unique index fails

I addressed this by giving more info on the duplicate error for databases service.

Before:

```js
{"message":"Document with the requested ID already exists.","code":409,"type":"document_already_exists"}
```

After:

```js
{"message":"Duplicated document: Duplicate entry 'Matej' for key 'name'","code":409,"type":"document_already_exists"}
```

A problem that was created is a message about duplicate ID:

```
Duplicated document: Duplicate entry '123' for key '_index1'
```

This may be solved on library-level by giving indexes a proper name depending on what they hold: https://github.com/utopia-php/database/pull/171

New message would be:

```
Duplicated document: Duplicate entry '123' for key '_uid'
```

## Test Plan

Current tests must pass.

## Related PRs and Issues

Issues:
- https://github.com/appwrite/appwrite/issues/3627

PRs:
- https://github.com/utopia-php/database/pull/171

Yes 💪

### DIFF
--- a/app/controllers/api/databases.php
+++ b/app/controllers/api/databases.php
@@ -1858,7 +1858,7 @@ App::post('/v1/databases/:databaseId/collections/:collectionId/documents')
         } catch (StructureException $exception) {
             throw new Exception(Exception::DOCUMENT_INVALID_STRUCTURE, $exception->getMessage());
         } catch (DuplicateException $exception) {
-            throw new Exception(Exception::DOCUMENT_ALREADY_EXISTS);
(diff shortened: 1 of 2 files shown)


abnegate:
Accounts are affected by this too - but in the reverse way. Creating an account with a duplicate _ID_ gives the error:
`A user with the same email already exists in your project.`

PR #3720, opened 2022-08-26
Allow flaky tests to be automatically retried with an attribute

abnegate:
## What does this PR do?

Add `Retry` attribute and `Retryable` trait to allow marking tests as flaky and automatically retrying them the given amount of times. 

`Scope` now has the `Retryable` trait so all you need to do is add the attribute to the flaky test.

```php
#[Retry(count: 1)]
public function testThatSometimesFails()
````

Should save time by not requiring re-running the entire test suite for the few tests that randomly fail.

## Test Plan

Existing tests

Yes

### CODE REVIEW

File: tests/e2e/Services/Webhooks/WebhooksBase.php [hunk 1]
```diff
@@ -296,6 +297,7 @@ public function testUpdateDocument(array $data): array
     /**
      * @depends testCreateCollection
      */
+    #[FlakyTest(retries: 1)]
```
stnguyen90, line 300: I love the idea of adding this! Is this type of syntax common? Or are docblock tags more common? When looking into this, did you take a look at [bshaffer/phpunit-retry-annotations](https://github.com/bshaffer/phpunit-retry-annotations)?

abnegate: I think annotations are likely far more common as attributes were only added in PHP 8. However, attributes are objectively better, check out why in [the RFC here](https://wiki.php.net/rfc/attributes_v2#why_not_extending_doc_comments).

 I did not, I found a retry trait example on some Drupal forum and modified it to fit. Basically the same thing though

tests/extensions/FlakyTest.php
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Appwrite\Tests;
+
+/**
+ * Allows test methods to be retried if they fail.
+ *
+ * Requires that the test class extends {@see TestCase} and has trait {@see Retryable}.
+ */
+#[\Attribute(\Attribute::TARGET_METHOD)]
+class FlakyTest
```
christyjacob4, line 11: I think its better to call this `Retry` with the constructor parameter as `$count` . That way we would call it as

`#[Retry(count: 1)]`

christyjacob4 (changes requested):
Please address the comment

PR #3727, opened 2022-08-29
Allow guest write scope for document, file and execution

abnegate:
## What does this PR do?

Allow guests write scope for document, files, executions. Missed from spec for permissions V2.

## Test Plan

Expanded tests to cover guest create/update/delete

## Related PRs and Issues

https://github.com/appwrite/appwrite/pull/3700

Yes


app/controllers/api/databases.php
```diff
@@ -1802,6 +1802,8 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
     ->label('audits.resource', 'database/{request.databaseId}/collection/{request.collectionId}')
     ->label('usage.metric', 'documents.{scope}.requests.create')
     ->label('usage.params', ['databaseId:{request.databaseId}', 'collectionId:{request.collectionId}'])
+    ->label('abuse-limit', 120)
```
lohanidamodar, line 1805: Why is this different than other ones?

lohanidamodar (approved):

app/controllers/api/databases.php
abnegate: @Meldiron Raised a good point that some workflows require creating multiple documents at once and we don't want to be too limiting to client apps, especially when the change is already reducing the amount of requests they can make.

stnguyen90: @abnegate, @Meldiron, since the [abuse key is the URL](https://github.com/appwrite/appwrite/blob/a9cfb06ee92f701877642b7880d513bbe28332fb/app/controllers/shared/api.php#L48), this would allow creating 120 documents in 1 collection in 1 minute, right? Is that reasonable?

stnguyen90: Also, after reviewing the abuse check a bit more, it looks like [params are part of the abuse check](https://github.com/appwrite/appwrite/blob/a9cfb06ee92f701877642b7880d513bbe28332fb/app/controllers/shared/api.php#L72) so wouldn't a bad actor be able to change the param value to bypass the abuse check?

abnegate: @stnguyen90 What is reasonable is open for discussion - what are your thoughts?

We can configure the key to be whatever we want: URL, IP, any of the params, or a combination of all.

The block you referenced lets us explicitly set request params as a part of the key - if we don't use them, it won't have any effect. For example, the abuse key for creating a magic URL contains the `userId`, but none of the other params:

```php
    ->label('abuse-key', 'url:{url},userId:{param-userId}')
```

christyjacob4 (approved):

app/controllers/api/databases.php
stnguyen90: @abnegate, 

> What is reasonable is open for discussion - what are your thoughts?

1 doc in a particular collection per second should be plenty, right?

> The block you referenced lets us explicitly set request params as a part of the key

I see now! Thanks for the clarification!

abnegate: @stnguyen90 Please check the refactor here and lets follow up there: https://github.com/appwrite/appwrite/pull/3741

> 1 doc in a particular collection per second should be plenty, right?

From further investigation today, URL keys are not populated with ID's, however, the route method was not taken into account which led to limits applying across PATCH/PUT/DELETE routes with the same URL. Following the refactor, keys will look like this:

Screen Shot 2022-08-31 at 4 18 08 PM

cc @eldadfux

PR #3736, opened 2022-08-30
Refactor cache date

shimonewman:
## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)


app/controllers/shared/api.php
```diff
@@ -358,30 +358,33 @@
...
                     'resource' => $resource,
-                    'accessedAt' => \time(),
+                    'accessedAt' => $now,
                     'signature' => $signature,
                     ])));
-                } elseif (date('Y/m/d', \time()) > date('Y/m/d', $cacheLog->getAttribute('accessedAt'))) {
-                    $cacheLog->setAttribute('accessedAt', \time());
+                } elseif (date('Y/m/d', \time()) > date('Y/m/d', strtotime($cacheLog->getAttribute('accessedAt')))) {
+                    var_dump('update');
+                    var_dump($now);
```
christyjacob4, line 386: we should remove these

app/controllers/shared/api.php
```diff
@@ -358,30 +358,33 @@
...
                 $key = md5($request->getURI() . implode('*', $request->getParams()));
-
                 $data = json_encode([
                 'content-type' => $response->getContentType(),
                 'payload' => base64_encode($data['payload']),
                 ]) ;
 
                 $signature = md5($data);
                 $cacheLog  = $dbForProject->getDocument('cache', $key);
+                $now = date('d-m-Y H:i:s', time());
```
christyjacob4, line 376: With the new `datetime` changes we should use the Datetime class to get the current time. 

Reference 
https://github.com/appwrite/appwrite/blob/0.16.x/app/controllers/general.php#L310

app/controllers/shared/api.php
```diff
@@ -358,30 +358,33 @@
...
                     Authorization::skip(fn () => $dbForProject->createDocument('cache', new Document([
                     '$id' => $key,
                     'resource' => $resource,
-                    'accessedAt' => \time(),
+                    'accessedAt' => $now,
                     'signature' => $signature,
                     ])));
-                } elseif (date('Y/m/d', \time()) > date('Y/m/d', $cacheLog->getAttribute('accessedAt'))) {
-                    $cacheLog->setAttribute('accessedAt', \time());
+                } elseif (date('Y/m/d', \time()) > date('Y/m/d', strtotime($cacheLog->getAttribute('accessedAt')))) {
```
christyjacob4, line 384: For comparisons as well. Let's use the new datetime class

app/controllers/shared/api.php
shimonewman: fixed

app/controllers/shared/api.php
shimonewman: fixed

app/controllers/shared/api.php
shimonewman: fixed, using the DateTime::now() still needs to be formatted

app/controllers/shared/api.php
```diff
@@ -362,30 +363,33 @@
...
                 $cacheLog  = $dbForProject->getDocument('cache', $key);
+                $accessedAt = $cacheLog->getAttribute('accessedAt', '');
+                $accessedAt = (new \DateTime($accessedAt))->format('Y/m/d');
+
                 if ($cacheLog->isEmpty()) {
                     Authorization::skip(fn () => $dbForProject->createDocument('cache', new Document([
                     '$id' => $key,
                     'resource' => $resource,
-                    'accessedAt' => \time(),
+                    'accessedAt' => DateTime::now(),
```
christyjacob4, line 388:
```suggestion
                    'accessedAt' => \DateTime(),
```


`DateTime::now()` returns a formatted string. We need to use the DateTime object directly as it then goes through the database filters

app/controllers/shared/api.php
```diff
@@ -362,30 +363,33 @@
...
-
                 $data = json_encode([
                 'content-type' => $response->getContentType(),
                 'payload' => base64_encode($data['payload']),
                 ]) ;
 
                 $signature = md5($data);
                 $cacheLog  = $dbForProject->getDocument('cache', $key);
+                $accessedAt = $cacheLog->getAttribute('accessedAt', '');
+                $accessedAt = (new \DateTime($accessedAt))->format('Y/m/d');
```
christyjacob4, line 382: Why aren't we using this variable in Line 388 ?

app/controllers/shared/api.php
```diff
@@ -362,30 +363,33 @@
...
                     Authorization::skip(fn () => $dbForProject->createDocument('cache', new Document([
                     '$id' => $key,
                     'resource' => $resource,
-                    'accessedAt' => \time(),
+                    'accessedAt' => DateTime::now(),
                     'signature' => $signature,
                     ])));
-                } elseif (date('Y/m/d', \time()) > date('Y/m/d', $cacheLog->getAttribute('accessedAt'))) {
-                    $cacheLog->setAttribute('accessedAt', \time());
+                } elseif ((new \DateTime(DateTime::now()))->format('Y/m/d') > $accessedAt) {
```
christyjacob4, line 391: Let's be more consistent and use something like this 
https://github.com/appwrite/appwrite/blob/0.16.x/app/controllers/general.php#L310

app/workers/deletes.php
```diff
@@ -225,7 +225,7 @@ protected function deleteUsageStats(string $datetime1d, string $datetime30m)
             ], $dbForProject);
 
             $this->deleteByGroup('stats', [
-                Query::lessThan('time', [$datetime30m]),
+                Query::lessThan('time', $datetime30m),
```
lohanidamodar, line 228: This doesn't make sense? being array or not it should be same right? we have the same below `Query::equal('period', ['30m']`?

However, the change should not make a difference as far as I know how database query works

shimonewman: The DateTime validator was throwing errors.
The diff is period is a string  while time is dateTime attr

PR #3737, opened 2022-08-30
Expose and introduce File Compression support

lohanidamodar:
## What does this PR do?

- [x] Add compression attribute on buckets collection
- [x] Add compression attribute on create and update bucket endpoint
- [x] Account file create, preview, view, download to account for new compression algorithm support
- [x] Update dockerfile to include Zstd compressor
- [x] Compression in response
- [x] Test for zstd compression
- [x] UI update for compression

### UI
image

## Test Plan

Yet to add test for compression

## Related PRs and Issues

- Added new test for Zstd compression

YES


abnegate (changes requested):
With this, it will be possible to get out of sync, where the file's `algorithm` is not the same as the bucket's `compression`, which might be unexpected for developers. For example:

1. Create bucket with gzip `compression`
2. Create file, will use gzip and set `algorithm` to gzip
3. Update bucket `compression` to zstd
4. Get file, will use gzip from `algorithm`, would expect zstd

But I understand it would not make sense to decompress and recompress every file in the bucket when `compression` changes. Can't see a clean way around it, as long we make it clear in the docs or something I think it should be fine.

lohanidamodar:
> With this, it will be possible to get out of sync, where the file's `algorithm` is not the same as the bucket's `compression`, which might be unexpected for developers. For example:
> 
> 1. Create bucket with gzip `compression`
> 2. Create file, will use gzip and set `algorithm` to gzip
> 3. Update bucket `compression` to zstd
> 4. Get file, will use gzip from `algorithm`, would expect zstd
> 
> But I understand it would not make sense to decompress and recompress every file in the bucket when `compression` changes. Can't see a clean way around it, as long we make it clear in the docs or something I think it should be fine.

This is known, but think it would be okay

Dockerfile
```diff
@@ -34,7 +34,8 @@ ENV PHP_REDIS_VERSION=5.3.7 \
     PHP_SWOOLE_VERSION=v4.8.10 \
     PHP_IMAGICK_VERSION=3.7.0 \
     PHP_YAML_VERSION=2.2.2 \
-    PHP_MAXMINDDB_VERSION=v1.11.0
+    PHP_MAXMINDDB_VERSION=v1.11.0 \
+    PHP_ZSTD_VERSION="master"
```
eldadfux, line 38: better us a tag or a commit, master can break

app/config/collections.php
```diff
@@ -2875,6 +2875,16 @@
                 'required' => true,
                 'array' => true,
             ],
+            [
```
eldadfux, line 2878: Is it a new field? What did we use before?

app/controllers/api/storage.php
```diff
@@ -66,12 +64,13 @@
     ->param('enabled', true, new Boolean(true), 'Is bucket enabled?', true)
     ->param('maximumFileSize', (int) App::getEnv('_APP_STORAGE_LIMIT', 0), new Range(1, (int) App::getEnv('_APP_STORAGE_LIMIT', 0)), 'Maximum file size allowed in bytes. Maximum allowed value is ' . Storage::human(App::getEnv('_APP_STORAGE_LIMIT', 0), 0) . '. For self-hosted setups you can change the max limit by changing the `_APP_STORAGE_LIMIT` environment variable. [Learn more about storage environment variables](docs/environment-variables#storage)', true)
     ->param('allowedFileExtensions', [], new ArrayList(new Text(64), APP_LIMIT_ARRAY_PARAMS_SIZE), 'Allowed file extensions. Maximum of ' . APP_LIMIT_ARRAY_PARAMS_SIZE . ' extensions are allowed, each 64 characters long.', true)
+    ->param('compression', 'none', new WhiteList(['none', 'gzip', 'zstd']), 'Compression algorithm choosen for compression. For file size above ' . Storage::human(APP_STORAGE_READ_BUFFER, 0) . ' compression is skipped even if it\'s enabled', true)
```
eldadfux, line 67: We should name the options here, and would be cool if we can have links to wikipedia pages of each compression.

composer.json
```diff
@@ -57,7 +57,7 @@
         "utopia-php/preloader": "0.2.*",
         "utopia-php/domains": "1.1.*",
         "utopia-php/swoole": "0.3.*",
-        "utopia-php/storage": "0.9.*",
+        "utopia-php/storage": "dev-feat-zstd-compression",
```
eldadfux, line 60: Reminder to replace to the tag.

src/Appwrite/Utopia/Response/Model/Bucket.php
```diff
@@ -66,6 +66,13 @@ public function __construct()
                 'example' => ['jpg', 'png'],
                 'array' => true
             ])
+            ->addRule('compression', [
+                'type' => self::TYPE_STRING,
+                'description' => 'Compression method chosen.',
```
eldadfux, line 71: Let's list the options with links to wikipedia.

eldadfux (changes requested):

app/config/collections.php
lohanidamodar: it's a new field, previously we only had it on file as `algorithm` attribute.

app/controllers/api/storage.php
lohanidamodar: May be adding constant in Storage and getting list of supported compression from as `Storage::supportedCompression` would make sense?

src/Appwrite/Utopia/Response/Model/Bucket.php
lohanidamodar: In the description in both places?

composer.json
lohanidamodar: done

app/controllers/api/storage.php
lohanidamodar: done

src/Appwrite/Utopia/Response/Model/Bucket.php
lohanidamodar: done

app/controllers/api/storage.php
```diff
@@ -64,7 +64,7 @@
...
     ->param('allowedFileExtensions', [], new ArrayList(new Text(64), APP_LIMIT_ARRAY_PARAMS_SIZE), 'Allowed file extensions. Maximum of ' . APP_LIMIT_ARRAY_PARAMS_SIZE . ' extensions are allowed, each 64 characters long.', true)
-    ->param('compression', 'none', new WhiteList(['none', 'gzip', 'zstd']), 'Compression algorithm choosen for compression. For file size above ' . Storage::human(APP_STORAGE_READ_BUFFER, 0) . ' compression is skipped even if it\'s enabled', true)
+    ->param('compression', 'none', new WhiteList([COMPRESSION_TYPE_NONE, COMPRESSION_TYPE_GZIP, COMPRESSION_TYPE_ZSTD]), 'Compression algorithm choosen for compression. Can be one of ' . COMPRESSION_TYPE_NONE . ',' . COMPRESSION_TYPE_GZIP . '(https://en.wikipedia.org/wiki/Gzip), or ' . COMPRESSION_TYPE_ZSTD . '(https://en.wikipedia.org/wiki/Zstd), For file size above ' . Storage::human(APP_STORAGE_READ_BUFFER, 0) . ' compression is skipped even if it\'s enabled', true)
```
abnegate, line 67: We could use markdown here to make the compression types themselves clickable from docs like:

`'[' . COMPRESSION_TYPE_GZIP . '](https://en.wikipedia.org/wiki/Gzip)'`

Applies to all doc updates

abnegate: Let's make sure there's a space following each `,` too

Dockerfile
```diff
@@ -35,7 +35,7 @@ ENV PHP_REDIS_VERSION=5.3.7 \
     PHP_IMAGICK_VERSION=3.7.0 \
     PHP_YAML_VERSION=2.2.2 \
     PHP_MAXMINDDB_VERSION=v1.11.0 \
-    PHP_ZSTD_VERSION="master"
+    PHP_ZSTD_VERSION="4504e4186e79b197cfcb75d4d09aa47ef7d92fe9 "
```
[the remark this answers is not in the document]
christyjacob4: Don't we have a tag to use ?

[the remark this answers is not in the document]
lohanidamodar: We do, but it's from October, and @abnegate  is using master for test?

app/controllers/api/storage.php
```diff
@@ -1018,12 +1031,21 @@
...
+                $compressor = new Zstd();
+                $source = $compressor->decompress($source);
+                break;
+            case 'gzip':
+                if (empty($source)) {
+                    $source = $deviceFiles->read($path);
+                }
+                $compressor = new GZIP();
+                $source = $compressor->decompress($source);
+                break;
```
christyjacob4, line 1048: Dont we need a default case?  
Because when we compress, we are defaulting to GZIP 
https://github.com/appwrite/appwrite/pull/3737/files#diff-02d3b2d220ab36c4733052fd1e3e017df738b47f7b3033dc8a93efbb7752436bR513

app/controllers/api/storage.php
```diff
@@ -882,8 +888,15 @@
...
-            $source = $compressor->decompress($source);
+        switch ($algorithm) {
+            case 'zstd':
+                $compressor = new Zstd();
+                $source = $compressor->decompress($source);
+                break;
+            case 'gzip':
+                $compressor = new GZIP();
+                $source = $compressor->decompress($source);
+                break;
```
christyjacob4, line 899: Dont we need a default case?  
Because when we compress, we are defaulting to GZIP 
https://github.com/appwrite/appwrite/pull/3737/files#diff-02d3b2d220ab36c4733052fd1e3e017df738b47f7b3033dc8a93efbb7752436bR513

app/controllers/api/storage.php
```diff
@@ -1159,12 +1181,21 @@
...
+                $compressor = new Zstd();
+                $source = $compressor->decompress($source);
+                break;
+            case 'gzip':
+                if (empty($source)) {
+                    $source = $deviceFiles->read($path);
+                }
+                $compressor = new GZIP();
+                $source = $compressor->decompress($source);
+                break;
```
christyjacob4, line 1198: Dont we need a default case?  
Because when we compress, we are defaulting to GZIP 
https://github.com/appwrite/appwrite/pull/3737/files#diff-02d3b2d220ab36c4733052fd1e3e017df738b47f7b3033dc8a93efbb7752436bR513

christyjacob4 (changes requested):
Please address the comments

app/controllers/api/storage.php
lohanidamodar: fixed

Dockerfile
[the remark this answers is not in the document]
christyjacob4: Its always safer to use a stable tag. 
The test is fine we can update it later 👍  
@abnegate  please have a task to update the tag in utopia-php/storage

app/controllers/api/storage.php
lohanidamodar: In those cases we are already checking algorithm is not equal to `none` and having default there to gzip to keep it consistent if people upgrade as previously we were not exposing it but were using gzip compression and when we upgrade we can set it to gzip, but yes we can remove the default case in above methods where you pointed if we want to, and make sure when we migrate, we set `compression` to `gzip`. Let me know what you think?

christyjacob4: As a general idea, its always good to have a default case in every switch case. I prefer we keep a default for the sake of readability so anyone who reads it in the future knows what the logic is

lohanidamodar:
@christyjacob4 Those default cases would just be empty, so no required.

christyjacob4 (approved):

PR #3742, opened 2022-08-31
Fix: UI with queries v2

Meldiron:
## What does this PR do?

Applies changes to Appwrite Console to use `queries` parameter instead of `limit`, `offset` and `order`.

## Test Plan

- [x] Manual QA

## Related PRs and Issues

x

Yes 🤠


app/init.php
```diff
@@ -81,7 +81,7 @@
 const APP_USERAGENT = APP_NAME . '-Server v%s. Please report abuse at %s';
 const APP_MODE_DEFAULT = 'default';
 const APP_MODE_ADMIN = 'admin';
-const APP_PAGING_LIMIT = 12;
+const APP_PAGING_LIMIT = 1;
```
abnegate, line 84: Is this change intentional?

app/views/console/databases/index.phtml
```diff
@@ -16,9 +16,9 @@
                 data-service="databases.list"
                 data-event="load,databases.create,databases.update,databases.delete"
                 data-param-search="{{router.params.search}}"
-                data-param-limit="<?php echo APP_PAGING_LIMIT; ?>"
-                data-param-offset="{{router.params.offset}}"
-                data-param-order-type="ASC"
+                data-param-queries="orderAsc(''),limit(<?php echo APP_PAGING_LIMIT; ?>),offset({{router.params.offset|orZero}})"
```
abnegate, line 19: Do we need to pass empty string to `orderAsc()`?

abnegate, line 19: We could replace `orZero` with the existing `pageCurrent` filter

app/init.php
christyjacob4, line 84: I believe this was done for testing and we might need to revert it back ?

christyjacob4 (commented):
I like the approach! Lets go ahead

app/views/console/databases/index.phtml
Meldiron: Not required, seems to be default behaviour. Removing it.

app/init.php
Meldiron: Gonna keep it for testing purposes, will remove at the end.

Meldiron: Gonna keep it for testing purposes, will remove at the end. Keeping Christy's comment opened so I don't forget.

app/views/console/databases/index.phtml
Meldiron: Are you sure? Formula there is `Math.ceil(parseInt($value || 0) / env.PAGING_LIMIT) + 1;`
With limit of 12 and value of 5, it gives 2. I think there is more logic to that filter.

All I need here is to have 0 if the value is missing.

abnegate: My mistake, looks likes (unsurprisingly) that's used to display the current number 😅

christyjacob4 (approved):

PR #3755, opened 2022-09-04
audit event decoupling

shimonewman:
## What does this PR do?

Decouple Audit events from Events in order to get more human-readable events.


eldadfux (approved):

app/controllers/api/functions.php
```diff
@@ -1240,6 +1246,7 @@
     ->desc('Retry Build')
     ->label('scope', 'functions.write')
     ->label('event', 'functions.[functionId].deployments.[deploymentId].update')
+    ->label('audits.event', 'build.create')
```
christyjacob4, line 1249: sholdnn't this be `deployment.update` ?

eldadfux: It should.

src/Appwrite/Event/Audit.php
```diff
@@ -108,6 +108,29 @@ public function getIP(): string
...
+    /**
+     * Returns the  audit event description.
+     *
+     * @return string
+     */
+    public function getEvent(): string
+    {
+        return $this->event;
+    }
+
```
TorstenDittmann, line 133: This is already part of the parent class 👍🏻

TorstenDittmann (changes requested):

src/Appwrite/Event/Audit.php
TorstenDittmann: So we don't need it here.

TorstenDittmann: https://github.com/appwrite/appwrite/blob/342c4a3253816b9fb72b6ede26a617e31e0c362e/src/Appwrite/Event/Event.php#L81-L101

shimonewman: Cool

app/controllers/api/functions.php
shimonewman: Ok

christyjacob4 (approved):

PR #3772, opened 2022-09-06
Fix datetime attr UI

Meldiron:
## What does this PR do?

- Fixes UI icon for datetime attribute
- Fixes format of datetime on UI to properly fill the input in existing document
- Fixes timezone awareness for datetime input fields in UI

## Test Plan

- [x] Manual QA

## Related PRs and Issues

Depends on: https://github.com/utopia-php/database/pull/181

Yes 😇


composer.json
```diff
@@ -51,7 +51,7 @@
         "utopia-php/cache": "0.6.*",
         "utopia-php/cli": "0.13.*",
         "utopia-php/config": "0.2.*",
-        "utopia-php/database": "0.24.*",
+        "utopia-php/database": "dev-fix-force-datetime-filter as 0.24.0",
```
Meldiron, line 54: Note to address before merging

Meldiron:
Some pics showcasing changes:

CleanShot 2022-09-07 at 14 51 11

CleanShot 2022-09-07 at 14 51 02

Aaaand a vid:


https://user-images.githubusercontent.com/19310830/188884051-1127c3ca-77d0-4539-a529-e778267a9400.mp4

public/scripts/services/form.js
```diff
@@ -25,6 +25,26 @@
                         value = null;
                     }
                     break;
+                case 'string-datetime':
+                    value = value.toString();
```
stnguyen90, line 29: What are the possible values that `value` can be?

public/scripts/services/form.js
```diff
@@ -25,6 +25,26 @@
...
+                        
+                            while (value.length < minLength) {
+                                value = `0${value}`;
+                            }
+                        
+                            return value;
+                        }
+                        
+                        const date = new Date(value);
+                        const isoTime = `${date.getUTCFullYear()}-${prefixWithZero(date.getUTCMonth() + 1)}-${prefixWithZero(date.getUTCDate())}T${prefixWithZero(date.getUTCHours())}:${prefixWithZero(date.getUTCMinutes())}:${prefixWithZero(date.getUTCSeconds())}.${prefixWithZero(date.getUTCMilliseconds(), 3)}+00:00`;
```
stnguyen90, line 44: Can we use [`Date.toISOString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) instead?

public/scripts/services/form.js
Meldiron: Good catch, always string as it's value form input. Removing toString()

public/scripts/services/form.js
Meldiron: What toISOString gives is `2022-09-11T13:00:00.000Z`, while our standard is ISO 8601 that looks like `2022-09-11T13:00:00.000+00:00`.

It.. Seems to work with both? Not sure why, might be some PHP magic that we shouldn't relay on. Doing it this hectic way ensures we use exact format the format of datetime that we have many tests for, I would keep it.

composer.json
Meldiron: Uses 0.25.x now.

christyjacob4 (approved):

stnguyen90 (approved):

TorstenDittmann (approved):

PR #3789, opened 2022-09-08
Feat app region

lohanidamodar:
## What does this PR do?

- Add regions support to Projects

## Test Plan

## Related PRs and Issues

(Write your answer here.)


.env
```diff
@@ -81,6 +81,9 @@ _APP_USAGE_DATABASE_INTERVAL=15
 _APP_USAGE_STATS=enabled
 _APP_LOGGING_PROVIDER=
 _APP_LOGGING_CONFIG=
+_APP_REGION_DEFAULT=default
```
eldadfux, line 84: @lohanidamodar @christyjacob4 we might want to use a key value syntax here for having proper labels we can use in the UI.

New York=nyc001,Frankfurt=fra001

WDYT?

eldadfux: @lohanidamodar we need to check if there is any way to support spaces in var values.

app/controllers/api/projects.php
```diff
@@ -364,6 +366,7 @@
     ->label('sdk.response.model', Response::MODEL_PROJECT)
     ->param('projectId', '', new UID(), 'Project unique ID.')
     ->param('name', null, new Text(128), 'Project name. Max length: 128 chars.')
+    ->param('region', '', new Whitelist([App::getEnv('_APP_REGIONS', [])]), 'Project URL.', true)
```
eldadfux, line 369: Description is missing.

src/Appwrite/Usage/Calculators/Database.php
```diff
@@ -95,6 +96,7 @@ protected function createOrUpdateMetric(string $projectId, string $metric, strin
                     'time' => $time,
                     'metric' => $metric,
                     'value' => $value,
+                    'region' => App::getEnv('_APP_REGION', 'default'),
```
eldadfux, line 99: Will be better to pass this to the constructor so we don't couple the calculator with the class `App`

src/Appwrite/Usage/Calculators/TimeSeries.php
```diff
@@ -315,6 +316,7 @@ private function createOrUpdateMetric(string $projectId, string $time, string $p
                     'metric' => $metric,
                     'value' => $value,
                     'type' => $type,
+                    'region' => App::getEnv('_APP_REGION', 'default'),
```
eldadfux, line 319: Same, will be better to pass this to the constructor so we don't couple the calculator with the class App.

eldadfux (changes requested):

eldadfux (approved):

eldadfux (approved):

PR #3792, opened 2022-09-08
feat: migration 1.0.0

TorstenDittmann:
> tbd


src/Appwrite/Migration/Version/V15.php
```diff
@@ -0,0 +1,1294 @@
...
+    /**
+     * Migrates a permission string
+     *
+     * @param string $permission
+     * @return string
+     */
+    protected function migratePermission(string $permission): string
+    {
+        return match ($permission) {
+            'role:all' => 'any',
```
stnguyen90, line 330: Do we want to do this as it would change the behavior to write operations between 0.15 and 1.0 (`role:all` for write operations in 0.15 is actually `role:member`).

TorstenDittmann: iirc role:all also includes guests right?

talking about role:member, did it stay the same?

stnguyen90: `role:all` includes guests, but it doesn't actually give guests access for write operations. For example, in 0.15, if you put `role:all` for function execute permission, a guest still can't actually execute the function.

TorstenDittmann: ah okay! gotcha 👍🏻 

Gonna add this case for write permission.

app/tasks/migrate.php
```diff
@@ -60,6 +63,13 @@
 
         while (!empty($projects)) {
             foreach ($projects as $project) {
+                /**
+                 * Skip user projects with id 'console'
+                 */
+                if ($project->getId() === 'console' && $project->getInternalId() !== 'console') {
```
fogelito, line 69: Is this getInternalId or getId?

TorstenDittmann: It's intended 👍🏻 

Some time ago you were able to create projects with the id `console` and they didn't work. This is just making sure to not migrate them.

composer.json
```diff
@@ -51,7 +51,7 @@
         "utopia-php/cache": "0.6.*",
         "utopia-php/cli": "0.13.*",
         "utopia-php/config": "0.2.*",
-        "utopia-php/database": "0.25.*",
+        "utopia-php/database": "dev-fix-update-attribute-datetime as 0.25.99",
```
christyjacob4, line 54:
```suggestion
        "utopia-php/database": "0.25.*",
```

stnguyen90 (commented):
I still haven't looked through everything thoroughly enough, but here are my questions/comments so far...

app/http.php
```diff
@@ -122,9 +124,9 @@
                 continue;
             }
             /**
-             * Skip to prevent 0.15 migration issues.
+             * Skip to prevent 0.16 migration issues.
              */
-            if ($key === 'databases' && $dbForConsole->exists(App::getEnv('_APP_DB_SCHEMA', 'appwrite'), 'collections')) {
+            if (in_array($key, ['cache', 'variables']) && $dbForConsole->exists(App::getEnv('_APP_DB_SCHEMA', 'appwrite'), 'bucket_1')) {
```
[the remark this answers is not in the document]
TorstenDittmann: when I start the server, it tries to create collection. But can't if its on 0.15, due to the new database library. This skips creating them when the project bucket is created and therefore everything had been initialized.

src/Appwrite/Migration/Migration.php
```diff
@@ -39,18 +41,7 @@ abstract class Migration
...
-        '0.13.3' => 'V12',
-        '0.13.4' => 'V12',
-        '0.14.0' => 'V13',
-        '0.14.1' => 'V13',
-        '0.14.2' => 'V13',
-        '0.15.0' => 'V14',
-        '0.15.1' => 'V14',
-        '0.15.2' => 'V14',
-        '0.15.3' => 'V14'
+        '1.0.0-RC1' => 'V15'
```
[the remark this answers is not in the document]
TorstenDittmann: yes 👍🏻

src/Appwrite/Migration/Migration.php
```diff
@@ -287,13 +293,15 @@ public function createAttributeFromCollection(Database $database, string $collec
...
+     * @param string|null $from
      * @return void
      * @throws \Exception
      * @throws \Utopia\Database\Exception\Duplicate
      * @throws \Utopia\Database\Exception\Limit
      */
-    public function createIndexFromCollection(Database $database, string $collectionId, string $indexId): void
+    public function createIndexFromCollection(Database $database, string $collectionId, string $indexId, string $from = null): void
     {
+        $from ??= $collectionId;
```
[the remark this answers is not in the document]
TorstenDittmann: in the `app/config/collections.php` we have a list of collections. In most cases the collection id is the same. But not for Collections and Buckets 👍🏻

src/Appwrite/Migration/Version/V15.php
```diff
@@ -0,0 +1,1464 @@
...
+                $this->pdo->prepare("ALTER TABLE IF EXISTS `{$this->projectDB->getDefaultDatabase()}`.`_{$this->project->getInternalId()}_{$table}` MODIFY {$attribute} DATETIME(3)")->execute();
+            } catch (\Throwable $th) {
+                Console::warning($th->getMessage());
+            }
+        }
+
+        /**
+         * Skip adding filter on internal attributes.
+         */
+        if (!str_starts_with($attribute, '_')) {
```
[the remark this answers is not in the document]
TorstenDittmann: createdAt and updateAt are not part of the mariadb metadata and are hardcoded in the database library. if we try to migrate them it fails 👍🏻

src/Appwrite/Migration/Migration.php
```diff
@@ -105,64 +97,76 @@ public function setProject(Document $project, Database $projectDB, Database $con
...
-                                if ($document && $new->getId() !== $document->getId()) {
-                                    throw new Exception('Duplication Error');
-                                }
-                            }
-                        }, $document, $callback);
-                    }
-                }, $documents, $callback);
+            Console::log($sum . ' / ' . $collectionCount);
+            foreach ($documents as $document) {
+                yield $document;
```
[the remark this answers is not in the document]
TorstenDittmann: thank you 👀

src/Appwrite/Migration/Migration.php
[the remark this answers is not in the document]
stnguyen90: I see...but where is this variable used in the function body? 

And are we using it in the V15 migration? I don't see it being passed into any of the calls 👀

PR #3794, opened 2022-09-09
Improve DX on Appwrite using Composer Scripts

stnguyen90:
## What does this PR do?

1. Update composer.json so that we can run `composer update` rather than `composer update --ignore-platform-reqs --optimize-autoloader --no-plugins --no-scripts --prefer-dist`
2. Add composer scripts to start appwrite and run tests

## Test Plan

N/A

## Related PRs and Issues

None

Yes

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -94,12 +94,13 @@
         "optimize-autoloader": true,
         "platform": {
             "php": "8.0",
-            "ext-imagick": "3.7.0",
-            "ext-yaml": "2.2.2",
-            "ext-redis": "5.3.7",
-            "ext-swoole": "4.8.10",
-            "ext-zstd": "0.11.0",
-            "ext-mongodb": "1.13.0"
+            "ext-imagick": "0.0.0",
+            "ext-yaml": "0.0.0",
+            "ext-redis": "0.0.0",
+            "ext-swoole": "0.0.0",
+            "ext-zstd": "0.0.0",
+            "ext-mongodb": "1.8.1",
+            "ext-sockets": "0.0.0"
         }
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -12,8 +12,8 @@
     "scripts": {
         "start": "docker compose up -d",
         "test": "docker compose exec appwrite test",
-        "test-unit": "@test /usr/src/code/tests/unit",
-        "test-e2e": "@test /usr/src/code/tests/e2e",
+        "test-unit": "@test --testsuite unit",
+        "test-e2e": "@test --testsuite e2e",
         "lint": "vendor/bin/phpcs",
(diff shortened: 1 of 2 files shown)


stnguyen90:
@TorstenDittmann, what do you think about these changes?

eldadfux (commented):
Would be great to add this shortcuts to the contribution guide.

composer.json
```diff
@@ -88,8 +91,15 @@
...
+        "optimize-autoloader": true,
         "platform": {
-            "php": "8.0"
+            "php": "8.0",
+            "ext-imagick": "3.7.0",
+            "ext-yaml": "2.2.2",
+            "ext-redis": "5.3.7",
+            "ext-swoole": "4.8.10",
+            "ext-zstd": "0.11.0",
+            "ext-mongodb": "1.13.0"
```
TorstenDittmann, line 102: I think it might be fair to say that we should probably use a complete wildcard or only the major version.

`"ext-imagick": "3.*",` or `"ext-imagick": "*",`

TorstenDittmann (approved):

composer.json
stnguyen90: Unfortunately, we can't:

```
$ composer update
Loading composer repositories with package information

                                
  [UnexpectedValueException]    
  Invalid version string "3.*"  
                                
```

TorstenDittmann: bummer, what about `^3`? 😄

stnguyen90: same:

```
$ docker run --rm --interactive --tty \
  --volume $PWD:/app \
  composer update
Loading composer repositories with package information

In VersionParser.php line 186:
                               
  Invalid version string "^3"  
```

composer.json
```diff
@@ -10,7 +10,10 @@
         }
     ],
     "scripts": {
-        "test": "vendor/bin/phpunit",
+        "start": "docker compose up -d",
+        "test": "docker compose exec appwrite test",
+        "test-unit": "@test /usr/src/code/tests/unit",
+        "test-e2e": "@test /usr/src/code/tests/e2e",
```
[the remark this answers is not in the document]
stnguyen90: done!

composer.json
stnguyen90: How about using a fake 0.0.0 value when possible? ext-mongodb requires at least 1.8.1 so I put that.

abnegate:
Note that this adds system PHP and composer as dev dependencies, or the need to wrap commands with `docker run` anyway

TorstenDittmann:
> Note that this adds system PHP and composer as dev dependencies, or the need to wrap commands with `docker run` anyway

These were already present with the other dependencies.

Right now I already need to do `composer update --ignore-platform-reqs` 👍🏻

PR #3798, opened 2022-09-09
Fix: 1.0 bugs from QA (3)

Meldiron:
## What does this PR do?

- Implements `enabled` boolean attribute on functions
- Fix bug when listVariables returned all variables (not just for a specific function)
- Remove queries from listVariables endpoint

## Test Plan

- [x] Manual QA

## Related PRs and Issues

x

Yes


app/controllers/api/functions.php
```diff
@@ -424,12 +424,13 @@
     ->param('events', [], new ArrayList(new ValidatorEvent(), APP_LIMIT_ARRAY_PARAMS_SIZE), 'Events list. Maximum of ' . APP_LIMIT_ARRAY_PARAMS_SIZE . ' events are allowed.', true)
     ->param('schedule', '', new Cron(), 'Schedule CRON syntax.', true)
     ->param('timeout', 15, new Range(1, (int) App::getEnv('_APP_FUNCTIONS_TIMEOUT', 900)), 'Maximum execution time in seconds.', true)
+    ->param('enabled', true, new Boolean(), 'Is collection enabled?', true)
```
eldadfux, line 427:
```suggestion
    ->param('enabled', true, new Boolean(), 'Is function enabled?', true)
```

app/controllers/api/functions.php
```diff
@@ -424,12 +424,13 @@
...
     ->inject('response')
     ->inject('dbForProject')
     ->inject('project')
     ->inject('user')
     ->inject('events')
-    ->action(function (string $functionId, string $name, array $execute, array $events, string $schedule, int $timeout, Response $response, Database $dbForProject, Document $project, Document $user, Event $eventsInstance) {
+    ->action(function (string $functionId, string $name, array $execute, array $events, string $schedule, int $timeout, bool $enabled, Response $response, Database $dbForProject, Document $project, Document $user, Event $eventsInstance) {
```
eldadfux, line 433: Don't we need this param in the function update endpoint as well?

app/controllers/api/functions.php
```diff
@@ -1376,6 +1389,9 @@
             $queries[] = Query::search('search', $search);
         }
 
+        // Apply internal queries
+        $queries[] = Query::equal('functionInternalId', [$function->getInternalId()]);
```
eldadfux, line 1393: What exactly does this line do?

eldadfux (changes requested):

app/controllers/api/functions.php
Meldiron: If it was not there, you would get **all** variables. I have a few functions, this is what my "variables" tab look like: (same on all of them)
CleanShot 2022-09-11 at 13 57 47

app/controllers/api/functions.php
Meldiron: Good point.

app/controllers/api/functions.php
eldadfux: Got it. Thanks

eldadfux: Wait, why aren't we using a subquery to get all the vars?!

Meldiron: To support pagination. As part of feature of 1.0 - queries on all list endpoints. Stuff like

```
Query.equal("name", "API_KEY")
```
(when you want to find by name, instead of by ID)

eldadfux: Why don't we do it for sessions?

eldadfux: I doubt we really need pagination for this kind of entity. With the new console we didn't even consider it.

Meldiron: Understandable. Let's remove queries support from this endpoint and use subQueries.

app/controllers/api/functions.php
Meldiron: We now have this in both createFunction and updateFunction endpoints.

app/controllers/api/functions.php
Meldiron: Removed queries and search feature from variables, and used subquery in the listVariables endpoint.

christyjacob4 (approved):

PR #3809, opened 2022-09-11
Fix functon scheduling loop logic, and double-loop prevention

Meldiron:
⚠️ This requires database migration 

## What does this PR do?

Currently, even a change to scheduling cron causes double executions:

CleanShot 2022-09-11 at 13 41 47

I fixed that by fixing `scheduleOriginal`. It no longer exists. So I use $function instead. Double execution still possible with rapid cron changes:

CleanShot 2022-09-11 at 13 46 36

By introducing `scheduleUpdatedAt` I also solved rapid cron changes, and there are no more double executions:

CleanShot 2022-09-11 at 14 15 59

---

What does rapid cron mean?

- Have cron set to `* * * * *`. It executes every minute, once.
- Set cron to `*/2 * * * *`, and then quickly back to `* * * * *`
- There are now 2 schedule recursive loops, and neither gets cancelled because both started with `* * * * *` and now value is `* * * * *`. So it works

Third screenshot solves it with `scheduleUpdatedAt`, since that is a value that would be different for these loops.

---

## Test Plan

- [x] Manual QA

## Related PRs and Issues

x

Yes


app/workers/functions.php
```diff
@@ -150,21 +150,25 @@ function: $function,
                     throw new Exception('Function not found (' . $function->getId() . ')');
                 }
 
-                if ($scheduleOriginal && $scheduleOriginal !== $function->getAttribute('schedule')) { // Schedule has changed from previous run, ignore this run.
+                if ($functionOriginal->getAttribute('schedule') !== $function->getAttribute('schedule')) { // Schedule has changed from previous run, ignore this run.
+                    return;
+                }
+
+                if ($functionOriginal->getAttribute('scheduleUpdatedAt') !== $function->getAttribute('scheduleUpdatedAt')) { // Double execution due to rapid cron changes, ignore this run.
```
stnguyen90, line 157: Instead of introducing a new attribute, can we make use of scheduledNext? If there's a duplicate execution scheduled, one would execute 1st and update the scheduledNext. When the 2nd one executes and fetches the function from the DB, there would be a different scheduledNext than from the event. What do you think?

Meldiron: Nice idea! I will attempt this and see if it works as expected. Sounds logical.

Meldiron: Gave it some more thinking, and I would be more confident with separate attribute. That way, the attribute is directly linked to corn itself.

The problem I can see is that if you scale and have multiple workers, it will result in multiple parallel jobs. That means both jobs would be taken at the same time, so they both pass the check. 

Also, I know Torsten plans to work on a new queuing library that would solve this with jobIds. I don't see new attribute as too big of a concern, since it will be removed and replaced with whole different attribute in future - no data migration needed.

stnguyen90: Okay sure. Let's at least add a comment about why the `scheduleUpdatedAt` attribute is needed (maybe in `app/config/collections.php`) and that it can be removed once jobIds come.

Meldiron: Added comment into collections.php

stnguyen90 (approved):

PR #3824, opened 2022-09-13
Re-use same image for all appwrite containers during dev

stnguyen90:
## What does this PR do?

During build, every appwrite service builds an image even though they are all the same. This change speeds up the build step by only having 1 service build and then every other container re-use that built image.

## Test Plan

Automated tests still pass and Build Appwrite step is faster:

image

https://github.com/appwrite/appwrite/actions/runs/3047580615/jobs/4915521735

## Related PRs and Issues

None

Yes


eldadfux (approved):

eldadfux:
As @Meldiron suggested, I would check if it's worth to mention something on the contribution guide.

TorstenDittmann (approved):
There is some downside to it, commands like `docker compose up -d --build appwrite-realtime` will not work anymore.

But I think its a fair trade off 👍🏻

stnguyen90:
> As @Meldiron suggested, I would check if it's worth to mention something on the contribution guide.

@eldadfux, I've updated the contributing guide.

TorstenDittmann (approved):

Issue #3707, opened 2022-08-21
Can't read env vars in Swift 5.5 function


mbazs:
### 👟 Reproduction steps

I've created a function in Swift 5.5, it is compiled and deployed correctly by Appwrite:

```
func main(req: RequestValue, res: RequestResponse) async throws -> RequestResponse {
    return res.json(data: [
        "areDevelopersAwesome": true,
        "hello": 100,
        "reqdata": "endpoint: \(req.env["APPWRITE_FUNCTION_ENDPOINT"] ?? "<none>")"
    ])
}
```

I do the deploy from CLI by such **appwrite.json** (snippet):

```
...
 "functions": [
        {
            "$id": "entry",
            "name": "entry",
            "runtime": "swift-5.5",
            "path": "entry",
            "entrypoint": "Sources/entry/index.swift",
            "vars": {
                "APPWRITE_FUNCTION_ENDPOINT": "https://myhost.com/v1",
                "APPWRITE_FUNCTION_API_KEY": "e73b4fc631215e9b6391ec.....f54322ed",
                "APPWRITE_FUNCTION_PROJECT_ID": "62.....6d8ac8"
            },
            "ignore": [
                ".build",
                ".swiftpm"
            ],
            "execute": [],
            "events": [],
            "schedule": "",
            "timeout": 15
        }
    ]
...
```

**These env vars can be seen correctly in the console after deployment.**

The function runs successfully from the console (by the **Execute Now** button) except no env values can be read during execution.

```
{
  "areDevelopersAwesome" : true,
  "hello" : 100,
  "reqdata" : "endpoint: <none>"
}
```

Also tried with APPWRITE_FUNCTION_PROJECT_ID which mentioned in the docs but it can't be read when I set it in the CLI JSON as above and when I unset it also doesn't come by default as stated in the docs. It seems no env vars come ever at all.

**UPDATE: the env var shows up when the same function is called from client code, not from the console! So the Execute Now button behaves differently??**

### 👍 Expected behavior

Env vars can be read in the function Swift code.

### 👎 Actual Behavior

Can't read the env vars in Swift function.

### 🎲 Appwrite version

Version 0.15.x

### 💻 Operating system

Linux

### 🧱 Your Environment

No special settings, everything runs correctly except above.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

[...]

PR #3885, opened 2022-09-19
Fix: Function null variables

Meldiron:
## What does this PR do?

In strictly-typed languages we noticed people were getting `null` values in variables when the type states that the value is always a string.

This PR makes sure there are no null values in variables - instead, an empty string will be present. I consider this a better solution as it will still allow people to do `console.log(req.variables)` to see what they can expect, instead of searching over docs.

## Test Plan

- [x] Current tests must pass
 
## Related PRs and Issues

- Issue: https://github.com/appwrite/appwrite/issues/3707
- PoC in OPR: https://github.com/open-runtimes/open-runtimes/pull/88 (do not merge)

Yes 👍


abnegate (approved):
Looks good 👍

app/workers/functions.php
```diff
@@ -270,18 +270,18 @@ private function execute(
...
-            'APPWRITE_FUNCTION_RUNTIME_NAME' => $runtime['name'],
-            'APPWRITE_FUNCTION_RUNTIME_VERSION' => $runtime['version'],
-            'APPWRITE_FUNCTION_TRIGGER' => $trigger,
-            'APPWRITE_FUNCTION_EVENT' => $event,
-            'APPWRITE_FUNCTION_EVENT_DATA' => $eventData,
-            'APPWRITE_FUNCTION_DATA' => $data,
-            'APPWRITE_FUNCTION_PROJECT_ID' => $project->getId(),
-            'APPWRITE_FUNCTION_USER_ID' => $user->getId(),
-            'APPWRITE_FUNCTION_JWT' => $jwt,
+            'APPWRITE_FUNCTION_DEPLOYMENT' => $deploymentId ?? '',
```
christyjacob4, line 275: Variables like functionID and projectID are never empty, so we can leave them as it is

PR #3900, opened 2022-09-22
Add get team membership tests

abnegate:
## What does this PR do?

Fix get team membership response model label

## Test Plan

Added team get membership tests

Yes


christyjacob4 (changes requested):
If the test case is the same can we not add it in the underlying base class ( or trait ) ?

abnegate:
> If the test case is the same can we not add it in the underlying base class ( or trait ) ?

@christyjacob4 They depend on a previous test; if I move both into the base trait, it breaks lots of the other tests due to counts and different expected outputs for server vs client. I could update all of them but I think this is a case where a little duplication is not so bad. What do you think?

christyjacob4:
@abnegate sounds good. 👍

PR #3908, opened 2022-09-23
Fix schedule loop triggering

Meldiron:
## What does this PR do?

If you set schedule in `createFunction`, the schedule loop is never started. This occurs in Appwrite CLI. This PR solves it.

Also, this PR makes it so `updateFunction` forcefully re-triggers schedule loop. Old loop will be stopped, so no double executions. This is useful, for instance, when you clean your Redis instance and loop is gone. By clicking "Update" on function settings you can easily re-trigger the loop.

## Test Plan

Manual QA for all edge cases I can think of:

CleanShot 2022-09-23 at 08 12 10@2x

## Related PRs and Issues

x

Yes 😇

### DIFF
--- a/app/controllers/api/functions.php
+++ b/app/controllers/api/functions.php
@@ -71,8 +71,13 @@ App::post('/v1/functions')
     ->param('enabled', true, new Boolean(), 'Is function enabled?', true)
     ->inject('response')
     ->inject('dbForProject')
+    ->inject('project')
+    ->inject('user')
     ->inject('events')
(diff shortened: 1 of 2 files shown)


Meldiron:
Also did QA from CLI to ensure the specific use case is addressed.

CleanShot 2022-09-23 at 08 19 10@2x

TorstenDittmann (approved):

s3ppo:
😍

Issue #3893, opened 2022-09-21
🚀 Feature: Add Email Verification with a verification code instead of a URL


timurbutt:
### 🔖 Feature description

A User wants to create an account and needs to verify his email address. He will receive a 4-6 digit code inside that mail and can then enter that code inside the app. This could be done by linking a randomly generated number to the UserID that is created at signup and also adding a timestamp so the app knows how old the code. This would also make it possible to give the code an expiration time.

### 🎤 Pitch

The main reason why this feature would be great is because I could imagine that its way easier to implement, and also it would be a cool way to verify the User with a code instead of an URL. 

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

christyjacob4:
We have a draft PR ready. Few more things to fix and it should be out in the next Appwrite version 😉

FaisalMohammadi:
is their any update to this issue?

christyjacob4:
@FaisalMohammadi @timurbutt sorry for the delay but this issue was deprioritised during the previous releases.

We plan to release it alongside the next Appwrite version.

FaisalMohammadi:
Thats perfect thank you. Is their any fixed date, when would be the next version released?

cddqssc:
Email OTP is very very good, I can’t wait to use it

alangautham:
Please release OTP email verification... the present implementation URL verification is definitely the way to irritate users

FaisalMohammadi:
is their any update to this?

Vedsaga:
is there any update...?

PR #3912, opened 2022-09-23
Feat email code

christyjacob4:
## What does this PR do?

Until now we could only perform email verification using a redirect URL sent via email. A recent feature request asked for the same verification to be performed using an OTP style code. This PR allows us to perform email verification using a short lived code

https://user-images.githubusercontent.com/20852629/191921011-3bd8e42b-fd9b-4e48-9015-0a3855a8a94c.mov

TODOs

- [ ] Tests
- [ ] Decide whether we need a new endpoint for this or reuse the existing one ?
- [ ] Update language templates for other languages
- [ ] Maybe introduce an env var to set the token expiration time

## Test Plan

TBD

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/3893

YES


christyjacob4:
I realised if we reuse the existing endpoint, it will lead to a breaking change . However, with a new endpoint, we can ship this feature without a breaking change. Thoughts ?
@stnguyen90 @Meldiron @eldadfux @TorstenDittmann

stnguyen90:
> I realised if we reuse the existing endpoint, it will lead to a breaking change . However, with a new endpoint, we can ship this feature without a breaking change.

@christyjacob4, what makes it breaking?

christyjacob4:
Screenshot 2022-10-25 at 7 27 14 PM
@stnguyen90 The new parameter added to the endpoint would be a breaking change for languages like Java, Kotlin, etc.

app/config/locale/translations/en.json
```diff
@@ -3,12 +3,18 @@
...
+    "emails.verificationUrl.body": "Follow this link to verify your email address.",
+    "emails.verificationUrl.footer": "If you didn’t ask to verify this address, you can ignore this message.",
+    "emails.verificationUrl.thanks": "Thanks",
+    "emails.verificationUrl.signature": "{{project}} team",
+    "emails.verificationCode.subject": "Account Verification",
+    "emails.verificationCode.hello": "Hey {{name}}",
+    "emails.verificationCode.body": "Please use the following code to verify your email address. Note that this code is only valid for 15 minutes.",
+    "emails.verificationCode.footer": "If you didn’t ask to verify this address, you can ignore this message.",
+    "emails.verificationCode.thanks": "Thanks",
+    "emails.verificationCode.signature": "{{project}} team",
```
abnegate, line 17: Lots of these are duplicate, what do you think about keeping ones that are as `"emails.verification.subject"` and specifying per type as `"emails.verification.code.body"` for example? Then we could fetch them without having to check the type like `"emails.verification.{$type}.body"`

abnegate:
> The new parameter added to the endpoint would be a breaking change for languages like Java, Kotlin, etc.

This would not be breaking as the parameter is optional and is added as the last parameter; it would be automatically set to null, then excluded from the request so the API default would be used

eldadfux:
Please address Jake's comments. I agree calling the param `code` might be problematic. Also, this makes translations and custom email templates much more complex.

Have we considered adding a new endpoint for verification with code with its own template for easier translation maintenance? Not sure how it should be structured, but would love to hear your thoughts.

binaryfire:
It'd be great to move this forward. @Cvr421 offered to assist with this in another thread. Perhaps he could help?

Some new info re: OTPs:

- Apple have (finally) started supporting PWAs. They're recommending against Magic Links and suggest OTP codes instead. See timestamp 7:40 in this video: https://developer.apple.com/videos/play/wwdc2023/10120/

- We've been using OTPs with Supabase and it's been great: https://supabase.com/docs/reference/javascript/auth-signinwithotp This PR would bring Appwrite's auth up to feature-parity with Supabase.

FaisalMohammadi:
@binaryfire you are absolutely right. I am also waiting for this feature for very long time. Will be also happy to have this feature as soon as possible.

ajitsinghkaler:
This will be very helpful

christyjacob4:
Not relevant anymore since it was released in #7422

PR #3913, opened 2022-09-23
Fix: loose booleans

Meldiron:
## What does this PR do?

In a few places we allowed loose boolean (meaning string or integer should work), but never converted it properly to boolean. That caused all deployments to be always marked as active if you provided `active: "false"` instead of `active: false`, even tho validator allowed it.

This PR:

- Fixes bugs regarding loose booleans
- Makes approach consistent everywhere

## Test Plan

- [x] Manual QA. Also tested with CLI - there we stringify body if a file is present in the request (making boolean value `activate` loose)

## Related PRs and Issues

x

Ýes

### DIFF
--- a/app/controllers/api/avatars.php
+++ b/app/controllers/api/avatars.php
@@ -313,9 +313,9 @@ App::get('/v1/avatars/qr')
     ->param('margin', 1, new Range(0, 10), 'Margin from edge. Pass an integer between 0 to 10. Defaults to 1.', true)
     ->param('download', false, new Boolean(true), 'Return resulting image with \'Content-Disposition: attachment \' headers for the browser to start downloading it. Pass 0 for no header, or 1 for otherwise. Default value is set to 0.', true)
     ->inject('response')
-    ->action(function (string $text, int $size, int $margin, bool $download, Response $response) {
+    ->action(function (string $text, int $size, int $margin, mixed $downloadLoose, Response $response) {
+        $download = in_array($downloadLoose, ['1', 'true', 1, true], true);
 
-        $download = ($download === '1' || $download === 'true' || $download === 1 || $download === true);
         $options = new QROptions([
             'addQuietzone' => true,
             'quietzoneSize' => $margin,
--- a/app/controllers/api/functions.php
+++ b/app/controllers/api/functions.php
@@ -610,7 +610,8 @@ App::post('/v1/functions/:functionId/deployments')
     ->inject('project')
     ->inject('deviceFunctions')
     ->inject('deviceLocal')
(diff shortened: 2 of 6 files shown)


app/controllers/api/avatars.php
```diff
@@ -313,9 +313,9 @@
     ->param('margin', 1, new Range(0, 10), 'Margin from edge. Pass an integer between 0 to 10. Defaults to 1.', true)
     ->param('download', false, new Boolean(true), 'Return resulting image with \'Content-Disposition: attachment \' headers for the browser to start downloading it. Pass 0 for no header, or 1 for otherwise. Default value is set to 0.', true)
     ->inject('response')
-    ->action(function (string $text, int $size, int $margin, bool $download, Response $response) {
+    ->action(function (string $text, int $size, int $margin, mixed $downloadLoose, Response $response) {
```
eldadfux, line 316:
```suggestion
    ->action(function (string $text, int $size, int $margin, mixed $download, Response $response) {
```

No point in adding the postfix, I was a bit confused by it, and how dependency injection even allow it to work. We can stay with original name and overwrite the value. True to all use cases in this PR.

eldadfux (changes requested):

TorstenDittmann (commented):
Can't we fix this on utopia level?

I feel like it feels kinda unnecessary here and just adds noise to the controllers.

app/controllers/api/avatars.php
Meldiron: Addressed in all places

app/controllers/api/functions.php
```diff
@@ -610,7 +610,8 @@
     ->inject('project')
     ->inject('deviceFunctions')
     ->inject('deviceLocal')
-    ->action(function (string $functionId, string $entrypoint, mixed $code, bool $activate, Request $request, Response $response, Database $dbForProject, Event $events, Document $project, Device $deviceFunctions, Device $deviceLocal) {
+    ->action(function (string $functionId, string $entrypoint, mixed $code, mixed $activate, Request $request, Response $response, Database $dbForProject, Event $events, Document $project, Device $deviceFunctions, Device $deviceLocal) {
+        $activate = in_array($activate, ['1', 'true', 1, true], true);
```
christyjacob4, line 614: The boolean validator should only check for true and false. Not 1 and 0.
And the loose flag in the validator should also allow the strings 'true' and 'false'

christyjacob4: So lets update the Boolean Validator

Meldiron:
Solved already by another PR

PR #4309, opened 2022-10-06
fix: docker compose syntax

christyjacob4:
## What does this PR do?

The Syntax we use for storing storage environment variables no longer works since docker compose fixed a bug that was allowing it in the first place. This PR reverts this change 

## Test Plan

NA

## Related PRs and Issues

NA

YES


eldadfux:
Should we also change it in `compose.phtml`? Also, what was the bug, does it affect `x-logging`?

christyjacob4:
@eldadfux We don't use this syntax in `compose.phtml`
https://github.com/appwrite/appwrite/blob/fix-update-docker-compose/app/views/install/compose.phtml#L108-L133

No it doesn't affect `x-logging` . That is still valid syntax 👍

PR #4418, opened 2022-10-10
Add build time to deployments

PineappleIOnic:
## What does this PR do?

This PR adds the build time to deployments for easier access

Yes

### DIFF
--- a/app/controllers/api/functions.php
+++ b/app/controllers/api/functions.php
@@ -832,6 +832,7 @@ App::get('/v1/functions/:functionId/deployments')
             $result->setAttribute('status', $build->getAttribute('status', 'processing'));
             $result->setAttribute('buildStderr', $build->getAttribute('stderr', ''));
             $result->setAttribute('buildStdout', $build->getAttribute('stdout', ''));
+            $result->setAttribute('buildTime', $build->getAttribute('duration', 0));
         }
 
         $response->dynamic(new Document([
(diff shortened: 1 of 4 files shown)


TorstenDittmann (commented):
Can we extend the existing tests and also check if this needs to be added to the functions worker?

app/controllers/api/functions.php
```diff
@@ -832,6 +832,7 @@
             $result->setAttribute('status', $build->getAttribute('status', 'processing'));
             $result->setAttribute('buildStderr', $build->getAttribute('stderr', ''));
             $result->setAttribute('buildStdout', $build->getAttribute('stdout', ''));
+            $result->setAttribute('buildTime', $build->getAttribute('duration', 0));
```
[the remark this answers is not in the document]
PineappleIOnic: They are both correct, The deployment just gets the buildTime by checking how long the build took. We can rename the buildTime to duration if you like

app/workers/builds.php
```diff
@@ -181,6 +181,10 @@ protected function buildDeployment(Document $project, Document $function, Docume
             $build->setAttribute('stderr', $response['stderr']);
             $build->setAttribute('stdout', $response['response']);
 
+            /* Also update the deployment buildTime */
+            $deployment->setAttribute('buildTime', $response['duration']);
+            $deployment = $dbForProject->updateDocument('deployments', $deployment->getId(), $deployment);
```
TorstenDittmann, line 186: we don't need to update this right?

PineappleIOnic: I added this to help with events not receiving buildTime, if you don't mind about events getting it, we can remove it

TorstenDittmann: yeah, but do we need to update the document?

TorstenDittmann (approved):

Issue #4441, opened 2022-10-13
🚀 Feature: Make membership secret available


Bendix20:
### 🔖 Feature description

The secret of the membership should be available. Either after creating a new membership with the `Teams` class or as part of the `Membership` model (but only accessible from the own user or the Team owners).

### 🎤 Pitch

For now it's not possible to send the team invitation again. If the secret is available, it would be possible to send and resend a custom invitation email to the users. 
Also it would be possible to join the team in other ways. There could be a function inside the application to display open invitations with a possibility to directly join these teams.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

stnguyen90:
@Bendix20 , @Meldiron is this a duplicate of https://github.com/appwrite/appwrite/issues/3115?

rafagazani:
@stnguyen90 can we close this issue?

PR #4445, opened 2022-10-13
Fix team memberships secret exposure

Meldiron:
## What does this PR do?

When team membership is created by an API key or console user, a plain secret should be returned so developers can share the secret over custom channels, or implement re-invitation logic.

## Test Plan

- [x] Update existing tests
- [x] Add new tests to ensure the secret is not exposed
- [x] Add tests to ensure secret persistency

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/4441

Yes

Yes

### DIFF
--- a/app/controllers/api/teams.php
+++ b/app/controllers/api/teams.php
@@ -449,7 +449,18 @@ App::post('/v1/teams/:teamId/memberships')
         $events
             ->setParam('teamId', $team->getId())
             ->setParam('membershipId', $membership->getId())
+            ->setPayload($response->output(
+                $membership
+                ->setAttribute('teamName', $team->getAttribute('name'))
+                ->setAttribute('userName', $invitee->getAttribute('name'))
+                ->setAttribute('userEmail', $invitee->getAttribute('email'))
+                ->setAttribute('secret', $secret),
+                Response::MODEL_MEMBERSHIP
+            ))
         ;
+
+        // Hide secret for clients
+        $membership->setAttribute('secret', ($isPrivilegedUser || $isAppUser) ? $secret : '');
 
         $response
             ->setStatusCode(Response::STATUS_CODE_CREATED)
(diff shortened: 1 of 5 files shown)


app/controllers/api/teams.php
```diff
@@ -449,8 +449,19 @@
...
             ->setParam('teamId', $team->getId())
             ->setParam('membershipId', $membership->getId())
+            ->setPayload($response->output(
+                $membership
+                ->setAttribute('teamName', $team->getAttribute('name'))
+                ->setAttribute('userName', $invitee->getAttribute('name'))
+                ->setAttribute('userEmail', $invitee->getAttribute('email'))
+                ->setAttribute('secret', $secret),
+                Response::MODEL_MEMBERSHIP
+            ))
```
eldadfux, line 459: This seems repetitive. We do almost the same in the response method.

eldadfux (changes requested):
One comment, in the future, we should try to enforce secret hiding from the response model instead of multiple places in the codebase for better maintainability and reducing the chances of this being missed.

app/controllers/api/teams.php
Meldiron: I believe we already do that regarding Account/User response format. Should I try to refactor it as part of this PR?

eldadfux: Might be simpler on a separate PR that won't block this one. Where have you seen it?

Meldiron: Account response model: https://github.com/appwrite/appwrite/blob/master/src/Appwrite/Utopia/Response/Model/Account.php

Notice it extends User, but removes rules regarding passwords. That way password stuff never gets into account model, even if you return it from the action.

eldadfux: This would still leave the maintainer with the responsibility of returning different response for each auth method. Would also create too many response models. I was thinking more something in the direction of a a filter that could be assigned once.

HarshMN2345:

PR #4330, opened 2022-10-08
Feat db pools eldad

eldadfux:
## What does this PR do?

Updated for the database pools PR.

- [x] Renamed DatabasePool to Pools - no need to include namespace in the name itself.
- [ ] Decouple Utopia libraries from Pools:
    - [ ] `App::getEnv()`
    - [ ] `App::isDevelopment()`
    - [ ] Appwrite Exception
    - [ ] RedisCache
    - [ ] MariaDB
- [ ] Deprecate or adjust `Pools::getDatabase()` to use generic DSN connections 
- [x] Fix `_APP_DB__APP_DB_CONSOLEROOT_PASS` to be `_APP_DB_CONSOLE `
- [x] Hide `_APP_DB_CONSOLE` and `_APP_DB_PROJECT` from docs

Open Questions:
- Do we have to use Swoole Pool as a dependency or can we be completely independent? 

## Test Plan

Use existing tests

## Related PRs and Issues

#3423 

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

Yes

PR #4540, opened 2022-10-20
Feat redis sync

shimonewman:
- Cache re-invoking issue


app/workers/syncsIn.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+require_once __DIR__ . '/../../vendor/autoload.php';
```
eldadfux, line 3: Naming convention is wrong, should be `sync-in.php`

app/workers/syncsIn.php
```diff
@@ -0,0 +1,49 @@
...
+
+use Utopia\App;
+use Utopia\Cache\Adapter\Redis as RedisCache;
+use Utopia\Queue;
+use Utopia\Queue\Message;
+
+require_once __DIR__ . '/../init.php';
+
+define("CURRENT_REGION", App::getEnv('_APP_REGION', 'nyc1'));
```
eldadfux, line 12: Defining constants in multiple files in the code is not suggested. We can just use the env var directly. Assuming `nyc` is the default region is wrong. We can have `default` as the default region.

app/workers/syncsIn.php
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * @return RedisCache
+ */
+function getCache(): RedisCache
+{
+    global $register;
+    return new RedisCache($register->get('cache'));
+}
+
+$connection = new Queue\Connection\Redis('redis');
```
eldadfux, line 23: As mentioned on init.php. This is not the way to create a connection. You have to sync branches an imitate what we do for DB connections and instances.

docker-compose.yml
```diff
@@ -206,6 +207,7 @@ services:
     volumes:
       - ./app:/usr/src/code/app
       - ./src:/usr/src/code/src
+      - ./vendor/utopia-php/cache:/usr/src/code/vendor/utopia-php/cache
```
eldadfux, line 210: Reminder - Do not commit these change

docker-compose.yml
```diff
@@ -108,6 +108,7 @@ services:
       - ./public:/usr/src/code/public
       - ./src:/usr/src/code/src
       - ./dev:/usr/local/dev
+      - ./vendor/utopia-php/cache:/usr/src/code/vendor/utopia-php/cache
```
eldadfux, line 111: Reminder - Do not commit these change

docker-compose.yml
```diff
@@ -253,6 +255,54 @@ services:
...
+      - appwrite
+    volumes:
+      - ./app:/usr/src/code/app
+      - ./src:/usr/src/code/src
+    depends_on:
+      - mariadb
+      - redis
+    environment:
+        - _APP_ENV
+        - _APP_REDIS_HOST
```
eldadfux, line 275: In my new branch, we have new env vars to support new connections. We should also keep the old `_APP_REDIS` vars for backward compatibility. You can see examples in the docker-compose file in my db-pool branch: https://github.com/appwrite/appwrite/pull/4330

eldadfux (changes requested):
Left my review, please read all comments before starting to adapt the PR as they kind of complete each other.

app/config/regions.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+return [
```
eldadfux, line 3: This should be updated before merging. No need for default region.

app/config/services.php
```diff
@@ -186,4 +186,17 @@
         'optional' => false,
         'icon' => '',
     ],
+    'syncs' => [
+        'key' => 'edge',
+        'name' => 'edge',
+        'subtitle' => 'Appwrite\'s cache edge sync Endpoint',
```
eldadfux, line 192:
```suggestion
        'subtitle' => 'TBD,
```

app/config/services.php
```diff
@@ -186,4 +186,17 @@
...
+        'description' => 'Cache edge sync Endpoint',
```
eldadfux, line 193:
```suggestion
        'description' => 'TBD',
```

app/controllers/api/edge.php
```diff
@@ -0,0 +1,72 @@
...
+use Utopia\Validator\ArrayList;
+use Utopia\Validator\Text;
+
+App::init()
+    ->groups(['edge'])
+    ->inject('request')
+    ->action(function (Request $request) {
+
+        $token = $request->getHeader('authorization');
+        $token = str_replace(["Bearer"," "], "", $token);
```
eldadfux, line 24: Aren't we missing here a space that needs to be removed?
`Authorization: Bearer {token}`

app/controllers/api/edge.php
```diff
@@ -0,0 +1,72 @@
...
+
+        if (empty($dsns)) {
+            Console::error("No Dsn found");
+        }
+
+        $dsn = explode('=', $dsns[0]);
+        $dsn = $dsn[1] ?? '';
+        $dsn = new DSN($dsn);
+
+        $client = new SyncIn('syncIn', new QueueRedis($dsn->getHost(), $dsn->getPort()));
```
eldadfux, line 65: This is not how we should implement it. By doing this, you create a new Redis connection on every API call. This should be handled as a resource that can use dependency injection.

app/init.php
```diff
@@ -529,12 +533,12 @@ function (mixed $value) {
             'multiple' => true,
             'schemes' => ['mariadb', 'mysql'],
         ],
-        'queue' => [
-            'type' => 'queue',
-            'dsns' => App::getEnv('_APP_CONNECTIONS_QUEUE', $fallbackForRedis),
-            'multiple' => false,
-            'schemes' => ['redis'],
-        ],
+//        'queue' => [
```
eldadfux, line 536: We can remove this part and create an independent resource for the `queue`. Would be nice if we can have a similar preparation for using other adapters than Redis for future flexibility and extendability.

app/init.php
```diff
@@ -1024,6 +1028,12 @@ function (mixed $value) {
     ]);
 }, []);
 
+//App::setResource('queue', function (Group $pools) {
```
eldadfux, line 1031: We probably need this now, but please avoid leaving comments in the code. If we need to trace back we can utilize Git for that.

app/init.php
```diff
@@ -928,6 +931,7 @@ function (mixed $value) {
         if ($project->isEmpty()) {
             $user = new Document(['$id' => ID::custom(''), '$collection' => 'users']);
         } else {
+            //Todo fix cache save re-invoked
```
eldadfux, line 934: Reminder to fix this before merging.

app/init.php
```diff
@@ -1066,8 +1075,28 @@ function (mixed $value) {
             ->getResource()
         ;
     }
+    $cache  = new Cache(new Sharding($adapters));
+
+    $client = new SyncOut('syncOut', new QueueRedis(App::getEnv('_APP_REDIS_HOST', 'redis'), App::getEnv('_APP_REDIS_PORT', '6379')));
```
eldadfux, line 1080: Multiple cache write events can happen in the source code. We said we should aggregate them in an event and only fire them on the API shutdown (if needed). This will fire a message to the queue on every write.

app/tasks/maintenance.php
```diff
@@ -103,7 +103,8 @@ function notifyDeleteCache($interval)
                 ->trigger();
         }
 
-        // # of days in seconds (1 day = 86400s)
+
```
eldadfux, line 106: Extra line break, should be removed.

composer.json
```diff
@@ -48,7 +48,7 @@
         "utopia-php/abuse": "0.14.*",
         "utopia-php/analytics": "0.2.*",
         "utopia-php/audit": "0.15.*",
-        "utopia-php/cache": "0.8.*",
+        "utopia-php/cache": "dev-feat-redis-sync as 0.8.1",
```
eldadfux, line 51: Reminder to set proper version.

composer.json
```diff
@@ -48,7 +48,7 @@
...
         "utopia-php/cli": "0.13.*",
         "utopia-php/config": "0.2.*",
         "utopia-php/database": "dev-feat-update-cache-lib as 0.26.1",
```
eldadfux, line 54: Reminder to set proper version.

docker-compose.yml
```diff
@@ -268,6 +270,60 @@ services:
...
+        - _APP_REDIS_PORT
+        - _APP_REDIS_USER
+        - _APP_REDIS_PASS
+        - _APP_DB_HOST
+        - _APP_DB_PORT
+        - _APP_DB_SCHEMA
+        - _APP_DB_USER
+        - _APP_DB_PASS
+        - _APP_CONNECTIONS_DB_PROJECT
```
eldadfux, line 297: I don't think we use project specific databases in this container.

docker-compose.yml
```diff
@@ -268,6 +270,60 @@ services:
...
+      - ./src:/usr/src/code/src
+    depends_on:
+      - redis
+    environment:
+      - _APP_ENV
+      - _APP_REDIS_HOST
+      - _APP_REDIS_PORT
+      - _APP_REDIS_USER
+      - _APP_REDIS_PASS
+      - _APP_CONNECTIONS_DB_PROJECT
```
eldadfux, line 321: I don't think we use project specific databases in this container.

app/controllers/api/edge.php
```diff
@@ -0,0 +1,72 @@
...
+
+        $client->enqueue(['value' => ['keys' => $keys]]);
+
+        $response->dynamic(new Document([
+            'keys' => $keys
```
eldadfux, line 70: We probably need just a simple acknowledgement message, returning the full payload can be skipped. We can work the exact structure later.

docker-compose.yml
```diff
@@ -608,6 +664,37 @@ services:
...
+      - _APP_REDIS_PORT
+      - _APP_REDIS_USER
+      - _APP_REDIS_PASS
+      - _APP_DB_HOST
+      - _APP_DB_PORT
+      - _APP_DB_SCHEMA
+      - _APP_DB_USER
+      - _APP_DB_PASS
+      - _APP_SYNC_EDGE_INTERVAL
+      - _APP_CONNECTIONS_DB_PROJECT
```
eldadfux, line 692: I don't think we use any database connection in this container.

src/Appwrite/Utopia/Response/Model/EdgeSync.php
```diff
@@ -0,0 +1,43 @@
...
+use Appwrite\Utopia\Response\Model;
+
+class EdgeSync extends Model
+{
+    public function __construct()
+    {
+
+        $this
+            ->addRule('keys', [
+                'type' => self::TYPE_STRING,
```
eldadfux, line 15: As mentioned before, we don't need to return the full payload. It is important to mention this is also not the way to declare an array of strings. You can check one of the other examples in the source when it does needed.

app/worker.php
```diff
@@ -0,0 +1,123 @@
...
+            ->get($value)
+            ->pop()
+            ->getResource()
+        ;
+    }
+
+    return new Cache(new Sharding($adapters));
+}, ['register']);
+
+Server::setResource('ErrorLog', function (Logger $logger) {
```
eldadfux, line 56:
```suggestion
Server::setResource('logger', function (Logger $logger) {
```

app/worker.php
```diff
@@ -0,0 +1,123 @@
...
+    return function (Throwable $error, $action) use ($logger) {
```
eldadfux, line 57: Why is the callback returning a callback?

app/worker.php
```diff
@@ -0,0 +1,123 @@
...
+]);
+
+$connection = App::getEnv('_APP_CONNECTIONS_QUEUE', $fallbackForRedis);
+$dsns = explode(',', $connection ?? '');
+
+if (empty($dsns)) {
+    Console::error("Dsn not found");
+}
+
+$dsn = explode('=', $dsns[0]);
```
eldadfux, line 121: This parsing part can be skipped. We expect only a single DSN record in the variable.

app/tasks/sync-edge.php
```diff
@@ -0,0 +1,79 @@
...
+            if (App::getEnv('_APP_REGION', 'default') === 'default') {
+                return;
+            }
+
+            Console::info("[{$time}] Notifying workers with edges tasks every {$interval} seconds");
+
+            $time = DateTime::now();
+            $chunks = $database->find('syncs', [
+                Query::equal('region', [$region]),
+                Query::limit(500)
```
eldadfux, line 58: This is not good enough. We should have pagination logic here.

app/workers/sync-In.php
```diff
@@ -0,0 +1,53 @@
...
+use Utopia\Database\DateTime;
+use Utopia\Queue;
+use Utopia\Queue\Message;
+
+if (App::getEnv('_APP_REGION', 'default') === 'default') {
+    throw new Exception(Exception::GENERAL_SERVER_ERROR);
+}
+
+global $register;
+global $dsn;
```
eldadfux, line 18: Use dependency injection, even if need to copy-paste. We are working on a mechanism for share dependencies between http, realtime, workers, cli.

app/workers/sync-In.php
```diff
@@ -0,0 +1,53 @@
...
+            Console::info("[{$time}] Purging  {$key}");
+            $cache->purge($key);
+        }
+        $cache->setDisableListeners(false);
+    });
+
+$server
+    ->error()
+    ->inject('error')
+    ->inject('logError')
```
eldadfux, line 41: We should call this logger, and use the logger to log the error, we would usually avoid saving functions or primitives as dependencies.

app/workers/sync-out.php
```diff
@@ -0,0 +1,184 @@
...
+
+    return $payload;
+}
+
+/**
+ * @throws Authorization
+ * @throws Structure
+ * @throws Exception
+ */
+function handle($dbForConsole, $regions, $stack): void
```
eldadfux, line 85: Very generic name. Let's try to find a better one.

app/workers/sync-out.php
```diff
@@ -0,0 +1,184 @@
...
+            try {
+                $dbForConsole->createDocument('syncs', new Document([
+                    'region' => App::getEnv('_APP_REGION'),
+                    'target' => $code,
+                    'keys' => $stack,
+                    'status' => $response['status'],
+                    'payload' => $response['payload'],
+                ]));
+            } catch (\Throwable $th) {
+                $register->get('pools')->reclaim();
```
eldadfux, line 107: It would be much safer to use this on the error handler rather than here.

eldadfux (changes requested):

PR #4575, opened 2022-10-25
feat : Add error attribute to Collection Indexes and Attributes

everly-gif:
## What does this PR do?

Add error attribute to Collection Indexes and Attributes

## Test Plan

- [ ] Manual

## Related PRs and Issues

NIL

yes

yes


app/config/collections.php
```diff
@@ -263,6 +263,17 @@
                 'array' => false,
                 'filters' => [],
             ],
+            [
+                '$id' => ID::custom('error'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 255,
```
eldadfux, line 270: We might want to use a bigger limit here.

app/workers/databases.php
```diff
@@ -268,7 +280,13 @@ protected function createIndex(Document $database, Document $collection, Documen
             $dbForProject->updateDocument('indexes', $index->getId(), $index->setAttribute('status', 'available'));
         } catch (\Throwable $th) {
             Console::error($th->getMessage());
-            $dbForProject->updateDocument('indexes', $index->getId(), $index->setAttribute('status', 'failed'));
+            $dbForProject->updateDocument(
```
eldadfux, line 283: Some errors might expose sensitive server information. I suggest to only catch errors of a specific type, and that we have thrown ourselves. Avoiding native PDO errors that might expose information that we should keep private.

This might require us to add some base Database Exception that we could use to distinguish between our errors and native errors. This are the database exceptions we support today: https://github.com/utopia-php/database/tree/main/src/Database/Exception

src/Appwrite/Utopia/Response/Model/Attribute.php
```diff
@@ -28,6 +28,12 @@ public function __construct()
                 'default' => '',
                 'example' => 'available',
             ])
+            ->addRule('error', [
+                'type' => self::TYPE_STRING,
+                'description' => 'Error message',
```
eldadfux, line 33: Let's also explain when this field is relevant.

src/Appwrite/Utopia/Response/Model/Index.php
```diff
@@ -28,6 +28,12 @@ public function __construct()
                 'default' => '',
                 'example' => 'available',
             ])
+            ->addRule('error', [
+                'type' => self::TYPE_STRING,
+                'description' => 'Error message',
```
eldadfux, line 33: Let's also explain when this field is relevant.

app/workers/databases.php
```diff
@@ -321,7 +339,13 @@ protected function deleteIndex(Document $database, Document $collection, Documen
             $dbForProject->deleteDocument('indexes', $index->getId());
         } catch (\Throwable $th) {
             Console::error($th->getMessage());
-            $dbForProject->updateDocument('indexes', $index->getId(), $index->setAttribute('status', 'stuck'));
+            $dbForProject->updateDocument(
+                'indexes',
+                $index->getId(),
+                $index
+                ->setAttribute('status', 'stuck')
```
eldadfux, line 346:
```suggestion
                ->setAttribute('status', 'stuck')
```

What status types do we support? can we change this to error? cc @christyjacob4

eldadfux (changes requested):
Great work, left a few comments and points for discussion. Some extra adjustments in utopia/database might be required.

app/config/collections.php
christyjacob4: 2048 can be a good starting point.

app/config/collections.php
```diff
@@ -440,6 +451,17 @@
...
```
christyjacob4, line 458: Here as well. Let's use `2048`

app/workers/databases.php
```diff
@@ -158,7 +164,13 @@ protected function deleteAttribute(Document $database, Document $collection, Doc
             $dbForProject->deleteDocument('attributes', $attribute->getId());
         } catch (\Throwable $th) {
             Console::error($th->getMessage());
-            $dbForProject->updateDocument('attributes', $attribute->getId(), $attribute->setAttribute('status', 'stuck'));
+            $dbForProject->updateDocument(
+                'attributes',
+                $attribute->getId(),
+                $attribute
+                ->setAttribute('status', 'stuck')
```
christyjacob4, line 171: We should deprecate `stuck` and just use `failed` instead

app/workers/databases.php
christyjacob4: We currently support 
1. available
2. failed
3. stuck
4. deleting
5. processing

everly-gif: considering this comment [here](https://github.com/appwrite/appwrite/blob/master/app/workers/databases.php#L148-#L153), i think changing it to `failed` can be misleading.

`failed` indicates the attribute was never created in the db, whereas `stuck`
indicates the attribute is available in the db but the deletion operation failed. Since the exception thrown [here](https://github.com/appwrite/appwrite/blob/master/app/workers/databases.php#L318) would only be of a case where the index is available in the db, I think `stuck` makes sense by the definition.  If we want to deprecate `stuck` and go with `failed` i think we should redefine `failed` and what it means before doing so.

app/workers/databases.php
everly-gif: https://github.com/utopia-php/database/pull/218 has a base exception class implemented.

composer.json
```diff
@@ -51,10 +51,10 @@
         "utopia-php/cache": "0.8.*",
         "utopia-php/cli": "0.13.*",
         "utopia-php/config": "0.2.*",
-        "utopia-php/database": "0.28.*",
         "utopia-php/locale": "0.4.*",
         "utopia-php/registry": "0.5.*",
         "utopia-php/preloader": "0.2.*",
+        "utopia-php/database": "dev-add-exception-class as 0.28.0",
```
everly-gif, line 57: TODO : to revert/update composer after https://github.com/utopia-php/database/pull/218 is merged

PR #4605, opened 2022-10-29
Images updated

mishmanners:
## What does this PR do?

Added new images to the links section for Git Pod and Digital Ocean. These were broken. Now that they are uploaded directly, the links will never break!

Also added lots of updates for better readability and to encourage newbies to come as well, not only seasoned developers.

Yes I did!


stnguyen90:
Thanks for the PR! 🙏🏼 For now, I don't think we'll merge this in because their are errors and we'd rather contributors focus on the [issues we've created for Hacktoberfest](https://hacktoberfest.appwrite.io/).

PR #4618, opened 2022-11-01
Implement Variable Session Duration

PineappleIOnic:
## What does this PR do?

This PR adds a new endpoint: `PATCH /v1/projects/:projectId/auth/authDuration`that will edit the `authDuration` of sessions. Thie default is set to a year in minutes but can be altered using this endpoint. Editing this attribute will change how long sessions should last when they are generated.

## Test Plan

Tests have been created and implemented to make sure all functionality works without breaking current functionality.

Tests will set the session time to 1 minute, then create a new user account followed by a new session; it will then wait just over a minute and check if the session is still active. If the session is not active, then the test is passed and the session duration is reset back to a year.

Yes

Yes


app/config/collections.php
```diff
@@ -633,6 +633,17 @@
                 'array' => false,
                 'filters' => [],
             ],
+            [
+                '$id' => ID::custom('sessionDuration'),
```
eldadfux, line 637: Lets have `auth` as a prefix and group it with other auth related settings.

app/controllers/api/account.php
```diff
@@ -185,9 +186,11 @@
             throw new Exception(Exception::USER_BLOCKED); // User is in status blocked
         }
 
+        $sessionDuration = ($project->getAttribute('sessionDuration', 0) * 60) ?? Auth::TOKEN_EXPIRATION_LOGIN_LONG;
```
eldadfux, line 189:
```suggestion
        $duration = ($project->getAttribute('sessionDuration', 0) * 60) ?? Auth::TOKEN_EXPIRATION_LOGIN_LONG;
```

eldadfux: same for others. when we can and when not used as a namespaces in other areas, let's prioritize a single word as the name.

app/controllers/api/projects.php
```diff
@@ -67,10 +68,11 @@
     ->param('legalCity', '', new Text(256), 'Project legal City. Max length: 256 chars.', true)
     ->param('legalAddress', '', new Text(256), 'Project legal Address. Max length: 256 chars.', true)
     ->param('legalTaxId', '', new Text(256), 'Project legal Tax ID. Max length: 256 chars.', true)
+    ->param('sessionDuration', 525600, new Integer(), 'Session duration in minutes. Defaults to 1 year', true)
     ->inject('response')
     ->inject('dbForConsole')
     ->inject('dbForProject')
-    ->action(function (string $projectId, string $name, string $teamId, string $description, string $logo, string $url, string $legalName, string $legalCountry, string $legalState, string $legalCity, string $legalAddress, string $legalTaxId, Response $response, Database $dbForConsole, Database $dbForProject) {
+    ->action(function (string $projectId, string $name, string $teamId, string $description, string $logo, string $url, string $legalName, string $legalCountry, string $legalState, string $legalCity, string $legalAddress, string $legalTaxId, int $sessionDuration, Response $response, Database $dbForConsole, Database $dbForProject) {
```
eldadfux, line 75: No need to have this param in the creation of a project. This make the signature very busy (probably already too busy I know). We should try to replicate the pattern we use for other auth settings.

app/tasks/maintenance.php
```diff
@@ -99,7 +99,7 @@ function notifyDeleteExpiredSessions()
         {
             (new Delete())
                 ->setType(DELETE_TYPE_SESSIONS)
-                ->setDatetime(DateTime::addSeconds(new \DateTime(), -1 * Auth::TOKEN_EXPIRATION_LOGIN_LONG))
+                ->setDatetime(DateTime::addSeconds(new \DateTime(), -1 * Auth::TOKEN_EXPIRATION_LOGIN_LONG)) //TODO: Update to use project session expiration instead of default.
```
eldadfux, line 102: Reminder.

src/Appwrite/Utopia/Response/Model/Project.php
```diff
@@ -101,6 +101,12 @@ public function __construct()
                 'default' => '',
                 'example' => '131102020',
             ])
+            ->addRule('sessionDuration', [
```
eldadfux, line 104: @PineappleIOnic try to stick with existing pattern as much as possible. `authDuration`

eldadfux (changes requested):
@TorstenDittmann let's make sure we mention this feature on the release announcement or notes and social media at least.

app/tasks/maintenance.php
PineappleIOnic: I'll probably have to remove this comment as I don't think it's possible to access the project document from the maintenance task

eldadfux: In theory, I don't even think we need this timestamp. We can just delete all expired session in the deletes worker according to current timestamp + a safety factor of a few days and remove this var altogether. WDYT?

app/controllers/api/projects.php
```diff
@@ -374,16 +376,21 @@
...
 
         $project = $dbForConsole->getDocument('projects', $projectId);
 
         if ($project->isEmpty()) {
             throw new Exception(Exception::PROJECT_NOT_FOUND);
         }
 
+        if ($authDuration < 0 || $authDuration > 525600) {
+            throw new Exception('Session length must be between 0 and 525600 minutes');
+        }
```
TorstenDittmann, line 392: This should be handled by the validator

app/tasks/maintenance.php
TorstenDittmann: how do we know they are expired if the session length is variable?

TorstenDittmann (approved):

app/init.php
```diff
@@ -837,7 +837,7 @@ function (mixed $value) {
 
     if (
         $user->isEmpty() // Check a document has been found in the DB
-        || !Auth::sessionVerify($user->getAttribute('sessions', []), Auth::$secret)
+        || !Auth::sessionVerify($user->getAttribute('sessions', []), Auth::$secret, $project->getAttribute('authDuration', 0))
```
TorstenDittmann, line 840: shouldn't we fallback to 1 year instead of zero when `authDuration` is not set?

app/realtime.php
```diff
@@ -539,7 +539,7 @@ function () use ($register, $db, $redis) {
 
                 if (
                     empty($user->getId()) // Check a document has been found in the DB
-                    || !Auth::sessionVerify($user->getAttribute('sessions', []), Auth::$secret) // Validate user has valid login token
+                    || !Auth::sessionVerify($user->getAttribute('sessions', []), Auth::$secret, $project->getAttribute('authDuration', 0)) // Validate user has valid login token
```
TorstenDittmann, line 542: Same here, we should fall back to the Token const I guess

TorstenDittmann (commented):
I think we also need to add the `expire` field to getSession and listSessions for both account and users service

TorstenDittmann (approved):

app/controllers/api/projects.php
```diff
@@ -374,9 +376,10 @@
     ->param('legalCity', '', new Text(256), 'Project legal city. Max length: 256 chars.', true)
     ->param('legalAddress', '', new Text(256), 'Project legal address. Max length: 256 chars.', true)
     ->param('legalTaxId', '', new Text(256), 'Project legal tax ID. Max length: 256 chars.', true)
+    ->param('authDuration', 525600, new Range(0, 525600), 'Project session length in minutes. Max length: 525600 minutes.', true)
     ->inject('response')
     ->inject('dbForConsole')
-    ->action(function (string $projectId, string $name, string $description, string $logo, string $url, string $legalName, string $legalCountry, string $legalState, string $legalCity, string $legalAddress, string $legalTaxId, Response $response, Database $dbForConsole) {
+    ->action(function (string $projectId, string $name, string $description, string $logo, string $url, string $legalName, string $legalCountry, string $legalState, string $legalCity, string $legalAddress, string $legalTaxId, int $authDuration, Response $response, Database $dbForConsole) {
```
eldadfux, line 382: @PineappleIOnic this param should be updated by a standalone route just like we have for other auth settings. We should remove it from the project creation endpoint.

eldadfux: It should replicate this method, same for how we store the setting, no need for a dedicated column.
https://github.com/appwrite/appwrite/blob/842185e45d84f5fcfcc2330f3dfa680fe61504e9/app/controllers/api/projects.php#L471-L500

app/controllers/api/projects.php
```diff
@@ -395,6 +398,7 @@
                 ->setAttribute('legalCity', $legalCity)
                 ->setAttribute('legalAddress', $legalAddress)
                 ->setAttribute('legalTaxId', $legalTaxId)
+                ->setAttribute('authDuration', $authDuration * 60)
```
eldadfux, line 401: Same as the comment above. Replicate the behavior of previous auth related routes in the projects service.

app/init.php
```diff
@@ -917,6 +917,7 @@ function (mixed $value) {
         'legalCity' => '',
         'legalAddress' => '',
         'legalTaxId' => '',
+        'authDuration' => Auth::TOKEN_EXPIRATION_LOGIN_LONG, // 1 Year in seconds
```
eldadfux, line 920: Store the same way as other auth endpoints.

eldadfux (changes requested):

app/controllers/api/projects.php
```diff
@@ -495,6 +497,38 @@
         $response->dynamic($project, Response::MODEL_PROJECT);
     });
 
+
+App::patch('/v1/projects/:projectId/auth/authDuration')
```
eldadfux, line 501:
```suggestion
App::patch('/v1/projects/:projectId/auth/duration')
```

`auth` is already being used as the namespace in the parent directory.

app/controllers/api/projects.php
```diff
@@ -112,6 +113,7 @@
             'legalCity' => $legalCity,
             'legalAddress' => $legalAddress,
             'legalTaxId' => ID::custom($legalTaxId),
+            'authDuration' => Auth::TOKEN_EXPIRATION_LOGIN_LONG / 60,
```
eldadfux, line 116:
```suggestion
```

app/controllers/api/projects.php
```diff
@@ -495,6 +497,38 @@
...
+    ->groups(['api', 'projects'])
+    ->label('scope', 'projects.write')
+    ->label('sdk.auth', [APP_AUTH_TYPE_ADMIN])
+    ->label('sdk.namespace', 'projects')
+    ->label('sdk.method', 'updateAuthDuration')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_PROJECT)
+    ->param('projectId', '', new UID(), 'Project unique ID.')
+    ->param('authDuration', 525600, new Range(0, 525600), 'Project session length in minutes. Max length: 525600 minutes.')
```
eldadfux:
```suggestion
    ->param('duration', 525600, new Range(0, 525600), 'Project session length in minutes. Max length: 525600 minutes.')
```

The `Auth` context is already clear from the path. Always prefer using a single word for a name when possible.

src/Appwrite/Utopia/Response/Model/Project.php
```diff
@@ -101,6 +101,12 @@ public function __construct()
                 'default' => '',
                 'example' => '131102020',
             ])
+            ->addRule('authDuration', [
```
eldadfux, line 104: We also need to update the filter. See how `authLimit` behaves.

tests/e2e/Scopes/ProjectCustom.php
```diff
@@ -53,6 +53,7 @@ public function getProject(bool $fresh = false): array
             'legalCity' => '',
             'legalAddress' => '',
             'legalTaxId' => '',
+            'authDuration' => 525600
```
eldadfux, line 56: No need for this param anymore.

app/controllers/api/projects.php
```diff
@@ -495,6 +496,38 @@
         $response->dynamic($project, Response::MODEL_PROJECT);
     });
 
+
```
eldadfux, line 499:
```suggestion
```

Extra space

eldadfux (approved):

PR #4620, opened 2022-11-01
Sync 1.1.x

christyjacob4:
## What does this PR do?

Sync 1.1.x branch with `master` and `1.0.x`

## Test Plan

NA

## Related PRs and Issues

NA

YES

YES


app/config/collections.php
```diff
@@ -2248,21 +2248,21 @@
                 '$id' => ID::custom('_key_runtime'),
                 'type' => Database::INDEX_KEY,
                 'attributes' => ['runtime'],
-                'lengths' => [2048],
+                'lengths' => [768],
```
wess, line 2251: do we want to drop these back down, and has this been tested against Mongo?

stnguyen90 (approved):

app/config/collections.php
christyjacob4: Yes this is to ensure support for MySQL. MySQL has some limits on the maximum index length. 
I don't think it has been tested on Mongo, @fogelito can confirm. 
Does mongo use index lengths ?

Even if that were the case, it wouldn't be an issue for Appwrite 1.1 👍

fogelito: I Do not know I have a test for this on the new Database Tag and Mongo was passing , but better check

PR #4654, opened 2022-11-09
Fix null warnings

PineappleIOnic:
## What does this PR do?

This PR fixes various null safety issues that get warned against when running tests in Appwrite,

- Fixed the Avatar favicon API isset operator being converted to num then using isset operator causing the warning
- Added an isset operator to OpenSSL fields on files for Appwrite installations without encryption enabled
- dns_get_record does not throw errors, instead it logs a warning and returns false. Altered code accordingly. PHP Bug Report: https://bugs.php.net/bug.php?id=73149
- The dns_get_record will always give a warning, however the second null warning was fixed.

## Test Plan

Run the tests and check all Warnings were removed.

Yes

Yes


app/controllers/api/storage.php
```diff
@@ -581,10 +581,10 @@
...
                         ->setAttribute('sizeActual', $sizeActual)
                         ->setAttribute('algorithm', $algorithm)
-                        ->setAttribute('openSSLVersion', $openSSLVersion)
-                        ->setAttribute('openSSLCipher', $openSSLCipher)
-                        ->setAttribute('openSSLTag', $openSSLTag)
-                        ->setAttribute('openSSLIV', $openSSLIV)
+                        ->setAttribute('openSSLVersion', $openSSLVersion ?? null)
+                        ->setAttribute('openSSLCipher', $openSSLCipher ?? null)
+                        ->setAttribute('openSSLTag', $openSSLTag ?? null)
+                        ->setAttribute('openSSLIV', $openSSLIV ?? null)
```
christyjacob4, line 587: Do we really need null safety here? I dont think there is a different if we do `$openSSLVersion ?? null` vs just `$openSSLVersion`

app/controllers/api/storage.php
```diff
@@ -565,10 +565,10 @@
...
                         'chunksTotal' => $chunks,
                         'chunksUploaded' => $chunksUploaded,
-                        'openSSLVersion' => $openSSLVersion,
-                        'openSSLCipher' => $openSSLCipher,
-                        'openSSLTag' => $openSSLTag,
-                        'openSSLIV' => $openSSLIV,
+                        'openSSLVersion' => $openSSLVersion ?? null,
+                        'openSSLCipher' => $openSSLCipher ?? null,
+                        'openSSLTag' => $openSSLTag ?? null,
+                        'openSSLIV' => $openSSLIV ?? null,
```
christyjacob4, line 571: Same comment as below

christyjacob4 (changes requested):
Please address the comments

app/controllers/api/storage.php
PineappleIOnic: Without the null safety here we get a warning logged when we run tests since we are accessing a undefined value

christyjacob4 (approved):

christyjacob4 (approved):

christyjacob4:
@PineappleIOnic please resolve the conflicts?

PR #4655, opened 2022-11-09
feat: console submodule

TorstenDittmann:


app/controllers/web/console.php
```diff
@@ -1,609 +1,19 @@
...
-            ->setParam('header', [$header])
-            ->setParam('footer', [$footer])
-        ;
-
-        $response->html($layout->render());
-    });
-
-App::get('/error/:code')
-    ->groups(['web', 'console'])
+App::get('/console/*')
```
eldadfux, line 6: Does * actually work?!

app/controllers/web/console.php
```diff
@@ -1,609 +1,19 @@
...
+    ->alias('/')
+    ->alias('/invite')
+    ->alias('/login')
+    ->alias('/recover')
+    ->alias('/register')
```
eldadfux, line 11: Very elegant!

app/http.php
```diff
@@ -51,7 +51,7 @@
     Console::success('Reload completed...');
 });
 
-Files::load(__DIR__ . '/../public');
+Files::load(__DIR__ . '/../console');
```
eldadfux, line 54: In theory, I think we can load both folders.

eldadfux (commented):
We also need to remove a lot of .phtml files, but not all!

app/http.php
TorstenDittmann: we don't need the public anymore 👍🏻

app/controllers/web/console.php
TorstenDittmann: It looks like 😄

Dockerfile
```diff
@@ -14,12 +14,9 @@ RUN composer install --ignore-platform-reqs --optimize-autoloader \
...
-WORKDIR /usr/local/src/
+COPY app/console /usr/local/src/console
 
-COPY package-lock.json /usr/local/src/
-COPY package.json /usr/local/src/
-COPY gulpfile.js /usr/local/src/
-COPY public /usr/local/src/public
+WORKDIR /usr/local/src/console
 
 RUN npm ci
```
stnguyen90, line 21: To make better of the image cache, shouldn't we only copy the package.json and package-lock.json first, run `npm ci` and then copy the rest of the files and run `npm run build`?

app/controllers/general.php
```diff
@@ -566,32 +566,6 @@
         );
     });
 
-App::get('/manifest.json')
```
stnguyen90, line 569: why was this removed?

TorstenDittmann: This should be handled by the console, not appwrite 👍🏻 

If at some point we are going to add it for a PWA - we will do it there.

abnegate (approved):

PR #4665, opened 2022-11-13
Feat refactor tasks only using platform library

lohanidamodar:
## What does this PR do?

Refactors tasks in Appwrite using the new Platform library

## Test Plan

Existing tests

YES


src/Appwrite/CLI/Tasks.php
```diff
@@ -0,0 +1,13 @@
...
+
+use Utopia\Platform\Platform;
+
+class Tasks extends Platform
+{
+    public function __construct()
+    {
+        $this->addService('cliTasks', new TasksService());
+    }
+}
```
eldadfux, line 13: Feels to me like the platform should be named `Appwrite` and `Appwrite` should have a `tasks` service, not `cliTasks`. WDYT?

eldadfux: This will be a placeholder for other services.

src/Appwrite/CLI/Tasks/Specs.php
```diff
@@ -1,12 +1,14 @@
 <?php
 
-global $cli;
+namespace Appwrite\CLI\Tasks;
```
eldadfux, line 3:
```suggestion
namespace Appwrite\Appwrite\Tasks;
```

Make sense?

eldadfux: Or maybe `Appwrite\Platform\Tasks`

src/Appwrite/CLI/Tasks/Usage.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        $this
+            ->desc('Schedules syncing data from influxdb to Appwrite console db')
+            ->param('type', 'timeseries', new WhiteList(['timeseries', 'database']))
+            ->inject('dbForConsole')
+            ->inject('influxdb')
+            ->inject('logError')
+            ->callback(fn ($type, $dbForConsole, $influxDB, $logError) => $this->action($type, $dbForConsole, $influxDB, $logError));
+    }
+
```
eldadfux, line 34:
```suggestion
```

Extra line break.

src/Appwrite/CLI/TasksService.php
```diff
@@ -0,0 +1,36 @@
...
+use Appwrite\CLI\Tasks\Version;
+use VolumeSync;
+
+class TasksService extends Service
+{
+    public function __construct()
+    {
+        $this->type = self::TYPE_CLI;
+        $this
+            ->addAction(Version::getName(), new Version())
```
eldadfux, line 24: Rethinking about this. Do we still need the name if it's part of the interface? Can't we just take it inside the constructor?

eldadfux: If we do that. I think we can change getName to not be a static class.

src/Appwrite/CLI/TasksService.php
```diff
@@ -0,0 +1,36 @@
...
+use Appwrite\CLI\Tasks\Install;
+use Appwrite\CLI\Tasks\Maintenance;
+use Appwrite\CLI\Tasks\Migrate;
+use Appwrite\CLI\Tasks\SDKs;
+use Appwrite\CLI\Tasks\Specs;
+use Appwrite\CLI\Tasks\SSL;
+use Appwrite\CLI\Tasks\Usage;
+use Appwrite\CLI\Tasks\Vars;
+use Appwrite\CLI\Tasks\Version;
+use VolumeSync;
```
eldadfux, line 16: What is `VolumeSync`?

eldadfux (commented):
I think we should change the hierarchy of files to better support future services and platforms. Right now it seems to focused on the CLI and also create some confusion between CLI and Tasks

src/Appwrite/CLI/Tasks.php
lohanidamodar: renamed platform to Appwrite and service to Tasks and moved both to new namespace `Appwrite\Platform`

src/Appwrite/CLI/Tasks/Specs.php
lohanidamodar: Moved to `Appwrite\Platform\Tasks`

src/Appwrite/CLI/TasksService.php
lohanidamodar: updated the missing namespacing

src/Appwrite/CLI/TasksService.php
lohanidamodar: what do we do? move getName as instance function? that will have to be done in the `Action` and we need the name in order to reference the action, later to remove action?
Not sure I fully understand this

eldadfux (approved):

christyjacob4 (approved):

PR #4673, opened 2022-11-15
Feat realtime dbpools

abnegate:
## What does this PR do?

- Add `_APP_DB_MAX_CONNECTIONS` env var
- Add a `getWorkerPoolSize` function: allows calculating an appropriate pool size based on the number of workers and number of max connections as defined in env. Throws an exception if the pool size is smaller than the number of workers (can not satisfy a pool size of 1).
- Use `getWorkerPoolSize` for realtime server to set database pool size
- Update Traefik 2.7 -> 2.9
- Increase Traefik max established sockets (with an ACK) waiting to be accepted by listening process (`net.core.somaxconn`)
- Increase Traefik max open file limits (`ulimit.nofile.(hard|soft)`)
- Increase Traefik per-socket read/write buffer size (`net.ipv4.tcp_(w|r)mem`)
- Increase Traefik max outgoing port limit (`net.ipv4.ip_local_port_range`)
- Pass through `_APP_DB_MAX_CONNECTIONS` to local database container

## Test Plan

Existing tests + manual benchmark tests

## Related PRs and Issues

https://github.com/utopia-php/registry/pull/12
https://github.com/utopia-php/pools/pull/10

Yes

Yes


app/init.php
```diff
@@ -1122,6 +1123,26 @@ function getDevice($root): Device
...
+ */
+function getWorkerPoolSize(): int
+{
+    $reservedConnections = APP_DEFAULT_POOL_SIZE; // Pool of default size is reserved for the HTTP API
+    $workerCount = swoole_cpu_num() * intval(App::getEnv('_APP_WORKER_PER_CORE', 6));
+    $maxConnections = App::getenv('_APP_DB_MAX_CONNECTIONS', 251);
+    $workerConnections = $maxConnections - $reservedConnections;
+
+    if ($workerCount > $workerConnections) {
+        throw new \Exception('Worker pool size is too small. Increase the number of allowed database connections or decrease the number of workers.');
```
christyjacob4, line 1140: Can we use the 500 error code for this exception ?

christyjacob4 (approved):
Looks great 🆗

christyjacob4 (approved):

.env
```diff
@@ -23,6 +23,7 @@ _APP_DB_SCHEMA=appwrite
 _APP_DB_USER=user
 _APP_DB_PASS=password
 _APP_DB_ROOT_PASS=rootsecretpassword
+_APP_DB_MAX_CONNECTIONS=251
```
eldadfux, line 26: We should stop using _APP_DB as a prefix for env vars. We aim to eventually move to `_APP_CONNECTIONS` for all connection related stuff.

app/config/variables.php
```diff
@@ -306,24 +306,33 @@
...
+                'name' => '_APP_DB_MAX_CONNECTIONS',
+                'description' => 'MariaDB server maximum connections.',
+                'introduction' => 'TBD',
+                'default' => 251,
+                'required' => false,
+                'question' => '',
+                'filter' => ''
+            ],
+             [
+                 'name' => '_APP_CONNECTIONS_DB_PROJECT',
```
eldadfux, line 319: This was hidden on intent, please hide this vars.

app/init.php
```diff
@@ -1122,6 +1123,26 @@ function getDevice($root): Device
     }
 }
 
+/**
+ * Get database connection pool size for worker processes.
+ *
+ * @return int
+ * @throws \Exception
+ */
+function getWorkerPoolSize(): int
```
eldadfux, line 1132: Can't we just have the same logic for both and avoid the extra complexity at least for now? Also, we should try our best to avoid adding functions like this to the global scope.

app/realtime.php
```diff
@@ -55,7 +56,8 @@ function getProjectDB(Document $project): Database
 {
     global $register;
 
-    $pools = $register->get('pools'); /** @var \Utopia\Pools\Group $pools */
+    /** @var \Utopia\Pools\Group $pools */
+    $pools = $register->get('pools', args: [getWorkerPoolSize()]);
```
eldadfux, line 60: How was this merged if we decide to not apply this feature to the registry?

PR #4685, opened 2022-11-16
fix: github images

TorstenDittmann:
## What does this PR do?

- adds github images

## Test Plan
*n/a*

## Related PRs and Issues
*n/a*

*n/a*

✅


eldadfux (approved):
We should have the logo without the io
Also, I don't think we should have deleted all the phtml files. Some are used by the homepage and for OAuth redirects.

TorstenDittmann:
> We should have the logo without the io Also, I don't think we should have deleted all the phtml files. Some are used by the homepage and for OAuth redirects.

OAuth is handled by the console itself. If we need the old `*.phtml` files - we can go back to a previous tag.

Issue #4714, opened 2022-11-17
🐛 Bug Report: can’t create oauth sessions on native apps


netapy:
### 👟 Reproduction steps

When I try to open an oauth session by calling  `await account.createAuth2Session (provider: provider);` on my native mobile app (both android and iOS), I see a white screen and the url is : 
`https://example.com/v1/auth/oauth2/success?project=[PROJECT
ID]&domain=[DOMAIN]&key=[KEY]&secret=[SECRET]#` 

### 👍 Expected behavior

It should close the window and open the session in my app.

### 👎 Actual Behavior

The browser window is white and the session isn’t opened on my app.

### 🎲 Appwrite version

Version 1.1.x

### 💻 Operating system

Linux

### 🧱 Your Environment

My client runs the Flutter SDK.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

stnguyen90:
Thanks for raising this issue! 🙏🏼 I think the problem might be because we implemented the route as

```
https://example.com/auth/oauth/success
```

rather than

```
https://example.com/v1/auth/oauth2/success
```

netapy:
@stnguyen90  thanks for handling this.

Can't we add this to the 1.1.1 ? I would really need this as fast as possible and unfortunately I can't downgrade to 1.0.3 because it causes some database issues (unless you know how to solve these?).
Cheers

stnguyen90:
@netapy we can't add it to 1.1.1 because 1.1.1 was already released 😕

netapy:
@stnguyen90 Oh right !
I saw your pull request, any idea how long it'll take to implement the fix ? Sorry for insisting but it's really important for me.

stnguyen90:
@netapy, hopefully in the next couple of days.

Shiba-Kar:
Facing the same issue in Flutter

sampathkumar7:
I'm facing same issue

sampathkumar7:
anyway i can change that url manually, i'm using digitalocean onclick droplet

netapy:
Hi, I believe the PR for this issue has been accepted (https://github.com/appwrite/appwrite/pull/4725).
[...]

PR #4725, opened 2022-11-18
Fix the default oauth paths to match the route

stnguyen90:
## What does this PR do?

Previously, Utopia handled `/auth/oauth2/success` and `/v1/auth/oauth2/success` exactly the same (with the same handler). So on mobile, when users were redirected `$oauthDefaultSuccess` (`/v1/auth/oauth2/success`), it was still handled. After we deleted that route and [moved the code to the new console](https://github.com/appwrite/console/commit/2171d7f22d27b3a9ee99693e6b71b4a445a4eda3), the default `/v1/auth/oauth2/success` resulted in a 404. This changes the default path to remove the `/v1/` prefix so that the console can handle the redirect.

## Test Plan

Manual

## Related PRs and Issues

- #4714 

Yes

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -41,8 +41,8 @@ use Utopia\Validator\Assoc;
 use Utopia\Validator\Text;
 use Utopia\Validator\WhiteList;
 
-$oauthDefaultSuccess = '/v1/auth/oauth2/success';
-$oauthDefaultFailure = '/v1/auth/oauth2/failure';
+$oauthDefaultSuccess = '/auth/oauth2/success';
+$oauthDefaultFailure = '/auth/oauth2/failure';
 
 App::post('/v1/account')
     ->desc('Create Account')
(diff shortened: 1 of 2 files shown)


app/controllers/api/account.php
```diff
@@ -41,8 +41,8 @@
 use Utopia\Validator\Text;
 use Utopia\Validator\WhiteList;
 
-$oauthDefaultSuccess = '/v1/auth/oauth2/success';
-$oauthDefaultFailure = '/v1/auth/oauth2/failure';
+$oauthDefaultSuccess = '/auth/oauth2/success';
```
stnguyen90, line 44: This is actually `/auth/oauth/success` in the [Console](https://github.com/appwrite/console/tree/2.0.1/src/routes/auth/oauth/success). @ArmanNik, @TorstenDittmann, should we change it in the console code or here?

stnguyen90: Here's the PR if we want to go down the route of updating the path in the Console: https://github.com/appwrite/console/pull/170

ArmanNik (approved):

PR #4831, opened 2022-12-09
feat (projects): auth session limit

lohanidamodar:
Add endpoint on project to set max number of sessions for project users.

## What does this PR do?

- introduces feature to set max session limits per project

### UI
image

## Test Plan

- New test added to `ProjectConsoleClientTest`

## Related PRs and Issues

- UI - https://github.com/appwrite/console/pull/215

YES


app/init.php
```diff
@@ -84,6 +84,7 @@
 const APP_PAGING_LIMIT = 12;
 const APP_LIMIT_COUNT = 5000;
 const APP_LIMIT_USERS = 10000;
+const APP_LIMIT_USER_SESSIONS = 100;
```
eldadfux, line 87: I feel we should aim lower. Maybe 20.

app/controllers/shared/api.php
```diff
@@ -318,6 +318,39 @@
         }
     });
 
+App::shutdown()
+    ->groups(['session'])
```
eldadfux, line 322: Can we add a description to each hook explaining its responsibility?

app/controllers/shared/api.php
```diff
@@ -318,6 +318,39 @@
...
+    ->inject('utopia')
+    ->inject('request')
+    ->inject('response')
+    ->inject('project')
+    ->inject('dbForProject')
+    ->action(function (App $utopia, Request $request, Response $response, Document $project, Database $dbForProject) {
+        $sessionLimit = $project->getAttribute('auths', [])['maxSessions'] ?? APP_LIMIT_USER_SESSIONS;
+        $session = $response->getPayload();
```
eldadfux, line 330: Does this works well for OAuth sessions as well?

src/Appwrite/Utopia/Response/Model/Project.php
```diff
@@ -114,6 +114,12 @@ public function __construct()
                 'default' => 0,
                 'example' => 100,
             ])
+            ->addRule('authSessionsLimit', [
+                'type' => self::TYPE_INTEGER,
+                'description' => 'Max sessions allowed per user. 100 maximum.',
+                'default' => 100,
```
eldadfux, line 120:
```suggestion
                'default' => 100,
```

Lower...

eldadfux (changes requested):

app/init.php
lohanidamodar: This is the maximum ever allowed? Are we sure to keep it low as 20?, default is 10 hardcoded

app/controllers/shared/api.php
lohanidamodar: it should as I added the group in the endpoint that actually creates the session object

app/init.php
lohanidamodar: this is also used the in the range validation of the endpoint `1-100` that's the max allowed and default. Do we keep separate constant for max allowed value and default? or make it lower as 10-20 ?

app/controllers/shared/api.php
lohanidamodar: I'll add to this one, but others are handling so many things

eldadfux: Agree. Eventually I think we should start splitting them to have one responsibility.

src/Appwrite/Utopia/Response/Model/Project.php
eldadfux: Let's change the default to 10 as well

eldadfux (approved):

PR #4860, opened 2022-12-14
Fix: expired session, schedule deletes

Meldiron:
## What does this PR do?

Fixes:
- delete worker deleting schedules for existing functions
- expired session deletion having syntax error after merge conflict
- Prepared patch CLI script for missing schedule documents

## Test Plan

- [x] Manual QA

## Related PRs and Issues

x

Not user facing chage

Yes


src/Appwrite/Platform/Services/Tasks.php
```diff
@@ -29,6 +30,7 @@ public function __construct()
             ->addAction(Doctor::getName(), new Doctor())
             ->addAction(Install::getName(), new Install())
             ->addAction(Maintenance::getName(), new Maintenance())
+            ->addAction(CreateMissingSchedules::getName(), new CreateMissingSchedules())
```
eldadfux, line 33: Let add a patch prefix for this command so we know how to treat it and remove it when times comes.

docker-compose.yml
```diff
@@ -33,15 +33,6 @@ services:
       - 8080:80
       - 443:443
       - 9500:8080
-    ulimits:
```
eldadfux, line 36: Why are we removing this as part of this PR?

eldadfux (changes requested):

docker-compose.yml
Meldiron: It's something that is not supposed to be here anyway.

src/Appwrite/Platform/Services/Tasks.php
Meldiron: Renamed to `PatchCreateMissingSchedules`.

docker-compose.yml
eldadfux: Can you explain? Where should it be in that case?

eldadfux (approved):

docker-compose.yml
Meldiron: It breaks Gitpod workspaces.

PR #4866, opened 2022-12-16
feat: check password history

lohanidamodar:
## What does this PR do?

- adds password history setting for project auth, allows saving the history of passwords

## Test Plan

- New test on ProjectsConsoleClientTest

## Related PRs and Issues

YES

YES


app/controllers/api/account.php
```diff
@@ -109,7 +112,8 @@
                 'email' => $email,
                 'emailVerification' => false,
                 'status' => true,
-                'password' => Auth::passwordHash($password, Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS),
+                'passwordHistory' => $passwordHistory > 0 ? [$password] : [],
```
eldadfux, line 115: Let's group all password related attributes together in all controllers.

app/controllers/api/projects.php
```diff
@@ -576,6 +576,37 @@
...
+    ->groups(['api', 'projects'])
+    ->label('scope', 'projects.write')
+    ->label('sdk.auth', [APP_AUTH_TYPE_ADMIN])
+    ->label('sdk.namespace', 'projects')
+    ->label('sdk.method', 'updateAuthLimit')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_PROJECT)
+    ->param('projectId', '', new UID(), 'Project unique ID.')
+    ->param('limit', false, new Range(0, APP_LIMIT_USER_SESSIONS_MAX), 'Set the max number of password history to keep for users. Value allowed is between 0-' . APP_LIMIT_USER_PASSWORD_HISTORY . '. Default is 0')
```
eldadfux, line 590: We should create a dedicated constant for passwords.

eldadfux, line 590:
```suggestion
    ->param('limit', false, new Range(0, APP_LIMIT_USER_SESSIONS_MAX), 'Set the max number of password to store in user history. User can't choose a new password that is already stored in the password history list.  Max number of password allowed in history is' . APP_LIMIT_USER_PASSWORD_HISTORY . '. Default value is 0')
```

eldadfux, line 590: Default value should be 0 not false.

app/controllers/api/teams.php
```diff
@@ -358,6 +360,7 @@
                     'email' => $email,
                     'emailVerification' => false,
                     'status' => true,
+                    'passwordHistory' => $passwordHistory > 0 ? [Auth::passwordHash(Auth::passwordGenerator(), Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS)] : [],
```
eldadfux, line 363: Because this is a random password, maybe we should not store it in the history list?

app/controllers/api/users.php
```diff
@@ -781,18 +792,39 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
-    ->action(function (string $userId, string $password, Response $response, Database $dbForProject, Event $events) {
+    ->action(function (string $userId, string $password, Response $response, Document $project, Database $dbForProject, Event $events) {
 
         $user = $dbForProject->getDocument('users', $userId);
 
         if ($user->isEmpty()) {
             throw new Exception(Exception::USER_NOT_FOUND);
         }
 
+        $newPassword = Auth::passwordHash($password, Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS);
```
eldadfux, line 806: It will make more sense to abstract all this logic in a validator. We should try to extend the current password validator and let the framework input validation take care of this or if not possible create a new validator and check inside the controller. Option 1 seems better if possible.

app/controllers/api/users.php
```diff
@@ -781,18 +792,39 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+            foreach ($history as $hash) {
+                if (Auth::passwordVerify($password, $hash, $user->getAttribute('hash'), $user->getAttribute('hashOptions'))) {
+                    throw new Exception(Exception::USER_PASSWORD_RECENTLY_USED, 'The password was recently used', 409);
+                }
+            }
+
+            $history[] = $newPassword;
+            while (count($history) > $historyLimit) {
+                array_pop($history);
+            }
```
eldadfux, line 822: Array slice would work better here: https://www.php.net/array_slice

eldadfux (changes requested):

app/controllers/api/projects.php
lohanidamodar: ah I do, used the wrong one.

app/controllers/api/teams.php
lohanidamodar: Okay

src/Appwrite/Utopia/Response/Model/Project.php
```diff
@@ -120,6 +120,12 @@ public function __construct()
                 'default' => 10,
                 'example' => 10,
             ])
+            ->addRule('authPasswordHistory', [
+                'type' => self::TYPE_INTEGER,
+                'description' => 'Max password history to save per user. 20 maximum. 0 for ignoring the password history.',
```
eldadfux, line 125:
```suggestion
                'description' => 'Max password history to save per user. Max passwords limit allowed in history is 20. Use 0 for disabling password history.',
```

app/controllers/api/projects.php
lohanidamodar: resolved.

lohanidamodar: resolved.

app/controllers/api/teams.php
lohanidamodar: done.

app/controllers/api/users.php
lohanidamodar: done.

app/controllers/api/users.php
lohanidamodar: updated.

app/controllers/api/account.php
```diff
@@ -1507,17 +1515,36 @@
...
             throw new Exception(Exception::USER_INVALID_CREDENTIALS);
         }
 
+        $newPassword = Auth::passwordHash($password, Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS);
+
+        $historyLimit = $project->getAttribute('auths', [])['passwordHistory'] ?? 0;
+        $history = [];
+        if ($historyLimit > 0) {
+            $history = $user->getAttribute('passwordHistory', []);
+            $validator = new PasswordHistory($history, $user->getAttribute('hash'), $user->getAttribute('hashOptions'));
```
eldadfux, line 1534: I think it will be better to just pass the hashed string instead of hashing it inside the validator to create less dependency between them.

app/controllers/api/account.php
```diff
@@ -1507,17 +1515,36 @@
...
+            if (!$validator->isValid($password)) {
+                throw new Exception(Exception::USER_PASSWORD_RECENTLY_USED, 'The password was recently used', 409);
+            }
+
+            $history[] = $newPassword;
+            if (count($history) > $historyLimit) {
```
eldadfux, line 1540: Do we need this condition? Won't it work the same without it?

app/controllers/api/account.php
```diff
@@ -1507,17 +1515,36 @@
...
+                array_slice($history, (count($history) - $historyLimit), $historyLimit);
+            }
+        }
+
         $user = $dbForProject->updateDocument('users', $user->getId(), $user
-                ->setAttribute('password', Auth::passwordHash($password, Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS))
+                ->setAttribute('a', $history)
```
eldadfux, line 1546: I think this is an error. Let's group attributes here as well.

app/controllers/api/projects.php
```diff
@@ -576,6 +576,37 @@
...
+    ->groups(['api', 'projects'])
+    ->label('scope', 'projects.write')
+    ->label('sdk.auth', [APP_AUTH_TYPE_ADMIN])
+    ->label('sdk.namespace', 'projects')
+    ->label('sdk.method', 'updateAuthLimit')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_PROJECT)
+    ->param('projectId', '', new UID(), 'Project unique ID.')
+    ->param('limit', false, new Range(0, APP_LIMIT_USER_SESSIONS_MAX), 'Set the max number of password to store in user history. User can\'t choose a new password that is already stored in the password history list.  Max number of password allowed in history is' . APP_LIMIT_USER_PASSWORD_HISTORY . '. Default value is 0')
```
eldadfux, line 590:
```suggestion
    ->param('limit', 0, new Range(0, APP_LIMIT_USER_SESSIONS_MAX), 'Set the max number of passwords to store in user history. User can\'t choose a new password that is already stored in the password history list.  Max number of passwords allowed in history is' . APP_LIMIT_USER_PASSWORD_HISTORY . '. Default value is 0')
```

eldadfux: We also still use the wrong constant.

app/controllers/api/users.php
```diff
@@ -781,18 +793,37 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+
+            $history[] = $newPassword;
+            if (count($history) > $historyLimit) {
+                array_slice($history, (count($history) - $historyLimit), $historyLimit);
+            }
+        }
+
         $user
-            ->setAttribute('password', Auth::passwordHash($password, Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS))
+            ->setAttribute('passwordHistory', $history)
```
eldadfux, line 825: Let's group attributes here.

src/Appwrite/Auth/Validator/PasswordHistory.php
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * Is valid.
+     *
+     * @param mixed $value
+     *
+     * @return bool
+     */
+    public function isValid($value): bool
+    {
+        foreach ($this->history as $hash) {
```
eldadfux, line 44: Why do we validate the password here? Don't we just need to check if the new hash in the history list of hashes?

src/Appwrite/Utopia/Response/Model/Project.php
```diff
@@ -120,6 +120,12 @@ public function __construct()
                 'default' => 10,
                 'example' => 10,
             ])
+            ->addRule('authPasswordHistory', [
+                'type' => self::TYPE_INTEGER,
+                'description' => 'Max password history to save per user. Max passwords limit allowed in history is 20. Use 0 for disabling password history.',
```
eldadfux, line 125:
```suggestion
                'description' => 'Max allowed passwords in the history list per user. Max passwords limit allowed in history is 20. Use 0 for disabling password history.',
```

eldadfux (changes requested):

app/controllers/api/account.php
lohanidamodar: I tried checking string against hashed password, it didn't work, I had to use Auth::passwordVerify which needs these options
https://github.com/appwrite/appwrite/blob/28ac6c93555fbf53ceb48523e53a431d4a4913d5/src/Appwrite/Auth/Auth.php#L216

app/controllers/api/account.php
lohanidamodar: It should right.

app/controllers/api/account.php
lohanidamodar: fixed

app/controllers/api/account.php
lohanidamodar: updated.

app/controllers/api/users.php
lohanidamodar: it is grouped. 
https://github.com/appwrite/appwrite/blob/28ac6c93555fbf53ceb48523e53a431d4a4913d5/app/controllers/api/users.php#L824

src/Appwrite/Auth/Validator/PasswordHistory.php
lohanidamodar: tried using `in_array` didn't work. I had to use `Auth::passwordVerify` not sure why.

CHANGES.md
```diff
@@ -1,3 +1,8 @@
+## Version 1.3.0
+
+## Features
+- Password history setting allows to save user's last used password so that it may not be used again. Maximum number of history saved is 20, which can be configured [#4866](https://github.com/appwrite/appwrite/pull/4866)
```
Meldiron, line 4: No need to explain minimum, I think. Tho if we do, let's also mention default value.

src/Appwrite/Migration/Version/V18.php
```diff
@@ -0,0 +1,103 @@
...
+            switch ($id) {
+                case 'users':
+                    try {
+                        /**
+                         * Create 'passwordHistory' attribute
+                         */
+                        $this->createAttributeFromCollection($this->projectDB, $id, 'passwordHistory');
+                        $this->projectDB->deleteCachedCollection($id);
+                    } catch (\Throwable $th) {
+                        Console::warning("'region' from {$id}: {$th->getMessage()}");
```
Meldiron, line 59:
```suggestion
                        Console::warning("'passwordHistory' from {$id}: {$th->getMessage()}");
```

app/controllers/api/account.php
```diff
@@ -501,6 +508,7 @@
                         'email' => $email,
                         'emailVerification' => true,
                         'status' => true, // Email should already be authenticated by OAuth2 provider
+                        'passwordHistory' => $passwordHistory > 0 ? [Auth::passwordHash(Auth::passwordGenerator(), Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS)] : null,
```
Meldiron, line 511: Let's make a variable to prevent code diplication (passwordHistory, password)

CHANGES.md
lohanidamodar: default is 0 which mean it's not enabled, do we need to mention that?

app/controllers/api/account.php
lohanidamodar: updated

CHANGES.md
eldadfux: I think we can, just in case we expose this API in the future.

app/controllers/api/account.php
lohanidamodar: done in dictionary PR

CHANGES.md
lohanidamodar: updated

eldadfux (approved):

PR #4898, opened 2022-12-22
add internalId to collections

shimonewman:
added internalId to collections that have relations and missing  internalId column


tests/unit/Docker/EnvTest.php
```diff
@@ -37,6 +37,7 @@ public function testExport(): void
         $this->assertEquals("_APP_X=value1
 _APP_Y=value2
 _APP_Z=value3
+_APP_W=value5=
```
lohanidamodar, line 40: What is this change?

lohanidamodar (approved):
Looks good. Just did not understand the change with Docker/EnvTest.php

tests/unit/Docker/EnvTest.php
shimonewman: This env variable (_APP_W = value5=)   was added lately  to the resources/docker/.env
inside the test, there is a check against these values.
I can not see how it will pass without this value being added to the test.

PR #4906, opened 2022-12-26
password dictionary

lohanidamodar:
## What does this PR do?

- check user's new password against common passwords dictionary

## Test Plan

## Related PRs and Issues

(Write your answer here.)


app/init.php
```diff
@@ -605,6 +605,9 @@ function (mixed $value) {
 $register->set('geodb', function () {
     return new Reader(__DIR__ . '/assets/dbip/dbip-country-lite-2022-06.mmdb');
 });
+$register->set('passwordsDB', function () {
+    return \file_get_contents(__DIR__ . '/assets/security/10k-common-passwords');
```
eldadfux, line 609: It will be much more efficient to parse the list here as a hashmap with the password used as the key, and then using array_key_exists for very fast validation on this specific data structure.

app/init.php
```diff
@@ -1022,6 +1025,11 @@ function getDevice($root): Device
     return $register->get('geodb');
 }, ['register']);
 
+App::setResource('passwordsDB', function ($register) {
```
eldadfux, line 1028:
```suggestion
App::setResource('passwordsDictionary', function ($register) {
```

Let's introduce less new terms

eldadfux: Same in other places.

app/controllers/api/users.php
```diff
@@ -806,6 +818,15 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
 
         $newPassword = Auth::passwordHash($password, Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS);
 
+        $passwordDictionary = $project->getAttribute('auths', [])['passwordDictionary'] ?? false;
```
eldadfux, line 821: Can we abstract the logic behind the framework validator? We can pass both the project document and dictionary into the validator is DI and using a callback.

eldadfux (changes requested):

src/Appwrite/Utopia/Response/Model/Project.php
```diff
@@ -247,6 +253,7 @@ public function filter(Document $document): Document
         $document->setAttribute('authDuration', $authValues['duration'] ?? Auth::TOKEN_EXPIRATION_LOGIN_LONG);
         $document->setAttribute('authSessionLimit', $authValues['maxSessions'] ?? APP_LIMIT_USER_SESSIONS_DEFAULT);
         $document->setAttribute('authPasswordHistory', $authValues['passwordHistory'] ?? 0);
+        $document->setAttribute('authPasswordDictionary', $authValues['passwordDictionary'] ?? false);
```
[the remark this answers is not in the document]
lohanidamodar: this is dictionary, `enable` vs `disable` in project setting

app/controllers/api/users.php
lohanidamodar: we can, but not in the parameter validation because, we will need the project 's `auths/passwordDictionary` value

lohanidamodar: we will have to do it like the password history validator

lohanidamodar: updated

app/init.php
lohanidamodar: updated

app/controllers/api/projects.php
```diff
@@ -607,6 +607,37 @@
         $response->dynamic($project, Response::MODEL_PROJECT);
     });
 
+App::patch('/v1/projects/:projectId/auth/password-dictionary')
+    ->desc('Update Project users limit')
```
Meldiron, line 611: Wrong description

app/controllers/api/projects.php
```diff
@@ -607,6 +607,37 @@
...
+    ->groups(['api', 'projects'])
+    ->label('scope', 'projects.write')
+    ->label('sdk.auth', [APP_AUTH_TYPE_ADMIN])
+    ->label('sdk.namespace', 'projects')
+    ->label('sdk.method', 'updateAuthLimit')
```
Meldiron, line 616: Wrong sdk method

app/controllers/api/projects.php
```diff
@@ -607,6 +607,37 @@
...
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_PROJECT)
+    ->param('projectId', '', new UID(), 'Project unique ID.')
+    ->param('enabled', false, new Boolean(true), 'Set whether or not to enable checking user\'s password against most commonly used passwords. Default is false.')
```
Meldiron, line 621: if `true` in validator means `loose`, we should not allow that. We should set loose=false. Loose is only needed for GET requests or file upload requests. With PATCH we can be strict about types

app/controllers/api/projects.php
```diff
@@ -607,6 +607,37 @@
...
+    ->action(function (string $projectId, bool $enabled, Response $response, Database $dbForConsole) {
+
+        $project = $dbForConsole->getDocument('projects', $projectId);
+
+        if ($project->isEmpty()) {
+            throw new Exception(Exception::PROJECT_NOT_FOUND);
+        }
+
+        $auths = $project->getAttribute('auths', []);
+        $auths['passwordDictionary'] = (bool) filter_var($enabled, FILTER_VALIDATE_BOOLEAN);
```
Meldiron, line 633: With loose=false, we dont need the convertion. Simple `= $enabled` will work fine.

app/init.php
```diff
@@ -607,6 +607,12 @@ function (mixed $value) {
 $register->set('geodb', function () {
     return new Reader(__DIR__ . '/assets/dbip/dbip-country-lite-2022-06.mmdb');
 });
+$register->set('passwordsDictionary', function () {
+    $content = \file_get_contents(__DIR__ . '/assets/security/10k-common-passwords');
+    $content = explode("\n", $content);
+    $content = array_flip($content);
```
Meldiron, line 613: Why do we flip array? Does it make faster?

app/controllers/api/projects.php
Meldiron, line 633: We are missing migration for this. No need for collection migration, but w eneed to update each project document and add default value for passwordDictionary

src/Appwrite/Extend/Exception.php
```diff
@@ -66,6 +66,7 @@ class Exception extends \Exception
     public const USER_SESSION_ALREADY_EXISTS       = 'user_session_already_exists';
     public const USER_NOT_FOUND                    = 'user_not_found';
     public const USER_PASSWORD_RECENTLY_USED       = 'password_recently_used';
+    public const USER_PASSWORD_IN_DICTIONARY       = 'password_in_dictionary';
```
Meldiron, line 69: I cant see where we use this new excepiton

Meldiron (changes requested):

app/controllers/api/projects.php
lohanidamodar: default is false, which we make sure to test for null when reading so it will work without migration. You think we need to?

app/controllers/api/projects.php
lohanidamodar: updated

app/controllers/api/projects.php
lohanidamodar: updated

app/controllers/api/projects.php
lohanidamodar: updated

app/controllers/api/projects.php
lohanidamodar: updated

src/Appwrite/Extend/Exception.php
lohanidamodar: removed

app/controllers/api/projects.php
Meldiron: Not sure what our strategy here is, let's maybe confirm with Eldad.
My call would be to have migration. In future someone might want to use the attribute but won't know there **has to** be default specified in `getAttribute`, and it can cause a bug.
Having migration ensures the attribute is always set, so default can be set on database-level, not in the controller code.

lohanidamodar: Updated migration

src/Appwrite/Auth/Validator/PasswordDictionary.php
```diff
@@ -0,0 +1,78 @@
...
+ * Password.
+ *
+ * Validates user password string
+ */
+class PasswordDictionary extends Password
+{
+    protected array $dictionary;
+    protected Document $project;
+
+    public function __construct(array $dictionary, Document $project)
```
eldadfux, line 17: We should always strive to simplify our signatures and use primitives when can. In this example, we can use an array of strings.

eldadfux (changes requested):

src/Appwrite/Auth/Validator/PasswordDictionary.php
lohanidamodar: I thought array_key_exists was more performant than in_array, so I used array_flip to get strings into keys. 10k strings are a lot.

eldadfux: I was referring to the 2nd argument, the project document. Can we only get the relevant array from it?

lohanidamodar: updated

eldadfux (approved):

PR #4973, opened 2023-01-09
Team Invite Improvements

lohanidamodar:
## What does this PR do?

- Supports UserID, email and Phone in team invite

## Test Plan

## Related PRs and Issues

YES

### DIFF
--- a/app/controllers/api/teams.php
+++ b/app/controllers/api/teams.php
@@ -1,6 +1,7 @@
 <?php
 
 use Appwrite\Auth\Auth;
+use Appwrite\Auth\Validator\Phone;
 use Appwrite\Detector\Detector;
 use Appwrite\Event\Delete;
 use Appwrite\Event\Event;
@@ -39,6 +40,7 @@ use Utopia\Validator\Text;
 use Utopia\Validator\Range;
 use Utopia\Validator\ArrayList;
 use Utopia\Validator\WhiteList;
+use Appwrite\Event\Phone as EventPhone;
 
 App::post('/v1/teams')
     ->desc('Create Team')
@@ -304,7 +306,9 @@ App::post('/v1/teams/:teamId/memberships')
     ->label('sdk.response.model', Response::MODEL_MEMBERSHIP)
     ->label('abuse-limit', 10)
     ->param('teamId', '', new UID(), 'Team ID.')
-    ->param('email', '', new Email(), 'Email of the new team member.')
+    ->param('userId', '', new UID(), 'User ID.', true)
+    ->param('email', '', new Email(), 'Email of the new team member.', true)
+    ->param('phone', '', new Phone(), 'Phone number. Format this number with a leading \'+\' and a country code, e.g., +16175551212.', true)
     ->param('roles', [], new ArrayList(new Key(), APP_LIMIT_ARRAY_PARAMS_SIZE), 'Array of strings. Use this param to set the user roles in the team. A role can be any string. Learn more about [roles and permissions](/docs/permissions). Maximum of ' . APP_LIMIT_ARRAY_PARAMS_SIZE . ' roles are allowed, each 32 characters long.')
     ->param('url', '', fn($clients) => new Host($clients), 'URL to redirect the user back to your app from the invitation email.  Only URLs from hostnames in your project platform list are allowed. This requirement helps to prevent an [open redirect](https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html) attack against your project API.', false, ['clients']) // TODO add our own built-in confirm page
     ->param('name', '', new Text(128), 'Name of the new team member. Max length: 128 chars.', true)
@@ -314,9 +318,13 @@ App::post('/v1/teams/:teamId/memberships')
     ->inject('dbForProject')
     ->inject('locale')
     ->inject('mails')
+    ->inject('messaging')
     ->inject('events')
-    ->action(function (string $teamId, string $email, array $roles, string $url, string $name, Response $response, Document $project, Document $user, Database $dbForProject, Locale $locale, Mail $mails, Event $events) {
+    ->action(function (string $teamId, string $userId, string $email, string $phone, array $roles, string $url, string $name, Response $response, Document $project, Document $user, Database $dbForProject, Locale $locale, Mail $mails, EventPhone $messaging, Event $events) {
 
+        if(empty($userId) && empty($email) && empty($phone)) {
+            throw new Exception(Exception::GENERAL_ARGUMENT_INVALID, 'At least one of userId, email, or phone is required');
+        }
         $isPrivilegedUser = Auth::isPrivilegedUser(Authorization::getRoles());
         $isAppUser = Auth::isAppUser(Authorization::getRoles());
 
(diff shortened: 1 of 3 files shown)


app/controllers/api/teams.php
```diff
@@ -304,7 +306,9 @@
     ->label('sdk.response.model', Response::MODEL_MEMBERSHIP)
     ->label('abuse-limit', 10)
     ->param('teamId', '', new UID(), 'Team ID.')
-    ->param('email', '', new Email(), 'Email of the new team member.')
+    ->param('userId', '', new UID(), 'User ID.', true)
```
Meldiron, line 309: Let's keep `email` first to prevent breaking changes on SDK side.

app/controllers/api/teams.php
```diff
@@ -314,9 +318,13 @@
     ->inject('dbForProject')
     ->inject('locale')
     ->inject('mails')
+    ->inject('messaging')
     ->inject('events')
-    ->action(function (string $teamId, string $email, array $roles, string $url, string $name, Response $response, Document $project, Document $user, Database $dbForProject, Locale $locale, Mail $mails, Event $events) {
+    ->action(function (string $teamId, string $userId, string $email, string $phone, array $roles, string $url, string $name, Response $response, Document $project, Document $user, Database $dbForProject, Locale $locale, Mail $mails, EventPhone $messaging, Event $events) {
 
+        if (empty($userId) && empty($email) && empty($phone)) {
```
Meldiron, line 325: I think we should make this `exactly one`. If you pass all 3, how does it behave?

Or if it's on purpose for creating, we need to make documentation around what is the priority.

app/controllers/api/teams.php
```diff
@@ -434,16 +464,25 @@
...
+                    ->setUrl($url)
+                    ->setName($name)
+                    ->setLocale($locale->default)
+                    ->setTeam($team)
+                    ->setUser($user)
+                    ->trigger()
+                ;
+            }
+
+            if (@empty($phone)) {
```
Meldiron, line 480: Is `@empty` on purpose?

app/controllers/api/teams.php
```diff
@@ -434,16 +464,25 @@
...
-                ->setType(MAIL_TYPE_INVITATION)
-                ->setRecipient($email)
-                ->setUrl($url)
-                ->setName($name)
-                ->setLocale($locale->default)
-                ->setTeam($team)
-                ->setUser($user)
-                ->trigger()
-            ;
+            if (!empty($email)) {
```
Meldiron, line 467: If both email and phone are provided, does it send both? Is that intended? I would again try to setup a priority and write documentation for it.

app/controllers/api/teams.php
```diff
@@ -332,7 +340,28 @@
...
+            }
+            if (!empty($phone) && $invitee->getAttribute('phone', '') != $phone) {
+                throw new Exception(Exception::USER_ALREADY_EXISTS, 'Given userId and phone doesn\'t match', 409);
+            }
+        } elseif (!empty($email)) {
+            $invitee = $dbForProject->findOne('users', [Query::equal('email', [$email])]); // Get user by email address
+            if (!$invitee->isEmpty() && !empty($phone) && $invitee->getAttribute('phone', '') != $phone) {
+                throw new Exception(Exception::USER_ALREADY_EXISTS, 'Given email and phone doesn\'t match', 409);
+            }
+        } elseif (!empty($phone)) {
```
Meldiron, line 359: This seems to ignore `userId`. How would it behave if all 3 information are provided? Is it considered expected behaviour?

app/controllers/api/teams.php
lohanidamodar: noe, supposed to be !empty 😄

app/controllers/api/teams.php
```diff
@@ -356,6 +385,7 @@
                         Permission::delete(Role::user($userId)),
                     ],
                     'email' => $email,
+                    'phone' => $phone,
```
Meldiron, line 388: Above this we define `$userId = ID::unique();`. Should we take it from parameters instead, if provided?

app/controllers/api/teams.php
lohanidamodar: we have a priority ID > email > phone for this purpose

app/controllers/api/teams.php
lohanidamodar: done!

app/controllers/api/teams.php
lohanidamodar: Priority is ID > email > Phone, however if both  is provided, I think it's logical to send on to both?

app/controllers/api/teams.php
lohanidamodar: ID is the first priority and if the ID is provided and user with ID doesn't exist we throw 404, so the case when phone exists and id also exists doesn't arise.

app/controllers/api/teams.php
lohanidamodar: No, if the ID is provided and user doesn't exist with that ID we throw 404 don't create new with the given ID

app/controllers/api/teams.php
eldadfux: Sending the message on both seems a bit annoying, maybe we can try to prioritize email?

Meldiron: Makes sense to me to prioritize email if both provided.
If they want to send SMS, they can invite without email and later add email (with event / function / ...)

Alternatively we could add parameter to provide array of strings where to send notificaton. `email` and `phone` would be valid options.
If we did that, in theory this endpoint could behave same for client and for server. In the end, currently server-SDK does not trigger email. With this parameter, server could decide if they want to send email or not.

lohanidamodar: fixed

Meldiron (approved):

eldadfux (commented):
Looks good, but we need to update the docs to better explain how this endpoint behaves now.

app/controllers/api/teams.php
```diff
@@ -304,7 +306,9 @@
     ->label('sdk.response.model', Response::MODEL_MEMBERSHIP)
     ->label('abuse-limit', 10)
     ->param('teamId', '', new UID(), 'Team ID.')
-    ->param('email', '', new Email(), 'Email of the new team member.')
+    ->param('email', '', new Email(), 'Email of the new team member.', true)
+    ->param('userId', '', new UID(), 'User ID.', true)
```
gewenyu99, line 310:
```suggestion
    ->param('userId', '', new UID(), 'User ID. Choose your own unique ID or pass the string `ID.unique()` to auto generate it. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.', true)
```

If they can generate a UID, let's keep the description consistent with other endpoints.

gewenyu99, line 310:
```suggestion
    ->param('userId', '', new UID(), 'Unique Id. Choose your own unique ID or generate a random ID with `ID.unique()`. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.', true)
```

gewenyu99 (dismissed):
This looks good. Made a suggestion to keep the UID description the same everywhere.

app/controllers/api/teams.php
lohanidamodar: @gewenyu99 we don't allow using new ID here, it should be an existing User ID so that we can add the existing user to the team.

lohanidamodar:
@gewenyu99 we need to update this docs as well
https://appwrite.io/docs/client/teams?sdk=web-default#teamsCreateMembership

eldadfux (approved):

karbonnanroid (approved):

PR #4974, opened 2023-01-09
Oauth2 check if provider is enabled

lohanidamodar:
## What does this PR do?

- check if provider is enabled for OAuth2

## Test Plan

## Related PRs and Issues

(Write your answer here.)

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -280,6 +280,12 @@ App::get('/v1/account/sessions/oauth2/:provider')
 
         $protocol = $request->getProtocol();
         $callback = $protocol . '://' . $request->getHostname() . '/v1/account/sessions/oauth2/callback/' . $provider . '/' . $project->getId();
+        $providerEnabled = $project->getAttribute('authProviders', [])[$provider.'Enabled'] ?? false;
+        
+        if (!$providerEnabled) {
+            throw new Exception(Exception::PROJECT_PROVIDER_DISABLED, 'This provider is disabled. Please enable the provider from your ' . APP_NAME . ' console to continue.');
+        }
+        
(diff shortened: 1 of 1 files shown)


app/controllers/api/account.php
```diff
@@ -280,6 +280,12 @@
 
         $protocol = $request->getProtocol();
         $callback = $protocol . '://' . $request->getHostname() . '/v1/account/sessions/oauth2/callback/' . $provider . '/' . $project->getId();
+        $providerEnabled = $project->getAttribute('authProviders', [])[$provider.'Enabled'] ?? false;
+        
+        if (!$providerEnabled) {
+            throw new Exception(Exception::PROJECT_PROVIDER_DISABLED, 'This provider is disabled. Please enable the provider from your ' . APP_NAME . ' console to continue.');
```
Meldiron, line 286: Do we need to specify message? Cant we use default from errors.php?
If so, did you copy this from somewhere? Should we fix this elsewhere too?

Meldiron (approved):

app/controllers/api/account.php
lohanidamodar: Copied from few lines below.

stnguyen90:
This is good for most cases, but we may also want the check in the callback as well go prevent a bad actor from, somehow, going directly to the oauth provider without using this API call.

lohanidamodar:
@stnguyen90 added a check in the redirect as well. Do you think we need to add in callbacks as well?

stnguyen90:
> @stnguyen90 added a check in the redirect as well. Do you think we need to add in callbacks as well?

@lohanidamodar, I think this is good enough!

christyjacob4 (approved):

eldadfux (approved):

PR #4977, opened 2023-01-09
feat: add code to account creation endpoint

christyjacob4:
## What does this PR do?

Creates a temporary endpoint to restrict console user signups using invite codes.

## Test Plan

## Related PRs and Issues

(Write your answer here.)


app/controllers/api/account.php
```diff
@@ -44,6 +44,96 @@
...
+    ->label('sdk.description', '/docs/references/account/create.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_ACCOUNT)
+    ->label('abuse-limit', 10)
+    ->param('userId', '', new CustomId(), 'Unique Id. Choose your own unique ID or pass the string `ID.unique()` to auto generate it. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.')
+    ->param('email', '', new Email(), 'User email.')
+    ->param('password', '', new Password(), 'User password. Must be at least 8 chars.')
+    ->param('name', '', new Text(128), 'User name. Max length: 128 chars.', true)
+    ->param('code', '', new Text(64), 'An invite code to restrict user signups on the Appwrite console. Users with an invite code will be able to create accounts irrespective of email and IP whitelists.', true)
```
TorstenDittmann, line 69: Shouldn't the Whitelist validator be used here? 🤔

eldadfux: Might work smoother, I agree.

app/config/variables.php
```diff
@@ -88,6 +88,15 @@
                 'question' => '',
                 'filter' => ''
             ],
+            [
+                'name' => '_APP_CONSOLE_WHITELIST_CODES',
```
eldadfux, line 92: Because this is just an internal feature, I don't think we should officially expose it, as we won't be able to provide proper long term support for it.

app/controllers/api/account.php
eldadfux: Thinking again, both can work OK. @christyjacob4 I'd leave for you to decide.

app/controllers/api/account.php
```diff
@@ -44,6 +44,96 @@
 $oauthDefaultSuccess = '/auth/oauth2/success';
 $oauthDefaultFailure = '/auth/oauth2/failure';
 
+App::post('/v1/account/invite')
+    ->desc('Create Account using an invite code')
+    ->groups(['api', 'account', 'auth'])
```
eldadfux, line 49: How do we skip the auth protection we have in the special auth init method?

Referring to these checks: 
https://github.com/appwrite/appwrite/blob/master/app/controllers/shared/api.php#L267-L268

eldadfux (commented):
Might be a good idea to add a test as well.

app/config/variables.php
christyjacob4: Alright will remove it 👍

app/controllers/api/account.php
christyjacob4: True we could use the Whitelist Validator, however this check only needs to run on the `console` project. Adding it as part of the whitelist validator would mean we run this check for other projects as well

app/controllers/api/account.php
christyjacob4: I don't see how its a problem ? 
Also for the console project these checks are skipped anyways 
https://github.com/appwrite/appwrite/blob/master/app/controllers/shared/api.php#L279

eldadfux: Got it what about the whitelist check? how is that being skipped?

app/controllers/api/account.php
eldadfux: OK

eldadfux (approved):

Issue #4428, opened 2022-10-11
🐛 Bug Report: Deleting Collection Doesn't Delete from Database


stnguyen90:
### 👟 Reproduction steps

1. Delete a collection

### 👍 Expected behavior

The collection is deleted from the database

### 👎 Actual Behavior

The collection is no longer available via APIs, but is still in the database. Looking at the appwrite-worker-deletes logs shows:

```
[...]

PR #4983, opened 2023-01-11
Fix deletes worker not deleting database collections

abnegate:
## What does this PR do?

The delete collection endpoint would remove a collection from the projects `databases` tables, but collection tables themselves (e.g. `_1_database_1_collection_1`) were not being deleted from the database because the deletes worker was checking if the document `$collection` attribute was `collections` which is no longer true.

This PR changes the worker to check if the `$collection` attribute starts with `database_` instead.

## Test Plan

Manual tests

Screenshot 2023-01-11 at 5 25 06 PM

## Related PRs and Issues

Fixes https://github.com/appwrite/appwrite/issues/4428

Yes

Yes

### DIFF
--- a/app/workers/deletes.php
+++ b/app/workers/deletes.php
@@ -44,9 +44,6 @@ class DeletesV1 extends Worker
                     case DELETE_TYPE_DATABASES:
                         $this->deleteDatabase($document, $project->getId());
                         break;
-                    case DELETE_TYPE_COLLECTIONS:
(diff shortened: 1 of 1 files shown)


stnguyen90 (approved):
Tested and it does look like the table for the collection and the rows for the collection's attributes and indexes are all properly deleted.

christyjacob4 (approved):

Issue #4930, opened 2022-12-28
🐛 Bug Report: deleted project still exists in appwrite


usamashafiqs:
### 👟 Reproduction steps

Deleted project not showing in appwrite console but showing during migration.

I migrated to v1.2.0 after migration it showing me 8 projects are migrated but only have 4 project in appwrite console.

![Screenshot 2022-12-28 225857](https://user-images.githubusercontent.com/18592683/209853814-31f5203a-9559-4ac3-bb7a-89930a91f463.png)

I deleted these projects a mouth ago.
It showing me all of my deleted project during migration.

![image](https://user-images.githubusercontent.com/18592683/209853687-5e8c30a0-9ba8-4fc0-b21e-1b5a45465387.png)

### 👍 Expected behavior

Delete project should delete completely form appwrite.

### 👎 Actual Behavior

Deleted project not showing in appwrite console but showing during migration.

### 🎲 Appwrite version

Version 1.2.x

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

stnguyen90:
@usamashafiqs thanks for raising this issue! 🙏🏼 Do you happen to remember what version you were on when you deleted the projects?

usamashafiqs:
@stnguyen90 can't remember the exact date of deletion 1 to 2 mouths ago.

suleimanas:
@stnguyen90  I can confirm this issue on version 1.1.2

suleimanas:
And version 1.2.0 all the project's tables and data remains after delete

suleimanas:
@usamashafiqs yes a quick way is to use **phpmyadmin**

https://hub.docker.com/_/phpmyadmin

in my `docker-compose.override.yml`
```
services:
  phpmyadmin:
    image: phpmyadmin/phpmyadmin
    restart: unless-stopped
    ports:
      - "8090:80"
    environment:
      PMA_HOST: mariadb
      UPLOAD_LIMIT: 999999999
    depends_on:
      - mariadb
    networks:
      - appwrite
```

suleimanas:
@usamashafiqs you can check the **appwrite-uploads** volume

[...]

PR #4984, opened 2023-01-11
Fix deletes worker not deleting project database tables

abnegate:
## What does this PR do?

The delete project endpoint would delete the project from the `_console_projects` table, then the deletes worker tried to delete the entire `appwrite` database when a project was deleted, which would fail, but because the project was deleted from the projects table, the project was no longer accessible.

This meant all the project-specific tables were present in the database with no way to remove them except manually deleting them via SQL.

Project certificates, domains, functions, builds and cache were also not deleted with a project

This PR fixes the deletes worker by:

- Deleting all project domains and certificates
- Getting a project-specific database
- Listing all the project collection tables
- Deleting them all page by page
- Deleting the project metadata tables
- Deleting function, builds and cache

It also updates the existing `getProjectDB` function in the worker base to accept a project document to allow getting a project-specific database with a project document instead of a project ID. This avoids a race condition where the delete project endpoint could delete a project document before the worker can fetch it.

## Test Plan

Manual test

## Related PRs and Issues

Fixes https://github.com/appwrite/appwrite/issues/4930

Yes

Yes

### DIFF
--- a/app/workers/deletes.php
+++ b/app/workers/deletes.php
@@ -253,8 +253,10 @@ class DeletesV1 extends Worker
      */
     protected function deleteProject(Document $document): void
     {
+        $projectId = $document->getId();
+
         // Delete project tables
-        $dbForProject = $this->getProjectDB($document->getId(), $document);
+        $dbForProject = $this->getProjectDB($projectId, $document);
 
         $limit = 50;
         $offset = 0;
@@ -282,10 +284,14 @@ class DeletesV1 extends Worker
         }
 
         // Delete all storage directories
-        $uploads = $this->getFilesDevice($document->getId());
-        $cache = new Local(APP_STORAGE_CACHE . '/app-' . $document->getId());
+        $uploads = $this->getFilesDevice($projectId);
+        $functions = $this->getFunctionsDevice($projectId);
+        $builds = $this->getBuildsDevice($projectId);
+        $cache = $this->getCacheDevice($projectId);
 
         $uploads->delete($uploads->getRoot(), true);
+        $functions->delete($functions->getRoot(), true);
+        $builds->delete($builds->getRoot(), true);
         $cache->delete($cache->getRoot(), true);
     }
 
--- a/app/workers/deletes.php
+++ b/app/workers/deletes.php
@@ -172,6 +172,7 @@ class DeletesV1 extends Worker
     /**
      * @param Document $document database document
      * @param string $projectId
+     * @throws Exception
      */
     protected function deleteDatabase(Document $document, string $projectId): void
     {
@@ -191,6 +192,7 @@ class DeletesV1 extends Worker
     /**
      * @param Document $document teams document
      * @param string $projectId
+     * @throws Exception
      */
     protected function deleteCollection(Document $document, string $projectId): void
     {
@@ -217,6 +219,7 @@ class DeletesV1 extends Worker
 
     /**
      * @param string $hourlyUsageRetentionDatetime
+     * @throws Exception
      */
     protected function deleteUsageStats(string $hourlyUsageRetentionDatetime)
     {
@@ -232,6 +235,7 @@ class DeletesV1 extends Worker
     /**
      * @param Document $document teams document
      * @param string $projectId
+     * @throws Exception
      */
     protected function deleteMemberships(Document $document, string $projectId): void
     {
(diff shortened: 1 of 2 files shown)


src/Appwrite/Resque/Worker.php
```diff
@@ -177,56 +179,84 @@ protected function getProjectDB(string $projectId): Database
...
 
+    /**
+     * Get internal project database given the project document
+     *
+     * Allows avoiding race conditions when modifying the projects collection
+     * @param Document $project
+     * @return Database
+     * @throws Exception
+     */
+    protected function getProjectDBFromDocument(Document $project): Database
```
stnguyen90, line 190: Hmm is this going to cause a lot of conflict with:

https://github.com/appwrite/appwrite/blob/708994d4a66838ece5ae0b37a2902bf6bda16e2e/src/Appwrite/Resque/Worker.php#L175

app/workers/deletes.php
```diff
@@ -245,25 +249,56 @@ protected function deleteMemberships(Document $document, string $projectId): voi
...
+
+            if (empty($collections)) {
+                break;
+            }
+
+            foreach ($collections as $collection) {
+                $dbForProject->deleteCollection($collection->getId());
+            }
+
+            $offset += $limit;
```
lohanidamodar, line 275: shouldn't after delete, the offset should still be 0?

lohanidamodar (commented):
Looks mostly good.
Do we need to delete the project document from the console DB?

app/workers/deletes.php
abnegate: This is to delete all collections page by page, so 0-50, then 50-100 etc

abnegate:
> Do we need to delete the project document from the console DB?

@lohanidamodar The projects controller handles that

app/workers/deletes.php
abnegate: Oh now I see what you mean, you're right, will update

app/workers/deletes.php
```diff
@@ -245,25 +249,56 @@ protected function deleteMemberships(Document $document, string $projectId): voi
...
+            // which will throw an exception here because the metadata collection is already deleted.
+        }
 
         // Delete all storage directories
-        $uploads = $this->getFilesDevice($document->getId());
-        $cache = new Local(APP_STORAGE_CACHE . '/app-' . $document->getId());
+        $uploads = $this->getFilesDevice($projectId);
+        $functions = $this->getFunctionsDevice($projectId);
+        $builds = $this->getBuildsDevice($projectId);
+        $cache = $this->getCacheDevice($projectId);
```
christyjacob4, line 290: Don't we also need to delete the `config` and `certificates` associated with a project? 
https://github.com/appwrite/appwrite/blob/master/app/init.php#L114-L115

src/Appwrite/Resque/Worker.php
```diff
@@ -151,82 +152,100 @@ public function tearDown(): void
...
                 $database = new Database(new MariaDB($register->get('db')), $cache);
                 $database->setDefaultDatabase(App::getEnv('_APP_DB_SCHEMA', 'appwrite'));
                 $database->setNamespace($namespace); // Main DB
 
-                if (!empty($projectId) && !$database->getDocument('projects', $projectId)->isEmpty()) {
-                    throw new \Exception("Project does not exist: {$projectId}");
+                if (
+                    $project === null
+                    && !empty($projectId)
+                    && !$database->getDocument('projects', $projectId)->isEmpty()
```
christyjacob4, line 246: Will this condition ever be true ? 

If `project` is null, `projectId` will also be empty right? 
which means the condition `$project == null && !empty(projectId)` is always false ?

christyjacob4 (changes requested):
Please address the comments

app/workers/deletes.php
abnegate: Both are associated with an installation instead of a project :+1:

src/Appwrite/Resque/Worker.php
abnegate: You can pass `$projectId`, or the `$project` document, so this case will be true when a document was not passed, and a project with the given ID can not be found.

I have just added support for passing the entire document, to handle the race condition in this particular case where the project document is deleted from the database in the controller, but is required here in the worker

app/workers/deletes.php
```diff
@@ -245,25 +249,51 @@ protected function deleteMemberships(Document $document, string $projectId): voi
...
+                break;
+            }
+
+            foreach ($collections as $collection) {
+                $dbForProject->deleteCollection($collection->getId());
+            }
+        }
+
+        // Delete metadata tables
+        try {
```
lohanidamodar, line 274: I don't see where we delete the project document itself?

christyjacob4: Thats done in the controller itself 
https://github.com/appwrite/appwrite/blob/master/app/controllers/api/projects.php#L642

app/workers/deletes.php
christyjacob4: Yes, but we still need to delete the certificates of the domains associated with a project right ?

abnegate: Ah yep my mistake!

christyjacob4 (approved):

Issue #4990, opened 2023-01-11
🚀 Feature: Re-hash Password on Phone Update


stnguyen90:
### 🔖 Feature description

Users can be imported with a different password hash, but they should be re-hashed to the default algorithm. This is done for the update email API call:

https://github.com/appwrite/appwrite/blob/514b42bea2116444082553ea758eeb9f256421dc/app/controllers/api/account.php#L1564

but not for the update phone API call.

### 🎤 Pitch

Update the update phone API call to make sure it's consistent with update email.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

harshmange44:
@stnguyen90 I'd like to contribute to this feature.

stnguyen90:
@harshmange44 assigned! thanks for your interest! 🙏🏼

harsh020:
@stnguyen90, I would also be interested in working on this. 

If I understand this correctly, you want the `updatePhone` api to re-hash the password if the user is anonymous. Am I right?

stnguyen90:
> @stnguyen90, I would also be interested in working on this.

@harsh020, thanks! 🙏🏼 However, this has already been assigned to @harshmange44 and it's enough to have 1 person working on this.
 
> If I understand this correctly, you want the `updatePhone` api to re-hash the password if the user is anonymous. Am I right?

Yes, the logic should be the same as the updateEmail API.

harshmange44:
@stnguyen90 I've raised a PR: [PR](https://github.com/appwrite/appwrite/pull/5003)

abhishek818:
@stnguyen90  @Meldiron  can i pick up this issue , have a MR --> https://github.com/appwrite/appwrite/pull/5269

stnguyen90:
Sorry, looks like we're going to change the behavior a bit. 

1. We don't want to update the password for users without a password
2. We don't want to verify the password if the user doesn't have a password

PR #5003, opened 2023-01-13
Modified PATCH '/v1/account/phone' API to re-hash password

harshmange44:
## What does this PR do?

Re-hash password using default algo on update phone (PATCH '/v1/account/phone' API call) if it's an anonymous user

## Test Plan

-

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/4990

Yes

Yes

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -1635,6 +1635,8 @@ App::patch('/v1/account/phone')
 
         $user
             ->setAttribute('password', $isAnonymousUser ? Auth::passwordHash($password, Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS) : $user->getAttribute('password', ''))
+            ->setAttribute('hash', $isAnonymousUser ? Auth::DEFAULT_ALGO : $user->getAttribute('hash', ''))
+            ->setAttribute('hashOptions', $isAnonymousUser ? Auth::DEFAULT_ALGO_OPTIONS : $user->getAttribute('hashOptions', ''))
             ->setAttribute('phone', $phone)
             ->setAttribute('phoneVerification', false) // After this user needs to confirm phone number again
             ->setAttribute('search', implode(' ', [$user->getId(), $user->getAttribute('name', ''), $user->getAttribute('email', ''), $phone]));
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -1634,6 +1634,7 @@ App::patch('/v1/account/phone')
         }
 
         $user
+            ->setAttribute('password', $isAnonymousUser ? Auth::passwordHash($password, Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS) : $user->getAttribute('password', ''))
             ->setAttribute('phone', $phone)
             ->setAttribute('phoneVerification', false) // After this user needs to confirm phone number again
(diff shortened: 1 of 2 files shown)


harshmange44:
Fixed #4990 https://github.com/appwrite/appwrite/issues/4990

app/controllers/api/account.php
```diff
@@ -1634,6 +1634,7 @@
         }
 
         $user
+            ->setAttribute('password', $isAnonymousUser ? Auth::passwordHash($password, Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS) : $user->getAttribute('password', ''))
```
stnguyen90, line 1637: Please also make sure to update the `hash` and `hashOptions` attributes.

stnguyen90 (changes requested):

app/controllers/api/account.php
harshmange44: @stnguyen90 Done, thanks for the review!

app/controllers/api/account.php
```diff
@@ -1634,6 +1634,9 @@
...
+            ->setAttribute('hash', $isAnonymousUser ? Auth::DEFAULT_ALGO : $user->getAttribute('hash', ''))
+            ->setAttribute('hashOptions', $isAnonymousUser ? Auth::DEFAULT_ALGO_OPTIONS : $user->getAttribute('hashOptions', ''))
```
stnguyen90, line 1639: @Meldiron is this logic right? Is it supposed to be `!$isAnonymousUser`?

stnguyen90, line 1639: @Meldiron is this logic right? Is it supposed to be `!$isAnonymousUser`?

Meldiron: Honestly not sure why we ask for password in update phone endpoint. What password is expected in case of anonymous user? Why does "update phone" also update password? If anonymous user used this endpoint and entered password, do I understand correctly this would claim account for him not only by phone, but also by password?

If all of that is expected, then the logic makes sense. If anonymous, generate password hash. If anonymous, configure hash algo. If not, leave as it was.

stnguyen90: @Meldiron,

> Honestly not sure why we ask for password in update phone endpoint.

Probably as an extra security measure. Same idea behind asking for password for update email. However, the user is less likely to have a password in this case...

> Why does "update phone" also update password?

Probably because update email does. And I think the justification was we want to rehash as soon as we have the password to make sure it's hashed using the default algorithm

> If anonymous user used this endpoint and entered password, do I understand correctly this would claim account for him not only by phone, but also by password?

What do you mean by claim? I think it would set the password, just like the update email API

> If all of that is expected, then the logic makes sense. If anonymous, generate password hash. If anonymous, configure hash algo. If not, leave as it was.

I'd like to get your insight into what should happen since you were involved in adding this type of code to the update email endpoint.

harshmange44: @stnguyen90 Any update, are we going to merge this one?

stnguyen90: Sorry, looks like we're going to change the behavior a bit. 

1. We don't want to update the password for users without a password
2. We don't want to verify the password if the user doesn't have a password

stnguyen90 (approved):

stnguyen90 (changes requested):
See my previous comment

PR #5014, opened 2023-01-17
Update create-team-membership.md

gewenyu99:
## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)

### DIFF
--- a/docs/references/teams/create-team-membership.md
+++ b/docs/references/teams/create-team-membership.md
@@ -1,7 +1,7 @@
-Invite a new member to join your team. Provide an ID for existing users, or invite unregistered users using an email or phone number. If initiated from a Client SDK, an email or sms with a link to join the team will be sent to the invited user and an account will be created for them if one doesn't exist. If initiated from a Server SDKs, the new member will be added automatically to the team.
+Invite a new member to join your team. Provide an ID for existing users, or invite unregistered users using an email or phone number. If initiated from a Client SDK, Appwrite will send an email or sms with a link to join the team to the invited user, and an account will be created for them if one doesn't exist. If initiated from a Server SDK, the new member will be added automatically to the team.
 
-You only need to provide one of User ID, email, or phone number. When multiple are provided, the priority will be user ID > email > phone number, lower priority parameters are ignored.
+You only need to provide one of a user ID, email, or phone number. Appwrite will prioritize accepting the user ID > email > phone number if you provide more than one of these parameters.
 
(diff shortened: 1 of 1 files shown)


docs/references/teams/create-team-membership.md
```diff
@@ -1,7 +1,7 @@
-Invite a new member to join your team. You can also add existing users with their unique ID. You can also invite using a phone number. If initiated from the client SDK, an email or sms with a link to join the team will be sent to the member's email address or phone number and an account will be created for them should they not exist already. If initiated from server-side SDKs, the new member will automatically be added to the team.
+Invite a new member to join your team. Provide an ID for existing users, or invite unregistered users using an email or phone number. If initiated from a Client SDK, Appwrite will send an email or sms with a link to join the team to the invited user, and an account will be created for them if one doesn't exist. If initiated from a Server SDK, the new member will be added automatically to the team.
 
-You can only provide one of the User ID, email or phone number to add member to the team. And the priority will be User ID > email and > phone number.
+You only need to provide one of a user ID, email, or phone number. Appwrite will prioritize accepting the user ID > email > phone number if you provide more than one of these parameters.
```
lohanidamodar, line 3: when more than one parameter is provided, we also check to verify it matches, for example if userID and email is provided, then we check if the user with that userID has the same email, and throw exception otherwise. Can we add this info somehow, same for `email` and `phone` with priority over email.

gewenyu99: Maybe we don't need this information. I think, practically speaking, it shouldn't make a difference. I think getting an error is enough cause this only happens when they make a mistake.

If we need to document this:
```suggestion
You only need to provide one of a user ID, email, or phone number. Appwrite will prioritize accepting the user ID > email > phone number if you provide more than one of these parameters. If more than one parameter is provided, Appwrite will verify if the user ID matches the provided email or phone and throw an error otherwise.
```

lohanidamodar: I'll leave the final decision to @eldadfux ?

eldadfux: Agree with @gewenyu99. We provide enough information, we can reconsider if we see this becomes a popular error. @gewenyu99 might be good to monitor our error stats on Sentry as part of our DX workflows, and try to detect problematic places in the dev-journey.

lohanidamodar: okay

PR #5024, opened 2023-01-20
Update Update Document API to check X-Appwrite-Timestamp header

stnguyen90:
## What does this PR do?

Appwrite will refer to the X-Appwrite-Timestamp header for when the request was originally sent and reject the request if the Timestamp is older than the updatedAt timestamp of the document.

## Test Plan

E2E tests

## Related PRs and Issues

Dependent PR:

* https://github.com/utopia-php/database/pull/233
* https://github.com/utopia-php/audit/pull/43
* https://github.com/utopia-php/abuse/pull/43

Yes


stnguyen90:
@eldadfux I've added the check in the Update Document API. What other APIs should also do this?

app/controllers/api/databases.php
```diff
@@ -2330,11 +2333,38 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
...
         $data['$permissions'] = $permissions;
 
+        $timestamp = null;
+        $timestampHeader = $request->getHeader('x-appwrite-timestamp');
+        if (!empty($timestampHeader)) {
+            $timestamp = new \DateTime($timestampHeader);
+        }
+
         try {
+            $internalCollectionId = 'database_' . $database->getInternalId() . '_collection_' . $collection->getInternalId();
```
abnegate, line 2343: The name `$internalCollectionId` is a little confusing since it does not equal `$collection->getInternalId()`

app/controllers/general.php
```diff
@@ -485,6 +485,10 @@
                     $error->setType(AppwriteException::GENERAL_ROUTE_NOT_FOUND);
                     break;
             }
+        } elseif ($error instanceof Utopia\Database\Exception\Conflict) {
+            $error = new AppwriteException(AppwriteException::DOCUMENT_UPDATE_CONFLICT, null, null, $error);
```
abnegate, line 489: I think this can just be:

```suggestion
            $error = new AppwriteException(AppwriteException::DOCUMENT_UPDATE_CONFLICT);
```

abnegate (commented):
I think we need to cover the users and storage APIs too

app/controllers/general.php
stnguyen90: I wanted to make sure to pass the underlying error through as well.

app/controllers/api/databases.php
stnguyen90: @abnegate do you have a recommendation for a different variable name?

stnguyen90: @abnegate, I've updated the variable name to `$databaseCollectionId`.

abnegate: I think `$privateCollectionId` might be a bit clearer

stnguyen90:
> I think we need to cover the users and storage APIs too

@abnegate, offline support will only be supported in client sdks and it won't support file operations (at least for now).

app/controllers/shared/api.php
```diff
@@ -111,6 +109,18 @@
             throw new Exception(Exception::PROJECT_UNKNOWN);
         }
 
+        // Validate x-appwrite-timestamp header
+        $timestampHeader = $request->getHeader('x-appwrite-timestamp');
```
eldadfux, line 113: We usually don't use `App::setResource` from inside a route / hook. We can set it on [init.php](https://github.com/appwrite/appwrite/blob/master/app/init.php) and still access the request header.

eldadfux (changes requested):

app/controllers/shared/api.php
stnguyen90: @eldadfux are we able to do validation and throw an exception in `App::setResource()`?

stnguyen90: @eldadfux updated!

abnegate (approved):

eldadfux (approved):
Let's change the target branch to 1.3.x

PR #5050, opened 2023-01-26
Optimized and added comments to UsageUsers model class

DeepakKumarKhatri:
## What does this PR do?

- I added class level documentation to provide a clear understanding of the class's purpose
- I replaced the use of the array_fill() function with an empty array for the 'default' and 'example' properties in the commonProperties array. This eliminates the unnecessary use of the array_fill() function and makes the code more readable.
- I added additional comments to the class's methods, properties and constants to provide a clear understanding of their purpose and usage.
- I added missing return type to getType() method.
- Changed the return value of getName() method to static::class which will return the fully qualified class name at runtime.

## Test Plan

1. Verified that the UsageUsers class extends the Model class and implements the getName() and getType() methods as expected.
2. Tested the various properties and rules added to the class using the addRule() method, such as range, usersCount, usersCreate, etc. Verify that the properties have the correct type, description, default value, and example values.
3. Tested the getName() method and ensure that it returns the correct class name.
4. Tested the getType() method and ensure that it returns the correct constant value, Response::MODEL_USAGE_USERS.
5. Verified that the $commonProperties array is used correctly and that it is correctly merged with the properties of each rule.
6. Tested the UsageUsers class in the context of the larger application and ensure that it is used correctly to handle usage statistics for users and sessions.
7. Verified that the class is working with the correct data types and handling the data as expected.
8. Tested the class with different test cases to verify its functionality under different scenarios.
9. Tested the class with different combinations of parameters for the range and check the results.

## Related PRs and Issues

- None

## Checklist

- [ Yes ] Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [ No ] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

### DIFF
--- a/src/Appwrite/Utopia/Response/Model/UsageUsers.php
+++ b/src/Appwrite/Utopia/Response/Model/UsageUsers.php
@@ -5,10 +5,31 @@ namespace Appwrite\Utopia\Response\Model;
 use Appwrite\Utopia\Response;
 use Appwrite\Utopia\Response\Model;
 
+/**
+ * Class UsageUsers
+ *
+ * Model class representing a usage statistics for users and sessions.
+ */
 class UsageUsers extends Model
 {
+    /**
+     * Constant representing the type string.
+     *
+     * @var string
+     */
+    const TYPE_STRING = 'string';
+    const TYPE_METRIC = 'metric';
+
     public function __construct()
     {
+        $commonProperties = [
+            'type' => self::TYPE_METRIC,
+            'description' => '',
+            'default' => [],
+            'example' => [],
+            'array' => true
+        ];
+
         $this
             ->addRule('range', [
                 'type' => self::TYPE_STRING,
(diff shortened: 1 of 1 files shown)


src/Appwrite/Utopia/Response/Model/UsageUsers.php
```diff
@@ -82,14 +71,14 @@ public function __construct()
      */
     public function getName(): string
     {
-        return 'UsageUsers';
+        return static::class;
```
stnguyen90, line 74: Unfortunately, I think this will also return the namespace, which we don't want.

src/Appwrite/Utopia/Response/Model/UsageUsers.php
```diff
@@ -5,72 +5,61 @@
...
+ */
 class UsageUsers extends Model
 {
+    /**
+     * Constant representing the type string.
+     *
+     * @var string
+     */
+    const TYPE_STRING = 'string';
+    const TYPE_METRIC = 'metric';
```
stnguyen90, line 21: We don't do this in other models, and it doesn't look like this is used in many places, so I don't think it's worth it to add.

src/Appwrite/Utopia/Response/Model/UsageUsers.php
```diff
@@ -5,72 +5,61 @@
...
                 'example' => '30d',
             ])
-            ->addRule('usersCount', [
-                'type' => Response::MODEL_METRIC,
+            ->addRule('usersCount', $commonProperties + [
                 'description' => 'Aggregated stats for total number of users.',
-                'default' => [],
-                'example' => [],
-                'array' => true
             ])
```
stnguyen90, line 42: I think this works differently than expected...According to the [docs](https://www.php.net/manual/en/language.operators.array.php), the keys in the right array do not replace the matching ones in the left array.

This type of nuance with how the operator works suggest it might be simpler and clearer to leave it the way it is.

stnguyen90 (changes requested):

PR #5075, opened 2023-02-03
Enable SMTP debugging in doctor

stnguyen90:
## What does this PR do?

Enable debug messages when checking SMTP to help troubleshoot SMTP problems.

## Test Plan

image

## Related PRs and Issues

- #1390

## Checklist

### DIFF
--- a/app/tasks/doctor.php
+++ b/app/tasks/doctor.php
@@ -134,7 +134,9 @@ $cli
         }
 
         try {
-            $mail = $register->get('smtp'); /* @var $mail \PHPMailer\PHPMailer\PHPMailer */
+            /** @var PHPMailer\PHPMailer\PHPMailer $mail */
+            $mail = $register->get('smtp');
+            $mail->SMTPDebug = PHPMailer\PHPMailer\SMTP::DEBUG_SERVER;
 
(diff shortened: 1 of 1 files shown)


app/tasks/doctor.php
```diff
@@ -134,7 +134,9 @@
         }
 
         try {
-            $mail = $register->get('smtp'); /* @var $mail \PHPMailer\PHPMailer\PHPMailer */
+            /** @var PHPMailer\PHPMailer\PHPMailer $mail */
+            $mail = $register->get('smtp'); 
+            $mail->SMTPDebug = PHPMailer\PHPMailer\SMTP::DEBUG_SERVER;
```
stnguyen90, line 139: @eldadfux, [you said it would be good to enable debugging in doctor](https://github.com/appwrite/appwrite/issues/1390#issuecomment-880077189), but will it be too messy?

LyVanBong:
I am also encountering a problem with the SMTP configuration of the Outlook Office 365 email.

app/tasks/doctor.php
eldadfux: Yeh, this feels quite messy. We can try and have a different command for doing this.

Something like:
`docker exec appwrite debug smtp`

PR #5126, opened 2023-02-15
switch SDK response model from account to user

lohanidamodar:
## What does this PR do?

- Switch SDK response model from account to User

## Test Plan

N/A

## Related PRs and Issues

- https://github.com/appwrite/sdk-for-flutter/issues/93

## Checklist


Meldiron (commented):
Why do we want this change? I think we have separate model so account API cannot show password hash and password hash options, while Users API can.

eldadfux:
> Why do we want this change? I think we have separate model so account API cannot show password hash and password hash options, while Users API can.

This will only affect and simplify the SDKs + Docs.

Meldiron (approved):

PR #5137, opened 2023-02-17
Docs: Fix Max Size of Function Response

gewenyu99:
## What does this PR do?

The response size has been 1MB for a while instead of 4,000 bytes.

### DIFF
--- a/src/Appwrite/Utopia/Response/Model/Execution.php
+++ b/src/Appwrite/Utopia/Response/Model/Execution.php
@@ -62,19 +62,19 @@ class Execution extends Model
             ])
             ->addRule('response', [
                 'type' => self::TYPE_STRING,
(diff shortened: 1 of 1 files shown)


Meldiron (approved):
Looks good. Please keep this in mind for next function iteration, it might be changing again 😅 I have plan to go as high as 20MB

gewenyu99:
> Looks good. Please keep this in mind for next function iteration, it might be changing again 😅 I have plan to go as high as 20MB

Can we maybe expose this as a variable we can hook into instead? 👀 So we don't forget.

Hardanish-Singh (commented):
LGTM

gewenyu99:
@christyjacob4 @Meldiron Can I close this one? ;)

PR #5143, opened 2023-02-20
Remove count, limit and offset limitations for queries

abnegate:
## What does this PR do?

Remove count, limit and offset limitations for queries. Specifically:

- Remove limit of 5000 for page offset
- Remove limit of 100 for page size

## Test Plan

Existing tests

## Related PRs and Issues

https://github.com/utopia-php/database/pull/237
https://github.com/utopia-php/audit/pull/42
https://github.com/utopia-php/abuse/pull/42

## Checklist


eldadfux (commented):
1. If we remove the 5,000 count limit, on big tables every query would fail with a timeout, no matter how well the tables/collections are indexed.
2. Where are we removing the index validation when passing queries?

fogelito:
Today we fetch 5000 rows with a limit , and count all fetched rows VS regular aggregation count with no limit, hard to say from what number each is better. what we are doing today is not optimal too.
Can we save the result in Redis, with an auto-expiration key? 
We do not have timeouts on count only on find. Count method does not use the find method.

eldadfux (approved):

PR #5196, opened 2023-03-06
feat: team preferences

TorstenDittmann:
## What does this PR do?

- adds team preferences
- adds `prefs` to Team Response Model
- adds events for updating team prefs

## Test Plan

- teams API
- webhooks
- realtime

## Related PRs and Issues

n/a

## Checklist


docs/references/teams/get-team-prefs.md
```diff
@@ -0,0 +1 @@
+Get the team preferences by its unique ID.
```
eldadfux, line 1: We might want to add a reference to user prefs if they prefer the settings to be stored per user and link to that part of the docs and do the same vice versa. This is relevant for all the endpoints. @gewenyu99 wdyt? I feel we should advocate for similar features.

eldadfux (changes requested):

app/config/events.php
```diff
@@ -190,6 +190,9 @@
         ],
         'update' => [
             '$description' => 'This event triggers when a bucket is updated.',
+            'prefs' => [
+                '$description' => 'This event triggers when a team\'s preferences is updated.',
```
gewenyu99, line 194:
```suggestion
                '$description' => 'This event triggers when a team\'s preferences are updated.',
```

docs/references/teams/get-team-prefs.md
gewenyu99: This description is pretty short; we can do this.

Something like:

>Get a team's shared preferences by its unique ID. If a preference doesn't need to be shared by all team members, prefer storing them in [user preferences](url).

docs/references/teams/update-team-name.md
```diff
@@ -0,0 +1 @@
+Update the team name by its unique ID.
```
gewenyu99, line 1:
```suggestion
Update the team's name by its unique ID.
```

docs/references/teams/update-team-prefs.md
```diff
@@ -0,0 +1 @@
+Update the team preferences by its unique ID. The object you pass is stored as is, and replaces any previous value. The maximum allowed prefs size is 64kB and throws error if exceeded.
```
gewenyu99, line 1:
```suggestion
Update the team's preferences by its unique ID. The object you pass is stored as is and replaces any previous value. The maximum allowed prefs size is 64kB and throws an error if exceeded.
```

docs/references/teams/get-team-prefs.md
gewenyu99, line 1:
```suggestion
Get the team's preferences by its unique ID.
```

gewenyu99 (changes requested):

app/config/events.php
```diff
@@ -190,6 +190,9 @@
         ],
         'update' => [
             '$description' => 'This event triggers when a bucket is updated.',
```
christyjacob4, line 192:
```suggestion
            '$description' => 'This event triggers when a team is updated.',
```
Don't we wanna update this description as well 😅

docs/references/teams/update-team-name.md
```diff
@@ -0,0 +1 @@
+Update the team's name by its unique ID.
```
christyjacob4, line 1: We should mention that only members with the owner role can update the team name

christyjacob4 (changes requested):
Looks good. please address the comments

app/config/events.php
christyjacob4: Similarly, some of the other descriptions in the teams events use `bucket`

docs/references/teams/update-team-name.md
TorstenDittmann: Actually I don't think its protected by the `owner` role.

app/controllers/api/teams.php
```diff
@@ -241,6 +271,51 @@
         $response->dynamic($team, Response::MODEL_TEAM);
     });
 
+App::patch('/v1/teams/:teamId/prefs')
```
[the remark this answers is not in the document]
TorstenDittmann: Actually PUT should be used here, but PATCH is actually the right way Method for name - since it only changes the name of the team resource - not all.

app/controllers/api/teams.php
```diff
@@ -241,6 +271,51 @@
...
+    ->inject('events')
+    ->action(function (string $teamId, array $prefs, Response $response, Database $dbForProject, Event $events) {
+
+        $team = $dbForProject->getDocument('teams', $teamId);
+
+        if ($team->isEmpty()) {
+            throw new Exception(Exception::TEAM_NOT_FOUND);
+        }
+
+        $roles = Authorization::getRoles();
```
eldadfux, line 303: Are we doing this check just for the proper error? Isn't the write operation on the team's document already protected by permissions?

docs/references/teams/update-team-name.md
eldadfux: @TorstenDittmann please check, sounds weird if it doesn't.

TorstenDittmann: no, works as intended. I somehow mixed things up 👍🏻

app/config/events.php
```diff
@@ -189,7 +189,10 @@
             '$description' => 'This event triggers when a bucket is deleted.',
```
abnegate, line 189: Looks like a couple of the other team events are using `bucket` instead of `team` in the description, might as well fix them too

app/controllers/api/teams.php
```diff
@@ -208,7 +238,7 @@
     ->label('sdk.auth', [APP_AUTH_TYPE_SESSION, APP_AUTH_TYPE_KEY, APP_AUTH_TYPE_JWT])
     ->label('sdk.namespace', 'teams')
     ->label('sdk.method', 'update')
```
abnegate, line 240: Should we change the SDK method too?

TorstenDittmann: good question, that would end up in a breaking change 🤔

abnegate: What do you think @eldadfux?

TorstenDittmann: actually not a breaking change, since the url doesnt change. so will work with old sdks

abnegate (approved):

PR #5197, opened 2023-03-06
Dynamic attibute for related Primary key

fogelito:
In order to relate an attribute for a Primary key we need a dynamic attribute , since in some adapters the primary key can be a string not always an integer


abnegate:
@fogelito Do we need both this one and https://github.com/utopia-php/database/pull/243?

PR #5210, opened 2023-03-10
Fix yasd so that the debugger works

stnguyen90:
## What does this PR do?

* Update the base image since the yasd extension wasn't installed correctly
* Fix the path to the yasd_init.php file
* Use `host.docker.internal` as the remote host because that should always work inside docker
* Mount the dev folder into the appwrite container so the yasd_init.php file is picked up.

## Test Plan

Manual

## Related PRs and Issues

None

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -346,8 +346,7 @@ If you are in PHP Storm you don't need any plugin. Below are the settings requir
 
 1. Create an init file. 
 2. Duplicate **dev/yasd_init.php.stub** file and name it **dev/yasd_init.php**.
-3. Change the IP address to your development machine's IP. Without the proper IP address, the debugger won't connect. 
-4. Set **DEBUG** build arg in **appwrite** service in **docker-compose.yml** file.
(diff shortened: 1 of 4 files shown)


zvikfir:
I cloned the repo yesterday, and I was actually using `host.docker.internal`, too. At first I didn't understand what "development machine's IP" meant, so I think this is a really good PR. 

I went through the changes and I have no comments, but I want to test that it works. However, I haven't been able to run the debugger until now.

Do you think you could help me figure what the problem is so I can test the changes you've made work?

I followed the instructions and changed the `DEBUG` env variable to true. I also used the launch configurations provided.
When I run them I get an orange footer in my vscode, and the appwrite is running through docker compose.

However, breakpoints are not working, and in the debug console when I try to type it says: 
"Cannot evaluate code without a connection".

Am I doing something wrong?

stnguyen90:
@zvikfir, did you start appwrite after launching the debugger?

lohanidamodar (approved):

PR #5247, opened 2023-03-19
Fix: scheduler env variables

Meldiron:
## What does this PR do?

Fix missing env variables on functions executed over scheduler.

## Test Plan

- [x] Manual QA

## Related PRs and Issues

x

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -116,7 +116,6 @@ RUN mkdir -p /storage/uploads && \
 
 # Executables
 RUN chmod +x /usr/local/bin/doctor && \
-    chmod +x /usr/local/bin/patch-create-missing-schedules && \
     chmod +x /usr/local/bin/patch-delete-schedule-updated-at-attribute && \
     chmod +x /usr/local/bin/maintenance && \
     chmod +x /usr/local/bin/volume-sync && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -117,6 +117,7 @@ RUN mkdir -p /storage/uploads && \
 # Executables
 RUN chmod +x /usr/local/bin/doctor && \
     chmod +x /usr/local/bin/patch-create-missing-schedules && \
+    chmod +x /usr/local/bin/patch-delete-schedule-updated-at-attribute && \
     chmod +x /usr/local/bin/maintenance && \
     chmod +x /usr/local/bin/volume-sync && \
     chmod +x /usr/local/bin/usage && \
--- a/app/controllers/api/functions.php
+++ b/app/controllers/api/functions.php
@@ -84,7 +84,6 @@ App::post('/v1/functions')
             'deployment' => '',
             'events' => $events,
             'schedule' => $schedule,
-            'scheduleUpdatedAt' => DateTime::now(),
             'timeout' => $timeout,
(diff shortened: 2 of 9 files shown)


Dockerfile
```diff
@@ -117,6 +117,7 @@ RUN mkdir -p /storage/uploads && \
 # Executables
 RUN chmod +x /usr/local/bin/doctor && \
     chmod +x /usr/local/bin/patch-create-missing-schedules && \
```
christyjacob4, line 119: can you now delete the old patch file ?

Meldiron: Just bin file, or the while patch task?
We might want to keep it for historical reasons. Allows copy&pasting when the similar patch is needed. Feels more safer than relaying on Migration files

src/Appwrite/Platform/Tasks/PatchDeleteScheduleUpdatedAtAttribute.php
```diff
@@ -0,0 +1,70 @@
...
+            }
+            $projects = $dbForConsole->find('projects', $projectsQueries);
+
+            if (count($projects) === 0) {
+                break;
+            }
+
+            foreach ($projects as $project) {
+                Console::log("Checking Project " . $project->getAttribute('name') . " (" . $project->getId() . ")");
+                $dbForProject = $getProjectDB($project);
```
christyjacob4, line 53: shouldn't this be reclaimed back in the pool ?

app/controllers/api/functions.php
```diff
@@ -786,22 +772,6 @@
...
-
-        $active = !empty($function->getAttribute('schedule'));
-
-        if ($active) {
-            $schedule->setAttribute('resourceUpdatedAt', datetime::now());
-        }
-
-        $schedule->setAttribute('active', $active);
-
-        Authorization::skip(fn () => $dbForConsole->updateDocument('schedules', $schedule->getId(), $schedule));
```
christyjacob4, line 803: why did we remove this ? How is the `active` attribute set when there is a new deployment ?

src/Appwrite/Platform/Tasks/PatchDeleteScheduleUpdatedAtAttribute.php
Meldiron: Added reclaim logic.

Dockerfile
christyjacob4: hmm, okay you can keep the actual file but remove the binary

app/controllers/api/functions.php
christyjacob4: Okay so is the builds worker updating the schedule ?

Meldiron: Correct: https://github.com/appwrite/appwrite/blob/567cb01b9940948575abb08a1654b92cc61fe6d9/app/workers/builds.php#L214

Dockerfile
Meldiron: Removed bin file

app/controllers/api/functions.php
christyjacob4: okay cool 👍

christyjacob4 (approved):

PR #5269, opened 2023-03-23
Update password for users having password (for Update Phone/Email APIs)

abhishek818:
## What does this PR do?
1. We don't want to update the password for users without a password
2. We don't want to verify the password if the user doesn't have a password

## Related PRs and Issues
4990, 4975

## Checklist

- [] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -1602,6 +1602,7 @@ App::patch('/v1/account/email')
 
         if (
             !$isAnonymousUser &&
+            !empty($user->getAttribute('password')) &&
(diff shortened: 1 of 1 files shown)


app/controllers/api/account.php
```diff
@@ -1602,17 +1602,22 @@
 
         if (
             !$isAnonymousUser &&
+            !empty($user->getAttribute('password')) &&
```
stnguyen90, line 1605: Unfortunately, this won't work because we actually set a random password for users if they sign in using OAuth2.

stnguyen90 (commented):
Thanks for your PR! 🙏🏼 Unfortunately, the change won't work in all cases. Also, we've decided to take this on internally so we won't need a PR anymore.

Issue #364, opened 2020-05-30
Create Kubernetes Operator


javib51:
Create a Kubernetes Operator to allow easy setup for appwrite stack in any Kubernetes cluster.

As discussed in issue #24.

javib51:
My idea is to start with a MVP of the operator and add more functionality afterward.

For now, development setup: 
- one instance of each container
- redis, influxdb and mariadb are included

What do you think @eldadfux?

eldadfux:
We would probably need Traefik container as well. ClamAV is going to be optional in next release (0.6.2).

You can use this compose file as a source of truth:
https://appwrite.io/docker-compose.yml

javib51:
@eldadfux thanks! I'm going to follow it.  

Another thing. For now, Im going to develop it in a private [repo](https://github.com/javib51/appwrite-operator). Whenever it's ready, we can decide where to place it.

eldadfux:
Sounds great, feel free to reach out on Discord in you want to chat more about this.

eldadfux:
hey @javib51, any update on this one?

pant03:
+1

kodumbeats:
Might be able to leverage [Kompose](https://kompose.io/user-guide/#alternative-conversions) to convert `docker-compose.yml` -> helm chart (tackling #24), and according to [RedHat](https://docs.okd.io/latest/operators/operator_sdk/osdk-helm.html), the Operator SDK knows what to do:

> One of the Operator SDK options for generating an Operator project includes leveraging an existing Helm chart to deploy Kubernetes resources as a unified application, without having to write any Go code.

I'll update if I can get it to work.

mcchin:
I have been following `Appwrite` for a while now, looking forward to setup in k8s environment for my personal projects. Armed with basic knowledge of k8s, I thought to give it a shot this weekend to at least able to deploy `Appwrite` on GKE, and see if I could at least get the admin dashboard up.

Looking through threads here at Github and Discord, I tried `Kompose` (ver 1.21.0) and as mentioned in previous threads on discord it won't run outright (pods keep crashing), unless I exposed some of the deployments as a service (redis, influxdb, telegraf, clamav), and update the PVC with r+w permission for config and certificate volume. (There is probably a correct way for this, but for now I just want to get it up)

So with that alone and with default env variables, I could do a portfoward and access `Appwrite` dashboard locally (e.g. `http://127.0.0.1:8080`), and able to signup, login, and start creating projects.

[...]

Issue #24, opened 2019-09-23
Create a Helm Chart for Appwrite


eldadfux:
Create a Helm chart for Appwrite to allow DevOps to easily setup the server stack in any Kubernetes based cluster.

RanadeepPolavarapu:
I'll take this.

eldadfux:
@RanadeepPolavarapu That will be great.

I guess I don't have much input on this task as I don't have too much experience with creating helm charts.

I guess it will be a good practice to follow our general guidelines of keeping things as lean, well documented and as standard as possible.

@RanadeepPolavarapu Do you think we need to add the chart in this repo or do you want me to open a new repo just for it (appwrite/helm-chart), what is more common across other projects maintaining a helm chart?

Should we aim to create the helm chart with production ready settings or maybe we create multiple versions available? Is there any convention for replacing or choosing ENV settings?

KadenLNelson:
> Do you think we need to add the chart in this repo or do you want me to open a new repo just for it (appwrite/helm-chart), what is more common across other projects maintaining a helm chart?

Just create the chart in this repository, it's redundant to manage it within another repository.

> Should we aim to create the helm chart with production ready settings or maybe we create multiple versions available? Is there any convention for replacing or choosing ENV settings?

The default `values.yaml` should container production ready settings. Since you're publishing/providing the `AppWrite` application/helm charts, I wouldn't recommend including a `values.dev.yaml` within the same VCS as the helm chart being provided. This is something that should be created within a `deployment` repository that end users of `appwrite` shouldn't be able to see..

eldadfux:
OK great! if you need any help feel free to contact me, we can also chat on Gitter.

Ishaan28malik:
Is it still open ? @eldadfux

RanadeepPolavarapu:
Going to work on this week.

CQ10:
มันทำงานยังไง

RanadeepPolavarapu:
Going to work on it this weekend.

CQ10:
คราบ

eldadfux:
@CQ10 please use english.

mbtamuli:
It's been 15 days since the last response. If no one is working on it, I'd like to take this up.

eldadfux:
@mbtamuli yeh I guess you can work on it. It's been 26 days with no updates.

You can reach me out on our Discord channel, feel free to ask me anything.

https://discord.gg/JbdEMd4

[...]

PR #5276, opened 2023-03-24
Helm chart

fpoussin:
## What does this PR do?

This is a Helm chart for deploying AppWrite on Kubernetes.
It's based of https://github.com/k3env/appwrite-helm.

## Test Plan

- Use Minikube to create a cluster: `minikube start --network-plugin=cni --cni=calico --addons ingress`
- Install InfluxDB: `helm -n monitoring install --create-namespace influxdb influxdata/influxdb`
- Create a key for secure storage: `openssl genrsa -out key.pem 2048`

Finally, install AppWrite: 
```bash
helm -n appwrite upgrade --install --create-namespace --set-file appwrite.sslKey=key.pem --dependency-update appwrite ./helm/
```
You can find the cluster's ip with `minikube ip`
Update your `/etc/hosts` with e.g: `192.168.49.2 appwrite.local`
Open `http://appwrite.local` with your browser, that's it.

Stuff that doesn't work:
- Usage stats
  - The usage service connects fine but I could not make it write data. (stats always at 0)
  - Doctor says everything is OK
- Executor
  - It would require interaction with the K8S API instead of using docker commands
  - Docker in Docker via Sysbox might work if installed on cluster.

Some more testing would be needed as I don't know the stack very well.
I originally used the SQL script bundled here (from `appwrite/mariadb`) but it seems like it's not needed.

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/24
- https://github.com/appwrite/appwrite/issues/364

## Checklist

### DIFF
--- a/helm/Chart.lock
+++ b/helm/Chart.lock
@@ -1,0 +1,12 @@
+dependencies:
+- name: mariadb
+  repository: https://charts.bitnami.com/bitnami
+  version: 11.5.5
+- name: redis
+  repository: https://charts.bitnami.com/bitnami
+  version: 17.9.0
+- name: telegraf
+  repository: https://helm.influxdata.com/
+  version: 1.8.26
+digest: sha256:3101feaa5de1854ff25673f059988104e28f2481a8c6ec0d434fb8a5e78076ad
+generated: "2023-03-24T12:30:07.256605462+01:00"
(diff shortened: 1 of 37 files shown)


stnguyen90:
@fpoussin thanks for creating this! 🙏🏼 Unfortunately, I don't think we want this included in the appwrite repo at the moment. It would be best if you created a separate repo and create a PR to https://github.com/appwrite/awesome-appwrite.

fpoussin:
Sounds good, I've created https://github.com/fpoussin/appwrite-helm

PR #5363, opened 2023-04-13
Fix auto-setting custom ID on nested documents

abnegate:
## What does this PR do?

- Fix auto-setting custom ID on nested documents

## Test Plan

- Added tests for nested ID not included

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/controllers/api/databases.php
+++ b/app/controllers/api/databases.php
@@ -2743,8 +2743,11 @@ App::post('/v1/databases/:databaseId/collections/:collectionId/documents')
                 if (empty($related)) {
                     continue;
                 }
-                if (!\is_array($related)) {
-                    $related = [$related];
+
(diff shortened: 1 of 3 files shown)


app/controllers/api/databases.php
```diff
@@ -2743,16 +2743,27 @@ function updateAttribute(
                 if (empty($related)) {
                     continue;
                 }
-                if (!\is_array($related)) {
-                    $related = [$related];
+
+                if (\is_array($related) && \array_values($related) === $related) {
```
stnguyen90, line 2747: If we're doing this here and after the for loop, maybe it would be good to store this into a descriptive variable?

stnguyen90 (approved):

PR #5377, opened 2023-04-14
Update description for _APP_FUNCTIONS_INACTIVE_THRESHOLD

gewenyu99:
## What does this PR do?

Update the description to reflect better what actually happens in the code/reflect real behavior.

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/config/variables.php
+++ b/app/config/variables.php
@@ -764,7 +764,7 @@ return [
             ],
             [
                 'name' => '_APP_FUNCTIONS_INACTIVE_THRESHOLD',
-                'description' => 'The minimum time a function can be inactive before it\'s container is shutdown and put to sleep. The default value is 60 seconds',
+                'description' => 'The minimum time a function has been inactive before it could be shut down and put to sleep. Containers are checked once every hour, and if they have been inactive for an interval longer than this threshold, they are put to sleep to free server resources. This feature intends to prevent the buildup of unused containers. Containers may remain active for longer than the interval before being put to sleep. If no value is provided, defaults to 60 seconds.',
                 'introduction' => '0.13.0',
                 'default' => '60',
                 'required' => false,


stnguyen90 (approved):

PineappleIOnic (approved):

app/config/variables.php
```diff
@@ -764,7 +764,7 @@
             ],
             [
                 'name' => '_APP_FUNCTIONS_INACTIVE_THRESHOLD',
-                'description' => 'The minimum time a function can be inactive before it\'s container is shutdown and put to sleep. The default value is 60 seconds',
+                'description' => 'The minimum time a function has been inactive before it could be shut down and put to sleep. Containers are checked once every hour, and if they have been inactive for an interval longer than this threshold, they are put to sleep to free server resources. This feature intends to prevent the buildup of unused containers. Containers may remain active for longer than the interval before being put to sleep. If no value is provided, defaults to 60 seconds.',
```
Meldiron, line 767: Let's remove "put to sleep". sound like pausing, while in reality we are destroying the container.

Meldiron: "Containers may remain active for longer than the interval before being put to sleep", let's explain why, if we can do that easily.

gewenyu99, line 767:
```suggestion
                'description' => 'The minimum time a function must be inactive before it can be shut down and cleaned up. This feature is intended to clean up unused containers. Containers may remain active for longer than the interval before being shut down, as Appwrite only cleans up unused containers every hour. If no value is provided, the default is 60 seconds.',
```

gewenyu99: @Meldiron How's this?

Meldiron: Looks amazing 😍

Meldiron (approved):

christyjacob4 (approved):

PR #5393, opened 2023-04-17
Update container images

Emiliaaah:
## What does this PR do?

This PR aims to update some images that should probably be updated once in a while.

The main reason for this PR is that I discovered that traefik was running on v2.7 which has a critical [CVE(-2022-37434)](https://dso.docker.com/cve/CVE-2022-37434)

## Test Plan

I updated the images in the docker-compose file and then played around a bit with the web-ui which still seemed to work fine.

## Related PRs and Issues

- None

## Checklist

## Changes

- [x] Update traefik
- [x] Update telegraf
- [x] Update redis
- [ ] Update influxdb?

### DIFF
--- a/app/views/install/compose.phtml
+++ b/app/views/install/compose.phtml
@@ -15,7 +15,7 @@ $image = $this->getParam('image', '');
 
 services:
   traefik:
-    image: traefik:2.7
+    image: traefik:2.9
     container_name: appwrite-traefik
     <<: *x-logging
     command:
--- a/app/views/install/compose.phtml
+++ b/app/views/install/compose.phtml
@@ -619,7 +619,7 @@ services:
     command: 'mysqld --innodb-flush-method=fsync'
 
   redis:
-    image: redis:7.0.4-alpine
+    image: redis:7.0.10-alpine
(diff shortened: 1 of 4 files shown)


harshsingh32 (approved):

christyjacob4:
@Emiliaaah The main reason we used a custom docker image for influx is such that we could initialise it with our own startup scripts. 

1. https://github.com/appwrite/docker-influxdb/blob/main/init-influxdb.sh
2. https://github.com/appwrite/docker-influxdb/blob/main/influxdb.conf
3. https://github.com/appwrite/docker-influxdb/blob/main/entrypoint.sh

However we are soon deprecating influx from our stack as we will be storing the usage data in our main MariaDB itself.

Emiliaaah:
@christyjacob4 Ah okay, that makes sense. I'll let influxdb stay the way it is then.

Something about this PR though, since it's been a while since these images have been updated, it might be a good idea to check the possibility of upgrading to newer versions every once in a while or say, every release, etc. as long as there is some check in place so we don't suddenly end up with 3 year old images.

PR #5395, opened 2023-04-18
Fix relationship filtering

stnguyen90:
## What does this PR do?

In the previous release, filtering was moved to the Appwrite layer rather than being done in the database. This resulted in some inconsistencies with the data returned. In particular, the `$total` ended up being higher than expected because some documents that shouldn't be counted were being counted.

Tasks:

- [x] Figure out what to do with `permissions` and `documentSecurity` in the `'database_' . $database->getInternalId()` collection
- [x] Update the Update Collection endpoint to update the `permissions` and `documentSecurity` in the metadata collection
- [x] Ensure all database tests pass
- [x] Ensure all tests pass
- [x] Ensure a user doesn't see teams they're not in
- [x] Ensure a user doesn't see other users' function executions
- [x] Add any additional relevant tests
   - [x] Test for non-existing database
   - [x] Test for non-existing collection
   - [x] Test for collection the user doesn't have access to
   - [x] Test abuse

## Test Plan

Update test cases

## Related PRs and Issues

Depends on:

- https://github.com/utopia-php/database/pull/262

## Checklist


app/controllers/api/databases.php
```diff
@@ -2885,12 +2885,16 @@ function updateAttribute(
...
         $collection = Authorization::skip(fn() => $dbForProject->getDocument('database_' . $database->getInternalId(), $collectionId));
 
-        if ($collection->isEmpty() || !$collection->getAttribute('enabled')) {
-            if (!($mode === APP_MODE_ADMIN && Auth::isPrivilegedUser(Authorization::getRoles()))) {
-                throw new Exception(Exception::COLLECTION_NOT_FOUND);
+        if (!($mode === APP_MODE_ADMIN && Auth::isPrivilegedUser(Authorization::getRoles()))) {
+            if (!$collection->getAttribute('documentSecurity', false)) {
+                $collection = $dbForProject->getDocument('database_' . $database->getInternalId(), $collectionId);
             }
         }
```
stnguyen90, line 2892: This actually changes the behavior a bit. Do we want to return 404 if the collection exists, but the user doesn't have access to it? I think the way it was before, list documents would return 200 with 0 documents.

stnguyen90: We're going to move forward with returning 404

PR #5436, opened 2023-04-27
Feat: Console OAuth & Cards

Meldiron:
## What does this PR do?

- Adds GitHub OAuth for console
- Adds cards generation

## Test Plan

- Manual QA

## Related PRs and Issues

x

## Checklist


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "app/console"]
 	path = app/console
 	url = https://github.com/appwrite/console
-	branch = 2.2.2
+	branch = feat/cloud-beta-card
```
christyjacob4, line 4:
```suggestion
	branch = cloud-current
```

.env
```diff
@@ -76,4 +76,6 @@ _APP_LOGGING_PROVIDER=
 _APP_LOGGING_CONFIG=
 _APP_REGION=default
 _APP_DOCKER_HUB_USERNAME=
-_APP_DOCKER_HUB_PASSWORD=
\ No newline at end of file
+_APP_DOCKER_HUB_PASSWORD=
+_APP_CONSOLE_GITHUB_SECRET=
```
christyjacob4, line 80: Add to Cloud CI

docker-compose.yml
```diff
@@ -165,6 +165,8 @@ services:
       - _APP_SMS_PROVIDER
       - _APP_SMS_FROM
       - _APP_REGION
+      - _APP_CONSOLE_GITHUB_APP_ID
```
christyjacob4, line 168: Add to stack file

app/controllers/api/avatars.php
```diff
@@ -49,11 +53,92 @@
...
+        ->file($data);
     unset($image);
 };
 
+$getUserGitHub = function (string $userId, Document $project, Database $dbForProject, Database $dbForConsole) {
+    try {
+        $user = Authorization::skip(fn () => $dbForConsole->getDocument('users', $userId));
+
+        $sessions = $user->getAttribute('sessions', []);
+        $session = $sessions[0] ?? new Document();
```
christyjacob4, line 65: shouldn't it throw an exception if there is no session?

app/controllers/api/avatars.php
```diff
@@ -49,11 +53,92 @@
...
+                        break;
+                    }
+
+                    $index++;
+                    \usleep(500000);
+                } while ($index < 10);
+            }
+        }
+
+        $oauth2 = new $className($appId, $appSecret, '', [], []);
```
christyjacob4, line 125:
```suggestion
```
it's being reinitialised ?

app/controllers/api/avatars.php
```diff
@@ -49,11 +53,92 @@
...
+
+        return [
+            'name' => $githubUser,
+            'id' => $githubId
+        ];
+    } catch (Exception $err) {
+        \var_dump($err->getMessage());
+        \var_dump($err->getTraceAsString());
+        \var_dump($err->getLine());
+        \var_dump($err->getFile());
```
christyjacob4, line 137: we should rethrow the exception so it can be handled by our normal exception handling flow

app/controllers/api/avatars.php
```diff
@@ -49,11 +53,92 @@
...
+        return [];
+    }
```
christyjacob4, line 139: You could get rid of this catch block and replace it with a `finally` block

christyjacob4 (approved):

app/controllers/api/avatars.php
Meldiron: Yes, on purpose, to lose all the user data from previous check that is intended for locking concurrent requests. By re-initializing we make sure that user is gone, and we can safely fetch new user data with our safe access token

app/controllers/api/avatars.php
Meldiron: It will end up throwing exception PROJECT_PROVIDER_UNSUPPORTED a little below.

app/controllers/api/avatars.php
Meldiron: I wanted to print logs. If someone complains, we can check what's wrong on cloud. Maybe there are some GitHub profile privacy options that would break something. maybe.

christyjacob4: You can just throw a 500 error and it will report it to sentry

PR #5437, opened 2023-04-27
Implement Migrations API

PineappleIOnic:
## What does this PR do?

Implement Migrations API

## Checklist


app/config/services.php
```diff
@@ -199,4 +199,17 @@
         'optional' => false,
         'icon' => '',
     ],
+    'imports' => [
+        'key' => 'imports',
+        'name' => 'Imports',
+        'subtitle' => 'The Imports service allows you to migrate third-party data to your Appwrite server.',
+        'description' => '/docs/services/imports.md',
```
eldadfux, line 206: We need to add this file with service description for the docs.

app/controllers/api/functions.php
```diff
@@ -845,6 +845,93 @@
...
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', '*/*')
+    ->label('sdk.methodType', 'location')
+    ->param('functionId', '', new UID(), 'Function ID.')
+    ->param('deploymentId', '', new UID(), 'Deployment ID.')
+    ->inject('response')
+    ->inject('request')
+    ->inject('dbForProject')
+    ->inject('deviceFunctions')
+    ->action(function (string $functionId, string $deploymentId, Response $response, Request $request, Database $dbForProject, Device $deviceFunctions) {
```
eldadfux, line 865: Do we also read this file in chunks from the imports worker? It is important to make sure we don't load the entire file to memory during migration.

composer.json
```diff
@@ -70,12 +71,17 @@
...
         "slickdeals/statsd": "3.1.0",
-        "webonyx/graphql-php": "14.11.*"
+        "webonyx/graphql-php": "14.11.*",
+        "psr/cache": "^1.0||^2.0"
     },
     "repositories": [
         {
             "url": "https://github.com/appwrite/runtimes.git",
             "type": "git"
+        },
```
eldadfux, line 81: Reminder to add a proper tag when the library is released.

app/config/collections.php
```diff
@@ -3580,6 +3580,139 @@
...
+                'size' => Database::LENGTH_KEY,
+                'signed' => true,
+                'required' => true,
+                'default' => [],
+                'array' => true,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('statusCounters'),
+                'type' => Database::VAR_STRING,
```
eldadfux, line 3635: Apply proper changes according to changes required in the response model. Both structures should be pretty similar.

app/controllers/api/imports.php
```diff
@@ -0,0 +1,403 @@
...
+        $deletes
+            ->setType(DELETE_TYPE_DOCUMENT)
+            ->setDocument($import);
+
+        $events->setParam('importId', $import->getId());
+
+        $response->noContent();
+    });
+
+App::post('/v1/imports/appwrite')
```
eldadfux, line 179: Let's add all CREATE operation at the start of the controller.

app/controllers/api/imports.php
```diff
@@ -0,0 +1,403 @@
...
+    ->param('database', '', new Text(512), "Source's Database Name")
+    ->param('username', '', new Text(512), "Source's Database Username")
+    ->param('password', '', new Text(512), "Source's Database Password")
+    ->param('port', 5432, new Integer(), "Source's Database Port", true)
+    ->inject('response')
+    ->inject('dbForProject')
+    ->inject('project')
+    ->inject('user')
+    ->inject('events')
+    ->action(function (array $resources, string $subdomain, string $region, string $adminSecret, string $database, string $username, string $password, int $port, Response $response, Database $dbForProject, Document $project, Document $user, Event $eventsInstance) {
```
eldadfux, line 369:
```suggestion
    ->action(function (array $resources, string $subdomain, string $region, string $adminSecret, string $database, string $username, string $password, int $port, Response $response, Database $dbForProject, Document $project, Document $user, Event $events) {
```

Should this work? why did we go with a different name?

app/controllers/api/imports.php
```diff
@@ -0,0 +1,403 @@
...
+    ->inject('dbForProject')
+    ->inject('project')
+    ->inject('user')
+    ->inject('events')
+    ->action(function (array $resources, string $serviceAccount, Response $response, Database $dbForProject, Document $project, Document $user, Event $eventsInstance) {
+        $import = $dbForProject->createDocument('imports', new Document([
+            '$id' => ID::unique(),
+            'status' => 'pending',
+            'stage' => 'init',
+            'source' => json_encode([
```
eldadfux, line 258: No need to use json_encode here, we already have a database filter we can auto apply for this. Also, wouldn't it make more sense to call this attribute credentials and save the source type in a separate attribute?

app/controllers/api/imports.php
```diff
@@ -0,0 +1,403 @@
...
+    ->groups(['api', 'imports'])
+    ->desc('Retry Import')
+    ->label('scope', 'imports.write')
+    ->label('event', 'imports.[importId].retry')
+    ->label('audits.event', 'import.retry')
+    ->label('audits.resource', 'imports/{request.importId}')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'imports')
+    ->label('sdk.method', 'retry')
+    ->label('sdk.description', '/docs/references/imports/retry-import.md')
```
eldadfux, line 104: 1. Not sure this one should be POST vs PATCH
2. Let's ensure we use descriptions that are consistent with other services
3. Have we added md files for the docs?

eldadfux (changes requested):

app/config/collections.php
```diff
@@ -3580,6 +3580,139 @@
...
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['source'],
+                'lengths' => [Database::LENGTH_KEY],
+                'orders' => [Database::ORDER_ASC],
+            ],
+            [
+                '$id' => '_key_statusCounters',
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['statusCounters'],
+                'lengths' => [Database::LENGTH_KEY],
```
fogelito, line 3704: statusCounters attribute size is 3000, Only 255 chars (Database::LENGTH_KEY) will be indexes.
Please double-check if it is enough for your queries, can be up to 768 MAX.

app/config/scopes.php
```diff
@@ -76,4 +76,10 @@
     'health.read' => [
         'description' => 'Access to read your project\'s health status',
     ],
+    'migrations.read' => [
+        'description' => 'Access to read your project\'s migration logs',
```
christyjacob4, line 80: migration logs or migrations?

docker-compose.yml
```diff
@@ -575,6 +575,39 @@ services:
...
+    container_name: appwrite-worker-migrations
+    restart: unless-stopped
+    image: appwrite-dev
+    networks:
+      - appwrite
+    volumes:
+      - ./app:/usr/src/code/app
+      - ./src:/usr/src/code/src
+      - ./tests:/usr/src/code/tests
+      - ./vendor:/usr/src/code/vendor
```
christyjacob4, line 590: Lets remove the vendor mount

docker-compose.yml
```diff
@@ -575,6 +575,39 @@ services:
...
+    depends_on:
+      - redis
```
christyjacob4, line 592: let's depend on the appwrite container instead.

app/controllers/api/functions.php
```diff
@@ -845,6 +845,93 @@
...
+
+        $path = $deployment->getAttribute('path', '');
+        if (!$deviceFunctions->exists($path)) {
+            throw new Exception(Exception::DEPLOYMENT_NOT_FOUND);
+        }
+
+        $response
+            ->setContentType('application/gzip')
+            ->addHeader('Expires', \date('D, d M Y H:i:s', \time() + (60 * 60 * 24 * 45)) . ' GMT') // 45 days cache
+            ->addHeader('X-Peak', \memory_get_peak_usage())
```
christyjacob4, line 889: whats this header used for ?

app/controllers/api/functions.php
```diff
@@ -845,6 +845,93 @@
...
+            ->addHeader('Content-Disposition', 'attachment; filename="deployment.tar.gz"')
```
christyjacob4, line 890: maybe we want to use `$deploymentID` as the filename ?

app/controllers/api/teams.php
```diff
@@ -405,6 +405,10 @@
             throw new Exception(Exception::GENERAL_SMTP_DISABLED);
         }
 
+        if (!$isPrivilegedUser && !$isAppUser && empty($url)) {
+            throw new Exception(Exception::GENERAL_ACCESS_FORBIDDEN, 'URL is required for non-privileged users');
```
christyjacob4, line 409:
```suggestion
            throw new Exception(Exception::GENERAL_ACCESS_FORBIDDEN, 'Param 'url' is required for non-privileged users');
```

src/Appwrite/Event/Migration.php
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setProject(Document $project): self
+    {
+        parent::setProject($project);
+
+        return $this;
+    }
```
christyjacob4, line 117: These methods are already implemented in the base class. So why do we need it here?

src/Appwrite/Utopia/Response/Model/Migration.php
```diff
@@ -0,0 +1,96 @@
...
+            ])
+            ->addRule('$updatedAt', [
+                'type' => self::TYPE_DATETIME,
+                'description' => 'Variable creation date in ISO 8601 format.',
+                'default' => '',
+                'example' => self::TYPE_DATETIME_EXAMPLE,
+            ])
+            ->addRule('status', [
+                'type' => self::TYPE_STRING,
+                'description' => 'Migration status.',
```
christyjacob4, line 33:
```suggestion
                'description' => 'Migration status (pending, ....)',
```
would be good to add the possible values in the description

src/Appwrite/Utopia/Response/Model/Migration.php
```diff
@@ -0,0 +1,96 @@
...
+                'default' => '',
+                'example' => 'pending',
+            ])
+            ->addRule('stage', [
+                'type' => self::TYPE_STRING,
+                'description' => 'Migration stage.',
```
christyjacob4, line 39:
```suggestion
                'description' => 'Migration stage ( init, ... )',
```
Please mention all the stages in the description

src/Appwrite/Utopia/Response/Model/ImportValidationError.php
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * Get Type
+     *
+     * @return string
+     */
+    public function getType(): string
+    {
+        return Response::MODEL_IMPORT_VALIDATION_ERROR;
+    }
+}
```
christyjacob4, line 28: Why do we need a new model here if it's not adding anything over the `Any` base class ? I think its better to reuse that

app/workers/migrations.php
```diff
@@ -0,0 +1,314 @@
...
+                return new Firebase(
+                    json_decode($source['serviceAccount'], true),
+                );
+                break;
+            case 'supabase':
+                return new Supabase(
+                    $source['endpoint'],
+                    $source['apiKey'],
+                    $source['databaseHost'],
+                    "postgres",
```
christyjacob4, line 91: is this the database name ?

app/workers/migrations.php
```diff
@@ -0,0 +1,314 @@
...
+                    $source['region'],
+                    $source['adminSecret'],
+                    $source['database'],
+                    $source['username'],
+                    $source['password'],
+                    $source['port'],
+                );
+                break;
+            case 'appwrite':
+                return new Appwrite($source['projectId'], str_starts_with($source['endpoint'], 'http://localhost/v1') ? 'http://appwrite/v1' : $source['endpoint'], $source['apiKey']);
```
christyjacob4, line 109: how does the http://appwrite resolve? Don't we have to register this with the system ?

app/workers/migrations.php
```diff
@@ -0,0 +1,314 @@
...
+                'databases.write',
+                'collections.read',
+                'collections.write',
+                'documents.read',
+                'documents.write',
+                'buckets.read',
+                'buckets.write',
+                'files.read',
+                'files.write',
+
```
christyjacob4, line 186:
```suggestion
```

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+        if ($migration->isEmpty()) {
+            throw new Exception(Exception::MIGRATION_NOT_FOUND, 'Migration not found', 404);
+        }
+
+        $response->dynamic($migration, Response::MODEL_MIGRATION);
+    });
+
+App::post('/v1/migrations/:migrationId')
+    ->groups(['api', 'migrations'])
+    ->desc('Retry Migration')
```
christyjacob4, line 100: POST endpoints are usually to create migrations

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+    ->label('scope', 'migrations.write')
+    ->label('event', 'migrations.[migrationId].retry')
+    ->label('audits.event', 'migration.retry')
+    ->label('audits.resource', 'migrations/{request.migrationId}')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'migrations')
+    ->label('sdk.method', 'retry')
+    ->label('sdk.description', '/docs/references/migrations/retry-migration.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
```
christyjacob4, line 109: the status code should be 202 Accepted rather than 200 OK

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+    ->inject('response')
+    ->inject('dbForProject')
+    ->inject('deletes')
+    ->inject('events')
+    ->action(function (string $migrationId, Response $response, Database $dbForProject, Delete $deletes, Event $events) {
+
+        $migration = $dbForProject->getDocument('migrations', $migrationId);
+
+        if ($migration->isEmpty()) {
+            throw new Exception(Exception::MIGRATION_NOT_FOUND, 'Migration not found', 404);
```
christyjacob4, line 167:
```suggestion
            throw new Exception(Exception::MIGRATION_NOT_FOUND);
```
Check this for all exceptions.. If the type MIGRATION_NOT_FOUND has a message, we dont need to use the same message and code again unless we want to over write it

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+        }
+
+        if (!$dbForProject->deleteDocument('migrations', $migration->getId())) {
+            throw new Exception(Exception::GENERAL_SERVER_ERROR, 'Failed to remove migration from DB', 500);
+        }
+
+        $deletes
+            ->setType(DELETE_TYPE_DOCUMENT)
+            ->setDocument($migration);
```
christyjacob4, line 176: Why are we sending the deleted document to the deletes worker again ? Isnt it already deleted on L170 ?

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+    ->groups(['api', 'migrations'])
+    ->desc('Migrate Appwrite Data')
+    ->label('scope', 'migrations.write')
+    ->label('event', 'migrations.create')
+    ->label('audits.event', 'migration.create')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'migrations')
+    ->label('sdk.method', 'migrateAppwrite')
+    ->label('sdk.description', '/docs/references/migrations/migration-appwrite.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
```
christyjacob4, line 193: The status code here should be 202 accepted

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+        // Trigger Transfer
+        $event = new Migration();
+        $event
+            ->setMigration($migration)
+            ->setProject($project)
+            ->setUser($user)
+            ->trigger();
+
+        $response
+            ->setStatusCode(Response::STATUS_CODE_CREATED)
```
christyjacob4, line 233: The status code here should be 202 accepted

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+    });
+
+App::post('/v1/migrations/appwrite/report')
+    ->groups(['api', 'migrations'])
+    ->desc('Generate a report on Appwrite Data')
+    ->label('scope', 'migrations.write')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'migrations')
+    ->label('sdk.method', 'generateAppwriteReport')
+    ->label('sdk.description', '/docs/references/migrations/migration-appwrite-report.md')
```
christyjacob4, line 244: Please add the docs... All the markdown files for the migration service is missing at the moment

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+    ->inject('response')
+    ->inject('dbForProject')
+    ->inject('project')
+    ->inject('user')
+    ->action(function (array $resources, string $endpoint, string $projectID, string $key, Response $response) {
+
+        $appwrite = new Appwrite($projectID, $endpoint, $key);
+
+        $response
+            ->setStatusCode(Response::STATUS_CODE_CREATED)
```
christyjacob4, line 261: this should be a 200 OK 
We have not created any resource in the database

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+    ->groups(['api', 'migrations'])
+    ->desc('Migrate Firebase Data')
+    ->label('scope', 'migrations.write')
+    ->label('event', 'migrations.create')
+    ->label('audits.event', 'migration.create')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'migrations')
+    ->label('sdk.method', 'migrateFirebase')
+    ->label('sdk.description', '/docs/references/migrations/migration-firebase.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
```
christyjacob4, line 276: The status code here should be 202 accepted

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+        // Trigger Transfer
+        $event = new Migration();
+        $event
+            ->setMigration($migration)
+            ->setProject($project)
+            ->setUser($user)
+            ->trigger();
+
+        $response
+            ->setStatusCode(Response::STATUS_CODE_CREATED)
```
christyjacob4, line 312: The status code here should be 202 accepted

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+    ->inject('project')
+    ->inject('user')
+    ->inject('events')
+    ->action(function (array $resources, string $serviceAccount, Response $response, Database $dbForProject, Document $project, Document $user, Event $eventsInstance) {
+        $migration = $dbForProject->createDocument('migrations', new Document([
+            '$id' => ID::unique(),
+            'status' => 'pending',
+            'stage' => 'init',
+            'source' => json_encode([
+                'type' => 'firebase',
```
christyjacob4, line 292: Why dont we have constants in the migration library to represent these strings?

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+    ->label('audits.event', 'migration.create')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'migrations')
+    ->label('sdk.method', 'migrateFirebase')
+    ->label('sdk.description', '/docs/references/migrations/migration-firebase.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_MIGRATION)
+    ->param('resources', [], new ArrayList(new WhiteList(Firebase::getSupportedResources())), 'List of resources to migrate')
+    ->param('serviceAccount', '', new Text(512), "Source's Service Account")
```
christyjacob4, line 280: Source's service account sounds vague. This is a firebase specific endpoint so lets make the description firebase specific

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+            ->trigger();
+
+        $response
+            ->setStatusCode(Response::STATUS_CODE_CREATED)
+            ->dynamic($migration, Response::MODEL_MIGRATION);
+    });
+
+App::post('/v1/migrations/firebase/report')
+    ->groups(['api', 'migrations'])
+    ->desc('Generate a report on Firebase Data')
```
christyjacob4, line 318: similar feedback as the generate appwrite report endpoint

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+            ->setProject($project)
+            ->setUser($user)
+            ->trigger();
+
+        $response
+            ->setStatusCode(Response::STATUS_CODE_CREATED)
+            ->dynamic($migration, Response::MODEL_MIGRATION);
+    });
+
+App::post('/v1/migrations/appwrite/report')
```
christyjacob4, line 237: This should be a GET endpoint as we're not creating any resources in here

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_MIGRATION)
+    ->param('resources', [], new ArrayList(new WhiteList(Supabase::getSupportedResources(), true)), 'List of resources to migrate')
+    ->param('endpoint', '', new URL(), "Source's Supabase Endpoint")
+    ->param('apiKey', '', new Text(512), "Source's API Key")
+    ->param('databaseHost', '', new Text(512), "Source's Database Host")
+    ->param('username', '', new Text(512), "Source's Database Username")
+    ->param('password', '', new Text(512), "Source's Database Password")
+    ->param('port', 5432, new Integer(), "Source's Database Port", true)
```
christyjacob4, line 365: Similar to the feedback on the appwrite and firebase endpoints

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+        // Trigger Transfer
+        $event = new Migration();
+        $event
+            ->setMigration($migration)
+            ->setProject($project)
+            ->setUser($user)
+            ->trigger();
+
+        $response
+            ->setStatusCode(Response::STATUS_CODE_CREATED)
```
christyjacob4, line 402: 202 Accepted

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+            ->dynamic($migration, Response::MODEL_MIGRATION);
+    });
+
+
+App::post('/v1/migrations/supabase/report')
+    ->groups(['api', 'migrations'])
+    ->desc('Generate a report on Supabase Data')
```
christyjacob4, line 409: Similar to the feedback on the appwrite and firebase endpoints

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+    ->inject('events')
+    ->action(function (array $resources, string $endpoint, string $apiKey, string $databaseHost, string $username, string $password, int $port, Response $response, Database $dbForProject, Document $project, Document $user, Event $eventsInstance) {
+        try {
+            $supabase = new Supabase($endpoint, $apiKey, $databaseHost, 'postgres', $username, $password, $port);
+
+            $response
+                ->setStatusCode(Response::STATUS_CODE_CREATED)
+                ->dynamic(new Document($supabase->report($resources)), Response::MODEL_MIGRATION_REPORT);
+        } catch (\Exception $e) {
+            var_dump($e->getMessage());
```
christyjacob4, line 438: leftovers

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+
+            $response
+                ->setStatusCode(Response::STATUS_CODE_BAD_REQUEST)
+                ->json([
+                    'status' => 'error',
+                    'message' => $e->getMessage()
+                ]);
+        }
```
christyjacob4, line 446: We also don't handle errors this way. We simply throw them and let the global error handler capture them

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+    ->param('adminSecret', '', new Text(512), "Source's Admin Secret")
+    ->param('database', '', new Text(512), "Source's Database Name")
+    ->param('username', '', new Text(512), "Source's Database Username")
+    ->param('password', '', new Text(512), "Source's Database Password")
+    ->param('port', 5432, new Integer(), "Source's Database Port", true)
+    ->inject('response')
+    ->inject('dbForProject')
+    ->inject('project')
+    ->inject('user')
+    ->inject('events')
```
christyjacob4, line 474: Similar to the feedback on the appwrite and firebase endpoints

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+        // Trigger Transfer
+        $event = new Migration();
+        $event
+            ->setMigration($migration)
+            ->setProject($project)
+            ->setUser($user)
+            ->trigger();
+
+        $response
+            ->setStatusCode(Response::STATUS_CODE_CREATED)
```
christyjacob4, line 507: 202 Accepted

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+    ->param('adminSecret', '', new Text(512), "Source's Admin Secret")
+    ->param('database', '', new Text(512), "Source's Database Name")
+    ->param('username', '', new Text(512), "Source's Database Username")
+    ->param('password', '', new Text(512), "Source's Database Password")
+    ->param('port', 5432, new Integer(), "Source's Database Port", true)
+    ->inject('response')
+    ->inject('dbForProject')
+    ->inject('project')
+    ->inject('user')
+    ->inject('events')
```
christyjacob4, line 536: Similar to the feedback on the appwrite and firebase endpoints

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+                ->dynamic(new Document($nhost->report($resources)), Response::MODEL_MIGRATION_REPORT);
+        } catch (\Exception $e) {
+            var_dump($e->getMessage());
+
+            $response
+                ->setStatusCode(Response::STATUS_CODE_BAD_REQUEST)
+                ->json([
+                    'status' => 'error',
+                    'message' => $e->getMessage()
+                ]);
```
christyjacob4, line 552: left overs and, this is not how we handle errors in the endpoint. Unless there is a very specific error we want to capture... 
else we just throw them and let the global handler do the rest

christyjacob4 (changes requested):
Please address the comments

app/workers/migrations.php
PineappleIOnic: It resolves internally within the Docker, we have to rewrite it since localhost doesn't work

app/controllers/api/imports.php
PineappleIOnic: Mainly just to simplify the variable name to just one word, we can roll it back if you think it's too vague?

app/workers/migrations.php
PineappleIOnic: Yep, from what I've observed from their database. The default database is always default to "postgres"

app/controllers/api/migrations.php
PineappleIOnic: Hm, yes I agree. CREATE may be a better HTTP call to use. Will update accordingly

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,554 @@
...
+            ->trigger();
+
+        $response
+            ->setStatusCode(Response::STATUS_CODE_CREATED)
+            ->dynamic($migration, Response::MODEL_MIGRATION);
+    });
+
+App::post('/v1/migrations/appwrite/report')
+    ->groups(['api', 'migrations'])
+    ->desc('Generate a report on Appwrite Data')
```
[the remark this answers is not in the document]
PineappleIOnic: The data hasn't been migrated yet at this stage, this is just used to check a source before migration to give us a report on resources/version/etc...

app/controllers/api/imports.php
PineappleIOnic: I'm not sure I understand what you mean by a CREATE operation?

app/controllers/api/migrations.php
PineappleIOnic: Sorry, this was a fallback while I was trying to debug the issue with messages not being sent

app/controllers/api/migrations.php
PineappleIOnic: Sorry, the comment has moved from where it was originally. Could you please correct me on what part of the code this is referrencing?

app/controllers/api/functions.php
PineappleIOnic: I'm not entirely sure, it was taken from the File Download route I assume the Expires is to cache it but as for the X-Peak I will have to ask Damodar.

app/controllers/api/migrations.php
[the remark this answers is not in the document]
abnegate: I think we should still describe this as `Get a report on Appwrite data` for consistency

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,824 @@
...
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_MIGRATION_REPORT)
+    ->param('resources', [], new ArrayList(new WhiteList(Appwrite::getSupportedResources())), 'List of resources to migrate')
+    ->param('endpoint', '', new URL(), "Source's Appwrite Endpoint")
+    ->param('projectID', '', new Text(512), "Source's Project ID")
+    ->param('key', '', new Text(512), "Source's API Key")
+    ->inject('response')
+    ->inject('dbForProject')
+    ->inject('project')
+    ->inject('user')
```
abnegate, line 254: These aren't being injected in the action function signature, please double check all the injections/parameters and function signatures

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,824 @@
...
+    ->label('sdk.namespace', 'migrations')
+    ->label('sdk.method', 'getAppwriteReport')
+    ->label('sdk.description', '/docs/references/migrations/migration-appwrite-report.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_MIGRATION_REPORT)
+    ->param('resources', [], new ArrayList(new WhiteList(Appwrite::getSupportedResources())), 'List of resources to migrate')
+    ->param('endpoint', '', new URL(), "Source's Appwrite Endpoint")
+    ->param('projectID', '', new Text(512), "Source's Project ID")
+    ->param('key', '', new Text(512), "Source's API Key")
```
abnegate, line 250: Let's use single quotes where we're not interpolating, applies a few places in this controller

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,824 @@
...
+
+App::post('/v1/migrations/appwrite')
+    ->groups(['api', 'migrations'])
+    ->desc('Migrate Appwrite Data')
+    ->label('scope', 'migrations.write')
+    ->label('event', 'migrations.create')
+    ->label('audits.event', 'migration.create')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'migrations')
+    ->label('sdk.method', 'migrateAppwrite')
```
abnegate, line 190: Let's change the POST method names to `create[SOURCE]Migration`

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,824 @@
...
+        // $event = new Migration();
+        // $event
+        //     ->setMigration($migration)
+        //     ->setProject($project)
+        //     ->setUser($user)
+        //     ->trigger();
+
+        // $response
+        //     ->setStatusCode(Response::STATUS_CODE_ACCEPTED)
+        //     ->dynamic($migration, Response::MODEL_MIGRATION);
```
abnegate, line 399: Do we need this code or can it be removed?

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,824 @@
...
+            ->setProject($project)
+            ->setUser($user)
+            ->trigger();
+
+        $response
+            ->setStatusCode(Response::STATUS_CODE_ACCEPTED)
+            ->dynamic($migration, Response::MODEL_MIGRATION);
+    });
+
+App::post('/v1/migrations/firebaseOAuth')
```
abnegate, line 317: We should use single words in the path, let's move this to `/v1/migrations/firebase/oauth`

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,824 @@
...
+            'total' => count($output),
+        ]), Response::MODEL_MIGRATION_FIREBASE_PROJECT_LIST);
+    });
+
+App::get('/v1/migrations/firebase/deauthorize')
+    ->desc('Revoke Appwrite\'s authorization to access Firebase Projects')
+    ->groups(['api', 'migrations'])
+    ->label('scope', 'public')
+    ->label('sdk.namespace', 'migrations')
+    ->label('sdk.method', 'firebaseDeauthorize')
```
abnegate, line 609: Let's change this to `deleteFirebaseAuth` to match the `createFirebaseAuth` method

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,824 @@
...
+            ->setStatusCode(Response::STATUS_CODE_ACCEPTED)
+            ->dynamic($migration, Response::MODEL_MIGRATION);
+    });
+
+App::get('/v1/migrations/nhost/report')
+    ->groups(['api', 'migrations'])
+    ->desc('Generate a report on NHost Data')
+    ->label('scope', 'migrations.write')
+    // ->label('event', 'migrations.report')
+    // ->label('audits.event', 'migration.report')
```
abnegate, line 793: Let's remove this if they're unused, the same applies to other source's endpoints

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,824 @@
...
+use Utopia\Validator\ArrayList;
+use Utopia\Validator\Host;
+use Utopia\Validator\Integer;
+use Utopia\Validator\Text;
+use Utopia\Validator\URL;
+use Utopia\Validator\WhiteList;
+
+include_once __DIR__ . '/../shared/api.php';
+
+App::get('/v1/migrations')
```
abnegate, line 31: Let's reorder the routes so that we have them ordered like create, list, read, update, delete (post, get, get, patch, put, delete) for each resource type. Take a look at the other controllers for an example

abnegate (commented):
Looking good, just a few comments on style and consistency. Let's merge in 1.4.x to bring it up to date and resolve conflicts as well

app/controllers/api/migrations.php
PineappleIOnic: This code was commented out as I was still working on the code that generates a service account and didn't want to start a transfer everytime

PineappleIOnic: It will be uncommented once I'm done with that

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,824 @@
...
+    ->label('sdk.response.model', Response::MODEL_MIGRATION)
+    ->param('resources', [], new ArrayList(new WhiteList(Appwrite::getSupportedResources())), 'List of resources to migrate')
+    ->param('endpoint', '', new URL(), "Source's Appwrite Endpoint")
+    ->param('projectId', '', new UID(), "Source's Project ID")
+    ->param('apiKey', '', new Text(512), "Source's API Key")
+    ->inject('response')
+    ->inject('dbForProject')
+    ->inject('project')
+    ->inject('user')
+    ->inject('events')
```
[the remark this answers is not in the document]
PineappleIOnic: No, it's being used by `$eventsInstance` to fire off the request to the migrations worker, will rename to $events to make it a little clearer what the injection is doing

app/config/collections.php
PineappleIOnic: We don't perform any queries on this attribute, we may be best off just not having a index for it

PineappleIOnic:
Hm, I wonder where all those changes came from. It's rolling back loads

PineappleIOnic:
Due to the weird collision issues that occured, I've gone ahead and move to a new branch with a clean slate based off 1.4.x. Please review this PR instead: https://github.com/appwrite/appwrite/pull/5938

PR #5442, opened 2023-04-27
fix: project deletion leftovers

TorstenDittmann:
## What does this PR do?

- uses the `projectInternalId` to fetch platforms
- removes domains, keys and platforms with the deletes worker if project is deleted

## Test Plan

- manual

## Checklist

### DIFF
--- a/app/controllers/api/projects.php
+++ b/app/controllers/api/projects.php
@@ -1237,7 +1237,7 @@ App::get('/v1/projects/:projectId/platforms')
         }
 
         $platforms = $dbForConsole->find('platforms', [
-            Query::equal('projectId', [$project->getId()]),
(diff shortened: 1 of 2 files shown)


app/controllers/api/projects.php
```diff
@@ -1237,7 +1237,7 @@
         }
 
         $platforms = $dbForConsole->find('platforms', [
-            Query::equal('projectId', [$project->getId()]),
+            Query::equal('projectInternalId', [$project->getInternalId()]),
```
stnguyen90, line 1240: So the internal project ID is different even if a project is recreated with the same ID so this would prevent any old platform from ever being fetched, right?

stnguyen90 (approved):

PR #5460, opened 2023-04-29
fix: throw correct exception when project ID already exists

christyjacob4:
## What does this PR do?

Throw the correct exception when creating a project with a requested ID already exists 

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/config/errors.php
+++ b/app/config/errors.php
@@ -492,7 +492,7 @@ return [
     Exception::PROJECT_ALREADY_EXISTS => [
         'name' => Exception::PROJECT_ALREADY_EXISTS,
         'description' => 'Project with the requested ID already exists.',
-        'code' => 403,
+        'code' => 409,
     ],
     Exception::PROJECT_UNKNOWN => [
         'name' => Exception::PROJECT_UNKNOWN,
--- a/app/config/errors.php
+++ b/app/config/errors.php
@@ -489,6 +489,11 @@ return [
         'description' => 'Project with the requested ID could not be found. Please check the value of the X-Appwrite-Project header to ensure the correct project ID is being used.',
         'code' => 404,
     ],
+    Exception::PROJECT_ALREADY_EXISTS => [
+        'name' => Exception::PROJECT_ALREADY_EXISTS,
+        'description' => 'Project with the requested ID already exists.',
+        'code' => 403,
+    ],
     Exception::PROJECT_UNKNOWN => [
         'name' => Exception::PROJECT_UNKNOWN,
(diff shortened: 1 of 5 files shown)


app/config/errors.php
```diff
@@ -489,6 +489,11 @@
         'description' => 'Project with the requested ID could not be found. Please check the value of the X-Appwrite-Project header to ensure the correct project ID is being used.',
         'code' => 404,
     ],
+    Exception::PROJECT_ALREADY_EXISTS => [
+        'name' => Exception::PROJECT_ALREADY_EXISTS,
+        'description' => 'Project with the requested ID already exists.',
+        'code' => 403,
```
stnguyen90, line 495: We use 409 for others so maybe use the same for this.

stnguyen90 (changes requested):

app/config/errors.php
christyjacob4: Yes true but @TorstenDittmann suggested using 403. Which I feel is also relevant in this case 🤔

TorstenDittmann: should be 409, I might have remembered the wrong one when I mentioned it :)

tests/e2e/Services/Projects/ProjectsConsoleClientTest.php
```diff
@@ -98,7 +99,37 @@ public function testCreateProject(): array
...
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()), [
+            'projectId' => $projectId,
+            'name' => 'Project Duplicate',
+            'teamId' => $teamId,
+            'region' => 'default'
+        ]);
+
+        $this->assertEquals(403, $response['headers']['status-code']);
+        $this->assertEquals(403, $response['body']['code']);
```
christyjacob4, line 130:
```suggestion
        $this->assertEquals(409, $response['headers']['status-code']);
        $this->assertEquals(409, $response['body']['code']);
```

phpunit.xml
```diff
@@ -6,7 +6,7 @@
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
          processIsolation="false"
-         stopOnFailure="false"
+         stopOnFailure="true"
```
christyjacob4, line 9:
```suggestion
         stopOnFailure="false"
```

app/config/errors.php
christyjacob4, line 495:
```suggestion
        'code' => 409,
```

christyjacob4: Okay 👍  updated it

stnguyen90 (approved):
LGTM!

This is what the error looks like now:

image

TorstenDittmann (approved):

PR #5514, opened 2023-05-07
remove audit userInternalId

fogelito:
Remove the audit userInternalId dependency and use the current userId attribute.


stnguyen90:
@fogelito looks like there's a merge conflict now.

PR #5538, opened 2023-05-15
Feat dotnet sdk

abnegate:
## What does this PR do?

- Update .NET platform and SDK generator
- Add .NET examples

## Test Plan

N/A

## Related PRs and Issues

N/A

## Checklist


app/config/platforms.php
```diff
@@ -357,12 +357,12 @@
             [
                 'key' => 'dotnet',
                 'name' => '.NET',
-                'version' => '2.0.0',
+                'version' => '1.0.0',
```
lohanidamodar, line 360: Are we planning for 1.0? we are releasing as beta right?

lohanidamodar (commented):
Looks good, just a query.

app/config/platforms.php
abnegate: Updated to bump minor version from latest release on NuGet :+1:

lohanidamodar (commented):
looks good but why are we regenerating specs for SDK release?

abnegate:
> looks good but why are we regenerating specs for SDK release?

@lohanidamodar Just habit 😅 nothing in them had changed but there's always a diff because it's a new file. Have reverted now 👍

lohanidamodar (approved):

Issue #5542, opened 2023-05-16
🐛 Bug Report: x-appwrite-id should disallow unique()


Suven-p:
### 👟 Reproduction steps

When passing `unique()` to `x-appwrite-id` header with CreateFile endpoint, the request completes successfully but breaks stuff on appwrite's side.

### 👍 Expected behavior

Return an error.

### 👎 Actual Behavior

Request completes successfully. The uploaded file can't be accessed, modified or deleted.

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

stnguyen90:
@Suven-p, thanks for raising this issue! 🙏🏼 It looks like we should add some validation to prevent malformed file ids passed in the x-appwrite-id header:

https://github.com/appwrite/appwrite/blob/953485299a2a4040406cfec290a68f69187ec5e8/app/controllers/api/storage.php#L447

Suven-p:
@stnguyen90 Could you assign this to me?

PR #5550, opened 2023-05-17
Validate value of x-appwrite-id header

Suven-p:
## What does this PR do?

validate the value of x-appwrite-id header

## Test Plan

added test to e2e suite
manually test using REST API

## Related PRs and Issues

Fix https://github.com/appwrite/appwrite/issues/5542

## Checklist


Suven-p:
Does this require tests? and where do they go, I see that there isn't github actions configured for the tests

Suven-p:
@stnguyen90 Added the test. Invalid file size validation uses a short message
https://github.com/appwrite/appwrite/blob/5108ab7b061ffb1846b1cbdbd3077110e026482c/app/controllers/api/storage.php#L480
instead of the descriptive one at
https://github.com/appwrite/appwrite/blob/5108ab7b061ffb1846b1cbdbd3077110e026482c/app/config/errors.php#L291-L295
and the test:
https://github.com/appwrite/appwrite/blob/5108ab7b061ffb1846b1cbdbd3077110e026482c/tests/e2e/Services/Storage/StorageBase.php#L198

Should I do the same? The last commit throws and tests for the descriptive version of error message:
https://github.com/appwrite/appwrite/blob/582698bd03235a7ee1b458865bd0b32fa368d369/app/config/errors.php#L321-L325

stnguyen90 (changes requested):
Would you please fix the linting error?

stnguyen90 (approved):
LGTM!

app/controllers/api/storage.php
```diff
@@ -449,6 +449,11 @@
                 throw new Exception(Exception::STORAGE_INVALID_CONTENT_RANGE);
             }
 
+            $idValidator = new UID();
+            if (!$idValidator->isValid($request->getHeader('x-appwrite-id'))) {
```
lohanidamodar, line 453: x-appwrite-id is already assigned to `$fileId` why are we not using that instead?
Also, in the chunked upload, the first request will not have `x-appwrite-id` header, will that not cause error to throw?

lohanidamodar (commented):
Great work. Looks good. Just a small query before I approve.

app/controllers/api/storage.php
Suven-p: > x-appwrite-id is already assigned to $fileId why are we not using that instead

because the error message specifically mentions that the value of `x-appwrite-id` is invalid. Using `$fileId` may throw an error for invalid value of `$fileId`. I am not sure where the value of $fileId is validated so I might be wrong.

> the first request will not have x-appwrite-id header

This should work because the default value for `getHeader` is empty string
https://github.com/utopia-php/swoole/blob/2949a1bd93f5c772fc7de974e6b9da01106c704a/src/Swoole/Request.php#L270
and the validator at 
https://github.com/utopia-php/database/blob/d821d1f18dfe83b0350ed208be3658764aec99d7/src/Database/Validator/Key.php#L7
considers empty string to be valid https://onlinephp.io/c/21d30

lohanidamodar (approved):

PR #5551, opened 2023-05-18
Separate OAuth2 data out of Sessions and into User Identities

stnguyen90:
## What does this PR do?

#### New User - New Session - No Matching Identities

image

#### New User - New Session - Matching provider identity

No new user or identity created. Only new session.

Before:

image

After:

image

#### New Session - Matching Another User's Email Identity

If there is an identity with the same email already, there should be an error:

Before:

image

Error:

image

#### Existing User - No Identities - New Session

This should create a new session successfully for backwards compatibility on mobile platforms where we still need to make sure an existing session is used.

Before:

image

After:

image

#### Existing (matching) User - No Identies - Existing Session

This should create a new identity:

Before:

image

After:

image

#### Existing (non-matching) User - No Identities - New Session

Trying to connect an identity where providerEmail matches another user should result in error.

Before:

image

Error:

image

#### Deleting a User Deletes the Identities

Delete job deleted 64b1c0422411984aa897:

image

Which was the identity:

image

## Related PRs and Issues

* https://github.com/appwrite/rfc/pull/57

## Checklist


app/config/collections.php
```diff
@@ -1870,6 +1870,129 @@
...
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => ['encrypt'],
+            ],
+        ],
+        'indexes' => [
+            [
+                '$id' => ID::custom('_key_provider_providerUid'),
```
eldadfux, line 1980: Let's ask @fogelito to review structure and indexes.

app/controllers/api/users.php
```diff
@@ -47,6 +47,14 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
         $email = \strtolower($email);
     }
 
+    // Makes sure this email is not already used in another identity
+    $identityWithMatchingEmail = $dbForProject->findOne('identities', [
+        Query::equal('providerEmail', [$email]),
```
eldadfux, line 52: Are we making sure to attach only verified emails? Otherwise anyone could fake an email in a 3rd party provider and gain access to it on Appwrite.

src/Appwrite/Auth/OAuth2.php
```diff
@@ -195,8 +204,21 @@ protected function request(string $method, string $url = '', array $headers = []
         // Send the request & save response to $response
         $response = \curl_exec($ch);
 
+        $code = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+
         \curl_close($ch);
 
+        if ($code != 200) {
+            $dump = function($value) {
```
eldadfux, line 212: Reminder, remove debugs.

app/controllers/api/account.php
```diff
@@ -98,6 +99,14 @@
             }
         }
 
+        // Makes sure this email is not already used in another identity
+        $identityWithMatchingEmail = $dbForProject->findOne('identities', [
```
eldadfux, line 103: Should a user be able to log in with an email associated solely with the provider that is different than the main email?

Are we blocking users to attach OAuth providers with emails associated to other users?

app/workers/deletes.php
```diff
@@ -399,6 +399,11 @@ protected function deleteUser(Document $document, string $projectId): void
         $this->deleteByGroup('tokens', [
             Query::equal('userId', [$userId])
         ], $this->getProjectDB($projectId));
+
+        // Delete identities
+        $this->deleteByGroup('identities', [
+            Query::equal('userId', [$userId])
```
eldadfux, line 405: Use internalId for the deletion of child documents.

src/Appwrite/Auth/OAuth2/Exception.php
```diff
@@ -0,0 +1,41 @@
...
+            401 => AppwriteException::USER_OAUTH2_UNAUTHORIZED,
+            default => AppwriteException::GENERAL_SERVER_ERROR
+        };
+
+        parent::__construct($type, $this->message, $code, $previous);
+    }
+
+    public function getError() : string
+    {
+        return $this->error;
```
eldadfux, line 34: Why aren't we using the native getMessage interface?

app/controllers/api/users.php
stnguyen90: We're actually not really checking verified at all today because not all providers verify

app/controllers/api/account.php
stnguyen90: > Should a user be able to log in with an email associated solely with the provider that is different than the main email?

Using OAuth2, If the email is tied to an existing user via an identity, they won't be able to log in. That's what this check does:

https://github.com/appwrite/appwrite/blob/8fe688a95253b2b78126b3607fbee4ad60f328be/app/controllers/api/account.php#L564-L570

Using OAuth2, If the email is not associated with an existing identity or user, a new user will be created.

Using Create Email Session, a user must use their main user email and cannot use an identity email.

> Are we blocking users to attach OAuth providers with emails associated to other users?

Yes, here:

https://github.com/appwrite/appwrite/blob/8fe688a95253b2b78126b3607fbee4ad60f328be/app/controllers/api/account.php#L500-L511

src/Appwrite/Auth/OAuth2/Exception.php
stnguyen90: This is the error from the oauth2 provider. The message will include error and error_description.

See https://datatracker.ietf.org/doc/html/rfc6749#section-5.2

app/config/collections.php
```diff
@@ -1870,6 +1870,129 @@
...
+                'type' => Database::INDEX_UNIQUE,
+                'attributes' => ['provider', 'providerUid'],
+                'lengths' => [100, 100],
```
fogelito, line 1983: Provider size is 128 , and providerUid is 2048 (does it have to be so long?)
So will do :
'lengths' => [128, 640],
Since both together is the max of 768

app/controllers/api/account.php
```diff
@@ -441,27 +442,70 @@
...
-            throw new Exception(Exception::USER_MISSING_ID);
+        $name = $oauth2->getUserName($accessToken);
+        $email = $oauth2->getUserEmail($accessToken);
+
+        // Check if this identity is connected to a different user
+        if (!$user->isEmpty()) {
+            $userId = $user->getId();
+
+            $identitiesWithMatchingEmail = $dbForProject->find('identities', [
+                Query::equal('providerEmail', [$email]),
```
fogelito, line 503: I do not see an index for providerEmail

app/controllers/api/account.php
```diff
@@ -535,13 +593,59 @@
...
 
+        Authorization::setRole(Role::user($user->getId())->toString());
+        Authorization::setRole(Role::users()->toString());
+
         if (false === $user->getAttribute('status')) { // Account is blocked
-            throw new Exception(Exception::USER_BLOCKED); // User is in status blocked
+            $failureRedirect(Exception::USER_BLOCKED); // User is in status blocked
+        }
+
+        $identity = $dbForProject->findOne('identities', [
```
fogelito, line 608: To the existing userInternalId index , you can add these 2 attributes: provider, providerUid
['userInternalId', 'provider', 'providerUid']

app/controllers/api/account.php
```diff
@@ -624,6 +728,86 @@
...
+            if ($cursorDocument->isEmpty()) {
+                throw new Exception(Exception::GENERAL_CURSOR_NOT_FOUND, "Identity '{$identityId}' for the 'cursor' value not found.");
+            }
+
+            $cursor->setValue($cursorDocument);
+        }
+
+        $filterQueries = Query::groupByType($queries)['filters'];
+
+        $results = $dbForProject->find('identities', $queries);
```
fogelito, line 771: Since $queries are dynamic, will have to add indexes for the most common attributes, that will be queried for

app/config/collections.php
stnguyen90: > and providerUid is 2048 (does it have to be so long?)

`providerUid` in the sessions collection is 2048, so I wanted to make it consistent. Is that okay?

app/controllers/api/account.php
stnguyen90: You're suggesting to make sure there's an index on this query (userInternalId, provider, and providerUid)?

fogelito: yes , since there is a unique index on [provider,providerUid],
I think adding an additional index on the above will be good

PR #5552, opened 2023-05-18
Collection Index Validation

fogelito:


stnguyen90:
@fogelito looks like there's a merge conflict now.

PR #5556, opened 2023-05-19
Return Relationship Metadata for Offline Support

stnguyen90:
## What does this PR do?

If the client has offline persistency enabled (indicated by the X-SDK-Offline request header), the server will return the collection's relationships attributes so that while the client is offline, it knows where to store related data.

The header will include something like:

```json
[
  {
    "databaseId": "many-to-one",
    "collectionId": "level1",
    "attributes": {
      "level2": {
        "key": "level2",
        "type": "relationship",
        "relatedCollection": "level2",
        "relationType": "manyToOne",
        "twoWay": true,
        "twoWayKey": "level1",
        "onDelete": "setNull",
        "side": "parent"
      }
    }
  },
  {
    "databaseId": "many-to-one",
    "collectionId": "level3",
    "attributes": {
      "level2": {
        "key": "level2",
        "type": "relationship",
        "relatedCollection": "level2",
        "relationType": "manyToOne",
        "twoWay": true,
        "twoWayKey": "level3",
        "onDelete": "setNull",
        "side": "child"
      },
      "level4": {
        "key": "level4",
        "type": "relationship",
        "relatedCollection": "level4",
        "relationType": "manyToOne",
        "twoWay": true,
        "twoWayKey": "level3",
        "onDelete": "setNull",
        "side": "parent"
      }
    }
  },
  {
    "databaseId": "many-to-one",
    "collectionId": "level2",
    "attributes": {
      "level1": {
        "key": "level1",
        "type": "relationship",
        "relatedCollection": "level1",
        "relationType": "manyToOne",
        "twoWay": true,
        "twoWayKey": "level2",
        "onDelete": "setNull",
        "side": "child"
      },
      "level3": {
        "key": "level3",
        "type": "relationship",
        "relatedCollection": "level3",
        "relationType": "manyToOne",
        "twoWay": true,
        "twoWayKey": "level2",
        "onDelete": "setNull",
        "side": "parent"
      }
    }
  }
]
```

## Test Plan

TODO

## Related PRs and Issues

TBD

## Checklist


app/controllers/api/databases.php
```diff
@@ -368,6 +369,65 @@ function updateAttribute(
     return $attribute;
 }
 
+/**
+ * Recursively processes a collection to get it's relationships
+ *
+ * Returns true if the user has access to the collection
+ */
+function getCollectionRelationships(Document $collection, bool $mode, Database $dbForProject, int $depth, array &$collections): bool
```
abnegate, line 377: Can we set a resource and inject it instead of using a global function?

app/controllers/api/databases.php
```diff
@@ -368,6 +369,65 @@ function updateAttribute(
...
+    if (!($mode === APP_MODE_ADMIN && Auth::isPrivilegedUser(Authorization::getRoles()))) {
+        if (!$collection->getAttribute('documentSecurity', false)) {
+            $validator = new Authorization(Database::PERMISSION_READ);
+            if (!$validator->isValid($collection->getRead())) {
+                return false;
+            }
+        }
+    }
+
+    if ($depth + 1 > Database::RELATION_MAX_DEPTH) {
```
abnegate, line 389: Might be clearer to use `>=` instead of `$depth + 1`

app/controllers/api/databases.php
```diff
@@ -368,6 +369,65 @@ function updateAttribute(
...
+
+    $relationshipAttributes = \array_filter(
+        $collection->getAttribute('attributes', []),
+        fn ($attribute) => $attribute->getAttribute('type') === Database::VAR_RELATIONSHIP
+    );
+
+    $attributes = [];
+    foreach ($relationshipAttributes as $attribute) {
+        /** @var Document $attribute */
+        $relatedCollection = Authorization::skip(fn() => $dbForProject->getDocument('database_' . $collection->getAttribute('databaseInternalId'), $attribute->getAttribute('relatedCollection')));
```
abnegate, line 401: Instead of skipping auth, can we utilize the permission check at the database level and check if `$relatedCollection->isEmpty()`, then skip the recursion if it is? Then we can remove the manual check at the start of this function too

app/controllers/api/databases.php
stnguyen90: @eldadfux, considering your comment over https://github.com/appwrite/appwrite/pull/5587#discussion_r1253434658, do you not want us to set any global functions as a resource? Some other cases where we have functions we want to re-use between endpoints are:

1. create user
2. [createAttribute](https://github.com/appwrite/appwrite/blob/9f2bf844c7acc4932ca377941061665cd9b96645/app/controllers/api/databases.php#L62)
3. [updateAttribute](https://github.com/appwrite/appwrite/blob/9f2bf844c7acc4932ca377941061665cd9b96645/app/controllers/api/databases.php#L212)
4. [avatarCallback](https://github.com/appwrite/appwrite/blob/9f2bf844c7acc4932ca377941061665cd9b96645/app/controllers/api/avatars.php#L19)
5. some graphql methods

eldadfux: There are a few places where we compromised on this as we didn't had any other reasonable choice. Me and @lohanidamodar are working on a longer term solution to define routes as classes. This will allow us to better reuse or extend routes logic.

Issue #5490, opened 2023-05-02
🐛 Bug Report: "User has been blocked" error on creating account on Appwrite


rapid-killer-9:
After Creating an account on [Appwrite](https://t.co/CrxNdOjteN) l.
I have noticed that my details are incorrect. so, I deleted the account to recreate but after that whenever I click on signup or singing it gives me the same error again and again

![image](https://user-images.githubusercontent.com/83023263/235745461-a7d8fcfc-961c-4a70-9a4d-343596112654.png)

Hi-ImKyle:
+1, is there not a default admin account that can be used?

stnguyen90:
We've updated the console such that deleting will actually delete the account rather than block it.

PR #5561, opened 2023-05-19
Update the Update Account Status API to clear the cookie

stnguyen90:
## What does this PR do?

After a user updates their status, their session no longer works however, the cookie may still exist in their browser, preventing other API calls from completing successfully.

## Test Plan

E2E test updated

## Related PRs and Issues

* https://github.com/appwrite/appwrite/issues/5490

## Checklist

### DIFF
--- a/app/config/errors.php
+++ b/app/config/errors.php
@@ -112,7 +112,7 @@ return [
     ],
     Exception::USER_BLOCKED => [
         'name' => Exception::USER_BLOCKED,
-        'description' => 'The current user has been blocked. You can unblock the user from the Appwrite console.',
+        'description' => 'The current user has been blocked.',
         'code' => 401,
     ],
     Exception::USER_INVALID_TOKEN => [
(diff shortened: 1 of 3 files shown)


lohanidamodar (approved):

eldadfux:
How does having an invalid cookie interfere with other calls?

stnguyen90:
> How does having an invalid cookie interfere with other calls?

Since the cookie is still there, API calls result in "User is blocked error" which can prevent progressing in the app/Console.

For example, this shows deleting the account and then refreshing the page/navigating to the console again:

https://github.com/appwrite/appwrite/assets/1477010/debe45e5-3665-4aae-9e05-0ced83ca9313

eldadfux:
Can we also change the error message? "You can unblock the user from the console" part is not relevant to the person receiving the error.

eldadfux (approved):

PR #5577, opened 2023-05-24
Feat: Improve logs

Meldiron:
## What does this PR do?

Makes sure function errors are logged to Sentry

## Test Plan

- [x] Manual QA

CleanShot 2023-05-24 at 21 00 25@2x

CleanShot 2023-05-24 at 21 00 41@2x

CleanShot 2023-05-24 at 21 13 32@2x

## Related PRs and Issues

x

## Checklist

### DIFF
--- a/app/workers/functions.php
+++ b/app/workers/functions.php
@@ -20,6 +20,7 @@ use Utopia\Database\Permission;
 use Utopia\Database\Query;
 use Utopia\Database\Role;
 use Utopia\Database\Validator\Authorization;
+use Utopia\Logger\Log;
 use Utopia\Queue\Server;
 
 Authorization::disable();
@@ -27,6 +28,7 @@ Authorization::setDefaultStatus(false);
 
 Server::setResource('execute', function () {
     return function (
+        Log $log,
         Func $queueForFunctions,
         Database $dbForProject,
         Client $statsd,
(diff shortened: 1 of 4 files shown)


app/workers/functions.php
```diff
@@ -217,6 +221,11 @@
                 ->setParam('networkResponseSize', 0)
                 ->submit();
         }
+
+        if (!empty($error)) {
+            $projectId = $project->getId();
+            throw new Exception("Error in project #{$projectId}, function #{$functionId}: " . $error);
```
eldadfux, line 227: I'm not sure this the best way to utilize Sentry. It will treat each error as its own separate things. Can we make our exception accept more metadata and them pass it properly to Sentry or any other logging provider?

Meldiron: Made log a resource and updated it from worker

app/workers/functions.php
```diff
@@ -153,10 +161,8 @@
                 ->setAttribute('statusCode', $th->getCode())
                 ->setAttribute('stderr', $th->getMessage());
 
-            Console::error($th->getTraceAsString());
-            Console::error($th->getFile());
-            Console::error($th->getLine());
-            Console::error($th->getMessage());
```
christyjacob4, line 159: why did we remove this ? I think this should still stay

Meldiron: At the end of job we re-throw error if one was cought. That goes to worker.php into ::error(). There see send error to sentry, but also log error.

This was removed because same error was printed to CLI twice

christyjacob4: okay 👍

christyjacob4 (approved):

PR #5587, opened 2023-05-27
Add a new labels attribute to the Users collection

stnguyen90:
## What does this PR do?

This labels attribute can be used apply labels to users. These labels can eventually be used:

* with permissions to grant access
* to filter users in the Users collection

## Test Plan

Automated

## Related PRs and Issues

Depends on:

* https://github.com/utopia-php/database/pull/280

Blocks:

* https://github.com/appwrite/console/pull/457

## Checklist

### DIFF
--- a/app/controllers/api/users.php
+++ b/app/controllers/api/users.php
@@ -686,8 +686,7 @@ App::put('/v1/users/:userId/labels')
     ->inject('response')
     ->inject('dbForProject')
     ->inject('events')
-    ->inject('prepareUserSearch')
-    ->action(function (string $userId, array $labels, Response $response, Database $dbForProject, Event $events, callable $prepareUserSearch) {
+    ->action(function (string $userId, array $labels, Response $response, Database $dbForProject, Event $events) {
 
         $user = $dbForProject->getDocument('users', $userId);
 
@@ -696,7 +695,6 @@ App::put('/v1/users/:userId/labels')
         }
 
         $user->setAttribute('labels', (array) \array_values(\array_unique($labels)));
-        $user->setAttribute('search', $prepareUserSearch($user));
 
         $user = $dbForProject->updateDocument('users', $user->getId(), $user);
 
@@ -799,8 +797,7 @@ App::patch('/v1/users/:userId/name')
     ->inject('response')
     ->inject('dbForProject')
     ->inject('events')
-    ->inject('prepareUserSearch')
-    ->action(function (string $userId, string $name, Response $response, Database $dbForProject, Event $events, callable $prepareUserSearch) {
+    ->action(function (string $userId, string $name, Response $response, Database $dbForProject, Event $events) {
 
         $user = $dbForProject->getDocument('users', $userId);
 
@@ -809,7 +806,6 @@ App::patch('/v1/users/:userId/name')
         }
 
         $user->setAttribute('name', $name);
-        $user->setAttribute('search', $prepareUserSearch($user));
 
         $user = $dbForProject->updateDocument('users', $user->getId(), $user);
 
@@ -898,8 +894,7 @@ App::patch('/v1/users/:userId/email')
     ->inject('response')
     ->inject('dbForProject')
     ->inject('events')
-    ->inject('prepareUserSearch')
-    ->action(function (string $userId, string $email, Response $response, Database $dbForProject, Event $events, callable $prepareUserSearch) {
+    ->action(function (string $userId, string $email, Response $response, Database $dbForProject, Event $events) {
 
         $user = $dbForProject->getDocument('users', $userId);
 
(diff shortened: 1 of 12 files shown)


app/controllers/api/users.php
```diff
@@ -663,6 +665,45 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+    ->label('usage.metric', 'users.{scope}.requests.update')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'users')
+    ->label('sdk.method', 'updateLabels')
+    ->label('sdk.description', '/docs/references/users/update-user-labels.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_USER)
+    ->param('userId', '', new UID(), 'User ID.')
+    ->param('labels', [], new ArrayList(new Text(128, allowList: [...Text::NUMBERS, ...Text::ALPHABET_UPPER, ...Text::ALPHABET_LOWER]), APP_LIMIT_ARRAY_PARAMS_SIZE), 'Array of user labels. Replaces the previous labels. Maximum of 100 labels are allowed, each 128 characters long. Valid chars are a-z, A-Z, and 0-9.')
```
eldadfux, line 685: @TorstenDittmann @stnguyen90 how will too many labels affect permissions performance? Should we consider a smaller limit?

src/Appwrite/Utopia/Response/Model/Membership.php
```diff
@@ -80,7 +80,7 @@ public function __construct()
                 'type' => self::TYPE_STRING,
                 'description' => 'User list of roles',
                 'default' => [],
-                'example' => 'admin',
+                'example' => ['owner'],
```
eldadfux, line 83: If we use array=true one line below do we still need to provide the example like this? do we have issues in other places as well?

docs/references/users/update-user-labels.md
```diff
@@ -0,0 +1 @@
+Update the user labels by its unique ID. The array you pass is stored as is, and replaces any previous value. Each label has a maximum of 128 characters.
```
eldadfux, line 1: I would also add another line explaining where this can come useful and how it connect to the permissions system. Might also be useful to add another line for labels vs teams.

eldadfux (commented):
@stnguyen90 looks great, left a few suggestions.

src/Appwrite/Utopia/Response/Model/Membership.php
stnguyen90: Yes, we need to. Our [membership docs](https://appwrite.io/docs/models/membership) shows an incorrect example:

image

docs/references/users/update-user-labels.md
stnguyen90: @eldadfux are we adding the permission side of things as part of this release?

app/controllers/api/users.php
```diff
@@ -663,6 +665,47 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+    ->inject('prepareUserSearch')
+    ->action(function (string $userId, array $labels, Response $response, Database $dbForProject, Event $events, callable $prepareUserSearch) {
+
+        $user = $dbForProject->getDocument('users', $userId);
+
+        if ($user->isEmpty()) {
+            throw new Exception(Exception::USER_NOT_FOUND);
+        }
+
+        $user->setAttribute('labels', (array) $labels);
```
TorstenDittmann, line 698: this will not cover unique values in the array right? means we can have the same label twice?

app/controllers/shared/api.php
```diff
@@ -87,6 +87,25 @@
     }
 };
 
+App::setResource('prepareUserSearch', function () {
```
eldadfux, line 90: We do not use resource for global functions, this could become a very messy pattern over time. For this scenario, using database filters would work great.

app/controllers/api/users.php
```diff
@@ -663,6 +665,47 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
         $response->dynamic($user, Response::MODEL_USER);
     });
 
+App::put('/v1/users/:userId/labels')
```
eldadfux, line 668: We have something similar for buckets where we update an array of strings. I wonder if we should have a dedicated endpoint vs having another attribute for creating and updating users.

eldadfux (changes requested):

app/controllers/api/users.php
stnguyen90: @eldadfux, sorry I don't quite understand. Would you please clarify what's the similar thing for buckets?

stnguyen90: This matches the existing APIs where we have separate APIs for updating name, email, etc. Do you recall why it was done that way?

stnguyen90: I synced with Eldad and we may want to introduce an Update User API in the future, but, for now, we'll just use this specific endpoint since it matches the other update user endpoints.

TorstenDittmann (approved):

PR #5603, opened 2023-05-30
Update framework to fix route mismatches

abnegate:
## What does this PR do?

- Update framework to include a check to ensure that the matched URL has the same amount of segments (separated by /) as the route. This ensures that a URL will not match an incorrect route if parameters are missing.

## Test Plan

- Add assertion to ensure routes are not mismatched
- Fixed existing tests

## Checklist


eldadfux (approved):

wilhantian:
There are omission in `/app/controllers/general.php`,  need to modify `App:: get ('/.well-know/acme-challenge ')` the code has correctly matched the URL `xxx.com/.well-know/acme-challenge/xxx`

PR #5604, opened 2023-05-30
feat: add encrypt param to string attribute

### LINKED ISSUE #3150: 🚀 Feature: Encypted database fields
### 🔖 Feature description

It would be very handy especially for string fields if you could have AppWrite encrypt them natively.

### 🎤 Pitch

Here I think of the use-case for keys that are stored directly in the database, which should not be readable for everyone with database access. Of course, there are other examples of sensitive data that should not be stored as plain text in the database. Of course you would lose some queries and indexes for such fields, but "equal" would still work without problems. As key for these fields you could either take the already existing `_APP_OPENSSL_KEY_V1` or create an own ENV variable for it.

- [x] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

2002Bishwajeet:
## What does this PR do?

Add and optional encrypt param to encrypt attributes

## Test Plan

Unit and E2E testing

## Related PRs and Issues

Fix #3150 

## Checklist


app/controllers/api/databases.php
```diff
@@ -1103,27 +1103,35 @@ function updateAttribute(
     ->param('key', '', new Key(), 'Attribute Key.')
     ->param('size', null, new Range(1, APP_DATABASE_ATTRIBUTE_STRING_MAX_LENGTH, Range::TYPE_INTEGER), 'Attribute size for text attributes, in number of characters.')
     ->param('required', null, new Boolean(), 'Is attribute required?')
+    ->param('encrypt', false, new Boolean(), 'Encrypt attribute? Encrypting an attribute means that the attribute can not be queried.', true)
```
abnegate, line 1106: Let's move this to be the last param so it's a non-breaking change

abnegate, line 1106: @gewenyu99 Would love your feedback on the description

app/controllers/api/databases.php
```diff
@@ -1167,7 +1175,7 @@ function updateAttribute(
             'required' => $required,
             'default' => $default,
             'array' => $array,
-            'format' => APP_DATABASE_ATTRIBUTE_EMAIL,
+            'format' => APP_DATABASE_ATTRIBUTE_EMAIL
```
abnegate, line 1178: Let's keep all trailing commas, it makes it easier to add a new item or re-order the existing ones

tests/e2e/Services/Databases/DatabasesCustomServerTest.php
```diff
@@ -593,6 +593,200 @@ public function testUpdateCollection(array $data)
...
+
+        $alias = $this->client->call(Client::METHOD_POST, $attributesPath . '/string', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'x-appwrite-key' => $this->getProject()['apiKey']
+        ]), [
+            'key' => 'alias',
+            'size' => 256,
+            'required' => false,
+        ]);
```
abnegate, line 674: Let's remove these since they aren't needed to test to encrypt functionality

tests/e2e/Services/Databases/DatabasesCustomServerTest.php
```diff
@@ -593,6 +593,200 @@ public function testUpdateCollection(array $data)
...
+        $this->assertEquals(202, $alias['headers']['status-code']);
+        $this->assertEquals('alias', $alias['body']['key']);
+        $this->assertEquals('string', $alias['body']['type']);
+
+        $this->assertEquals(202, $age['headers']['status-code']);
+        $this->assertEquals('age', $age['body']['key']);
+        $this->assertEquals('integer', $age['body']['type']);
+
+        // Wait for database worker to finish creating attributes
+        sleep(5);
```
abnegate, line 696: We can probably reduce this to 2 seconds, always good to reduce test time where we can

tests/e2e/Services/Databases/DatabasesCustomServerTest.php
```diff
@@ -593,6 +593,200 @@ public function testUpdateCollection(array $data)
...
+        ]), [
+            'key' => 'key_lastName',
+            'type' => 'key',
+            'attributes' => [
+                'lastName',
+            ],
+        ]);
+
+        // Wait for database worker to finish creating index
+        sleep(2);
```
abnegate, line 731: Let's remove this, doesn't make much sense to add an index on an encrypted attribute since it can't be queried

tests/e2e/Services/Databases/DatabasesCustomServerTest.php
```diff
@@ -593,6 +593,200 @@ public function testUpdateCollection(array $data)
...
+        // Delete attribute
+        $attribute = $this->client->call(Client::METHOD_DELETE, '/databases/' . $databaseId . '/collections/' . $actors ['body']['$id'] . '/attributes/' . $unneededId, array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'x-appwrite-key' => $this->getProject()['apiKey']
+        ]));
+
+        $this->assertEquals(204, $attribute['headers']['status-code']);
+
+        sleep(2);
```
abnegate, line 760: Let's remove these since they aren't needed to test to encrypt functionality

tests/e2e/Services/Databases/DatabasesCustomServerTest.php
```diff
@@ -593,6 +593,200 @@ public function testUpdateCollection(array $data)
...
+        $this->assertIsArray($collection['body']['attributes']);
+        $this->assertCount(3, $collection['body']['attributes']);
+        $this->assertEquals($collection['body']['attributes'][0]['key'], $firstName['body']['key']);
+        $this->assertEquals($collection['body']['attributes'][1]['key'], $lastName['body']['key']);
+
+        return [
+            'collectionId' => $actors['body']['$id'],
+            'key' => $index['body']['key'],
+            'databaseId' => $databaseId
+        ];
```
abnegate, line 787: Let's remove these since they aren't needed to test to encrypt functionality

tests/e2e/Services/Databases/DatabasesCustomServerTest.php
```diff
@@ -593,6 +593,200 @@ public function testUpdateCollection(array $data)
...
+
+        sleep(2);
+
+        // Check document to ensure cache is purged on schema change
+        $document = $this->client->call(Client::METHOD_GET, '/databases/' . $databaseId . '/collections/' . $actors['body']['$id'] . '/documents/' . $document['body']['$id'], array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'x-appwrite-key' => $this->getProject()['apiKey']
+        ]));
+
```
abnegate, line 768: Let's add an assertion to check that `lastName` is decrypted properly when getting the document

abnegate (commented):
Nice work

tests/e2e/Services/Databases/DatabasesCustomServerTest.php
```diff
@@ -593,6 +593,110 @@ public function testUpdateCollection(array $data)
...
+            'key' => 'lastName',
+            'size' => 256,
+            'required' => true,
+            'encrypt' => true,
+        ]);
+
+
+     /**
+      * Check Status of every Attribute
+      */
```
abnegate, line 658: Let's fix the indentation here

abnegate (approved):

app/controllers/api/databases.php
gewenyu99: I suggest giving more context. "Encrypt attribute? [Describe what encrypting an attribute does] Encrypted attributes cannot be queried."

app/controllers/api/databases.php
```diff
@@ -1105,25 +1105,33 @@ function updateAttribute(
     ->param('required', null, new Boolean(), 'Is attribute required?')
     ->param('default', null, new Text(0, 0), 'Default value for attribute when not provided. Cannot be set when attribute is required.', true)
     ->param('array', false, new Boolean(), 'Is attribute an array?', true)
+    ->param('encrypt', false, new Boolean(), 'Encrypt attribute? Encrypting an attribute means that the attribute can not be queried.', true)
     ->inject('response')
     ->inject('dbForProject')
     ->inject('database')
     ->inject('events')
-    ->action(function (string $databaseId, string $collectionId, string $key, ?int $size, ?bool $required, ?string $default, bool $array, Response $response, Database $dbForProject, EventDatabase $database, Event $events) {
+    ->action(function (string $databaseId, string $collectionId, string $key, ?int $size, ?bool $required, ?string $default, bool $array, bool $encrypt, Response $response, Database $dbForProject, EventDatabase $database, Event $events) {
```
christyjacob4, line 1113: We also need to add this param to the update String Attribute endpoint. 

Was it discussed to only implement this feature for string attributes?

christyjacob4 (changes requested):
Please address the comments

app/controllers/api/databases.php
2002Bishwajeet: @gewenyu99  What about now??

gewenyu99: > @gewenyu99 What about now??

What does encrypting an attribute do? Can you describe it to me briefly?

2002Bishwajeet: > > @gewenyu99 What about now??
> 
> What does encrypting an attribute do? Can you describe it to me briefly?

Encrypting an attribute would keep the data encrypted in the database. Incase if a bad actor gets accessed to your data, it wouldn't be able to access the data of the encrypted attribute. Encrypting would also means that we would loose indexing and queries support

gewenyu99: > Encrypt attribute? Appwrite Database encrypts data in encrypted attributes. Even if malicious actors gain access to your data, they wouldn't be able to access the data of the encrypted attribute. You can't query or index encrypted data.

gewenyu99 (approved):

gewenyu99:
Sorry for taking 2 weeks to get to this Biswa!

app/controllers/api/databases.php
```diff
@@ -1778,17 +1789,26 @@ function updateAttribute(
...
+        }
+
         $attribute = updateAttribute(
             databaseId: $databaseId,
             collectionId: $collectionId,
             key: $key,
             dbForProject: $dbForProject,
             events: $events,
             type: Database::VAR_STRING,
+            filter: empty($filter) ? null : $filter,
```
abnegate, line 1811: Let's remove this for now, adding/removing encryption to an existing attribute would require quite a bit more work. We can address this in future

app/controllers/api/databases.php
abnegate: We can not add this for update, otherwise we'd need to migrate all the data when changed. 

It was discussed to allow for strings only - it's the only attribute that it can be used for as it's stored as a string in the database, otherwise validation needed to be skipped. If you really need to encrypt a number, you can still do so as a string

christyjacob4 (approved):

app/controllers/api/databases.php
```diff
@@ -1739,6 +1749,8 @@ function updateAttribute(
     ->inject('dbForProject')
     ->inject('events')
     ->action(function (string $databaseId, string $collectionId, string $key, ?bool $required, ?string $default, Response $response, Database $dbForProject, Event $events) {
+
+
```
abnegate, line 1753:
```suggestion
```

abnegate (approved):

PR #5616, opened 2023-05-31
Refactor cache poc

shimonewman:
Saving  output to the cache instead of  the encoded json data

### DIFF
--- a/app/controllers/shared/api.php
+++ b/app/controllers/shared/api.php
@@ -207,7 +207,7 @@ App::init()
             $timestamp = 60 * 60 * 24 * 30;
             $data = $cache->load($key, $timestamp);
 
-            if (!empty($data) && !empty($cacheLog)) {
+            if (!empty($data) && !$cacheLog->isEmpty()) {
                 $parts = explode('/', $cacheLog->getAttribute('resourceType'));
                 $type = $parts[0] ?? null;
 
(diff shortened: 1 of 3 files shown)


app/controllers/shared/api.php
```diff
@@ -487,22 +484,17 @@
...
+                $key = md5($request->getURI() . '*' . implode('*', $request->getParams())) . '*' . APP_CACHE_BUSTER;
+                $signature = md5($data['payload']);
+                $cacheLog  =  Authorization::skip(fn () => $dbForProject->getDocument('cache', $key));
                 $accessedAt = $cacheLog->getAttribute('accessedAt', '');
                 $now = DateTime::now();
                 if ($cacheLog->isEmpty()) {
                     Authorization::skip(fn () => $dbForProject->createDocument('cache', new Document([
                     '$id' => $key,
                     'resource' => $resource,
+                    'resourceType' => $resourceType,
```
stnguyen90, line 496: `$resourceType` looks like it's the `$id` of the bucket, but should we use the internal id instead? same with `$resource`.

shimonewman: We also use bucketId and fileId in getDocument() to check cache file permissions.

stnguyen90: I see, we get the IDs from the cacheLog:

https://github.com/appwrite/appwrite/blob/2f7b0c9938e55cae647ac72651d2f91574843342/app/controllers/shared/api.php#L211-L216

We may end up deleting the wrong cache in cases where the delete job runs after a file/bucket was recreated with the same ID, but it's just a cache, so it's okay.

stnguyen90 (approved):

PR #5620, opened 2023-06-01
Fix system performance

stnguyen90:
## What does this PR do?

* Prevent triggering a function off of a function
* Don't clear cache on startup 
* Update console to include new fixes from https://github.com/appwrite/console/pull/458
* Add Check for recursive function execution in the functions worker. 

## Test Plan

Tested adding a functions.* event

image

## Related PRs and Issues

None

## Checklist

### DIFF
--- a/src/Appwrite/Event/Validator/Event.php
+++ b/src/Appwrite/Event/Validator/Event.php
@@ -45,7 +45,7 @@ class Event extends Validator
          * Identify all sections of the pattern.
          */
         $type = $parts[0] ?? false;
-            
+
         if ($type == 'functions') {
             $this->message = 'Triggering a function on a function event is not allowed.';
             return false;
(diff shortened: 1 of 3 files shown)


src/Appwrite/Event/Validator/Event.php
```diff
@@ -40,6 +45,12 @@ public function isValid($value): bool
          * Identify all sections of the pattern.
          */
         $type = $parts[0] ?? false;
+            
+        if ($type == 'functions') {
+            $this->message = 'Triggering a function on a function event is not allowed.';
+            return false;
```
stnguyen90, line 51: This is one approach. 

Alternatively, we can remove the event from:

https://github.com/appwrite/appwrite/blob/0765aac5d27cb7f39e839f85bc6eada87a011cc7/app/config/events.php#L195-L236

christyjacob4: If we remove the functions events from the events.php file we will also disable it for webhooks and realtime.

TorstenDittmann (approved):

christyjacob4 (approved):

PR #5622, opened 2023-06-01
Refactor collections config

lohanidamodar:
## What does this PR do?

- Update collections config into sections so that we can create only relevant collections for console and each projects separately. This will help extend console only collection easily and also improve performance while creating new projects.

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/config/collections.php
```diff
@@ -3591,6 +3603,13 @@
             ],
         ],
     ],
+], $commonCollections);
+
+$collections = [
```
eldadfux, line 3608: I was thinking, maybe it will be easier to read if we just add a property to each collection stating where it should go.

src/Appwrite/Utopia/Database/Validator/Queries/Base.php
```diff
@@ -24,7 +24,9 @@ class Base extends Queries
      */
     public function __construct(string $collection, array $allowedAttributes)
     {
-        $collection = Config::getParam('collections', [])[$collection];
+        $config = Config::getParam('collections', []);
```
eldadfux, line 27: What exactly are we validating here?

eldadfux (changes requested):

src/Appwrite/Utopia/Database/Validator/Queries/Base.php
lohanidamodar: Query allowed attributes, validating the attribute is valid in the collection structure.

app/config/collections.php
lohanidamodar: Not sure I understand.

eldadfux (approved):

lohanidamodar:
This change is already in main, no sure why there's this PR as well, I don't remember.

PR #5630, opened 2023-06-02
Extract db from list while backing

shimonewman:


app/controllers/api/projects.php
```diff
@@ -90,7 +89,27 @@
         }
 
         $projectId = ($projectId == 'unique()') ? ID::unique() : $projectId;
+
+        $backups[0] = ['from' => '4:30', 'to' => '5:30'];
+        $backups[1] = ['from' => '11:30', 'to' => '12:30'];
+        $backups[2] = ['from' => '21:30', 'to' => '22:30'];
```
christyjacob4, line 95: let's not index it by the 0,1,2. 

If we change the order of the env variables, it will get messed up. 

use the database names as keys like so `db_fra1_01` and so on. Please verify the db names

app/controllers/api/projects.php
```diff
@@ -90,7 +89,34 @@
...
+                }
+                $backup = $backups[$database];
+                $from = DateTime::createFromFormat('H:i', $backup['from']);
+                $to = DateTime::createFromFormat('H:i', $backup['to']);
+                if ($now >= $from && $now <= $to) {
+                    unset($databases[$index]);
+                    break;
+                }
+            }
+            $databases = array_values($databases);
```
christyjacob4, line 117:
```suggestion
```
This line seems unnecessary 🤔 

Screenshot 2023-06-06 at 5 51 05 PM

christyjacob4 (approved):

PR #5649, opened 2023-06-03
Sync master with 1.3.x 

christyjacob4:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/web/console.php
```diff
@@ -5,10 +5,11 @@
 
 App::get('/console/*')
     ->alias('/')
+    ->alias('auth/*')
```
Meldiron, line 8: Is this on purpose? Looking at code around, I would expect `/` at beginning

Meldiron (approved):

app/controllers/web/console.php
christyjacob4: I think it works both ways ( with and without the leading / ) I had tested it and it seemed to work

PR #5695, opened 2023-06-13
cache deletion bug fix

shimonewman:
- separated  cache deleteCacheByDate() and cacheDeleteByResource() 
- added  cacheDeleteByResource()  to put file API

### DIFF
--- a/app/controllers/api/storage.php
+++ b/app/controllers/api/storage.php
@@ -1263,7 +1263,8 @@ App::put('/v1/storage/buckets/:bucketId/files/:fileId')
     ->inject('user')
     ->inject('mode')
     ->inject('events')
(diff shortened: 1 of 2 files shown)


app/controllers/api/storage.php
```diff
@@ -1329,6 +1330,11 @@
             $file = Authorization::skip(fn() => $dbForProject->updateDocument('bucket_' . $bucket->getInternalId(), $fileId, $file));
         }
 
+        $deletes
+            ->setType(DELETE_TYPE_CACHE_BY_RESOURCE)
+            ->setResource('file/' . $file->getId())
+        ;
+
```
christyjacob4, line 1337: deleting cache when updating a file makes sense. But why do we need to delete cache when creating a file ?

christyjacob4 (approved):

PR #5698, opened 2023-06-13
Fix cache delete bug master

shimonewman:
fix delete cache bug to master

### DIFF
--- a/app/controllers/api/storage.php
+++ b/app/controllers/api/storage.php
@@ -654,11 +654,6 @@ App::post('/v1/storage/buckets/:bucketId/files')
             ->setContext('bucket', $bucket)
         ;
 
-        $deletes
-            ->setType(DELETE_TYPE_CACHE_BY_RESOURCE)
-            ->setResource('file/' . $file->getId())
-        ;
-
         $metadata = null; // was causing leaks as it was passed by reference
 
         $response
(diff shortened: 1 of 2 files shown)


stnguyen90 (approved):

app/controllers/api/storage.php
```diff
@@ -654,11 +653,6 @@
             ->setContext('bucket', $bucket)
         ;
 
-        $deletes
```
eldadfux, line 657: Didn't we have this to avoid a conflict when a file with a similar ID was recently deleted?

stnguyen90: @eldadfux this event should already be triggered when the file is deleted, right?

app/workers/deletes.php
```diff
@@ -130,31 +130,56 @@ public function shutdown(): void
...
-        $this->deleteCacheFiles([
-            Query::lessThan('accessedAt', $this->args['datetime']),
-        ]);
+        $cache = new Cache(
+            new Filesystem(APP_STORAGE_CACHE . DIRECTORY_SEPARATOR . 'app-' . $projectId)
+        );
+
+        $query = [
+            Query::equal('resource', [$resource])
+        ];
```
stnguyen90, line 148: Wait...how does this handle the same file id in multiple buckets?

stnguyen90 (commented):
See https://github.com/appwrite/appwrite/pull/5698/files#r1269913474

stnguyen90 (approved):
Approving this, but created https://github.com/appwrite/appwrite/issues/5861

eldadfux (approved):

PR #5699, opened 2023-06-13
fix delete cache file to dbp-master

shimonewman:
fix delete cache file to dbp-master

### DIFF
--- a/app/workers/deletes.php
+++ b/app/workers/deletes.php
@@ -169,7 +169,7 @@ class DeletesV1 extends Worker
     protected function deleteCacheByResource(Document $project, string $resource): void
     {
         $projectId = $project->getId();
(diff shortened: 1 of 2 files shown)


app/workers/deletes.php
```diff
@@ -201,10 +201,9 @@ protected function deleteCacheByDate(string $datetime): void
     {
         $this->deleteForProjectIds(function (Document $project) use ($datetime) {
 
-            $projectId = $project->getId();
```
christyjacob4, line 204: why did we remove this ?

christyjacob4: But $projectId was equal to $project->getId()

christyjacob4 (approved):

Issue #5198, opened 2023-03-09
🐛 Bug Report: Trigger a function `users.*.create` when an Anonymous user is created


chuiizeet:
### 👟 Reproduction steps

Try to use `users.*.create` when an anonymous user is created but does not works.

### 👍 Expected behavior

The function must be triggered

### 👎 Actual Behavior

The function does not triggered

### 🎲 Appwrite version

Version 1.2.x

### 💻 Operating system

MacOS

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

stnguyen90:
@chuiizeet, thanks for opening this issue! 🙏🏼 It looks like the [Create Anonymous Session](https://appwrite.io/docs/client/account?sdk=web-default#accountCreateAnonymousSession) endpoint only triggers a `users.[userId].sessions.[sessionId].create` event:

https://github.com/appwrite/appwrite/blob/e8c74d7204258133a61db09ed518211170e751d9/app/controllers/api/account.php#L1134

We definitely need to also trigger the `users.[userId].create` event. The big challenge right now is our event mechanism only supports 1 event per endpoint, so we've been discussing rewriting the system to support multiple events.

Related issue: https://github.com/appwrite/appwrite/issues/2406

PR #5703, opened 2023-06-14
Add support for multiple events per endpoint

stnguyen90:
## What does this PR do?

Trigger events using database listeners

We initially designed events to trigger via labels on the endpoint. the problem with this approach was triggering multiple events per endpoint became very messy.

## Test Plan

TBD

## Related PRs and Issues

- #2406 
- https://github.com/appwrite/appwrite/issues/5198

## Checklist

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -219,7 +219,7 @@ App::post('/v1/account/sessions/email')
                 ->setAttribute('password', Auth::passwordHash($password, Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS))
                 ->setAttribute('hash', Auth::DEFAULT_ALGO)
                 ->setAttribute('hashOptions', Auth::DEFAULT_ALGO_OPTIONS);
-            $dbForProject->updateDocument('users', $profile->getId(), $profile);
+            $dbForProject->silent(fn () => $dbForProject->updateDocument('users', $profile->getId(), $profile));
         }
 
         $dbForProject->deleteCachedDocument('users', $profile->getId());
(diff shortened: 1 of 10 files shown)


app/controllers/api/databases.php
```diff
@@ -79,12 +79,16 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
         throw new Exception(Exception::DATABASE_NOT_FOUND);
     }
 
+    $events->setContext('database', $db);
```
stnguyen90, line 82: This context is required by some events so we have to make sure to set it before the DB operation so it's available in the handler.

app/controllers/api/account.php
```diff
@@ -703,6 +703,10 @@
         $loginSecret = Auth::tokenGenerator();
         $expire = DateTime::addSeconds(new \DateTime(), Auth::TOKEN_EXPIRATION_CONFIRM);
 
+        $events->setContext('token', new Document([
+            'secret' => $loginSecret,
+        ]));
```
stnguyen90, line 708: The unhashed secret is required in the event, so pass it to the handler using context.

app/init.php
```diff
@@ -1180,3 +1185,240 @@ function getDevice($root): Device
...
+            $collection = $document->getCollection();
+            $dump($databaseEvent);
+            $dump($collection);
+
+            switch ($collection) {
+                case 'users':
+                    if (!\str_starts_with($eventPattern, 'users.[userId]')) {
+                        $eventPattern = 'users.[userId]';
+                        $eventPatternReady = false;
+                    }
```
stnguyen90, line 1225: reverse engineering the event pattern given the collection feels a little hacky.

app/controllers/api/account.php
```diff
@@ -219,7 +219,7 @@
                 ->setAttribute('password', Auth::passwordHash($password, Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS))
                 ->setAttribute('hash', Auth::DEFAULT_ALGO)
                 ->setAttribute('hashOptions', Auth::DEFAULT_ALGO_OPTIONS);
-            $dbForProject->updateDocument('users', $profile->getId(), $profile);
+            $dbForProject->silent(fn () => $dbForProject->updateDocument('users', $profile->getId(), $profile));
```
stnguyen90, line 222: Wrapping with silent to disable all listeners might be dangerous.

PR #5726, opened 2023-06-19
Feat:  Whitelist enums

### LINKED ISSUE #426: OAuth provider String to Enum
Current 0Auth providers are strings. 
Appwrite should define an enum with all the allowed providers for each of use. 

Existing flutter example;
```
return await account.createOAuth2Session(provider: 'facebook');
```

Proposed improvment;
```
return await account.createOAuth2Session(provider: Provider.Facebook);
```

2002Bishwajeet:
## What does this PR do?

Add whitelist to the `enum` field in the spec files

## Test Plan

Manual Testing

## Related PRs and Issues

Fixes #426 

## Checklist


src/Appwrite/Specification/Format.php
```diff
@@ -97,4 +98,106 @@ public function getParam(string $key, string $default = ''): string
...
+                        switch ($param) {
+                            case 'gravity':
+                                return 'ImageGravity';
+                            case 'output':
+                                return  'ImageFormat';
+                        }
+                        break;
+                }
+                break;
+        }
```
abnegate, line 133: We're missing functions from the database, functions, projects and users services here

src/Appwrite/Specification/Format/Swagger2.php
```diff
@@ -413,7 +413,11 @@ public function parse(): array
                         /** @var \Utopia\Validator\WhiteList $validator */
                         $node['type'] = $validator->getType();
                         $node['x-example'] = $validator->getList()[0];
-
+                        if (!($route->getLabel('sdk.namespace', '') == 'users' && $route->getLabel('sdk.method', '') == 'getUsage' && $name == 'provider')) {
```
abnegate, line 416: I don't think we should explicitly exclude the usage provider

2002Bishwajeet: This was done because of [this](https://github.com/2002Bishwajeet/sdk-for-web/commit/9492f2f0914cc5e2600d448dcbefd87876d2a220#diff-a9c15c236009c0168cd4548b068c3a0994d8f3eb30f10995373cff94d6cf3d47)

So the files were getting overwritten and we didn't need the providers prefixed with Oauth-provider , thus excluded

src/Appwrite/Specification/Format.php
2002Bishwajeet: They don't have enum field set in the spec file (https://raw.githubusercontent.com/appwrite/appwrite/faece71145f680a839ec70d41e6eefe9e341fdbe/app/config/specs/swagger2-latest-console.json).

Is that expected or we are missing something?

abnegate: We're missing something 🤔 every parameter that uses the WhiteList validator should have the enum property. A project search for `new WhiteList` should show all the routes that need added

src/Appwrite/Specification/Format/Swagger2.php
abnegate: Let's try create a blacklist instead of specifically excluding a single route. That way it will be extensible if we need to exclude any others in future

2002Bishwajeet: Done

tests/e2e/Services/Teams/TeamsBaseServer.php
```diff
@@ -4,6 +4,7 @@
 
 use Tests\E2E\Client;
 use Utopia\Database\Validator\Datetime as DatetimeValidator;
+use Utopia\Database\Helpers\ID;
```
abnegate, line 7:
```suggestion
```

src/Appwrite/Specification/Format/OpenAPI3.php
```diff
@@ -414,6 +414,25 @@ public function parse(): array
...
+                                $blacklist['namespace'] == $route->getLabel('sdk.namespace', '')
+                                && $blacklist['method'] == $route->getLabel('sdk.method', '')
+                                && $blacklist['parameter'] == $name
+                            ) {
+                                $allowed = false;
+                                break;
+                            }
+                        }
+
+                        if (!false) {
```
abnegate, line 431:
```suggestion
                        if ($allowed) {
```

abnegate (approved):

eldadfux (approved):

eldadfux:
Great work @2002Bishwajeet!

PR #5738, opened 2023-06-22
Improve Appwrite's error messages.

gewenyu99:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/config/errors.php
+++ b/app/config/errors.php
@@ -10,52 +10,52 @@ return [
     /** General Errors */
     Exception::GENERAL_UNKNOWN => [
         'name' => Exception::GENERAL_UNKNOWN,
-        'description' => 'An unknown error has occured. Please check the logs for more information.',
+        'description' => 'An unknown error has occured. Please submit a support ticket at []() with reproduction steps for further assistance.',
         'code' => 500,
     ],
     Exception::GENERAL_MOCK => [
         'name' => Exception::GENERAL_MOCK,
-        'description' => 'General errors thrown by the mock controller used for testing.',
+        'description' => 'General errors thrown by the mock controller used for testing. If you\'re seeing this on a production build if Appwrite, please raise an issue at ()[]',
         'code' => 400,
     ],
     Exception::GENERAL_ACCESS_FORBIDDEN => [
         'name' => Exception::GENERAL_ACCESS_FORBIDDEN,
-        'description' => 'Access to this API is forbidden.',
+        'description' => 'Access to this API is forbidden. Please [action to get access]',
         'code' => 401,
     ],
     Exception::GENERAL_UNKNOWN_ORIGIN => [
         'name' => Exception::GENERAL_UNKNOWN_ORIGIN,
-        'description' => 'The request originated from an unknown origin. If you trust this domain, please list it as a trusted platform in the Appwrite console.',
+        'description' => 'The request originated from an unknown origin. If you trust this domain, please add it as a trusted platform in the Appwrite console.',
         'code' => 403,
     ],
     Exception::GENERAL_SERVICE_DISABLED => [
         'name' => Exception::GENERAL_SERVICE_DISABLED,
-        'description' => 'The requested service is disabled. You can enable the service from the Appwrite console.',
+        'description' => 'The requested service is disabled. You can enable the service in the settings of your Appwrite project.',
         'code' => 503,
     ],
     Exception::GENERAL_UNAUTHORIZED_SCOPE => [
         'name' => Exception::GENERAL_UNAUTHORIZED_SCOPE,
-        'description' => 'The current user or API key does not have the required scopes to access the requested resource.',
+        'description' => 'The current user or API key does not have the required scopes to access the requested resource. Update the permissions of the requested resource or add appropriate scopes to your API key.',
         'code' => 401,
     ],
     Exception::GENERAL_RATE_LIMIT_EXCEEDED => [
         'name' => Exception::GENERAL_RATE_LIMIT_EXCEEDED,
-        'description' => 'Rate limit for the current endpoint has been exceeded. Please try again after some time.',
+        'description' => 'Rate limit for the current endpoint has been exceeded. Please try again after some time. Learn more about rate-limits here https://appwrite.io/docs/rate-limits.',
         'code' => 429,
     ],
     Exception::GENERAL_SMTP_DISABLED => [
         'name' => Exception::GENERAL_SMTP_DISABLED,
-        'description' => 'SMTP is disabled on your Appwrite instance. You can <a href="/docs/email-delivery">learn more about setting up SMTP</a> in our docs.',
+        'description' => 'Email delivery is not possible because SMTP is disabled on your Appwrite instance. You can learn more about setting up SMTP here: https://appwrite.io/docs/email-delivery.',
         'code' => 503,
     ],
     Exception::GENERAL_PHONE_DISABLED => [
         'name' => Exception::GENERAL_PHONE_DISABLED,
-        'description' => 'Phone provider is not configured. Please check the _APP_SMS_PROVIDER environment variable of your Appwrite server.',
+        'description' => 'SMS delivery is not possible because a phone provider is not configured. You can learn more about setting up SMS delivery here: https://appwrite.io/docs/sms-delivery.',
         'code' => 503,
     ],
     Exception::GENERAL_ARGUMENT_INVALID => [
         'name' => Exception::GENERAL_ARGUMENT_INVALID,
-        'description' => 'The request contains one or more invalid arguments. Please refer to the endpoint documentation.',
+        'description' => 'The argument {argument} is invalid. Please refer to the endpoint documentation {docs}.',
         'code' => 400,
     ],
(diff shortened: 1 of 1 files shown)


app/config/errors.php
```diff
@@ -10,61 +10,64 @@
     /** General Errors */
     Exception::GENERAL_UNKNOWN => [
         'name' => Exception::GENERAL_UNKNOWN,
-        'description' => 'An unknown error has occured. Please check the logs for more information.',
+        'description' => 'An unknown error has occured. Please submit a support ticket at [TODO: Where do people get help?]() with reproduction steps for further assistance.',
```
gewenyu99, line 13: Is there a places we want to direct users who need support? Maybe an email address?

app/config/errors.php
```diff
@@ -10,61 +10,64 @@
...
         'code' => 500,
     ],
     Exception::GENERAL_MOCK => [
         'name' => Exception::GENERAL_MOCK,
-        'description' => 'General errors thrown by the mock controller used for testing.',
+        'description' => 'General errors thrown by the mock controller used for testing. If you\'re not building Appwrite from source, please raise an issue at [TODO: Where do people get help?]()',
```
gewenyu99, line 18: Is there a places we want to direct users who need support? Maybe an email address?

app/config/errors.php
```diff
@@ -10,61 +10,64 @@
...
         'code' => 400,
     ],
     Exception::GENERAL_ACCESS_FORBIDDEN => [
         'name' => Exception::GENERAL_ACCESS_FORBIDDEN,
-        'description' => 'Access to this API is forbidden.',
+        'description' => 'Access to this API is forbidden. Please [action to get access]',
```
gewenyu99, line 23: When is this error API ever used? Why would access be forbidden? What are practical next steps here?

app/config/errors.php
```diff
@@ -10,61 +10,64 @@
...
     Exception::GENERAL_PHONE_DISABLED => [
         'name' => Exception::GENERAL_PHONE_DISABLED,
-        'description' => 'Phone provider is not configured. Please check the _APP_SMS_PROVIDER environment variable of your Appwrite server.',
+        'description' => 'SMS delivery is not possible because a phone provider is not configured. You can learn more about setting up SMS delivery here: https://appwrite.io/docs/sms-delivery.',
         'code' => 503,
     ],
     Exception::GENERAL_ARGUMENT_INVALID => [
         'name' => Exception::GENERAL_ARGUMENT_INVALID,
-        'description' => 'The request contains one or more invalid arguments. Please refer to the endpoint documentation.',
+        // TODO: This is used by the DB, can we make this more specific instead of a generic error or add some kind of templating into the message?
```
gewenyu99, line 58: Argument invalid is very generic and difficult to debug, what would be a better way to handle these errors? I'm unfamiliar with the exact logic behind the many places that use this error.

app/config/errors.php
```diff
@@ -10,61 +10,64 @@
...
+        'description' => 'The argument {argument} is invalid. Please refer to the endpoint documentation {docs}.',
         'code' => 400,
     ],
     Exception::GENERAL_QUERY_LIMIT_EXCEEDED => [
         'name' => Exception::GENERAL_QUERY_LIMIT_EXCEEDED,
+        // TODO: Would it be possible to indicate which attribute? It is really hard to debug when it just says "current attribute"
```
gewenyu99, line 64: Is it possible to also say exactly which attribute exceeded the limit? It is much easier to debug when we can be more specific.

app/config/errors.php
```diff
@@ -10,61 +10,64 @@
...
         'description' => 'Query limit exceeded for the current attribute. Usage of more than 100 query values on a single attribute is prohibited.',
         'code' => 400,
     ],
     Exception::GENERAL_QUERY_INVALID => [
         'name' => Exception::GENERAL_QUERY_INVALID,
+        // TODO: Can we provide more detailed feedback about which part of the query is invalid?
```
gewenyu99, line 70: How do we parse queries? Is it possible to indicate exactly which part of a query is invalid? This message is vague.

app/config/errors.php
stnguyen90: This is kind of a catch-all exception. I'd like to see a GitHub issue for this with:

1. reproduction steps
2. if self-hosting, logs from the appwrite container
3. if on cloud, project ID so we can try to lookup the logs

Then, we can look into how to catch the generic exception and return an better exception.

app/config/errors.php
stnguyen90: I've never seen anyone ask about this exception. Maybe they should reach out on discord so we can have a discussion about what they're doing and why they're getting this.

app/config/errors.php
stnguyen90: Theoretically, this happens when someone tries to access the console or project APIs, but specifies a different project:

https://github.com/appwrite/appwrite/blob/27adf0788ff9809b9e3853480e498e5e169c81b2/app/controllers/api/console.php#L13

https://github.com/appwrite/appwrite/blob/27adf0788ff9809b9e3853480e498e5e169c81b2/app/controllers/api/projects.php#L45

We can say to please use `console` as the project for this request.

app/config/errors.php
```diff
@@ -75,17 +78,19 @@
     ],
     Exception::GENERAL_CURSOR_NOT_FOUND => [
         'name' => Exception::GENERAL_CURSOR_NOT_FOUND,
-        'description' => 'The cursor is invalid. This can happen if the item represented by the cursor has been deleted.',
+        'description' => 'The cursor used in pagination is invalid. This can happen if the item represented by the cursor has been deleted. Retry this request using another item from the previous as the cursor.',
         'code' => 400,
     ],
     Exception::GENERAL_SERVER_ERROR => [
+        // TODO: This is used often. Can we take a look at the usage here and make sure it's not abused?
```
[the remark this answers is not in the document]
stnguyen90: This is a generic 500. We could try to create a specific exception for each of these...or it might be better to just start with some of the most common ones.

app/config/errors.php
stnguyen90: This is also not currently used.

image

Search: https://github.com/search?q=repo%3Aappwrite%2Fappwrite+GENERAL_QUERY_INVALID&type=code

app/config/errors.php
stnguyen90: I don't think this exception is even used:

image

Search: https://github.com/search?q=repo%3Aappwrite%2Fappwrite%20GENERAL_QUERY_LIMIT_EXCEEDED&type=code

app/config/errors.php
stnguyen90: This description is not used/shown to the user when the API throws this error. The message is determined by what the validator found to be incorrect:

https://github.com/appwrite/appwrite/blob/27adf0788ff9809b9e3853480e498e5e169c81b2/app/controllers/api/databases.php#L2907

app/config/errors.php
gewenyu99: Okay cooolllll. Should this exist then?

app/config/errors.php
gewenyu99: That is better, I think for "contributor devex" this is important still. Thanks

app/config/errors.php
gewenyu99: Okay, this should never occur on Cloud so I'm gonna say "raise an issue on GitHub or reach out to us on Discord."

app/config/errors.php
gewenyu99: Sounds good,

app/config/errors.php
christyjacob4: @gewenyu99 yes. I remember that it was used when we first implemented it. 
https://github.com/appwrite/appwrite/blob/0.13.x/app/controllers/api/database.php#L1728-L1736

I think it was then moved to the query validator. Please check with @abnegate that we throw this error somewhere else. If we do, we can remove it from the error codes.

app/config/errors.php
christyjacob4: @gewenyu99 similar to my comment above. It was used when it was initially implemented. But it has then been moved to the queries validator. Lets check with @abnegate

app/config/errors.php
```diff
@@ -75,17 +78,19 @@
     ],
     Exception::GENERAL_CURSOR_NOT_FOUND => [
         'name' => Exception::GENERAL_CURSOR_NOT_FOUND,
-        'description' => 'The cursor is invalid. This can happen if the item represented by the cursor has been deleted.',
+        'description' => 'The cursor used in pagination is invalid. This can happen if the item represented by the cursor has been deleted. Retry this request using another item from the previous as the cursor.',
```
christyjacob4, line 81:
```suggestion
        'description' => 'The cursor used in pagination is invalid. This can happen if the item represented by the cursor has been deleted or doesn't exist. Retry this request using another item from the previous as the cursor.',
```
Also the last part `Retry this request using another item from the previous as the cursor.` is not very clear and looks like a mouthful to read & parse 👀

Issue #5670, opened 2023-06-06
🚀 Feature: restrict unsupported characters for project ID.


byawitz:
### 🔖 Feature description

As of now, Appwrite's using `flutter_web_auth_2` for managing OAuth2 login within any Flutter app.

This package has some limitations on what character project ID can have, as you can see [here](https://github.com/ThexXTURBOXx/flutter_web_auth_2#troubleshooting-callbackurlscheme).

It will be good to have one of these.
1. A complete restriction on what project id characters
2. No restriction at all. But, some note that says something like so:

> Be aware that Project ID cannot contain special characters if you're planning to use Flutter and OAuth2 login methods. Click [here](https://github.com/ThexXTURBOXx/flutter_web_auth_2#troubleshooting-callbackurlscheme) to learn more.

### 🎤 Pitch

This will avoid confusion in many cases, where users won't be able to debug it as there isn't any problem with Appwrite.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

joeyouss:
Hi
Thanks for opening this. Let me share this feedback with my team and keep this open meanwhile for community's feedback.

gewenyu99:
> > To be honest, I'm on board with just not allowing `-`, `+`, and `.`, I don't see a good reason for people to need other characters. I mean imagine the chaos of `，` vs `,` because we allow special commas :P Would also reduce bugs.
> > ;-)
> 
> That package also doesn't like uppercase characters.

😆 That's a little extreme, but understandable.

byawitz:
It looks like there's no need for restricting the upper-case letters, as Appwrite [uses](https://github.com/utopia-php/database/blob/main/src/Database/Adapter/MariaDB.php#L31)  `utf8mb4` for the charset. that will make MariaDB choose the `utf8mb4_general_ci` collation, so the user can just lowercase the project id.

Check this for more details: https://discord.com/channels/564160730845151244/1120835483849724036/1120908499036160090

joeyouss:
@byawitz we are waiting to review and merge the PR to fix this issue, once done we can close this! Thank you so much for opening this.

byawitz:
Cool 👍

PR #5743, opened 2023-06-23
Restrict characters for project ID

stnguyen90:
## What does this PR do?

Only allow lowercase alphanumeric chars and hyphens because the project ID is used as part of url
schemes to redirect users back to mobile apps
and certain characters are not allowed.

Closes: https://github.com/appwrite/appwrite/issues/5670

## Test Plan

Automated tests should pass.

I also confirmed i was able to create a new project while still accessing the project i created with the weird project ID.

## Related PRs and Issues

* https://github.com/appwrite/appwrite/issues/5670

## Checklist

### DIFF
--- a/app/controllers/api/projects.php
+++ b/app/controllers/api/projects.php
@@ -9,7 +9,7 @@ use Appwrite\Network\Validator\CNAME;
 use Utopia\Validator\Domain as DomainValidator;
 use Appwrite\Network\Validator\Origin;
 use Utopia\Validator\URL;
-use Appwrite\Utopia\Database\Validator\CustomId;
+use Appwrite\Utopia\Database\Validator\ProjectId;
 use Appwrite\Utopia\Response;
 use Utopia\Abuse\Adapters\TimeLimit;
 use Utopia\App;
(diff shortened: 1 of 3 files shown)


stnguyen90:
@TorstenDittmann, would you please review this when yo get a chance?

eldadfux:
Can't we just urlencode the ID where needed?

eldadfux:
I also wonder if we should block the ID `console`

stnguyen90:
@eldadfux, 

> Can't we just urlencode the ID where needed?

We use the project ID in the URL scheme for deep links. The URL scheme has a very limited list of allowed characters. The [RFC](https://www.rfc-editor.org/rfc/rfc3986#section-3.1) states:

>    Scheme names consist of a sequence of characters beginning with a
   letter and followed by any combination of letters, digits, plus
   ("+"), period ("."), or hyphen ("-").  Although schemes are case-
   insensitive, the canonical form is lowercase and documents that
   specify schemes must do so with lowercase letters.  An implementation
   should accept uppercase letters as equivalent to lowercase in scheme
   names (e.g., allow "HTTP" as well as "http") for the sake of
   robustness but should only produce lowercase scheme names for
   consistency.
>
>      scheme      = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )

> I also wonder if we should block the ID console

It looks like 'console' is already blocked as reserved:

image

but `Console` isn't:

image

PR #5764, opened 2023-07-04
Refactor workers

shimonewman:


.env
```diff
@@ -1,4 +1,4 @@
-_APP_ENV=development
+_APP_ENV=production
```
lohanidamodar, line 1: Why set to production any particular reason?

lohanidamodar (changes requested):
Looking in great shape already. couple of suggestions to think of.

.env
shimonewman: No reason, tried the cert worker, but i guess I need a live deployment

app/workers/audits.php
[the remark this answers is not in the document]
shimonewman: Yes , waited for the last sec (:

Issue #3891, opened 2022-09-21
🚀 Feature: Allow full server side impersonation


aviNrich:
### 🔖 Feature description

Hi,
In order to migrate complex apps such as ERP/CRM, I think we need a way to wrap/proxy appwrite from the server.
If everything that is possible from the client side will be available on the server side it will give huge benefit migrating complex apps. 
This way we will be less depend on `functions` and can wrap all database calls with our own logic.
I think most of the problem today is the ability to impersonate the use on the server side for unlimited time and do everything on behalf of him..

### 🎤 Pitch

1. Exposing server to server APIs for our customers is almost impossible today...
2. Imagine if server-side create session function will return a jwt token with unlimited time and this token will be used in the server the same way on as we use it from the client side, The variety of apps we can implement in appwrite is much more wide.  

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

Meldiron:
Not directly, but related to: https://github.com/appwrite/appwrite/issues/3120
Thanks for raising this 🙏

We have seen many different use cases, and we should address all as part of one feature. We will do our best to brainstorm a secure, easy-to-use feature to address them all.

izznatsir:
This will be really useful to use on server side js framework such as Remix and Sveltekit. A small e-commerce that I created implement the session creation on the server side. So I could create and manage a custom session cookie myself.

CannonPhelps:
Having this especially for sveltekit would be amazing as handing over client-side info to the server is overly complicated

stnguyen90:
This should be doable now by creating a [custom token](https://appwrite.io/docs/products/auth/custom-token) server side and exchanging it for a session. Then, use that session for any calls on behalf of the user.

PR #5777, opened 2023-07-06
feat: SSR

loks0n:
## What does this PR do?

**Session Secrets for API key authenticated clients**
- Adds the `secret` field to the Session response object when authenticated using API key, including the response of all existing Account session creation endpoints.
- Adds the `sessions` scope, used for session creation endpoints. It is added available to guests by default and will be optional for API key.

**Tokens**
- 🟢 New endpoint `Create session` in users service `POST /users/:userId/session`
- 🟢 New endpoint `Create token` in users service `POST /users/:userId/token`
- 🟢 New endpoint `Create session [from token]` in account service `POST /account/sessions/token`
- 🔴 Removes and aliases `PUT /account/sessions/phone`  to `POST /account/sessions/token`
- 🔴 Removes and aliases `PUT /account/sessions/magic-url` to `POST /account/sessions/token`

**Token for SSR OAuth**
- Appends URL parameters `userId` & `secret` from a generated token to the final OAuth2 redirect to allow SSR clients to exchange this for a session.

## Test Plan

- [x] e2e tests
- [x] manual w/o sdk
- [ ] manual w/ sdk

## Related PRs and Issues

[RFC](https://github.com/appwrite/rfc/pull/59)

## Checklist


tests/e2e/Services/Account/AccountBase.php
```diff
@@ -1389,6 +1390,7 @@ public function testCreateSessionWithMagicUrl($data): array
         $this->assertNotEmpty($response['body']);
         $this->assertNotEmpty($response['body']['$id']);
         $this->assertNotEmpty($response['body']['userId']);
+        $this->assertEmpty($response['body']['secret']);
```
christyjacob4, line 1393: Is it right to assert Empty here? 

The AccountBase class is used by AccountsCustomClientTest and AccountCustomServerTest. In the first case, the secret should be empty since the tests are executed in the client scope and should not be empty in the latter since its executed in the server scope

app/controllers/api/account.php
```diff
@@ -1076,39 +1101,40 @@
...
-    ->groups(['api', 'account', 'session'])
+App::put('/v1/account/sessions/token')
+    ->alias('/v1/account/sessions/magic-url')
+    ->alias('/v1/account/sessions/phone')
+    ->desc('Update token session')
+    ->groups(['api', 'account'])
     ->label('scope', 'public')
-    ->label('event', 'users.[userId].sessions.[sessionId].create')
-    ->label('audits.event', 'session.update')
+    ->label('auth.type', 'token')
```
christyjacob4, line 1124: This token needs to be added in the following places
1. https://github.com/appwrite/appwrite/blob/342deb49896842ace0a80b60da1d5048a8dd993e/app/controllers/shared/api.php#L280
2. https://github.com/appwrite/appwrite/blob/342deb49896842ace0a80b60da1d5048a8dd993e/app/controllers/shared/api/auth.php#L27

In addition I also see that the `phone` auth type is missing in the above two places. You might wanna add it there too 🤔

app/controllers/api/account.php
```diff
@@ -1076,39 +1101,40 @@
...
     ->label('audits.resource', 'user/{response.userId}')
     ->label('audits.userId', '{response.userId}')
     ->label('usage.metric', 'sessions.{scope}.requests.create')
-    ->label('usage.params', ['provider:magic-url'])
     ->label('sdk.auth', [])
     ->label('sdk.namespace', 'account')
-    ->label('sdk.method', 'updateMagicURLSession')
-    ->label('sdk.description', '/docs/references/account/update-magic-url-session.md')
-    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.method', 'updateTokenSession')
```
christyjacob4, line 1131: shouldn't this be `updateSessionToken`

christyjacob4 (commented):
Looks good for the most part. The direction looks good 👍

app/controllers/general.php
```diff
@@ -495,7 +495,7 @@ function router(App $utopia, Database $dbForConsole, SwooleRequest $swooleReques
                 ]);
 
                 $role = Auth::USER_ROLE_APPS;
-                $scopes = \array_merge($roles[$role]['scopes'], $key->getAttribute('scopes', []));
+                $scopes = \array_merge($roles[$role]['scopes'], $key->getAttribute('scopes', []), ['public']);
```
loks0n, line 498: API key authenticated clients need `public` scope in order to use account APIs.

Thoughts @christyjacob4 ?

app/config/roles.php
```diff
@@ -8,6 +8,7 @@
     'home',
     'console',
     'graphql',
+    'sessions',
```
loks0n, line 11: Create a new `sessions` scope.

This scopes is available to guests, but only to app users (API keys) if created with the scope.

app/config/roles.php
```diff
@@ -98,6 +100,6 @@
     ],
     Auth::USER_ROLE_APPS => [
         'label' => 'Applications',
-        'scopes' => ['global', 'health.read', 'graphql'],
+        'scopes' => ['global', 'sessions', 'health.read', 'graphql'],
```
loks0n, line 103: For testing, revert this

app/controllers/api/account.php
```diff
@@ -765,27 +769,48 @@
...
+                Permission::read(Role::user($user->getId())),
+                Permission::update(Role::user($user->getId())),
+                Permission::delete(Role::user($user->getId())),
+            ]));
+
+        $dbForProject->deleteCachedDocument('users', $user->getId());
+
+        $state['success'] = URLParser::parse($state['success']);
+        $query = URLParser::parseQuery($state['success']['query']);
+        $query['secret'] = $tokenSecret;
```
stnguyen90, line 799: I think this would be a breaking change 😬

app/controllers/api/account.php
```diff
@@ -1072,39 +1097,40 @@
...
 
-App::put('/v1/account/sessions/magic-url')
-    ->desc('Create magic URL session (confirmation)')
-    ->groups(['api', 'account', 'session'])
-    ->label('scope', 'public')
-    ->label('event', 'users.[userId].sessions.[sessionId].create')
-    ->label('audits.event', 'session.update')
+App::put('/v1/account/sessions/token')
+    ->alias('/v1/account/sessions/magic-url')
+    ->alias('/v1/account/sessions/phone')
```
stnguyen90, line 1116: we might have gotten rid of aliases 🧐

src/Appwrite/Utopia/Response/Model/Session.php
```diff
@@ -160,6 +160,12 @@ public function __construct()
                 'default' => false,
                 'example' => true,
             ])
+            ->addRule('secret', [
+                'type' => self::TYPE_STRING,
+                'description' => 'Secret used to authenticate the user. Only included if the request was made with an API key',
+                'default' => '',
+                'example' => '5e5bb8c16897e',
+            ])
```
stnguyen90, line 168: We're going to need to be very careful and not leak this 😬

app/controllers/api/users.php
```diff
@@ -1080,6 +1080,61 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+        $secret = Auth::tokenGenerator();
+        $expire = DateTime::formatTz(DateTime::addSeconds(new \DateTime(), Auth::TOKEN_EXPIRATION_CONFIRM));
+
+        $token = new Document([
+            '$id' => ID::unique(),
+            'userId' => $user->getId(),
+            'userInternalId' => $user->getInternalId(),
+            'type' => Auth::TOKEN_TYPE_UNIVERSAL,
+            'secret' => Auth::hash($secret), // One way hash encryption to protect DB leak
+            'expire' => $expire,
```
stnguyen90, line 1118: I think the expiration time for the universal token needs to be really short to reduce the risk of an unauthorized user using it.

app/controllers/api/users.php
```diff
@@ -1080,6 +1080,61 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
         $response->dynamic(new Document($prefs), Response::MODEL_PREFERENCES);
     });
 
+App::post('/v1/users/:userId/tokens')
+    ->desc('Create session token')
```
stnguyen90, line 1084: What's the use case for this again? I can see this being used for [full server-side impersonation](https://github.com/appwrite/appwrite/issues/3891), but it would require an additional API call to exchange the token for a session.

app/controllers/api/account.php
loks0n: The tests are working for me right now?

app/controllers/api/users.php
loks0n: Mainly for using a custom messaging method for authentication.
For example, if you want to implement login via WhatsApp link.
Server side impersonation is also a good use case.

app/controllers/api/users.php
loks0n: Had a think about this and added a default of 15 minutes, alongside an optional parameter `expire` to set the seconds until expiry

src/Appwrite/Utopia/Response/Model/Session.php
loks0n: We could do it differently:
How about: 
1) adding a new rule option in Model object called `admin` (or similiar) that defaults to false
2) In the response->dynamic() function we could filter out fields with admin=true, unless the user is privileged

app/controllers/api/account.php
stnguyen90: It would be best to do it in a backwards compatible way to give people time to update their code.

app/controllers/api/account.php
stnguyen90: @TorstenDittmann, did we get rid of aliases?

src/Appwrite/Utopia/Response/Model/Session.php
stnguyen90: Maybe...I'm not sure what the best approach is. Maybe we need to make sure we have tests to catch it? 🤷🏼‍♂️

app/controllers/api/users.php
stnguyen90: An additional API call for the exchange when doing full server-side impersonation might be a pain, though.

app/controllers/api/account.php
```diff
@@ -1072,39 +1107,41 @@
...
     ->label('audits.resource', 'user/{response.userId}')
     ->label('audits.userId', '{response.userId}')
     ->label('usage.metric', 'sessions.{scope}.requests.create')
-    ->label('usage.params', ['provider:magic-url'])
     ->label('sdk.auth', [])
     ->label('sdk.namespace', 'account')
-    ->label('sdk.method', 'updateMagicURLSession')
-    ->label('sdk.description', '/docs/references/account/update-magic-url-session.md')
-    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.method', ['exchangeTokenForSession', 'updateMagicURLSession', 'updatePhoneSession'])
```
Meldiron, line 1138: `exchangeTokenForSession` doesnt follow our patterns. I think `createSession ` would make more sense.

loks0n: Slightly concerned about this as we already have `createSession` on the users API

app/controllers/api/account.php
```diff
@@ -1123,154 +1170,192 @@
...
+    ->label('sdk.namespace', 'account')
+    ->label('sdk.method', 'createSession')
+    ->label('sdk.description', '/docs/references/account/create-session.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_SESSION)
+    ->label('abuse-limit', 10)
+    ->label('abuse-key', 'ip:{ip},userId:{param-userId}')
+    ->param('userId', '', new CustomId(), 'User ID.')
+    ->param('secret', '', new Text(256), 'Valid verification token.')
```
Meldiron, line 1335:
```suggestion
    ->param('secret', '', new Text(256), 'Secert of a token generated when authentication flow was initialized.')
```

Meldiron: Something along those lines

app/controllers/api/users.php
```diff
@@ -1420,6 +1422,186 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+    ->groups(['api', 'users'])
+    ->label('event', 'users.[userId].sessions.create')
+    ->label('scope', 'users.write')
+    ->label('audits.event', 'session.create')
+    ->label('audits.resource', 'user/{request.userId}')
+    ->label('usage.metric', 'sessions.{scope}.requests.create')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'users')
+    ->label('sdk.method', 'createSession')
+    ->label('sdk.description', '/docs/references/users/create-user-session.md')
```
Meldiron, line 1436: We are missing this markdown file. Also lets rename to just `create-session`, its in `users` folder already.

I think it will be worth doing a link between `create-session` and `create-token`, and explain difference. For example something like `Use this to create session for user without authentication flow. bla bla bla... If you instead want to include authentication flow, please use [link to createToken] instead.`. And something similar for create token, linking here if you want to go without authentication.

app/controllers/api/users.php
```diff
@@ -1420,6 +1422,186 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+    ->inject('project')
+    ->inject('locale')
+    ->inject('geodb')
+    ->inject('queueForEvents')
+    ->action(function (string $userId, Request $request, Response $response, Database $dbForProject, Document $project, Locale $locale, Reader $geodb, Event $queueForEvents) {
+        $user = $dbForProject->getDocument('users', $userId);
+
+        if ($user->isEmpty()) {
+            throw new Exception(Exception::USER_NOT_FOUND);
+        }
```
Meldiron, line 1453: We agreed to do magic URL behaviour here, which is:

If there is no user, create the user. Make it empty (like anonymous).

app/controllers/api/users.php
```diff
@@ -1420,6 +1422,186 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+    ->label('audits.resource', 'user/{request.userId}')
+    ->label('usage.metric', 'sessions.{scope}.requests.create')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'users')
+    ->label('sdk.method', 'createSession')
+    ->label('sdk.description', '/docs/references/users/create-user-session.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_SESSION)
+    ->param('userId', '', new UID(), 'User ID.')
```
Meldiron, line 1440: Not sure, but we might need different validator here that will support normal IDs, custom IDs as well as `unique()`. Check other endpoints like createDocument

app/controllers/api/users.php
```diff
@@ -1420,6 +1422,186 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+    ->inject('geodb')
+    ->inject('queueForEvents')
+    ->action(function (string $userId, Request $request, Response $response, Database $dbForProject, Document $project, Locale $locale, Reader $geodb, Event $queueForEvents) {
+        $user = $dbForProject->getDocument('users', $userId);
+
+        if ($user->isEmpty()) {
+            throw new Exception(Exception::USER_NOT_FOUND);
+        }
+
+        $secret = Auth::codeGenerator();
```
Meldiron, line 1455: In magic URL we use `$tokenSecret = Auth::tokenGenerator(Auth::TOKEN_LENGTH_MAGIC_URL);`

Lets use same here, define a const for it. like `TOKEN_LENGTH_CUSTOM`. I think 32 is reasonable length.

app/controllers/api/users.php
```diff
@@ -1420,6 +1422,186 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+        $detector = new Detector($request->getUserAgent('UNKNOWN'));
+        $record = $geodb->get($request->getIP());
+
+        $duration = $project->getAttribute('auths', [])['duration'] ?? Auth::TOKEN_EXPIRATION_LOGIN_LONG;
```
Meldiron, line 1459: This is 1 year. I dont like tokens being valid for so long. I would use `TOKEN_EXPIRATION_CONFIRM`, which is 1 hour. should be more than enough.

When we get feedback, we could (and should, for 1.6) make length and expiry configurable. Not only here, in many places.

app/controllers/api/users.php
```diff
@@ -1420,6 +1422,186 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'users')
+    ->label('sdk.method', 'createToken')
+    ->label('sdk.description', '/docs/references/users/create-token.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_TOKEN)
+    ->param('userId', '', new UID(), 'User ID.')
+    ->param('length', 6, new Range(4, 128), 'Token length in chars.', true)
+    ->param('expire', Auth::TOKEN_EXPIRATION_UNIVERSAL, new Range(1, Auth::TOKEN_EXPIRATION_LOGIN_LONG), 'Token expiration in seconds from now.', true)
```
Meldiron, line 1513: Lets set minimum to 60 instead of 1. Auth0 uses that minimum

Meldiron: also `from now` doesnt sound nessessary.

app/controllers/api/users.php
```diff
@@ -1420,6 +1422,186 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+    ->label('audits.resource', 'user/{request.userId}')
+    ->label('usage.metric', 'tokens.requests.create')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'users')
+    ->label('sdk.method', 'createToken')
+    ->label('sdk.description', '/docs/references/users/create-token.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_TOKEN)
+    ->param('userId', '', new UID(), 'User ID.')
```
Meldiron, line 1511: Should be `CustomId` validator I think, to allow `unique()`. I might be wrong, please double check.

app/controllers/api/users.php
```diff
@@ -1420,6 +1422,186 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+        }
+
+        $secret = Auth::tokenGenerator($length);
+        $expire = DateTime::formatTz(DateTime::addSeconds(new \DateTime(), $expire));
+
+        $token = new Document([
+            '$id' => ID::unique(),
+            'userId' => $user->getId(),
+            'userInternalId' => $user->getInternalId(),
+            'type' => Auth::TOKEN_TYPE_GENERIC,
```
Meldiron, line 1583: Here we call it GENERIC but above we call it UNIVERSAL. I would rename them to use same, for sick of consistency. Unless I am missing something.

app/controllers/api/users.php
Meldiron: If we do this, we need UID to be CusotmId validator on param. I think.

docs/references/users/create-token.md
```diff
@@ -0,0 +1 @@
+Returns a custom token with a secret key for creating a session. If the provided user ID has not be registered, a new user will be created. Use the returned user ID and secret and submit a request to the [PUT /account/sessions/custom](https://appwrite.io/docs/references/cloud/client-web/account#updateCustomSession) endpoint to complete the login process. By default, the secret sent in the response is valid for 15 minutes.
```
Meldiron, line 1: I think endpoint is different, no more `custom`.
I don't think we need sentence about expiry, it will be there on the param

app/controllers/api/users.php
Meldiron, line 1513: Lets mention what are defaults in the descriptions

docs/references/account/create-email-session.md
Meldiron, line 1: I think its createEmailPassword session now

docs/references/account/create-session.md
```diff
@@ -0,0 +1 @@
+Use this endpoint to complete creating the session with a custom session token. Both the **userId** and **secret** arguments will be returned in the response body when sending your request to the [POST /users/:userId/sessions](https://appwrite.io/docs/references/cloud/server-nodejs/users#createToken) endpoint.
```
Meldiron, line 1: `custom` not needed here anymore, we resue it for all types of tokens.

We should not point directly to createToken .we just say that userId and secret will be returned from request to create a token, or by the flow initiated by token creation (meaning that with phone auth, you get it in your SMS)

src/Appwrite/Auth/Auth.php
```diff
@@ -52,6 +52,8 @@ class Auth
     public const TOKEN_TYPE_INVITE = 4;
     public const TOKEN_TYPE_MAGIC_URL = 5;
     public const TOKEN_TYPE_PHONE = 6;
+    public const TOKEN_TYPE_OAUTH2 = 7;
+    public const TOKEN_TYPE_GENERIC = 8;
```
Meldiron, line 56: Generic or universal? 🤔

src/Appwrite/Auth/Auth.php
```diff
@@ -60,6 +62,9 @@ class Auth
     public const SESSION_PROVIDER_ANONYMOUS = 'anonymous';
     public const SESSION_PROVIDER_MAGIC_URL = 'magic-url';
     public const SESSION_PROVIDER_PHONE = 'phone';
+    public const SESSION_PROVIDER_OAUTH2 = 'oauth2';
+    public const SESSION_PROVIDER_TOKEN = 'token';
+    public const SESSION_PROVIDER_SERVER = 'server';
```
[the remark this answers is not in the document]
Meldiron: Hmm whats difference between provider token and server? I might be missing something

app/controllers/api/account.php
```diff
@@ -1123,154 +1172,192 @@
...
-        $countryName = $locale->getText('countries.' . strtolower($session->getAttribute('countryCode')), $locale->getText('locale.country.unknown'));
+    $response->dynamic($session, Response::MODEL_SESSION);
+};
 
-        $session
-            ->setAttribute('current', true)
-            ->setAttribute('countryName', $countryName)
-            ->setAttribute('expire', $expire);
+App::put('/v1/account/sessions/magic-url')
+    ->alias('/v1/account/sessions/phone')
```
Meldiron, line 1286: Can we find a way to hide it from docs?

docs/references/account/create-anonymous-session.md
Meldiron, line 1: We have `create-session-oauth2`. Lets find consistency. Either session always second, or always last

app/controllers/api/users.php
loks0n: Added

app/controllers/api/users.php
loks0n: Done 👍

app/controllers/api/users.php
loks0n: Yup, you're correct 👍

app/controllers/api/users.php
loks0n: Went with `GENERIC`

docs/references/users/create-token.md
loks0n: Updated this

app/controllers/api/users.php
loks0n: Done

docs/references/account/create-email-session.md
loks0n: Updated

docs/references/account/create-session.md
loks0n: Updated

src/Appwrite/Auth/Auth.php
loks0n: Prefer generic, went with this

src/Appwrite/Auth/Auth.php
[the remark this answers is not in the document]
loks0n: SESSION_PROVIDER_SERVER refers to a session created by `users.createSession`
SESSION_PROVIDER_TOKEN refers to a session create by `account.createSession` from a token created by `users.createToken`

docs/references/account/create-anonymous-session.md
loks0n: Changed to `create-session-oauth2`

app/controllers/api/account.php
loks0n: We just need to not give examples in `docs/examples/${version}/**/*.md`

Meldiron (approved):
AMAZING PR 🔥🔥🔥

tests/e2e/Services/Users/UsersBase.php
```diff
@@ -230,6 +230,27 @@ public function testCreateUserSessionHashed(array $data): void
...
+        $token = $this->client->call(Client::METHOD_POST, '/users/' . $data['userId'] . '/tokens', array_merge([
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()), [
+            'expire' => 60,
+        ]);
+
+        $this->assertEquals(201, $token['headers']['status-code']);
+        $this->assertEquals($data['userId'], $token['body']['userId']);
+        $this->assertNotEmpty($token['body']['secret']);
+        $this->assertNotEmpty($token['body']['expire']);
```
christyjacob4, line 250: what are the possible values for the `expires` parameter? Can we have some test cases to test for failure?

src/Appwrite/Utopia/Request.php
```diff
@@ -25,7 +25,11 @@ public function getParams(): array
         $parameters = parent::getParams();
 
         if (self::hasFilter() && self::hasRoute()) {
-            $endpointIdentifier = self::getRoute()->getLabel('sdk.namespace', 'unknown') . '.' . self::getRoute()->getLabel('sdk.method', 'unknown');
+            $method = self::getRoute()->getLabel('sdk.method', ['unknown']);
+            if (\is_array($method)) {
+                $method = $method[0];
+            }
```
christyjacob4, line 31: in what scenario is this an array

tests/e2e/Services/Account/AccountCustomServerTest.php
```diff
@@ -6,35 +6,154 @@
...
 use Utopia\Database\Helpers\ID;
 
 class AccountCustomServerTest extends Scope
 {
     use ProjectCustom;
     use SideServer;
 
-    public function testCreateAccount(): array
+
+    public function testCreateAnonymousAccount()
```
christyjacob4, line 18: what about the email and password route? Is that not accessible to the server?

src/Appwrite/Platform/Tasks/Specs.php
```diff
@@ -108,6 +114,24 @@ public function action(string $version, string $mode, Registry $register): void
...
+                    'in' => 'header',
+                ],
+                'ForwardedFor' => [
+                    'type' => 'apiKey',
+                    'name' => 'X-Forwarded-For',
+                    'description' => 'The IP address of the client that made the request',
+                    'in' => 'header',
+                ],
+                'ForwardedUserAgent' => [
+                    'type' => 'apiKey',
```
christyjacob4, line 130: the type parameters need to be updated here. I dont think these are apiKeys

docs/references/account/create-oauth2-session.md
christyjacob4, line 1: What is the naming convention being followed here? 

I thought we would have
create phone token
create magic url token
create oauth2 token
create email password token 

and then a `create session` endpoint to accept any of the above tokens and create a session with it?

christyjacob4: Looking at the above pattern, anonymous seems like an edge case then? 🤔

app/console
christyjacob4, line 1: lets remove this change from the PR

app/controllers/api/users.php
```diff
@@ -1420,6 +1422,235 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
             ->dynamic($target, Response::MODEL_TARGET);
     });
 
+App::post('/v1/users/:userId/sessions')
+    ->desc('Create session')
+    ->groups(['api', 'users'])
+    ->label('event', 'users.[userId].sessions.create')
```
christyjacob4, line 1428: This needs to be `users.[userId].sessions.[sessionId].create`

app/controllers/api/users.php
```diff
@@ -1420,6 +1422,235 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+    ->inject('project')
+    ->inject('locale')
+    ->inject('geodb')
+    ->inject('queueForEvents')
+    ->action(function (string $userId, Request $request, Response $response, Database $dbForProject, Document $project, Locale $locale, Reader $geodb, Event $queueForEvents) {
+        $user = $dbForProject->getDocument('users', $userId);
+
+        if ($user !== false && !$user->isEmpty()) {
+            $user->setAttributes($user->getArrayCopy());
+        } else {
```
christyjacob4, line 1453: what is the need for this ?

app/controllers/api/users.php
```diff
@@ -1420,6 +1422,235 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+                'sessions' => null,
+                'tokens' => null,
+                'memberships' => null,
+                'search' => implode(' ', [$userId]),
+                'accessedAt' => DateTime::now(),
+            ]);
+
+            $user->removeAttribute('$internalId');
+            Authorization::skip(fn () => $dbForProject->createDocument('users', $user));
+        }
```
christyjacob4, line 1502: This is a `createSession` endpoint right? I see this is performing 2 functions
1. Creating a user 
2. Creating a session for the user

this goes against REST principles where each endpoint should handle only one responsibility unless absolutely necessary. 

What is the need to create a user here when we already have the `createUser` endpoint?

This `createSession` endpoint should return 404 if the user does not exist

app/controllers/api/users.php
```diff
@@ -1420,6 +1422,235 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+                'passwordUpdate' => null,
+                'registration' => DateTime::now(),
+                'reset' => false,
+                'prefs' => new \stdClass(),
+                'sessions' => null,
+                'tokens' => null,
+                'memberships' => null,
+                'search' => implode(' ', [$userId]),
+                'accessedAt' => DateTime::now(),
+            ]);
```
christyjacob4, line 1619: same comment as above. The endpoint should only perform one function.

app/controllers/api/users.php
```diff
@@ -1420,6 +1422,235 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+
+        return $response
+            ->setStatusCode(Response::STATUS_CODE_CREATED)
+            ->dynamic($session, Response::MODEL_SESSION);
+    });
+
+App::post('/v1/users/:userId/tokens')
+    ->desc('Create token')
+    ->groups(['api', 'users'])
+    ->label('event', 'users.[userId].tokens.create')
```
christyjacob4, line 1548: This event needs to be added to the `events.php` file

christyjacob4: And it should also be `users.[userId].tokens.[tokenId].create`

app/controllers/api/users.php
christyjacob4, line 1: How are these two endpoints supposed to be used? What is the usecase? 

I was under the impression that we have `createToken*()` endpoints in the accounts API and then use that token to create a session using the `createSession*()` endpoints also from the accounts API

app/controllers/shared/api.php
```diff
@@ -311,6 +311,18 @@
                 }
                 break;
 
+            case 'token':
+                if (($auths['token'] ?? true) === false) {
+                    throw new Exception(Exception::USER_AUTH_METHOD_UNSUPPORTED, 'Token authentication is disabled for this project');
+                }
+                break;
```
christyjacob4, line 318: This needs to be added to the auth.php file in the config directory. Else it will break the `updateAuthStatus` endpoint

christyjacob4: Also what is "token" based auth? How does it sit well in this screen? 
Screenshot 2024-01-16 at 1 35 49 AM

I am to understand that tokens are created by emailPassword, phone, magic URL etc. So if we add a new check box for `token` here, what does it mean to disable `token` based auth? 

does it mean, we also disable emailPassword, phone, magicURL etc? Because a token is generated by all those methods right?

docs/references/account/create-email-password-session.md
christyjacob4, line 1: We need to fix the naming convention in these files. 

When we have a set of words, we try to make sure they maintain a common prefix.
In this case it would be `create-session-anonymous`, `create-session-email-password`, `create-session-oauth`, `create-token-magic-url`, `create-token-phone`

docs/references/account/update-magic-url-session.md
christyjacob4, line 1: `update-session-magic-url`

docs/references/account/update-phone-session.md
christyjacob4, line 1: `update-session-phone`

app/config/auth.php
```diff
@@ -42,7 +42,7 @@
         'name' => 'Phone',
         'key' => 'phone',
         'icon' => '/images/users/phone.png',
-        'docs' => 'https://appwrite.io/docs/client/account?sdk=web-default#accountCreatePhoneSession',
+        'docs' => 'https://appwrite.io/docs/references/cloud/client-web/account#accountCreatePhoneToken',
         'enabled' => true,
     ],
```
christyjacob4, line 47: we're missing a new array item for `token`

app/config/scopes.php
```diff
@@ -1,6 +1,12 @@
 <?php
 
 return [ // List of publicly visible scopes
+    'account' => [
+        'description' => 'Access to make actions on behalf of a user account',
+    ],
```
christyjacob4, line 6: Can you give me some reasons why this is not `account.read` and `account.write`

app/config/scopes.php
```diff
@@ -1,6 +1,12 @@
...
+    'sessions' => [
```
christyjacob4, line 7: similarly why is this not `sessions.write` ?

app/config/scopes.php
loks0n: There was an existing hidden scope called `account`.
I wanted keep this name to reduce the scope of the changes.

app/config/scopes.php
loks0n: We can change this.

app/controllers/api/users.php
loks0n: We have:

- `users.createSession` Returns a session for a user ID. Could be used by a customer service admin to assume control of a users account.

- `users.createToken` Returns a token for a user ID. Could be used to implement a custom authentication method, for example, send the token secret via discord chat.

- `account.createSession` Returns a session for a token and user ID. Used as part of existing token based authentication flows.

docs/references/account/create-email-password-session.md
loks0n: Done

app/controllers/api/users.php
loks0n: Updated

src/Appwrite/Utopia/Request.php
loks0n: https://github.com/appwrite/appwrite/blob/77b1762514125c319b2ae4d34f063812c86229d2/app/controllers/api/account.php#L198

We use array to define alias methods for the SDK to maintain backwards compatibility.

app/controllers/api/users.php
loks0n: I agree it seems to violate REST principle, but it offers pontentially better DX, and Matej proposed this change justifying it as the existing behaviour for endpoints in account service.

Should we:
1. Auto-create users for endpoints in account service and NOT users service?
2. Auto-create users for endpoints in both account and users services?
3. Change accounts service so we never auto-create users?

src/Appwrite/Platform/Tasks/Specs.php
loks0n: This field is currently unused in SDK generators. Given these objects are headers only string makes sense.

Therefore I think we should remove it and make it implicitly string

docs/references/account/create-oauth2-session.md
loks0n: We have two(ish) kinds of authentication methods.

Token flows endpoints:
- create phone token
- create magic url token

Session flows:
- create anon session
- create email session

Oauth is special, as by default it is a session flow, but becomes a token flow if you pass `token=true` as a param

app/controllers/api/users.php
loks0n: Updated

tests/e2e/Services/Users/UsersBase.php
loks0n: Added some more test cases

tests/e2e/Services/Account/AccountCustomServerTest.php
loks0n: Added a test for it :)

app/console
loks0n: Done

app/controllers/api/users.php
loks0n: Done

app/controllers/api/users.php
loks0n: Removed

app/config/scopes.php
loks0n: Changed

app/config/scopes.php
loks0n: Changed

app/controllers/shared/api.php
loks0n: Good point, I think we should remove these. It doesn't fit in this screen. 

The user can already control where a token is created with the existed toggles.
If we let them disable token auth it would break magic url and phone auth.

app/controllers/api/account.php
```diff
@@ -3021,6 +2995,7 @@
     ->inject('dbForProject')
     ->inject('queueForEvents')
     ->action(function (string $userId, string $secret, Response $response, Document $user, Database $dbForProject, Event $queueForEvents) {
+        var_dump($secret);
```
christyjacob4, line 2998: lets remove this

app/controllers/api/account.php
```diff
@@ -2875,7 +2848,8 @@
         $roles = Authorization::getRoles();
         $isPrivilegedUser = Auth::isPrivilegedUser($roles);
         $isAppUser = Auth::isAppUser($roles);
-        $verificationSecret = Auth::tokenGenerator();
+        $verificationSecret = Auth::tokenGenerator(Auth::TOKEN_LENGTH_VERIFICATION);
+        var_dump($verificationSecret);
```
christyjacob4, line 2852: lets remove this too

app/controllers/api/account.php
loks0n: 👍

PR #5781, opened 2023-07-07
Track a user's last activity

stnguyen90:
## What does this PR do?

A user will have an accessedAt timestamp that will update at most once per day if they make some API call. This timestamp can then be used find active users and calculate daily, weekly, and monthly active users.

## Test Plan

Automated tests

## Related PRs and Issues

Blocks:

- https://github.com/appwrite/console/pull/475

Depends on:

- https://github.com/utopia-php/database/pull/289

## Checklist


app/controllers/shared/api.php
```diff
@@ -551,6 +552,22 @@
...
+        if (!$user->isEmpty()) {
+            $accessedAt = $user->getAttribute('accessedAt', '');
+            if (DateTime::formatTz(DateTime::addSeconds(new \DateTime(), -APP_USER_ACCCESS)) > $accessedAt) {
+                $user->setAttribute('accessedAt', DateTime::now());
+
+                if (APP_MODE_ADMIN !== $mode) {
+                    $dbForProject->updateDocument('users', $user->getId(), $user);
+                } else {
+                    $dbForConsole->updateDocument('users', $user->getId(), $user);
+                }
```
TorstenDittmann, line 568: I dont think we need that condition 👍🏻

app/controllers/api/account.php
```diff
@@ -476,10 +479,15 @@
...
-        ]) : $user;
+        if ($user->isEmpty()) {
+            $session = $dbForProject->findOne('sessions', [ // Get user by provider id
+                Query::equal('provider', [$provider]),
+                Query::equal('providerUid', [$oauth2ID]),
+            ]);
+            if ($session !== false && !$session->isEmpty()) {
+                $user->setAttributes($dbForProject->getDocument('users', $session->getAttribute('userId'))->getArrayCopy());
+            }
+        }
```
stnguyen90, line 490: Should we really try to look up a user by provider? That sounds dangerous...Maybe we should delete this block altogether.

composer.json
```diff
@@ -49,7 +49,7 @@
         "utopia-php/cache": "0.8.*",
         "utopia-php/cli": "0.13.*",
         "utopia-php/config": "0.2.*",
-        "utopia-php/database": "0.37.*",
+        "utopia-php/database": "dev-feat-document-set-attributes as 0.37.99",
```
eldadfux, line 52: Reminder to add tag.

eldadfux (approved):

app/controllers/api/account.php
eldadfux: What exactly is this part responsible for?

stnguyen90: This is the old code that (I assume) tried to find the user from an old session with matching provider values:

https://github.com/appwrite/appwrite/blob/c8689d1d416716e74a47c95957f96d15ea23baf0/app/controllers/api/account.php#L479-L482

However, this never worked because:

1. an unauthenticated user doesn't have access to another user's sessions
2. assigning a session to $user wouldn't work

composer.json
stnguyen90: Updated

app/controllers/shared/api.php
```diff
@@ -551,6 +544,22 @@
...
+         */
+        if (!$user->isEmpty()) {
+            $accessedAt = $user->getAttribute('accessedAt', '');
+            if (DateTime::formatTz(DateTime::addSeconds(new \DateTime(), -APP_USER_ACCCESS)) > $accessedAt) {
+                $user->setAttribute('accessedAt', DateTime::now());
+
+                if (APP_MODE_ADMIN !== $mode) {
+                    $dbForProject->updateDocument('users', $user->getId(), $user);
+                } else {
+                    $dbForConsole->updateDocument('users', $user->getId(), $user);
```
christyjacob4, line 559: These database updates are better handled in a worker. We are adding avoidable latency to every authenticated request

app/controllers/shared/api.php
TorstenDittmann: but `$user` in this case should still be the admin user right?

TorstenDittmann: oh, I think you are right :-D

eldadfux (approved):

PR #5785, opened 2023-07-07
doc: Update getting started examples for .NET SDK

adityaoberai:
## What does this PR do?

Update the getting started example for .NET:
- Align code to C# coding conventions
- Add missing using directives
- Correct playground link title

## Related PRs and Issues

- No

## Checklist


docs/sdks/dotnet/GETTING_STARTED.md
```diff
@@ -1,44 +1,47 @@
...
 ```csharp
 using Appwrite;
+using Appwrite.Services;
+using Appwrite.Models;
 
-static async Task Main(string[] args)
-{
-  var client = Client();
+var client = new Client()
+  .SetEndpoint("http://cloud.appwrite.io/v1")  // Make sure your endpoint is accessible
```
gewenyu99, line 12: You don't need to make sure endpoint is accessible anymore ;) It's cloud

docs/sdks/dotnet/GETTING_STARTED.md
```diff
@@ -1,44 +1,47 @@
...
+  .SetProject("5ff3379a01d25")                 // Your project ID
+  .SetKey("cd868db89")                         // Your secret API key
+  .SetSelfSigned();                            // Use only on dev mode with a self-signed SSL cert
```
gewenyu99, line 15: I don't think you should have this here if it's pointed to cloud.

gewenyu99 (commented):
Left some comments.

docs/sdks/dotnet/GETTING_STARTED.md
adityaoberai: Might still be a good note for people who replace it with self-hosted links :)

docs/sdks/dotnet/GETTING_STARTED.md
adityaoberai: Should we do that, or maybe we can update the comment to mention that it should only be used if the user has a self-hosted instance without a domain pointed to it?

gewenyu99: Our attitude at the moment is cloud first, so make this copy pasteable for cloud.

Maybe, we can split this into cloud and self-hosted (we should do for all SDKs) but you can leave a comment in experience channel later and I'll make it happen following ur example here ;)

adityaoberai: Made the update :)

christyjacob4 (approved):
Can you also update the sdk version in the platforms.php file ?

adityaoberai:
> Can you also update the sdk version in the platforms.php file ?

Updated 👍

christyjacob4:
The SDK generator version also needs to be updated in the composer and composer-lock files @adityaoberai

adityaoberai:
@christyjacob4 updated those too

christyjacob4 (approved):

christyjacob4 (approved):

PR #5800, opened 2023-07-11
Update method name for `updateMembershipRoles`

lohanidamodar:
## What does this PR do?

- rename `updateMembershipRoles` to `updateMembership`. REST API should remain the same

## Test Plan

N/A

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/teams.php
```diff
@@ -727,8 +727,8 @@
     ->label('audits.resource', 'team/{request.teamId}')
```
eldadfux, line 727: We also need to update `->desc()` value.

lohanidamodar: done!

eldadfux (approved):

abnegate (approved):

Issue #3905, opened 2022-09-22
Atomic Document Update Operations


merabtenei:
### 🔖 Feature description

There is a clear need of having some sort of atomic transactions at least for updating a field value based on its current value, like Incrementing or decrementing a field's value on the database.

### 🎤 Pitch

An example for this use case is Counters. In a blog post case we may want to count the number of likes or comments on every user operation (like or comment) by incrementing a counter (ex: likes_count, comments_count).
We can do that using server functions at the moment but there is no way to prevent race condition. (https://en.wikipedia.org/wiki/Race_condition)

Firestore have some solutions for this using either transactions or using an update based on the field's value.
```js
update({
'likes_count': FieldValue.increment(1),
})
```

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

Meldiron:
This became really requested feature recently. @eldadfux would it be possible to incorporate this into our near-future release plans?

---

@merabtenei This is what I did in my project, it may help solve your case in the meantime:
I have discussions which nest discussions recursively. Each discussion is either positive or negative. For each discussion, I need to know positive&negative ratio of all sub-discussions.

I created a function that updates counters on all parent discussions: https://github.com/Meldiron/almost-kialo/blob/master/backend/functions/onDiscussionCreate/src/index.js

This function now has the same problem as you - I am getting and updating a document, and there is a possibility of race condition: https://github.com/Meldiron/almost-kialo/blob/master/backend/functions/onDiscussionCreate/src/index.js#L65-L75

This only happens rarely with huge traffic, so I was not too worried. Also, it's must tiny metadata, not a huge feature.

Either way, I write a function that is triggered by CRON every 60 minutes. This function goes over all discussions and calculates their upvotes and downvotes based on how many discussions there are. This is a pretty slow function, tho it gets everything back in sync: https://github.com/Meldiron/almost-kialo/blob/master/backend/functions/aggregateVotes/src/index.js

These two combined create a counting solution that is always only a few numbers off, if not exact. I understand if it feels like extra work, and that is why I fully support this feature idea.

merabtenei:
@Meldiron I totally agree, this may not happen with small traffic, but we can not rely on this, if the project gets big it would be really hard to change things from there.
[...]

PR #5803, opened 2023-07-11
Add APIs to atomically increment/decrement numeric attributes

stnguyen90:
## What does this PR do?

Fixes https://github.com/appwrite/appwrite/issues/3905

## Test Plan

E2E tests added

## Related PRs and Issues

- 3905

## Checklist


app/controllers/api/databases.php
```diff
@@ -3439,6 +3440,282 @@ function updateAttribute(
...
+        $collection = Authorization::skip(fn() => $dbForProject->getDocument('database_' . $database->getInternalId(), $collectionId));
+
+        if ($collection->isEmpty() || !$collection->getAttribute('enabled')) {
+            if (!($mode === APP_MODE_ADMIN && Auth::isPrivilegedUser(Authorization::getRoles()))) {
+                throw new Exception(Exception::COLLECTION_NOT_FOUND);
+            }
+        }
+
+        // Validate attribute type
+        $attribute = $collection->find('key', $key, 'attributes');
```
fogelito, line 3495: Do we need to check the attribute status?

app/controllers/api/databases.php
```diff
@@ -3439,6 +3440,282 @@ function updateAttribute(
...
+        if (!$attribute) {
+            throw new Exception(Exception::DOCUMENT_INVALID_STRUCTURE, "Attribute '$key' not found");
+        }
+
+        $validator = match ($attribute->getAttribute('type')) {
+            Database::VAR_INTEGER => new Integer(),
```
fogelito, line 3501: If we use the find method above,  can we filter "type" for integers and floats?

app/controllers/api/databases.php
```diff
@@ -3439,6 +3440,282 @@ function updateAttribute(
...
+        $processDocument($collection, $document);
+
+        $events
+            ->setParam('databaseId', $databaseId)
+            ->setParam('collectionId', $collection->getId())
+            ->setParam('documentId', $document->getId())
+            ->setContext('collection', $collection)
+            ->setContext('database', $database)
+            ->setPayload($response->output($document, Response::MODEL_DOCUMENT));
+
```
fogelito, line 3577: Will it make sense to return only the updated value?

app/controllers/api/databases.php
```diff
@@ -3439,6 +3440,282 @@ function updateAttribute(
...
+    ->param('documentId', '', new UID(), 'Document ID.')
+    ->param('key', '', new Key(), 'Attribute Key.')
+    ->param('value', null, new Numeric(), 'Value to decrement attribute by')
+    ->inject('requestTimestamp')
+    ->inject('response')
+    ->inject('dbForProject')
+    ->inject('events')
+    ->inject('mode')
+    ->action(function (string $databaseId, string $collectionId, string $documentId, string $key, int|float $value, ?\DateTime $requestTimestamp, Response $response, Database $dbForProject, Event $events, string $mode) {
+        if ($value <= 0) {
```
fogelito, line 3614: We can use a Range validator?

app/controllers/api/databases.php
fogelito: or query for "available"

app/controllers/api/databases.php
stnguyen90: That find method works differently than the database find so no we can't. Anyways, I think it's good to provide a clear error message that the attribute was found, but the type was incorrect.

stnguyen90: No, I don't think that [find](https://github.com/utopia-php/database/blob/4035d3f7e3393385eabc7816055047659c6fb4d3/src/Database/Document.php#L265) can filter like that.

app/controllers/api/databases.php
```diff
@@ -3439,6 +3440,282 @@ function updateAttribute(
...
+
+        $events
+            ->setParam('databaseId', $databaseId)
+            ->setParam('collectionId', $collection->getId())
+            ->setParam('documentId', $document->getId())
+            ->setContext('collection', $collection)
+            ->setContext('database', $database)
+            ->setPayload($response->output($document, Response::MODEL_DOCUMENT));
+
+        $response->dynamic($document, Response::MODEL_DOCUMENT);
```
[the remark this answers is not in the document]
stnguyen90: That's how the update document works so 🤷🏼

app/controllers/api/databases.php
stnguyen90: We can, but `Range(0)` allows for 0, so we would need to check for 0 here. Would you rather me switch it to `Range()`?

app/controllers/api/databases.php
fogelito: you are right 👍

app/controllers/api/databases.php
fogelito: Yea forgot about the floats.
This looks bad:
new Range(0.00001, PHP_INT_MAX, \Utopia\Validator::TYPE_FLOAT)
I will resolve 👍

Issue #5768, opened 2023-07-05
🐛 Bug Report: Creating new team with existed teamId throws error 500


huy-lv:
### 👟 Reproduction steps

I create a new team with existed team id:
Method: Post
Url: `Host/v1/teams`
Body: 
```
{
    "teamId": "63fc4cbb567e0d7d7f7d", //this team id already existed
    "name": "admin"
}
```
Response: 
```
{
    "message": "Server Error",
    "code": 500,
    "type": "general_unknown",
    "version": "1.3.4"
}
```

### 👍 Expected behavior

This should throw error "Team ID already existed" or something like that, error code should not be 500

### 👎 Actual Behavior

Response Server Error code 500
Log from docker:
![image](https://github.com/appwrite/appwrite/assets/7688070/b4643442-17c8-4ed5-a567-fd6426ae5996)

### 🎲 Appwrite version

Version 1.3.x

### 💻 Operating system

Linux

### 🧱 Your Environment

Appwrite 1.3.4

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

stnguyen90:
@huy-lv, thanks for creating this issue! 🙏 We should definitely return a clearer error here.

Abhijnan-Bajpai:
Hey there! I would like to try working on this. Could I be assigned?

jaivix:
@stnguyen90 would love to help with this, can you assign it to me?

sahilrawat001:
@stnguyen90 just a check must be include to overcome this error .Can you assign this issue to me

stnguyen90:
@jaivix, assigned! Thanks for your interest! 🙏🏼

TheOjasSingh:
can I work on this issue

PR #5808, opened 2023-07-12
fixed team creation response if teamId already exists

jaivix:
## What does this PR do?

Fixes the linked issue 

## Test Plan
- Created a simple web project, tried to create team with same teamId - 
image

- Wrote test case for this - 
image

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/5768

## Checklist

### DIFF
--- a/src/Appwrite/Extend/Exception.php
+++ b/src/Appwrite/Extend/Exception.php
@@ -83,6 +83,7 @@ class Exception extends \Exception
     public const TEAM_INVALID_SECRET               = 'team_invalid_secret';
     public const TEAM_MEMBERSHIP_MISMATCH          = 'team_membership_mismatch';
     public const TEAM_INVITE_MISMATCH              = 'team_invite_mismatch';
+    public const TEAM_ID_CONFLICT                  = 'team_id_conflict';
 
     /** Membership */
     public const MEMBERSHIP_NOT_FOUND              = 'membership_not_found';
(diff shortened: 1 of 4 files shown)


src/Appwrite/Extend/Exception.php
```diff
@@ -83,6 +83,7 @@ class Exception extends \Exception
     public const TEAM_INVALID_SECRET               = 'team_invalid_secret';
     public const TEAM_MEMBERSHIP_MISMATCH          = 'team_membership_mismatch';
     public const TEAM_INVITE_MISMATCH              = 'team_invite_mismatch';
+    public const TEAM_ID_CONFLICT                  = 'team_id_conflict';
```
stnguyen90, line 86: It's always good to take a look at existing code to try and match existing conventions. In this case, for other similar errors, we have `<RESOURCE>_ALREADY_EXISTS` such as `TEAM_INVITE_ALREADY_EXISTS`. So, this should probably be:

```suggestion
    public const TEAM_ALREADY_EXISTS                = 'team_already_exists';
```

stnguyen90 (changes requested):

Also, it looks like the linter failed. Please run the formatter to fix the lint error.

jaivix:
not sure why test cases were failing since locally test cases seems to run fine, maybe because the repo wasn't synced with appwrite master, synced with master so this time the test cases should pass hopefully 🤞 
@stnguyen90 can you please review again.

eldadfux (approved):

jaivix:
thanks 🙏

PR #5815, opened 2023-07-13
Fix 5404 update doc without relation permission

### LINKED ISSUE #5404: 🐛 Bug Report: Nested update, permission issue
### 👟 Reproduction steps

**SETUP**
I have a collection Player that is related to the collection playerAttributes and playerAttributes is related to the collection Attributes:
Player -> playerAttributes -> Attributes.

The user A create multiples documents in Attributes. He has all permission on these documents.
The user B create a player and has all permission for the player and playerAttributes documents created.
However, the playerAttributes documents are linked to already created documents in Attributes and the user B only has read permission for the Attributes collection.

### 👍 Expected behavior

The user B should be able to update the player name as he has update permissions on the Player collection.

### 👎 Actual Behavior

When the user B want to update the player name he get the following error :

`{"message":"The current user is not authorized to perform the requested action.","code":401,"type":"user_unauthorized","version":"1.3.1"}`

This is because he doesn't have update permissions in the Attributes collection.

### 🎲 Appwrite version

Version 1.3.x

### 💻 Operating system

Linux

### 🧱 Your Environment

I used appwrite web SDK 

fanatic75:
## What does this PR do?

Fixes #5404  Update of document with relations without permissions of relation collection

## Test Plan

Added E2E test case

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/5404

## Checklist

### DIFF
--- a/app/controllers/api/databases.php
+++ b/app/controllers/api/databases.php
@@ -73,7 +73,7 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
     $default = $attribute->getAttribute('default');
     $options = $attribute->getAttribute('options', []);
 
-    $db = Authorization::skip(fn () => $dbForProject->getDocument('databases', $databaseId));
+    $db = Authorization::skip(fn() => $dbForProject->getDocument('databases', $databaseId));
 
     if ($db->isEmpty()) {
         throw new Exception(Exception::DATABASE_NOT_FOUND);
@@ -193,16 +193,14 @@ function createAttribute(string $databaseId, string $collectionId, Document $att
         ->setType(DATABASE_TYPE_CREATE_ATTRIBUTE)
         ->setDatabase($db)
         ->setCollection($collection)
-        ->setDocument($attribute)
-    ;
+        ->setDocument($attribute);
 
     $events
         ->setContext('collection', $collection)
         ->setContext('database', $db)
         ->setParam('databaseId', $databaseId)
         ->setParam('collectionId', $collection->getId())
-        ->setParam('attributeId', $attribute->getId())
-    ;
+        ->setParam('attributeId', $attribute->getId());
 
     $response->setStatusCode(Response::STATUS_CODE_CREATED);
 
@@ -224,7 +222,7 @@ function updateAttribute(
     array $elements = null,
     array $options = []
 ): Document {
-    $db = Authorization::skip(fn () => $dbForProject->getDocument('databases', $databaseId));
+    $db = Authorization::skip(fn() => $dbForProject->getDocument('databases', $databaseId));
 
     if ($db->isEmpty()) {
         throw new Exception(Exception::DATABASE_NOT_FOUND);
@@ -685,13 +683,11 @@ App::delete('/v1/databases/:databaseId')
 
         $deletes
             ->setType(DELETE_TYPE_DOCUMENT)
-            ->setDocument($database)
-        ;
+            ->setDocument($database);
 
         $events
             ->setParam('databaseId', $database->getId())
-            ->setPayload($response->output($database, Response::MODEL_DATABASE))
-        ;
+            ->setPayload($response->output($database, Response::MODEL_DATABASE));
 
         $response->noContent();
     });
@@ -2364,7 +2360,7 @@ App::post('/v1/databases/:databaseId/collections/:collectionId/indexes')
         }
 
         // Convert Document[] to array of attribute metadata
-        $oldAttributes = \array_map(fn($a) => $a->getArrayCopy(), $collection->getAttribute('attributes'));
+        $oldAttributes = \array_map(fn ($a) => $a->getArrayCopy(), $collection->getAttribute('attributes'));
 
         $oldAttributes[] = [
             'key' => '$id',
@@ -2538,7 +2534,7 @@ App::get('/v1/databases/:databaseId/collections/:collectionId/indexes/:key')
         $indexes = $collection->getAttribute('indexes');
 
         // Search for index
-        $indexIndex = array_search($key, array_map(fn($idx) => $idx['key'], $indexes));
+        $indexIndex = array_search($key, array_map(fn ($idx) => $idx['key'], $indexes));
 
         if ($indexIndex === false) {
             throw new Exception(Exception::INDEX_NOT_FOUND);
@@ -2737,7 +2733,7 @@ App::post('/v1/databases/:databaseId/collections/:collectionId/documents')
 
             $relationships = \array_filter(
                 $collection->getAttribute('attributes', []),
-                fn($attribute) => $attribute->getAttribute('type') === Database::VAR_RELATIONSHIP
+                fn ($attribute) => $attribute->getAttribute('type') === Database::VAR_RELATIONSHIP
             );
 
             foreach ($relationships as $relationship) {
@@ -2816,7 +2812,7 @@ App::post('/v1/databases/:databaseId/collections/:collectionId/documents')
 
             $relationships = \array_filter(
                 $collection->getAttribute('attributes', []),
-                fn($attribute) => $attribute->getAttribute('type') === Database::VAR_RELATIONSHIP
+                fn ($attribute) => $attribute->getAttribute('type') === Database::VAR_RELATIONSHIP
             );
 
             foreach ($relationships as $relationship) {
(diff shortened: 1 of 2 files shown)


app/controllers/api/databases.php
```diff
@@ -3302,9 +3303,31 @@ function updateAttribute(
...
+                                return true;
+                        }
+
+                        foreach ($oldRelationDocument as $key => $obj1) {
+                            $obj2 = $newRelationDocuemntFromRequestData[$key] ?? null;
+                            if (($obj1 instanceof Document && $obj2 instanceof Document && $obj1->getArrayCopy() !== $obj2->getArrayCopy())  || !($obj2 instanceof Document)) {
+                                // Return true if a difference is found in the relationships
+                                return true;
+                            }
+                        }
```
stnguyen90, line 3326: I didn't get a chance to test this, but what if we have a setup like this:

```
A <-- two way, one-to-one --> B <-- two way, one-to-one --> C
```

Where the user has update permissions to A and C and they attempt to update A and C, but not B like:

```json
{
  "someString": "updated value",
  "b": {
    "$id": "asdf",
    "someString": "unchanged value",
    "c": {
      "$id": "zxcv",
      "someString": "updated value"
    }
  }
}
```

What would happen?

stnguyen90: Never mind. Let's not worry about this case. We should just reject this case because the user doesn't have access to B. However, this API call should succeed:

```
{
  "someString": "updated value",
  "b": "any_id"
}
```

fanatic75: I think we should fix this case as well. It's perfectly fine for the user to update A and C through B right? 
I moved to a new logic as filtering out relationships is not a good way. 

I have changed my logic to skip Authorization if there is no changes to the document instead of filtering out relationships as I cannot go to C if B is filtered out.

app/controllers/api/databases.php
```diff
@@ -3280,7 +3276,6 @@ function updateAttribute(
         }
 
         $data = \array_merge($document->getArrayCopy(), $data); // Merge existing data with new data
-        $data['$collection'] = $collection->getId();            // Make sure user doesn't switch collectionID
```
abnegate, line 3283: Do we need to remove this?

app/controllers/api/databases.php
```diff
@@ -3366,8 +3382,29 @@ function updateAttribute(
...
+            break;
+        }
+    }
+    if ($skipCheckingPermission) {
+        Authorization::skip(
+            fn() => $checkPermissions($collection, $newDocument, $document, Database::PERMISSION_UPDATE)
+        );
+    } else {
         $checkPermissions($collection, $newDocument, $document, Database::PERMISSION_UPDATE);
+        ;
```
abnegate, line 3405: Let's remove this

app/controllers/api/databases.php
```diff
@@ -3366,8 +3382,29 @@ function updateAttribute(
...
+    }
+
```
abnegate, line 3407: Let's keep just one line between code blocks

tests/e2e/Services/Databases/DatabasesCustomClientTest.php
```diff
@@ -316,4 +317,146 @@ public function testUpdateTwoWayRelationship(): void
         $this->assertEquals($relation['body']['relatedCollection'], $collection1RelationAttribute['relatedCollection']);
         $this->assertEquals('restrict', $collection1RelationAttribute['onDelete']);
     }
+
+    public function testUpdateWithoutRelationPermission(): void
+    {
+
```
abnegate, line 323: Let's have no blank lines at the start of function

tests/e2e/Services/Databases/DatabasesCustomClientTest.php
```diff
@@ -316,4 +317,146 @@ public function testUpdateTwoWayRelationship(): void
...
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'x-appwrite-key' => $this->getProject()['apiKey']
+        ], [
+            'databaseId' => ID::unique(),
+            'name' => ID::unique(),
+        ]);
+
+        $databaseId = $database['body']['$id'];
+
```
abnegate, line 341: Let's keep just one line between code blocks

tests/e2e/Services/Databases/DatabasesCustomClientTest.php
```diff
@@ -316,4 +317,146 @@ public function testUpdateTwoWayRelationship(): void
...
+        ]), [
+            'collectionId' => ID::unique(),
+            'name' => ID::unique(),
+            'documentSecurity' => false,
+            'permissions' => [
+                Permission::read(Role::user($userId)),
+            ]
+        ]);
+
+        \sleep(2);
```
abnegate, line 374: We don't need to sleep after collection creation as it's done synchronously (only attribute and index creation are async)

tests/e2e/Services/Databases/DatabasesCustomClientTest.php
```diff
@@ -316,4 +317,146 @@ public function testUpdateTwoWayRelationship(): void
...
+            'x-appwrite-key' => $this->getProject()['apiKey']
+        ]), [
+            'relatedCollectionId' => $collection2['body']['$id'],
+            'type' => 'oneToMany',
+            'twoWay' => false,
+            'onDelete' => 'setNull',
+            'key' => $collection2['body']['$id']
+        ]);
+
+        \sleep(2);
```
abnegate, line 389: We can remove this too

tests/e2e/Services/Databases/DatabasesCustomClientTest.php
```diff
@@ -316,4 +317,146 @@ public function testUpdateTwoWayRelationship(): void
...
+            'x-appwrite-key' => $this->getProject()['apiKey']
+         ]), [
+            'key' => "Rating",
+            'size' => 100,
+            'required' => false,
+            'array' => false,
+            'default' => null,
+         ]);
+
+        \sleep(3);
```
abnegate, line 415: We should be able to reduce this to 2 or 1

tests/e2e/Services/Databases/DatabasesCustomClientTest.php
```diff
@@ -316,4 +317,146 @@ public function testUpdateTwoWayRelationship(): void
...
+
```
abnegate, line 416: Let's keep just one blank line between code blocks

tests/e2e/Services/Databases/DatabasesCustomClientTest.php
```diff
@@ -316,4 +317,146 @@ public function testUpdateTwoWayRelationship(): void
...
+            'x-appwrite-key' => $this->getProject()['apiKey']
+        ]), [
+            'documentId' => ID::unique(),
+            'data' => [
+                'Rating' => '10',
+            ],
+            'permissions' => [],
+        ]);
+
+        \sleep(3);
```
abnegate, line 430: Let's remove this too

tests/e2e/Services/Databases/DatabasesCustomClientTest.php
```diff
@@ -316,4 +317,146 @@ public function testUpdateTwoWayRelationship(): void
...
+            'documentId' => ID::unique(),
+            'data' => [
+                'Title' => 'Captain America',
+                $collection2['body']['$id'] => [$childDocument['body']['$id']]
+            ],
+            'permissions' => [],
+        ]);
+        $this->assertEquals(201, $parentDocument['headers']['status-code']);
+
+        \sleep(3);
```
abnegate, line 447: Let's remove this too

app/controllers/api/databases.php
```diff
@@ -3337,25 +3334,44 @@ function updateAttribute(
...
+                                //No need to compare values of relations as for each relation we are recursively checking permission.
+                                if ($relatedDocumentOldVersion->getAttribute($key) instanceof Document) {
+                                    continue;
+                                }
+                                //If any of the values are different, we need to check permission.
+                                if ($relatedDocumentOldVersion->getAttribute($key) !== $value) {
+                                    $skipCheckingPermission = false;
+                                    $relation->removeAttribute('$collectionId');
+                                    $relation->removeAttribute('$databaseId');
+                                    $relation->setAttribute('$collection', $relatedCollection->getId());
```
abnegate, line 3363: I think we can do this outside the foreach loop

app/controllers/api/databases.php
```diff
@@ -3366,8 +3382,29 @@ function updateAttribute(
...
+    foreach ($newDocument as $key => $value) {
+        if ($document->getAttribute($key) instanceof Document) {
+            continue;
+        }
+        //If any of the values are different, we need to check permission.
+        if ($newDocument->getAttribute($key) !== $value) {
+            $skipCheckingPermission = false;
+            $newDocument->removeAttribute('$collectionId');
+            $newDocument->removeAttribute('$databaseId');
+            $newDocument->setAttribute('$collection', $collection->getId());
```
abnegate, line 3395: I think we can do this outside the foreach loop

PR #5825, opened 2023-07-17
Adding the ability to change Appwrite hard limit

### LINKED ISSUE #5506: 🐛 Bug Report: Increase Pagination maximum offset from 5000
### 👟 Reproduction steps

Pagination: The maximum offset is 5000. But in some cases, i want use offset > 5000 . I want have default max offset = 5000 and i can config max offset from ENV.

### 👍 Expected behavior

Pagination: The maximum offset is 5000. But in some cases, i want use offset > 5000 . I want have default max offset = 5000 and i can config max offset from ENV.

### 👎 Actual Behavior

Pagination: The maximum offset is 5000. But in some cases, i want use offset > 5000 . I want have default max offset = 5000 and i can config max offset from ENV.

### 🎲 Appwrite version

Version 1.3.x

### 💻 Operating system

Linux

### 🧱 Your Environment

_No response_

byawitz:
Closes #5506 

## What does this PR do?
Gives the ability to change Appwrite database-related limit.

## Test Plan

Check that it worked.

## Related PRs and Issues

#5506

## Checklist


app/controllers/api/account.php
```diff
@@ -1421,7 +1421,7 @@
 
         $queries = Query::parseQueries($queries);
         $grouped = Query::groupByType($queries);
-        $limit = $grouped['limit'] ?? APP_LIMIT_COUNT;
+        $limit = $grouped['limit'] ?? (int) App::getEnv('_APP_LIMIT_COUNT', APP_LIMIT_COUNT);
```
christyjacob4, line 1424: Instead of changing  it in every file, it might be worth considering changing it in one place here

https://github.com/appwrite/appwrite/blob/c8689d1d416716e74a47c95957f96d15ea23baf0/app/init.php#L87

```php
define("APP_LIMIT_COUNT", App::getEnv('_APP_LIMIT_COUNT', 5000));
```

My only issue with this approach is its not consistent with the way we define other constants.. 🤔

byawitz: Agreed!

That's why I've chosen the first approach.

christyjacob4: Let's wait for more feedback from the team, maybe we can make an exception for this or refactor all the constants to use `define`

abnegate: I'm not opposed to changing all to `define`, some other constants may need to be set at runtime in future

byawitz: I can create new PR changing all `const`s to `define`s

PR #5830, opened 2023-07-17
Bump clamav to get it working again

stnguyen90:
## What does this PR do?

freshclam is used to properly update the virus database. It's best to use a separate container so that the clamav container is not impacted by database updates.

## Test Plan

Manual:

## Related PRs and Issues

- https://github.com/appwrite/docker-clamav/pull/21

## Checklist

### DIFF
--- a/app/views/install/compose.phtml
+++ b/app/views/install/compose.phtml
@@ -633,13 +633,26 @@ services:
     volumes:
       - appwrite-redis:/data:rw
 
+  # freshclam:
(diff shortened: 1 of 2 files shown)


eldadfux:
This is nice, but I think the problem we had back in the day was that every database update that didn't sync with our clamAV version broke the setup. 

Maybe we can run this setup on a server for a few weeks and have a script to monitor it's working as expected. If it does, I think we can enable it back by default.

HarshMN2345:

PR #5846, opened 2023-07-19
change to use find instead of iterating index array in Get Index route

fanatic75:
## What does this PR do?

Uses findOne method in getIndex route to fetch an index instead of iterating index array

## Test Plan

No need of new test case as it's the same output just use a different method to achieve it.

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/databases.php
```diff
@@ -2535,20 +2535,19 @@ function updateAttribute(
             throw new Exception(Exception::COLLECTION_NOT_FOUND);
         }
 
-        $indexes = $collection->getAttribute('indexes');
-
-        // Search for index
-        $indexIndex = array_search($key, array_map(fn($idx) => $idx['key'], $indexes));
+        $index = $dbForProject->findOne('indexes', [
+            Query::equal('$id', [$database->getInternalId() . '_' . $collection->getInternalId() . '_' . $key])
+        ]);
```
stnguyen90, line 2540: I don't think we should use `findOne()` because it incurs a db operation while the array_search is a local operation.

app/controllers/api/databases.php
```diff
@@ -2535,20 +2535,19 @@ function updateAttribute(
             throw new Exception(Exception::COLLECTION_NOT_FOUND);
         }
 
-        $indexes = $collection->getAttribute('indexes');
-
-        // Search for index
-        $indexIndex = array_search($key, array_map(fn($idx) => $idx['key'], $indexes));
```
stnguyen90, line 2541: Instead of doing getAttribute() and then searching through the result, we should probably be using `Document->find()`: https://github.com/utopia-php/database/blob/main/src/Database/Document.php#L280

app/controllers/api/databases.php
fanatic75: removed this

app/controllers/api/databases.php
fanatic75: done this

fanatic75:
@stnguyen90  please review whenever you get some time

app/controllers/api/databases.php
```diff
@@ -2524,31 +2524,15 @@ function updateAttribute(
...
-        $database = Authorization::skip(fn() => $dbForProject->getDocument('databases', $databaseId));
-
-        if ($database->isEmpty()) {
-            throw new Exception(Exception::DATABASE_NOT_FOUND);
-        }
-        $collection = $dbForProject->getDocument('database_' . $database->getInternalId(), $collectionId);
-
-        if ($collection->isEmpty()) {
-            throw new Exception(Exception::COLLECTION_NOT_FOUND);
-        }
```
stnguyen90, line 2536: But these are helpful to give feedback that the databaseId and collectionId passed in are correct 🧐 

Also, the `getDocument()` calls here aren't too bad because `getDocument()` goes to the cache rather than the database.

fanatic75: Got it, added them back

fanatic75:
@stnguyen90  please review again

app/controllers/api/databases.php
```diff
@@ -2535,21 +2535,16 @@ function updateAttribute(
             throw new Exception(Exception::COLLECTION_NOT_FOUND);
         }
 
-        $indexes = $collection->getAttribute('indexes');
-
-        // Search for index
-        $indexIndex = array_search($key, array_map(fn($idx) => $idx['key'], $indexes));
+        $index = $dbForProject->find('indexes', [Query::equal('$id', [$database->getInternalId() . '_' . $collection->getInternalId() . '_' . $key]), Query::limit(1)]);
```
stnguyen90, line 2538: Why do we need to make another database query for the index?

fanatic75: understood what you mean when using Document->find(). 
I have used the same now.

stnguyen90 (approved):

christyjacob4 (approved):

Issue #2616, opened 2022-01-15
📚 Documentation: Upgrade Instructions Unclear


stnguyen90:
### 💭 Description

Despite the [Upgrade docs](https://appwrite.io/docs/upgrade) saying:

> The first step is to install the latest version of Appwrite. Head to the directory where you first installed Appwrite. This is where you will find the **appwrite** directory and the **appwrite/docker-compose.yml** file. Make sure you are at the same level as the **appwrite** directory and then execute the following command

I had 2 instances today where people had confusion about where to run the upgrade command. 

[basatS7T said](https://discord.com/channels/564160730845151244/564160731327758347/931988183074365480):

> Yep, I saw that.  It is not clear to me at this point what that says

[godfather said](https://discord.com/channels/564160730845151244/564161855862603789/932001419349860403):

> ohh cool, one question i have to run update command inside appwrite directory (i.e. which has docker-compose.yml file)?

basatS7T suggested to include example paths to help clarify.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

eldadfux:
Created a PR here: https://github.com/appwrite/docs/pull/133/files I hope it makes things more clear.

stnguyen90:
I like this change. @eldadfux, what are your thoughts on adding examples to show that `appwrite` should be in the current directory?

eldadfux:
@stnguyen90 did you had any snippet in mind?

stnguyen90:
The user probably wanted to see something like this:

```bash
/path/to/parent/of/appwrite $ docker run -it --rm \
    --volume /var/run/docker.sock:/var/run/docker.sock \
    --volume "$(pwd)"/appwrite:/usr/src/code/appwrite:rw \
    --entrypoint="install" \
    appwrite/appwrite:0.12.1
```

but this doesn't match the convention of any of the other example commands so I don't think this should be done.

Perhaps the docs could be updated to:

> Head to the directory where you first installed Appwrite. This is the parent directory, where you will find the **appwrite** directory.
> ```bash
> ls appwrite
> ```
> Verify that shows the **docker-compose.yml** and **.env** files. Then, install the latest version of Appwrite.
> ```bash
> docker run -it --rm \
[...]

PR #5860, opened 2023-07-21
Create a dedicated upgrade command

stnguyen90:
## What does this PR do?

Before, we used the same command for both installation and upgrades. This lead to problems because developers would try to upgrade in the wrong folder and end up creating a new installation.

This new upgrade command validates the existence of an existing installation before proceeding with the upgrade to ensure no new installation is created when upgrading.

## Test Plan

The following screenshot shows:

1. Running the upgrade command w/o an installation results in an error
2. After adding an installation (appwrite folder w/ docker-compose.yml file), the install command runs as usual
   * The error at the end occurs because I'm running in an appwrite instance where the docker socket is unavailable. When someone tries to upgrade, they will have the docker socket available and this error won't show.

Screen Shot 2023-07-20 at 7 11 29 PM

Latest error output:

image

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/2616

## Checklist

### DIFF
--- a/src/Appwrite/Task/Setup.php
+++ b/src/Appwrite/Task/Setup.php
@@ -1,0 +1,271 @@
+<?php
+
+namespace Appwrite\Task;
+
+use Appwrite\Auth\Auth;
+use Appwrite\Docker\Compose;
+use Appwrite\Docker\Env;
+use Appwrite\Utopia\View;
+use Utopia\CLI\Console;
+use Utopia\Config\Config;
+use Utopia\Analytics\GoogleAnalytics;
+
+class Setup
+{
+    public const INSTALLATION_PATH = '/usr/src/code/appwrite';
+
+    private string $httpPort;
+    private string $httpsPort;
+    private string $organization;
+    private string $image;
+    private string $interactive;
+
+    public function __construct(string $httpPort, string $httpsPort, string $organization, string $image, string $interactive)
+    {
+        $this->httpPort = $httpPort;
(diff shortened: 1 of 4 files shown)


eldadfux (commented):
@stnguyen90 this is not exactly how we use class based tasks, we've created a new abstraction layer for them here: https://github.com/appwrite/cloud/tree/feat-billing/src/Appwrite/Cloud/Platform/Tasks we might be able to reuse this logic as we need to rewrite them anyway.

Please check with @lohanidamodar if we can adapt this approach here.

src/Appwrite/Task/Setup.php
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    public function upgrade()
+    {
+        // Check for previous installation
+        $data = @file_get_contents(SETUP::INSTALLATION_PATH . '/docker-compose.yml');
+        if (empty($data)) {
+            Console::error('Appwrite installation not found.');
+            Console::log('The command was not run in the parent folder of your appwrite installation.');
+            Console::log('Please navigate to the parent of the Appwrite installation and try again.');
```
gewenyu99, line 262:
```suggestion
            Console::log('Please navigate to the parent directory of the Appwrite installation and try again.');
```

gewenyu99 (changes requested):
LGTM, but I want to add one word to feel useful.

(Jokes aside I think we should be verbose here.

stnguyen90:
Need to restructure the task.

PR #5874, opened 2023-07-24
Feat upgrade task

stnguyen90:
## What does this PR do?

Before, we used the same command for both installation and upgrades. This lead to problems because developers would try to upgrade in the wrong folder and end up creating a new installation.

This new upgrade command validates the existence of an existing installation before proceeding with the upgrade to ensure no new installation is created when upgrading.

## Test Plan

The following screenshot shows:

Running the upgrade command w/o an installation results in an error:

image

Installation still works:

image

Upgrading with an existing installation:

image

image

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Tasks/Install.php
+++ b/src/Appwrite/Platform/Tasks/Install.php
@@ -198,8 +198,8 @@ class Install extends Action
             }
         }
 
-        $templateForCompose = new View(__DIR__ . '/../views/install/compose.phtml');
-        $templateForEnv = new View(__DIR__ . '/../views/install/env.phtml');
+        $templateForCompose = new View(__DIR__ . '/../../../../app/views/install/compose.phtml');
+        $templateForEnv = new View(__DIR__ . '/../../../../app/views/install/env.phtml');
 
         $templateForCompose
             ->setParam('httpPort', $httpPort)
--- a/src/Appwrite/Platform/Tasks/Install.php
+++ b/src/Appwrite/Platform/Tasks/Install.php
@@ -15,6 +15,8 @@ use Utopia\Platform\Action;
 
 class Install extends Action
 {
+    protected string $path = '/usr/src/code/appwrite';
+
     public static function getName(): string
     {
         return 'install';
@@ -51,7 +53,6 @@ class Install extends Action
          * 6. Run data migration
          */
         $config = Config::getParam('variables');
-        $path = '/usr/src/code/appwrite';
         $defaultHTTPPort = '80';
         $defaultHTTPSPort = '443';
         $vars = [];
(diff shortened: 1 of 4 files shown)


src/Appwrite/Platform/Tasks/Install.php
```diff
@@ -50,7 +53,6 @@ public function action(string $httpPort, string $httpsPort, string $organization
          * 6. Run data migration
          */
         $config = Config::getParam('variables');
-        $path = '/usr/src/code/appwrite';
         $defaultHTTPPort = '80';
```
eldadfux, line 56: Can we also add a confirmation dialog of an existing setup is detected. This way we can ask the developer to confirm if they want to overwrite it and understand the implications of it.

src/Appwrite/Platform/Tasks/Upgrade.php
```diff
@@ -0,0 +1,42 @@
...
+            Console::error('Appwrite installation not found.');
+            Console::log('The command was not run in the parent folder of your appwrite installation.');
+            Console::log('Please navigate to the parent directory of the Appwrite installation and try again.');
+            Console::log('  parent_directory <= you run the command in this directory');
+            Console::log('  └── appwrite');
+            Console::log('      └── docker-compose.yml');
+            Console::exit(1);
+        }
+        parent::action($httpPort, $httpsPort, $organization, $image, $interactive);
+    }
```
eldadfux, line 41: It could also be a good idea to drop the migration command and initiate the migration directly here. We could add a confirmation dialog.

stnguyen90: We should keep the migrate command the same since there can be scenarios where you only need to run the migrate (for example, if you manually upgrade by modifying the docker-compose.yml file).

I also prefer to avoid automatically running the migration after an upgrade. I think it would be better to focus on a specific thing (the upgrade should be focused on upgrading the docker-compose.yml and .env files and migrate on migrating the db schema). Long run, I think it would be best to decouple migrations from the upgrade even more so that you don't need to run the migration after upgrading, and maybe you can do them in pieces per project or something.

On another note, I hate how the install command automatically starts appwrite because the docker compose context is in the container rather than on the host machine.

src/Appwrite/Platform/Tasks/Install.php
stnguyen90: How's this?

install:

image

upgrade:

image

eldadfux: I like it.

eldadfux (approved):

src/Appwrite/Platform/Tasks/Upgrade.php
```diff
@@ -0,0 +1,42 @@
...
+        // Check for previous installation
+        $data = @file_get_contents($this->path . '/docker-compose.yml');
+        if (empty($data)) {
+            Console::error('Appwrite installation not found.');
+            Console::log('The command was not run in the parent folder of your appwrite installation.');
+            Console::log('Please navigate to the parent directory of the Appwrite installation and try again.');
+            Console::log('  parent_directory <= you run the command in this directory');
+            Console::log('  └── appwrite');
+            Console::log('      └── docker-compose.yml');
+            Console::exit(1);
```
lohanidamodar, line 38: It seems above we are setting `protected string $path = '/usr/src/code/appwrite';` so it doesn't where they run the install/upgrade command it's always looking for this path, so this log doesn't make sense. They don't need to run in any specific directory when we are not using their current directory. Was this intended?
 OR am I missing something?

lohanidamodar (approved):

PR #5639, opened 2023-06-02
Fix default project

stnguyen90:
## What does this PR do?

Don't default to console project
    
Defaulting to the console project leads to misleading error messages. For example, you can get a CORS error saying you haven't configured a platform yet, but the actual error is missing project ID.

## Test Plan

Passing unit and e2e tests

## Related PRs and Issues

Depends on:

* https://github.com/appwrite/sdk-generator/pull/681

## Checklist

- [x] Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [x] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

PR #5884, opened 2023-07-26
Remove the OAuth2 Callback endpoints from api group

stnguyen90:
## What does this PR do?

Prevent missing project ID error.

## Test Plan

All tests should pass

## Related PRs and Issues

- https://github.com/appwrite/appwrite/pull/5639

## Checklist


eldadfux:
This might lead to some future side-effects. This group is also taking care of event triggering, abuse control, and auth. In what cases exactly do we face this error?

Another approach could be to use a label that will allow these endpoints to skip the project ID check.

stnguyen90:
> This might lead to some future side-effects.

I don't think anything in the `api` init or shutdown are needed for these 2 endpoints because they're simple redirects to transform the request into something more suited for Appwrite. After redirecting to the redirect endpoint, the init and shutdown will run to do any required check.

>  In what cases exactly do we face this error?

Without this, any OAuth2 redirect will throw an unknown project error.

> Another approach could be to use a label that will allow these endpoints to skip the project ID check.

I'm just concerned we already have a lot of labels. Let me know if you'd prefer label rather than removing the api group and I'll update this.

stnguyen90:
@eldadfux, 

> This might lead to some future side-effects.

I don't think anything in the `api` init or shutdown are needed for these 2 endpoints because they're simple redirects to transform the request into something more suited for Appwrite. After redirecting to the redirect endpoint, the init and shutdown will run to do any required check.

>  In what cases exactly do we face this error?

Without this, any OAuth2 redirect will throw an unknown project error.

> Another approach could be to use a label that will allow these endpoints to skip the project ID check.

I'm just concerned we already have a lot of labels. Let me know if you'd prefer label rather than removing the api group and I'll update this.

eldadfux (approved):

PR #5885, opened 2023-07-26
Feat query support  in list attributes &  list indexes endpoint

fanatic75:
## What does this PR do?

Adds support for queries in the List Attributes endpoint and List Indexes endpoint

## Test Plan

Adds E2E test case to verify list attributes and list indexes with queries

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/databases.php
```diff
@@ -1673,11 +1676,54 @@ function updateAttribute(
...
 
-        $attributes = $collection->getAttribute('attributes');
+        $queries = Query::parseQueries($queries);
+        foreach ($queries as $query) {
+            if ($query->getMethod() ===  Query::TYPE_SELECT) {
+                throw new Exception(Exception::GENERAL_QUERY_INVALID, 'Select queries are not valid.');
+            }
+        }
+        \array_push($queries, Query::equal('collectionId', [$collectionId]), Query::equal('databaseId', [$databaseId]));
+
```
abnegate, line 1686: Let's just have one blank line

app/controllers/api/databases.php
```diff
@@ -1673,11 +1676,54 @@ function updateAttribute(
             throw new Exception(Exception::COLLECTION_NOT_FOUND);
         }
 
-        $attributes = $collection->getAttribute('attributes');
+        $queries = Query::parseQueries($queries);
+        foreach ($queries as $query) {
+            if ($query->getMethod() ===  Query::TYPE_SELECT) {
+                throw new Exception(Exception::GENERAL_QUERY_INVALID, 'Select queries are not valid.');
+            }
+        }
```
abnegate, line 1684: Why do we disallow select queries?

fanatic75: Select queries are not supported due to the limitation of no "key" in the metadata table, which is being used to validate select queries. Also, @TorstenDittmann  said most of the keys are not optional so it's not worth it. 
https://github.com/utopia-php/database/blob/a9f706034a12c7c87e6fd38eca69587fd3def0e0/src/Database/Database.php#L4509

app/controllers/api/databases.php
fanatic75: done this

app/controllers/api/databases.php
abnegate: What's the error if they're enabled? I think the `key` attribute should be there since it's in the collections config here: https://github.com/appwrite/appwrite/blob/887abad1904936bf9d328443598fec504c01107e/app/config/collections.php#L245

Select queries might not particularly useful for these resources but I would suggest we allow them if easy enough, so we can remove this code and keep the controller complexity to a minimum

fanatic75: Screenshot 2023-08-01 at 4 21 21 PM
Screenshot 2023-08-01 at 4 20 33 PM


I think if we use $id instead of $key in validating select queries then we can maybe use that because $id exists in all rows of metadata attributes and indexes

abnegate: We can try get the key and fallback to $id if it's not found, we're doing this a few other places I think

fanatic75: Made the changes. But sadly list attributes endpoint will have to be much more complex for  adding select queries support, we need to have type attribute for validating the response model for Attributes list. 

If select query does not include type support, we will have to add type in select query and then remove type property from output.  

discussed two options with @christyjacob4, either we use validator to not support select queries or we support select queries but we add type in select query when it doesn't include it. 

We decided to keep endpont consistent with select queries, we need to add it in select queries.

app/controllers/api/databases.php
```diff
@@ -1673,12 +1676,77 @@ function updateAttribute(
             throw new Exception(Exception::COLLECTION_NOT_FOUND);
         }
 
-        $attributes = $collection->getAttribute('attributes');
+        $queriesFromRequest = Query::parseQueries($queries);
```
abnegate, line 1679: Let's override `$queries` with the parsed value

app/controllers/api/databases.php
```diff
@@ -1673,12 +1676,77 @@ function updateAttribute(
...
+        ];
+        foreach ($queriesFromRequest as $query) {
+            if ($query->getMethod() ===  Query::TYPE_SELECT) {
+                $hasDetailsinQuery['selectQuery'] = true;
+            }
+            if (\array_search('type', $query->getValues())) {
+                $hasDetailsinQuery['typeAttributeInSelectQuery'] = true;
+            }
+        }
+        $transformedQueries = $queriesFromRequest;
```
abnegate, line 1694: We can remove this and push directly onto `$queries`

app/controllers/api/databases.php
```diff
@@ -1673,12 +1676,77 @@ function updateAttribute(
...
+            $cursor->setValue($cursorDocument[0]);
+        }
+
+        $attributes = $dbForProject->find('attributes', $transformedQueries);
+        $filterQueries = Query::groupByType($transformedQueries)['filters'];
+        $total = $dbForProject->count('attributes', $filterQueries, APP_LIMIT_COUNT);
+
+         //Add relationship data from options to attributes as it loses options during response setup
+        foreach ($attributes as $attribute) {
+            if ($attribute->getAttribute('type') === Database::VAR_RELATIONSHIP) {
```
abnegate, line 1725: Do we need the `type` for the response models or just to check if it's a relationship here?

app/controllers/api/databases.php
```diff
@@ -1673,12 +1676,77 @@ function updateAttribute(
...
+        $queriesFromRequest = Query::parseQueries($queries);
 
-        $response->dynamic(new Document([
-            'total' => \count($attributes),
-            'attributes' => $attributes
+        // Add type property in query if select query exists and type property doesn't exist as type is required for response model
+        $hasDetailsinQuery = [
+            'selectQuery' => false,
+            'typeAttributeInSelectQuery' => false,
+        ];
```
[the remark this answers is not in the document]
fanatic75: fixed this

app/controllers/api/databases.php
fanatic75: fixed this

app/controllers/api/databases.php
fanatic75: fixed this

app/controllers/api/databases.php
fanatic75: This logic adds relationship attributes to each attribute object as the options key gets filtered out from Attribute Relationship Type as defined in the model. 
I have moved this logic in the filter function of the Relationship Model so we can avoid this from controller logic.

abnegate: If we left it in the controller, would we be able to remove the type attribute after fetching from the database but before we call `$response->dynamic`? It would be much cleaner that way as opposed to calling `$response->output` then `$response->static` after filtering

fanatic75: This part of code is not really related to do with $type attribute. 

This is just to move from options key which contains relationship data to the appropriate keys that are required according to the models.
Regarding output and static 
That is not really possible, even I didn't like doing it but the whole reason of adding that type in select query is required because output method requires the type property. 

We cannot remove type property before output is run

src/Appwrite/Specification/Format/Swagger2.php
```diff
@@ -342,6 +342,8 @@ public function parse(): array
                     case 'Utopia\Validator\ArrayList':
                     case 'Appwrite\Utopia\Database\Validator\Queries\Buckets':
                     case 'Appwrite\Utopia\Database\Validator\Queries\Collections':
+                    case 'Appwrite\Utopia\Database\Validator\Queries\Indexes':
+                    case 'Appwrite\Utopia\Database\Validator\Queries\Attributes':
```
abnegate, line 346: We will also need to add these cases in `src/Appwrite/GraphQL/Types/Mapper.php`

src/Appwrite/Utopia/Database/Validator/Queries/Base.php
```diff
@@ -22,7 +23,7 @@ class Base extends Queries
      * @param string[] $allowedAttributes
      * @throws \Exception
      */
-    public function __construct(string $collection, array $allowedAttributes)
+    public function __construct(string $collection, array $allowedAttributes, ?array $prohibitedQueries = [])
```
abnegate, line 26: Since we want to iterate the array and it has a default of empty, let's just use `array` as the type so that `null` can't be passed

src/Appwrite/Utopia/Response.php
```diff
@@ -491,44 +491,45 @@ public function output(Document $document, string $model): array
...
 
-            if ($rule['array']) {
-                if (!is_array($document[$key])) {
-                    throw new Exception($key . ' must be an array of type ' . $rule['type']);
-                }
+            if ($data->isSet($key)) {
+                if ($rule['array']) {
+                    if (!is_array($document[$key])) {
+                        throw new Exception($key . ' must be an array of type ' . $rule['type']);
+                    }
```
abnegate, line 498: Do we need this change?

fanatic75: Yes there's a bug, 
We only wanna do this if the key exists, because if it's not required attribute and it doesn't exist, this code produces bug

src/Appwrite/Utopia/Database/Validator/Queries/Base.php
fanatic75: done this

src/Appwrite/Specification/Format/Swagger2.php
fanatic75: done this

src/Appwrite/Utopia/Response.php
fanatic75: Okay the error is when the attribute is an array and it is unset. We are throwing exception if the type is not array when rule says it is array, it's not array because it's optional and not set. 

we should throw exception when it's array and the value is not array. 
right now value is not set so it's throwing exception when it should not

this is the exception "orders must be an array of type string"

abnegate: Is this caused by using a select query but not selecting `orders`? I think the problem will be larger than this. For any resource, if we use select queries and don't select any attributes that are required, this function will set them to their default values even though they should not be included in the response at all.

I think we should block select queries for all resources _except_ documents. We can do this by removing `new Select($attributes)` from `Base`. We can then revert the changes here as well as the `$prohibitedQueries` in `Base`

fanatic75: I have removed the Select validator and the prohibitQueries from the Base validator.

I would still like to keep `$data->isSet($key)` because this seems like a reasonable condition for the code to work. 

No sense in throwing an exception that the rule is not an array if it's unset in the response since it's not a required rule

tests/e2e/Services/Databases/DatabasesBase.php
```diff
@@ -316,6 +316,23 @@ public function testCreateAttributes(array $data): array
...
+     */
+    public function testListAttributes(array $data): void
+    {
+        $databaseId = $data['databaseId'];
+        $attributes = $this->client->call(Client::METHOD_GET, '/databases/' . $databaseId . '/collections/' . $data['moviesId'] . '/attributes', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'x-appwrite-key' => $this->getProject()['apiKey'],
+        ]), [
+            'queries' => ['equal("type", "string")'],
```
christyjacob4, line 330: You can use the query class itself here.

tests/e2e/Services/Databases/DatabasesBase.php
```diff
@@ -1048,6 +1064,23 @@ public function testCreateIndexes(array $data): array
...
+     */
+    public function testListIndexes(array $data): void
+    {
+        $databaseId = $data['databaseId'];
+        $indexes = $this->client->call(Client::METHOD_GET, '/databases/' . $databaseId . '/collections/' . $data['moviesId'] . '/indexes', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'x-appwrite-key' => $this->getProject()['apiKey'],
+        ]), [
+            'queries' => ['equal("type", "key")'],
```
christyjacob4, line 1078: same as above

tests/e2e/Services/Databases/DatabasesBase.php
```diff
@@ -316,6 +316,23 @@ public function testCreateAttributes(array $data): array
...
+        $databaseId = $data['databaseId'];
+        $attributes = $this->client->call(Client::METHOD_GET, '/databases/' . $databaseId . '/collections/' . $data['moviesId'] . '/attributes', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'x-appwrite-key' => $this->getProject()['apiKey'],
+        ]), [
+            'queries' => ['equal("type", "string")'],
+        ]);
+        $this->assertEquals(200, $attributes['headers']['status-code']);
+        $this->assertEquals(4, $attributes['body']['total']);
```
christyjacob4, line 333: lets also add a test to make sure select queries are not allowed

christyjacob4: lets also try some other queries like limit , offset

composer.json
```diff
@@ -43,13 +43,13 @@
...
-        "utopia-php/abuse": "0.27.*",
+        "utopia-php/abuse": "0.29.*",
         "utopia-php/analytics": "0.10.*",
-        "utopia-php/audit": "0.29.*",
+        "utopia-php/audit": "0.31.*",
         "utopia-php/cache": "0.8.*",
         "utopia-php/cli": "0.15.*",
         "utopia-php/config": "0.2.*",
-        "utopia-php/database": "0.38.*",
+        "utopia-php/database": "0.40.*",
```
christyjacob4, line 52: we should update the db library here as well right ?

fanatic75: done this

tests/e2e/Services/Databases/DatabasesBase.php
fanatic75: done this

tests/e2e/Services/Databases/DatabasesBase.php
fanatic75: kept it same as it's done the same way in other tests

src/Appwrite/Utopia/Response.php
abnegate: If we block the select queries for indexes (and all other resources), I don't think the exception would be thrown. In the database, `orders` always has a value set, because when not included in the request to create an index, the default of an empty array is set. So the only way that exception could be thrown, is if using a select query and not selecting `orders`.

abnegate:
@fanatic75 Looks like we'll need to fix up some of the tests for other changes in the database library (select queries now no longer select internal attributes by default)

src/Appwrite/Utopia/Response.php
fanatic75: Hmm, that makes sense. But this condition is also reasonable. I think we should let this stay even if this case might not happen for us because, logically, this code block should run on set keys only, and we should check if the key is set or not in the output if the rule is an array.

fanatic75:
> @fanatic75 Looks like we'll need to fix up some of the tests for other changes in the database library (select queries now no longer select internal attributes by default)

I have fixed the tests, 
One thing I find weird is, If you select only a single attribute that has a null value, we get an error 404 that document is not found.

src/Appwrite/Utopia/Response.php
fanatic75: Removed this condition.

abnegate:
> One thing I find weird is, If you select only a single attribute that has a null value, we get an error 404 that document is not found.

@fanatic75 This will likely happen whenever `$id` is not selected, as the document empty check just looks at the `$id` value to determine if the document is empty. I've fixed this to check for any value instead [here](https://github.com/utopia-php/database/pull/305)

stnguyen90 (approved):
Looks good. Can this be merged into branch `cl-1.4.x`, though?

abnegate:
> Can this be merged into branch cl-1.4.x, though?

@stnguyen90 There's a lot of conflicts there, so we will merge this into 1.4.x and resolve them later all together :+1:

abnegate (approved):

christyjacob4 (approved):

Issue #5893, opened 2023-07-27
The Oauth sessions from deep links expires on mobile when two different apps runs on the same project


rohankm:
### 👟 Reproduction steps

The OAuth sessions from deep links expire on mobile when two different apps are running on the same project.

### 👍 Expected behavior

Each of the two different apps should have distinct sessions that do not get replaced by one another.

### 👎 Actual Behavior

At present, the sessions are being replaced by one another. This means that when one app creates a session, it overwrites the session of the other app, causing conflicts and undesirable behavior. To ensure that each app maintains its own separate session without interfering with the other, appropriate changes or improvements to the session handling mechanism should be implemented.

### 🎲 Appwrite version

Version 1.3.x

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

rohankm:
> Related: #1515

Even if we have different callbacks for the deeplink.. still the session conflict takes place.. bec the cookie is stored in the browser end which is unique per project.. so whenever the 2nd app logins the 1st apps session would be deleted...

the solution to this is to not set any cookies for the default success page.

gewenyu99:
@rohankm Hey, I'm following up on this. Looks like an interesting issue 😜 and I'm super curious how this happened.

Can I get more information about your setup?
- What platforms are you using
- Why do you need more than one apps on the same project.
- Why are you hitting session conflicts when deeplinking for two different apps, are the apps hosted on the same domain?

Thanks 👍 Once I know what your setup is like, I'll try to find someone who can look into this deeper.

rohankm:
Hi,

1. this affects android platform
2. there can be multiple reasons for having two apps on the same project (admin and user)
3. Yes the apps are hosted on the same domain

PR #5894, opened 2023-07-27
Fix 5893 oauth deeplink multiple sessions

rohankm:
## What does this PR do?

The default OAuth URL is primarily utilized as a deep link on mobile applications to establish sessions. However, the current approach has a flaw as it sets the cookies on the appwrite domain. This presents an issue, particularly when two mobile apps are concurrently running on the same project, as one app's session might overwrite another's.

To rectify this, it is advisable not to include the cookies in the default success link. By doing so, a session will not be created on the browser end, enabling two apps for the same project to function independently without any session conflicts.

## Test Plan

I have manually tested it with two different apps running on react native for the same project.

## Related PRs and Issues

#5893 

- (Related PR or issue)

## Checklist

- [yes] Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [no] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -613,15 +613,22 @@ App::get('/v1/account/sessions/oauth2/:provider/redirect')
             $query['secret'] = Auth::encodeSession($user->getId(), $secret);
             $state['success']['query'] = URLParser::unparseQuery($query);
             $state['success'] = URLParser::unparse($state['success']);
+            $response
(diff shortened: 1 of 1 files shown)


HarshMN2345:

Issue #4753, opened 2022-11-22
🐛 Bug Report: redirect to the failure url instead of throwing an error in OAuth2


gepd:
### 👟 Reproduction steps

1. Login with any OAuth2 Provider
2. Block the user in the console
3. Login with the OAuth2 provider again

### 👍 Expected behavior

I would expect to be redirected to the failure url. I have some doubt about if the failure url should have any extra information as for example, we have `USER_BLOCKED`, `USER_MISSING_ID`, `GENERAL_SERVER_ERROR` and many other errors, and go to the failure url without any other data will be too generic

### 👎 Actual Behavior

Currently if we block a user and this user try to login with any OAuth2 provider, you will get a scarry 401 error.
The current implementation makes imposible to catch this error

I have also noted that the error trace shows a lot of information, and some of them could be sensitive.

<img width="1237" alt="image" src="https://user-images.githubusercontent.com/7091609/203394630-06b2ca6c-c4c0-46bb-8431-3f7d34d64b62.png">

### 🎲 Appwrite version

Version 1.1.x

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

stnguyen90:
@gepd, thanks for raising this issue! 🙏🏼 Let me discuss this with the team

stnguyen90:
Related issue: https://github.com/appwrite/appwrite/issues/2591

PR #5895, opened 2023-07-28
Improve OAuth2 error handling

stnguyen90:
## What does this PR do?

There are two problems when the OAuth2 flow fails.

First, if there is a problem during authentication, the user will be redirected back to Appwrite without the authorization code, and Appwrite will throw an error like this:

image

The other problem occurs if Appwrite receives an authorization code but fails to exchange it for an access token. In that case, an error like this occurs:

image

This PR improves both scenarios to show more detailed info or send the data to the failure URL for the developer.

Without a failure URL, the user will see a more detailed error:

image

With a failure URL, the user will be redirected to it like:

```
https://localhost/auth/oauth2/failure?error=%7B%22message%22%3A%22Failed+to+obtain+access+token.+The+Google+OAuth2+provider+returned+an+error%3A+invalid_grant%3A+Malformed+auth+code.%22%2C%22type%22%3A%22user_oauth2_bad_request%22%2C%22code%22%3A400%7D#
```

The developer can then pull the `error` from the query string with something like:

```javascript
u = new URL(window.location);
e = JSON.parse(u.searchParams.get('error'));
```

which would yield:

```json
{
    "message": "Failed to obtain access token. The Google OAuth2 provider returned an error: invalid_grant: Malformed auth code.",
    "type": "user_oauth2_bad_request",
    "code": 400
}
```

## Test Plan

Manual

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/2591
- https://github.com/appwrite/appwrite/issues/4753

## Checklist

### DIFF
--- a/app/config/errors.php
+++ b/app/config/errors.php
@@ -195,6 +195,21 @@ return [
         'description' => 'Missing ID from OAuth2 provider.',
         'code' => 400,
     ],
+    Exception::USER_OAUTH2_BAD_REQUEST => [
+        'name' => Exception::USER_OAUTH2_BAD_REQUEST,
+        'description' => 'OAuth2 provider rejected the bad request.',
+        'code' => 400,
+    ],
+    Exception::USER_OAUTH2_UNAUTHORIZED => [
+        'name' => Exception::USER_OAUTH2_UNAUTHORIZED,
+        'description' => 'OAuth2 provider rejected the unauthorized request.',
+        'code' => 401,
+    ],
+    Exception::USER_OAUTH2_PROVIDER_ERROR => [
+        'name' => Exception::USER_OAUTH2_PROVIDER_ERROR,
+        'description' => 'OAuth2 provider returned some error.',
+        'code' => 424,
+    ],
 
     /** Teams */
     Exception::TEAM_NOT_FOUND => [
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -2,6 +2,7 @@
 
 use Ahc\Jwt\JWT;
 use Appwrite\Auth\Auth;
+use Appwrite\Auth\OAuth2\Exception as OAuth2Exception;
 use Appwrite\Auth\Validator\Password;
 use Appwrite\Auth\Validator\Phone;
 use Appwrite\Detector\Detector;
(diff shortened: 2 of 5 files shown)


app/config/errors.php
```diff
@@ -195,6 +195,21 @@
         'description' => 'Missing ID from OAuth2 provider.',
         'code' => 400,
     ],
+    Exception::USER_OAUTH2_BAD_REQUEST => [
+        'name' => Exception::USER_OAUTH2_BAD_REQUEST,
+        'description' => 'OAuth2 provider rejected the bad request.',
```
gewenyu99, line 200: This is great!

I would still elaborate on these a little:
**[What went wrong] [Why we think it went wrong] [How you could fix it]**

OAuth2 provider rejected the bad request. We think [SOMETHING] is misconfigured. Check [PLACE] to see if there are any mistakes.

gewenyu99 (changes requested):
Love where this is going.

I'd still like to see more context, even if we're being too wordy, or giving educated guesses. More information for the end user to use to debug is always valuable.

app/config/errors.php
stnguyen90: @gewenyu99, these are the default descriptions shown in the docs, but they can be different from what's returned to the developer. For example, one error response we have is:

```json
{
    "message": "Failed to obtain access token. The Google OAuth2 provider returned an error: invalid_grant: Malformed auth code.",
    "type": "user_oauth2_bad_request",
    "code": 400
}
```

gewenyu99: Okay, would I be so bold to suggest saying: OAuth2 provider rejected the bad request. This error will be replaced by an error from the provider at runtime.

I know it's excessive, but maybe this is a good hint for maintainers, but also in docs.

I err on the side of over explaining with errors cause it's extra frustrating. You can make the call.

gewenyu99 (approved):

app/config/errors.php
stnguyen90: > This error will be replaced by an error from the provider at runtime.

That doesn't really make sense to me because it's just a description of the error type.

app/controllers/api/account.php
```diff
@@ -447,27 +461,66 @@
...
+        try {
+            $accessToken = $oauth2->getAccessToken($code);
+            $refreshToken = $oauth2->getRefreshToken($code);
+            $accessTokenExpiry = $oauth2->getAccessTokenExpiry($code);
+        } catch (OAuth2Exception $ex) {
+            $failureRedirect(
+                $ex->getType(),
+                'Failed to obtain access token. The ' . $providerName . ' OAuth2 provider returned an error: ' . $ex->getMessage(),
+                $ex->getCode(),
+            );
```
stnguyen90, line 518: @gewenyu99 ,

Here, `$ex->getType()` can either be:

1. `USER_OAUTH2_BAD_REQUEST`
2. 'USER_OAUTH2_UNAUTHORIZED`
3. `USER_OAUTH2_PROVIDER_ERROR`

So, I can put some additional text based on that. I just need to know what you'd want to say.

So far, from my tests, `USER_OAUTH2_BAD_REQUEST` occurs if the code is old or incorrect.

`USER_OAUTH2_UNAUTHORIZED` can occur if the secret in appwrite is incorrect.

eldadfux:
@stnguyen90 also, something has happened to the style of the HTML error page. Let's see if we can replicated the standalone error style we use for the console.

app/config/errors.php
gewenyu99: > > This error will be replaced by an error from the provider at runtime.
> 
> That doesn't really make sense to me because it's just a description of the error type.

:/ Maybe the real issue is how we present these errors. I never know which ones will have their description replaced at runtime, which ones will actually use the description in the error file.

I'll let it slide, cause it's not really developer facing, more a maintainer facing issue.

gewenyu99 (approved):
I'm okay with the words used in this PR. Can't speak to the code :P

eldadfux (approved):

Issue #5680, opened 2023-06-09
🐛 Bug Report: 500 Error Accessing Disabled Collection


Lexazan:
### 👟 Reproduction steps

If collection is disabled via UI in Database->Some collection->Settings-> Enabled toggle
it can't be open using UI menus anymore, you will get an error 500
![image](https://github.com/appwrite/appwrite/assets/2877709/44698cb8-a7a9-4830-bd71-8cd1dc29b54d)
(AppwriteException: Collection with the requested ID could not be found.)
But you can still access the settings if you type in the collection settings url directly, E.g:
![image](https://github.com/appwrite/appwrite/assets/2877709/e96cc1ce-bf02-44b3-8a5e-acd9125afbff)

Seems like that's happening because Documents page of the collection is restricted when it is disabled.

### 👍 Expected behavior

It should be possible to open it, even when disabled.

### 👎 Actual Behavior

Throws error 500

### 🎲 Appwrite version

Version 1.3.x

### 💻 Operating system

Linux

### 🧱 Your Environment

Version 1.3.6, Docker

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

joeyouss:
Hi @Lexazan , did you try running it again? is it working fine now for you?

joeyouss:
Hi @Lexazan I think this should be working now, can you please confirm once?

Lexazan:
Hey, sorry guys, I am on vacation and can't check it now. Maybe someone else can?

stnguyen90:
I still see the error on 1.3.8

![image](https://github.com/appwrite/appwrite/assets/1477010/ddc8f24d-3509-43e4-9b02-4b5822ee17a1)

but not on cloud.

abnegate:
Closing as fixed in 1.4.0

PR #5900, opened 2023-07-28
Fix collection disabled mode checks

abnegate:
## What does this PR do?

Adds the missing admin mode checks when accessing collections

## Test Plan

Updated tests to check for when collection is disabled

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/5680

## Checklist


app/controllers/api/databases.php
```diff
@@ -3235,7 +3235,7 @@ function updateAttribute(
 
         $collection = Authorization::skip(fn() => $dbForProject->getDocument('database_' . $database->getInternalId(), $collectionId));
 
-        if ($collection->isEmpty() || !$collection->getAttribute('enabled')) {
+        if ($collection->isEmpty() || (!$collection->getAttribute('enabled') && $mode !== APP_MODE_ADMIN)) {
```
stnguyen90, line 3238: `Auth::isPrivilegedUser(Authorization::getRoles())` implies `$mode === APP_MODE_ADMIN` so I don't think we need both.

abnegate: I don't think it does, you can be a privileged user but not make the request in admin mode and vice versa

stnguyen90: `isPrivilegedUser()` returns true for Console users, right? I think a console user would only be detected if mode was admin:

https://github.com/appwrite/appwrite/blob/fcd4a69ad1b3db67c6d89f02c5f230c2038a4b35/app/init.php#L821-L824

app/controllers/api/databases.php
```diff
@@ -2664,14 +2664,20 @@ function updateAttribute(
...
+            if (!($isAdminMode && ($isAppUser || $isPrivilegedUser))) {
+                throw new Exception(Exception::DATABASE_NOT_FOUND);
+            }
         }
 
         $collection = Authorization::skip(fn() => $dbForProject->getDocument('database_' . $database->getInternalId(), $collectionId));
 
         if ($collection->isEmpty() || !$collection->getAttribute('enabled')) {
-            if (!($mode === APP_MODE_ADMIN && Auth::isPrivilegedUser(Authorization::getRoles()))) {
+            if (!($isAdminMode && ($isAppUser || $isPrivilegedUser))) {
```
stnguyen90, line 2680: I don't think admin mode is related to app user (API key) at all. If you look around here:

https://github.com/appwrite/appwrite/blob/fcd4a69ad1b3db67c6d89f02c5f230c2038a4b35/app/controllers/general.php#L317

You'll see there's never any check for $mode.

app/controllers/api/databases.php
```diff
@@ -2664,14 +2664,21 @@ function updateAttribute(
 
         $database = Authorization::skip(fn() => $dbForProject->getDocument('databases', $databaseId));
 
-        if ($database->isEmpty() || (!$database->getAttribute('enabled') && $mode !== APP_MODE_ADMIN)) {
-            throw new Exception(Exception::DATABASE_NOT_FOUND);
+        $isAPIKey = Auth::isAppUser(Authorization::getRoles());
+        $isPrivilegedUser = Auth::isPrivilegedUser(Authorization::getRoles());
+        $isAdminMode = $mode === APP_MODE_ADMIN;
+        $isConsole = $isAdminMode && $isPrivilegedUser;
```
stnguyen90, line 2670: I don't think we need these additional checks; `$isPrivelegedUser` is enough.

app/controllers/api/databases.php
```diff
@@ -2664,14 +2664,21 @@ function updateAttribute(
...
+
+        if ($database->isEmpty() || !$database->getAttribute('enabled')) {
+            if (!$isConsole && !$isAPIKey) {
+                throw new Exception(Exception::DATABASE_NOT_FOUND);
+            }
         }
```
stnguyen90, line 2676: I wonder if this logic is correct 🧐. If the database doesn't exist, but I'm accessing this with an API key or as a Console User, this won't throw an exception, and it will continue to try and fetch the collection, causing unwanted/unexpected behavior.

stnguyen90 (changes requested):
also looks like there are merge conflicts.

stnguyen90 (approved):
LGTM. The only thing is the tests are failing. I re-ran them so 🤞🏼

PR #5921, opened 2023-08-01
Separate OAuth2 info from Sessions into Identities

stnguyen90:
## What does this PR do?

This allows us to retain the OAuth2 info even if the session is deleted. This also provides a foundation for allowing multiple emails, phone numbers, etc, not from an OAuth2 provider.

## Test Plan

Manual

## Related PRs and Issues

- https://github.com/appwrite/appwrite/pull/5551

## Checklist


src/Appwrite/Utopia/Database/Validator/Queries/Identities.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Appwrite\Utopia\Database\Validator\Queries;
+
+class Identities extends Base
+{
+    public const ALLOWED_ATTRIBUTES = [
```
Meldiron, line 7: Might be worth adding index for token expiration

src/Appwrite/Utopia/Response/Model/Identity.php
```diff
@@ -0,0 +1,101 @@
...
+            ])
+            ->addRule('provider', [
+                'type' => self::TYPE_STRING,
+                'description' => 'Identity Provider.',
+                'default' => '',
+                'example' => 'email',
+            ])
+            ->addRule('status', [
+                'type' => self::TYPE_STRING,
+                'description' => 'Connection status. Can be connected or disconnected',
```
Meldiron, line 45: Why not delete the identity if disconnected?

src/Appwrite/Utopia/Response/Model/Identity.php
```diff
@@ -0,0 +1,101 @@
...
+            ])
+            ->addRule('providerUid', [
+                'type' => self::TYPE_STRING,
+                'description' => 'ID of the User in the Identity Provider.',
+                'default' => '',
+                'example' => '5e5bb8c16897e',
+            ])
+            ->addRule('providerEmail', [
+                'type' => self::TYPE_STRING,
+                'description' => 'Email of the User in the Identity Provider.',
```
Meldiron, line 57: A bit skeptic about having email here. OAuth1 might not provide email.

If we use email, what do we do about OAUth1 providers?
Id we dotn use email, do we really need to store it?

app/controllers/api/account.php
```diff
@@ -602,7 +644,7 @@
                         'email' => $email,
                         'emailVerification' => true,
                         'status' => true, // Email should already be authenticated by OAuth2 provider
-                        'password' => Auth::passwordHash(Auth::passwordGenerator(), Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS),
+                        'password' => null,
```
Meldiron, line 647: Can we ensure those accounts without password can do password-protected actions? Like change password, or delete project

src/Appwrite/Utopia/Response/Model/Identity.php
Meldiron: I couldn' find place where we set it to disconnected.

app/controllers/api/account.php
stnguyen90: Yes, things work better if password is null.

src/Appwrite/Utopia/Database/Validator/Queries/Identities.php
stnguyen90: I can add an index if you need to query for identities based on `providerAccessTokenExpiry`, but I'm hesitant to expose it on the API in case we decide to restructure the data. What do you think?

app/controllers/api/account.php
```diff
@@ -705,6 +793,86 @@
...
     });
 
+App::get('/v1/account/identities')
+    ->desc('List Identities')
+    ->groups(['api', 'account'])
+    ->label('scope', 'account')
+    ->label('usage.metric', 'users.{scope}.requests.read')
+    ->label('sdk.auth', [APP_AUTH_TYPE_SESSION, APP_AUTH_TYPE_JWT])
+    ->label('sdk.namespace', 'account')
+    ->label('sdk.method', 'listConnected')
```
Meldiron, line 803: `listIdentities` makes more sense. Have we considered `getIdentity` also?

christyjacob4: 👆 I agree, `listIdenties` abides more by our conventions

src/Appwrite/Utopia/Database/Validator/Queries/Identities.php
eldadfux: It make sense to me to have the expiry, in which case you think we might want to hide it?

stnguyen90: @eldadfux, I really want identities to support other things besides oauth so there might be cases where the identity doesn't have a providerAccessTokenExpiry. For example, an additional email or phone.

src/Appwrite/Utopia/Response/Model/Identity.php
stnguyen90: I've removed the status attribute.

app/controllers/api/account.php
stnguyen90: This was leftover from when we called it connected accounts. Thanks for catching it!

src/Appwrite/Utopia/Database/Validator/Queries/Identities.php
eldadfux: We could examine the approach we used for attributes where we have a base model that is extended. Attributes is also the only place where an array of models can contain models of different types. So, I think we might have the foundation for it.

stnguyen90: @Meldiron, @eldadfux, I've added providerAccessTokenExpiry as a filterable attribute.

eldadfux (approved):

app/controllers/api/account.php
```diff
@@ -650,10 +688,54 @@
...
+                Query::notEqual('userId', $user->getId()),
+            ]);
+            if (!empty($identitiesWithMatchingEmail)) {
+                throw new Exception(Exception::USER_EMAIL_ALREADY_EXISTS);
+            }
+
+            $dbForProject->createDocument('identities', new Document([
+                '$id' => ID::unique(),
+                '$permissions' => [
+                    Permission::read(Role::any()),
```
Meldiron, line 718: Is it okay to allow anyone to read user's identity? It could be exposing email

Meldiron (approved):

PR #5922, opened 2023-08-01
Style Appwrite Error Page

stnguyen90:
## What does this PR do?

If an endpoint throws an exception, an error like this is shown:

image

This PR adds some styling so that the page matches the Console

## Test Plan

Manual:

Prod mode:

image

Dev mode:

image

## Related PRs and Issues

None

## Checklist

### DIFF
--- a/app/controllers/general.php
+++ b/app/controllers/general.php
@@ -561,6 +561,7 @@ App::error()
                 ->setParam('projectName', $project->getAttribute('name'))
                 ->setParam('projectURL', $project->getAttribute('url'))
                 ->setParam('message', $error->getMessage())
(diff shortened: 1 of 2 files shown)


eldadfux (changes requested):
Let's use proper escape when printing stuff in the template.

Inside the template, we can use:
```php
$this->print($value, View::FILTER_ESCAPE);
```

https://github.com/utopia-php/framework/blob/master/src/View.php#L9C18-L9C31
https://github.com/utopia-php/framework/blob/master/src/View.php#L208

eldadfux (approved):

PR #5923, opened 2023-08-01
V19 migration - uid varchar

fogelito:
Updating Collections attribute UID from CHAR => VARCHAR


abnegate:
@fogelito Looks good, we need the [database side PR](https://github.com/utopia-php/database/pull/245) merged and released and the version updated here before merge

abnegate (approved):

PR #5924, opened 2023-08-01
fix: error in initialising the Exception constructor in api.php

christyjacob4:
## What does this PR do?
The exception constrictor was being initialised incorrectly in api.php

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/controllers/shared/api.php
+++ b/app/controllers/shared/api.php
@@ -27,7 +27,7 @@ $parseLabel = function (string $label, array $responsePayload, array $requestPar
         $parts = explode('.', $match);
 
         if (count($parts) !== 2) {
-            throw new Exception('Too less or too many parts', 400, Exception::GENERAL_ARGUMENT_INVALID);
+            throw new Exception(Exception::GENERAL_SERVER_ERROR, "The server encountered an error while parsing the label: $label");
         }
 
         $namespace = $parts[0] ?? '';


app/controllers/shared/api.php
```diff
@@ -27,7 +27,7 @@
         $parts = explode('.', $match);
 
         if (count($parts) !== 2) {
-            throw new Exception('Too less or too many parts', 400, Exception::GENERAL_ARGUMENT_INVALID);
+            throw new Exception(Exception::GENERAL_SERVER_ERROR, "The server encountered an error while parsing the label: $label");
```
gewenyu99, line 30: I know it's not always possible, but try to include the following information in error messages:

[What went wrong] [Why we think it happened] [What we suggest as a next step/ a suggested fix]

gewenyu99: I feel like this error is missing some context. If you're a developer seeing this issue, what should you do as a next step?

gewenyu99 (changes requested):

app/controllers/shared/api.php
christyjacob4:
```suggestion
            throw new Exception(Exception::GENERAL_SERVER_ERROR, "The server encountered an error while parsing the label: $label. Please create an issue on GitHub to allow us to investigate further https://github.com/appwrite/appwrite/issues/new/choose");
```

gewenyu99 (approved):
LGTM

PR #5928, opened 2023-08-01
Fixed bug where memberships remained after a team is deleted

### LINKED ISSUE #5517: 🐛 Bug Report: Users can see stale membership
### 👟 Reproduction steps

* Create a user.
* Create a new team.
* Add the user as team member.
* Delete the Team.
* For the user view Membership.

### 👍 Expected behavior

The membership invite for the deleted team should not be present and the membership should be empty.

### 👎 Actual Behavior

The membership invite shows stale invite which can't be deleted as well.

Screenshot for stale invite:
Screenshot 2023-05-09 at 11 18 14 AM

Screenshot when deleting the stale invite:
Screenshot 2023-05-09 at 11 21 33 AM

### 🎲 Appwrite version

Version 1.3.x

### 💻 Operating system

MacOS

### 🧱 Your Environment

Local Docker image.

safwanyp:
## What does this PR do?

This PR closes #5517. 
- Fixed a bug that caused team memberships to remain linked to a user even after a team is deleted.

## Test Plan

Did a couple of tests by manually creating teams and memberships randomly. Kept track of the memberships and made sure they persisted and/or deleted as needed.

## Related PRs and Issues

- #5517

## Checklist


app/controllers/api/teams.php
```diff
@@ -343,11 +343,17 @@
...
             if (!$dbForProject->deleteDocument('memberships', $membership->getId())) {
                 throw new Exception(Exception::GENERAL_SERVER_ERROR, 'Failed to remove membership for team from DB');
             }
+
+            $user = $dbForProject->getDocument('users', $membership->getAttribute('userId'));
+            $user->setAttribute('memberships', array_values(array_filter(
+                $user->getAttribute('memberships', []),
+                fn($um) => $um['teamId'] !== $membership->getAttribute('teamId')
+            )));
+            $dbForProject->updateDocument('users', $user->getId(), $user);
```
stnguyen90, line 356: The big drawback about doing it in here is that the API can then take much longer to finish since it must iterate over every user, but the API should be as quick as possible.

stnguyen90 (changes requested):
Left a comment about your approach.

Could you do some investigation to actually see if there is any data in the database that needs to be deleted for each user? Or is it maybe somewhere else?

app/controllers/api/teams.php
safwanyp: I'm gonna spend some time to see how and where the membership data is stored exactly. Will also look into background workers. Will leave a comment soon!

safwanyp: Okay so based on what I see in mariadb, there is a `memberships` column in the `_1_users` table, but nothing gets stored in it even after I create a membership. As far as I can understand, when a membership is created, it does various checks and creates an entry in the `_1_memberships` table and that's it. There's no more entries in any other db as far as I can tell from the code and the db.

Going through the code in the console, it seems like the memberships page simply calls the `listMemberships` endpoint with user params. So the issue must be something to do with the frontend because the memberships get deleted from the `_1_memberships` table as expected, and memberships are displayed from a Teams API endpoint and not an Account API. I'll be reverting the code back to it's original state, and fixing the frontend issue. 

Does that sound good?

safwanyp: Okay after some more digging, looks like I missed something super obvious. The sdk calls `users.listMemberships` which uses the `teams` table. Gonna look more into this and see what's up. Sorry for the notifs, and let me know if you want me to stop lmao

safwanyp:
Hey @stnguyen90 just pushed a commit to this PR. 
I solved it by checking if the `$team` variable is empty. After going through a lot of code in the Utopia package, as well as keeping track of changes in the mariadb tables, seems like this is a decent fix. Let me know if this is better.

Since there aren't any more API calls being made than there were originally, I think it's good. Regardless, I'm willing to learn and if you think I could have found a better solution, let me know!!

app/controllers/api/users.php
```diff
@@ -530,6 +530,10 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
         $memberships = array_map(function ($membership) use ($dbForProject, $user) {
             $team = $dbForProject->getDocument('teams', $membership->getAttribute('teamId'));
 
+            if ($team->isEmpty()) {
+                throw new Exception(Exception::TEAM_NOT_FOUND);
+            }
+
```
stnguyen90, line 536: Hmm...this would cause an error so when you browse to the page, nothing would show even if they had other teams they were in, right?

stnguyen90 (changes requested):

safwanyp:
Just pushed a fix that definitely works 😶‍🌫️ 

HUGE thanks to @stnguyen90 for the absolutely amazing support he provided till now. I have learnt so much more about the system design of Appwrite, and was mesmerized to say the least.

That being said, the fix was actually really simple, and I just needed to implement cache invalidation when a team is deleted. According to network timings in the dev tools, the cache invalidation has added ~0.4ms per user to the response time. With 7 users, it added 2.7ms, and with 2 users, it took less than 1ms.

Feedback is appreciated!

stnguyen90 (commented):
Looks good! Can you add something to the tests to ensure the user's memberships are updated?

safwanyp:
> Looks good! Can you add something to the tests to ensure the user's memberships are updated?

Will do that now

safwanyp:
Added a test to check the same and it was a success on localhost

tests/e2e/Services/Teams/TeamsBase.php
```diff
@@ -428,4 +428,67 @@ public function testUpdateAndGetUserPrefs(array $data): void
...
+
+            $this->assertEquals(201, $mem['headers']['status-code']);
+            $this->assertNotEmpty($mem['body']['$id']);
+            $this->assertNotEmpty($mem['body']['userId']);
+            $this->assertEquals('User ' . $i, $mem['body']['userName']);
+            $this->assertEquals('email' . $i . '@example.com', $mem['body']['userEmail']);
+            $this->assertNotEmpty($mem['body']['teamId']);
+            $this->assertCount(2, $mem['body']['roles']);
+        }
+
```
stnguyen90, line 470: I would add a check here to confirm the user's memberships are as expected.

tests/e2e/Services/Teams/TeamsBase.php
```diff
@@ -428,4 +428,67 @@ public function testUpdateAndGetUserPrefs(array $data): void
...
+        $this->client->call(Client::METHOD_DELETE, '/teams/' . $team['body']['$id'], array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()));
+
+        foreach ($users as $user) {
```
stnguyen90, line 476: It doesn't look like you populated $users

tests/e2e/Services/Teams/TeamsBase.php
```diff
@@ -428,4 +428,67 @@ public function testUpdateAndGetUserPrefs(array $data): void
...
+            $user = $this->client->call(Client::METHOD_GET, '/users/' . $user['body']['$id'], array_merge([
+                'content-type' => 'application/json',
+                'x-appwrite-project' => $this->getProject()['$id'],
+            ], $this->getHeaders()));
+
+            $this->assertEquals(200, $user['headers']['status-code']);
+            $this->assertEquals(0, $user['body']['total']);
+            $this->assertEquals([], $user['body']['memberships']);
```
stnguyen90, line 484: These don't get returned by the Get User API.

app/controllers/api/teams.php
```diff
@@ -343,11 +343,12 @@
             Query::limit(2000), // TODO fix members limit
         ]);
 
-        // TODO delete all members individually from the user object
         foreach ($memberships as $membership) {
+            // Memberships are deleted here instead of in the worker to make sure user permisions are updated instantly
```
stnguyen90, line 347: Would you please move the comment to before the for loop because the comment applies to the whole block?

stnguyen90 (changes requested):
2 additional things:

1. when you're done making changes, make sure to re-request a review from your reviewer in GitHub.
2. some suggestions and resources on git commit messages:
   1. [How to Write a Git Commit Message](https://cbea.ms/git-commit/)
   2. [My favourite Git commit | dhwthompson.com](https://dhwthompson.com/2019/my-favourite-git-commit)

safwanyp:
Re-requested review.

Thanks for the helpful resources Steven, I honestly used to try and write descriptive commit messages, but that habit died down slowly lol. Will definitely start writing more descriptive commits from now on!

tests/e2e/Services/Teams/TeamsBaseServer.php
```diff
@@ -281,4 +282,101 @@ public function testDeleteUserUpdatesTeamMembershipCount($data)
...
+            ], false);
+
+            $user_body = json_decode($user['body'], true);
+
+            $this->assertEquals(201, $user['headers']['status-code']);
+            $this->assertEquals('newuser' . $i . '@localhost.test', $user_body['email']);
+            $this->assertEquals('New User ' . $i, $user_body['name']);
+            $this->assertEquals($user_body['status'], true);
+
+            array_push($new_users, $user_body['$id']);
```
stnguyen90, line 311: PHP's shorthand notation for this is:

```suggestion
            $new_users[] = $user_body['$id'];
```

tests/e2e/Services/Teams/TeamsBaseServer.php
```diff
@@ -281,4 +282,101 @@ public function testDeleteUserUpdatesTeamMembershipCount($data)
...
+
+    public function testTeamDeleteUpdatesUserMembership()
+    {
+        $new_users = [];
+
+        /**
+         * Create 5 new users and add their IDs to an array
+         */
+        for ($i = 0; $i < 5; $i++) {
+            $user = $this->client->call(Client::METHOD_POST, '/users', array_merge([
```
stnguyen90, line 294: Can we just create the team membership without creating the user first?

stnguyen90 (commented):
> Will definitely start writing more descriptive commits from now on!

Sounds good! 👍🏼 Also, don't forget to use imperative mood instead of past tense!

tests/e2e/Services/Teams/TeamsBaseServer.php
safwanyp: Yeah using the create membership endpoint to create users is fine as well. I just thought it would be better to create the user separately. If it's somewhat of a deal-breaker in terms of test performance, I'll go ahead and do that instead. Let me know!

tests/e2e/Services/Teams/TeamsBaseServer.php
safwanyp: Did not know that, thanks a lot! I'll make this change 💡

tests/e2e/Services/Teams/TeamsBaseServer.php
```diff
@@ -281,4 +282,101 @@ public function testDeleteUserUpdatesTeamMembershipCount($data)
...
+                'content-type' => 'application/json',
+                'x-appwrite-project' => $this->getProject()['$id'],
+            ], $this->getHeaders()), [
+                'userId' => ID::unique(),
+                'email' => 'newuser' . $i . '@localhost.test',
+                'password' => 'password',
+                'name' => 'New User ' . $i,
+            ], false);
+
+            $user_body = json_decode($user['body'], true);
```
stnguyen90, line 304: We typically don't call this 🧐 is it really needed?

tests/e2e/Services/Teams/TeamsBaseServer.php
stnguyen90: Ya, our tests take a long time as it is so minimizing as much as possible would be good.

tests/e2e/Services/Teams/TeamsBaseServer.php
safwanyp: You mean the `json_decode` function? I saw a similar line in the create user test, so I decided to include it. If you're talking about the name, well, yeah it isn't really necessary. 

I also just saw that the `call` function takes a `decode` param, so I'm assuming that even if I want to decode the response, I can just do it by passing `true` instead of `false`?

stnguyen90: Ah yes, it looks like the decode param is optional and defaults to true so let's remove the false from the function call so it decodes automatically.

safwanyp: I'm getting rid of that whole block and using the create membership endpoint for user creation anyways. So all good I guess!

stnguyen90 (changes requested):
looks like linter failed.

safwanyp:
> looks like linter failed.

Honestly forgot to run the formatter and linter. Fixed 1/1 issue and pushed.

stnguyen90 (approved):
Great work!

safwanyp:
> Great work!

Huge thanks to you Steven! Really appreciate all the help!

eldadfux (approved):

eldadfux:
Thank you @safwanyp!

PR #5930, opened 2023-08-02
add collections to config file and messaging scopes to config file

fanatic75:
## What does this PR do?

This PR adds collections required for messaging service in config. 

## Related PRs and Issues

- (Related PR or issue)
For more details, check out [RFC](https://github.com/appwrite/rfc-internal/blob/main/003-messaging/003-messaging.md#new-collections) 

## Checklist


fanatic75:
@abnegate  This is base out of 1.4.x branch. You can just check my ending commits for reviewing the PR.

app/config/collections.php
```diff
@@ -1366,7 +1377,495 @@
...
+                'lengths' => [128],
+                'orders' => [Database::ORDER_ASC],
+            ],
+            [
+                '$id' => ID::custom('_key_name'),
+                'type' => Database::INDEX_FULLTEXT,
+                'attributes' => ['name'],
+                'lengths' => [128],
+                'orders' => [Database::ORDER_ASC],
+            ]
```
abnegate, line 1456: Do we internally query on any other attributes in this collection? I'm thinking maybe `type`?

app/config/collections.php
```diff
@@ -1366,7 +1377,495 @@
...
+                'format' => '',
+                'size' => 0,
+                'signed' => false,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => ['datetime'],
+            ],
+            [
+                '$id' => ID::custom('deliveryError'),
```
abnegate, line 1521: As discussed lets update this to a string array `deliveryErrors`

app/config/collections.php
```diff
@@ -1366,7 +1377,495 @@
...
+                'required' => true,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('data'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 16384,
```
abnegate, line 1491: Let's update this to `65535` to allow storing as much data as possible per message

app/config/collections.php
```diff
@@ -1366,7 +1377,495 @@
...
+                'required' => true,
+                'default' => null,
+                'array' => false,
+                'filters' => ['json'],
+            ],
+            [
+                '$id' => ID::custom('to'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 16834,
```
abnegate, line 1502: `16384` will limit each message to ~400 recipients, let's set this to the max of `65535` as well

app/config/collections.php
```diff
@@ -1366,7 +1377,495 @@
...
+                '$id' => ID::custom('credentials'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 16384,
+                'signed' => true,
+                'required' => true,
+                'default' => null,
+                'array' => false,
+                'filters' => ['json', 'encrypt'],
+            ]
```
abnegate, line 1440: Let's add the `search` attribute here too, as well a fulltext index on it

app/config/collections.php
```diff
@@ -1366,7 +1377,495 @@
...
+                'lengths' => [128],
+                'orders' => [Database::ORDER_ASC],
+            ],
+            [
+                '$id' => ID::custom('_key_search'),
+                'type' => Database::INDEX_FULLTEXT,
+                'attributes' => ['search'],
+                'lengths' => [],
+                'orders' => [],
+            ],
```
abnegate, line 1586: Do we internally query on any other attributes in this collection? I'm thinking maybe `deliveryTime` and `delivered`?

app/config/collections.php
```diff
@@ -1366,7 +1377,495 @@
...
+            [
+                '$id' => ID::custom('_key_description'),
+                'type' => Database::INDEX_FULLTEXT,
+                'attributes' => ['description'],
+                'lengths' => [],
+                'orders' => [],
+            ],
```
abnegate, line 1679: In hindsight not sure if we'll need this index since we likely won't query the description

app/config/collections.php
```diff
@@ -1366,7 +1377,495 @@
...
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 16384,
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => ['subQueryTopicTargets'],
+            ]
+        ],
```
abnegate, line 1650: Let's add the `search` attribute here too, as well a fulltext index on it

app/config/collections.php
```diff
@@ -1366,7 +1377,495 @@
...
+                '$id' => ID::custom('providerType'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => Database::LENGTH_KEY,
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => ['subQueryProviderType'],
+            ],
```
abnegate, line 1824: Do we actually need this 🤔 might be just as easy to get the provider and check it's `type` after

app/config/collections.php
```diff
@@ -1366,7 +1377,495 @@
...
+            [
+                '$id' => ID::custom('identifier'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => Database::LENGTH_KEY,
```
abnegate, line 1829: Let's change this to 2048, push tokens may be longer than 255 chars, same for some people's emails

app/init.php
```diff
@@ -530,6 +530,19 @@ function (mixed $value) {
...
+        return null;
+    },
+    function (mixed $value, Document $document, Database $database) {
+        return Authorization::skip(fn() => $database
+            ->find('targets', [
+                Query::equal('userInternalId', [$document->getInternalId()]),
+                Query::limit(APP_LIMIT_SUBQUERY),
+            ]));
+    }
+);
```
abnegate, line 545: I think we're missing `subQueryTopicTargets` here

abnegate (commented):
Let's revert the console changes and add the scopes in `roles.php` as well. All except subscribers and target should be admins only, and subscribers and targets should be both admins and members

app/config/collections.php
```diff
@@ -1366,7 +1377,495 @@
...
+                'array' => false,
+                'filters' => ['subQueryTopicTargets'],
+            ]
+        ],
+        'indexes' => [
+            [
+                '$id' => ID::custom('_key_providerId'),
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['providerId'],
+                'lengths' => [128],
```
fogelito, line 1656: We can leave the index length empty since attribute _key_providerId is 128 in length. 
We do not exceed the length max.

app/config/collections.php
```diff
@@ -1366,7 +1377,495 @@
...
+                'orders' => [Database::ORDER_ASC],
+            ],
+            [
+                '$id' => ID::custom('_key_providerInternalId'),
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['providerInternalId'],
+                'lengths' => [128],
```
fogelito, line 1663: Same

app/config/collections.php
fogelito: If it is possible we prefer types as constant integers.

app/config/collections.php
```diff
@@ -1366,7 +1377,495 @@
...
+                'array' => false,
+                'filters' => [],
+            ],
+        ],
+        'indexes' => [
+            [
+                '$id' => ID::custom('_key_userId'),
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['userId'],
+                'lengths' => [128],
```
fogelito, line 1842: Same here with index length

app/config/collections.php
```diff
@@ -1366,7 +1377,495 @@
...
+                'orders' => [Database::ORDER_ASC],
+            ],
+            [
+                '$id' => ID::custom('_key_userInternalId'),
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['userInternalId'],
+                'lengths' => [128],
```
fogelito, line 1849: Same here with index length

app/config/collections.php
fanatic75: updated it to deliveryErrors

app/config/collections.php
fanatic75: done this

app/config/collections.php
fanatic75: done this

app/config/collections.php
fanatic75: hmm, I haven't written messages endpoint yet for console but I am not sure if we are gonna query in list messages by deliveryTime or delivered

app/config/collections.php
fanatic75: removed this

app/config/collections.php
fanatic75: done this

app/config/collections.php
fanatic75: hmm, we can change this to provider as well. although we are only selecting type from provider sub query so that would still be faster than fetching the whole provider doc but your call @abnegate

app/config/collections.php
fanatic75: done this

app/config/collections.php
fanatic75: done this

app/config/collections.php
fanatic75: done this

app/config/collections.php
fanatic75: done this

app/config/collections.php
fanatic75: done this

app/init.php
```diff
@@ -530,6 +530,60 @@ function (mixed $value) {
...
+                $document->getAttribute('providerId'),
+                [Query::select(['type'])]
+            ));
+        if ($provider) {
+            return $provider->getAttribute('type');
+        }
+        return null;
+    }
+);
+
```
abnegate, line 565:
```suggestion
```

app/config/collections.php
abnegate: Let's remove it, and call `getDocument` using the `providerId` from here where we need it. Then if we need any other provider attributes we can use them, without making a second query to get the same document but with more attributes

app/init.php
fanatic75: done this

app/config/collections.php
```diff
@@ -1366,7 +1377,545 @@
...
+                '$id' => ID::custom('provider'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => Database::LENGTH_KEY,
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => ['subQueryProvider'],
+            ],
```
abnegate, line 1874: Let's remove this (as well as the sub-query), we can manually get the provider by ID when needed

fanatic75: done this

abnegate (approved):

app/init.php
```diff
@@ -530,6 +530,40 @@ function (mixed $value) {
...
+Database::addFilter(
+    'subQueryTargets',
+    function (mixed $value) {
+        return null;
+    },
+    function (mixed $value, Document $document, Database $database) {
+        return Authorization::skip(fn() => $database
+            ->find('targets', [
+                Query::equal('userInternalId', [$document->getInternalId()]),
+                Query::limit(APP_LIMIT_SUBQUERY),
```
fogelito, line 542: If we added limit? do we want to order by desc to get the latest?

PR #5934, opened 2023-08-03
Feat appwrite assistant

lohanidamodar:
## What does this PR do?

- Integrate assistant endpoint and container

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/config/services.php
```diff
@@ -199,4 +199,17 @@
         'optional' => false,
         'icon' => '',
     ],
+    'assistant' => [
+        'key' => 'assistant',
+        'name' => 'Assistant',
+        'subtitle' => 'The Appwrite assistant service allows you to interact with appwrite assistant.',
```
eldadfux, line 205:
```suggestion
        'subtitle' => 'The Appwrite assistant service allows you to interact with Appwrite'\s AI assistant.',
```

lohanidamodar: Removing this for now, adding it to console service

app/controllers/api/console.php
```diff
@@ -38,3 +39,57 @@
...
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_TEXT)
+    ->param('query', '', new Text(2000), 'Query')
+    ->inject('response')
+    ->action(function (string $query, Response $response) {
+        $ch = curl_init('http://appwrite-assistant:3003/');
+        $responseHeaders = [];
+        $query = json_encode(['prompt' => $query]);
+
+        $headers = ['accept: text/event-stream', 'authorization: ' . APP::getEnv('ASSISTANT_SECRET')];
```
eldadfux, line 60: Why do we need this secret? This container is going to be accessible only from private network

eldadfux: I think we can skip it, for the executor it was to protect different runtimes from reaching each other.

eldadfux:
One more thing, we need to add abuse protection

app/controllers/api/console.php
```diff
@@ -38,3 +39,59 @@
...
+    ->desc('Ask Query')
+    ->groups(['api', 'assistant'])
+    ->label('scope', 'public')
+    ->label('sdk.auth', [APP_AUTH_TYPE_SESSION])
+    ->label('sdk.namespace', 'assistant')
+    ->label('sdk.method', 'chat')
+    ->label('sdk.description', '/docs/references/assistant/chat.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_TEXT)
+    ->label('abuse-limit', 15)
```
eldadfux, line 54: Is the default time interval 1 hour?

eldadfux (approved):

app/controllers/api/console.php
lohanidamodar: yes https://github.com/appwrite/appwrite/blob/fcd4a69ad1b3db67c6d89f02c5f230c2038a4b35/app/controllers/shared/api.php#L121

eldadfux (approved):

PR #5938, opened 2023-08-04
Implement Migrations

PineappleIOnic:
## What does this PR do?

This PR Implements the API and Service for Migrations to work

## Test Plan

Most testing will be handled by the Transfer project instead due to the reliance on third party/mock API's to test it.

## Checklist


src/Appwrite/Utopia/Response.php
```diff
@@ -198,6 +201,13 @@ class Response extends SwooleResponse
     public const MODEL_BUILD_LIST = 'buildList';  // Not used anywhere yet
     public const MODEL_FUNC_PERMISSIONS = 'funcPermissions';
 
+    // Migrations
+    public const MODEL_MIGRATION = 'migration';
+    public const MODEL_MIGRATION_LIST = 'migrationList';
+    public const MODEL_MIGRATION_REPORT = 'migrationReport';
+    public const MODEL_MIGRATION_FIREBASE_PROJECT = 'firebaseProject';
+    public const MODEL_MIGRATION_FIREBASE_PROJECT_LIST = 'firebaseProjectList';
```
christyjacob4, line 209: aren't we adding the other vendors here?

app/config/collections.php
```diff
@@ -2507,6 +2551,143 @@
...
+            [
+                '$id' => ID::custom('credentials'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 65536,
+                'signed' => true,
+                'required' => false,
+                'default' => [],
+                'array' => false,
+                'filters' => ['json'],
```
christyjacob4, line 2602: this field should probably have the encrypt filter

app/views/general/error.phtml
```diff
@@ -1,6 +1,5 @@
 <?php
 $development = $this->getParam('development', false);
-$type = $this->getParam('type', 'general_server_error');
```
christyjacob4, line 3: this shouldn't be removed

composer.json
```diff
@@ -66,6 +66,7 @@
         "utopia-php/storage": "0.14.*",
         "utopia-php/swoole": "0.8.*",
         "utopia-php/websocket": "0.1.*",
+        "utopia-php/transfer": "dev-feat-improve-features",
```
christyjacob4, line 69: lets remember to update the library version

christyjacob4 (approved):
@PineappleIOnic PR looks good! Lets make sure we update the library and we can merge this!!

src/Appwrite/Auth/OAuth2/Exception.php
```diff
@@ -16,9 +16,15 @@ public function __construct(string $response = '', int $code = 0, \Throwable $pr
...
         $decoded = json_decode($response, true);
         if (\is_array($decoded)) {
-            $this->error = $decoded['error'];
-            $this->errorDescription = $decoded['error_description'];
-            $this->message =  $this->error . ': ' . $this->errorDescription;
+            if (\is_array($decoded['error'])) {
+                $this->error = $decoded['error']['status'];
+                $this->errorDescription = $decoded['error']['message'];
+                $this->message =  $this->error . ': ' . $this->errorDescription;
+            } else {
```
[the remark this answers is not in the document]
PineappleIOnic: When the OAuth provider gives a JSON Response, this just makes it so it can handle JSON OAuth errors correctly

app/workers/migrations.php
```diff
@@ -0,0 +1,315 @@
...
+            $transfer = new Transfer(
+                $source,
+                $destination
+            );
+
+            $migrationDocument->setAttribute('stage', 'source-check');
+            $this->updateMigrationDocument($migrationDocument, $projectDocument);
+
+            $migrationDocument->setAttribute('stage', 'destination-check');
+            $this->updateMigrationDocument($migrationDocument, $projectDocument);
```
christyjacob4, line 240: you're updating the status here to `destination-check` but I dont see any code that does it 🤔

app/controllers/api/migrations.php
```diff
@@ -0,0 +1,973 @@
...
+            'resources' => $resources,
+            'statusCounters' => '{}',
+            'resourceData' => '{}',
+            'errors' => []
+        ]));
+
+        $events->setParam('migrationId', $migration->getId());
+
+        // Trigger Transfer
+        var_dump($project);
```
christyjacob4, line 227:
```suggestion
```

christyjacob4, line 227:
```suggestion
```

PR #5725, opened 2023-06-19
Feat: Functions G4

Meldiron:
## What does this PR do?

(Provide a description of what this PR does and why it's needed.)

## Test Plan

(Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified your changes work. Screenshots may also be helpful.)

## Related PRs and Issues

- (Related PR or issue)

## Checklist

- [ ] Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [ ] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

PR #5953, opened 2023-08-07
Separate OAuth2 info from Sessions into Identities

stnguyen90:
## What does this PR do?

This allows us to retain the OAuth2 info even if the session is deleted. This also provides a foundation for allowing multiple emails, phone numbers, etc, not from an OAuth2 provider.

This PR was created to target cl-1.4.x.

In addition, this PR adds a secrets attribute to the identities collection. These secrets can be used to store data from the provider that may or may not be sensitive.

For example, this will be used by the migration API when connecting to Firebase to store the service account used for the migration.

This data will only be used internally inside Appwrite and not exposed to an end user or developer.

Data here is tied to the user identity so that it can be removed when the user or the user identity is removed.

## Test Plan

Manual

## Related PRs and Issues

- https://github.com/appwrite/appwrite/pull/5551
- https://github.com/appwrite/appwrite/pull/5920

## Checklist


src/Appwrite/Utopia/Database/Validator/Queries/Identities.php
```diff
@@ -0,0 +1,23 @@
...
+namespace Appwrite\Utopia\Database\Validator\Queries;
+
+class Identities extends Base
+{
+    public const ALLOWED_ATTRIBUTES = [
+        'userId',
+        'provider',
+        'providerUid',
+        'providerEmail',
+        'providerAccessTokenExpiry',
```
christyjacob4, line 12: in what scenario do we want to allow querying by the access token expiry ?

docs/references/account/list-identities.md
```diff
@@ -0,0 +1 @@
+Get currently logged in user list of identities.
```
christyjacob4, line 1:
```suggestion
Get the list of identities for the currently logged in user.
```

src/Appwrite/Utopia/Database/Validator/Queries/Identities.php
stnguyen90: Let's say we or a developer wants to get all the expired or soon-to-be expired tokens to delete the identity or try and refresh it.

christyjacob4 (approved):

src/Appwrite/Utopia/Database/Validator/Queries/Identities.php
christyjacob4: okay alright 👍

app/controllers/api/account.php
```diff
@@ -690,8 +729,7 @@
                         'email' => $email,
                         'emailVerification' => true,
                         'status' => true, // Email should already be authenticated by OAuth2 provider
-                        'passwordHistory' => $passwordHistory > 0 ? [$password] : null,
-                        'password' => $password,
```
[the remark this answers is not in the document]
stnguyen90: Setting a password can cause problems with other APIs that expect the password to be null. In addition, it doesn't match the implementation for the other APIs that create a user without a password (Create Magic URL Session, Create Phone Session, Create Anonymous Session, etc).

docs/references/account/list-identities.md
stnguyen90: I matched the convention of some of the other descriptions like list-sessions.md so I'll update that too like:

```diff
-Get currently logged in user list of active sessions across different devices.
+Get the list of active sessions across different devices for the currently logged in user.
```

and how about these two?

account/get-prefs.md:

```diff
-Get currently logged in user preferences as a key-value object.
+Get the preferences as a key-value object for the currently logged in user
```

account/get.md:

```diff
-Get currently logged in user data as JSON object.
+Get currently logged in user.
```

christyjacob4: Yeap lets do it 👍

christyjacob4 (approved):
Lets also implement your recommendations for the descriptions

Meldiron:
This has been merged into `feat-git-inregration`.
We can close this PR if we merge that one

docs/references/account/list-identities.md
stnguyen90: updated!

stnguyen90:
> This has been merged into `feat-git-inregration`. We can close this PR if we merge that one

@Meldiron, but when will https://github.com/appwrite/appwrite/pull/5725 be merged considering this is needed for https://github.com/appwrite/appwrite/pull/5938?

PR #5971, opened 2023-08-11
Chore cleanup dockerfile

christyjacob4:
## What does this PR do?

A lot of the environment variables defined in the Dockerfile were redundant since

1. They had no significance during build time
2. All environment variables were defined in the .env file anyways
3. The make it harder for maintenance 

In this PR we remove the redundant environment variables from the Dockerfile

Doctor has been updated to use the new endpoint
Screenshot 2023-09-01 at 5 10 13 PM

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


loks0n (approved):

stnguyen90 (approved):
LGTM except the `composer.lock` doesn't need to be part of the PR, right?

Dockerfile
```diff
@@ -37,99 +37,7 @@ ARG VERSION=dev
...
-ENV DOCKER_COMPOSE_VERSION=v2.5.0
-
-ENV _APP_SERVER=swoole \
-    _APP_ENV=production \
-    _APP_LOCALE=en \
-    _APP_WORKER_PER_CORE= \
-    _APP_DOMAIN=localhost \
-    _APP_DOMAIN_FUNCTIONS=functions.localhost \
-    _APP_DOMAIN_TARGET=localhost \
-    _APP_HOME=https://appwrite.io \
```
abnegate, line 50: This value isn't set anywhere else (variables.php, docker-compose.yml, compose.phtml) and will cause the `doctor` to fail on any remote host:

https://github.com/appwrite/appwrite/blob/891099c7c3dc1fe86c4e64467f9249651991b603/src/Appwrite/Platform/Tasks/Doctor.php#L248

 Let's double check all the vars with a value are set elsewhere

abnegate (approved):

docker-compose.yml
```diff
@@ -95,6 +95,7 @@ services:
       - -dopcache.preload=opcache.preload=/usr/src/code/app/preload.php
     environment:
       - _APP_ENV
+      - _APP_HOME
```
abnegate, line 98: We'll need this in `compose.phtml` as well

abnegate, line 98: We'll also either need to hardcode the value here, or add it in `variables.php` with a default value

christyjacob4: @abnegate I had second thoughts about it since we also use it in the `specs.php` file where its used to generate some very specific URLs like our terms and conditions etc. 
https://github.com/appwrite/appwrite/blob/master/src/Appwrite/Platform/Tasks/Specs.php#L244

So it made sense it add a default inside the dockerfile and not override it.

abnegate (approved):

abnegate (approved):

Meldiron (approved):

PR #5951, opened 2023-08-07
Feat add messaging response models

fanatic75:
## What does this PR do?

Adds response models for messaging service

## Test Plan

(Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified your changes work. Screenshots may also be helpful.)

## Related PRs and Issues

This PR is dependent on https://github.com/appwrite/appwrite/pull/5937

## Checklist

- [ ] Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [ ] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

PR #5986, opened 2023-08-14
Feat messages event config

fanatic75:
## What does this PR do?

This PR adds event config for messages and adds target attribute in User's response Model

## Test Plan

## Related PRs and Issues

This PR is dependent on https://github.com/appwrite/appwrite/pull/5951

## Checklist


app/config/events.php
```diff
@@ -237,6 +251,51 @@
...
+            'create' => [
+                '$description' => 'This event triggers when a provider is created.',
+            ],
+            'update' => [
+                '$description' => 'This event triggers when a provider is updated.',
+            ],
+            'delete' => [
+                '$description' => 'This event triggers when a provider is deleted.'
+            ],
+        ],
```
abnegate, line 274: Let's move providers to the top level instead of nesting (contrary to the RFC), they're generic enough that we could potentially use them for other provider types like auth in future

abnegate (approved):

Issue #4952, opened 2023-01-05
🐛 Bug Report: test email address cause SMTP disconnected 


gepd:
### 👟 Reproduction steps

1. Add your SMTP credentials in your `.env` file
2. Run `docker exec appwrite doctor`
3. You can get `SMTP.............disconnected 👎`

### 👍 Expected behavior

You should get `SMTP................connected 👍`

### 👎 Actual Behavior

Currently the doctor try to send an email to `demo@example.com` 

https://github.com/appwrite/appwrite/blob/15107f0f5cf40b647e8d282e2c84322ba0bfd370/app/tasks/doctor.php#L139

debuggin the error returned by the code I have found this message:

```
PHPMailer\PHPMailer\Exception: SMTP Error: The following recipients failed: demo@example.com: The mail server could not deliver mail to demo@example.com.  The account or
domain may not exist, they may be blacklisted, or missing the proper dns
entries.
``` 

So my SMTP server haven't any problem but as the server can't find the address `demo@example.com` it fails

In my instance I have added the env var `_APP_SMTP_TEST_EMAIL` to change this address

Would you accept a PR with this change? 

### 🎲 Appwrite version

Version 1.2.x

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

stnguyen90:
@gepd thanks for raising this! let me check with the team on this.

LyVanBong:
I am also encountering a problem with the SMTP configuration of the Outlook Office 365 email. Has anyone encountered this issue before? Can anyone help me with this, I would greatly appreciate it. Thank you.

stnguyen90:
Actually, instead of an environment variable, I think it would be good to expose the email as a parameter to the CLI command.

mustansirgodhrawala:
Hey there, can this issue be assigned to me? I would like to go a step forward and include a small feature with your permission to create a quick smtp test button with CSRF in the UI for quick smtp testing for easier onboarding as well.

I'm quite intrigued by Appwrite and would like an opportunity to delve into the codebase and contribute beyond bugs.

mustansirgodhrawala:
Hey there @stnguyen90, can I handle this issue?

Haimantika:
[...]

PR #6014, opened 2023-08-18
Fix 4952 read doctor email receiver from console

mustansirgodhrawala:
## What is the issue? Where did it come from?
This issue arised because RFC 2606, does not allow sending emails to restricted domains such as example.com. This is why, even with a valid SMTP configuration, PHP Mailer will mark SMTP as disconnected.
A different fix, such as changing the default receiver to perhaps smtp-test[@]appwrite.io could also fix this issue. But perhaps this is a better fix? 

## What does this PR do?

This PR addresses the issue #4952, as @stnguyen90 mentions exposing the receiver email to the CLI would be the best approach I've implemented the same, with also added a interactive param. 

The interactive param is defaulted to 'n', which allows our e2e tests to stay the same. But requires updating our documentation to recommend the user to run doctor in the following way-
```bash
docker exec appwrite sh -c 'doctor --interactive=y'
```

## Test Plan

I've not made much changes, that would require adding extra tests. Since this is my first pull request, I'm not fully aware of what to fill in here.

## Related PRs and Issues

- Issue #4952 

## Checklist

### DIFF
--- a/app/tasks/doctor.php
+++ b/app/tasks/doctor.php
@@ -10,12 +10,14 @@ use Utopia\App;
 use Utopia\CLI\Console;
 use Utopia\Domains\Domain;
 use Utopia\Validator\Text;
+use Utopia\Validator\Wildcard;
 
 $cli
     ->task('doctor')
     ->desc('Validate server health')
     ->param('interactive', 'n', new Text(1), 'Run an interactive session', true)
-    ->action(function ($interactive) use ($register) {
+    ->param('receiver', '', new Wildcard(), 'Send an email to this address to test SMTP', true)
+    ->action(function ($interactive, $receiver) use ($register) {
         Console::log("  __   ____  ____  _  _  ____  __  ____  ____     __  __  
  / _\ (  _ \(  _ \/ )( \(  _ \(  )(_  _)(  __)   (  )/  \ 
 /    \ ) __/ ) __/\ /\ / )   / )(   )(   ) _)  _  )((  O )
@@ -139,7 +141,10 @@ $cli
         try {
             $mail = $register->get('smtp'); /* @var $mail \PHPMailer\PHPMailer\PHPMailer */
            
-            if ($interactive == "Y" || $interactive == "y") {
+            if ($receiver != '') {
+                $mail->addAddress($receiver);
+            }
+            else if ($interactive == "Y" || $interactive == "y") {
                 $receiver = Console::confirm('Enter your email address to test SMTP connection: ');
                 $mail->addAddress($receiver);
             } else {
--- a/app/tasks/doctor.php
+++ b/app/tasks/doctor.php
@@ -141,7 +141,7 @@ $cli
             $mail = $register->get('smtp'); /* @var $mail \PHPMailer\PHPMailer\PHPMailer */
             if (empty($receiver)) {
                 $mail->addAddress($receiver);
-            } elseif ($interactive == "Y" || $interactive == "y") {
+            } elseif (strtolower($interactive === "y")) {
                 $receiver = Console::confirm('Enter your email address to test SMTP connection: ');
                 $mail->addAddress($receiver);
             } else {
--- a/app/tasks/doctor.php
+++ b/app/tasks/doctor.php
@@ -9,11 +9,13 @@ use Utopia\Storage\Storage;
 use Utopia\App;
 use Utopia\CLI\Console;
 use Utopia\Domains\Domain;
+use Utopia\Validator\Text;
 
 $cli
     ->task('doctor')
     ->desc('Validate server health')
-    ->action(function () use ($register) {
+    ->param('interactive', 'n', new Text(1), 'Run an interactive session', true)
+    ->action(function ($interactive) use ($register) {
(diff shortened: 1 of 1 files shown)


app/tasks/doctor.php
```diff
@@ -9,11 +9,13 @@
 use Utopia\App;
 use Utopia\CLI\Console;
 use Utopia\Domains\Domain;
+use Utopia\Validator\Text;
 
 $cli
     ->task('doctor')
     ->desc('Validate server health')
-    ->action(function () use ($register) {
+    ->param('interactive', 'n', new Text(1), 'Run an interactive session', true)
```
stnguyen90, line 17: Good approach to make sure existing functionality still works!

Would you please also add a param to allow setting the address so that you can still do non-interactive but set a different email?

app/tasks/doctor.php
```diff
@@ -133,10 +135,17 @@
             }
         }
 
+
         try {
             $mail = $register->get('smtp'); /* @var $mail \PHPMailer\PHPMailer\PHPMailer */
+
```
stnguyen90, line 141: Also, make sure to run the formatter to fix linting problems.

mustansirgodhrawala: Absolutely, my apologies.

app/tasks/doctor.php
mustansirgodhrawala: Hey there, this is done and tested. I've also linted the file. Take a look when you can, thanks.

fanatic75 (changes requested):

app/tasks/doctor.php
```diff
@@ -133,10 +137,18 @@
             }
         }
 
+
```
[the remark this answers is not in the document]
mustansirgodhrawala: All done

app/tasks/doctor.php
```diff
@@ -135,8 +139,15 @@
 
         try {
             $mail = $register->get('smtp'); /* @var $mail \PHPMailer\PHPMailer\PHPMailer */
+            if ($receiver != '') {
```
fanatic75, line 142: Let's use 
```suggestion
            if (empty($receiver )) {
```

mustansirgodhrawala: Done, thanks for the help. I was not aware of that keyword.

app/tasks/doctor.php
```diff
@@ -135,8 +139,15 @@
...
+                $mail->addAddress($receiver);
+            } elseif ($interactive == "Y" || $interactive == "y") {
```
fanatic75, line 144: Let's convert interactive to lower case and keep one condition check
```suggestion
            } elseif (strtolower($interactive === "y")) {
```

mustansirgodhrawala: Thanks, looks like I really need to read up a bit on php....

app/tasks/doctor.php
```diff
@@ -135,8 +139,15 @@
...
         try {
             $mail = $register->get('smtp'); /* @var $mail \PHPMailer\PHPMailer\PHPMailer */
+            if (empty($receiver)) {
+                $mail->addAddress($receiver);
+            } elseif (strtolower($interactive === "y")) {
+                $receiver = Console::confirm('Enter your email address to test SMTP connection: ');
+                $mail->addAddress($receiver);
+            } else {
+                $mail->addAddress('demo@example.com', 'Example.com');
+            }
```
fanatic75, line 149: If the interactive option is set to 'Y' and a value for `$receiver` is also provided, what will occur?

app/tasks/doctor.php
```diff
@@ -135,8 +139,15 @@
 
         try {
             $mail = $register->get('smtp'); /* @var $mail \PHPMailer\PHPMailer\PHPMailer */
+            if (empty($receiver)) {
```
fanatic75, line 142: Hmm, looks like you used empty but the condition is opposite now. 
Let's change it to 

```suggestion
            if (!empty($receiver)) {
```

mustansirgodhrawala:
Could we merge this pull request now?

mustansirgodhrawala:
@fanatic75 @stnguyen90 Could you please take a look and let me know anything else is needed to merge this pull?

app/tasks/doctor.php
```diff
@@ -9,11 +9,15 @@
...
 use Utopia\Domains\Domain;
+use Utopia\Validator\Text;
+use Utopia\Validator\Wildcard;
 
 $cli
     ->task('doctor')
     ->desc('Validate server health')
-    ->action(function () use ($register) {
+    ->param('interactive', 'n', new Text(1), 'Run an interactive session', true)
+    ->param('receiver', '', new Wildcard(), 'Send an email to this address to test SMTP', true)
```
stnguyen90, line 19: We have a [validator for email](https://github.com/appwrite/appwrite/blob/master/src/Appwrite/Network/Validator/Email.php) so let's use that instead of wildcard

stnguyen90 (changes requested):
Would you please also fix the merge conflicts? 🙏🏼

stnguyen90:
Closing due to inactivity.

Issue #4519, opened 2022-10-17
🐛 Bug Report: Incorrect Error When Creating Duplicate Domain


stnguyen90:
### 👟 Reproduction steps

1. Create a custom domain for a project
2. Create another custom domain that's the same as in step 1

### 👍 Expected behavior

Error indicates domain already exists

### 👎 Actual Behavior

Error says duplicate ID:

```json
{
    "message": "A Domain with the requested ID already exists.",
    "code": 409,
    "type": "domain_already_exists",
    "version": "dev"
}
```

### 📝 Tasks

1. [ ] Update error:

https://github.com/appwrite/appwrite/blob/6c220467b8cb33e3784dc99a580cc322b370f9f1/app/config/errors.php#L534

2. [ ] Add a test case:

https://github.com/appwrite/appwrite/blob/939f974d19f30a30c204fa8a7108b857e88673e7/tests/e2e/Services/Projects/ProjectsConsoleClientTest.php#L2158-L2203

3. [ ] Create PR w/ a screenshot of the updated error message

### 🎲 Appwrite version

Version 1.0.x

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

20RitikSingh:
I want to contribute to this.

Atharvk7:
Could you assign this issue to me
I want to contribute

Meldiron:
@20RitikSingh , thanks for your interest! 🙏 Happy hacking! 🎃

Meldiron:
@Atharvk7 , we're assigning people on a first come first serve basis. If @20RitikSingh  isn't able to complete this issue, we'll reassign accordingly. In the meantime, you can find another hacktoberfest issue via our [website](https://hacktoberfest.appwrite.io/).

yeya:
The error type says: `domain_already_exists`. Isn't it the expected behavior? You need the message say that too? Or maybe each domain has `name` field AND `id` filed?

stnguyen90:
@yeya it's just misleading that it says the `ID` already exists but it's the `name` that's causing the error.

vedant-z:
@stnguyen90 If you allow me I am working on this issue.

mustansirgodhrawala:
[...]

PR #6019, opened 2023-08-19
Added new testcase and modified duplicate domain error

mustansirgodhrawala:
## What does this PR do?

This PR addresses issue #4519, it performs two things.
1. Creates a new test case, for testing out duplicate domain for the same project and checks the error code. 
2. Modifies the previous error, to clearly state that the domain is the duplicate and not the id.
3. It fixes a previous test which lacked an assert statement

## Test Plan

I attempt to create a new project, with the same name as the one created in the previous test.

## Related PRs and Issues

- #4519 

## Checklist

### DIFF
--- a/app/config/errors.php
+++ b/app/config/errors.php
@@ -571,7 +571,7 @@ return [
     ],
     Exception::DOMAIN_ALREADY_EXISTS => [
         'name' => Exception::DOMAIN_ALREADY_EXISTS,
-        'description' => 'A Domain with the requested NAME already exists.',
+        'description' => 'A Domain with the requested name already exists.',
         'code' => 409,
     ],
     Exception::VARIABLE_NOT_FOUND => [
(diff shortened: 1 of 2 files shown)


app/config/errors.php
```diff
@@ -571,7 +571,7 @@
     ],
     Exception::DOMAIN_ALREADY_EXISTS => [
         'name' => Exception::DOMAIN_ALREADY_EXISTS,
-        'description' => 'A Domain with the requested ID already exists.',
+        'description' => 'A Domain with the requested NAME already exists.',
```
stnguyen90, line 574: Name doesn't need to be all uppercase

stnguyen90 (changes requested):
Great job! I left one comment. Also, it looks like there's a lint problem somewhere.

app/config/errors.php
mustansirgodhrawala: All good! Thanks for the input

mustansirgodhrawala:
All recommended changes complete, we should be good to go! Thanks for the help

stnguyen90:
@mustansirgodhrawala sorry, but 1.4 made some changes to domains (replaced them with rules) so I don't think this is a problem anymore.

PR #6020, opened 2023-08-19
Update Dockerfile

Rithwik0480:
# Executables
`RUN chmod +x -R /usr/local/bin`  made this change in dockerfile so that every file in `/usr/local/bin` will get the exec permission.

## What does this PR do?

`RUN chmod +x -R /usr/local/bin`  made this change in dockerfile so that every file in `/usr/local/bin` will get the exec permission.

## Test Plan
I build the docker image locally and tested.

## Related PRs and Issues

- (Related PR or issue)

## Checklist


abnegate:
Hey @Rithwik0480, thanks for the PR! While your way in certainly less verbose and easier to maintain, it doesn't fit with our principle of security by default to give everything in `/usr/local/bin` execute access when we have no control over the binaries already in there.

PR #6022, opened 2023-08-20
Rollback to old usage

shimonewman:


app/controllers/api/functions.php
```diff
@@ -430,7 +430,7 @@
 
 App::get('/v1/functions/:functionId/usage')
```
eldadfux, line 431: We're missing `->label('usage.metric'` label in this controller.

app/controllers/api/project.php
```diff
@@ -3,13 +3,10 @@
 use Appwrite\Extend\Exception;
 use Appwrite\Utopia\Response;
 use Utopia\App;
```
eldadfux, line 5: Some endpoints that we do need have been removed here.

app/controllers/api/projects.php
```diff
@@ -95,6 +95,7 @@
         $backups['database_db_fra1_03'] = ['from' => '10:30', 'to' => '11:15'];
         $backups['database_db_fra1_04'] = ['from' => '13:30', 'to' => '14:15'];
         $backups['database_db_fra1_05'] = ['from' => '4:30', 'to' => '5:15'];
+        $backups['database_db_fra1_06'] = ['from' => '16:30', 'to' => '17:15'];
```
eldadfux, line 98: How is this related to usage PR? it seems like you got to branches mixed.

composer.json
```diff
@@ -70,6 +70,7 @@
         "resque/php-resque": "1.3.6",
         "matomo/device-detector": "6.1.*",
         "dragonmantank/cron-expression": "3.3.2",
+        "influxdb/influxdb-php": "1.15.2",
         "phpmailer/phpmailer": "6.8.0",
         "chillerlan/php-qrcode": "4.3.4",
         "adhocore/jwt": "1.1.2",
```
eldadfux, line 76: We forgot to remove statsd last time, let's make a note when we recover this feature for next release.

eldadfux (changes requested):

app/controllers/api/projects.php
shimonewman: It was intently.
I was comparing against usage from the cloud branch
and noticed that a db was added I thought it should be updated too.

PR #6029, opened 2023-08-22
Fix colliding ID overrides

abnegate:
## What does this PR do?

- Fix upload file with colliding ID overwriting original

## Test Plan

- Added test

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Extend/Exception.php
+++ b/src/Appwrite/Extend/Exception.php
@@ -105,6 +105,7 @@ class Exception extends \Exception
     public const AVATAR_ICON_NOT_FOUND             = 'avatar_icon_not_found';
 
     /** Storage */
+    public const STORAGE_FILE_ALREADY_EXISTS     = 'storage_file_already_exists';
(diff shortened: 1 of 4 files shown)


src/Appwrite/Extend/Exception.php
```diff
@@ -105,6 +105,7 @@ class Exception extends \Exception
     public const AVATAR_ICON_NOT_FOUND             = 'avatar_icon_not_found';
 
     /** Storage */
+    public const STORAGE_FILE_ALREADY_EXISTS     = 'storage_file_already_exists';
```
eldadfux, line 108: Missing space

eldadfux (approved):

src/Appwrite/Extend/Exception.php
abnegate, line 108:
```suggestion
    public const STORAGE_FILE_ALREADY_EXISTS       = 'storage_file_already_exists';
```

PR #6047, opened 2023-08-28
Feat SMTP Template Simplification

PineappleIOnic:
## What does this PR do?

This PR removes HTML from the SMTP Templates, moving it into it's own template and making updating them easier

## Checklist


eldadfux (changes requested):
- Please share some screenshot of the templates look at the console now after this change.
- I think we miss a few places where we send emails, like the teams controller, please double check we haven't miss any other places.
- Not sure we need to change the syntax of how we set all the settings from emails in the controllers, we can leave as is and probably improve it in a later iteration, as it's more complex than it should be.

PineappleIOnic:
> * Please share some screenshot of the templates look at the console now after this change.
> * I think we miss a few places where we send emails, like the teams controller, please double check we haven't miss any other places.
> * Not sure we need to change the syntax of how we set all the settings from emails in the controllers, we can leave as is and probably improve it in a later iteration, as it's more complex than it should be.

image
image
image
image

eldadfux (approved):

PR #6068, opened 2023-08-30
Feat upgrade task

stnguyen90:
## What does this PR do?

Brings code from https://github.com/appwrite/appwrite/pull/5874 into 1.4.x

> Before, we used the same command for both installation and upgrades. This led to problems because developers would try to upgrade in the wrong folder and end up creating a new installation.
> 
> This new upgrade command validates the existence of an existing installation before proceeding with the upgrade to ensure no new installation is created when upgrading.

## Test Plan

Manual:

### Install

image

### Upgrade in wrong folder

image

### Upgrade cancel

image

### Upgrade confirm

image

## Related PRs and Issues

- https://github.com/appwrite/appwrite/pull/5874

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Tasks/Install.php
+++ b/src/Appwrite/Platform/Tasks/Install.php
@@ -230,9 +230,9 @@ class Install extends Action
             }
         }
 
-        Console::log("Running \"docker compose -f $this->path/docker-compose.yml up -d --remove-orphans --renew-anon-volumes\"");
+        Console::log("Running \"docker compose up -d --remove-orphans --renew-anon-volumes\"");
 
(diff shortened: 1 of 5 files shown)


src/Appwrite/Platform/Tasks/Install.php
```diff
@@ -56,19 +57,28 @@ public function action(string $httpPort, string $httpsPort, string $organization
...
                 Console::exit(1);
             }
         }
 
-        $data = @file_get_contents($path . '/docker-compose.yml');
+        $data = @file_get_contents($this->path . '/docker-compose.yml');
 
         if ($data !== false) {
+            if ($interactive == 'Y' && Console::isInteractive()) {
+                $answer = Console::confirm('Previous installation found, do you want to overwrite it? (Y/n)');
```
abnegate, line 71: Let's add a note that the previous config will be backed up in case of error

abnegate (approved):

PR #6072, opened 2023-08-30
1.4 Fixes

abnegate:
## What does this PR do?

- Add upgrade task
- Fix VCS/migrations/assistant scopes
- Fix domains migration
- Fix missing variable in email templates
- Fix call on missing route
- Fix missing mock headers endpoint
- Update console with scope fixes
- Update specs

## Test Plan

Test suite + manual tests

## Related PRs and Issues

- (Related PR or issue)

## Checklist


stnguyen90 (approved):

app/workers/mails.php
```diff
@@ -35,9 +35,9 @@ public function run(): void
 
         $recipient = $this->args['recipient'];
         $subject = $this->args['subject'];
-        $name = $this->args['name'];
         $body = $this->args['body'];
         $variables = $this->args['variables'];
+        $name = $this->args['name'];
```
christyjacob4, line 40:
```suggestion
        $name = $variables['user'] ?? '';
```

app/controllers/api/migrations.php
```diff
@@ -580,7 +580,7 @@
 App::get('/v1/migrations/firebase/connect')
     ->desc('Authorize with firebase')
     ->groups(['api', 'migrations'])
-    ->label('origin', '*')
+    ->label('scope', 'migrations.read')
```
christyjacob4, line 583: hmm shouldn't this be a write scope as there's an updateDocument happening here?

christyjacob4 (approved):

app/workers/mails.php
abnegate: This should `args['name']` as it is used as `from` name, `variables['user']` is the `to` name

app/controllers/api/migrations.php
abnegate: Makes senses, PR to update it [here](https://github.com/appwrite/appwrite/pull/6083)

christyjacob4 (approved):

PR #6077, opened 2023-08-30
fix: missing variable in email templates

christyjacob4:
## What does this PR do?

The email templates in the console expect {{user}}, {{team}}, {{redirect}} and {{project}} as template variables. 
eg:
1. https://github.com/appwrite/console/blob/main/src/routes/console/project-%5Bproject%5D/auth/templates/emailInviteTemplate.svelte#L40-L43

But some of these were missing in our implementation.

## Before 
Screenshot 2023-08-30 at 6 28 08 PM

## After
Screenshot 2023-08-30 at 6 25 32 PM

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/account.php
```diff
@@ -2503,15 +2507,18 @@
             'hello' => $locale->getText("emails.recovery.hello"),
             'name' => $profile->getAttribute('name'),
```
abnegate, line 2508: Do we still need `name` if we updated it to `user`?

abnegate (commented):
Looks good, just one question and needs a linter run

app/controllers/api/account.php
christyjacob4: Not necessarily. I can refactor the mails worker to not use `name` anymore 👍

abnegate (approved):

PR #6079, opened 2023-08-30
Fix router error

stnguyen90:
## What does this PR do?

It's possible for the route to be null (no match) which leads to an error like:

```
appwrite  | [Error] Timestamp: 2023-08-30T21:39:58+00:00
appwrite  | [Error] Type: Error
appwrite  | [Error] Message: Call to a member function label() on null
appwrite  | [Error] File: /usr/src/code/app/controllers/general.php
appwrite  | [Error] Line: 50
```

This adds a null check to prevent the error.

## Test Plan

Manually made an API call to a custom endpoint and confirmed no more "Call to a member function label() on null" error.

## Related PRs and Issues

None

## Checklist


christyjacob4 (approved):
Good catch 👀

PR #6080, opened 2023-08-30
Create a GH workflow to cleanup the GH cache

stnguyen90:
## What does this PR do?

The GitHub actions cache is helpful to speed up CI jobs. Once a branch is closed, we don't need the cacche anymore so we should delete it so the disk space can be reclaimed.

This workflow was copied from https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries.

## Test Plan

https://github.com/appwrite/appwrite/actions/runs/6031274668/job/16364664657?pr=6080

## Related PRs and Issues

None

## Checklist

### DIFF
--- a/.github/workflows/cleanup-cache.yml
+++ b/.github/workflows/cleanup-cache.yml
@@ -1,0 +1,42 @@
+name: Cleanup Cache
+
+# for testing
+on: [pull_request]
+
+# on:
+#   pull_request:
+#     types:
+#       - closed
+
+jobs:
+  cleanup:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+        
+      - name: Cleanup
+        run: |
+          gh extension install actions/gh-actions-cache
+          
+          REPO=${{ github.repository }}
+          BRANCH="refs/pull/${{ github.event.pull_request.number }}/merge"
+
+          while true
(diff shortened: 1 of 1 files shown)


christyjacob4 (approved):

.github/workflows/cleanup-cache.yml
```diff
@@ -0,0 +1,37 @@
...
+        
+      - name: Cleanup
+        run: |
+          gh extension install actions/gh-actions-cache
+          
+          REPO=${{ github.repository }}
+          BRANCH="refs/pull/${{ github.event.pull_request.number }}/merge"
+
+          echo "Fetching list of cache key"
+          cacheKeysForPR=$(gh actions-cache list -R $REPO -B $BRANCH -L 100 | cut -f 1 )
```
abnegate, line 26: I think if we remove `-L 100` it will clear the entire cache instead of only the oldest 100

stnguyen90: Max limit is 100: https://github.com/actions/gh-actions-cache#list

abnegate: Can we paginate and delete them all?

stnguyen90: Updated

PR #6098, opened 2023-08-31
Fix the default executor host since it's changed in 1.4

stnguyen90:
## What does this PR do?

Since the executor's hostname is now `executor`

https://github.com/appwrite/appwrite/blob/36698d180c44c631576607569f4255afc2ead7ca/app/views/install/compose.phtml#L592-L595

the default variable value should be updated.

## Test Plan

Manual

## Related PRs and Issues

None

## Checklist


abnegate:
I think we'll need to change the Dockerfile default as well

stnguyen90:
@abnegate, the env var in the Dockerfile shouldn't matter after https://github.com/appwrite/appwrite/pull/5971

app/config/variables.php
```diff
@@ -748,7 +748,7 @@
                 'name' => '_APP_EXECUTOR_HOST',
                 'description' => 'The host used by Appwrite to communicate with the function executor!',
                 'introduction' => '0.13.0',
-                'default' => 'http://appwrite-executor/v1',
+                'default' => 'http://executor/v1',
```
Meldiron, line 751: Looks good, but maybe we should go back to using `appwrite-executor`, or have smart migration script (if value is exactly `http://appwrite-executor/v1`, then replace. If not, keep using their modified value)

stnguyen90: let's leave it as executor. smarter installation command coming in: https://github.com/appwrite/appwrite/issues/6137

christyjacob4: @stnguyen90 the override PR has been merged. I think now we should change it back to `appwrite-executor`

PR #6127, opened 2023-09-03
fix-6117-empty-users-getting-created-in-appwrite-console

btme0011:
## What does this PR do?
Implemented validation checks for the name, email, and password fields. If any of these validations fail, an exception is thrown and the corresponding error message is displayed to the user.

## Test Plan

Before changes 
We were able to create empty users

After changes 
Screenshot 2023-09-03 at 11 08 51 AM

## Checklist

### DIFF
--- a/app/config/errors.php
+++ b/app/config/errors.php
@@ -240,6 +240,27 @@ return [
         'description' => 'OAuth2 provider returned some error.',
         'code' => 424,
     ],
+    Exception::USER_NAME_INVALID => [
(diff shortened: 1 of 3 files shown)


eldadfux:
@btme0011 we appreciate the effort and contribution, but this is the expected behavior. All this details are optional to allow you to simulate different type of users through different login methods (email+password, phone, magic URL, anonymous and more).

It is always best to open an issue and discuss before investing time in making a PR. That said, great job with the code implementation.

Issue #6128, opened 2023-09-03
🐛 Bug Report: Pre 1.4 function doesn't work


stnguyen90:
### 👟 Reproduction steps

1. start on Appwrite version 1.3.8
2. create a function that works
3. upgrade to 1.4.x
4. execute the function

### 👍 Expected behavior

The function executes successfully.

### 👎 Actual Behavior

Error occurs. 

Here's the exeuction result:

```json
{
  "$id": "64f41a33aa19a4085416",
  "$createdAt": "2023-09-03T05:31:31.697+00:00",
  "$updatedAt": "2023-09-03T05:31:42.595+00:00",
  "$permissions": [
    "read(\"user:62f160ce77c57e6e6ac3\")"
  ],
  "functionId": "dump",
  "trigger": "http",
  "status": "failed",
  "requestMethod": "GET",
  "requestPath": "/",
  "requestHeaders": [],
  "responseStatusCode": 500,
  "responseBody": "",
  "responseHeaders": [],
  "logs": "",
  "errors": "An internal curl error has occurred within the executor! Error Msg: Operation timed out\\nError Code: 500",
  "duration": 10.888585090637
}
```

The function seems to try to build, but then fails. You can see it exited:

<img width="1096" alt="image" src="https://github.com/appwrite/appwrite/assets/1477010/fb82e08e-9131-4a3d-9458-59cf07e606ef">

Inspecting the container gives:

```json
[
    {
        "Id": "f9452024db1d3f24dfb96f8355d8d0f21c2d752410b68171ecc11a77795c319e",
        "Created": "2023-09-03T05:31:31.804240093Z",
        "Path": "sh",
        "Args": [
            "-c",
            "cp /tmp/code.tar.gz /mnt/code/code.tar.gz \u0026\u0026 nohup helpers/start.sh \"src/function/server\""
        ],
        "State": {
            "Status": "exited",
            "Running": false,
            "Paused": false,
            "Restarting": false,
            "OOMKilled": false,
            "Dead": false,
            "Pid": 0,
            "ExitCode": 127,
            "Error": "",
            "StartedAt": "2023-09-03T05:31:32.047404259Z",
            "FinishedAt": "2023-09-03T05:31:32.055148676Z"
        },
        "Image": "sha256:4c0bda687e945d844c251a796c585e910c7df8ab4f1cd22baeea4f4c2fed7dea",
[...]

PR #6142, opened 2023-09-04
Fix: v2 functions

Meldiron:
## What does this PR do?

Ensures pre-1.4 functions work after upgrading to 1.4.

## Test Plan

Manual QA. I was able to execute, and even make new deployment with old code.

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/6128

## Checklist


app/controllers/api/functions.php
```diff
@@ -1523,7 +1523,8 @@
             throw new Exception(Exception::FUNCTION_NOT_FOUND);
         }
 
-        $runtimes = Config::getParam('runtimes', []);
+        $version = $function->getAttribute('version', 'v2');
+        $runtimes = Config::getParam($version === 'v2' ? 'runtimes-v2' : 'runtimes', []);
```
christyjacob4, line 1527: Dont we also need to update it here
https://github.com/appwrite/appwrite/blob/fc3ff188b143d55bc7a6c1fda3072b3913f89a6b/app/controllers/api/vcs.php#L478

and here
https://github.com/appwrite/appwrite/blob/fc3ff188b143d55bc7a6c1fda3072b3913f89a6b/app/controllers/api/vcs.php#L558

Meldiron: That's for runtime auto-detection IIRC. If you are connecting v2 function to git, you are weird.
Also, if detection fails, Console gives you dropdown so you can pick

Meldiron: Similiarly, listRuntimes endpoint will return v3 stuff even if your function is v2. But same rule, you shouldn't really be changing your runtime if your function is v2.

If you are actively working on your function (connecting it to git and upgrading runtime), you should also upgrade to v3. V2 support should be purely to not break existing things

christyjacob4: lets say someone has an existing v2 function and only want to connect it to git without really ugrading the code/syntax.

In that case we need to account for v2 runtimes here too right ?

stnguyen90:
I tried this code on an old dart function and it looks like the right image is created now, but t still exits:

image

docker logs:

```
$ docker logs f7b37dc71597
qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
```

app/controllers/api/functions.php
Meldiron: library appwrite/runtimes isn't really aware of v2 list anyway.
Let's say there was V2 Node 16, and now there is V3 Node 18..
In Appwrite 1.4, it will only know "Node 18". So it will think "v2 Node 18" is a thing, even if it isn't.

If we want to fully be aware of v2 runtimes, we need to introduce and maintain them here: https://github.com/appwrite/runtimes

I personally don't really see value in that. If you want to connect git, you actively work on project. You should migrate to V3.

Also, from technical point of view, this VCS endpoint isn't aware of function. It just gets repo, and does detection. We would need to add `version` parameter to the VCS endpoint. I really dont like touching code this much for simple backwards compatibility.

app/controllers/mock.php
```diff
@@ -605,6 +607,32 @@
             ]);
     });
 
+App::patch('/v1/mock/functions-v2')
+    ->desc('OAuth2 Failure')
```
christyjacob4, line 611: can you please update the descriptions, and also add the `api` & `functions` group to the `->groups()` function ?

christyjacob4 (changes requested):
Please address the comments

Issue #6113, opened 2023-09-01
🐛 Bug Report: Project Global Variables Don't Work in Functions


stnguyen90:
### 👟 Reproduction steps

1. Create a project global variable
2. Create a function that reads the variable
3. Execute the function

### 👍 Expected behavior

The project variable is available in the function.

### 👎 Actual Behavior

The project variable is not available in the function.

I created a global variable named `APPWRITE_ENDPOINT`:

![image](https://github.com/appwrite/appwrite/assets/1477010/a454817f-3be5-439f-a5a5-f6875a7cfbad)

I deployed this code:

```dart
import 'dart:async';
import 'dart:io';

Future<dynamic> main(final context) async {
  return context.res.json({
    'environment': Platform.environment,
  });
}
```

I deployed the function and made sure it didn't have any pending changes to be deployed: 

![image](https://github.com/appwrite/appwrite/assets/1477010/ece254b4-e419-40fb-b8e6-9794c39a73a8)

Execution:

<img width="735" alt="image" src="https://github.com/appwrite/appwrite/assets/1477010/cd6c0306-26d7-47d0-8259-faf86eabe3eb">

### 🎲 Appwrite version

Version 1.4.x

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

PR #6150, opened 2023-09-05
Fix: Global variables

Meldiron:
## What does this PR do?

Global variables didn't work due to missing subquery. Now they should.

This includes change to migration, so after release, devs needs to re-run migration

## Test Plan

- [x] Manual QA

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/6113

## Checklist

### DIFF
--- a/app/controllers/api/project.php
+++ b/app/controllers/api/project.php
@@ -161,7 +161,6 @@ App::post('/v1/project/variables')
         } catch (DuplicateException $th) {
             throw new Exception(Exception::VARIABLE_ALREADY_EXISTS);
         }
-        $dbForConsole->deleteCachedDocument('projects', $project->getId());
 
         $functions = $dbForProject->find('functions', [
             Query::limit(APP_LIMIT_SUBQUERY)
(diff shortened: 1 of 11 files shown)


app/controllers/api/project.php
```diff
@@ -161,18 +161,16 @@
...
         }
-        $dbForConsole->deleteCachedDocument('projects', $project->getId());
 
         $functions = $dbForProject->find('functions', [
             Query::limit(APP_LIMIT_SUBQUERY)
         ]);
 
         foreach ($functions as $function) {
             $dbForProject->updateDocument('functions', $function->getId(), $function->setAttribute('live', false));
+            $dbForProject->deleteCachedDocument('functions', $function->getId());
```
stnguyen90, line 171: This isn't needed because the updateDocument() should purge the cache, right?

https://github.com/utopia-php/database/blob/cc0247f4f0c402b39f663bf9f77b29d69b95f9d6/src/Database/Database.php#L2928

stnguyen90 (approved):

app/controllers/api/project.php
Meldiron: True, removed from multiple places (everywhere we set live=false)

christyjacob4 (approved):

PR #6155, opened 2023-09-05
Fix: Untrusted custom domains + auto-ssl

Meldiron:
## What does this PR do?

- Automatically issues SSL for first domain connected to new instance
- Servers API+Console by default, if domain is not known

## Test Plan

- [x] Manual QA

## Related PRs and Issues

x

## Checklist

### DIFF
--- a/app/controllers/general.php
+++ b/app/controllers/general.php
@@ -158,6 +158,7 @@ function router(App $utopia, Database $dbForConsole, SwooleRequest $swooleReques
 
         return true;
     } elseif ($type === 'api') {
+        $utopia->getRoute()?->label('error', '');
         return false;
     } else {
         throw new AppwriteException(AppwriteException::GENERAL_SERVER_ERROR, 'Unknown resource type ' . $type);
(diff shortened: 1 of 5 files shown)


app/controllers/general.php
```diff
@@ -234,6 +230,61 @@ function router(App $utopia, Database $dbForConsole, SwooleRequest $swooleReques
...
+                            ->setDomain($domainDocument)
+                            ->setSkipRenewCheck(true)
+                            ->trigger();
+                    }
+                }
+                $domains[$domain->get()] = true;
+
+                Authorization::reset(); // ensure authorization is re-enabled
+            }
+            Config::setParam('domains', $domains);
```
christyjacob4, line 285: im not sure this works as intended when we have multiple workers, as each worker holds its own copy of config in its memory.

christyjacob4: I dont think config behaves as a static variable

christyjacob4 (approved):

PR #6219, opened 2023-09-12
Extracting function event validator to its own class.

### LINKED ISSUE #6218: 🐛 Bug Report: Can't set function event to trigger Webhook
### 👟 Reproduction steps

When trying to create a Webhook that will be triggered by any function-relate events you get the following error.
```
Invalid `events` param: Value must a valid array 
and Triggering a function on a function event is not allowed."
```
image

The payload that was sent in the test
```json
{
  "name": "Test",
  "events": [
    "functions.*.executions.*"
  ],
  "url": "https://test.com/test",
  "security": false
}
```

### 👍 Expected behavior

Appwrite should create a function-based triggered webhook.

### 👎 Actual Behavior

Appwrite blocks any function-based event creation, as seen [here](https://github.com/appwrite/appwrite/blob/master/src/Appwrite/Event/Validator/Event.php#L50) for all available triggers, function and webhooks and should block [only](https://discord.com/channels/564160730845151244/1150634893882175528/1150910514118008843) functions.

### 🎲 Appwrite version

Version 1.4.x

### 💻 Operating system

Linux

### 🧱 Your Environment

_No response_

byawitz:
## What does this PR do?
Closes #6218 

## Test Plan
1. Added `FunctionEventValidatorTest.php` which is the same as `EventValidatorTest.php` except that any event that starts with `functions.` will return false.
2. Tested it on a temporary local Appwrite instance, and I was able to add the `functions.*` events to a webhook but not to a function.

## Related PRs and Issues

#6218

## Checklist

### DIFF
--- a/app/controllers/api/functions.php
+++ b/app/controllers/api/functions.php
@@ -7,7 +7,7 @@ use Appwrite\Event\Delete;
 use Appwrite\Event\Event;
 use Appwrite\Event\Func;
 use Appwrite\Event\Usage;
-use Appwrite\Event\Validator\Event as ValidatorEvent;
+use Appwrite\Event\Validator\FunctionEvent;
 use Appwrite\Utopia\Response\Model\Rule;
 use Appwrite\Extend\Exception;
 use Appwrite\Utopia\Database\Validator\CustomId;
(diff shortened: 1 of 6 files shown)


stnguyen90 (changes requested):
Would you please also uncomment the functions test in the event validator unit test?

byawitz:
Done

stnguyen90 (approved):

abnegate (approved):

Issue #6189, opened 2023-09-07
🐛 Bug Report: Git deployments when using a storage provider doesn't work


stnguyen90:
### 👟 Reproduction steps

1. Configure Appwrite to use a storage provider like wasabi
2. Configure Appwrite with github integration
3. Deploy a function using github

### 👍 Expected behavior

The deployment works.

### 👎 Actual Behavior

The deployment doesn't work.

appwrite-worker-builds logs:

![image](https://github.com/appwrite/appwrite/assets/1477010/e5f5ef0a-afb5-49b8-849c-fba9d2a80af0)

It seems these things need to be addressed:

Perhaps the appwrite-worker-builds container needs the storage environment variables:

https://github.com/appwrite/appwrite/blob/1.4.x/app/views/install/compose.phtml#L346-L385

### 🎲 Appwrite version

Version 1.4.x

### 💻 Operating system

Linux

### 🧱 Your Environment

Tested on 1.4.1

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

PR #6227, opened 2023-09-13
Fix: Git deploys with S3

Meldiron:
## What does this PR do?

Ensures deployments triggered from VCS source code work as expected when any S3 storage is connected.

## Test Plan

- [x] Manual QA

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/6189

## Checklist


app/workers/builds.php
```diff
@@ -248,22 +249,25 @@ protected function buildDeployment(GitHub $github, Document $project, Document $
...
-                $deviceFunctions = $this->getFunctionsDevice($project->getId());
+                Console::execute('tar --exclude code.tar.gz -czf ' . $tmpPathFile . ' -C /tmp/builds/' . \escapeshellcmd($buildId) . '/code' . (empty($rootDirectory) ? '' : '/' . $rootDirectory) . ' .', '', $stdout, $stderr);
 
-                $fileName = 'code.tar.gz';
-                $fileTmpName = '/tmp/builds/' . $buildId . '/code.tar.gz';
+                $deviceFunctions = $this->getFunctionsDevice($project->getId());
 
-                $path = $deviceFunctions->getPath($deployment->getId() . '.' . \pathinfo($fileName, PATHINFO_EXTENSION));
+                $localDevice = new Local();
+                $buffer = $localDevice->read($tmpPathFile);
```
stnguyen90, line 260: This could be dangerous if tar files are larger than the max memory allowed, right?

I think we're okay using `move()` because [`move()` calls `transfer()`](https://github.com/utopia-php/storage/blob/efec5376c02d3d8330f1beb1469e6d6e313e21ee/src/Storage/Device.php#L165) and `transfer()` will [chunk](https://github.com/utopia-php/storage/blob/efec5376c02d3d8330f1beb1469e6d6e313e21ee/src/Storage/Device/S3.php#L293-L299) if needed.

stnguyen90 (changes requested):

app/workers/builds.php
stnguyen90: Actually, move will still move from the provider so we can't use that.

stnguyen90: We'll need to handle chunking later

stnguyen90:
I wasn't able to test, but the code looks good.

Meldiron:
No worries, I tested it end to end. With DO Spaces

app/workers/builds.php
```diff
@@ -248,22 +249,25 @@ protected function buildDeployment(GitHub $github, Document $project, Document $
...
+                $mimeType = $localDevice->getFileMimeType($tmpPathFile);
 
-                $result = $deviceFunctions->move($fileTmpName, $path);
+                $path = $deviceFunctions->getPath($deployment->getId() . '.' . \pathinfo('code.tar.gz', PATHINFO_EXTENSION));
+                $result = $deviceFunctions->write($path, $buffer, $mimeType);
```
christyjacob4, line 264: @Meldiron why are we replacing the usage of the `move` function with a read and write?

The `move` function is designed to handle chunking etc.

christyjacob4: Using move() does not support transferring between devices in the storage library version that we currently use. That support was introduced in https://github.com/utopia-php/storage/releases/tag/0.16.0

We will upgrade to the new storage library with Appwrite 1.5

Issue #6212, opened 2023-09-10
🐛 Bug Report: Issues with LinkedIn OAuth 2.0


Ronitron007:
### 👟 Reproduction steps

So I am trying to using the OAuth2.0 for LinkedIn and the bug probably occurs when account.createOAuth2Session('LinkedIn') is used. 

While signIn with linkedIn initially used to have Oauth2.0 it uses OIDC now for SSO. 

<img width="781" alt="Screenshot 2023-09-10 at 6 10 04 PM" src="https://github.com/appwrite/appwrite/assets/26198593/24eb807b-b3d6-4733-b1ec-4f2be02fec44">

I think the name of scopes for linkedIn should be changed from
```
 'r_liteprofile',
 'r_emailaddress',
```

to 
```
'openid',
'profile',
'email'
```

### 👍 Expected behavior

I should be redirected to my success page. And there should be a new user logged in.

### 👎 Actual Behavior

I get an error on linkedIn's page and I am redirected to an appWrite url. 

The appWrite url is 
[https://cloud.appwrite.io/v1/account/sessions/oauth2/callback/linkedin/64fc38061af407cce741?error=unauthorized_scope_error&error_description=Scope+%26quot%3Br_liteprofile%26quot%3B+is+not+authorized+for+your+application&state=%7B%22success%22%3A%22http%3A%5C%2F%5C%2Flocalhost%3A3001%5C%2F%22%2C%22failure%22%3A%22http%3A%5C%2F%5C%2Flocalhost%3A3001%5C%2Ffail%22%7D](https://cloud.appwrite.io/v1/account/sessions/oauth2/callback/linkedin/64fc38061af407cce741?error=unauthorized_scope_error&error_description=Scope+%26quot%3Br_liteprofile%26quot%3B+is+not+authorized+for+your+application&state=%7B%22success%22%3A%22http%3A%5C%2F%5C%2Flocalhost%3A3001%5C%2F%22%2C%22failure%22%3A%22http%3A%5C%2F%5C%2Flocalhost%3A3001%5C%2Ffail%22%7D)

I pasted the url since the url does have the error in it 

> error: unauthorized_scope_error
> error_description: Scope &quot;r_liteprofile&quot; is not authorized for your application
> state: cfa970987c9a3c2178dddeda0d81dab3e00dce4c683098d6

[...]

PR #6229, opened 2023-09-13
fix-6212-Issue-With-Linkedin-OAuth

btme0011:
## What does this PR do?
Fix related to the issue - #6212 

Reference document - https://learn.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/sign-in-with-linkedin-v2?context=linkedin%2Fconsumer%2Fcontext

LinkedIn has transitioned to using OpenID for authentication and authorization, necessitating updates to the existing codebase to accommodate these changes.
Screenshot 2023-09-13 at 9 42 03 PM

## Test Plan

Screenshot 2023-09-14 at 1 18 31 AM

## Related PRs and Issues

- #6212 

## Checklist

### DIFF
--- a/src/Appwrite/Auth/OAuth2/Linkedin.php
+++ b/src/Appwrite/Auth/OAuth2/Linkedin.php
@@ -20,8 +20,9 @@ class Linkedin extends OAuth2
      * @var array
      */
     protected array $scopes = [
-        'r_liteprofile',
-        'r_emailaddress',
+        'openid',
+        'profile',
+        'email'
     ];
 
     /**
@@ -118,7 +119,7 @@ class Linkedin extends OAuth2
     {
         $user = $this->getUser($accessToken);
 
-        return $user['id'] ?? '';
+        return $user['sub'] ?? '';
     }
 
     /**
@@ -128,9 +129,8 @@ class Linkedin extends OAuth2
      */
     public function getUserEmail(string $accessToken): string
     {
-        $email = \json_decode($this->request('GET', 'https://api.linkedin.com/v2/emailAddress?q=members&projection=(elements*(handle~))', ['Authorization: Bearer ' . \urlencode($accessToken)]), true);
-
-        return $email['elements'][0]['handle~']['emailAddress'] ?? '';
+        $user = $this->getUser($accessToken);
+        return $user['email'] ?? '';
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Appwrite/Auth/OAuth2/Linkedin.php
```diff
@@ -159,12 +158,12 @@ public function getUserName(string $accessToken): string
...
-        if (isset($user['localizedFirstName'])) {
-            $name = $user['localizedFirstName'];
+        if (isset($user['given_name'])) {
+            $name = $user['given_name'];
         }
 
-        if (isset($user['localizedLastName'])) {
-            $name = (empty($name)) ? $user['localizedLastName'] : $name . ' ' . $user['localizedLastName'];
+        if (isset($user['family_name'])) {
+            $name = (empty($name)) ? $user['family_name'] : $name . ' ' . $user['family_name'];
```
stnguyen90, line 166: Let's just use the name from LinkedIn:

```php
        if (isset($user['name'])) {
            return $user['name'];
        }
```

src/Appwrite/Auth/OAuth2/Linkedin.php
```diff
@@ -144,9 +144,8 @@ public function getUserEmail(string $accessToken): string
      */
     public function isEmailVerified(string $accessToken): bool
     {
-        $email = $this->getUserEmail($accessToken);
-
-        return !empty($email);
+        $user = $this->getUser($accessToken);
+        return $user['email_verified'] ?? '';
```
stnguyen90, line 148: the function signature says this should return bool, but this line hear means it's possible for it to return a string.

```suggestion
        return $user['email_verified'] ?? false;
```

stnguyen90 (changes requested):

src/Appwrite/Auth/OAuth2/Linkedin.php
btme0011: in some cases name might be not present and given_name or family_name might be present so should i add this case also with the other two cases? like if name is not present take info from given_name or family_name ?

src/Appwrite/Auth/OAuth2/Linkedin.php
stnguyen90: Theoretically, an exception would be thrown elsewhere. For now, to keep things consistent, let's just return false.

src/Appwrite/Auth/OAuth2/Linkedin.php
stnguyen90: Oh really? That's odd behavior...sure, go ahead and fallback to given and family name.

venus-3:
@btme0011 

Can you please let me know when are you going to merge this PR, I am really waiting for this changes to be merged, I am experiencing the above issue while Integrating it with the Linkedin.

TorstenDittmann (approved):

stnguyen90 (approved):
I tested and was able to successfully log in:

image

PR #6270, opened 2023-09-18
Fix: build race condition

Meldiron:
## What does this PR do?

If you create execution right after build, it tries to create container but that (for some reason) still exist from build. You get error

```
An internal curl error has occurred while starting runtime! Error Msg: Docker Error: docker: Error response from daemon: Conflict. The container name "/exc0-6504acaa254213470be0-6507566cd981c5675a7f" is already in use by container "3db20471263767d391a7a0d2929eca60be43349b700c84467ca4efeec2c4de34". You have to remove (or rename) that container to be able to reuse that name.
See 'docker run --help'.
Preparing for build ...
Building ...

added 13 packages, and audited 14 packages in 1s

2 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
Packing build ...
Build finished.
\nError Code: 500
```

To prevent this, we give exec container different name than build container with the suffix. That should fully separate those concerns, and not cause this troble.

## Test Plan

None. This used to happen randomly and rarely. Current CI/CD must pass.

## Related PRs and Issues

x

## Checklist


lohanidamodar (approved):
This is a temporary fix, we should remember to debug and make better solution later.

christyjacob4 (approved):

Issue #6245, opened 2023-09-15
🐛 Bug Report: Functions Timing Out During Build


dishwasher-detergent:
### 👟 Reproduction steps

Deploy function using Github integration

### 👍 Expected behavior

Function is deployed and usable.

### 👎 Actual Behavior

Function says building for 15 minutes before timing out.

![image](https://github.com/appwrite/appwrite/assets/33816169/6b0c16ca-bf6b-4923-b349-7378256dcdfd)
![image](https://github.com/appwrite/appwrite/assets/33816169/f217d43e-1abf-464c-9645-8b9ad99d802b)

### 🎲 Appwrite version

Version 1.4.x

### 💻 Operating system

Linux

### 🧱 Your Environment

Using Appwrite version 1.4.2 migrated from 1.3.8 -> 1.4.0 -> 1.4.1 -> 1.4.2. 

Self hosted in Digital Ocean 
![image](https://github.com/appwrite/appwrite/assets/33816169/b1b880ad-b089-4a10-8463-e741055a5b17)

GitHub Repo size is 475 MB.

[Discord Support Ticket](https://discord.com/channels/564160730845151244/1151513242074624140)

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

stnguyen90:
@dishwasher-detergent thanks for creating this issue! 🙏 to clarify, the repo is 475MB, but the function itself is just a single folder that doesn't use anything else in the repo and is how large?

stnguyen90:
@dishwasher-detergent, also, in the git settings, did you set a Root directory?

<img width="1151" alt="image" src="https://github.com/appwrite/appwrite/assets/1477010/d6efe49f-a3e0-4305-a6e3-62ce89fd378c">

dishwasher-detergent:
> @dishwasher-detergent thanks for creating this issue! 🙏 to clarify, the repo is 475MB, but the function itself is just a single folder that doesn't use anything else in the repo and is how large?

Below is the exact code from the function I'm trying to deploy.

```type Context = {
  req: any;
  res: any;
  log: (msg: any) => void;
  error: (msg: any) => void;
};

export default async ({ req, res, log, error }: Context) => {

  return res.send("test");
};
```

[...]

PR #6273, opened 2023-09-18
Fix: Large builds

Meldiron:
## What does this PR do?

Allows big sizes for builds. Without this limit eas 130MB set by Swoole. After which the container crashed.

This ensures build is done with chunk read/writes.

TODO:

- [x] Ensure git builds respect _APP_FUNCTIONS_SIZE_LIMIT

## Test Plan

- [x] Manual QA

## Related PRs and Issues

* https://github.com/appwrite/appwrite/issues/6245

## Checklist


Meldiron:
QA:

Large build:
CleanShot 2023-09-18 at 13 56 57@2x

lohanidamodar (approved):

vermakhushboo:
If executor limit is 130 MB, why is functions limit only 28 MB in this code?

app/workers/builds.php
```diff
@@ -251,17 +251,24 @@ protected function buildDeployment(GitHub $github, Document $project, Document $
...
+
+                if (substr($tmpDirectory, -1) !== '/') {
+                    $tmpDirectory .= '/';
+                }
+
+                $directorySize = $localDevice->getDirectorySize($tmpDirectory);
+                $functionsSizeLimit = (int) App::getEnv('_APP_FUNCTIONS_SIZE_LIMIT', '30000000');
+                if ($directorySize > $functionsSizeLimit) {
+                    throw new Exception('Repository directory size should be less than ' . number_format($functionsSizeLimit / 1048576, 2) . ' MBs.');
+                }
```
stnguyen90, line 264: I'm a little concerned about the inconsistency between VCS and manual deployment. In manual deployment, the size limit is on the compressed archive:

https://github.com/appwrite/appwrite/blob/ddb73eb4f60aa807fd0bfe8622da32a69738077a/app/controllers/api/functions.php#L1082-L1120

but here, we're looking at the uncompressed directory size.

Meldiron: With manual, compression is done by client. With VCS, compression is done by server.
I am worried we would be spending resources, allowing DOS attack if we moved it lower.

I understand inconsistency concerns, but I think only very very few devs should be hitting the limit on Cloud. And if hit on self-hosted, its so easy to increase that it doesn't matter too much if it's before or after compression size

stnguyen90: > I think only very very few devs should be hitting the limit on Cloud. And if hit on self-hosted

@Meldiron, I don't think it's comparison between cloud vs self-hosted. It's manual deployment vs VCS. Both can be done on cloud and self-hosted.

christyjacob4 (approved):

Issue #6279, opened 2023-09-19
🐛 Bug Report: Not able to Create Project with Project id not equal to (Contain at most 36 chars. Valid chars are a-z, 0-9, and hyphen. Can't start with a special char.)


btme0011:
### 👟 Reproduction steps

Try to Create a project with project id 123abc456DEf789ghij0klmnopqrs_tuv.wx

### 👍 Expected behavior

Project should be Created.

### 👎 Actual Behavior

GIving exception - ✗ Error Invalid `projectId` param: Project IDs must contain at most 36 chars. Valid chars are a-z, 0-9, and hyphen. Can't start with a special char.

### 🎲 Appwrite version

Version 1.4.x

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

nick2432:
I would like to work on this issue , please assign it to me

stnguyen90:
For this, we're going to want to update the [CustomId component](https://github.com/appwrite/console/blob/9b4bcb8140484669421685b4ba89fa1c4d331360/src/lib/components/customId.svelte) to expose a `isProject` flag. If the flag is true, it should use a new InputProjectId rather than [InputId](https://github.com/appwrite/console/blob/9b4bcb8140484669421685b4ba89fa1c4d331360/src/lib/elements/forms/inputId.svelte).

@nick2432, I've assigned the issue to you. Thanks for your interest! 🙏🏼

MuskanLudhiyani:
@stnguyen90 is this still open as I see the PR was closed?

stnguyen90:
@nick2432, are you still working on this?

nick2432:
@stnguyen90  yes

Mann-tech13:
Can I work on this issue, if it's still open? @stnguyen90

priii-25:
assign this to me

nick2432:
Can this issue be labeled as Hacktoberfest?

stnguyen90:
@nick2432, unfortunately, this is a bit small to qualify for the Appwrite Hacktoberfest swag (even with the additional onboarding bugfix).

nick2432:
@stnguyen90 Thanks for the feedback

malay5:
[...]

PR #6298, opened 2023-09-21
Fix 6279 wrong validation of projectid

nick2432:
## What does this PR do?

It correct the validation for a project id . Project id cannot include capital letters, hyphens, or periods

## Test Plan

i modified the regex to include all necessary characters

## Before
Before

## After
After

## Before
bf

## After
af

## Related PRs and Issues

- (Bug Report: Not able to Create Project with Project id not equal to (Contain at most 36 chars. Valid chars are a-z, 0-9, and hyphen. Can't start with a special char.) #6279)

## Checklist

### DIFF
--- a/src/Appwrite/Utopia/Database/Validator/ProjectId.php
+++ b/src/Appwrite/Utopia/Database/Validator/ProjectId.php
@@ -17,7 +17,7 @@ class ProjectId extends Validator
      */
     public function isValid($value): bool
     {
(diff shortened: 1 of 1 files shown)


stnguyen90:
@nick2432, great job on this PR! The thing is, the validation is fine server side as we do want to restrict the characters used for the project id because of https://github.com/appwrite/appwrite/issues/5670. The thing that was incorrect was the helper text in the console.

Also, when making changes, make sure to look out for there are unit or E2E tests that need to be added or updated.

PR #6317, opened 2023-09-24
Feat: Parse event body

Meldiron:
## What does this PR do?

Allows you to easily get data from event-triggered execution without having to parse it. Thanks to header, Open RUntimes will parse it.

## Test Plan

- [x] Add new test

## Related PRs and Issues

Reported during Office Hours on Discord

## Checklist

### DIFF
--- a/tests/e2e/Services/Functions/FunctionsCustomServerTest.php
+++ b/tests/e2e/Services/Functions/FunctionsCustomServerTest.php
@@ -1294,6 +1294,9 @@ class FunctionsCustomServerTest extends Scope
 
         $this->assertEquals(200, $deployment['headers']['status-code']);
 
+        // Wait a little for activation to finish
+        sleep(5);
+
         // Create user to trigger event
         $user = $this->client->call(Client::METHOD_POST, '/users', array_merge([
             'content-type' => 'application/json',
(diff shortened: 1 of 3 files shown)


tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -191,8 +191,8 @@ public function testList(array $data): array
             'runtime' => 'php-8.0',
             'entrypoint' => 'index.php',
             'events' => [
-                'users.*.create',
-                'users.*.delete',
+                'buckets.*.create',
+                'buckets.*.delete',
```
christyjacob4, line 195: why did we change the event type here?

christyjacob4 (approved):

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
Meldiron: Below I added actual event-triggered test. Those are just sample values for array attribute.

my test below was failing because of those having user creation listener.

Meldiron: I would need to rename those, or increase timeout to wait for more execution to finish

christyjacob4: Alright 👍

tannermeade:
This change creates a big problem (at least for dart cloud functions).

Change is here: https://github.com/open-runtimes/open-runtimes/blob/2a069cf9bec98e18a096633b5848cde12b4fe319/runtime

By assuming that the body HAS to be json parseable, all cloud functions will throw an error if the request is not json.

So everyone who has existing functions that accept string request bodies or anything not json parsable (or anyone who wants to in the future) can no longer do that.

All changes to make this work going forward require not only backend refactors, but frontend changes too.

tannermeade:
It would be really nice if cloud functions stop assuming things like this. The recent big change helped reduce these assumptions by a lot, but this is still an issue.

PR #6370, opened 2023-09-30
Patch: Sync executions timeout

Meldiron:
## What does this PR do?

Protects API worker getting stuck because of sync execution, or execution over Appwrite Router.

## Test Plan

x

## Related PRs and Issues

x

## Checklist


christyjacob4 (approved):

christyjacob4:
The test failure in the python runtime is due to an automatic upgrade on an internal library that makes it incompatible with few other library versions. The fix for this will be done on openruntimes and is not a blocker for this PR

Issue #6281, opened 2023-09-20
🐛 Bug Report: Multiple many to one relations in a database fail 


hinnerkweiler:
### 👟 Reproduction steps

Create two Database collections and set a `many to one` or relation for a field from collection `a` to `b`. Now repeat the step for a second field. The second relation can not be created.

Tried this for `many to one` and `many to many`.

Originally raised and discussed this issue on Discord and was asked to file a bug report: https://discord.com/channels/564160730845151244/1152869638431068220

in depth sample: I have a collection `Ports` holding a list of available ports to travel to and from. I also have a collection `Reservations` where the user stores a travel plan. I need to reference `Ports`-Document two times in each `Reservation`, first for departure and than for destination. this seems to not be possible with the current relation management.  

### 👍 Expected behavior

Allow multiple references from one collection to another.

### 👎 Actual Behavior

The Reference is only set up once, more fields fail at creation.

### 🎲 Appwrite version

Version 1.4.x

### 💻 Operating system

Linux

### 🧱 Your Environment

Tried in 1.4.2 and 1.4.3 plain install according docs `docker run...` on Ubuntu 22.04

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

btme0011:
if ($attribute->getAttribute('type') === self::VAR_RELATIONSHIP
                && \strtolower($attribute->getAttribute('options')['twoWayKey']) === \strtolower($twoWayKey)
                && $attribute->getAttribute('options')['relatedCollection'] === $relatedCollection->getId()
            ) {
                throw new DuplicateException('Related attribute already exists');
            }
            
            As we can see from the code that if these three thing matched we throw DuplicateException. So if type is relationship , collecitonId is same and relatedColleciton id is also same it will throw this DuplicateException

btme0011:
As we can see from the code that if these three thing matched we throw DuplicateException. So if type is relationship , collecitonId is same and relatedColleciton id is also same it will throw this DuplicateException

btme0011:
@stnguyen90 can i work on this issue?

stnguyen90:
@hinnerkweiler, thanks for raising this issue! 🙏🏼 Let me talk to the team about this.

stnguyen90:
[...]

PR #6438, opened 2023-10-02
Same two way key

fogelito:
Fix for relations attributes when twoWayKey variable is not passed in API,
We do not want the worker to throw the Error
Until now, it was checked only for twoWay = true

Added a validation to prevent the creation of 2 manyToMany relations between same collections

Fixes: https://github.com/appwrite/appwrite/issues/6281

### DIFF
--- a/app/controllers/api/databases.php
+++ b/app/controllers/api/databases.php
@@ -1642,7 +1642,7 @@ App::post('/v1/databases/:databaseId/collections/:collectionId/attributes/relati
         /** @var Document[] $attributes */
         foreach ($attributes as $attribute) {
             if (\strtolower($attribute->getId()) === \strtolower($key)) {
-                throw new Exception(Exception::DOCUMENT_ALREADY_EXISTS);
+                throw new Exception(Exception::ATTRIBUTE_ALREADY_EXISTS);
             }
 
             if (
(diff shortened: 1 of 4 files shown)


app/controllers/api/databases.php
```diff
@@ -1615,8 +1615,49 @@ function updateAttribute(
...
+
+            if (
+                $attribute->getAttribute('type') === Database::VAR_RELATIONSHIP &&
+                \strtolower($attribute->getAttribute('options')['twoWayKey']) === \strtolower($twoWayKey) &&
+                $attribute->getAttribute('options')['relatedCollection'] === $relatedCollection->getId()
+            ) {
+                if ($twoWayKeyNull) {
+                    $twoWayKey .= '_' . uniqid();
+                } else {
+                    throw new Exception(Exception::DOCUMENT_ALREADY_EXISTS);
```
abnegate, line 1656: Same here for the exception type

app/controllers/api/databases.php
```diff
@@ -1615,8 +1615,49 @@ function updateAttribute(
...
+
+        if ($relatedCollection->isEmpty()) {
+            throw new Exception(Exception::COLLECTION_NOT_FOUND);
+        }
+
+        $attributes = $collection->getAttribute('attributes', []);
+        /** @var Document[] $attributes */
+        foreach ($attributes as $attribute) {
+            if (\strtolower($attribute->getId()) === \strtolower($key)) {
+                throw new Exception(Exception::DOCUMENT_ALREADY_EXISTS);
```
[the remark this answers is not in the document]
fogelito: Yep done

app/controllers/api/databases.php
fogelito: Done
I have removed the change of the attribute naming, as we spoke, currently will just throw an Error.

app/controllers/api/databases.php
```diff
@@ -1615,8 +1615,55 @@ function updateAttribute(
...
+                throw new Exception(Exception::ATTRIBUTE_ALREADY_EXISTS);
+            }
+
+            if (
+                $type === Database::RELATION_MANY_TO_MANY &&
+                $attribute->getAttribute('type') === Database::VAR_RELATIONSHIP &&
+                $attribute->getAttribute('options')['relationType'] === Database::RELATION_MANY_TO_MANY &&
+                $attribute->getAttribute('options')['relatedCollection'] === $relatedCollection->getId()
+            ) {
+                throw new Exception('Cannot create more than one ManyToMany relation.');
```
fogelito, line 1663: @abnegate what kind of Exception should we throw?

app/controllers/api/databases.php
```diff
@@ -1615,8 +1615,55 @@ function updateAttribute(
         Event $events
     ) {
         $key ??= $relatedCollectionId;
+        $twoWayKeyNull = is_null($twoWayKey);
         $twoWayKey ??= $collectionId;
 
+        $databaseDocument = Authorization::skip(fn() => $dbForProject->getDocument('databases', $databaseId));
```
abnegate, line 1621: Can we do this in the `createAttribute` function, I think we have a lot of this code there already

app/controllers/api/databases.php
abnegate: Maybe still `ATTRIBUTE_ALREADY_EXISTS` but with this message `Creating more than one "manyToMany" relationship on the same collection is currently not permitted.`

phpunit.xml
```diff
@@ -6,7 +6,7 @@
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
          processIsolation="false"
-         stopOnFailure="false"
+         stopOnFailure="true"
```
abnegate, line 9: Let's revert this

fogelito:
PR moved 
https://github.com/appwrite/appwrite/pull/7153

PR #6487, opened 2023-10-04
Updated error message to not user magic strings

gewenyu99:
## What does this PR do?

Fix use of outdated magic string.

image

## Test Plan
manual

### DIFF
--- a/app/config/errors.php
+++ b/app/config/errors.php
@@ -355,7 +355,7 @@ return [
     ],
     Exception::STORAGE_BUCKET_ALREADY_EXISTS => [
         'name' => Exception::STORAGE_BUCKET_ALREADY_EXISTS,
-        'description' => 'A storage bucket with the requested ID already exists. Try again with a different ID or use "unique()" to generate a unique ID.',
+        'description' => 'A storage bucket with the requested ID already exists. Try again with a different ID or use `ID.unique()` to generate a unique ID.',
         'code' => 409,
     ],
     Exception::STORAGE_BUCKET_NOT_FOUND => [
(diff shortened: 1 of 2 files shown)


gewenyu99:
Yeah @stnguyen90

app/config/errors.php
```diff
@@ -549,7 +549,7 @@
     ],
     Exception::ATTRIBUTE_ALREADY_EXISTS => [
         'name' => Exception::ATTRIBUTE_ALREADY_EXISTS,
-        'description' => 'Attribute with the requested ID already exists. Try again with a different ID or use "unique()" to generate a unique ID.',
+        'description' => 'Attribute with the requested ID already exists. Attribute IDs must be unique, try again with a different ID.',
```
stnguyen90, line 552: We call them keys rather than IDs so we should probably be consistent with that here.

app/config/errors.php
```diff
@@ -581,7 +581,7 @@
     ],
     Exception::INDEX_ALREADY_EXISTS => [
         'name' => Exception::INDEX_ALREADY_EXISTS,
-        'description' => 'Index with the requested ID already exists. Try again with a different ID or use "unique()" to generate a unique ID.',
+        'description' => 'Index with the requested ID already exists. Try again with a different ID or use ID.unique() to generate a unique ID.',
```
stnguyen90, line 584: We also call it keys for indexes...Also, I don't think you can use unique for index keys.

stnguyen90 (approved):
LGTM. I just merged main into here to hopefully fix the tests.

stnguyen90 (changes requested):
I lied 😬 Would you please remove the package-lock.json change?

gewenyu99:
I was so confused trying to revert my changes haha

christyjacob4 (approved):

gewenyu99:
> changes are fine, just curious if this works for all SDK's from the `ID.unique()`.

I think it's consistent here!

stnguyen90 (approved):

PR #6888, opened 2023-10-12
Feat db per worker

abnegate:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


docker-compose.yml
```diff
@@ -372,6 +372,7 @@ services:
       - mariadb
     environment:
       - _APP_ENV
+      - _APP_CONNECTIONS_DB_QUEUE_01
```
eldadfux, line 375: Do we need the number postfix? Would be more elegant if we can skip on the default self-hosted setup.

app/worker.php
```diff
@@ -216,17 +211,25 @@
...
+$workerIndex = $args[1] ?? '';
+
+if (!empty($workerNum)) {
+    $workerName .= '_' . $workerIndex;
 }
 
 try {
     $platform->init(Service::TYPE_WORKER, [
-        'workersNum' => strtolower($workerName) === 'databases'? 1 :swoole_cpu_num() * intval(App::getEnv('_APP_WORKER_PER_CORE', 6)),
+        'workersNum' => str_starts_with(strtolower($workerName), 'databases') ? 1 : swoole_cpu_num() * intval(App::getEnv('_APP_WORKER_PER_CORE', 6)),
```
eldadfux, line 229: This is not a great solution. We should always avoid specific conditions that complexity and make the story of the container much harder to explain. We can instead use a generic env var for deciding the number of workers to set.

For easier maintenance I would also add a doc block with a list of all specific env vars allowed on this file/container.

app/worker.php
```diff
@@ -216,17 +211,25 @@
...
         'connection' => $pools->get('queue')->pop()->getResource(),
         'workerName' => strtolower($workerName) ?? null,
+        'queueName' => App::getEnv('_APP_CONNECTIONS_DB_QUEUE', 'database_db_main')
```
eldadfux, line 232: Same as above. Let's not use any DB specific default but rather be very generic by adding custom capabilities through a list of env vars that acts as kind of a configuration API.

docker-compose.yml
```diff
@@ -356,10 +356,10 @@ services:
       - _APP_EXECUTOR_SECRET
       - _APP_EXECUTOR_HOST
 
-  appwrite-worker-databases:
-    entrypoint: worker-databases
+  appwrite-worker-databases-01:
+    entrypoint: worker-databases 01
     <<: *x-logging
-    container_name: appwrite-worker-databases
+    container_name: appwrite-worker-databases-01
```
eldadfux, line 362: Do we need the number postfix? Would be more elegant if we can skip on the default self-hosted setup. When needing to scale, like in Cloud, devs can have a dedicated and custom compose file.

eldadfux (changes requested):

app/worker.php
```diff
@@ -216,17 +211,31 @@
...
+if (!empty($workerNum)) {
+    $workerName .= '_' . $workerIndex;
 }
 
 try {
+    /**
+     * Any worker can be configured with the following env vars:
+     * - _APP_WORKERS_NUM           The total number of worker processes
+     * - _APP_WORKER_PER_CORE       The number of worker processes per core (ignored if _APP_WORKERS_NUM is set)
+     * - _APP_CONNECTIONS_DB_QUEUE  The name of the queue to read for database events
```
eldadfux, line 232:
```suggestion
     * - _APP_QUEUE_NAME  The name of the queue to read for database events
```

It would be better to not have special conditions just for the database worker. If empty we can try and generate the queue name as it is today or not change it and overwrite it only when a specific value has been given.

docker-compose.yml
```diff
@@ -356,7 +356,7 @@ services:
       - _APP_EXECUTOR_SECRET
       - _APP_EXECUTOR_HOST
 
-  appwrite-worker-databases:
+  appwrite-worker-databases-:
```
eldadfux, line 359: We need to drop the dash at the end.

eldadfux (changes requested):

.env
```diff
@@ -1,6 +1,7 @@
 _APP_ENV=development
 _APP_LOCALE=en
 _APP_WORKER_PER_CORE=6
+_APP_QUEUE_NAME=database_db_main
```
eldadfux, line 4: Shouldn't this be empty by default?

docker-compose.yml
```diff
@@ -372,6 +372,8 @@ services:
       - mariadb
     environment:
       - _APP_ENV
+      - _APP_WORKERS_NUM=1
```
eldadfux, line 375: Should we also update compose.phtml?

eldadfux (changes requested):

.env
abnegate: This is what will be set as the default `database` for a `project` in self-hosted, so that's the queue name that the database jobs will get pushed to as well

docker-compose.yml
abnegate: Yep will update :+1:

.env
eldadfux: Wouldn't this affect other workers as well?

abnegate: No, it will only affect containers that explicitly include it in the containers environment in the compose file

abnegate: We could hardcore it in the compose file for the database worker instead?

eldadfux: OK, yeh this will make more sense to me

PR #6890, opened 2023-10-12
1.5.x

abnegate:
## What does this PR do?

1.5.x

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


src/Appwrite/Utopia/Response.php
```diff
@@ -238,8 +255,8 @@ class Response extends SwooleResponse
     public const MODEL_WEBHOOK_LIST = 'webhookList';
     public const MODEL_KEY = 'key';
     public const MODEL_KEY_LIST = 'keyList';
-    public const MODEL_PROVIDER = 'provider';
-    public const MODEL_PROVIDER_LIST = 'providerList';
+    public const MODEL_AUTH_PROVIDER = 'authProvider';
```
eldadfux, line 258: How hard will it be for us to use same provider object for both auth and messaging in the future?

src/Appwrite/Utopia/Response/Model/Message.php
```diff
@@ -0,0 +1,104 @@
...
+            ])
+            ->addRule('deliveryErrors', [
+                'type' => self::TYPE_STRING,
+                'description' => 'Delivery errors if any.',
+                'required' => false,
+                'default' => '',
+                'array' => true,
+                'example' => ['Failed to send message to target 5e5ea5c16897e: Credentials not valid.'],
+            ])
+            ->addRule('deliveredTo', [
```
eldadfux, line 55: A bit misleading. My initial thought was that this will be an array of all targets.

src/Appwrite/Utopia/Response/Model/Message.php
```diff
@@ -0,0 +1,104 @@
...
+                'type' => self::TYPE_INTEGER,
+                'description' => 'Number of recipients the message was delivered to.',
+                'default' => 0,
+                'example' => 1,
+            ])
+            ->addRule('data', [
```
eldadfux, line 61: should this be data / payload / body? how does this compare to other similar fields we have?

src/Appwrite/Utopia/Response/Model/Project.php
```diff
@@ -138,9 +138,9 @@ public function __construct()
                 'default' => false,
                 'example' => true,
             ])
-            ->addRule('providers', [
-                'type' => Response::MODEL_PROVIDER,
-                'description' => 'List of Providers.',
+            ->addRule('authProviders', [
```
eldadfux, line 141: This is a breaking change for the console. Let's make sure we apply the relevant adjustments.

src/Appwrite/Utopia/Response/Model/Message.php
abnegate: How about `deliveryCount`? This will be how many targets the message has currently been delivered to, updated as the worker processes the target batches

src/Appwrite/Utopia/Response/Model/Message.php
abnegate: Actually I think this should be removed so we have the same as documents where the the attributes are added at the top level :thinking:

.env
```diff
@@ -98,4 +98,15 @@ _APP_VCS_GITHUB_CLIENT_SECRET=
 _APP_VCS_GITHUB_WEBHOOK_SECRET=
 _APP_MIGRATIONS_FIREBASE_CLIENT_ID=
 _APP_MIGRATIONS_FIREBASE_CLIENT_SECRET=
-_APP_ASSISTANT_OPENAI_API_KEY=
\ No newline at end of file
+_APP_ASSISTANT_OPENAI_API_KEY=
+_APP_MESSAGE_SMS_PROVIDER_MSG91_SENDER_ID=
```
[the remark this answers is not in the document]
Meldiron: I prefer DSN structure too

src/Appwrite/Utopia/Database/Validator/Queries/Messages.php
```diff
@@ -0,0 +1,27 @@
...
+    public const ALLOWED_ATTRIBUTES = [
+        'to',
+        'providerId',
+        'deliveredAt',
+        'deliveredTo',
+        'deliveryErrors',
+        'status',
+        'description',
+        'data',
+        'search'
```
stnguyen90, line 16: we typically don't allow querying on search. Instead, it's a separate param on the endpoint.

src/Appwrite/Utopia/Database/Validator/Queries/Messages.php
```diff
@@ -0,0 +1,27 @@
...
+namespace Appwrite\Utopia\Database\Validator\Queries;
+
+class Messages extends Base
+{
+    public const ALLOWED_ATTRIBUTES = [
+        'to',
+        'providerId',
+        'deliveredAt',
+        'deliveredTo',
+        'deliveryErrors',
```
stnguyen90, line 12: deliveryErrors is an array, but we don't support queries on arrays yet.

src/Appwrite/Utopia/Database/Validator/Queries/Messages.php
```diff
@@ -0,0 +1,27 @@
...
+        'status',
+        'description',
+        'data',
```
stnguyen90, line 15: data is JSON, but we don't really support queries on JSON.

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2138 @@
...
+            ->setType(DELETE_TYPE_SUBSCRIBERS)
+            ->setDocument($topic);
+
+        $response
+            ->setStatusCode(Response::STATUS_CODE_NOCONTENT)
+            ->noContent();
+    });
+
+App::post('/v1/messaging/topics/:topicId/subscribers')
+    ->desc('Adds a Subscriber to a Topic.')
```
stnguyen90, line 1391: Sentence case?

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2138 @@
...
+    ->action(function (string $providerId, string $name, bool $default, bool $enabled, bool $isEuRegion, string $from, string $apiKey, string $domain, Database $dbForProject, Response $response) {
+        $providerId = $providerId == 'unique()' ? ID::unique() : $providerId;
+
+        $provider = new Document([
+            '$id' => $providerId,
+            'name' => $name,
+            'provider' => 'mailgun',
+            'type' => 'email',
+            'default' => $default,
+            'enabled' => $enabled,
```
stnguyen90, line 61: should enabled be false until everything is populated?

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2138 @@
...
+use Utopia\Validator\ArrayList;
+use Utopia\Validator\Boolean;
+use Utopia\Validator\JSON;
+use Utopia\Validator\Text;
+use Utopia\Validator\WhiteList;
+
+App::post('/v1/messaging/providers/mailgun')
+    ->desc('Create Mailgun Provider')
+    ->groups(['api', 'messaging'])
+    ->label('audits.event', 'providers.create')
```
stnguyen90, line 32: missing events label?

stnguyen90, line 32: audits event should be singular:

```suggestion
    ->label('audits.event', 'provider.create')
```

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2138 @@
...
+    ->label('sdk.auth', [APP_AUTH_TYPE_ADMIN, APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'messaging')
+    ->label('sdk.method', 'createMsg91Provider')
+    ->label('sdk.description', '/docs/references/messaging/create-msg91-provider.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_PROVIDER)
+    ->param('providerId', '', new CustomId(), 'Provider ID. Choose a custom ID or generate a random ID with `ID.unique()`. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.')
+    ->param('name', '', new Text(128), 'Provider name.')
+    ->param('default', false, new Boolean(), 'Set as default provider.', true)
```
stnguyen90, line 166: what if i set multiple providers to be default? 👀

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2138 @@
...
+            throw new Exception(Exception::PROVIDER_ALREADY_EXISTS);
+        }
+
+        $response
+            ->setStatusCode(Response::STATUS_CODE_CREATED)
+            ->dynamic($provider, Response::MODEL_PROVIDER);
+    });
+
+App::post('/v1/messaging/providers/textmagic')
+    ->desc('Create Textmagic Provider')
```
stnguyen90, line 272:
```suggestion
    ->desc('Create TextMagic provider')
```

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2138 @@
...
+    ->groups(['api', 'messaging'])
+    ->label('audits.event', 'providers.create')
+    ->label('audits.resource', 'providers/{response.$id}')
+    ->label('scope', 'providers.write')
+    ->label('sdk.auth', [APP_AUTH_TYPE_ADMIN, APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'messaging')
+    ->label('sdk.method', 'createTextmagicProvider')
```
stnguyen90, line 279:
```suggestion
    ->label('sdk.method', 'createTextMagicProvider')
```

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2138 @@
...
+    ->label('sdk.description', '/docs/references/messaging/create-textmagic-provider.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_PROVIDER)
+    ->param('providerId', '', new CustomId(), 'Provider ID. Choose a custom ID or generate a random ID with `ID.unique()`. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.')
+    ->param('name', '', new Text(128), 'Provider name.')
+    ->param('default', false, new Boolean(), 'Set as default provider.', true)
+    ->param('enabled', true, new Boolean(), 'Set as enabled.', true)
+    ->param('username', '', new Text(0), 'Textmagic username.')
```
stnguyen90, line 288:
```suggestion
    ->param('username', '', new Text(0), 'TextMagic username.')
```

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2138 @@
...
+    ->param('apiKey', '', new Text(0), 'Textmagic apiKey.')
```
stnguyen90, line 289:
```suggestion
    ->param('apiKey', '', new Text(0), 'TextMagic API key.')
```

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2138 @@
...
+    ->label('sdk.auth', [APP_AUTH_TYPE_ADMIN, APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'messaging')
+    ->label('sdk.method', 'createTextmagicProvider')
+    ->label('sdk.description', '/docs/references/messaging/create-textmagic-provider.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_PROVIDER)
+    ->param('providerId', '', new CustomId(), 'Provider ID. Choose a custom ID or generate a random ID with `ID.unique()`. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.')
+    ->param('name', '', new Text(128), 'Provider name.')
+    ->param('default', false, new Boolean(), 'Set as default provider.', true)
```
stnguyen90, line 286: this should only be the default SMS provider, right? not default amongst all the providers?

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2138 @@
...
+            throw new Exception(Exception::PROVIDER_ALREADY_EXISTS);
+        }
+
+        $response
+            ->setStatusCode(Response::STATUS_CODE_CREATED)
+            ->dynamic($provider, Response::MODEL_PROVIDER);
+    });
+
+App::post('/v1/messaging/providers/twilio')
+    ->desc('Create Twilio Provider')
```
stnguyen90, line 330:
```suggestion
    ->desc('Create Twilio provider')
```

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2138 @@
...
+    ->label('sdk.method', 'createFcmProvider')
+    ->label('sdk.description', '/docs/references/messaging/create-fcm-provider.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_PROVIDER)
+    ->param('providerId', '', new CustomId(), 'Provider ID. Choose a custom ID or generate a random ID with `ID.unique()`. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.')
+    ->param('name', '', new Text(128), 'Provider name.')
+    ->param('default', false, new Boolean(), 'Set as default provider.', true)
+    ->param('enabled', true, new Boolean(), 'Set as enabled.', true)
+    ->param('serverKey', '', new Text(0), 'FCM Server Key.')
```
stnguyen90, line 462:
```suggestion
    ->param('serverKey', '', new Text(0), 'FCM server key.')
```

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2138 @@
...
+            $cursorDocument = Authorization::skip(fn () => $dbForProject->getDocument('providers', $providerId));
+
+            if ($cursorDocument->isEmpty()) {
+                throw new Exception(Exception::GENERAL_CURSOR_NOT_FOUND, "Provider '{$providerId}' for the 'cursor' value not found.");
+            }
+
+            $cursor->setValue($cursorDocument);
+        }
+
+        $filterQueries = Query::groupByType($queries)['filters'];
```
stnguyen90, line 597: we don't need this anymore since the count() method has been updated to use only filters.

app/config/errors.php
```diff
@@ -732,4 +742,62 @@
...
+        'code' => 404,
+    ],
+    Exception::PROVIDER_ALREADY_EXISTS => [
+        'name' => Exception::PROVIDER_ALREADY_EXISTS,
+        'description' => 'Provider with the requested ID already exists.',
+        'code' => 409,
+    ],
+    Exception::PROVIDER_INCORRECT_TYPE => [
+        'name' => Exception::PROVIDER_INCORRECT_TYPE,
+        'description' => 'Provider with the requested ID is of incorrect type: ',
```
stnguyen90, line 759: Should this be a colon at the end? Also,  @gewenyu99 would probably review the copy for these errors.

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2138 @@
...
+    ->label('sdk.method', 'updateMailgunProvider')
+    ->label('sdk.description', '/docs/references/messaging/update-mailgun-provider.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_PROVIDER)
+    ->param('providerId', '', new UID(), 'Provider ID.')
+    ->param('name', '', new Text(128), 'Provider name.', true)
+    ->param('enabled', null, new Boolean(), 'Set as enabled.', true)
+    ->param('isEuRegion', null, new Boolean(), 'Set as eu region.', true)
+    ->param('from', '', new Text(256), 'Sender Email Address.', true)
```
stnguyen90, line 645: Should this use the email validator?

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2138 @@
...
+        $provider->setAttribute('credentials', $credentials);
+
+        $provider = $dbForProject->updateDocument('providers', $provider->getId(), $provider);
+
+        $response
+            ->dynamic($provider, Response::MODEL_PROVIDER);
+    });
+
+App::patch('/v1/messaging/providers/textmagic/:providerId')
+    ->desc('Update Textmagic Provider')
```
stnguyen90, line 870:
```suggestion
    ->desc('Update TextMagic provider')
```

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2138 @@
...
+    ->groups(['api', 'messaging'])
+    ->label('audits.event', 'providers.update')
+    ->label('audits.resource', 'providers/{response.$id}')
+    ->label('scope', 'providers.write')
+    ->label('sdk.auth', [APP_AUTH_TYPE_ADMIN, APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'messaging')
+    ->label('sdk.method', 'updateTextmagicProvider')
```
stnguyen90, line 877:
```suggestion
    ->label('sdk.method', 'updateTextMagicProvider')
```

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2138 @@
...
+            throw new Exception(Exception::SUBSCRIBER_ALREADY_EXISTS);
+        }
+
+        $response
+            ->setStatusCode(Response::STATUS_CODE_CREATED)
+            ->dynamic($subscriber, Response::MODEL_SUBSCRIBER);
+    });
+
+App::get('/v1/messaging/topics/:topicId/subscribers')
+    ->desc('List topic\'s subscribers.')
```
stnguyen90, line 1448: will a developer need to be able to search for subscribers across multiple topics?

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2138 @@
...
+    ->label('sdk.description', '/docs/references/messaging/create-email.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_MESSAGE)
+    ->param('messageId', '', new CustomId(), 'Message ID. Choose a custom ID or generate a random ID with `ID.unique()`. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.')
+    ->param('providerId', '', new UID(), 'Email Provider ID.')
+    ->param('to', [], new ArrayList(new Text(Database::LENGTH_KEY)), 'List of Topic IDs or List of User IDs or List of Target IDs.')
+    ->param('subject', '', new Text(998), 'Email Subject.')
+    ->param('content', '', new Text(64230), 'Email Content.')
+    ->param('description', '', new Text(256), 'Description for Message.', true)
```
stnguyen90, line 1584: sentence case for description.

stnguyen90 (changes requested):

app/init.php
```diff
@@ -526,6 +522,38 @@ function (mixed $value) {
...
+Database::addFilter(
+    'subQueryTargets',
+    function (mixed $value) {
+        return null;
+    },
+    function (mixed $value, Document $document, Database $database) {
+        return Authorization::skip(fn() => $database
+            ->find('targets', [
+                Query::equal('userInternalId', [$document->getInternalId()]),
+            ]));
```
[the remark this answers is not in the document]
abnegate: Yep let's add this @fanatic75

app/config/collections.php
```diff
@@ -5,7 +5,7 @@
 use Utopia\Database\Database;
 use Utopia\Database\Helpers\ID;
 
-$providers = Config::getParam('providers', []);
+$providers = Config::getParam('authProviders', []);
```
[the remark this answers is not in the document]
abnegate: Makes sense since the file is specific to oauth, let's update @fanatic75

app/controllers/api/users.php
```diff
@@ -646,6 +735,35 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
         ]), Response::MODEL_LOG_LIST);
     });
 
+App::get('/v1/users/:userId/targets')
```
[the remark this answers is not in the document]
abnegate: I can't think of one; a target should belong to a user. User targets are following the same structure as user sessions

app/controllers/api/users.php
```diff
@@ -646,6 +735,35 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+
+        if ($user->isEmpty()) {
+            throw new Exception(Exception::USER_NOT_FOUND);
+        }
+
+        $targets = $user->getAttribute('targets', []);
+        $response->dynamic(new Document([
+            'targets' => $targets,
+            'total' => \count($targets),
+        ]), Response::MODEL_TARGET_LIST);
```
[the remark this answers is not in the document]
abnegate: We're following the same as sessions here, which is why we don't have querying on this route

app/controllers/api/users.php
```diff
@@ -1080,6 +1198,51 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
         $response->dynamic(new Document($prefs), Response::MODEL_PREFERENCES);
     });
 
+App::patch('/v1/users/:userId/targets/:targetId/identifier')
+    ->desc('Update user target\'s identifier')
+    ->groups(['api', 'users'])
+    ->label('audits.event', 'users.targets.update')
```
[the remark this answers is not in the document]
abnegate: @fanatic75 Let's fix the `audit.event` label here, as well as make sure we have the `event` label set on all the new routes

src/Appwrite/Platform/Workers/Deletes.php
```diff
@@ -149,6 +150,9 @@ public function action(Message $message, Database $dbForConsole, callable $getPr
             case DELETE_TYPE_SCHEDULES:
                 $this->deleteSchedules($dbForConsole, $getProjectDB, $datetime);
                 break;
+            case DELETE_TYPE_SUBSCRIBERS:
+                $this->deleteSubscribers($project, $getProjectDB, $document);
+                break;
```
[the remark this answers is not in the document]
abnegate: This actually is for delete topic but should be renamed. Deleting provider makes sense to add here too, making sure we delete all topics, subscribers, and messages related to the provider @fanatic75

app/init.php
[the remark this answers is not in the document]
fanatic75: @abnegate this might cause problem because if there are targets more than specified limit, they will not get messages

app/controllers/api/messaging.php
abnegate: I don't think it will make any difference here, nothing will be persisted until the `createDocument` call anyway

app/controllers/api/messaging.php
abnegate: @fanatic75 I don't think we actually want this parameter in the API 🤔 just in the DB

app/controllers/api/messaging.php
abnegate: Correct, when we look for a default, we check by the `default` value as well as the `type`

app/config/errors.php
abnegate: Yep should be, the message here should be a static fallback if it's not set in the constructor. All descriptions/docs will have their own review for sure :+1:

app/controllers/api/messaging.php
abnegate: I can't think of a use case, can you?

app/init.php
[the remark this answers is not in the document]
abnegate: `APP_LIMIT_SUBQUERY` has a default value of 1000, I think that's a reasonable limit in this context. I don't expect many users to have more than 1000 targets

.env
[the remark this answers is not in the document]
christyjacob4: Yea DSN would be worth considering here. 

We could do
* sms://username:password@host/path?paramA=valueA
* email://username:password@host/path?paramA=valueA

and so on .. 

this can then be accessed with 
```php
$dsn->getScheme() # sms
$dsn->getHost() # host
$dsn->getUser() #username
$dsn->getPassword() #password
$dsn->getPath() # path
$dsn->getParam('paramA') # valueA
```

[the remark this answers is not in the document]
fanatic75: These env variables are only for e2e tests of messaging endpoints. Since users configure the providers, we store the credentials directly in our db data.

.env
```diff
@@ -98,4 +98,15 @@ _APP_VCS_GITHUB_CLIENT_SECRET=
...
+_APP_MESSAGE_SMS_PROVIDER_MSG91_AUTH_KEY=
```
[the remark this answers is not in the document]
fanatic75: Since the send message endpoints are generic for each provider type (SMS, email and APNS), only one provider type's env variables are needed for test cases.

.env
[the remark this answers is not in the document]
fanatic75: Just a note here: These env variables are only for e2e tests of messaging endpoints. Since users configure the providers, we store the credentials directly in our db data.

app/controllers/api/messaging.php
fanatic75: done this

app/controllers/api/messaging.php
fanatic75: done this

src/Appwrite/Utopia/Response/Model/Message.php
stnguyen90: How about `deliveredTotal` considering we use total in other places?

app/controllers/api/users.php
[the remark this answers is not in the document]
stnguyen90: What if in Firebase, I see something has been sent to target `asdf`? Now, I want to figure out which user that was in Appwrite?

app/controllers/api/users.php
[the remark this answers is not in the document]
stnguyen90: Hmm it seems quite limiting to do it that way...

app/controllers/api/users.php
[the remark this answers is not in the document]
stnguyen90: @fanatic75 the audits.event label probably needs to be updated for other endpoints too.

app/controllers/api/messaging.php
stnguyen90: But there's nothing between here and the `createDocument()` call to validate or force `$enabled` to be false 🧐

app/controllers/api/messaging.php
stnguyen90: How would one update the default provider then? 🧐

app/controllers/api/messaging.php
stnguyen90: So if this stays in the API, perhaps the description can be updated to clarify that.

app/controllers/api/messaging.php
stnguyen90: I can imagine a developer needing to determine what topics a user is part of.

In general, I'm against nesting resources because it makes it harder to query for data.

app/controllers/api/users.php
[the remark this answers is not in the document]
abnegate: You could look up the target by `identifier = asdf` and look at it's `userId`

app/controllers/api/users.php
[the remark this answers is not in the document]
abnegate: Any suggestions? Open to any improvements, I agree it is quite limiting for both resources

app/controllers/api/messaging.php
abnegate: Not sure I understand, why would we need to do that?

.env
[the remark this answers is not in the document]
Meldiron: I like idea of having some kind of `TESTS` prefix. Still, it would be cool to use DSN I think, allowing us to have it all in 1 variable.

app/controllers/api/users.php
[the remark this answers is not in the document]
stnguyen90: How would you look up the target by identifier?

app/controllers/api/users.php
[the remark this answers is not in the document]
stnguyen90: Have an endpoint like `/v1/users/targets` that can support all the usual queries including userId, limit, offset, cursor, etc.

app/controllers/api/messaging.php
stnguyen90: To prevent people from enabling a provider that doesn't have everything filled out yet.

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2145 @@
...
+        }
+
+        $response
+            ->dynamic(new Document([
+                'subscribers' => $dbForProject->find('subscribers', $queries),
+                'total' => $dbForProject->count('subscribers', $queries, APP_LIMIT_COUNT),
+            ]), Response::MODEL_SUBSCRIBER_LIST);
+    });
+
+App::get('/v1/messaging/topics/:topicId/subscriber/:subscriberId')
```
stnguyen90, line 1539: subscribers should be plural:

```suggestion
App::get('/v1/messaging/topics/:topicId/subscribers/:subscriberId')
```

app/controllers/api/users.php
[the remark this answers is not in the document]
abnegate: I think these should still be a nested resource, but we can allow querying. @fanatic75 Let's add the querying support here by just pulling the data from the `targets` collection directly, filtering on the `userId`

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2145 @@
...
+    ->label('audits.resource', 'message/{response.$id}')
+    ->label('scope', 'messages.write')
+    ->label('sdk.auth', [APP_AUTH_TYPE_ADMIN, APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'messaging')
+    ->label('sdk.method', 'createEmail')
+    ->label('sdk.description', '/docs/references/messaging/create-email.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_MESSAGE)
+    ->param('messageId', '', new CustomId(), 'Message ID. Choose a custom ID or generate a random ID with `ID.unique()`. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.')
```
stnguyen90, line 1622: shouldn't these create email/sms/push also take a providerId as input?

stnguyen90: Jake said:

> Not anymore, we changed this so that topics can support multiple providers. One API call may actually send 3 messages to different providers, so the message needs to be agnostic to provider. In the worker the targets are grouped by provider and a message is sent to each

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2145 @@
...
+    ->param('to', [], new ArrayList(new Text(Database::LENGTH_KEY), 1), 'List of Topic IDs or List of User IDs or List of Target IDs.')
+    ->param('subject', '', new Text(998), 'Email Subject.')
+    ->param('content', '', new Text(64230), 'Email Content.')
+    ->param('description', '', new Text(256), 'Description for message.', true)
+    ->param('status', 'processing', new WhiteList(['draft', 'processing']), 'Message Status. Value must be either draft or processing.', true)
+    ->param('html', false, new Boolean(), 'Is content of type HTML', true)
+    ->param('deliveryTime', null, new DatetimeValidator(requireDateInFuture: true), 'Delivery time for message in ISO 8601 format. DateTime value must be in future.', true)
```
eldadfux, line 1629: Not sure about the naming convention. We use somethingAt in many places, what do we think about that?

eldadfux, line 1629: Can we make ISO 8601 linkable like in all other place? (and fix if it's missing somewhere)

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2145 @@
...
+    ->param('title', '', new Text(256), 'Title for push notification.')
+    ->param('body', '', new Text(64230), 'Body for push notification.')
+    ->param('description', '', new Text(256), 'Description for Message.', true)
+    ->param('data', null, new JSON(), 'Additional Data for push notification.', true)
+    ->param('action', '', new Text(256), 'Action for push notification.', true)
+    ->param('icon', '', new Text(256), 'Icon for push notification. Available only for Android and Web Platform.', true)
+    ->param('sound', '', new Text(256), 'Sound for push notification. Available only for Android and IOS Platform.', true)
+    ->param('color', '', new Text(256), 'Color for push notification. Available only for Android Platform.', true)
+    ->param('tag', '', new Text(256), 'Tag for push notification. Available only for Android Platform.', true)
+    ->param('badge', '', new Text(256), 'Badge for push notification. Available only for IOS Platform.', true)
```
eldadfux, line 1735: Is it correct from our side to state `IOS Platform` or should we be more generic when it come to Apple OS family?

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2145 @@
...
+    ->param('status', 'processing', new WhiteList(['draft', 'processing']), 'Message Status. Value must be either draft or processing.', true)
```
eldadfux, line 1736: Reminder. Let's ensure this works well with SDK enums.

eldadfux (changes requested):

app/controllers/api/messaging.php
fanatic75: Picked these details from Firebase cloud messaging provider documentation, it mentions IOS headers so I have kept the same.

https://firebase.google.com/docs/cloud-messaging/http-server-ref#notification-payload-support

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2145 @@
...
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_PROVIDER)
+    ->param('providerId', '', new CustomId(), 'Provider ID. Choose a custom ID or generate a random ID with `ID.unique()`. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.')
+    ->param('name', '', new Text(128), 'Provider name.')
+    ->param('from', '', new Email(), 'Sender email address.')
+    ->param('apiKey', '', new Text(0), 'Mailgun API Key.')
+    ->param('domain', '', new Text(0), 'Mailgun Domain.')
+    ->param('isEuRegion', false, new Boolean(), 'Set as EU region.')
+    ->param('enabled', true, new Boolean(), 'Set as enabled.', true)
```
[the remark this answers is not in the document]
fanatic75: @eldadfux  ,We have kept all the optional params below the mandatory params, I can however move enabled at the top of optional params, how does that sound?

app/controllers/api/messaging.php
fanatic75: I couldn't see it linkable in any controller. Does the wiki link work? 
https://en.wikipedia.org/wiki/ISO_8601

fanatic75: @eldadfux , we have two attributes, deliveredAt (signifies when the message was delivered to the user) and deliveryTime( time when message is scheduled at), we can rename this to scheduleAt, how does that sound?

.env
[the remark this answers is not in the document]
fanatic75: This has been changed to 

- _APP_MESSAGE_SMS_TEST_DSN=
- _APP_MESSAGE_EMAIL_TEST_DSN=
- _APP_MESSAGE_PUSH_TEST_DSN=

app/controllers/api/users.php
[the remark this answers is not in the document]
fanatic75: @stnguyen90 , added queries support for list targets.

src/Appwrite/Utopia/Response/Model/Message.php
fanatic75: Changed to ``deliveredTotal``

src/Appwrite/Utopia/Response/Model/Message.php
fanatic75: @abnegate, I think @stnguyen90 , might need ``data`` attribute in the message response

src/Appwrite/Utopia/Response/Model/Provider.php
```diff
@@ -7,44 +7,62 @@
...
-    /**
-     * @var bool
-     */
-    protected bool $public = false;
-
     public function __construct()
     {
         $this
-            ->addRule('key', [
+            ->addRule('$id', [
```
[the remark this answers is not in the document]
fanatic75: Added these

app/controllers/api/users.php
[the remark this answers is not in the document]
fanatic75: @abnegate , we don't have a search attribute in target collection, hope that doesn't limit us

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2145 @@
...
+    ->label('sdk.namespace', 'messaging')
+    ->label('sdk.method', 'updateMailgunProvider')
+    ->label('sdk.description', '/docs/references/messaging/update-mailgun-provider.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_PROVIDER)
+    ->param('providerId', '', new UID(), 'Provider ID.')
+    ->param('name', '', new Text(128), 'Provider name.', true)
+    ->param('enabled', null, new Boolean(), 'Set as enabled.', true)
+    ->param('isEuRegion', null, new Boolean(), 'Set as eu region.', true)
```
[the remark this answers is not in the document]
fanatic75: Fixed this

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2145 @@
...
+    ->label('sdk.auth', [APP_AUTH_TYPE_JWT, APP_AUTH_TYPE_SESSION, APP_AUTH_TYPE_ADMIN, APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'messaging')
+    ->label('sdk.method', 'createSubscriber')
+    ->label('sdk.description', '/docs/references/messaging/create-subscriber.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_SUBSCRIBER)
+    ->param('subscriberId', '', new CustomId(), 'Subscriber ID. Choose a custom Topic ID or a new Topic ID.')
+    ->param('topicId', '', new UID(), 'Topic ID.')
+    ->param('targetId', '', new UID(), 'Target ID.')
```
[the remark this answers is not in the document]
fanatic75: Updated the description here.

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2145 @@
...
+        }
+
+        $response
+            ->dynamic($subscriber, Response::MODEL_SUBSCRIBER);
+    });
+
+App::delete('/v1/messaging/topics/:topicId/subscriber/:subscriberId')
+    ->desc('Delete a subscriber from a topic.')
+    ->groups(['api', 'messaging'])
+    ->label('audits.event', 'subscriber.delete')
```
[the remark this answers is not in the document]
fanatic75: Updated it to subscribers for REST endpoint

app/controllers/api/messaging.php
fanatic75: Linked ISO 8601

app/controllers/api/messaging.php
stnguyen90: @fanatic75 wait...so sorry...did they rebrand or something?? Their website now says Textmagic 😬 Their [docs](https://docs.textmagic.com/) say TextMagic, though!!

app/controllers/api/messaging.php
fanatic75: Default and internal both are removed

app/controllers/api/users.php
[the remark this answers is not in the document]
fanatic75: This is added

app/controllers/api/messaging.php
fanatic75: Done this

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2145 @@
...
+        }
+        $providerAttr = $provider->getAttribute('provider');
+
+        if ($providerAttr !== 'mailgun') {
+            throw new Exception(Exception::PROVIDER_INCORRECT_TYPE);
+        }
+
+        if (!empty($name)) {
+            $provider->setAttribute('name', $name);
+            $provider->setAttribute('search', $provider->getId() . ' ' . $name . ' ' . 'mailgun' . ' ' . 'email');
```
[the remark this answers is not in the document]
fanatic75: Done this

app/controllers/api/messaging.php
fanatic75: These will work

src/Appwrite/Utopia/Response/Model/Message.php
abnegate: @stnguyen90 Does it need to be inside `data` or can it be treated like a Document, where the attributes inside `data` are hoisted to the top level?

app/controllers/api/users.php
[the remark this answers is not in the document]
abnegate: @stnguyen90 My mistake, you won't be able to via API. I don't think this is really a valid use case though, Firebase won't show you the individual tokens that each message was sent to, pretty sure they only show a count of how many people it was sent to

app/controllers/api/messaging.php
abnegate: All other parameters are required

app/controllers/api/messaging.php
abnegate: Looks like a rebrand, I would trust their main site to be most recent vs docs. Let's stick with `Textmagic`

app/controllers/api/messaging.php
abnegate: Subscribers will have a userId, and the userName will be dynamically included, the same as we have for memberships of a team

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2145 @@
...
+    ->label('sdk.method', 'createMsg91Provider')
+    ->label('sdk.description', '/docs/references/messaging/create-msg91-provider.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_PROVIDER)
+    ->param('providerId', '', new CustomId(), 'Provider ID. Choose a custom ID or generate a random ID with `ID.unique()`. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.')
+    ->param('name', '', new Text(128), 'Provider name.')
+    ->param('from', '', new Text(256), 'Sender number.')
+    ->param('senderId', '', new Text(0), 'Msg91 Sender ID.')
+    ->param('authKey', '', new Text(0), 'Msg91 Auth Key.')
```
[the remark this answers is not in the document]
abnegate: We'll need to somehow include this per message 🤔 we shouldn't force it to be set only once. It's not relevant for any other provider though, so doesn't make sense to include it in the send SMS route. Any ideas? cc @fanatic75

app/controllers/api/messaging.php
abnegate: @fanatic75 Renaming to scheduleAt sounds good :+1:

src/Appwrite/Utopia/Response/Model/Message.php
stnguyen90: Either is fine.

app/controllers/api/messaging.php
stnguyen90: > Subscribers will have a userId, and the userName will be dynamically included, the same as we have for memberships of a team

This only helps the front end show the user name w/o an extra API call. Is there any other approach to finding which topics a user is part of without making an API call per topic?

app/controllers/api/messaging.php
stnguyen90: According to the design, a developer needs to be able to start configuring a provider, save their progress, and go back to it to finish it at a later time.

stnguyen90 (commented):
Replied to some conversations

app/controllers/api/messaging.php
abnegate: Hmm no, I see the difficulty, how do we handle this for teams? Is there a way to list teams a user is a member of?

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2601 @@
...
+    ->label('sdk.auth', [APP_AUTH_TYPE_JWT, APP_AUTH_TYPE_SESSION, APP_AUTH_TYPE_ADMIN, APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'messaging')
+    ->label('sdk.method', 'createSubscriber')
+    ->label('sdk.description', '/docs/references/messaging/create-subscriber.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_SUBSCRIBER)
+    ->param('subscriberId', '', new CustomId(), 'Subscriber ID. Choose a custom Subscriber ID or a new Subscriber ID.')
+    ->param('topicId', '', new UID(), 'Topic ID. The topic ID to subscribe to.')
+    ->param('targetId', '', new UID(), 'Target ID. The target ID to link to the specified Topic ID.')
```
abnegate, line 1630: @fanatic75 This param doesn't make sense for the client side, are we able to auto-detect this in that case?

src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -60,48 +66,314 @@ public function action(Message $message): void
...
+        * @var array<string, array<string>> $identifiersByProviderId
+        */
+        $identifiersByProviderId = [];
+
+        /**
+        * @var Document[] $providers
+        */
+        $providers = [];
+        foreach ($recipients as $recipient) {
+            $providerId = $recipient->getAttribute('providerId');
```
stnguyen90, line 125: providerId isn't required in targets anymore so this won't work.

src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -60,48 +66,314 @@ public function action(Message $message): void
...
+        if (\count($targetsId) > 0) {
+            $targets = $dbForProject->find('targets', [Query::equal('$id', $targetsId)]);
+            $recipients = \array_merge($recipients, $targets);
         }
 
-        $from = App::getEnv('_APP_SMS_FROM');
+        $internalProvider = $dbForProject->findOne('providers', [
+            Query::equal('enabled', [true]),
+            Query::equal('type', [$recipients[0]->getAttribute('providerType')]),
+        ]);
```
stnguyen90, line 113: @fanatic75, what if `findOne()` doesn't find anything?

app/controllers/api/messaging.php
fanatic75: Targets can be iterated using getAccount api as we have discussed, closing this one

src/Appwrite/Platform/Workers/Messaging.php
fanatic75: this is fixed

app/controllers/api/users.php
```diff
@@ -379,6 +416,98 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+    ->label('scope', 'targets.write')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY, APP_AUTH_TYPE_ADMIN])
+    ->label('sdk.namespace', 'users')
+    ->label('sdk.method', 'createTarget')
+    ->label('sdk.description', '/docs/references/users/create-target.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_TARGET)
+    ->param('targetId', '', new CustomId(), 'Target ID. Choose a custom ID or generate a random ID with `ID.unique()`. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.')
+    ->param('userId', '', new UID(), 'User ID.')
```
stnguyen90, line 434: userId should go before targetId

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2961 @@
...
+    ->label('sdk.auth', [APP_AUTH_TYPE_ADMIN, APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'messaging')
+    ->label('sdk.method', 'createTopic')
+    ->label('sdk.description', '/docs/references/messaging/create-topic.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_TOPIC)
+    ->param('topicId', '', new CustomId(), 'Topic ID. Choose a custom Topic ID or a new Topic ID.')
+    ->param('name', '', new Text(128), 'Topic Name.')
+    ->param('description', '', new Text(2048), 'Topic Description.', true)
```
stnguyen90, line 1640: Does the validator need to be updated to allow empty string?

fanatic75: The test seems to work fine without description. If you are skipping the description property, the topic should be able to be created.

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,2961 @@
...
+        $queueForEvents
+            ->setParam('providerId', $provider->getId());
+
+        $response
+            ->setStatusCode(Response::STATUS_CODE_NOCONTENT)
+            ->noContent();
+    });
+
+App::post('/v1/messaging/topics')
+    ->desc('Create a topic.')
```
stnguyen90, line 1625: The desc of routes should not have a period at the end.

Bankrollshow (approved):

eldadfux (changes requested):

app/config/roles.php
```diff
@@ -8,7 +8,9 @@
     'home',
     'console',
     'graphql',
-    'account',
+    'sessions.write',
+    'accounts.read',
+    'accounts.write',
```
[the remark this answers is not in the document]
abnegate: Account/session one were but messaging weren't, added here https://github.com/appwrite/console/pull/761

app/controllers/api/functions.php
```diff
@@ -228,6 +228,7 @@
             fn () => $dbForConsole->createDocument('schedules', new Document([
                 'region' => App::getEnv('_APP_REGION', 'default'), // Todo replace with projects region
                 'resourceType' => 'function',
+                'resourceCollection' => 'functions',
```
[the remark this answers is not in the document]
abnegate: So that we can use abstracted code to fetch resources for multiple types, didn't want to assume that the collection would always be `$resourceType . 's';`

app/controllers/api/account.php
```diff
@@ -433,7 +462,9 @@
     ->label('abuse-limit', 50)
     ->label('abuse-key', 'ip:{ip}')
     ->label('docs', false)
-    ->param('provider', '', new WhiteList(\array_keys(Config::getParam('providers')), true), 'OAuth2 provider.')
+    ->label('usage.metric', 'sessions.{scope}.requests.create')
```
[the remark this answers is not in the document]
abnegate: Addressed here https://github.com/appwrite/appwrite/pull/7536

app/controllers/api/account.php
```diff
@@ -887,26 +969,28 @@
...
+    ->label('scope', 'sessions.write')
     ->label('auth.type', 'magic-url')
     ->label('audits.event', 'session.create')
     ->label('audits.resource', 'user/{response.userId}')
     ->label('audits.userId', '{response.userId}')
     ->label('sdk.auth', [])
     ->label('sdk.namespace', 'account')
-    ->label('sdk.method', 'createMagicURLSession')
-    ->label('sdk.description', '/docs/references/account/create-magic-url-session.md')
+    ->label('sdk.method', ['createMagicURLToken', 'createMagicURLSession'])
```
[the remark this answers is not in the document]
abnegate: We don't :thinking: looking at [spec generation](https://github.com/appwrite/appwrite/blob/1.5.x/src/Appwrite/Specification/Format/Swagger2.php#L126-L129) only the first will be generated for SDKs, cc @loks0n

app/controllers/api/account.php
```diff
@@ -3016,11 +3413,68 @@
...
+        $response
+            ->dynamic($target, Response::MODEL_TARGET);
+    });
+
 App::delete('/v1/account')
     ->desc('Delete account')
     ->groups(['api', 'account'])
     ->label('event', 'users.[userId].delete')
-    ->label('scope', 'account')
+    ->label('scope', 'accounts.write')
```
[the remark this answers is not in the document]
abnegate: Not sure if we can without some side-effects, @Meldiron could you please test?

src/Appwrite/Enum/MessageStatus.php
```diff
@@ -0,0 +1,27 @@
+<?php
```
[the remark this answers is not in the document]
abnegate: Addressed here https://github.com/appwrite/appwrite/pull/7537

app/controllers/api/functions.php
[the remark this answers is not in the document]
eldadfux: I would just handle a switch case according to the resource type, having both seems unnecessarily complex.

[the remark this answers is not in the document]
eldadfux: But I do agree we shouldn't try and build the collection name.

app/controllers/api/messaging.php
```diff
@@ -0,0 +1,3498 @@
...
+    ->param('port', null, new Range(1, 65535), 'SMTP port.', true)
+    ->param('username', '', new Text(0), 'Authentication username.', true)
+    ->param('password', '', new Text(0), 'Authentication password.', true)
+    ->param('encryption', '', new WhiteList(['none', 'ssl', 'tls']), 'Encryption type. Can be \'ssl\' or \'tls\'', true)
+    ->param('autoTLS', null, new Boolean(), 'Enable SMTP AutoTLS feature.', true)
+    ->param('fromName', '', new Text(128), 'Sender Name.', true)
+    ->param('fromEmail', '', new Email(), 'Sender email address.', true)
+    ->param('replyToName', '', new Text(128), 'Name set in the Reply To field for the mail. Default value is Sender Name.', true)
+    ->param('replyToEmail', '', new Text(128), 'Email set in the Reply To field for the mail. Default value is Sender Email.', true)
+    ->param('enabled', null, new Boolean(), 'Set as enabled.', true)
```
stnguyen90, line 1197: should this have the `mailer` param too?

TorstenDittmann (approved):

Issue #6881, opened 2023-10-11
🐛 Bug Report: Storage rate limit while uploading big file 


joern-h:
### 👟 Reproduction steps

Upload a big file (1GB) to a Storage Bucket
I use sdk-for-flutter which uploads files in 5MB chunks

### 👍 Expected behavior

It should upload the file without rate limiting.

Uploading the file from the Appwrite console works.

### 👎 Actual Behavior

Upload will be rate limited:

Client log:
```
Error: AppwriteException: general_rate_limit_exceeded, Rate limit for the current endpoint has been exceeded. Please try again after some time. (429)
```

Server log:
```
appwrite  | [Error] Timestamp: 2023-10-11T08:59:29+00:00
appwrite  | [Error] Method: POST
appwrite  | [Error] URL: /v1/storage/buckets/:bucketId/files
appwrite  | [Error] Type: Appwrite\Extend\Exception
appwrite  | [Error] Message: Rate limit for the current endpoint has been exceeded. Please try again after some time.
appwrite  | [Error] File: /usr/src/code/app/controllers/shared/api.php
appwrite  | [Error] Line: 169
```

### 🎲 Appwrite version

Version 1.4.x

### 💻 Operating system

Linux

### 🧱 Your Environment

Appwrite 1.4.5
Flutter SDK: 11.0.0

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

stnguyen90:
@joern-h, thanks for raising this issue! 🙏🏼 It looks like the rate limit (60 requests in 60 seconds) is being hit because the chunking causes the upload to be broken down into multiple requests. I'm curious to see how many others face this problem 🧐

PR #6893, opened 2023-10-12
Add _APP_STORAGE_ABUSE_LIMIT variable

joern-h:
## What does this PR do?

Implement a new environment variable to configure the abuse-limit for file creation.

## Test Plan

I verified my changes via integration testing by uploading a 1GB file with a Flutter application.
If I let the variable at its default (60) the upload will fail after chunk 60.
After increasing the limit, i am able to upload bigger files without running into the rate limit.

## Related PRs and Issues

Issue #6881

## Checklist


stnguyen90 (commented):
Great PR! 🤯 This is spot on for the implementation if we wanted to use environment variables to control this!

I'm unsure if we want to go with this approach to solve the underlying problem. We may wish to have a more comprehensive solution for handling rate limits for multiple endpoints and maybe even configurable per project.

PR #6894, opened 2023-10-12
Add VCS tests

vermakhushboo:
## What does this PR do?

Add tests for VCS(Version Control System) API controller and related functionality. 

## Test Plan

NA

## Related PRs and Issues

- (Related PR or issue)

## Checklist


.env
```diff
@@ -96,6 +96,7 @@ _APP_VCS_GITHUB_APP_ID=
 _APP_VCS_GITHUB_CLIENT_ID=
 _APP_VCS_GITHUB_CLIENT_SECRET=
 _APP_VCS_GITHUB_WEBHOOK_SECRET=
+_APP_VCS_TEST_GITHUB_INSTALLATION_ID=
```
Meldiron, line 99: Let's confirm variable name with Christy. I would probably suggest keeping it away from _APP scope. Or making TEXT primary. Soo something like

- `_APP_TESTS_VCS_GITHUB_INSTALLATION_ID`
- `_TESTS_VCS_GITHUB_INSTALLATION_ID`

app/controllers/mock.php
```diff
@@ -646,6 +649,69 @@
         $response->noContent();
     });
 
+App::get('/v1/mock/github/callback')
+    ->desc('Create installation document using GitHub installation id')
+    ->groups(['mock', 'api', 'vcs'])
```
Meldiron, line 654: Let's only keep `mock` group here, unless we really need other ones for something

app/controllers/mock.php
```diff
@@ -646,6 +649,69 @@
...
+    ->label('scope', 'public')
```
Meldiron, line 655: Let's add `->label('docs', false)`, I think it prevents this method to getting to SDKs

app/controllers/mock.php
```diff
@@ -646,6 +649,69 @@
...
+    ->param('installation_id', '', new Text(256, 0), 'GitHub installation ID', true)
+    ->param('projectId', '', new Text(2048), 'Project ID of the project where app is to be installed', true)
```
Meldiron, line 657: We can use `new UID()` validator instead of Text I think.

app/controllers/mock.php
```diff
@@ -646,6 +649,69 @@
...
+    ->inject('gitHub')
+    ->inject('project')
+    ->inject('response')
+    ->inject('dbForConsole')
+    ->action(function (string $providerInstallationId, string $projectId, GitHub $github, Document $project, Response $response, Database $dbForConsole) {
+        if (empty($projectId)) {
+            $error = 'Installation requests from organisation members for the Appwrite GitHub App are currently unsupported. To proceed with the installation, login to the Appwrite Console and install the GitHub App.';
+            throw new Exception(Exception::GENERAL_ARGUMENT_INVALID, $error);
+        }
```
Meldiron, line 666: To mark proejct ID required, you can remove last `true` param from `->param()` call.

```php
    ->param('projectId', '', new Text(2048), 'Project ID of the project where app is to be installed')
```

That way its required, so action wont even run if projected isn't provided.

app/controllers/mock.php
```diff
@@ -646,6 +649,69 @@
...
+    ->desc('Create installation document using GitHub installation id')
+    ->groups(['mock', 'api', 'vcs'])
+    ->label('scope', 'public')
+    ->param('installation_id', '', new Text(256, 0), 'GitHub installation ID', true)
+    ->param('projectId', '', new Text(2048), 'Project ID of the project where app is to be installed', true)
+    ->inject('gitHub')
+    ->inject('project')
+    ->inject('response')
+    ->inject('dbForConsole')
+    ->action(function (string $providerInstallationId, string $projectId, GitHub $github, Document $project, Response $response, Database $dbForConsole) {
```
Meldiron, line 662: `string $providerInstallationId`
It's best to have variable name same as parameter name, which is currently `installation_id`.

Let's call param `providerInstallationId` as well. After this change, you may need to update some tests to use new param name.

app/controllers/mock.php
```diff
@@ -646,6 +649,69 @@
...
+                'organization' => $owner,
+                'personal' => false
+            ]);
+
+            $installation = $dbForConsole->createDocument('installations', $installation);
+        } else {
+            $error = 'Installation of the Appwrite GitHub App on organization accounts is restricted to organization owners. As a member of the organization, you do not have the necessary permissions to install this GitHub App. Please contact the organization owner to create the installation from the Appwrite console.';
+
+            throw new Exception(Exception::GENERAL_ARGUMENT_INVALID, $error);
+        }
```
Meldiron, line 708: Same as comment above. If you remove `true` from `param()` definition, you don't need this empty check and exception.

app/controllers/mock.php
```diff
@@ -646,6 +649,69 @@
...
+
+        $project = $dbForConsole->getDocument('projects', $projectId);
+
+        if ($project->isEmpty()) {
+            $error = 'Project with the ID from state could not be found.';
+            throw new Exception(Exception::PROJECT_NOT_FOUND, $error);
+        }
+
+        // Create / Update installation
```
Meldiron, line 675: Let's remove this comment, seems like we only create the installation.

tests/e2e/Services/VCS/VCSBase.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Tests\E2E\Services\VCS;
+
+trait VCSBase
```
Meldiron, line 5: We define the trait but never use it.

I think you need `trait VCSBase extends Scope` here , and then in `tests/e2e/Services/VCS/VCSConsoleClientTest.php` you do `class VCSConsoleClientTest extends VCSBase`.

tests/e2e/Services/VCS/VCSConsoleClientTest.php
```diff
@@ -0,0 +1,316 @@
...
+    use ProjectCustom;
+    use SideConsole;
+
+    public ?string $providerRepositoryId = null;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->providerRepositoryId = '705764267';
+    }
```
Meldiron, line 27:
```suggestion
    public string $providerRepositoryId = '705764267';
```

We might not need setup for this.

app/controllers/mock.php
Meldiron, line 662: At beginning of action, lets add something like this: https://github.com/appwrite/appwrite/blob/main/app/controllers/mock.php#L632-L636

That ensures the mock endpoint cannot be used on production.

We might want to ask Chrsty about this, since we maaaaybe want to disable all mock endpoints on production. So we might want to do this in `::init()` of mock group. Maybe. But maybe we actually use some mock endpoints for something like SDK tests, so please confirm with Chrsity.

tests/e2e/Services/VCS/VCSConsoleClientTest.php
```diff
@@ -0,0 +1,316 @@
...
+         */
+
+        $installation = $this->client->call(Client::METHOD_GET, '/vcs/installations/' . $installationId, array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()), [
+            'installationId' => $installationId
+        ]);
+
+        $this->assertEquals(200, $installation['headers']['status-code']);
```
Meldiron, line 63: Let's add extra checks. Check value of provider, project ID, personal, providerInstallationId.. See what all is in the response, and assert what is expected.

tests/e2e/Services/VCS/VCSConsoleClientTest.php
```diff
@@ -0,0 +1,316 @@
...
+{
+    use ProjectCustom;
+    use SideConsole;
+
+    public ?string $providerRepositoryId = null;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->providerRepositoryId = '705764267';
```
Meldiron, line 26: If `providerRepositoryId` is hard-coded, we might not need `_APP_VCS_TEST_GITHUB_INSTALLATION_ID` env variable, and simply do variable `providerInstallationId`.
I would prefer having both as ENV variable, but if we want to avoid env vars, we can avoid it for both.

tests/e2e/Services/VCS/VCSConsoleClientTest.php
```diff
@@ -0,0 +1,316 @@
...
+
+        $this->assertEquals(200, $runtime['headers']['status-code']);
+        $this->assertEquals($runtime['body']['runtime'], 'ruby-3.1');
+
+        /**
+         * Test for FAILURE
+         */
+
+        $runtime = $this->client->call(Client::METHOD_POST, '/vcs/github/installations/' . $installationId . '/providerRepositories/randomRepositoryId/detection', array_merge([
+            'content-type' => 'application/json',
```
Meldiron, line 91: Content type should not be needed

Meldiron: Its in multiple places of this file. See if it works without it. If yes, lets remove it.

tests/e2e/Services/VCS/VCSConsoleClientTest.php
```diff
@@ -0,0 +1,316 @@
...
+        /**
+         * Test for SUCCESS
+         */
+
+        $runtime = $this->client->call(Client::METHOD_POST, '/vcs/github/installations/' . $installationId . '/providerRepositories/' . $this->providerRepositoryId . '/detection', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()), [
+            'installationId' => $installationId,
+            'providerRepositoryId' => $this->providerRepositoryId
```
Meldiron, line 80: This might not be needed. you already repository ID in the URL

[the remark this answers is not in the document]
Meldiron: I can see you provide installationId in body  and URL in almost all tests. Please check it on all places

tests/e2e/Services/VCS/VCSConsoleClientTest.php
```diff
@@ -0,0 +1,316 @@
...
+            'installationId' => 'randomInstallationId'
+        ]);
+
+        $this->assertEquals(404, $repositories['headers']['status-code']);
+    }
+
+    /**
+     * @depends testGitHubAuthorize
+     */
+    public function testGetRepository(string $installationId, string $providerRepositoryId2 = '705764449')
```
Meldiron, line 148: `providerRepositoryId2` should probably be class variable like `providerRepositoryId`. So its at the top of the file, easy to find and change if needed.

tests/e2e/Services/VCS/VCSConsoleClientTest.php
```diff
@@ -0,0 +1,316 @@
...
+        $repository = $this->client->call(Client::METHOD_GET, '/vcs/github/installations/' . $installationId . '/providerRepositories/' . $this->providerRepositoryId, array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()), [
+            'installationId' => $installationId,
+            'providerRepositoryId' => $this->providerRepositoryId
+        ]);
+
+        $this->assertEquals(200, $repository['headers']['status-code']);
+        $this->assertEquals($repository['body']['name'], 'ruby-starter');
```
Meldiron, line 163: Let's check all expected parameters, not just name.

tests/e2e/Services/VCS/VCSConsoleClientTest.php
```diff
@@ -0,0 +1,316 @@
...
+
+        $repositories = $this->client->call(Client::METHOD_GET, '/vcs/github/installations/' . $installationId . '/providerRepositories', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()), [
+            'installationId' => $installationId
+        ]);
+
+        $this->assertEquals(200, $repositories['headers']['status-code']);
+        $this->assertEquals($repositories['body']['total'], 3);
```
Meldiron, line 118: Let's get first repository and check all expected parameters.

tests/e2e/Services/VCS/VCSConsoleClientTest.php
```diff
@@ -0,0 +1,316 @@
...
+                'users.*.delete',
+            ],
+            'schedule' => '0 0 1 1 *',
+            'timeout' => 10,
+            'installationId' => $installationId,
+            'providerRepositoryId' => $this->providerRepositoryId,
+            'providerBranch' => 'main',
+        ]);
+
+        $this->assertEquals(201, $function['headers']['status-code']);
```
Meldiron, line 253: Let's also ensure that response has keys we just set. Sounds silly, but very reasonable in this case

tests/e2e/Services/VCS/VCSConsoleClientTest.php
```diff
@@ -0,0 +1,316 @@
...
+        return [
+            'installationId' => $installationId,
+            'functionId' => $function['body']['$id']
+        ];
+    }
+
+    /**
+     * @depends testCreateFunctionUsingVCS
+     */
+    public function testUpdateFunctionUsingVCS(array $data, string $providerRepositoryId2 = '705764449')
```
Meldiron, line 264: Same comment as above, this $providerRepositoryId2 should be defined at the top of the class.

tests/e2e/Services/VCS/VCSConsoleClientTest.php
```diff
@@ -0,0 +1,316 @@
...
+                'users.*.delete',
+            ],
+            'schedule' => '0 0 1 1 *',
+            'timeout' => 10,
+            'installationId' => $data['installationId'],
+            'providerRepositoryId' => $providerRepositoryId2,
+            'providerBranch' => 'main',
+        ]);
+
+        $this->assertEquals(200, $function['headers']['status-code']);
```
Meldiron, line 286: Same as above, lets check specific params on the response to ensure they were set properly.

tests/e2e/Services/VCS/VCSConsoleClientTest.php
```diff
@@ -0,0 +1,316 @@
...
+
+        return $function['body']['$id'];
+    }
+
+    /**
+     * @depends testGitHubAuthorize
+     */
+    public function testCreateRepository(string $installationId)
```
Meldiron, line 294: Let's include failure test. When you try to create repository twice, second one should fail as it will be name conflict.

tests/e2e/Services/VCS/VCSBase.php
vermakhushboo: I don't think `trait` supports `extends` syntax. Also, here the format I've followed is what we have in other tests like functions.

tests/e2e/Services/VCS/VCSConsoleClientTest.php
vermakhushboo: Removed it from wherever the tests were working without it.

tests/e2e/Services/VCS/VCSConsoleClientTest.php
vermakhushboo: The detect runtime only returns the detected runtime in response.

app/console
christyjacob4, line 1: we need to remove this file from the PR

tests/e2e/Services/VCS/VCSBase.php
```diff
@@ -0,0 +1,7 @@
...
+{
```
christyjacob4, line 6: The VCS endpoints are supposed to be accessed only by the console right? 

Lets have two more classes VCSCustomServerTest.php VCSCustomClientTest.php

In the first, you try to access the vcs endpoints using an API key and it should fail with the appropriate error code. 
Similarly access with a regular user cookie in the latter.

Just 1 test each to make sure the route is not accessible.

app/controllers/api/vcs.php
```diff
@@ -458,9 +458,9 @@
         $github->initializeVariables($providerInstallationId, $privateKey, $githubAppId);
 
         $owner = $github->getOwnerName($providerInstallationId);
-        $repositoryName = $github->getRepositoryName($providerRepositoryId);
-
-        if (empty($repositoryName)) {
+        try {
+            $repositoryName = $github->getRepositoryName($providerRepositoryId);
```
christyjacob4, line 462: To make sure our code is not affected by edge cases, you should still check if the repository name is empty and then through the same error accordingly.

christyjacob4 (changes requested):
Please address the changes. and also sync with main. We have adopted a new test mechanism to speed up tests. 

You can add your test here to ensure it runs in parallel
https://github.com/appwrite/appwrite/blob/main/.github/workflows/tests.yml#L85

app/controllers/api/vcs.php
Meldiron: I see same try/catch in multiple places. Lets make sure to address it everywhere.

tests/e2e/Services/VCS/VCSConsoleClientTest.php
```diff
@@ -0,0 +1,319 @@
...
+use Tests\E2E\Scopes\ProjectCustom;
+use Tests\E2E\Scopes\SideConsole;
+use Utopia\App;
+use Utopia\Database\Helpers\ID;
+use Utopia\Database\Helpers\Role;
+use Utopia\VCS\Adapter\Git\GitHub;
+use Utopia\Cache\Adapter\None;
+use Utopia\Cache\Cache;
+
+class VCSConsoleClientTest extends Scope
```
Meldiron, line 16: VCS tests won't work locally, until you set up some VCS env variables, right? For example stuff like secret keys we cant simply have in .env.

@christyjacob4 do you think we should see if secrets for VCS tests are missing, and if so, skip it? That way it doesn't run (and doesn't fail) locally, but it still runs in CI/CD. And if it fails on CI/CD and you want to investigate, you can set variables.

Meldiron: I am thinking something like this, in VCSBase: https://github.com/appwrite/appwrite/blob/073b9963d20945f4f1b6c6b35284fc3007940062/tests/e2e/General/AbuseTest.php#L20-L27

Meldiron: (in PM Christy confirmed we should try to do this)

app/console
vermakhushboo: Unable to understand how to remove it 😅

christyjacob4 (approved):

PR #6927, opened 2023-10-16
fixes#6844  UPDATED the update method files.

Salonikansal2:
In updateDocument Method  code is updated in all the files,

which increases readability for the users, by reading this they would know they have to pass the object with attributes they want to update.


stnguyen90:
@Salonikansal2, thanks for creating this PR! 🙏🏼 Unfortunately, the files you modified are automatically generated so the next time we generate, your changes would be wiped.

If you're looking to contribute, please look for issues labeled "good first issue".

PR #6934, opened 2023-10-16
Delegate custom deletes

abnegate:
## What does this PR do?

Move database and collections deletes to the database worker

## Test Plan

Manual tests

## Checklist


eldadfux (approved):

app/controllers/api/databases.php
```diff
@@ -3532,8 +3532,10 @@ function updateAttribute(
 
         $processDocument($collection, $document);
 
-        $queueForDeletes
-            ->setType(DELETE_TYPE_AUDIT)
+        $queueForDatabase
+            ->setType(DATABASE_TYPE_DELETE_DOCUMENT)
+            ->setDatabase($database)
+            ->setCollection($collection)
```
christyjacob4, line 3538: This is the delete document endpoint.. Why do we need to send this to a worker? We delete the document in the API itself right?

src/Appwrite/Platform/Workers/Databases.php
```diff
@@ -480,27 +455,179 @@ private function deleteIndex(Document $database, Document $collection, Document
...
+     * @param Document $project
+     * @param Database $dbForProject
+     * @return void
+     * @throws Authorization
+     * @throws Conflict
+     * @throws DatabaseException
+     * @throws Restricted
+     * @throws Structure
+     */
+    protected function deleteCollection(Document $database, Document $collection, Document $project, Database $dbForProject): void
```
christyjacob4, line 494: Shouldn't we now remove this code from the deletes worker ?

src/Appwrite/Platform/Workers/Databases.php
```diff
@@ -480,27 +455,179 @@ private function deleteIndex(Document $database, Document $collection, Document
...
     }
+
+    /**
+     * @param Document $database
+     * @param Document $project
+     * @param $dbForProject
+     * @return void
+     * @throws Exception
+     */
+    protected function deleteDatabase(Document $database, Document $project, $dbForProject): void
```
christyjacob4, line 471: Shouldn't we now remove this code from the deletes worker ?

app/controllers/api/databases.php
eldadfux: I think this is for delete the document audit and not the document itself. We can leave it as is.

src/Appwrite/Platform/Workers/Databases.php
eldadfux: I think we can.

src/Appwrite/Platform/Workers/Databases.php
eldadfux: I think we can.

PR #6940, opened 2023-10-17
chore: cleanup obsolete CLI tasks

christyjacob4:
## What does this PR do?
* Removes some obsolete CLI tasks that we don't use anymore
* Fix Flaky tests

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


composer.lock
shimonewman, line 1: Does the lock file needs  pushed?

app/worker.php
```diff
@@ -271,6 +271,7 @@
         Console::error('[Error] Message: ' . $error->getMessage());
         Console::error('[Error] File: ' . $error->getFile());
         Console::error('[Error] Line: ' . $error->getLine());
+        Console::error('[Error] Trace: ' . $error->getTraceAsString());
```
TorstenDittmann, line 274: general question, do we want the trace on production?

phpunit.xml
```diff
@@ -19,19 +19,19 @@
             <file>./tests/e2e/Client.php</file>
             <directory>./tests/e2e/General</directory>
             <directory>./tests/e2e/Scopes</directory>
+            <directory>./tests/e2e/Services/Teams</directory>
+            <directory>./tests/e2e/Services/Realtime</directory>
+            <directory>./tests/e2e/Services/Webhooks</directory>
```
TorstenDittmann, line 24: iirc we moved the webhooks to the bottom like functions - because they take the longest

src/Appwrite/Platform/Workers/Deletes.php
```diff
@@ -806,7 +806,7 @@ private function deleteDeployment(callable $getProjectDB, callable $getFunctions
          * Delete builds
          */
         Console::info("Deleting builds for deployment " . $deploymentId);
-        $storageBuilds = $$getBuildsDevice($projectId);
+        $storageBuilds = $getBuildsDevice($projectId);
```
TorstenDittmann, line 809: uff, did that even work in the first place?

phpunit.xml
christyjacob4: Moved it to the end 👍

src/Appwrite/Platform/Workers/Deletes.php
christyjacob4: Surprisingly this didn't crash the worker 😅 and only failed when we deleted a deployment

eldadfux (approved):

PR #6949, opened 2023-10-19
API DB timeout

abnegate:
## What does this PR do?

Force API queries to the database to timeout after 15 seconds of execution

## Test Plan

Added integration test for timeout

## Related PRs and Issues

- N/A

## Checklist


app/controllers/api/databases.php
```diff
@@ -381,6 +383,18 @@ function updateAttribute(
...
 }
 
+App::init()
+    ->groups(['api', 'database'])
+    ->inject('request')
+    ->inject('dbForProject')
+    ->action(function (Request $request, Database $dbForProject) {
+        $timeout = \intval($request->getHeader('x-appwrite-timeout'));
+
+        if (!empty($timeout) && App::isDevelopment()) {
```
fogelito, line 393: This will be active only on dev?

christyjacob4: good question.

app/controllers/api/databases.php
```diff
@@ -381,6 +383,18 @@ function updateAttribute(
     return $attribute;
 }
 
+App::init()
+    ->groups(['api', 'database'])
```
christyjacob4, line 387: why's this applied specifically to the `database` group ?  shouldn't `api` already cover `databases` as well?

christyjacob4: Also we have other `get` and `list` endpoints in other services which should also have the timeouts

composer.json
```diff
@@ -43,13 +43,13 @@
...
-        "utopia-php/abuse": "0.31.*",
+        "utopia-php/abuse": "0.32.*",
         "utopia-php/analytics": "0.10.*",
-        "utopia-php/audit": "0.33.*",
+        "utopia-php/audit": "0.34.*",
         "utopia-php/cache": "0.8.*",
         "utopia-php/cli": "0.15.*",
         "utopia-php/config": "0.2.*",
-        "utopia-php/database": "0.43.*",
+        "utopia-php/database": "dev-feat-global-timeouts as 0.44.1",
```
christyjacob4, line 52: reminder to change this once we release the library

app/controllers/general.php
```diff
@@ -587,7 +587,7 @@ function router(App $utopia, Database $dbForConsole, SwooleRequest $swooleReques
         $response
             ->addHeader('Server', 'Appwrite')
             ->addHeader('Access-Control-Allow-Methods', 'GET, POST, PUT, PATCH, DELETE')
-            ->addHeader('Access-Control-Allow-Headers', 'Origin, Cookie, Set-Cookie, X-Requested-With, Content-Type, Access-Control-Allow-Origin, Access-Control-Request-Headers, Accept, X-Appwrite-Project, X-Appwrite-Key, X-Appwrite-Locale, X-Appwrite-Mode, X-Appwrite-JWT, X-Appwrite-Response-Format, X-SDK-Version, X-SDK-Name, X-SDK-Language, X-SDK-Platform, X-SDK-GraphQL, X-Appwrite-ID, X-Appwrite-Timestamp, Content-Range, Range, Cache-Control, Expires, Pragma, X-Fallback-Cookies')
+            ->addHeader('Access-Control-Allow-Headers', 'Origin, Cookie, Set-Cookie, X-Requested-With, Content-Type, Access-Control-Allow-Origin, Access-Control-Request-Headers, Accept, X-Appwrite-Project, X-Appwrite-Key, X-Appwrite-Locale, X-Appwrite-Mode, X-Appwrite-JWT, X-Appwrite-Response-Format, X-Appwrite-Timeout, X-SDK-Version, X-SDK-Name, X-SDK-Language, X-SDK-Platform, X-SDK-GraphQL, X-Appwrite-ID, X-Appwrite-Timestamp, Content-Range, Range, Cache-Control, Expires, Pragma, X-Fallback-Cookies')
```
christyjacob4, line 590: Where do we send this header?

christyjacob4: I mean where do we set a value for it ?

app/controllers/api/databases.php
abnegate: Yep this is just for tests :+1:

PR #6988, opened 2023-10-23
Fix missing body in async function execution

shimonewman:
Fix a bug in which the body was not populated for async function executions.


stnguyen90:
Create execution with `async = true` includes global variables, function variables, headers, and body.

image

Scheduled execution includes global variables and function variables:

image

Execution triggered by event includes global variables, function variables, headers, raw body, and body:

image

PR #7003, opened 2023-10-25
Feat query metadata

abnegate:
## What does this PR do?

Add the host, project ID and user ID as metadata to database instances. Will be printed as a part of the query for debugging.

## Test Plan

Existing tests

## Related PRs and Issues

N/A

## Checklist


app/init.php
```diff
@@ -1049,6 +1049,9 @@ function (mixed $value) {
         }
     }
 
+    $dbForProject->setMetadata('user', $user->getId());
```
eldadfux, line 1052: Why not do it directly when we create the DB instance and have `user` as a dependency?

abnegate: We inject the db's when creating the user resource so we'd end up with a cyclic dependency

eldadfux (approved):

PR #7016, opened 2023-10-26
Sync 1.4.x with main

christyjacob4:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


TorstenDittmann (approved):

.github/workflows/tests.yml
```diff
@@ -4,57 +4,118 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  IMAGE: appwrite-dev
+  CACHE_KEY: appwrite-dev-${{ github.event.pull_request.head.sha }}
```
stnguyen90, line 9: wouldn't this make every subsequent commit in PR use a different cache so we can't rely on docker image layer caching?

christyjacob4: This cache is created so it can be shared across different jobs in the same workflow. I did try 

As a result the image is just built once in the first step and its consumed by all the other jobs. Which is much faster than rebuilding the image in each step.

We still leverage docker image layer caching here https://github.com/appwrite/appwrite/blob/main/.github/workflows/tests.yml#L33-L34

The cache key I use is not related to the internal caches that docker uses

stnguyen90 (approved):

PR #7056, opened 2023-10-31
Added cURL timeout and not store response data for webhooks

vermakhushboo:
## What does this PR do?

This PR targets to prevent webhook abuse by:
1. Setting a CURL timeout of 15 seconds for the webhook request.

2. Setting the CURL to avoid getting or downloading the request response. This will help us avoid a potential abusive behavior that tries to get us to download a large payload to the worker memory. 

Also, formatted the `Webhooks.php` file.

## Test Plan
Ran webhook service tests and they pass.

## Related PRs and Issues
https://app.asana.com/0/1203764161503708/1205842531745480/f

## Checklist


TorstenDittmann (approved):
have you double checked that actually the payload is not received?

You could try doing webhooks against one of those files http://hel.icmp.hetzner.com/ and see how fast its going.

vermakhushboo:
@TorstenDittmann I've verified that with the current changes, the operation times out after 15 seconds for a 1 GB file and runs fine for smaller files that require less than 15 seconds.

eldadfux:
@vermakhushboo I would just verify the memory status while downloading the 1GB file. In best case scenario it shouldn't load to memory at all.

christyjacob4 (approved):

PR #7024, opened 2023-10-27
Fix: Cookie Header

Meldiron:
## What does this PR do?

Fixes cookie headers not being passed properly by router.

## Test Plan

- [x] Maual QA

CleanShot 2023-10-27 at 10 30 45@2x

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/6924

## Checklist

- [x] Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [x] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

PR #7059, opened 2023-10-31
Fix 400 error on function domain execution

stnguyen90:
### Fix warning when cookie is null

`$this->swoole->cookie` was being referenced when it was null causing these warnings:

image

This PR checks for empty value before trying to iterate it.

### Fix incorrect genera_prototocol_unsupported error

`$requestHeaders` was being referenced before it was set. This PR reverts code back to using `$swooleRequest` to extract the header.

## Test Plan

Manual

## Related PRs and Issues

Previous PR:

* https://github.com/appwrite/appwrite/pull/7024

## Checklist

### DIFF
--- a/app/controllers/general.php
+++ b/app/controllers/general.php
@@ -408,7 +408,7 @@ App::init()
         * @see https://www.owasp.org/index.php/List_of_useful_HTTP_headers
         */
         if (App::getEnv('_APP_OPTIONS_FORCE_HTTPS', 'disabled') === 'enabled') { // Force HTTPS
-            if ($request->getProtocol() !== 'https' && ($requestHeaders['host'] ?? '') !== 'localhost' && ($requestHeaders['host'] ?? '') !== APP_HOSTNAME_INTERNAL) { // localhost allowed for proxy, APP_HOSTNAME_INTERNAL allowed for migrations
+            if ($request->getProtocol() !== 'https' && ($swooleRequest->header['host'] ?? '') !== 'localhost' && ($swooleRequest->header['host'] ?? '') !== APP_HOSTNAME_INTERNAL) { // localhost allowed for proxy, APP_HOSTNAME_INTERNAL allowed for migrations
                 if ($request->getMethod() !== Request::METHOD_GET) {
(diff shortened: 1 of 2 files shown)


app/controllers/general.php
```diff
@@ -408,7 +408,7 @@ function router(App $utopia, Database $dbForConsole, SwooleRequest $swooleReques
         * @see https://www.owasp.org/index.php/List_of_useful_HTTP_headers
         */
         if (App::getEnv('_APP_OPTIONS_FORCE_HTTPS', 'disabled') === 'enabled') { // Force HTTPS
-            if ($request->getProtocol() !== 'https' && ($requestHeaders['host'] ?? '') !== 'localhost' && ($requestHeaders['host'] ?? '') !== APP_HOSTNAME_INTERNAL) { // localhost allowed for proxy, APP_HOSTNAME_INTERNAL allowed for migrations
+            if ($request->getProtocol() !== 'https' && ($swooleRequest->header['host'] ?? '') !== 'localhost' && ($swooleRequest->header['host'] ?? '') !== APP_HOSTNAME_INTERNAL) { // localhost allowed for proxy, APP_HOSTNAME_INTERNAL allowed for migrations
```
christyjacob4, line 411: Lets keep this to `$requestHeaders` itself since the issue with empty cookies is not taken care of

stnguyen90: $requestHeaders is defined elsewhere. It shouldn't be used here.

christyjacob4: Screenshot 2023-10-31 at 11 07 53 PM

This was the original PR that affected this line of code.. so requestHeaders is something we declared.

stnguyen90: $requestHeadres is defined in the `router()` function:

https://github.com/appwrite/appwrite/blob/9fded16d3f90d3a0be61350bcbb73905611cf471/app/controllers/general.php#L48-L197

not in the block i updated.

stnguyen90: Also, I don't think the cookie change is related to the line I updated.

christyjacob4: Got it 👍

christyjacob4 (approved):

PR #7082, opened 2023-11-06
Add XDebug

PineappleIOnic:
## What does this PR do?

This PR adds XDebug's configuration files to Appwrite to leverage the XDebug extension implemented in the PR here: https://github.com/appwrite/docker-base/pull/20

## Test Plan

Build with the new docker-base and test if the debugger works

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,7 @@ ENV VITE_APPWRITE_GROWTH_ENDPOINT=$VITE_APPWRITE_GROWTH_ENDPOINT
 RUN npm ci
 RUN npm run build
 
-FROM appwrite/base:0.5.0 as final
+FROM appwrite/base:0.6.0 as final
 
 LABEL maintainer="team@appwrite.io"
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -56,6 +56,7 @@ COPY ./public /usr/src/code/public
 COPY ./bin /usr/local/bin
 COPY ./docs /usr/src/code/docs
 COPY ./src /usr/src/code/src
+COPY ./dev /usr/src/code/dev
```
abnegate, line 59: Let's remove this from the Dockerfile but mount it in docker-compose.yml, so that it's only present in dev mode

PineappleIOnic: The only problem is that we copy the file into the php config directories later in these steps, unlike YASD there doesn't appear to be an init_file config for xdebug's ini files.

https://github.com/appwrite/appwrite/blob/77caf64076c6f56bc18bd173a538cd6775cd2000/Dockerfile#L114C7-L114C7

abnegate: Makes sense, we can still remove it for non-debug, just needs an extra instruction _after_ the `COPY`:

Let's do the same to remove the xdebug extension file altogether if not in debug. Even if it's not loaded, it's bad practice to have XDebug present in production.

```dockerfile
RUN if [ "$DEBUG" = "false" ]; then rm -rf /usr/src/code/dev; fi
RUN if [ "$DEBUG" = "false" ]; then rm -f /usr/local/lib/php/extensions/no-debug-non-zts-20220829/xdebug.so
```

abnegate (commented):
We can now update docker-base to [0.6.0](https://github.com/appwrite/docker-base/releases/tag/0.6.0)

abnegate (approved):

PR #7092, opened 2023-11-07
Handle cases where password history could contain NULLs

PineappleIOnic:
## What does this PR do?

This adds a small extra check before we validate a password in password history that it isn't a null, while this shouldn't ever happen in a normal context we have seen reports of it occurring in our logs

## Test Plan

Manually modify my database to add nulls into a users password history and check it as I'm not entirely sure how the issue occurred in the first place

## Checklist


PineappleIOnic:
The || might not be correct after thinking this through a little more

christyjacob4 (approved):

PR #7095, opened 2023-11-08
Add commit flag to delete orphaned projects task

shimonewman:


src/Appwrite/Platform/Tasks/DeleteOrphanedProjects.php
```diff
@@ -110,6 +115,15 @@ public function action(Group $pools, Cache $cache, Database $dbForConsole, Regis
...
+        Console::log('Iterated through ' . $count - 1 . '/' . $totalProjects . ' projects found ' . $orphans - 1 . ' orphans');
+    }
+
+    private function deleteProject(Database $dbForConsole, $projectId): void
+    {
+        try {
+            $dbForConsole->deleteDocument('projects', $projectId);
+        } catch (\Throwable $th) {
+            Console::error('Error when trying to delete project (' . $projectId . ') ' . $th->getMessage());
+        }
```
christyjacob4, line 127: You probably also need to delete the other tables in the project's namespace.

src/Appwrite/Platform/Tasks/DeleteOrphanedProjects.php
```diff
@@ -26,25 +28,23 @@ public function __construct()
 
         $this
             ->desc('Get stats for projects')
```
stnguyen90, line 30: Desc should be updated

src/Appwrite/Platform/Tasks/DeleteOrphanedProjects.php
```diff
@@ -26,25 +28,23 @@ public function __construct()
...
+            ->param('commit', false, new boolean(true), 'Commit  project deletion', true)
```
stnguyen90, line 31: Remove extra space. Also shouldn't boolean be Boolean?

```suggestion
            ->param('commit', false, new Boolean(true), 'Commit project deletion', true)
```

src/Appwrite/Platform/Tasks/DeleteOrphanedProjects.php
```diff
@@ -80,17 +80,22 @@ public function action(Group $pools, Cache $cache, Database $dbForConsole, Regis
                     $dbForProject->setDefaultDatabase('appwrite');
                     $dbForProject->setNamespace('_' . $project->getInternalId());
                     $collectionsCreated = $dbForProject->count(Database::METADATA);
-                    $message = ' (' . $collectionsCreated . ') collections where found on project (' . $project->getId() . '))';
-                    if ($collectionsCreated < (count($collectionsConfig) + 2)) {
-                        Console::error($message);
+                    if ($collectionsCreated === 0) {
+                        if ($commit === true) {
```
stnguyen90, line 84: This can just be $commit only, right?

```suggestion
                        if ($commit) {
```

stnguyen90 (changes requested):

src/Appwrite/Platform/Tasks/DeleteOrphanedProjects.php
shimonewman: Not sure I understand what you mean by "other tables in the project's namespace"
Do you mean the metadata collection if it was created?

christyjacob4: No, its possible that certain tables of the project are still created like _buckets _databases _users. so we've to check if those tables exist and delete them too

christyjacob4 (approved):

PR #7097, opened 2023-11-08
Add support for geo-based SMS messaging

loks0n:
Related issue:

* https://github.com/utopia-php/messaging/pull/64


stnguyen90 (changes requested):
Please fix lint error

app/init.php
```diff
@@ -1329,18 +1332,7 @@ function getDevice($root): Device
...
-        'mock' => new Mock($user, $secret), // used for tests
-        'twilio' => new Twilio($user, $secret),
-        'text-magic' => new TextMagic($user, $secret),
-        'telesign' => new Telesign($user, $secret),
-        'msg91' => new Msg91($user, $secret),
-        'vonage' => new Vonage($user, $secret),
-        default => null
-    };
+    return SMSFactory::createFromDSN($dsn);
 });
```
abnegate, line 1336: Do we actually inject this resource anywhere? If not, let's remove it

loks0n: Appears not. Will remove it

src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -70,21 +68,13 @@ public function action(Message $message): void
...
-            'text-magic' => new TextMagic($this->user, $this->secret),
-            'telesign' => new Telesign($this->user, $this->secret),
-            'msg91' => new Msg91($this->user, $this->secret),
-            'vonage' => new Vonage($this->user, $this->secret),
-            default => null
-        };
-
-        if (empty(App::getEnv('_APP_SMS_PROVIDER'))) {
+
+        if (empty(App::getEnv('_APP_SMS_PROVIDER') && empty(App::getEnv('_APP_GEOSMS_PROVIDERS')))) {
```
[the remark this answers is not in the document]
loks0n: Legacy of previous approach - removed

app/init.php
```diff
@@ -72,6 +66,7 @@
 use Appwrite\Event\Build;
 use Appwrite\Event\Certificate;
 use Appwrite\Event\Func;
+use Appwrite\Messaging\Adapter\SMS\SMSFactory;
```
stnguyen90, line 69: This can be removed too, right?

src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -104,4 +93,66 @@ public function action(Message $message): void
...
+                $adapter = new Mock($dsn->getUser(), $dsn->getPassword());
+                break;
+            case 'msg91':
+                $adapter = new Msg91($dsn->getUser(), $dsn->getPassword());
+                $adapter->setTemplate($dsn->getParam('template', ''));
+                break;
+            case 'telesign':
+                $adapter = new Telesign($dsn->getUser(), $dsn->getPassword());
+                break;
+            case 'textmagic':
```
stnguyen90, line 112: This key has changed. Maybe we can include both for backwards compatibility?

src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -104,4 +93,66 @@ public function action(Message $message): void
...
+    {
+        $adapter = null;
+
+        switch ($dsn->getHost()) {
+            case 'mock':
+                $adapter = new Mock($dsn->getUser(), $dsn->getPassword());
+                break;
+            case 'msg91':
+                $adapter = new Msg91($dsn->getUser(), $dsn->getPassword());
+                $adapter->setTemplate($dsn->getParam('template', ''));
```
stnguyen90, line 107: Previously, `_APP_SMS_FROM ` was used as the Flow ID and passed to Msg91. Maybe we should fallback to that for backwards compatibility?

stnguyen90 (changes requested):

src/Appwrite/Platform/Workers/Messaging.php
loks0n: 👍

loks0n: This poses a larger problem:

- `_APP_SMS_FROM` currently has a single value. GEOSMS local adapters that depend on `_APP_SMS_FROM` will use the same value.
- Local adapters will need different values, as we have different numbers set up with different providers.
- This isn't a problem for MSG91, and maybe some others, because it uses the `username` DSN part as the sender ID/from.

We either need to:
- rethink the DSN approach, 
- update messaging to support `->setFrom()` function on the adapter/or individual providers, and set  `?from=` parameter in the DSN
- leave it for now as we just need MSG91 + Twilio which will work

abnegate: You can set the `from` value per **message** but not adapter, can that work without change in the library if we add it to the DSN?

app/config/variables.php
```diff
@@ -440,7 +440,7 @@
         'variables' => [
             [
                 'name' => '_APP_SMS_PROVIDER',
-                'description' => "Provider used for delivering SMS for Phone authentication. Use the following format: 'sms://[USER]:[SECRET]@[PROVIDER]'.\n\nEnsure `[USER]` and `[SECRET]` are URL encoded if they contain any non-alphanumeric characters.\n\nAvailable providers are twilio, text-magic, telesign, msg91, and vonage.",
+                'description' => "Provider used for delivering SMS for Phone authentication. Use the following format: 'sms://[USER]:[SECRET]@[PROVIDER]'.\n\nEnsure `[USER]` and `[SECRET]` are URL encoded if they contain any non-alphanumeric characters.\n\nAvailable providers are twilio, text-magic, telesign, msg91, vonage and geosms. When using the geosms provider, the adapter will parse `default` query parameter for a nested provider unless a `local-x` parameter is set for a country calling code. `default` and `local-x` parameters should be URL encoded.",
```
abnegate, line 443: This is very confusing to read, any ideas @gewenyu99?

src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -104,4 +93,70 @@ public function action(Message $message): void
...
+            case 'text-magic':
+                return new TextMagic($dsn->getUser(), $dsn->getPassword());
+            case 'twilio':
+                return new Twilio($dsn->getUser(), $dsn->getPassword());
+            case 'vonage':
+                return new Vonage($dsn->getUser(), $dsn->getPassword());
+            case 'geosms':
+                return self::createGEOSMSAdapter($dsn);
+            default:
+                return null;
```
abnegate, line 121: Let's throw here

src/Appwrite/Platform/Workers/Messaging.php
loks0n: Not really without duplicating the geosms adapter logic in the messaging library

src/Appwrite/Platform/Workers/Messaging.php
loks0n: Ok

app/config/variables.php
loks0n: Had another go

src/Appwrite/Platform/Workers/Messaging.php
abnegate: Let's add `from` as a default null constructor parameter to SMS adapters and use that instead of the message from if it's set. Not ideal as we'll overriding a value from a smaller scope with the one from larger scope which is counter-intuitive.. can't see any way around it though.

src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -104,4 +93,71 @@ public function action(Message $message): void
             throw new Exception('Error sending message: ' . $error->getMessage(), 500);
         }
     }
+
+    protected static function createAdapterFromDSN(DSN $dsn): SMSAdapter
+    {
+        $from = empty($dsn->getParam('from', '')) ? null : $dsn->getParam('from', '');
```
abnegate, line 99: Seems odd that you can't use a null default here 🤔 let's update the DSN lib to allow a nullable default, not a blocker for this PR though

abnegate (commented):
Looks good, any way we can add tests for it?

abnegate (approved):

stnguyen90 (changes requested):
merge conflict.

PR #7117, opened 2023-11-10
Fix delete sessions events

loks0n:
Delete sessions should also trigger each individual session - `users.[userId].sessions.[sessionId].delete`


app/controllers/api/users.php
```diff
@@ -1156,11 +1156,19 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+                ->setParam('userId', $user->getId())
+                ->setParam('sessionId', $session->getId())
+                ->setPayload($response->output($session, Response::MODEL_SESSION))
+                ->trigger();
         }
 
         $dbForProject->deleteCachedDocument('users', $user->getId());
 
         $queueForEvents
+            ->setEvent('users.[userId].sessions.delete')
```
abnegate, line 1171: We won't need this here - it will be set from the routes `event` label in [the shutdown hook](https://github.com/appwrite/appwrite/blob/5871178fc0bbc6edfcba6fe39cce3d2f3160a884/app/controllers/shared/api.php#L177)

app/controllers/api/users.php
```diff
@@ -1156,11 +1156,19 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
             /** @var Document $session */
             $dbForProject->deleteDocument('sessions', $session->getId());
             //TODO: fix this
+
+            $queueForEvents
+                ->setEvent('users.[userId].sessions.[sessionId].delete')
+                ->setParam('userId', $user->getId())
+                ->setParam('sessionId', $session->getId())
```
abnegate, line 1163: Do we have any way to remove this param after the loop?

loks0n: Added `->reset()`

app/controllers/api/users.php
loks0n: Removed

christyjacob4 (approved):

app/controllers/api/users.php
```diff
@@ -1156,8 +1156,17 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
             /** @var Document $session */
             $dbForProject->deleteDocument('sessions', $session->getId());
             //TODO: fix this
+
+            $queueForEvents
```
stnguyen90, line 1160: Shouldn't we have this logic in the account service too?

stnguyen90 (changes requested):

abnegate (commented):
Looks like we need to fix `testDeleteAccountSessions` for the account API changes

app/controllers/api/users.php
loks0n: Had to add it back, that hook is run before each request no? `init`

tests/e2e/Services/Webhooks/WebhooksCustomClientTest.php
```diff
@@ -347,6 +347,9 @@ public function testDeleteAccountSessions($data): array
 
         $this->assertEquals($accountSession['headers']['status-code'], 204);
 
+        // This is very flakey :/
+        \sleep(3);
```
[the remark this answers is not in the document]
loks0n: Problem was https://github.com/appwrite/appwrite/pull/7117#discussion_r1400701935

app/controllers/api/users.php
abnegate: Ah my bad, I thought it was in a shutdown hook. It probably should be, it's not really of any use before the route action runs 🤔 that's outside the scope of this PR though

app/controllers/api/account.php
```diff
@@ -2353,13 +2353,22 @@
...
         }
 
         $dbForProject->deleteCachedDocument('users', $user->getId());
 
         $queueForEvents
+            ->reset()
+            ->setEvent('users.[userId].sessions.delete')
             ->setParam('userId', $user->getId())
-            ->setParam('sessionId', $session->getId());
+            ->setPayload($response->output($session, Response::MODEL_SESSION));
```
abnegate, line 2371: This will just be the last session from the loop, based on [this comment](https://github.com/appwrite/appwrite/pull/7117/files#diff-93771e9bb8e57cbe7d0c0e91e7aea5d85238a362f51caf957049f1d2ccc4d312R2353), we should use the `current` session here (which is still kinda weird it should probably be the user model here, but at least we keep the behavior the same as now)

loks0n: Changed

app/controllers/api/account.php
```diff
@@ -2353,13 +2353,22 @@
...
+                ->setParam('sessionId', $session->getId())
+                ->setPayload($response->output($session, Response::MODEL_SESSION))
+                ->trigger();
         }
 
         $dbForProject->deleteCachedDocument('users', $user->getId());
 
         $queueForEvents
+            ->reset()
+            ->setEvent('users.[userId].sessions.delete')
```
stnguyen90, line 2369: Dow e need this event?

abnegate (approved):

app/controllers/api/account.php
eldadfux: @loks0n Maybe we can skip it for now, as we don't allow bulk deleting of other individual resources, but just "containers" like buckets or collections.

app/controllers/api/account.php
```diff
@@ -2353,14 +2352,17 @@
                 // Use current session for events.
                 $queueForEvents->setPayload($response->output($session, Response::MODEL_SESSION));
             }
+
+            $queueForEvents
+                ->setEvent('users.[userId].sessions.[sessionId].delete')
+                ->setParam('userId', $user->getId())
+                ->setParam('sessionId', $session->getId())
+                ->setPayload($response->output($session, Response::MODEL_SESSION))
+                ->trigger();
```
stnguyen90, line 2361: will the last event trigger twice since our shutdown hook will also trigger an event?

PR #7123, opened 2023-11-13
Implement health tresholds

Meldiron:
## What does this PR do?

Adds threshold parameter to queue health APIs. This is useful for uptime monitoring when you want to ensure queue size does not exceed the expected load.

## Test Plan

New tests

## Related PRs and Issues

x

## Checklist

### DIFF
--- a/app/controllers/api/health.php
+++ b/app/controllers/api/health.php
@@ -345,17 +345,17 @@ App::get('/v1/health/queue/webhooks')
     ->label('sdk.response.code', Response::STATUS_CODE_OK)
     ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
     ->label('sdk.response.model', Response::MODEL_HEALTH_QUEUE)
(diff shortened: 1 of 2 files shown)


app/controllers/api/health.php
```diff
@@ -344,11 +345,20 @@
     ->label('sdk.response.code', Response::STATUS_CODE_OK)
     ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
     ->label('sdk.response.model', Response::MODEL_HEALTH_QUEUE)
+    ->param('treshold', 5000, new Integer(true), 'Queue size treshold. When hit (equal or higher), endpoint returns server error. Default value is 5000.', true)
```
eldadfux, line 348:
```suggestion
    ->param('threshold', 5000, new Integer(true), 'Queue size threshold. When hit (equal or higher), endpoint returns 500 server error. Default value is 5000.', true)
```

eldadfux (changes requested):

eldadfux (approved):

PR #7126, opened 2023-11-13
Add new storage health routes

PineappleIOnic:
## What does this PR do?

This adds new health routes that are compatible with external devices from the storage library

## Checklist


christyjacob4:
This is not relevant as these endpoints dont exist. We have 1 endpoint to test this and its located here
https://github.com/appwrite/appwrite/blob/main/app/controllers/api/health.php#L759-L774

And its documentation is available here https://github.com/appwrite/appwrite/blob/main/docs/references/health/get-storage.md

PR #7128, opened 2023-11-14
Limit webhook failure attempts to 10

vermakhushboo:
## What does this PR do?

1. Limit failed webhook attempts - every time the API fails, count the error in a new errors attribute. Log the last error message so we can present it in the API. After 10 successive failures, pause the webhook by changing its status to disabled (show as stopped on console)
2. Allow users to enable or disable a webhook like storage buckets
3. Reset the attempts count whenever the webhook is re-enabled or on every successful request
4. Only allow URLs with public domains

## Test Plan
Tested locally

## Related PRs and Issues

https://app.asana.com/0/1148249895651207/1205842531745480/f
https://linear.app/appwrite/issue/PEA-7/prevent-webhooks-abuse

## Checklist


app/controllers/api/projects.php
```diff
@@ -1052,7 +1052,10 @@
             ->setAttribute('url', $url)
             ->setAttribute('security', $security)
             ->setAttribute('httpUser', $httpUser)
-            ->setAttribute('httpPass', $httpPass);
+            ->setAttribute('httpPass', $httpPass)
+            ->setAttribute('errors', 0)
+            ->setAttribute('status', true)
+            ->setAttribute('logs', '');
```
Meldiron, line 1058: I think we should add `status` and `logs` into response model of webhook, so we can show it in Console.

In console there should be warning icon (or possibly error one) explaining that what's going on. Showing logs can also help developer debug it.
100% we should also provide information on what to do next to resolve this. For example deleting a webhook and adding it back again.

I also think it should auto-re-enable when you do any change to webhook (I think updating is allowed).

Also I think there should be button to re-enable it. You might be having problem with external server that you now fixed, and simply want to re-enable webhook.

app/config/collections.php
```diff
@@ -4471,6 +4471,39 @@
                 'array' => false,
                 'filters' => [],
             ],
+            [
+                '$id' => ID::custom('status'),
```
Meldiron, line 4475: This is what we use in other collections instead:

```
            [
                '$id' => ID::custom('enabled'),
                'type' => Database::VAR_BOOLEAN,
                'signed' => true,
                'size' => 0,
                'format' => '',
                'filters' => [],
                'required' => true,
                'array' => false,
            ],
```

I would rename to `enabled`

app/config/collections.php
```diff
@@ -4471,6 +4471,39 @@
...
+                'format' => '',
+                'size' => 0,
+                'signed' => true,
+                'required' => false,
+                'default' => 0,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('logs'),
```
Meldiron, line 4497: Seing combination of `errors` and `logs` makes me think of `stderr` and `stdout`.

I would rename to `logs` and `attempts` - we call them like this in `certificates` collection

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -113,7 +120,19 @@ private function execute(array $events, string $payload, Document $webhook, Docu
         }
 
         if (false === \curl_exec($ch)) {
-            $this->errors[] = \curl_error($ch) . ' in events ' . implode(', ', $events) . ' for webhook ' . $webhook->getAttribute('name');
+            $errorCount = $webhook->getAttribute('errors', 0) + 1;
+            $lastErrorLogs = \curl_error($ch) . ' in events ' . implode(', ', $events) . ' for webhook ' . $webhook->getAttribute('name');
```
Meldiron, line 124: Webhook name isnt valuable here, as it's stored alongside all other webhook info on it's document.

Alongside curl_error I would also try to store HTTP status code and response body. Body should be truncated to ensure long ones don't cause updateDocument to fail.

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -113,7 +120,19 @@ private function execute(array $events, string $payload, Document $webhook, Docu
...
+            $webhook->setAttribute('errors', $errorCount);
+            $webhook->setAttribute('logs', $lastErrorLogs);
+
+            if ($errorCount > 9) {
```
Meldiron, line 128: 9 here should be const, or env variable. Check with Christy which one he prefers.

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -113,7 +120,19 @@ private function execute(array $events, string $payload, Document $webhook, Docu
...
+                $webhook->setAttribute('status', false);
+            }
+
+            $dbForConsole->updateDocument('webhooks', $webhook->getId(), $webhook);
+            $dbForConsole->deleteCachedDocument('projects', $project->getId());
+
+            $this->errors[] = $lastErrorLogs;
```
Meldiron, line 135: Can we please double check what scenario this errors array is for? It scares me to think about concurrency with this variable, and also I cant see where we clear it to empty array.

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -113,7 +120,19 @@ private function execute(array $events, string $payload, Document $webhook, Docu
         }
 
         if (false === \curl_exec($ch)) {
-            $this->errors[] = \curl_error($ch) . ' in events ' . implode(', ', $events) . ' for webhook ' . $webhook->getAttribute('name');
+            $errorCount = $webhook->getAttribute('errors', 0) + 1;
```
eldadfux, line 123: @vermakhushboo This could suffer from race conditions. We can use the increase method instead: 
https://github.com/utopia-php/database/blob/main/src/Database/Database.php#L3672

app/controllers/api/projects.php
vermakhushboo: The `attempts` count is reset everytime the `update` endpoint is called for Webhooks.

vermakhushboo: Also, who should I assign the frontend changes to?

src/Appwrite/Platform/Workers/Webhooks.php
TeamAppwrite: When I'm adding `$statusCode = curl_getinfo($ch, CURLINFO_RESPONSE_CODE);`, the status code is `200` even when exceptions are thrown. Any idea why that might be happening?

Also, how to fetch response body? In this case `\curl_exec($ch)` is `false`.

vermakhushboo: When I'm adding `$statusCode = curl_getinfo($ch, CURLINFO_RESPONSE_CODE);`, the status code is `200` even when exceptions are thrown. Any idea why that might be happening?

Also, how to fetch response body? In this case `\curl_exec($ch)` is `false`.

app/controllers/api/projects.php
```diff
@@ -1052,7 +1053,10 @@
             ->setAttribute('url', $url)
             ->setAttribute('security', $security)
             ->setAttribute('httpUser', $httpUser)
-            ->setAttribute('httpPass', $httpPass);
+            ->setAttribute('httpPass', $httpPass)
+            ->setAttribute('enabled', true)
+            ->setAttribute('logs', '')
+            ->setAttribute('attempts', 0);
```
christyjacob4, line 1059: Where do you use the param `enable`?

christyjacob4: Screenshot 2023-12-14 at 7 58 09 AM
Also looking at other APIs, we use the word `enabled` for the param name. Let's stay consistent with the name 👍

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -5,12 +5,14 @@
...
 use Utopia\App;
 use Utopia\Database\Document;
+use Utopia\Database\Database;
 use Utopia\Platform\Action;
 use Utopia\Queue\Message;
 
 class Webhooks extends Action
 {
     private array $errors = [];
+    private const FAILEDATTEMPTSCOUNT = 10;
```
christyjacob4, line 15:
```suggestion
    private const MAX_FAILED_ATTEMPTS = 10;
```

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -113,7 +121,21 @@ private function execute(array $events, string $payload, Document $webhook, Docu
         }
 
         if (false === \curl_exec($ch)) {
-            $this->errors[] = \curl_error($ch) . ' in events ' . implode(', ', $events) . ' for webhook ' . $webhook->getAttribute('name');
+            $dbForConsole->increaseDocumentAttribute('webhooks', $webhook->getId(), 'attempts', 1);
+            $webhook = $dbForConsole->getDocument('webhooks', $webhook->getId());
+            $errorCount = $webhook->getAttribute('attempts');
```
christyjacob4, line 126:
```suggestion
            $attempts = $webhook->getAttribute('attempts');
```

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -113,7 +121,21 @@ private function execute(array $events, string $payload, Document $webhook, Docu
...
+            $lastErrorLogs = \curl_error($ch) . ' in events ' . implode(', ', $events);
```
christyjacob4, line 127:
```suggestion
            $logs = \curl_error($ch) . ' in events ' . implode(', ', $events);
```

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -113,7 +121,21 @@ private function execute(array $events, string $payload, Document $webhook, Docu
...
+            $webhook->setAttribute('logs', $lastErrorLogs);
+            // $statusCode = curl_getinfo($ch, CURLINFO_RESPONSE_CODE);
+
+            if ($errorCount >= self::FAILEDATTEMPTSCOUNT) {
```
christyjacob4, line 131:
```suggestion
            if ($attempts >= self:: MAX_FAILED_ATTEMPTS) {
```

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -113,7 +121,21 @@ private function execute(array $events, string $payload, Document $webhook, Docu
...
+                $webhook->setAttribute('enabled', false);
```
christyjacob4, line 132: You should probably also reset the attempts count here.. right ?

app/controllers/api/projects.php
vermakhushboo: It'd be used from the console if the user just wants to `enable` the webhook without updating any other params.

src/Appwrite/Platform/Workers/Webhooks.php
vermakhushboo: It's already done when webhook is re-enabled from `projects.php` line 1059. It'd be redundant to do it here.

app/controllers/api/projects.php
```diff
@@ -898,13 +900,14 @@
     ->param('projectId', '', new UID(), 'Project unique ID.')
     ->param('name', null, new Text(128), 'Webhook name. Max length: 128 chars.')
     ->param('events', null, new ArrayList(new Event(), APP_LIMIT_ARRAY_PARAMS_SIZE), 'Events list. Maximum of ' . APP_LIMIT_ARRAY_PARAMS_SIZE . ' events are allowed.')
-    ->param('url', null, new URL(['http', 'https']), 'Webhook URL.')
+    ->param('url', '', fn ($request) => new Multiple([new URL(['http', 'https']), new PublicDomain(\parse_url($request->getPayload('url'), PHP_URL_HOST))], Multiple::TYPE_STRING), 'Webhook URL.', false, ['request'])
```
Meldiron, line 903: Looking at definiton of PublicDomain, it doesnt seem to accept any params to constructor.

I would suggest:

```php
    ->param('url', '', fn ($request) => new Multiple([new URL(['http', 'https']), new PublicDomain()], Validator::TYPE_STRING), 'Webhook URL.', false, ['request'])
```

Meldiron: Same applied in more places of this PR

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -47,8 +51,8 @@ public function action(Message $message): void
         $user = new Document($payload['user'] ?? []);
 
         foreach ($project->getAttribute('webhooks', []) as $webhook) {
-            if (array_intersect($webhook->getAttribute('events', []), $events)) {
-                $this->execute($events, $webhookPayload, $webhook, $user, $project);
+            if ($webhook->getAttribute('enabled') === true && array_intersect($webhook->getAttribute('events', []), $events)) {
```
Meldiron, line 54: This check is here, and also on line 75. Lets keep it in execute() only. We can remove it from here

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -63,10 +67,14 @@ public function action(Message $message): void
      * @param Document $webhook
      * @param Document $user
      * @param Document $project
+     * @param Database $dbForConsole
      * @return void
      */
-    private function execute(array $events, string $payload, Document $webhook, Document $user, Document $project): void
+    private function execute(array $events, string $payload, Document $webhook, Document $user, Document $project, Database $dbForConsole): void
     {
+        if ($webhook->getAttribute('enabled') === false) {
```
Meldiron, line 75:
```suggestion
        if ($webhook->getAttribute('enabled') !== true) {
```

nulls or undefined or 0 or empty string or other weird stuff that PHP does will also work fine if we do it like this. This will ensure only exactly `true` is allowed.

src/Appwrite/Utopia/Response/Model/Webhook.php
```diff
@@ -76,7 +76,18 @@ public function __construct()
...
                 'example' => 'ad3d581ca230e2b7059c545e5a',
             ])
-        ;
+            ->addRule('enabled', [
+                'type' => self::TYPE_BOOLEAN,
+                'description' => 'Indicates if this webhook is enabled.',
+                'default' => true,
+                'example' => true,
+            ])
+            ->addRule('logs', [
```
Meldiron, line 85: Attempts are not made public. Just making sure, is this on purpose? I don't see downside of allowing developers to see how many attempts were made.

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -113,7 +121,21 @@ private function execute(array $events, string $payload, Document $webhook, Docu
         }
 
         if (false === \curl_exec($ch)) {
-            $this->errors[] = \curl_error($ch) . ' in events ' . implode(', ', $events) . ' for webhook ' . $webhook->getAttribute('name');
+            $dbForConsole->increaseDocumentAttribute('webhooks', $webhook->getId(), 'attempts', 1);
+            $webhook = $dbForConsole->getDocument('webhooks', $webhook->getId());
+            $attempts = $webhook->getAttribute('attempts');
+            $logs = \curl_error($ch) . ' in events ' . implode(', ', $events);
+            $webhook->setAttribute('logs', $logs);
+            // $statusCode = curl_getinfo($ch, CURLINFO_RESPONSE_CODE);
```
Meldiron, line 129: I think logs are nto useful. CURL error is often just a weird number, not indicating too much. I would do longs along those lines:

```
URL: [WEBHOOK_URL]
Method: POST
Response status code: curl_getinfo($ch, CURLINFO_RESPONSE_CODE)
CURL error code: \curl_error($ch)
Events: implode(', ', $events)
Response body:
[RESPONSE_BODY]
```

(For response body, truncate to let's say 100k characters. Take inspiration from functions or builds worker, we do it there already I think)

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -113,7 +121,21 @@ private function execute(array $events, string $payload, Document $webhook, Docu
         }
 
         if (false === \curl_exec($ch)) {
-            $this->errors[] = \curl_error($ch) . ' in events ' . implode(', ', $events) . ' for webhook ' . $webhook->getAttribute('name');
+            $dbForConsole->increaseDocumentAttribute('webhooks', $webhook->getId(), 'attempts', 1);
```
Meldiron, line 124: If error increases attempts by 1, successful request should reset attempts back to 0.

If I have my webhook and it fails 10 times in 1 year after 1 million requests, it should not get blocked.

app/init.php
```diff
@@ -228,6 +229,11 @@
 
 App::setMode(App::getEnv('_APP_ENV', App::MODE_TYPE_PRODUCTION));
 
+$isProduction = App::getEnv('_APP_ENV', App::MODE_TYPE_DEVELOPMENT) === App::MODE_TYPE_PRODUCTION;
```
eldadfux, line 232: You can use this method instead: https://github.com/utopia-php/framework/blob/master/src/App.php#L395, no need for variable or function in most cases when that var/functions are used just once.

src/Appwrite/Platform/Workers/Webhooks.php
vermakhushboo: We do that in the API endpoint when `status` is reset.

app/init.php
```diff
@@ -228,6 +229,10 @@
 
 App::setMode(App::getEnv('_APP_ENV', App::MODE_TYPE_PRODUCTION));
 
+if (!App::isProduction()) {
```
eldadfux, line 232: Add a doc block to explain why we do this.

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -78,7 +86,7 @@ private function execute(array $events, string $payload, Document $webhook, Docu
         \curl_setopt($ch, CURLOPT_CUSTOMREQUEST, 'POST');
         \curl_setopt($ch, CURLOPT_POSTFIELDS, $payload);
         \curl_setopt($ch, CURLOPT_HEADER, 0);
-        \curl_setopt($ch, CURLOPT_RETURNTRANSFER, 0);
+        \curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
         \curl_setopt($ch, CURLOPT_TIMEOUT, 15);
         \curl_setopt($ch, CURLOPT_MAXFILESIZE, 5242880);
```
eldadfux, line 91: Use a constant or/and add the size in human readable format in a comment

eldadfux (changes requested):

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -117,10 +133,72 @@ private function execute(array $events, string $payload, Document $webhook, Docu
...
+
+                $body
+                    ->setParam('{{subject}}', $subject)
+                    ->setParam('{{message}}', $template->render())
+                    ->setParam('{{year}}', date("Y"));
+
+                $queueForMails
+                    ->setSubject($subject)
+                    ->setBody($body->render())
+                    ->setRecipient($user->getAttribute('email'))
```
Meldiron, line 190: User comes from `$user = new Document($payload['user'] ?? []);`.
Who is user in payload of webhook? 🤔 Most logical approach would be to send email to all members of organization, right?

app/controllers/api/projects.php
```diff
@@ -897,14 +899,15 @@
     ->label('sdk.response.model', Response::MODEL_WEBHOOK)
     ->param('projectId', '', new UID(), 'Project unique ID.')
     ->param('name', null, new Text(128), 'Webhook name. Max length: 128 chars.')
+    ->param('enabled', true, new Boolean(), 'Enable or disable a webhook.', true)
```
Meldiron, line 902: Can we have test for this? To mark webhook disabled, and trigger it's event. Then ensure the webhook didn't run (we have request catcher that can help for this)

app/controllers/api/projects.php
```diff
@@ -1020,14 +1024,15 @@
     ->param('projectId', '', new UID(), 'Project unique ID.')
     ->param('webhookId', '', new UID(), 'Webhook unique ID.')
     ->param('name', null, new Text(128), 'Webhook name. Max length: 128 chars.')
+    ->param('enabled', true, new Boolean(), 'Enable or disable a webhook.', true)
```
Meldiron, line 1027: Lets also have test for this, just like above one

app/controllers/api/projects.php
```diff
@@ -897,14 +899,15 @@
     ->label('sdk.response.model', Response::MODEL_WEBHOOK)
     ->param('projectId', '', new UID(), 'Project unique ID.')
     ->param('name', null, new Text(128), 'Webhook name. Max length: 128 chars.')
+    ->param('enabled', true, new Boolean(), 'Enable or disable a webhook.', true)
     ->param('events', null, new ArrayList(new Event(), APP_LIMIT_ARRAY_PARAMS_SIZE), 'Events list. Maximum of ' . APP_LIMIT_ARRAY_PARAMS_SIZE . ' events are allowed.')
-    ->param('url', null, new URL(['http', 'https']), 'Webhook URL.')
+    ->param('url', '', fn ($request) => new Multiple([new URL(['http', 'https']), new PublicDomain()], Multiple::TYPE_STRING), 'Webhook URL.', false, ['request'])
```
Meldiron, line 904: Let's have a failure test where we try to pass non-public domain

app/controllers/api/projects.php
```diff
@@ -1020,14 +1024,15 @@
     ->param('projectId', '', new UID(), 'Project unique ID.')
     ->param('webhookId', '', new UID(), 'Webhook unique ID.')
     ->param('name', null, new Text(128), 'Webhook name. Max length: 128 chars.')
+    ->param('enabled', true, new Boolean(), 'Enable or disable a webhook.', true)
     ->param('events', null, new ArrayList(new Event(), APP_LIMIT_ARRAY_PARAMS_SIZE), 'Events list. Maximum of ' . APP_LIMIT_ARRAY_PARAMS_SIZE . ' events are allowed.')
-    ->param('url', null, new URL(['http', 'https']), 'Webhook URL.')
+    ->param('url', '', fn ($request) => new Multiple([new URL(['http', 'https']), new PublicDomain()], Multiple::TYPE_STRING), 'Webhook URL.', false, ['request'])
```
Meldiron, line 1029: Same, lets have a failure test

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -117,10 +133,87 @@ private function execute(array $events, string $payload, Document $webhook, Docu
...
+                $logs .= 'Status code: ' . $statusCode . "\n";
+                $logs .= 'Body: ' . "\n" . \mb_strcut($responseBody, 0, 10000) . "\n"; // Limit to 10kb
+            }
+
+            $webhook->setAttribute('logs', $logs);
+
+            if ($attempts >= self::MAX_FAILED_ATTEMPTS) {
+                $webhook->setAttribute('enabled', false);
+
+                $memberships = $dbForConsole->find('memberships', [
```
Meldiron, line 163: Let's move all the logic for sending emails into a separate private method like `$this->sendEmailAlert()`. It will keep this code easier to read.

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -117,10 +133,87 @@ private function execute(array $events, string $payload, Document $webhook, Docu
...
+                $logs .= 'Events: ' . implode(', ', $events) . "\n";
+            } else {
+                $logs .= 'Status code: ' . $statusCode . "\n";
+                $logs .= 'Body: ' . "\n" . \mb_strcut($responseBody, 0, 10000) . "\n"; // Limit to 10kb
+            }
+
+            $webhook->setAttribute('logs', $logs);
+
+            if ($attempts >= self::MAX_FAILED_ATTEMPTS) {
+                $webhook->setAttribute('enabled', false);
```
Meldiron, line 161: Can we have test for this too? To make webhook and make it fail 10 times. Then when we get webhook, it will have enabled = false.

docker-compose.yml
```diff
@@ -289,12 +289,19 @@ services:
...
+      - _APP_DB_USER
+      - _APP_DB_PASS
       - _APP_REDIS_HOST
       - _APP_REDIS_PORT
       - _APP_REDIS_USER
       - _APP_REDIS_PASS
       - _APP_LOGGING_PROVIDER
       - _APP_LOGGING_CONFIG
+      - _APP_DOMAIN
+      - _APP_OPTIONS_FORCE_HTTPS
```
christyjacob4, line 304: why is this env variable required here? 
Also make sure to add these variables to the cloud repo

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -117,10 +133,87 @@ private function execute(array $events, string $payload, Document $webhook, Docu
...
+                $template = Template::fromFile(__DIR__ . '/../../../../app/config/locale/templates/email-webhook-failed.tpl');
+
+                $template->setParam('{{webhook}}', $webhook->getAttribute('name'));
+                $template->setParam('{{project}}', $project->getAttribute('name'));
+                $template->setParam('{{url}}', $webhook->getAttribute('url'));
+                $template->setParam('{{error}}', $curlError ??  'The server returned ' . $statusCode . ' status code');
+                $template->setParam('{{redirect}}', $protocol . '://' . $hostname . "/console/project-$projectId/settings/webhooks/$webhookId");
+                $template->setParam('{{attempts}}', $attempts);
+
+                $subject = 'Webhook deliveries have been paused';
```
christyjacob4, line 188: shouldn't this be part of the translation files?

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -117,10 +133,87 @@ private function execute(array $events, string $payload, Document $webhook, Docu
...
+                    ->setParam('{{year}}', date("Y"));
+
+                $queueForMails
+                    ->setSubject($subject)
+                    ->setBody($body->render());
+
+                foreach ($users as $user) {
+                    $queueForMails
+                        ->setVariables(['user' => $user->getAttribute('name', '')])
+                        ->setName($user->getAttribute('name', ''))
```
christyjacob4, line 203: why do we need to set variables and also the name?

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -117,10 +133,87 @@ private function execute(array $events, string $payload, Document $webhook, Docu
...
+            }
+
+            $dbForConsole->updateDocument('webhooks', $webhook->getId(), $webhook);
+            $dbForConsole->deleteCachedDocument('projects', $project->getId());
+
+            $this->errors[] = $logs;
+        } else {
+            $webhook->setAttribute('attempts', 0); // Reset attempts on success
+            $dbForConsole->updateDocument('webhooks', $webhook->getId(), $webhook);
+            $dbForConsole->deleteCachedDocument('projects', $project->getId());
```
christyjacob4, line 216: you should also reset the logs attribute here 👍

src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -117,10 +133,87 @@ private function execute(array $events, string $payload, Document $webhook, Docu
...
+                        ->setName($user->getAttribute('name', ''))
+                        ->setRecipient($user->getAttribute('email'))
+                        ->trigger();
+                }
+            }
+
+            $dbForConsole->updateDocument('webhooks', $webhook->getId(), $webhook);
+            $dbForConsole->deleteCachedDocument('projects', $project->getId());
+
+            $this->errors[] = $logs;
```
christyjacob4, line 212: what is this attribute used for?

src/Appwrite/Utopia/Response/Model/Webhook.php
```diff
@@ -76,7 +76,24 @@ public function __construct()
...
+                'type' => self::TYPE_STRING,
+                'description' => 'Webhooks last failed delivery attempt logs.',
```
christyjacob4, line 87:
```suggestion
                'description' => 'Webhook error logs from the most recent failure.',
```

app/controllers/api/projects.php
christyjacob4: lets also make sure there is strict validation for the Boolean Validator. Boolean(true)

app/controllers/api/projects.php
```diff
@@ -1052,7 +1057,12 @@
             ->setAttribute('url', $url)
             ->setAttribute('security', $security)
             ->setAttribute('httpUser', $httpUser)
-            ->setAttribute('httpPass', $httpPass);
+            ->setAttribute('httpPass', $httpPass)
+            ->setAttribute('enabled', $enabled);
+
+        if ($enabled) {
+            $webhook->setAttribute('attempts', 0);
```
christyjacob4, line 1064: we should also reset the logs attribute to be empty

docker-compose.yml
vermakhushboo: It's required here because we need to add button with link to console in email and to construct the url, we need these variables.

src/Appwrite/Platform/Workers/Webhooks.php
vermakhushboo: We are not using a translation file for this email since it's a complex one and cloud only supports english for now. So all the `body` of the email is in `.tpl` file and `subject` is in this file.

src/Appwrite/Platform/Workers/Webhooks.php
vermakhushboo: Because we needed a way to set `name` along with the `email` of the recipient in a loop. While debugging, we noticed there is a `setName` function that is not being used, so used it to setName as well. So that it looks like `Khushboo (khushboo@appwrite.io)` in to `to` field.

src/Appwrite/Platform/Workers/Webhooks.php
vermakhushboo: @eldadfux mentioned that we should not reset the logs, otherwise, the user will be confused where did the logs suddenly disappear. Let me know whatever is a better product decision.

src/Appwrite/Platform/Workers/Webhooks.php
vermakhushboo: No idea. It was part of the already existing code, so I didn't touch this bit. 😅

Meldiron (approved):

docker-compose.yml
eldadfux, line 304: Could we add this vars as global vars to all templates and then limit the env vars dependency to just the mails worker?

app/config/locale/templates/email-webhook-failed.tpl
```diff
@@ -0,0 +1,20 @@
...
+<p>To restore your webhook's functionality and reset attempts, we suggest to follow the below steps:</p>
+<ol>
+    <li>Examine the logs of both Appwrite Console and your webhook server to identify the issue.</li>
+    <li>Investigate potential network issues and use webhook testing tools to verify expected behaviour.</li>
+    <li>Ensure the webhook endpoint is reachable and configured to accept incoming POST requests.</li>
+    <li>Confirm that the webhook doesn't return error status codes such as 400 or 500.</li>
+</ol>
+<p>After the issue is resolved, please make sure to re-enable the webhook directly through the webhook settings.</p>
+
+<table border="0" cellspacing="0" cellpadding="0" style="padding-top: 10px; padding-bottom: 10px; margin-top: 32px">
```
eldadfux, line 14: Maybe not in this scope, but we really need to find a way to reuse this button style. It makes really bad maintenance design to have it like this, we will not be able to have any two button that look the same if we keep it like this.

vermakhushboo: Created a separate issue to resolve this.

eldadfux (approved):

src/Appwrite/Platform/Workers/Mails.php
```diff
@@ -63,6 +63,11 @@ public function action(Message $message, Registry $register, Log $log): void
         $name = $payload['name'];
         $body = $payload['body'];
 
+        $protocol = App::getEnv('_APP_OPTIONS_FORCE_HTTPS') == 'disabled' ? 'http' : 'https';
+        $hostname = App::getEnv('_APP_DOMAIN');
+
+        $body = str_replace(['{{protocol}}', '{{hostname}}'], [$protocol, $hostname], $body);
```
Meldiron, line 69: Let's instead add it to `variables` array. Code below ensures all placehodlers are replaced with variables in both body, and subject.

app/config/locale/templates/email-webhook-failed.tpl
```diff
@@ -14,7 +14,7 @@
 <table border="0" cellspacing="0" cellpadding="0" style="padding-top: 10px; padding-bottom: 10px; margin-top: 32px">
     <tr>
         <td style="border-radius: 8px; display: block; width: 100%;">
-            <a class="mobile-full-width" rel="noopener" target="_blank" href="{{redirect}}" style="font-size: 14px; font-family: Inter; color: #ffffff; text-decoration: none; background-color: #FD366E; border-radius: 8px; padding: 9px 14px; border: 1px solid #FD366E; display: inline-block; text-align:center; box-sizing: border-box;">Webhook settings</a>
+            <a class="mobile-full-width" rel="noopener" target="_blank" href="{{protocol}}://{{hostname}}{{redirect}}" style="font-size: 14px; font-family: Inter; color: #ffffff; text-decoration: none; background-color: #FD366E; border-radius: 8px; padding: 9px 14px; border: 1px solid #FD366E; display: inline-block; text-align:center; box-sizing: border-box;">Webhook settings</a>
```
Meldiron, line 17: Hmm, should we instead have just 1 param, `{{host}}`, which would be `https://instance.com`?

Meldiron: Reading such placeholder would be easier than protocol and hostname separated.

PR #7131, opened 2023-11-14
Only delete repositories linked to the particular project

vermakhushboo:
## What does this PR do?

The current condition only checked for `functionId`, which could be same for multiple projects, thus resulting in deletion of repositories linked to multiple Appwrite project. This fix ensures that we also filter out the results by `projectInternalId`.

## Test Plan

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Workers/Deletes.php
+++ b/src/Appwrite/Platform/Workers/Deletes.php
@@ -730,6 +730,7 @@ class Deletes extends Action
          */
         Console::info("Deleting VCS repositories and comments linked to function " . $functionId);
         $this->deleteByGroup('repositories', [
+            Query::equal('projectInternalId', [$projectInternalId]),
             Query::equal('resourceInternalId', [$functionInternalId]),
             Query::equal('resourceType', ['function']),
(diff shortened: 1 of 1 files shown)


src/Appwrite/Platform/Workers/Deletes.php
```diff
@@ -730,6 +730,7 @@ private function deleteFunction(Database $dbForConsole, callable $getProjectDB,
          */
         Console::info("Deleting VCS repositories and comments linked to function " . $functionId);
         $this->deleteByGroup('repositories', [
+            Query::equal('projectInternalId', [$projectInternalId]),
```
Meldiron, line 733:
```suggestion
            Query::equal('projectInternalId', [$project->getInternalId()]),
```

Variable is not defined in this scope. We can do exactly what we do here:  https://github.com/appwrite/appwrite/blob/fix-repository-deletion-bug/src/Appwrite/Platform/Workers/Deletes.php#L680

Meldiron (approved):

TorstenDittmann (approved):

PR #7132, opened 2023-11-14
Feat target provider type

fanatic75:
## What does this PR do?

Adds provider type in targets and makes providerId as optional unless type is push

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/account.php
```diff
@@ -2019,6 +2049,27 @@
...
+
+        /**
+        * @var Document $oldTarget
+        */
+        $oldTarget = $user->find('identifier', $oldEmail, 'targets');
+
+        if ($oldTarget !== false && !$oldTarget->isEmpty()) {
+            try {
+                $dbForProject->updateDocument('targets', $oldTarget->getId(), $oldTarget->setAttribute('identifier', $email));
+            } catch (Duplicate) {
```
abnegate, line 2068: Do we need the try/catch? I think duplicate should only be thrown on create

app/controllers/api/messaging.php
```diff
@@ -806,14 +806,21 @@
...
+        if ($enabled === true || $enabled === false) {
+            if ($provider->getAttribute('internal') === true && $enabled === false) {
+                throw new Exception(Exception::PROVIDER_INTERNAL_UPDATE_DISABLED);
+            }
+            if ($provider->getAttribute('internal') === true && $enabled === false) {
+                throw new Exception(Exception::PROVIDER_INTERNAL_UPDATE_DISABLED);
+            }
+
+            $provider->setAttribute('enabled', $enabled);
+        }
```
abnegate, line 822: Let's remove the outer `if` here

app/controllers/api/messaging.php
```diff
@@ -806,14 +806,21 @@
...
             $provider->setAttribute('internal', $internal);
         }
 
+        if ($enabled === true || $enabled === false) {
+            if ($provider->getAttribute('internal') === true && $enabled === false) {
+                throw new Exception(Exception::PROVIDER_INTERNAL_UPDATE_DISABLED);
+            }
+            if ($provider->getAttribute('internal') === true && $enabled === false) {
+                throw new Exception(Exception::PROVIDER_INTERNAL_UPDATE_DISABLED);
+            }
```
abnegate, line 819: Looks like a duplicate here

fanatic75: fixed this

app/controllers/api/account.php
fanatic75: removed this

app/controllers/api/messaging.php
fanatic75: need to keep it

abnegate (approved):

PR #7135, opened 2023-11-14
Remove database methods from deletes worker 

shimonewman:
We've moved the delete operations to the databases worker so these are no longer needed in the deletes worker.

Related PR:

* https://github.com/appwrite/appwrite/pull/6934


stnguyen90 (changes requested):
Lint errors

stnguyen90 (approved):

abnegate (approved):

Issue #6887, opened 2023-10-12
🐛 Bug Report: Adding multiple many to many relationship attributes to the same collection gets stuck at processing status 


nikkyeya:
### 👟 Reproduction steps

When I created multiple relationship attributes within the same collection, I observed that the second attribute became stuck in a 'failed' status, while the other side of the relationship remained in a 'processing' status. 

Also, adding relationship attributes with custom key names within the same collection sometimes failed.
Recording: https://www.awesomescreenshot.com/video/21552043?key=c3c8dde9e73f0bc09a0fdc67499d2bfa

Please

### 👍 Expected behavior

It should display a 'failed' status on both ends and automatically remove the attribute from the list. Alternatively, consider preventing users from adding multiple many-to-many relationship fields within the same collection if full support is not available at the moment.

### 👎 Actual Behavior

I actually was able to add the second many-to-many relationship, but it remained stuck in a 'failed/processing' status. The 'processing' status misled me into thinking I should wait for it to be resolved.

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

Windows

### 🧱 Your Environment

Appwrite ^13.0.0
Client SDK
Cloud Console

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

stnguyen90:
@nikkyeya, thanks for raising this issue! 🙏🏼 

FYI, the problem is a many-to-many relationship attribute creates a junction collection (extra table) with the name:

```php
'_' . $collection->getInternalId() . '_' . $relatedCollection->getInternalId()
```
[source](<https://github.com/utopia-php/database/blob/5f7b05189cfbcc0506090498c580c5765375a00a/src/Database/Database.php#L1600>)

This causes problems if you want multiple relationships on the same collections.

johnschatner:
This happened to me yesterday and the ominous part is that the backend completely stopped working, I couldn't access the data from any documents within a collection that had the "failed" and or "processing" status. I couldn't delete the attribute either. This led to me having to create three seperate collections and add the data all over. The logs were very misleading, it said that the attribute ID didn't exist when trying to delete the attribute, but when trying to re add the same attribute it would say that it already exists.

spencer237:
[...]

PR #7153, opened 2023-11-16
Add twoWayKey checks and multiple many-to-many restrictions

fogelito:
Fix for relations attributes when twoWayKey variable is not passed in API,
We do not want the worker to throw the Error
Until now, it was checked only for twoWay = true

Added a validation to prevent the creation of 2 manyToMany relations between the same collections

https://github.com/appwrite/appwrite/issues/6281


app/controllers/api/databases.php
```diff
@@ -1640,8 +1640,55 @@ function updateAttribute(
         Event $queueForEvents
```
fogelito, line 1640: @abnegate Since these are only relation validations, maybe should be cleaner to leave it here, for example, the loop through the attributes we do not have in the createAttribute function 
What do you think

app/controllers/api/databases.php
```diff
@@ -1640,8 +1640,55 @@ function updateAttribute(
...
     ) {
         $key ??= $relatedCollectionId;
+        $twoWayKeyNull = is_null($twoWayKey);
```
stnguyen90, line 1643: This is never used.

app/controllers/api/databases.php
```diff
@@ -1640,8 +1640,55 @@ function updateAttribute(
...
+                throw new Exception(Exception::ATTRIBUTE_ALREADY_EXISTS);
+            }
+
+            if (
+                $attribute->getAttribute('type') === Database::VAR_RELATIONSHIP &&
+                \strtolower($attribute->getAttribute('options')['twoWayKey']) === \strtolower($twoWayKey) &&
+                $attribute->getAttribute('options')['relatedCollection'] === $relatedCollection->getId()
+            ) {
+                // Currently, we always throw an Exception even when, We do not want to change $twoWayKsy on $twoWayKeyNull
+                throw new Exception(Exception::ATTRIBUTE_ALREADY_EXISTS);
```
stnguyen90, line 1679: Can we provide a clearer error message?

app/controllers/api/databases.php
```diff
@@ -1640,8 +1640,55 @@ function updateAttribute(
...
+            }
+
+            if (
+                $type === Database::RELATION_MANY_TO_MANY &&
+                $attribute->getAttribute('type') === Database::VAR_RELATIONSHIP &&
+                $attribute->getAttribute('options')['relationType'] === Database::RELATION_MANY_TO_MANY &&
+                $attribute->getAttribute('options')['relatedCollection'] === $relatedCollection->getId()
+            ) {
+                throw new Exception(Exception::ATTRIBUTE_ALREADY_EXISTS, 'Creating more than one "manyToMany" relationship on the same collection is currently not permitted.');
+            }
```
stnguyen90, line 1689: Is this supposed to address https://github.com/appwrite/appwrite/issues/6887? If so, i think it should be a separate PR linked to that https://github.com/appwrite/appwrite/issues/6887. We should probably improve these junction tables to:

1. properly allow multiple
2. make sure they work with multiple databases

app/controllers/api/databases.php
fogelito: Yep removed..
In the beginning, we wanted to auto-generate a key we realized that was not a good idea.

app/controllers/api/databases.php
fogelito: Done!

app/controllers/api/databases.php
fogelito: Currently we do not have this option in Utopia , Definitely something to improve in the future.
The issue you mention here sounds like the reason we are restricting this for now.

app/controllers/api/databases.php
```diff
@@ -1640,8 +1640,55 @@ function updateAttribute(
...
+        $relatedCollection = $dbForProject->getCollection('database_' . $databaseDocument->getInternalId() . '_collection_' . $relatedCollectionDocument->getInternalId());
+
+        if ($relatedCollection->isEmpty()) {
+            throw new Exception(Exception::COLLECTION_NOT_FOUND);
+        }
+
+        $attributes = $collection->getAttribute('attributes', []);
+        /** @var Document[] $attributes */
+        foreach ($attributes as $attribute) {
+            if (\strtolower($attribute->getId()) === \strtolower($key)) {
```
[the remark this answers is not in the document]
fogelito: Done!

app/controllers/api/databases.php
```diff
@@ -1640,8 +1640,55 @@ function updateAttribute(
         Event $queueForEvents
     ) {
         $key ??= $relatedCollectionId;
+        $twoWayKeyNull = is_null($twoWayKey);
         $twoWayKey ??= $collectionId;
 
+        $databaseDocument = Authorization::skip(fn() => $dbForProject->getDocument('databases', $databaseId));
```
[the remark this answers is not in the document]
fogelito: Done!

PR #7181, opened 2023-11-20
fix: templates copy dotgitignore

loks0n:
Fixes hidden files not getting copied when creating a function from a template.


src/Appwrite/Platform/Workers/Builds.php
```diff
@@ -242,7 +242,7 @@ protected function buildDeployment(callable $getFunctionsDevice, Func $queueForF
                     Console::execute('mkdir -p ' . $tmpDirectory . '/' . $rootDirectory, '', $stdout, $stderr);
 
                     // Merge template into user repo
-                    Console::execute('cp -rfn ' . $tmpTemplateDirectory . '/' . $templateRootDirectory . '/* ' . $tmpDirectory . '/' . $rootDirectory, '', $stdout, $stderr);
+                    Console::execute('cp -rfn ' . $tmpTemplateDirectory . '/' . $templateRootDirectory . '/. ' . $tmpDirectory . '/' . $rootDirectory, '', $stdout, $stderr);
```
loks0n, line 245: This should be sufficient.
We don't need to worry about copy .git or unwanted hidden files because we control the templates.

vermakhushboo (approved):
The change looks good. Did you verify it works as expected by testing locally?

loks0n:
> The change looks good. Did you verify it works as expected by testing locally?

I've tested the latest commit 👍

src/Appwrite/Platform/Workers/Builds.php
```diff
@@ -242,7 +242,7 @@ protected function buildDeployment(callable $getFunctionsDevice, Func $queueForF
                     Console::execute('mkdir -p ' . $tmpDirectory . '/' . $rootDirectory, '', $stdout, $stderr);
 
                     // Merge template into user repo
-                    Console::execute('cp -rfn ' . $tmpTemplateDirectory . '/' . $templateRootDirectory . '/* ' . $tmpDirectory . '/' . $rootDirectory, '', $stdout, $stderr);
+                    Console::execute('rsync -av --exclude \'.git\' ' . $tmpTemplateDirectory . '/' . $templateRootDirectory . '/ ' . $tmpDirectory . '/' . $rootDirectory, '', $stdout, $stderr);
```
loks0n, line 245: We have to use rsync to copy all files, including hidden, but exclude .git else there is problems related to sparse checkout config set by the clone command.

Meldiron (approved):

Issue #5595, opened 2023-05-29
🐛 Bug Report: Adding enums with length greater than before results in server error


nimit2801:
### 👟 Reproduction steps

Make a collection in a database for and an Enum Attribute
`paymentStatus` Attrib has
- 3 Enums: `Success` - `Pending` -  `Failure`
- if I edit this attribute and add `Intiated` this won't work, because `mariaDB` has already taken the highest length of `Pending`, `Success` and `Failure` that is 7

### 👍 Expected behavior

It should just add new data to the enum without an error

### 👎 Actual Behavior

This is the error it throws in the `Appwrite container logs`
![image](https://github.com/appwrite/appwrite/assets/53462249/6ceb0db3-1b27-41f7-9a48-738301fcbb95)

This is the error it throws on Client or on server while trying to add `Initiated` Enum to the attribute field

![image](https://github.com/appwrite/appwrite/assets/53462249/15c3249a-3764-4f6b-8059-41be1b58f13c)

### 🎲 Appwrite version

Version 1.3.x

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

fliitor:
i can reproduce. Set varchar(50) (or text..) on enum fields ?

joeyouss:
Hi @nimit2801 , is this fixed? If not, can you let me know?

nimit2801:
> Hi @nimit2801 , is this fixed? If not, can you let me know?

Will have to check on the newer versions but according to my discussion with @stnguyen90 he said to post this issue for future reference ig. The solution was still not found.

joeyouss:
Let me know if you check for newer versions

fliitor:
It's because of [this code](https://github.com/appwrite/appwrite/blob/9f2bf844c7acc4932ca377941061665cd9b96645/app/controllers/api/databases.php#L1208)

Possibilities

1. Fixed size (255 for ex)
2. Round up to 25 more
3. Add `size` input on form to let user decide

safwanyp:
Can I work on this 👀

joeyouss:
Sure @safwanyp 

The approach we want to take with issue as of now would be the fixed size one as stated by @fliitor : ) Assigned the issue to you.

[...]

PR #7183, opened 2023-11-20
migrates enum attribute size to 255

fanatic75:
## What does this PR do?

Migrates enum attribute size to 255 for bug fix https://github.com/appwrite/appwrite/issues/5595

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/init.php
+++ b/app/init.php
@@ -767,6 +767,25 @@ $register->set('pools', function () {
     return $group;
 });
 
+$register->set('db', function () {
+    // This is usually for our workers or CLI commands scope
+       $dbHost = App::getEnv('_APP_DB_HOST', '');
+       $dbPort = App::getEnv('_APP_DB_PORT', '');
+       $dbUser = App::getEnv('_APP_DB_USER', '');
+       $dbPass = App::getEnv('_APP_DB_PASS', '');
+       $dbScheme = App::getEnv('_APP_DB_SCHEMA', '');
+
+       $pdo = new PDO("mysql:host={$dbHost};port={$dbPort};dbname={$dbScheme};charset=utf8mb4", $dbUser, $dbPass, array(
+           PDO::ATTR_TIMEOUT => 3, // Seconds
+           PDO::ATTR_PERSISTENT => true,
+           PDO::ATTR_DEFAULT_FETCH_MODE => PDO::FETCH_ASSOC,
+           PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION,
+           PDO::ATTR_EMULATE_PREPARES => true,
+           PDO::ATTR_STRINGIFY_FETCHES => true,
+       ));
+
+       return $pdo;
+});
 $register->set('influxdb', function () {
 
  // Register DB connection
--- a/src/Appwrite/Migration/Migration.php
+++ b/src/Appwrite/Migration/Migration.php
@@ -195,17 +195,23 @@ abstract class Migration
      * @return iterable<Document>
      * @throws \Exception
      */
-    public function documentsIterator(string $collectionId): iterable
+    public function documentsIterator(string $collectionId, $queries = []): iterable
     {
         $sum = 0;
         $nextDocument = null;
         $collectionCount = $this->projectDB->count($collectionId);
+        $queries[] = Query::limit($this->limit);
 
         do {
-            $queries = [Query::limit($this->limit)];
             if ($nextDocument !== null) {
-                $queries[] = Query::cursorAfter($nextDocument);
+                $lastQuery = $queries[\count($queries) - 1];
+                if ($lastQuery->getMethod() === 'cursorAfter') {
+                    $queries[\count($queries) - 1] = Query::cursorAfter($nextDocument);
+                } else {
+                    $queries[] = Query::cursorAfter($nextDocument);
+                }
             }
+
             $documents = $this->projectDB->find($collectionId, $queries);
             $count = count($documents);
             $sum += $count;
--- a/src/Appwrite/Migration/Migration.php
+++ b/src/Appwrite/Migration/Migration.php
@@ -204,9 +204,10 @@ abstract class Migration
 
         do {
             if ($nextDocument !== null) {
-                $lastQuery = $queries[\count($queries) - 1];
-                if ($lastQuery->getMethod() === 'cursorAfter') {
-                    $queries[\count($queries) - 1] = Query::cursorAfter($nextDocument);
+                $cursorQueryIndex = \array_search('cursorAfter', \array_map(fn (Query $query) => $query->getMethod(), $queries));
+
+                if ($cursorQueryIndex !== false) {
+                    $queries[$cursorQueryIndex] = Query::cursorAfter($nextDocument);
                 } else {
                     $queries[] = Query::cursorAfter($nextDocument);
                 }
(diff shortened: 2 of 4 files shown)


app/init.php
```diff
@@ -767,6 +767,25 @@ function (mixed $value) {
...
+           PDO::ATTR_TIMEOUT => 3, // Seconds
+           PDO::ATTR_PERSISTENT => true,
+           PDO::ATTR_DEFAULT_FETCH_MODE => PDO::FETCH_ASSOC,
+           PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION,
+           PDO::ATTR_EMULATE_PREPARES => true,
+           PDO::ATTR_STRINGIFY_FETCHES => true,
+       ));
+
+       return $pdo;
+});
```
christyjacob4, line 788: why do we need this? you can use the db connection from the database pool directly

src/Appwrite/Migration/Version/V19.php
```diff
@@ -640,6 +646,21 @@ private function getFunctionCommands(Document $function): string
...
+    private function migrateEnumAttributeSize(): void
+    {
+        \Co\run(function () {
+            foreach ($this->documentsIterator('attributes') as $attribute) {
+                go(function ($attribute) {
+                    $attribute->setAttribute('size', Database::LENGTH_KEY);
+                    $this->projectDB->updateDocument('attributes', $attribute->getId(), $attribute);
+                    $databaseInternalId = $attribute->getAttribute('databaseInternalId');
+                    $collectionInternalId = $attribute->getAttribute('collectionInternalId');
+                    $this->changeAttributeInternalType('database_' . $databaseInternalId . '_collection_' . $collectionInternalId, $attribute->getAttribute('key'), 'varchar(255)');
```
stnguyen90, line 658: It would be good to avoid adapter specific things when possible. Can we use [`$this->projectDB->updateAttribute()`](https://github.com/utopia-php/database/blob/60a61d00570bb66e2277be9c88b947279c729623/src/Database/Database.php#L1310) instead?

src/Appwrite/Migration/Version/V19.php
```diff
@@ -640,6 +646,21 @@ private function getFunctionCommands(Document $function): string
         return $commands;
     }
 
+    private function migrateEnumAttributeSize(): void
+    {
+        \Co\run(function () {
+            foreach ($this->documentsIterator('attributes') as $attribute) {
+                go(function ($attribute) {
```
stnguyen90, line 653: Considering we have to update the collection record in the metadata table, it might be dangerous to do the updates concurrently.

src/Appwrite/Migration/Version/V19.php
```diff
@@ -640,6 +646,21 @@ private function getFunctionCommands(Document $function): string
...
+                    $attribute->setAttribute('size', Database::LENGTH_KEY);
```
stnguyen90, line 654: We should check the current size and make sure we're not making it smaller than what it was before.

app/init.php
fanatic75: We are missing setting up PDO in migrate file. This is required for using PDO during migrations

src/Appwrite/Migration/Version/V19.php
fanatic75: We are not updating metadata table here. We are updating attributes table and collection table

src/Appwrite/Migration/Version/V19.php
fanatic75: what is the adapter specific thing here?

src/Appwrite/Migration/Version/V19.php
fanatic75: removed co routines.

src/Appwrite/Migration/Version/V19.php
fanatic75: Added condition for checking current size, although something else will be needed to solve the bug. I think we should truncate the value and update the value to maximum 255 length size. 
cc @abnegate

src/Appwrite/Migration/Version/V19.php
fanatic75: Changed to updateAttribute

wess (approved):

app/init.php
abnegate: We can remove this now we are using `updateAttribute`, we should try keep moving towards using only the db lib functions :+1:

src/Appwrite/Platform/Tasks/Migrate.php
```diff
@@ -29,7 +30,8 @@ public function __construct()
             ->inject('cache')
             ->inject('dbForConsole')
             ->inject('getProjectDB')
-            ->callback(fn ($version, $cache, $dbForConsole, $getProjectDB) => $this->action($version, $cache, $dbForConsole, $getProjectDB));
+            ->inject('register')
```
abnegate, line 33: We can revert the changes here

src/Appwrite/Migration/Version/V19.php
abnegate: Actually I think it should be okay to ignore ones that are longer than 255 already 🤔 any developers in that scenario just won't be able to add any _more_ that are longer than 255

fanatic75: They will be able to update those enum attributes with length greater than 255. Update endpoint doesn't restrict size.

app/init.php
fanatic75: keeping it here @abnegate  for migration files that require PDO

src/Appwrite/Migration/Version/V19.php
fanatic75: Okay. resolving this then

src/Appwrite/Platform/Tasks/Migrate.php
fanatic75: Keeping it as it is required for setting PDO

src/Appwrite/Migration/Migration.php
```diff
@@ -195,17 +195,23 @@ public function forEachDocument(callable $callback): void
...
         $sum = 0;
         $nextDocument = null;
         $collectionCount = $this->projectDB->count($collectionId);
+        $queries[] = Query::limit($this->limit);
 
         do {
-            $queries = [Query::limit($this->limit)];
             if ($nextDocument !== null) {
-                $queries[] = Query::cursorAfter($nextDocument);
+                $lastQuery = $queries[\count($queries) - 1];
```
abnegate, line 207: What if the `cursorAfter` is not the last query passed but the first of multiple?

src/Appwrite/Migration/Version/V19.php
```diff
@@ -640,6 +646,17 @@ private function getFunctionCommands(Document $function): string
         return $commands;
     }
 
+    private function migrateEnumAttributeSize(): void
+    {
+        foreach ($this->documentsIterator('attributes', [Query::equal('format', ['enum']), Query::lessThan('size', Database::LENGTH_KEY)]) as $attribute) {
```
abnegate, line 651: Let's line break the queries for readability:

```suggestion
        foreach ($this->documentsIterator('attributes', [
            Query::equal('format', ['enum']), 
            Query::lessThan('size', Database::LENGTH_KEY)
        ]) as $attribute) {
```

abnegate (commented):
Looking good, please run a manual test to make sure it's working as expected too

abnegate (approved):

PR #7188, opened 2023-11-21
changes TextMagic to Textmagic in all places and uses email validator

fanatic75:
## What does this PR do?

Text magic and email validator PR 

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/users.php
```diff
@@ -415,6 +415,18 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+            $validator = new Email();
+            if (!$validator->isValid($identifier)) {
+                throw new Exception(Exception::GENERAL_INVALID_EMAIL);
+            }
+        } elseif ($providerType === 'sms') {
+            $validator = new Phone();
+            if (!$validator->isValid($identifier)) {
+                throw new Exception(Exception::GENERAL_INVALID_PHONE);
+            }
+        }
```
abnegate, line 428: Let's use `switch`es instead for these checks, easier to add more types in future without it getting messy

app/controllers/api/users.php
```diff
@@ -415,6 +415,23 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+                }
+                break;
+            case 'sms':
+                $validator = new Phone();
+                if (!$validator->isValid($identifier)) {
+                    throw new Exception(Exception::GENERAL_INVALID_PHONE);
+                }
+                break;
+            default:
+                throw new Exception(Exception::PROVIDER_INCORRECT_TYPE);
```
abnegate, line 432: We don't want to throw for `push`, let's just have an empty case for it for now

app/controllers/api/users.php
fanatic75: added case for push

app/controllers/api/users.php
fanatic75: added case for push

abnegate (approved):

PR #7195, opened 2023-11-22
makes provider creation fields optional and adds target info in subsc…

fanatic75:
…riber model

## What does this PR do?

Makes provider required field to optional when creating provider. 

Adds target identifier and providerType to subscriber response model 

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/messaging.php
```diff
@@ -1726,8 +2020,11 @@
         $subscribers = batch(\array_map(function (Document $subscriber) use ($dbForProject) {
             return function () use ($subscriber, $dbForProject) {
                 $user = Authorization::skip(fn () => $dbForProject->getDocument('users', $subscriber->getAttribute('userId')));
+                $target = Authorization::skip(fn () => $dbForProject->getDocument('targets', $subscriber->getAttribute('targetId')));
 
                 return $subscriber
+                    ->setAttribute('identifier', $target->getAttribute('identifier'))
```
abnegate, line 2026: Let's just add the entire `target` as an attribute here since we'll need the `name` too, and that way any future attributes we need are automatically added

fanatic75: done this

.env
```diff
@@ -87,7 +87,7 @@ _APP_LOGGING_PROVIDER=
 _APP_LOGGING_CONFIG=
 _APP_GRAPHQL_MAX_BATCH_SIZE=10
 _APP_GRAPHQL_MAX_COMPLEXITY=250
-_APP_GRAPHQL_MAX_DEPTH=3
+_APP_GRAPHQL_MAX_DEPTH=4
```
abnegate, line 90: Why this change? The updated tests should only be 3 levels 🤔

abnegate (approved):

.env
fanatic75: There's some counting error when calculating depth in GraphQL;

abnegate (approved):

PR #7206, opened 2023-11-27
Shared tables support

abnegate:
## What does this PR do?

Enables table sharing randomly for 1 in 20 projects

## Test Plan

Added tests + manual tests

## Related PRs and Issues

- https://github.com/utopia-php/database/pull/358

## Checklist


Bankrollshow (approved):

composer.json
```diff
@@ -43,13 +43,13 @@
...
-        "utopia-php/abuse": "0.33.*",
+        "utopia-php/abuse": "dev-feat-isolation-modes as 0.33.0",
         "utopia-php/analytics": "0.10.*",
-        "utopia-php/audit": "0.35.*",
+        "utopia-php/audit": "dev-feat-isolation-modes as 0.35.0",
         "utopia-php/cache": "0.8.*",
         "utopia-php/cli": "0.15.*",
         "utopia-php/config": "0.2.*",
-        "utopia-php/database": "0.45.*",
+        "utopia-php/database": "dev-feat-isolation-mode as 0.45.2",
```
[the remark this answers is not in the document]
Bankrollshow: - [ ]

app/controllers/api/projects.php
```diff
@@ -133,6 +141,14 @@
             throw new Exception(Exception::PROJECT_RESERVED_PROJECT, "'console' is a reserved project.");
         }
 
+        // One in 20 projects use shared tables
+        $shareTables = !\mt_rand(0, 19);
+
+        // Allow overriding in development mode
+        if (App::isDevelopment()) {
+            $shareTables = (bool) $request->getHeader('x-appwrite-share-tables', false);
```
eldadfux, line 149: This should probably be only temporary.

eldadfux (approved):

app/config/collections.php
```diff
@@ -3268,6 +3268,17 @@
                 'array' => false,
                 'filters' => [],
             ],
+            [
+                '$id' => ID::custom('shareTables'),
```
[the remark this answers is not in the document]
abnegate: Added a comment

app/controllers/api/projects.php
abnegate: Added a comment here as well

Bankrollshow (approved):
```

```

app/init.php
```diff
@@ -1266,21 +1281,14 @@ function getDevice($root): Device
...
-                return new Local($root);
-        }
+        return match ($device) {
+            Storage::DEVICE_S3 => new S3($root, $accessKey, $accessSecret, $bucket, $region, $acl),
+            Storage::DEVICE_DO_SPACES => new DOSpaces($root, $accessKey, $accessSecret, $bucket, $region, $acl),
+            Storage::DEVICE_BACKBLAZE => new Backblaze($root, $accessKey, $accessSecret, $bucket, $region, $acl),
+            Storage::DEVICE_LINODE => new Linode($root, $accessKey, $accessSecret, $bucket, $region, $acl),
+            Storage::DEVICE_WASABI => new Wasabi($root, $accessKey, $accessSecret, $bucket, $region, $acl),
+            default => new Local($root),
+        };
```
christyjacob4, line 1291: Im sure we use this block of code in a few other places too. Can we check those and fix them as well ?

app/init.php
```diff
@@ -1172,22 +1183,35 @@ function (mixed $value) {
...
         $databaseName = $project->getAttribute('database');
 
-        if (isset($databases[$databaseName])) {
-            $database = $databases[$databaseName];
-
+        $configure = (function (Database $database) use ($project) {
             $database
-                ->setNamespace('_' . $project->getInternalId())
                 ->setMetadata('host', \gethostname())
                 ->setMetadata('project', $project->getId())
```
christyjacob4, line 1196: I see we dont have these 2 metadata attributes in the workers. Is that expected? What are they for ?

app/controllers/api/projects.php
```diff
@@ -188,33 +217,19 @@
...
-                    'default' => $attribute['default'] ?? null,
-                    'format' => $attribute['format'] ?? ''
-                ]);
-            }
+            $attributes = \array_map(function (array $attribute) {
+                return new Document($attribute);
+            }, $collection['attributes']);
+
+            $indexes = \array_map(function (array $index) {
+                return new Document($index);
```
christyjacob4, line 225: are `default` and `format` handled appropriately when using the `Document()` constructor directly ?

app/controllers/api/projects.php
```diff
@@ -133,6 +141,14 @@
             throw new Exception(Exception::PROJECT_RESERVED_PROJECT, "'console' is a reserved project.");
         }
 
+        // One in 20 projects use shared tables
+        $shareTables = !\mt_rand(0, 19);
```
christyjacob4, line 145: Im not sure how this works? This will generate a random number between 0 to 19. 
Which means there is a 1 / 20 chance of it returning 0, which means it will create 19 projects on shared tables, contrary to what is mentioned in the comment ?

christyjacob4 (approved):
Looks good. please address the comments 👍

app/init.php
abnegate: Added now :+1: these values are added as comments in the SQL query, purely for debugging

app/init.php
abnegate: Updated

app/controllers/api/projects.php
abnegate: The `!` is inverting it so only when it returns 0 will it create the project on shared tables

app/controllers/api/projects.php
abnegate: Seems like it, the default check is redundant, it's coalescing to null if the value is null, and can't see anywhere it would matter if `format` is null instead of empty. All tests are passing too

app/controllers/api/projects.php
```diff
@@ -133,6 +141,23 @@
             throw new Exception(Exception::PROJECT_RESERVED_PROJECT, "'console' is a reserved project.");
         }
 
+        // TODO: One in 20 projects use shared tables. Temporary until all projects are using shared tables.
```
eldadfux, line 144: Maybe we can go simpler and after this is ready, just create a single new DB that uses shared tables?

app/init.php
```diff
@@ -1139,20 +1131,30 @@ function (mixed $value) {
...
+            ->setShareTables(true)
+            ->setTenant($project->getInternalId())
+            ->setNamespace('');
+    } else {
+        $database
+            ->setShareTables(false)
+            ->setTenant(null)
+            ->setNamespace('_' . $project->getInternalId());
+    }
+
```
christyjacob4, line 1149: @abnegate the value of `project->getAttribute('database')` is currently one of `db_fra1_v14x_02`, `db_fra1_v14x_03`, `db_fra1_v14x_04` etc. This allows us to allow new projects to be created on a new DB based on our needs and also gives us flexibility to route new projects to a new database in case there is some sort of a corruption. 

However with the introduction of the variable `DATABASE_SHARED_TABLES`, the `database` attribute can only be set to one specific database name which is (database_db_self_hosted_shared_tables). Which means we can only set it to one database. 

This can be problematic as we would be forced to use this one database forever?

abnegate: The `DATABASE_SHARED_TABLES` const will be temporary until we move all projects to use shared tables

docker-compose.yml
```diff
@@ -94,6 +92,7 @@ services:
       - app/http.php
     environment:
       - _APP_ENV
+      - _APP_EDITION
```
christyjacob4, line 95: lets make sure we also add this variable in the cloud repo

tests/e2e/Services/Projects/ProjectsConsoleClientTest.php
```diff
@@ -3461,4 +3462,498 @@ public function testDeleteProject(): array
...
+        $collection5 = $this->client->call(Client::METHOD_POST, '/databases/' . $database2['body']['$id'] . '/collections', [
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $project2['body']['$id'],
+            'x-appwrite-key' => $key2['body']['secret']
+        ], [
+            'databaseId' => $database2['body']['$id'],
+            'collectionId' => $collection4['body']['$id'],
+            'name' => 'Amazing Collection',
+        ]);
+
```
christyjacob4, line 3931: you're missing an assertion here

tests/e2e/Services/Projects/ProjectsConsoleClientTest.php
```diff
@@ -3461,4 +3462,498 @@ public function testDeleteProject(): array
...
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $project4['body']['$id'],
+            'x-appwrite-key' => $key4['body']['secret']
+        ], [
+            'userId' => 'user',
+            'email' => 'test@appwrite.io',
+            'password' => 'password',
+            'name' => 'Test User',
+        ]);
+
```
christyjacob4, line 3956: Aren't there a lot more resources we should check for? Or is it sufficient to just check for the users table and database table?

docker-compose.yml
abnegate: Yep done :+1:

tests/e2e/Services/Projects/ProjectsConsoleClientTest.php
abnegate: It's enough to check a per-project collection and a shared collection, since those are the only types of collections, any issues would present

tests/e2e/Services/Projects/ProjectsConsoleClientTest.php
abnegate: Added

app/config/collections.php
```diff
@@ -5833,14 +5833,14 @@
...
-                'lengths' => [768],
+                'lengths' => [256],
                 'orders' => [Database::ORDER_ASC],
             ],
             [
                 '$id' => ID::custom('_key_signature'),
                 'type' => Database::INDEX_KEY,
                 'attributes' => ['signature'],
-                'lengths' => [768],
+                'lengths' => [256],
```
christyjacob4, line 5843: How do we plan to migrate these indexes on cloud?

app/controllers/api/projects.php
```diff
@@ -130,18 +140,50 @@
             }
         }
 
-        $databaseOverride = System::getEnv('_APP_DATABASE_OVERRIDE', null);
-        $index = array_search($databaseOverride, $databases);
+        $databaseOverride = App::getEnv('_APP_DATABASE_OVERRIDE');
```
christyjacob4, line 143: I think we are using the `System::getEnv` function now 🤔  Ive seen it in use in other places too

christyjacob4, line 143: reminder to also change the value of this variable on cloud

app/controllers/api/projects.php
```diff
@@ -130,18 +140,50 @@
...
             throw new Exception(Exception::PROJECT_RESERVED_PROJECT, "'console' is a reserved project.");
         }
 
+        // TODO: One in 20 projects use shared tables. Temporary until all projects are using shared tables.
+        if (
+            !\mt_rand(0, 19)
+            && App::getEnv('_APP_DATABASE_SHARED_TABLES', 'enabled') === 'enabled'
+            && App::getEnv('_APP_EDITION', 'self-hosted') !== 'self-hosted'
+        ) {
+            $schema = 'appwrite';
```
christyjacob4, line 161: why is the schema `appwrite` ? 

Firstly shouldn't it be `scheme` ? and shouldn't the scheme be `mysql` ?

app/config/collections.php
abnegate: Drop and recreate, and alter the length in metadata

app/controllers/api/projects.php
abnegate: Yep done

app/controllers/api/projects.php
abnegate: The intention was to use it as the DB schema so we could support it for other engines that allow schema, mainly Postgres. In that case, hardcoding mysql wouldn't work well

christyjacob4 (approved):

PR #7207, opened 2023-11-27
Feat: Improve DB override

Meldiron:
## What does this PR do?

Allows multiple values in DB override. This is useful when you want to keep random behaviour (just like empty env var), but want to have allow-list approach (not use all existing connections as possibility).

I noticed this is useful when doing migration QA.

## Test Plan

- [ ] Maual QA

## Related PRs and Issues

x

## Checklist


Meldiron:
Not needed anymore

PR #7217, opened 2023-11-28
adds target when creating user via server endpoint

fanatic75:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -149,18 +149,20 @@ App::post('/v1/account')
             ]);
             $user->removeAttribute('$internalId');
             $user = Authorization::skip(fn() => $dbForProject->createDocument('users', $user));
-            $target = Authorization::skip(fn() => $dbForProject->createDocument('targets', new Document([
-                '$permissions' => [
-                    Permission::read(Role::any()),
-                    Permission::update(Role::user($userId)),
-                    Permission::delete(Role::user($userId)),
-                ],
-                'userId' => $user->getId(),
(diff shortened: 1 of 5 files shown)


app/controllers/api/account.php
```diff
@@ -2102,25 +2119,25 @@
...
         }
 
-        $target = $dbForProject->findOne('targets', [
+        $target = Authorization::skip(fn () => $dbForProject->findOne('targets', [
             Query::equal('identifier', [$email]),
-        ]);
+        ]));
 
-        if ($target && !$target->isEmpty()) {
+        if ($target instanceof Document && !$target->isEmpty()) {
```
abnegate, line 2126: Do we need this change? Should be good enough to check if it's truthy right?

app/controllers/api/users.php
```diff
@@ -1119,6 +1157,16 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
             throw new Exception(Exception::USER_EMAIL_ALREADY_EXISTS);
         }
 
+        $target = $dbForProject->findOne('targets', [
+            Query::equal('identifier', [$email]),
+        ]);
+
+        if ($target && !$target->isEmpty()) {
```
abnegate, line 1164: Let's keep the checks the same for all

fanatic75: fixed this and in one other place

abnegate (approved):

PR #7252, opened 2023-12-06
JSON + OR query support

fogelito:
Changing original query parsing to Json format


abnegate (approved):
Looks good, will wait until the SDK changes are done too otherwise it will break the console

abnegate (approved):

PR #7261, opened 2023-12-08
mail support string as attachment

lohanidamodar:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


TorstenDittmann (commented):
This limits us to a single attachment right?

Would it make sense to already consider multiple?

lohanidamodar:
> This limits us to a single attachment right?
> 
> Would it make sense to already consider multiple?

I don't yet have a use case for multiple attachments, for now it will just add complexity. There is also support to add file from path as attachment, however I don't have a use case for that either not yet

TorstenDittmann (approved):

Issue #7223, opened 2023-11-30
🚀 Feature: Update the install command to allow not starting Appwrite


stnguyen90:
### 🔖 Feature description

It would be nice to be able to run the install command to generate the files, but not start Appwrite.

### 🎤 Pitch

This is useful for cases where a developer only wants the files.

For the core team, it can also be used to generate the files for the website.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

zdanl:
Yes that.

PR #7271, opened 2023-12-11
Add a flag to install and upgrade commands to not start Appwrite

stnguyen90:
This can be useful for cases where the developer only wants the files to be generated and doesn't want to start Appwrite.

## What does this PR do?

Closes https://github.com/appwrite/appwrite/issues/7223

## Test Plan

Manually ran install:

image

Manually ran upgrade:

image

## Related PRs and Issues

- #7223

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Tasks/Install.php
+++ b/src/Appwrite/Platform/Tasks/Install.php
@@ -25,12 +25,12 @@ class Install extends Action
     {
         $this
             ->desc('Install Appwrite')
-            ->param('httpPort', '', new Text(4), 'Server HTTP port', true)
-            ->param('httpsPort', '', new Text(4), 'Server HTTPS port', true)
+            ->param('http-port', '', new Text(4), 'Server HTTP port', true)
+            ->param('https-port', '', new Text(4), 'Server HTTPS port', true)
             ->param('organization', 'appwrite', new Text(0), 'Docker Registry organization', true)
             ->param('image', 'appwrite', new Text(0), 'Main appwrite docker image', true)
             ->param('interactive', 'Y', new Text(1), 'Run an interactive session', true)
-            ->param('noStart', false, new Boolean(true), 'Run an interactive session', true)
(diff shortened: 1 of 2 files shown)


src/Appwrite/Platform/Tasks/Install.php
```diff
@@ -29,10 +30,11 @@ public function __construct()
             ->param('organization', 'appwrite', new Text(0), 'Docker Registry organization', true)
             ->param('image', 'appwrite', new Text(0), 'Main appwrite docker image', true)
             ->param('interactive', 'Y', new Text(1), 'Run an interactive session', true)
-            ->callback(fn ($httpPort, $httpsPort, $organization, $image, $interactive) => $this->action($httpPort, $httpsPort, $organization, $image, $interactive));
+            ->param('noStart', false, new Boolean(true), 'Run an interactive session', true)
```
eldadfux, line 33: Should this be camelCase or dash-case?

stnguyen90: I did camel because `httpPort` and `httpsPort` are camelCase. That said, in CLIs dash-case is more common.

eldadfux: Will it be a breaking change if we adapt our style? do we have inconsistencies in other places?

stnguyen90: Yes, it'll be a breaking change if people are using these scripts and passing the params camelCase. We would have to update utopia-php/cli to support both camelCase and dash-case for backwards compatbility.

This is the only command that has multiple words so it's the only one that has camelCase.

stnguyen90: Eldad said:

> if we correct the styling and update the docs, the impact should be 0 for current users. Any reasons not to fix it?

@eldadfux, if someone has some script that uses the old syntax, it will break. I do agree the impact is small, though.  That said, I've updated the arguments.

eldadfux (approved):

PR #7279, opened 2023-12-13
Feat new usage metric

shimonewman:


src/Appwrite/Platform/Workers/UsageHook.php
```diff
@@ -66,6 +67,38 @@ public function action($register, $cache, $pools): void
                             $time = 'inf' === $period ? null : date($format, time());
                             $id = \md5("{$time}_{$period}_{$key}");
 
+                            /**
+                             * Infinity daily
+                             */
+                            if ('infd' === $period) {
+                                $isInfDaily = array_reduce(["files", "buckets", "users"], function ($carry, $word) use ($key) {
```
christyjacob4, line 74: Do we only need to compute inf metrics for these 3 metric types?

christyjacob4: We also need the size of the storage

Meldiron (approved):

Issue #6166, opened 2023-09-06
🐛 Bug Report: Clarify error message when project ID is missing


stnguyen90:
### 👟 Reproduction steps

Make an API call like list documents for a database and collection that exists, but don't set the project ID.

### 👍 Expected behavior

The error says project ID is missing

### 👎 Actual Behavior

The error says database not found.

This happens because we default to the console project if no project ID is supplied. [This PR](https://github.com/appwrite/appwrite/pull/5639) attempted to fix this, but it was [reverted in 1.4.1](https://github.com/appwrite/appwrite/commit/345df3f2cafce9d8b9d66562cb83860c5835653e) when trying to make sure the [VCS callback](https://github.com/appwrite/appwrite/blob/1.4.1/app/controllers/api/vcs.php#L262-L277) doesn't throw an error like:

![image](https://github.com/appwrite/appwrite/assets/1477010/32bb17cd-0124-48f1-a996-950ddfe4189d)

We need to make sure to not default the project to console, while ensuring the VCS callback still works.

### 🎲 Appwrite version

Version 1.4.x

### 💻 Operating system

Linux

### 🧱 Your Environment

Appwrite Version 1.4.1

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

KanniShashankh:
Can i be assigned this issue, also a small breif description about the issue @stnguyen90

utsavbhattarai007:
Hi, Can you assign to me this issue?

nick2432:
can i work on this issue?

stnguyen90:
An example, when I create a document without a project ID:

```shell
$ curl 'https://localhost/v1/databases/default/collections/test/documents' -k \
  -H 'content-type: application/json' \
  --data-raw '{"documentId":"unique()","data":{"searchAttribute":"asdf"},"permissions":[]}'
```

I get a server error:

```json
{
    "message": "Server Error",
    "code": 500,
    "type": "general_unknown",
    "version": "dev"
}
```

The docker logs from the appwrite container shows:

```
appwrite  | [Error] Timestamp: 2024-04-08T20:00:03+00:00
appwrite  | [Error] Method: POST
appwrite  | [Error] URL: /v1/databases/:databaseId/collections/:collectionId/documents
[...]

PR #7341, opened 2023-12-27
databases.php collection not found

fogelito:
Fix Collection not found 500
In the Error I saw the projectId = "console". so perhaps we can disable console project in the API?

Fixes: https://github.com/appwrite/appwrite/issues/6166


app/controllers/api/databases.php
```diff
@@ -2973,7 +2974,11 @@ function updateAttribute(
     ->inject('dbForProject')
     ->inject('mode')
     ->action(function (string $databaseId, string $collectionId, array $queries, Response $response, Database $dbForProject, string $mode) {
-        $database = Authorization::skip(fn() => $dbForProject->getDocument('databases', $databaseId));
+        try {
+            $database = Authorization::skip(fn() => $dbForProject->getDocument('databases', $databaseId));
```
fogelito, line 2978: or should we try to getCollection('databases') and check exists?

app/controllers/api/databases.php
```diff
@@ -2973,7 +2974,11 @@ function updateAttribute(
...
+        } catch (DatabaseException $e) {
```
abnegate, line 2979: Shouldn't we just get an empty document here instead of a DatabaseException?

fogelito: Because the 'databases' collection does not exists, an Exception is being thrown.
We can see in the error  
$dbForProject is the console , that database does not exist in the console DB , if possible it should not even get to this route

fogelito: projectId=console

eldadfux: Don't we have the same thing in a gazillion places? @abnegate a different approach would be to disable databases, functions and other services the console don't use in the hardcoded console project we have.

abnegate: True that makes a lot more sense! I think it will fix other cases we've seen with similar issues @fogelito

app/controllers/general.php
```diff
@@ -357,6 +358,15 @@ function router(App $utopia, Database $dbForConsole, callable $getProjectDB, Swo
     return false;
 }
 
+App::init()
+    ->groups(['database', 'function', 'storage', 'messaging'])
+    ->inject('project')
+    ->action(function (Document $project) {
+        if ($project->getId() === 'console') {
+            throw new Exception(Exception::PROJECT_RESERVED_PROJECT, 'Please check X-Appwrite-Project header');
```
eldadfux, line 366: This error is not good enough. It assumes the consumer context is the API but ignores SDKs or console where the consumer has no idea what this error is.

app/controllers/general.php
```diff
@@ -357,6 +358,15 @@ function router(App $utopia, Database $dbForConsole, callable $getProjectDB, Swo
     return false;
 }
 
+App::init()
+    ->groups(['database', 'function', 'storage', 'messaging'])
```
eldadfux, line 362: Why are we being so specific on the groups and not using something like the API group?

eldadfux (changes requested):

eldadfux:
Also, are we getting `console` project ID by default even if no project ID was set? Where is this happening? what happens if the header value is empty or not provided at all?

app/controllers/general.php
abnegate: @gewenyu99 Any input for a message here? This is the case when the project is set as `console`, and you're calling an unsupported API like databases, functions etc.

app/controllers/general.php
abnegate: The console project needs access to some APIs (account/user) that are part of the `api` group, thought it would be easier to be specific so if any of these APIs need console access in future, we can just remove them instead of reworking the whole approach

abnegate:
> Also, are we getting `console` project ID by default even if no project ID was set? Where is this happening? what happens if the header value is empty or not provided at all?

Yep we are defaulting to console here: https://github.com/appwrite/appwrite/blob/main/app/init.php#L1233-L1235

app/controllers/general.php
gewenyu99: Questions here: 

1. Can I name my project ID `console`? We shouldn't allow this during creation.
2. Could this ever display on the console?

Either way:
[What went wrong][why we think it went wrong][steps to debug/fix]
We can try:
`This endpoint is not available for the `console` project. The Appwrite Console is a reserved project ID and cannot be used with the Appwrite SDKs and APIs. Please check if your project ID is correct.`

fogelito: @gewenyu99 No we can not you will get an error this project name is resolved.

I think this message is misleading since in 99% cases when  X-Appwrite-Project Header is not assigned, and 1%  the project 'console' is assigned, so we have a fallback to console default project.

fogelito: @gewenyu99 
1. You can not name a project ID console, it's disallowed during creation.
2. This happens in 99% when a user does not assign the X-Appwrite-Project header , We are the ones who fallback to console project

gewenyu99: @fogelito We should throw a separate error for when project ID is not assigned. Is that possible?

fogelito: I guess we can do something like:
https://github.com/appwrite/appwrite/pull/7341/files#diff-ff12d23a7b6584b45f7e669b24ae1637106903cbac9390cad416e2c21dfda3f4R367

@abnegate what do you think?

gewenyu99: @fogelito 
Let's work on the error message:
IF empty:
`No Appwrite project was specified. Please specify your Appwrite Project ID in the `x-appwrite-project` header or when you initialize your Appwrite SDK Client.

gewenyu99: If console:
`This endpoint is not available for the console project. The Appwrite Console is a reserved project ID and cannot be used with the Appwrite SDKs and APIs. Please check if your project ID is correct.`

app/controllers/general.php
```diff
@@ -350,13 +351,26 @@ function router(App $utopia, Database $dbForConsole, callable $getProjectDB, Swo
         $utopia->getRoute()?->label('error', '');
         return false;
     } else {
-        throw new AppwriteException(AppwriteException::GENERAL_SERVER_ERROR, 'Unknown resource type ' . $type);
+        throw new AppwriteException(AppwriteException::GENERAL_ACCESS_FORBIDDEN, 'Unknown resource type ' . $type);
```
abnegate, line 354: What's the reasoning for this change?

app/controllers/general.php
```diff
@@ -350,13 +351,26 @@ function router(App $utopia, Database $dbForConsole, callable $getProjectDB, Swo
...
     }
 
     $utopia->getRoute()?->label('error', '');
     return false;
 }
 
+App::init()
+    ->groups(['database', 'function', 'storage', 'messaging'])
```
abnegate, line 362: Functions is using a plural for group:

```suggestion
    ->groups(['database', 'functions', 'storage', 'messaging'])
```

app/controllers/general.php
```diff
@@ -350,13 +351,26 @@ function router(App $utopia, Database $dbForConsole, callable $getProjectDB, Swo
...
+    ->inject('project')
+    ->inject('request')
+    ->action(function (Document $project, Appwrite\Utopia\Request $request) {
+        if ($project->getId() === 'console') {
+            $message = empty($request->getHeader('x-appwrite-project')) ?
+                'No Appwrite project was specified. Please specify your Appwrite Project ID in the "x-appwrite-project" header or when you initialize your Appwrite SDK Client.' :
+                'This endpoint is not available for the console project. The Appwrite Console is a reserved project ID and cannot be used with the Appwrite SDKs and APIs. Please check if your project ID is correct.';
```
abnegate, line 369: Small change:

```suggestion
                'No Appwrite project was specified. Please specify your Appwrite Project ID in the "x-appwrite-project" header or when you initialize your Appwrite SDK.' :
                'This endpoint is not available for the console project. The Appwrite Console is a reserved project ID and cannot be used with the Appwrite SDKs and APIs. Please check if your project ID is correct.';
```

app/controllers/general.php
fogelito: Copied from this PR
https://github.com/appwrite/appwrite/pull/7954/files#diff-ff12d23a7b6584b45f7e669b24ae1637106903cbac9390cad416e2c21dfda3f4R375

fogelito: Wrong copy fixed now

app/controllers/general.php
eldadfux: 1. The console project is something internal, a normal dev using Appwrite has no idea what it is and its very confusing. 
2. We shouldn't mention `x-appwrite-project` just like that, this header is only visible to REST users, which are minority if we do mention the context should be clear and the first thing mention should be setting the project ID. We can also skip the header name altogether.

gewenyu99: 1. @eldadfux We have two errors now. IF they get an error that mentions console, they have it set as their project. IMO this is fine.

2. We can do: No Appwrite project was specified. Please specify your Appwrite Project ID on your Appwrite SDK Client.

app/controllers/general.php
```diff
@@ -357,6 +357,19 @@ function router(App $utopia, Database $dbForConsole, callable $getProjectDB, Swo
...
 }
 
+App::init()
+    ->groups(['database', 'functions', 'storage', 'messaging'])
+    ->inject('project')
+    ->inject('request')
+    ->action(function (Document $project, Appwrite\Utopia\Request $request) {
+        if ($project->getId() === 'console') {
+            $message = empty($request->getHeader('x-appwrite-project')) ?
+                'No Appwrite project was specified. Please specify your Appwrite Project ID in the "x-appwrite-project" header or when you initialize your Appwrite SDK.' :
```
abnegate, line 367:
```suggestion
                'No Appwrite project was specified. Please specify your project ID when you initialize your Appwrite SDK.' :
```

app/controllers/general.php
```diff
@@ -357,6 +357,19 @@ function router(App $utopia, Database $dbForConsole, callable $getProjectDB, Swo
     return false;
 }
 
+App::init()
+    ->groups(['database', 'functions', 'storage', 'messaging'])
+    ->inject('project')
+    ->inject('request')
+    ->action(function (Document $project, Appwrite\Utopia\Request $request) {
```
abnegate, line 364:
```suggestion
    ->action(function (Document $project, Request $request) {
```

abnegate (approved):

PR #7349, opened 2023-12-28
fix for file extension not supported

lohanidamodar:
- the error occured with jfif extension, which is essentially a jpeg file

## What does this PR do?

- Fixes file preview when default file extension is not one of supported ones but has supported mime/type and output format is not provided (occurred with .jfif image which has mime type of image/jpeg)

## Test Plan

- Update storage test to include test for this case

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/storage.php
```diff
@@ -915,6 +915,12 @@
         }
 
         if (empty($output)) {
+            // when file extension is provided but it's not one of our
+            // supported outputs we fallback to `jpg`
+            if (!empty($type) && !array_key_exists($type, $outputs)) {
```
eldadfux, line 920: But wouldn't this break everything else? what happened if someone tries to use `exe` as an image?

lohanidamodar: `exe` will not have the supported `mme/type` which case is already captured above where we default back to our default image preview logo

eldadfux (approved):

PR #7351, opened 2023-12-28
Improve large file handling

lohanidamodar:
## What does this PR do?

- improves preview and view endpoint

## Test Plan

- No new tests

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/controllers/api/storage.php
+++ b/app/controllers/api/storage.php
@@ -872,14 +872,6 @@ App::get('/v1/storage/buckets/:bucketId/files/:fileId/preview')
             throw new Exception(Exception::USER_UNAUTHORIZED);
         }
 
-        if ((\strpos($request->getAccept(), 'image/webp') === false) && ('webp' === $output)) { // Fallback webp to jpeg when no browser support
-            $output = 'jpg';
-        }
-
-        $inputs = Config::getParam('storage-inputs');
-        $outputs = Config::getParam('storage-outputs');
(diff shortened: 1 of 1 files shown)


app/controllers/api/storage.php
```diff
@@ -890,9 +887,17 @@
...
+        }
+
+        $inputs = Config::getParam('storage-inputs');
+        $outputs = Config::getParam('storage-outputs');
+        $fileLogos = Config::getParam('storage-logos');
+
         $path = $file->getAttribute('path');
         $type = \strtolower(\pathinfo($path, PATHINFO_EXTENSION));
-        $algorithm = $file->getAttribute('algorithm', 'none');
+        $algorithm = $file->getAttribute('algorithm', COMPRESSION_TYPE_NONE);
```
eldadfux, line 900: Where are this constants coming from? Can we have them come from the compression class? 
https://github.com/utopia-php/storage/blob/main/src/Storage/Compression/Compression.php

eldadfux (changes requested):

app/controllers/api/storage.php
lohanidamodar: @eldadfux These were already defined and used in other places, I'm just using it in places where it was missed. Can we do the refactor in separate PR?

eldadfux: Yes, but we should.

eldadfux (approved):

PR #7377, opened 2024-01-02
Add `_APP_CONSOLE_HOSTNAMES` env var to allow more hostnames to the console project

stnguyen90:
## What does this PR do?

- adds `_APP_CONSOLE_HOSTNAMES` environment variable to allow more hostnames to the console project

## Test Plan

- manual

## Related PRs and Issues

- https://github.com/appwrite/appwrite/pull/7360

## Checklist


TorstenDittmann (approved):

app/init.php
```diff
@@ -927,6 +927,18 @@ function (mixed $value) {
         'hostname' => $request->getHostname(),
     ], Document::SET_TYPE_APPEND);
 
+    $hostnames = explode(',', App::getEnv('_APP_CONSOLE_HOSTNAMES', ''));
+    if (is_array($hostnames)) {
```
stnguyen90, line 931: Hmm if `App::getEnv('_APP_CONSOLE_HOSTNAMES', '')` and an empty string, `$hostnames` will be an array with 1 empty string. That shouldn't be added as a platform, right?

TorstenDittmann: ohhh, one of those weird PHP things. We should fix it before releasing.

PR #7380, opened 2024-01-02
Create an enum for Message status

stnguyen90:
## What does this PR do?

Create an enum for Message status since there's a set of possible values for the status.

## Test Plan

None

## Related PRs and Issues

None

## Checklist

### DIFF
--- a/src/Appwrite/Enum/MessageStatus.php
+++ b/src/Appwrite/Enum/MessageStatus.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Appwrite\Enum;
+
+enum MessageStatus: string
+{
+    /**
+     * Message that is not ready to be sent
+     */
+    case Draft = 'draft';
+    /**
+     * Scheduled to be sent for a later time
+     */
+    case Scheduled = 'scheduled';
+    /**
+     * Picked up by the worker and starting to send
+     */
+    case Processing = 'processing';
+    /**
+     * Sent without errors
+     */
+    case Sent = 'sent';
+    /**
+     * Sent with some errors
+     */
+    case Failed = 'failed';
+}


src/Appwrite/Enum/MessageStatus.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace Appwrite\Enum;
+
+enum MessageStatus: string
```
eldadfux, line 5: Will we have more enums for `Message` or other entities? Maybe the structure should be Appwrite\Enum\[Entity]\[Type] or `Appwrite\Enum\Message\Type`?

TorstenDittmann: I tihnk it makes mroe sense to have it under the namespace where its actually used.

We already have `Appwrite\Messaging`

stnguyen90: > Will we have more enums for Message or other entities? Maybe the structure should be Appwrite\Enum[Entity][Type] or Appwrite\Enum\Message\Type?

@eldadfux, yes, we can have more. For example, `providerType` can be `email`, `sms`, or `push`.

> I think it makes more sense to have it under the namespace where it actually follows the story.
> 
> We already have `Appwrite\Messaging` for that imo

@TorstenDittmann, that path is for Realtime, though 😅. It's probably most related to [`Appwrite\Utopia\Response\Model\Message`](https://github.com/appwrite/appwrite/blob/1.5.x/src/Appwrite/Utopia/Response/Model/Message.php).

TorstenDittmann: I am kinda against having a separate namespace for enums, it's always nice if stuff like that is relative to its consuming classes (if there are).

However, we don't have enough enums implemented to know how it will look like.

TorstenDittmann (approved):

PR #7245, opened 2023-12-05
Adds uniform error logic for messaging worker and extra params for email 

fanatic75:
## What does this PR do?

Adds uniform error logic for messaging worker and extra params for email 
Blocked by https://github.com/utopia-php/messaging/pull/74

## Test Plan

(Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified your changes work. Screenshots may also be helpful.)

## Related PRs and Issues

- (Related PR or issue)

## Checklist

- [ ] Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [ ] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

PR #7394, opened 2024-01-04
Don't auto-delete targets

stnguyen90:
## What does this PR do?

Automatically deleting targets can cause confusion for developers. Instead, let the developer delete it.

## Test Plan

Manual

## Related PRs and Issues

- https://github.com/appwrite/appwrite/pull/7245

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Workers/Messaging.php
+++ b/src/Appwrite/Platform/Workers/Messaging.php
@@ -49,7 +49,7 @@ class Messaging extends Action
             ->desc('Messaging worker')
             ->inject('message')
             ->inject('dbForProject')
-            ->callback(fn(Message $message, Database $dbForProject) => $this->action($message, $dbForProject));
+            ->callback(fn (Message $message, Database $dbForProject) => $this->action($message, $dbForProject));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -195,14 +195,6 @@ private function processMessage(Database $dbForProject, Document $message): void
...
                                     $deliveryErrors[] = `Failed sending to target {$detail['recipient']} with error: {$detail['error']}`;
                                 }
-
-                                // Deleting push targets when token has expired.
-                                if ($detail['error'] === 'Expired device token.') {
-                                    $target = $dbForProject->findOne('targets', [Query::equal('identifier', [$detail['recipient']])]);
-                                    if ($target instanceof Document && !$target->isEmpty()) {
-                                        $dbForProject->deleteDocument('targets', $target->getId());
-                                    }
-                                }
```
abnegate, line 205: We should automatically delete a target if the token is expired/invalid. Imagine if I release my app, 1 million users install + sign up, and are added to the "all users" topic. The next day all but 1 uninstall the app. Now, sending a message to the "all users" topic will _always_ have 999,999 failures.

stnguyen90: @eldadfux, thoughts?

eldadfux: Can we let the maintenance and delete workers collaborate on this? we do similar things in other areas.

eldadfux: Another option is to skip them and still count them for usage, this way there's a bigger incentive for the devs to handle the list properly.

abnegate: We could do this in the deletes worker if we mark them as invalid here and then trigger the deletes worker. I don't think we can keep & skip them, since for mobile apps, there's no uninstall hook; developers have no way to do the removal themselves

eldadfux: I think it make sense to mark them here or maybe even count consecutive failures.

PR #7395, opened 2024-01-04
Replace DELETE_TYPE consts with enum

stnguyen90:
## What does this PR do?

A backed enum should be used for related constants that are a set of possible values.

## Test Plan

Tests should still pass.

## Related PRs and Issues

- None

## Checklist


stnguyen90:
The test are failing because Appwrite uses PHP version 8.0, but 8.1 is required for backed enums.

PR #7396, opened 2024-01-04
Delete subscribers and update topic totals when deleting target

stnguyen90:
## What does this PR do?

To ensure data is consistent, subscribers should be deleted if a target is deleted. In addition, the topic total should be updated when the subscriber is deleted.

## Test Plan

Manual

## Related PRs and Issues

None

## Checklist


stnguyen90:
This change will need to be decoupled from the enum change before it can be merged.

stnguyen90:
Not sure what happened to the GH action, but the tests passed when I ran them myself:

image

abnegate (approved):

PR #7402, opened 2024-01-06
Webhook attempts PR suggestions

Meldiron:


src/Appwrite/Platform/Workers/Webhooks.php
```diff
@@ -120,24 +120,26 @@ private function execute(array $events, string $payload, Document $webhook, Docu
...
+        if (!empty($curlError) || $statusCode >= 400) {
             $dbForConsole->increaseDocumentAttribute('webhooks', $webhook->getId(), 'attempts', 1);
             $webhook = $dbForConsole->getDocument('webhooks', $webhook->getId());
             $attempts = $webhook->getAttribute('attempts');
 
-            $logs = json_encode([
-                'URL' => $webhook->getAttribute('url'),
-                'Method' => 'POST',
-            ]);
+            $logs = '';
```
vermakhushboo, line 133: Since `logs` is now a string, do we need to parse it on console? Won't JSON be more helpful?

Meldiron: We wont be parsing it, just print as text in logs box.

The proper waywould be to have separate attribute for each, to also allow filtering. Ideally every log is it's own separate document  in `webhookIncident` collection, so developers can also see history, re-run, see how many times it failed, and so on.

For now, string is fine. I don't think design needs us to have them separated.

Meldiron: Double check with Eldad about the future if we want to be extra safe. I don't think we would need to migrate this once we introduce proper webhook logs, as this only has failed ones, and new solution will have all of them.

vermakhushboo (approved):
Good to go for now

PR #7408, opened 2024-01-08
fix: escape html in email params

TorstenDittmann:
## What does this PR do?

- use new framework version that escapes params by default
- dont escape for certain fields

## Test Plan

- tests added

## Checklist

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -1009,7 +1009,8 @@ App::post('/v1/account/sessions/magic-url')
 
         $message = Template::fromFile(__DIR__ . '/../../config/locale/templates/email-inner-base.tpl');
         $message
-            ->setParam('{{body}}', $body)
+            ->setParam('{{body}}', $body, escapeHtml: false)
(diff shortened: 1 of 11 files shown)


christyjacob4 (approved):

app/controllers/api/account.php
```diff
@@ -1064,11 +1065,10 @@
 
         $emailVariables = [
             'direction' => $locale->getText('settings.direction'),
-            /* {{user}} ,{{team}}, {{project}} and {{redirect}} are required in the templates */
+            /* {{user}} ,{{team}} and {{project}} are required in the templates */
             'user' => '',
             'team' => '',
-            'project' => $project->getAttribute('name'),
-            'redirect' => $url
+            'project' => $project->getAttribute('name')
```
stnguyen90, line 1071: I don't think redirect should be removed here. It should be a variable and the final rendered email uses the redirect variable in the worker.

stnguyen90:
Closing stale PR

PR #7409, opened 2024-01-08
fix: escape html in email params

TorstenDittmann:
see #7408

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -6,7 +6,7 @@
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
          processIsolation="false"
-         stopOnFailure="false"
+         stopOnFailure="true"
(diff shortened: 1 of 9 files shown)


phpunit.xml
```diff
@@ -6,7 +6,7 @@
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
          processIsolation="false"
-         stopOnFailure="false"
+         stopOnFailure="true"
```
christyjacob4, line 9: we should probably revert this ?

christyjacob4 (approved):

phpunit.xml
TorstenDittmann: yes, my bad

christyjacob4 (approved):

PR #7424, opened 2024-01-10
Added security phrase to magic URL

Meldiron:
## What does this PR do?

Adds security phrases to Magic URL emails. Useful to prevent attacks such as man in the middle.

CleanShot 2024-01-10 at 15 40 33@2x

## Test Plan

new tests implemented

## Related PRs and Issues

x

## Checklist


src/Appwrite/Platform/Tasks/DevGenerateTranslations.php
```diff
@@ -79,7 +79,7 @@ private function generateTranslation(string $targetLanguage, string $enTranslati
             'content-type' => Client::CONTENT_TYPE_APPLICATION_JSON,
             'Authorization' => 'Bearer ' . $this->apiKey
         ], Client::METHOD_POST, [
-            'model' => 'gpt-4-1106-preview', // https://platform.openai.com/docs/models/gpt-4-and-gpt-4-turbo
+            'model' => 'gpt-3.5-turbo-1106', // https://platform.openai.com/docs/models/gpt-4-and-gpt-4-turbo
```
Meldiron, line 82:
```suggestion
            'model' => 'gpt-4-1106-preview', // https://platform.openai.com/docs/models/gpt-4-and-gpt-4-turbo
```

christyjacob4:
@Meldiron can we make the message a little bit better? 

"If you are unable to sign it using the button above, please use the following link ... "

app/controllers/api/account.php
```diff
@@ -932,6 +934,7 @@
     ->param('userId', '', new CustomId(), 'Unique Id. Choose a custom ID or generate a random ID with `ID.unique()`. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.')
     ->param('email', '', new Email(), 'User email.')
     ->param('url', '', fn($clients) => new Host($clients), 'URL to redirect the user back to your app from the magic URL login. Only URLs from hostnames in your project platform list are allowed. This requirement helps to prevent an [open redirect](https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html) attack against your project API.', true, ['clients'])
+    ->param('securityPhrase', false, new Boolean(), 'Toggle for security phrase. If enabled, email will be send with a randomly generated phrase and the phrase will also be included in the response. Confirming phrases match incrases the security of authentication flow.', true)
```
christyjacob4, line 937:
```suggestion
    ->param('securityPhrase', false, new Boolean(), 'Toggle for security phrase. If enabled, email will be send with a randomly generated phrase and the phrase will also be included in the response. Confirming phrases match increases the security of authentication flow.', true)
```

Meldiron:
> @Meldiron can we make the message a little bit better?
> 
> "If you are unable to sign it using the button above, please use the following link ... "

Why would you not be able to? The only reason I could find was that the button doesn't show, as your client blocks it. I like my version better as it gives more simple verbose indication of a problem.

app/controllers/api/account.php
```diff
@@ -940,12 +943,16 @@
...
+    ->action(function (string $userId, string $email, string $url, bool $securityPhrase, Request $request, Response $response, Document $user, Document $project, Database $dbForProject, Locale $locale, Event $queueForEvents, Mail $queueForMails) {
 
         if (empty(App::getEnv('_APP_SMTP_HOST'))) {
             throw new Exception(Exception::GENERAL_SMTP_DISABLED, 'SMTP disabled');
         }
 
+        if ($securityPhrase === true) {
+            $securityPhrase = SecurityPhrase::generate();
+        }
+
```
christyjacob4, line 955: I think we might have an issue with boolean parameters in some SDKs. 

See what we do for the activate param in functions. I believe its still broken ?
https://github.com/appwrite/appwrite/blob/7e1b618769448fd222d51551f64080d49443c1fe/app/controllers/api/functions.php#L1067

Meldiron: Problem should only be there with GET requests, and non-JSON body requests. It also relates to `new Boolean()` being `loose: true`.

Here we have loose false (default), so we can always trust it to be boolean from JSON body (its POST request)

christyjacob4: The functions request is also a POST request. 
Have you tested your endpoint with JS? We noticed this issue in the console so it may be a JS specific thing.

Meldiron: Function request is file upload. Its not JSON body.

Meldiron: Ill test with JS SDK to be extra safe

christyjacob4: the check `$securityPhrase === true` will probably always be false. because the parameter `$securityPhrase` may be processed as a string by the server

christyjacob4: Yea what I mean to say is, JS ( and maybe more languages ) send parameters in a POST request body as strings. So the validator itself will block those requests

Meldiron: There are other places such as `required` on new string attribute, or `logging` on function update. There we do not convert booleans. If this was a problem, we would have noticed it by now. What we noticed so far is only new deployment, which is not JSON body.

I did QA with web SDK and it works as expected.

CleanShot 2024-01-11 at 10 55 37@2x

CleanShot 2024-01-11 at 10 55 45@2x

christyjacob4: What happens when you pass 'true' or 'false' as a parameter ? In both cases I would expect securityPhrase to be empty

christyjacob4:
> Why would you not be able to? The only reason I could find was that the button doesn't show, as your client blocks it. I like my version better as it gives more simple verbose indication of a problem.

Button doesn't show, maybe the URL in the button is incorrect, your browser has blocked opening links from emails because it look suspicious. 

Also saying something like "if the button above doesn't show" does not sound very professional. And rather than "use the following link" its better to say "please use the following link"

app/controllers/api/account.php
Meldiron: You get yelled at by IDE, since type is `securityPhrase?: boolean | undefined`.
Also by validator, you get 400 `Invalid `securityPhrase` param: Value must be a valid boolean`.

If you pass string there, you are not using SDKs correctly.

Meldiron:
Let's move quick, it's for 1.5. Ill use your suggestion, we can open discussion again after we have feedback from community.

app/config/locale/translations/en.json
```diff
@@ -13,8 +13,9 @@
     "emails.magicSession.hello": "Hello,",
     "emails.magicSession.optionButton": "Click the button below to securely sign in to your {{project}} account. It will expire in 1 hour.",
     "emails.magicSession.buttonText": "Sign in to {{project}}",
-    "emails.magicSession.optionUrl": "If the button above doesn't show, use the following link:",
+    "emails.magicSession.optionUrl": "If you are unable to sign it using the button above, please visit the following link:",
```
christyjacob4, line 16:
```suggestion
    "emails.magicSession.optionUrl": "If you are unable to sign in using the button above, please visit the following link:",
```

Meldiron: Nooo 😭

christyjacob4: I know 😭

Meldiron: And I am using GPT 4 to translate all, poor usage stats 😭

christyjacob4 (approved):

PR #7425, opened 2024-01-10
Delete console users

shimonewman:


src/Appwrite/Platform/Tasks/PatchDeleteConsoleUsers.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        return 'patch-delete-console-users';
+    }
+
+    public function __construct()
+    {
+        $this
+            ->desc('Delete console users patch')
+            ->inject('dbForConsole')
+            ->inject('queueForDeletes')
```
christyjacob4, line 25: we should have a parameter for a --dry-run ( be default it needs to be true ) 

and have a confirmation where we have to type "YES" to proceed

bin/patch-delete-console-users
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+php /usr/src/code/app/cli.php patch-delete-console-users $@
```
christyjacob4, line 3: Let's move these scripts to the cloud repository

shimonewman:
The API part was already merged by Thorsten

PR #7434, opened 2024-01-11
Make OTP template more contextual

vermakhushboo:
## What does this PR do?

Add more details to the OTP message and generate translations for all languages

## Test Plan

## Checklist


christyjacob4:
@vermakhushboo the PR looks good at first glance. But I dont think it addresses the issues mentioned. The issues mentioned talk about being able to customise the sms message which we plan to do in a separate feature.

christyjacob4 (approved):

vermakhushboo:
> @vermakhushboo the PR looks good at first glance. But I dont think it addresses the issues mentioned. The issues mentioned talk about being able to customise the sms message which we plan to do in a separate feature.

@christyjacob4 I agree, but the PR I have created is for this task that was assigned to me: https://app.asana.com/0/1201750942661294/1205916879308768/f

I think @stnguyen90 added related issues to it. Steven, can we create new tasks for those and assign separately?

app/config/locale/translations/en.json
```diff
@@ -36,6 +36,7 @@
     "emails.certificate.footer": "Your previous certificate will be valid for 30 days since the first failure. We highly recommend investigating this case, otherwise your domain will end up without a valid SSL communication.",
     "emails.certificate.thanks": "Thanks",
     "emails.certificate.signature": "{{project}} team",
+    "sms.verification.body": "{{secret}} is your {{project}} verification code.",
```
christyjacob4, line 39: Actually I think its important for the message to be structured as `Your [PROJECT_NAME] verification code is: [CODE]`
This will allow the phone operating systems to detect and suggest codes in apps

christyjacob4 (changes requested):
Please address the comments.

app/config/locale/translations/en.json
vermakhushboo: @christyjacob4, @Meldiron and I researched a bit into how other companies do it and we found that most of them follow this structure where code is in the beginning of the message. This would help in 2 things: 
1. The user can easily read the OTP from the notification without having to open the message
2. If the project name is too large, it would be hard to read the OTP/ the message may exceed allowed characters of one message.

We can finalise the approach and I'll make the changes accordingly.

christyjacob4: Okay I am fine with either messages as long as its detectable by the OS. Can we do a small test to trigger an SMS and check if the phone recognizes it?

christyjacob4 (approved):

PR #7437, opened 2024-01-11
fix: hotfix for redirect param in custom templates

TorstenDittmann:
## What does this PR do?

- fix redirect param not being set in custom templates
- dont escape redirect params in Mails worker

## Test Plan

- to be added

## Checklist

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -1011,7 +1011,6 @@ App::post('/v1/account/sessions/magic-url')
         $message = Template::fromFile(__DIR__ . '/../../config/locale/templates/email-inner-base.tpl');
         $message
             ->setParam('{{body}}', $body, escapeHtml: false)
-            ->setParam('{{redirect}}', $url, escapeHtml: false)
             ->setParam('{{hello}}', $locale->getText("emails.magicSession.hello"))
             ->setParam('{{footer}}', $locale->getText("emails.magicSession.footer"))
             ->setParam('{{thanks}}', $locale->getText("emails.magicSession.thanks"))
@@ -1069,6 +1068,7 @@ App::post('/v1/account/sessions/magic-url')
             /* {{user}} ,{{team}} and {{project}} are required in the templates */
             'user' => '',
             'team' => '',
+            'redirect' => $url,
             'project' => $project->getAttribute('name')
         ];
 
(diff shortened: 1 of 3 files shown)


src/Appwrite/Platform/Workers/Mails.php
```diff
@@ -67,7 +67,8 @@ public function action(Message $message, Registry $register): void
         $bodyTemplate = Template::fromFile($bodyTemplate);
         $bodyTemplate->setParam('{{body}}', $body, escapeHtml: false);
         foreach ($variables as $key => $value) {
-            $bodyTemplate->setParam('{{' . $key . '}}', $value);
+            // TODO: hotfix for redirect param
+            $bodyTemplate->setParam('{{' . $key . '}}', $value, escapeHtml: $key !== 'redirect');
```
stnguyen90, line 71: Why shouldn't we escape redirect?

app/controllers/api/account.php
```diff
@@ -1069,6 +1068,7 @@
             /* {{user}} ,{{team}} and {{project}} are required in the templates */
```
christyjacob4, line 1068: lets add it back to the comment. and mention not to remove it

src/Appwrite/Platform/Workers/Mails.php
TorstenDittmann: Bildschirmfoto 2024-01-11 um 20 43 46
source => https://www.php.net/manual/en/function.htmlspecialchars.php

app/controllers/api/account.php
TorstenDittmann: on it

src/Appwrite/Platform/Workers/Mails.php
stnguyen90: `$redirect` should only be a URL and when the worker receives the `$bodyTemplate` it should either be a custom template or one of ours. Ours is the [email-inner-base.tpl](https://github.com/appwrite/appwrite/blob/bd920ead40b50b8be1cec32fcc8299ec2e5454a3/app/config/locale/templates/email-inner-base.tpl) with the following from the following replaced:

https://github.com/appwrite/appwrite/blob/b0a8aac0b7b7511523d8127792342bea9d552b85/app/config/locale/translations/en.json#L7-L11

For this example, it'd be:

```html
<p>Hey {{user}}</p>

<br>

<p>Follow this link to verify your email address.</p>

<a href="{{redirect}}" target="_blank">{{redirect}}</a>

<p>If you didn’t ask to verify this address, you can ignore this message.</p>

<br>

<p>Thanks</p>

<p>{{project}} team</p>
```

christyjacob4 (approved):

PR #7440, opened 2024-01-12
Added cloud base templates to appwrite/appwrite

vermakhushboo:
## What does this PR do?

Added all base templates for emails to appwrite/appwrite

## Test Plan

NA

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/config/locale/templates/email-base-branded.tpl
+++ b/app/config/locale/templates/email-base-branded.tpl
@@ -1,0 +1,98 @@
+<!doctype html>
+<html>
+    <head>
+        <link
+            href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600&family=Poppins:wght@500;600&display=swap"
+            rel="stylesheet"
+        />
+        <style>
+            .main {
+                padding: 32px;
+                color: #616b7c;
+                font-size: 15px;
+                font-family: "Inter", sans-serif;
+            }
+
+            table {
+                width: 100%;
+                border-spacing: 0 !important;
+            }
+
+            table,
+            tr,
+            th,
+            td {
+                margin: 0;
+                padding: 0;
+            }
+
+            td {
+                vertical-align: top;
+            }
+
+            .main a {
+                color: currentColor;
+            }
+
+            h1,
+            h2,
+            h3,
+            h4,
+            h5,
+            h6 {
+                font-family: "Poppins", sans-serif;
+            }
+
+            hr {
+                border: none;
+                border-top: 1px solid #e8e9f0;
+            }
+
+            p {
+                margin-bottom: 10px;
+            }
+        </style>
+    </head>
+
+    <body>
+        <div
+            class="main"
+            style="
+                direction: {{direction}};
+                max-width: 650px;
+                word-wrap: break-word;
+                overflow-wrap: break-word;
+                word-break: break-all;
+                margin: 0 auto;
+            "
+        >
+            <table style="margin-top: 32px">
+                <tr>
+                    <td>{{body}}</td>
+                </tr>
+            </table>
+
+            <div
+                style="
+                    border-top: solid 1px #ededf0;
+                    width: 100%;
+                    color: #818186;
+                "
+            >
+                <table style="padding-top: 24px; margin: 0 auto; width: auto">
+                    <tr>
+                        <td style="display: flex">
+                            <a target="_blank" href="https://appwrite.io?utm_source=email&utm_medium=footer&utm_campaign=user_emails">
+                                <img
+                                    src="https://appwrite.io/email/footer.png"
+                                    width="378"
+                                    height="27"
+                                />
+                            </a>
+                        </td>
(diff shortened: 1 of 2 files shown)


app/config/locale/templates/email-base-branded.tpl
```diff
@@ -0,0 +1,98 @@
...
+
+    <body>
+        <div
+            class="main"
+            style="
+                direction: {{direction}};
+                max-width: 650px;
+                word-wrap: break-word;
+                overflow-wrap: break-word;
+                word-break: break-all;
```
Meldiron, line 65: Please confirm with Torsten or Elad about this. Sara didn't like screenshots I was sending her, it was breaking words in the middle.

What I did instead was `break-word`, and only set `break-all` on links (`a`)

app/config/locale/templates/email-base-branded.tpl
```diff
@@ -0,0 +1,98 @@
...
+            href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600&family=Poppins:wght@500;600&display=swap"
+            rel="stylesheet"
+        />
+        <style>
+            .main {
+                padding: 32px;
+                color: #616b7c;
+                font-size: 15px;
+                font-family: "Inter", sans-serif;
+            }
```
Meldiron, line 14: I think Sara didn't line line height. We should set it to 150% IIRC.

app/config/locale/templates/email-base-cloud.tpl
```diff
@@ -0,0 +1,234 @@
...
+                <tr>
+                    <td><a href="https://appwrite.io/terms">Terms</a></td>
+                    <td style="color: #e8e9f0">
+                        <div style="margin: 0 8px">|</div>
+                    </td>
+                    <td><a href="https://appwrite.io/privacy">Privacy</a></td>
+                </tr>
+            </table>
+            <p style="text-align: center" align="center">
+                &copy; 2023 Appwrite | 251 Little Falls Drive, Wilmington 19808,
```
Meldiron, line 229: 2024

app/config/locale/templates/email-base-cloud.tpl
```diff
@@ -0,0 +1,234 @@
...
+                font-family: "DM Sans", sans-serif;
+                font-weight: 600;
+                font-size: 12px;
+                color: #4f5769;
+                margin: 0;
+                padding: 0;
+            }
+            hr {
+                border: none;
+                border-top: 1px solid #e8e9f0;
```
Meldiron, line 64: If we set colors, we must set background color of body to white. Otherwise some mail clients with dark theme support would make it look bad. Example:
CleanShot 2024-01-12 at 10 30 39@2x

CleanShot 2024-01-12 at 10 30 58@2x

app/config/locale/templates/email-base-branded.tpl
vermakhushboo: I agree, `break-word` wasn't looking good. We should use `break-all`

Meldiron: Not sure if we are on same page.

`word-break: break-all;` will break word in the middle
`word-break: break-word;` will only break after a word

I propose to use second as default, and only use first for links, as they are long and unreadable anyway.

vermakhushboo: Oh, sorry. Yeah, that's what I mean too.

app/config/locale/templates/email-base-cloud.tpl
eldadfux: I agree, I prefer it to look good rather than have 1/2 support in dark mode.

app/config/locale/templates/email-base-cloud.tpl
eldadfux: We can use a function here instead of manually updating.

app/config/locale/templates/email-base-branded.tpl
eldadfux: No strong opinion, whatever design think is best.

app/config/locale/templates/email-base-cloud.tpl
Meldiron: Agree, {{year}} placeholder sounds like good idea (with setParam)

app/config/locale/templates/email-base-branded.tpl
Meldiron: Sara already didnt like it when reviewing my MagicURL email with this issue, you can go ahead and apply what I mentioned.

app/config/locale/templates/email-base-cloud.tpl
vermakhushboo: Forcing the background color to white for now until we make a decision

Meldiron (approved):

christyjacob4 (approved):

PR #7452, opened 2024-01-15
Feat: session renewal

Meldiron:
## What does this PR do?

Allows session renewal to extend it's expiry date. Very useful for when the session duration is set to short like 5 min.

## Test Plan

- [x] New time-sensitive tests

## Related PRs and Issues

x

## Checklist


app/controllers/api/account.php
```diff
@@ -232,7 +232,6 @@
         $user->setAttributes($profile->getArrayCopy());
 
         $duration = $project->getAttribute('auths', [])['duration'] ?? Auth::TOKEN_EXPIRATION_LOGIN_LONG;
-
         $detector = new Detector($request->getUserAgent('UNKNOWN'));
         $record = $geodb->get($request->getIP());
         $expire = DateTime::formatTz(DateTime::addSeconds(new \DateTime(), $duration));
```
christyjacob4, line 237: is this variable being used? Else we can remove it

app/controllers/api/account.php
```diff
@@ -291,7 +291,6 @@
         $session
             ->setAttribute('current', true)
             ->setAttribute('countryName', $countryName)
-            ->setAttribute('expire', $expire)
```
christyjacob4, line 294: why doesn't a session have the expiry anymore?

app/controllers/api/account.php
christyjacob4: You can either delete the variable or remove `DateTime::formatTz()` from it

app/controllers/api/account.php
```diff
@@ -1488,7 +1483,6 @@
         $record = $geodb->get($request->getIP());
         $secret = Auth::tokenGenerator();
         $expire = DateTime::formatTz(DateTime::addSeconds(new \DateTime(), $duration));
```
christyjacob4, line 1485: same comment as above

app/controllers/api/account.php
```diff
@@ -2446,72 +2434,79 @@
...
+        $session = null;
         foreach ($sessions as $key => $session) {/** @var Document $session */
-            if ($sessionId == $session->getId()) {
-                // Comment below would skip re-generation if token is still valid
-                // We decided to not include this because developer can get expiration date from the session
-                // I kept code in comment because it might become relevant in the future
+            if ($sessionId === $session->getId()) {
+                $session = $session;
+                break;
+            }
```
christyjacob4, line 2455: This doesn't look like a good practice. The loop variable has the same name as the variable outside the loop. It can have strange side effects later on

app/controllers/api/account.php
```diff
@@ -2446,72 +2434,79 @@
     ->label('sdk.response.model', Response::MODEL_SESSION)
     ->label('abuse-limit', 10)
     ->param('sessionId', '', new UID(), 'Session ID. Use the string \'current\' to update the current device session.')
-    ->inject('request')
+    ->param('identity', false, new Boolean(), 'Refresh OAuth2 identity. If enabled and session was created using an OAuth provider, the access token of the identity will be refreshed', true)
```
christyjacob4, line 2437: Hmm not sure this is a good name for the parameter. Please suggest some alternatives

christyjacob4: This endpoint takes a `sessionID` as input. The session already contains information if it was from an OAuth provider or not right? 

Why do we need an additional parameter `identity` 

If you make a PATCH request using your OAuth session, it will update ( renew ) that specific session,

christyjacob4 (changes requested):
Please address the comments

app/controllers/api/account.php
Meldiron: For all the expire variable comments, I checked once, its always used when we set cookie.

Meldiron: Ill move them all near the cookies so its clear.

app/controllers/api/account.php
Meldiron: Renamed to `session` and `loopSession`

app/controllers/api/account.php
```diff
@@ -753,7 +751,6 @@
         $record = $geodb->get($request->getIP());
         $secret = Auth::tokenGenerator();
         $expire = DateTime::formatTz(DateTime::addSeconds(new \DateTime(), $duration));
```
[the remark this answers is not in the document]
Meldiron: Same as above

app/controllers/api/account.php
Meldiron: Same as above

app/controllers/api/account.php
Meldiron: Another opened question, should it throw exception if you want to refresh identity and session doesn't have identity?

Maybe there should be no param and we always extend both session duration and refresh oauth token?

app/controllers/api/account.php
```diff
@@ -2396,7 +2394,7 @@
     });
 
 App::patch('/v1/account/sessions/:sessionId')
-    ->desc('Update OAuth session (refresh tokens)')
+    ->desc('Renew session')
```
christyjacob4, line 2397:
```suggestion
    ->desc('Update (or renew) a session')
```

tests/e2e/Services/Account/AccountCustomClientTest.php
```diff
@@ -1785,7 +1785,9 @@ public function testConvertAnonymousAccountOAuth2()
             'content-type' => 'application/json',
             'x-appwrite-project' => $this->getProject()['$id'],
             'cookie' => 'a_session_' . $this->getProject()['$id'] . '=' . $session,
-        ]));
+        ]), [
+            'identity' => true
+        ]);
```
christyjacob4, line 1790: lets remove this param since its not required anymore

app/controllers/api/account.php
```diff
@@ -2413,72 +2411,63 @@
...
-
-                $className = 'Appwrite\\Auth\\OAuth2\\' . \ucfirst($provider);
-
-                if (!\class_exists($className)) {
-                    throw new Exception(Exception::PROJECT_PROVIDER_UNSUPPORTED);
-                }
-
-                $oauth2 = new $className($appId, $appSecret, '', [], []);
+        $session = null;
+        foreach ($sessions as $key => $loopSession) {
```
christyjacob4, line 2427:
```suggestion
        foreach ($sessions as $key => $value) {
```
lets just use value..keeps it simple

christyjacob4 (changes requested):
Last set of comments and then we can merge

app/controllers/api/account.php
Meldiron: Param removed.

christyjacob4 (approved):

PR #7461, opened 2024-01-17
Fix create/update push target routes

abnegate:
## What does this PR do?

Fix account create push target

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/account.php
```diff
@@ -1770,10 +1769,6 @@
 
         $provider = Authorization::skip(fn () => $dbForProject->getDocument('providers', $providerId));
 
-        if ($provider->isEmpty()) {
-            throw new Exception(Exception::PROVIDER_NOT_FOUND);
-        }
-
```
stnguyen90, line 1776: Why aren't we doing any validation on the provider?

app/controllers/api/account.php
```diff
@@ -1796,8 +1791,8 @@
                     Permission::read(Role::user($user->getId())),
                     Permission::update(Role::user($user->getId())),
                 ],
-                'providerId' => $providerId ?? null,
-                'providerInternalId' => $provider->getInternalId() ?? null,
+                'providerId' => empty($providerId) ? $providerId : null,
+                'providerInternalId' => empty($providerId) ? $provider->getInternalId() : null,
```
stnguyen90, line 1795: Shouldn't this be:

```suggestion
                'providerId' => !empty($providerId) ? $providerId : null,
                'providerInternalId' => !empty($providerId) ? $provider->getInternalId() : null,
```

app/controllers/api/account.php
abnegate: It's optional here

stnguyen90 (approved):

PR #7463, opened 2024-01-18
Add endpoint to list a message's targets

stnguyen90:
## What does this PR do?

This endpoint is needed so that the Console can fetch the associated targets to display when viewing the details of a message or going in to edit a message.

## Test Plan

Manually tested the endpoint:

image

## Related PRs and Issues

None

## Checklist

### DIFF
--- a/app/controllers/api/messaging.php
+++ b/app/controllers/api/messaging.php
@@ -14,6 +14,7 @@ use Appwrite\Utopia\Database\Validator\CustomId;
 use Appwrite\Utopia\Database\Validator\Queries\Messages;
 use Appwrite\Utopia\Database\Validator\Queries\Providers;
 use Appwrite\Utopia\Database\Validator\Queries\Subscribers;
+use Appwrite\Utopia\Database\Validator\Queries\Targets;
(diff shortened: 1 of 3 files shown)


app/controllers/api/messaging.php
```diff
@@ -2640,6 +2641,57 @@
         ]), Response::MODEL_LOG_LIST);
     });
 
+App::get('/v1/messaging/messages/:messageId/targets')
+    ->desc('List message targets')
+    ->groups(['api', 'messaging'])
+    ->label('scope', 'messages.read')
+    ->label('sdk.auth', [APP_AUTH_TYPE_ADMIN, APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'messaging')
+    ->label('sdk.method', 'listMessageTargets')
```
abnegate, line 2650: Since the root resource is message, let's just use `listTargets` for the method

stnguyen90: Updated.

abnegate (commented):
Let's add a test as well

abnegate (approved):

PR #7471, opened 2024-01-20
Fix: Empty values in PATCH of users

Meldiron:
## What does this PR do?

PATCH methods in users service (like updateName, updatePassword, updateEmail, updatePhone) did not allow empty values. But when you create user, you can set empty values.

This PR fixes above inconsistency and allows empty values when updating user details with user service

## Test Plan

- [x] New tests
- [x] Manual QA

https://github.com/appwrite/appwrite/assets/19310830/8408cc6d-5ed1-4815-aa5e-3d4d813ffe9b

## Related PRs and Issues

x

## Checklist


app/controllers/api/users.php
```diff
@@ -1091,6 +1091,16 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+        if (\strlen($password) === 0) {
+            $user
+                ->setAttribute('password', '')
+                ->setAttribute('passwordUpdate', DateTime::now());
+
+            $user = $dbForProject->updateDocument('users', $user->getId(), $user);
+            $queueForEvents->setParam('userId', $user->getId());
+            $response->dynamic($user, Response::MODEL_USER);
+        }
+
```
christyjacob4, line 1103: setting the password to empty should not be allowed.

1. This is not how we should perform a password reset. We already have a procedure for password reset and its using tokens which is more secure and the right way.
2. We have another endpoint in the accounts service to reset password
3. This method is insecure and a security footgun. So its better we dont allow it in the first place

app/controllers/api/users.php
```diff
@@ -1150,21 +1160,23 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
-        }
+        if (\strlen($email) !== 0) {
+            // Makes sure this email is not already used in another identity
+            $identityWithMatchingEmail = $dbForProject->findOne('identities', [
+                Query::equal('providerEmail', [$email]),
+                Query::notEqual('userId', $user->getId()),
+            ]);
+            if ($identityWithMatchingEmail !== false && !$identityWithMatchingEmail->isEmpty()) {
+                throw new Exception(Exception::USER_EMAIL_ALREADY_EXISTS);
+            }
```
christyjacob4, line 1171: Just like passwords, this is not safe and should not be allowed. The only way we can identify users is through their email. If people even use this endpoint by mistake, it means a specific user can never login if email is their only option..

app/controllers/api/users.php
```diff
@@ -1033,7 +1033,7 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
     ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
     ->label('sdk.response.model', Response::MODEL_USER)
     ->param('userId', '', new UID(), 'User ID.')
-    ->param('name', '', new Text(128), 'User name. Max length: 128 chars.')
+    ->param('name', '', new Text(128, 0), 'User name. Max length: 128 chars.')
```
christyjacob4, line 1036: Yea this PR should have only made this possible.. to allow empty names. The other modifications don't feel secure and opens up the attack surface for more vulnerabilities

src/Appwrite/Auth/Validator/Password.php
```diff
@@ -11,6 +11,13 @@
  */
 class Password extends Validator
 {
+    protected bool $allowEmpty;
+
+    public function __construct(bool $allowEmpty = false)
+    {
+        $this->allowEmpty = $allowEmpty;
+    }
+
```
christyjacob4, line 20: This should not be allowed as I shared above

christyjacob4 (changes requested):
Please address the comments.. I think this PR should only address the primary issue which was "Being able to set empty names when updating a user"

The other changes should probably be in a different PR and needs more discussion as we need to evaluate if it will pose any security implications.

app/controllers/api/users.php
Meldiron: We can identify users without email. It's used for anonymous and phone sessions.

app/controllers/api/users.php
eldadfux: @christyjacob4 as Matej mentioned, this is not a password reset feature, but an ability to have the flexibility to convert the account to whatever the dev wants it to be. Its performed only from the server api. 

No reason why we should allow a user to be without a name, email or password but won't allow it to be edited to the same state.

eldadfux (approved):

PR #7482, opened 2024-01-23
Feat: CNAME validation logs

Meldiron:
## What does this PR do?

Adds logic to log cname validation errors into Sentry including many new details. Also adds parameter to SSL CLI task to configure if validation should be skipped

## Test Plan

- [x] Manual QA

Error from API:

CleanShot 2024-01-23 at 16 16 08@2x

CLI command:

CleanShot 2024-01-23 at 16 38 18@2x

Logs on worker:
CleanShot 2024-01-23 at 16 38 26@2x

Error from worker:

CleanShot 2024-01-23 at 16 39 16@2x

## Related PRs and Issues

x

## Checklist

### DIFF
--- a/app/controllers/api/proxy.php
+++ b/app/controllers/api/proxy.php
@@ -14,6 +14,7 @@ use Utopia\Database\Helpers\ID;
 use Utopia\Database\Query;
 use Utopia\Database\Validator\UID;
 use Utopia\Domains\Domain;
+use Utopia\Logger\Log;
 use Utopia\Validator\Domain as ValidatorDomain;
 use Utopia\Validator\Text;
 use Utopia\Validator\WhiteList;
@@ -278,7 +279,8 @@ App::patch('/v1/proxy/rules/:ruleId/verification')
     ->inject('queueForEvents')
     ->inject('project')
     ->inject('dbForConsole')
-    ->action(function (string $ruleId, Response $response, Certificate $queueForCertificates, Event $queueForEvents, Document $project, Database $dbForConsole) {
+    ->inject('log')
+    ->action(function (string $ruleId, Response $response, Certificate $queueForCertificates, Event $queueForEvents, Document $project, Database $dbForConsole, Log $log) {
         $rule = $dbForConsole->getDocument('rules', $ruleId);
 
         if ($rule->isEmpty() || $rule->getAttribute('projectInternalId') !== $project->getInternalId()) {
@@ -298,7 +300,14 @@ App::patch('/v1/proxy/rules/:ruleId/verification')
         $validator = new CNAME($target->get()); // Verify Domain with DNS records
         $domain = new Domain($rule->getAttribute('domain', ''));
 
+        $validationStart = \microtime(true);
         if (!$validator->isValid($domain->get())) {
+            $log->addExtra('dnsTiming', \strval(\microtime(true) - $validationStart));
+            $log->addTag('dnsDomain', $domain->get());
+
+            $error = $validator->getDnsResponse();
+            $log->addExtra('dnsResponse', \is_array($error) ? \json_encode($error) : \strval($error));
+
             throw new Exception(Exception::RULE_VERIFICATION_FAILED);
         }
 
--- a/app/controllers/api/proxy.php
+++ b/app/controllers/api/proxy.php
@@ -305,7 +305,7 @@ App::patch('/v1/proxy/rules/:ruleId/verification')
             $log->addExtra('dnsTiming', \strval(\microtime(true) - $validationStart));
             $log->addTag('dnsDomain', $domain->get());
 
-            $error = $validator->getDnsResponse();
+            $error = $validator->getLogs();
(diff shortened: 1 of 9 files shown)


src/Appwrite/Platform/Workers/Certificates.php
```diff
@@ -187,6 +190,10 @@ private function execute(Domain $domain, Database $dbForConsole, Mail $queueForM
             // Save all changes we made to certificate document into database
             $this->saveCertificateDocument($domain->get(), $certificate, $success, $dbForConsole, $queueForEvents, $queueForFunctions);
```
christyjacob4, line 191:
```suggestion
            $this->saveCertificateDocument($domain->get(), $certificate, $success, $dbForConsole, $queueForEvents, $queueForFunctions);
            throw $e;
```

Why not just this?

src/Appwrite/Platform/Workers/Certificates.php
```diff
@@ -167,6 +169,7 @@ private function execute(Domain $domain, Database $dbForConsole, Mail $queueForM
             $success = true;
         } catch (Throwable $e) {
             $logs = $e->getMessage();
+            $finalException = $e;
```
christyjacob4, line 172:
```suggestion
```

src/Appwrite/Platform/Workers/Certificates.php
```diff
@@ -187,6 +190,10 @@ private function execute(Domain $domain, Database $dbForConsole, Mail $queueForM
             // Save all changes we made to certificate document into database
             $this->saveCertificateDocument($domain->get(), $certificate, $success, $dbForConsole, $queueForEvents, $queueForFunctions);
         }
+
+        if (isset($finalException)) {
+            throw $finalException;
+        }
```
christyjacob4, line 196:
```suggestion
```

We dont need this if we use the code above

app/controllers/api/proxy.php
```diff
@@ -298,7 +300,14 @@
         $validator = new CNAME($target->get()); // Verify Domain with DNS records
         $domain = new Domain($rule->getAttribute('domain', ''));
 
+        $validationStart = \microtime(true);
         if (!$validator->isValid($domain->get())) {
+            $log->addExtra('dnsTiming', \strval(\microtime(true) - $validationStart));
+            $log->addTag('dnsDomain', $domain->get());
+
+            $error = $validator->getDnsResponse();
```
christyjacob4, line 308:
```suggestion
            $error = $validator->getLogs();
```
Why not just this?

app/controllers/general.php
```diff
@@ -615,54 +614,46 @@ function router(App $utopia, Database $dbForConsole, SwooleRequest $swooleReques
         }
 
         if ($logger && $publish) {
-            if ($error->getCode() >= 500 || $error->getCode() === 0) {
```
christyjacob4, line 618: why did we remove this check?

src/Appwrite/Platform/Workers/Certificates.php
Meldiron: Its in `finally`, we dont have exception there.
I don't want to re-throw inside catch, as I am not sure if finally would run after that properly. And Im afraid to experiment since this is untested piece of code.

app/controllers/general.php
Meldiron: Because it would never throw 400 error code, even if we set it to `true` in the exceptions definition.

Here we relay on `publish` having proper value, and this check was moved here https://github.com/appwrite/appwrite/pull/7482/files#diff-89eb85ffb7fda44e8ce1c91a90f72616a3d1bcaff0dbdf79486421df81b26f80R248

src/Appwrite/Platform/Workers/Certificates.php
Meldiron: Hmm or maybe finally does see exception? 🤔

CleanShot 2024-01-24 at 13 38 59@2x

@christyjacob4  can you confirm you are familiar with this behaviour? First time seeing this, worried to use it.

Meldiron: CleanShot 2024-01-24 at 13 40 05@2x

It still causes warning if no exception is there, so it doesn't solve issue with having to do if(isset))

christyjacob4: Okay then you can do this 
```php
try {
    // some code
} catch (Exception $e) {
  // Handle Error
   throw $e;
} finally {
   // Finally is always executed 
}
```

app/controllers/general.php
christyjacob4: Then we need to modify the check here.. We should not change how $publish works inside the Exception class..


```php
if ($logger && $publish) {
            if ($publish || $error->getCode() >= 500 || $error->getCode() === 0) {
```

This code should work for our case

Meldiron: Partly moved to exception, partly left in general.php

christyjacob4 (approved):

PR #7487, opened 2024-01-24
Feat block countries from certain regions

vermakhushboo:
## What does this PR do?

Add logic in `auth` group to block requests from certain IP Addresses

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/shared/api.php
```diff
@@ -592,3 +593,19 @@
...
     });
+
+App::init()
+    ->groups(['restrict'])
+    ->inject('request')
+    ->inject('geodb')
+    ->action(function (Request $request, Reader $geodb) {
+        if (!empty(app::getEnv('_APP_RESTRICTED_COUNTRIES', ''))) {
+            $countries = explode(',', App::getEnv('_APP_RESTRICTED_COUNTRIES', ''));
+            $record = $geodb->get($request->getHeader('x-forwarded-for'));
```
Meldiron, line 604:
```suggestion
            $record = $geodb->get($request->getIP());
```

eldadfux (commented):
Not sure we need to impose the limit on the users API. Also, I think we can settle on limiting access to the console and leave end-users signup open. 

Since this will be a console only limit, I would also consider using the existing pattern of `_APP_CONSOLE_WHITELIST` for the env prefix.

.env
```diff
@@ -103,3 +103,4 @@ _APP_ASSISTANT_OPENAI_API_KEY=
 _APP_MESSAGE_SMS_TEST_DSN=
 _APP_MESSAGE_EMAIL_TEST_DSN=
 _APP_MESSAGE_PUSH_TEST_DSN=
+_APP_RESTRICTED_COUNTRIES=AQ
```
christyjacob4, line 106:
```suggestion
_APP_COUNTRIES_DENYLIST=AQ
```

app/controllers/shared/api.php
```diff
@@ -592,3 +593,19 @@
             throw new Exception(Exception::GENERAL_USAGE_DISABLED);
         }
     });
+
+App::init()
+    ->groups(['restrict'])
+    ->inject('request')
+    ->inject('geodb')
+    ->action(function (Request $request, Reader $geodb) {
+        if (!empty(app::getEnv('_APP_RESTRICTED_COUNTRIES', ''))) {
```
christyjacob4, line 602:
```suggestion
        $denylist = App::getEnv('_APP_RESTRICTED_COUNTRIES', '');
        if (!empty($denylist)) {
```

app/controllers/shared/api.php
```diff
@@ -592,3 +593,19 @@
...
+            $countries = explode(',', App::getEnv('_APP_RESTRICTED_COUNTRIES', ''));
```
christyjacob4, line 603:
```suggestion
            $countries = explode(',', $denylist);
```

app/controllers/shared/api.php
```diff
@@ -592,3 +593,19 @@
...
+            $record = $geodb->get($request->getIP());
+            $country = $record['country']['iso_code'];
+            $countryName = $record['country']['names']['en'];
+            if (in_array($country, $countries)) {
+                throw new Exception(Exception::GENERAL_ACCESS_FORBIDDEN, "Sorry, access from $countryName is restricted");
```
christyjacob4, line 608: We need to return a 503 service unavailable error in this case. Check the errors.php file to see if we have something like that already 👍

tests/e2e/Services/Account/AccountBase.php
```diff
@@ -43,6 +43,20 @@ public function testCreateAccount(): array
         /**
          * Test for FAILURE
          */
+        $response = $this->client->call(Client::METHOD_POST, '/account', array_merge([
```
christyjacob4, line 46: Lets add a comment indicating what we are doing in this test case

app/controllers/shared/api.php
vermakhushboo: We don't have any relevant 503 errors at the moment. We'd need a new one.

vermakhushboo: Creating one with 451 as discussed

Meldiron (approved):

eldadfux (approved):

PR #7491, opened 2024-01-25
Add client target delete

abnegate:
## What does this PR do?

Add client side target delete route so that if a user logs out and another user logs in, we can detect a change in userID, delete the old target and create a new one. This will stop multiple users from having different targets with the same push token.

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/account.php
```diff
@@ -3412,10 +3342,114 @@
...
+        $queueForDeletes
+            ->setType(DELETE_TYPE_DOCUMENT)
+            ->setDocument($user);
+
+        $queueForEvents
+            ->setParam('userId', $user->getId())
+            ->setPayload($response->output($user, Response::MODEL_USER));
+
+        $response->noContent();
+    });
```
TorstenDittmann, line 3380: shouldn't this be already there?

abnegate: Yep it is, I just moved all the push related routes to the end so docs show them together, if you look near the bottom of the diff you'll see this one removed there

app/controllers/api/account.php
```diff
@@ -3469,39 +3500,44 @@
...
+    ->inject('request')
     ->inject('response')
     ->inject('dbForProject')
-    ->inject('queueForEvents')
-    ->inject('queueForDeletes')
-    ->action(function (Document $user, Response $response, Database $dbForProject, Event $queueForEvents, Delete $queueForDeletes) {
-        if ($user->isEmpty()) {
-            throw new Exception(Exception::USER_NOT_FOUND);
+    ->action(function (string $targetId, Event $queueForEvents, Delete $queueForDeletes, Document $user, Request $request, Response $response, Database $dbForProject) {
+        $target = Authorization::skip(fn() => $dbForProject->getDocument('targets', $targetId));
```
stnguyen90, line 3524: How are we making sure a user is only deleting their own target?

app/controllers/api/account.php
```diff
@@ -3469,39 +3500,44 @@
...
 
         $queueForDeletes
-            ->setType(DELETE_TYPE_DOCUMENT)
-            ->setDocument($user);
+            ->setType(DELETE_TYPE_TARGET)
+            ->setDocument($target);
 
         $queueForEvents
             ->setParam('userId', $user->getId())
-            ->setPayload($response->output($user, Response::MODEL_USER));
```
stnguyen90, line 3504: don't we need to set the payload?

stnguyen90 (changes requested):

app/controllers/api/account.php
abnegate: Addressed

app/controllers/api/account.php
abnegate: Addressed

app/controllers/api/account.php
```diff
@@ -3470,39 +3501,49 @@
...
-            throw new Exception(Exception::USER_NOT_FOUND);
+    ->action(function (string $targetId, Event $queueForEvents, Delete $queueForDeletes, Document $user, Request $request, Response $response, Database $dbForProject) {
+        $target = Authorization::skip(fn() => $dbForProject->getDocument('targets', $targetId));
+
+        if ($target->isEmpty()) {
+            throw new Exception(Exception::USER_TARGET_NOT_FOUND);
         }
 
-        $dbForProject->deleteDocument('users', $user->getId());
+        if ($user->getId() !== $target->getAttribute('userId')) {
```
stnguyen90, line 3531: should we use `userInternalId`?

abnegate: I don't think it's that important to just for single comparisons

stnguyen90: What do you mean? Why?

abnegate: We're just checking if they're equal, the difference is going to be unnoticeable. The benefit of using internal IDs comes when fetching from the database, as an incremental integer match is much quicker than a string match there

stnguyen90: I brought up checking using internal id

```suggestion
        if ($user->getInternalId() !== $target->getAttribute('userInternalId')) {
```

to make sure that the target belongs to the right user even if the user was recreated with the same ID. It not as detrimental because the target should be queued for deletion anyways, but I brought it up because we try to compare using internal IDs most of the time.

stnguyen90 (commented):
See https://github.com/appwrite/appwrite/pull/7491#discussion_r1475291833

app/controllers/api/account.php
abnegate: Ohh got you, will commit :+1:

abnegate:
Superseded by https://github.com/appwrite/appwrite/pull/7540

PR #7495, opened 2024-01-26
1.5 QA bug fixes

Meldiron:
## What does this PR do?

- Removes no longer needed translations
- Fixes unwanted empty spaces in email copy about client device
- Blocks email otp if disabled on project
- Fixes nested templates and problematic plaintext mails
- Fixes unescaped HTML in magic URL & email otp mails

## Test Plan

- [x] Manual QA

## Related PRs and Issues

x

## Checklist


src/Appwrite/Platform/Workers/Mails.php
```diff
@@ -78,8 +78,9 @@ public function action(Message $message, Registry $register, Log $log): void
         $bodyTemplate = Template::fromFile($bodyTemplate);
         $bodyTemplate->setParam('{{body}}', $body, escapeHtml: false);
         foreach ($variables as $key => $value) {
-            // TODO: hotfix for redirect param
-            $bodyTemplate->setParam('{{' . $key . '}}', $value, escapeHtml: $key !== 'redirect');
+            // TODO: hotfix for html params
+            $htmlParams = [ 'redirect', 'agentDevice', 'agentClient', 'agentOs', 'phrase' ];
```
TorstenDittmann, line 82: why cant those be escaped?

I think all of those expect redirect should be scaped.

Meldiron: They have boldness, so `<b>SOMETHING</b>`. Without it the email looks very plain, and I couldn't find better way to do it

Meldiron: This is what it looks likewithout changes in my PR
CleanShot 2024-01-31 at 10 05 57@2x

app/controllers/shared/api/auth.php
```diff
@@ -31,6 +31,12 @@
                 }
                 break;
 
+            case 'emailOtp':
```
TorstenDittmann, line 34: can we call it `email-otp` so it aligns with the other values like `magic-url`?

stnguyen90:
@Meldiron, is this PR still needed?

stnguyen90:
I'm closing this stale issue. Please create issues to track anything that still needs to be done.

PR #7500, opened 2024-01-28
PEA-233-prevent console user deletion before deleting their team

lohanidamodar:
## What does this PR do?

- Deleting a console user, throw an error if they are a member of active organization

## Test Plan

- updated the existing `testAccountDelete` in AccountConsoleClientTest.php

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/config/errors.php
+++ b/app/config/errors.php
@@ -251,7 +251,7 @@ return [
         'description' => 'User phone is already verified',
         'code' => 409
     ],
-    Exception::USER_DELETION_WITH_ACTIVE_TEAMS => [
+    Exception::USER_DELETION_PROHIBITED => [
         'name' => Exception::USER_DELETION_WITH_ACTIVE_TEAMS,
         'description' => 'User deletion is not allowed for users with active memberships. Please delete all confirmed memberships before deleting the account.',
         'code' => 400
--- a/app/config/errors.php
+++ b/app/config/errors.php
@@ -252,7 +252,7 @@ return [
         'code' => 409
     ],
     Exception::USER_DELETION_PROHIBITED => [
-        'name' => Exception::USER_DELETION_WITH_ACTIVE_TEAMS,
+        'name' => Exception::USER_DELETION_PROHIBITED,
         'description' => 'User deletion is not allowed for users with active memberships. Please delete all confirmed memberships before deleting the account.',
         'code' => 400
     ],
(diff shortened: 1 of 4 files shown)


app/controllers/api/account.php
```diff
@@ -3031,15 +3031,27 @@
...
+    ->action(function (Document $user, Document $project, Response $response, Database $dbForProject, Event $queueForEvents, Delete $queueForDeletes) {
         if ($user->isEmpty()) {
             throw new Exception(Exception::USER_NOT_FOUND);
         }
 
+        if ($project->getId() === 'console') {
+            // get all memberships
+            $memberships = $user->getAttribute('memberships', []);
+            foreach ($memberships as $membership) {
+                // prevent deletion if at lease one active membership
```
christyjacob4, line 3048:
```suggestion
                // prevent deletion if at least one active membership
```

app/config/errors.php
```diff
@@ -251,6 +251,11 @@
         'description' => 'User phone is already verified',
         'code' => 409
     ],
+    Exception::USER_DELETION_WITH_ACTIVE_TEAMS => [
+        'name' => Exception::USER_DELETION_WITH_ACTIVE_TEAMS,
+        'description' => 'Delete all your teams before trying to delete your account.',
```
christyjacob4, line 256:
```suggestion
        'description' => 'User deletion is not allowed for users with active memberships. Please delete all confirmed memberships before deleting the account.',
```

app/config/errors.php
```diff
@@ -251,6 +251,11 @@
         'description' => 'User phone is already verified',
         'code' => 409
     ],
+    Exception::USER_DELETION_WITH_ACTIVE_TEAMS => [
```
christyjacob4, line 254:
```suggestion
    Exception::USER_DELETION_NOT_ALLOWED => [
```

app/controllers/api/account.php
```diff
@@ -3031,15 +3031,27 @@
...
+            // get all memberships
+            $memberships = $user->getAttribute('memberships', []);
+            foreach ($memberships as $membership) {
+                // prevent deletion if at least one active membership
+                if ($membership->getAttribute('confirm', false)) {
+                    throw new Exception(Exception::USER_DELETION_WITH_ACTIVE_TEAMS);
+                }
+            }
+        }
+
```
christyjacob4, line 3054: actually @lohanidamodar this will not work. The memberships attribute of users is not used. We have to list memberships from the memberships collection

lohanidamodar: This is handled by sub-query, so we should still be good.

christyjacob4 (approved):

PR #7503, opened 2024-01-28
Usage queue poc

shimonewman:
The usage queue is constructed in two pieces:
 A worker and a swoole ticker that aggregates stats and fires on a fixed interval.
The worker sync behavior versus the ticker async behavior creates a conflict when the queue is busy and generates.
Swoole errors, which result in corrupted data.
In an attempt to solve it, we  refactored the architecture to use two queues :
The first queue aggregates the metrics based on the number of keys || last sent timestamp (the one that happens first)
and send them to the second queue, which inserts the stats into the DB.

The new architecture uses two sync elements and eliminates the necessary  to use a static array  that also conflicts
with multithreading.

From the test (e2e , manuel, k6) the results look promising


src/Appwrite/Platform/Workers/UsageDump.php
```diff
@@ -0,0 +1,116 @@
...
+                                '$id' => $id,
+                                'period' => $period,
+                                'time' => $time,
+                                'metric' => $key,
+                                'value' => $value,
+                                'region' => App::getEnv('_APP_REGION', 'default'),
+                            ]));
+                        } catch (Duplicate $th) {
+                            if ($value < 0) {
+                                //Todo debug (to be removed later @shimon)
```
eldadfux, line 89: Remove left overs

eldadfux (changes requested):

src/Appwrite/Platform/Workers/Usage.php
```diff
@@ -4,22 +4,22 @@
...
-        'inf' => '0000-00-00 00:00'
-    ];
-
-    protected const INFINITY_PERIOD = '_inf_';
+    private array $stats = [];
+    private int $lastTriggeredTime = 0;
+    private int $keys = 0;
+    private const INFINITY_PERIOD = '_inf_';
+    private const KEYS_THRESHOLD = 5;
+    private const KEYS_SENT_THRESHOLD = 60;
```
christyjacob4, line 19: Lets increase these values

src/Appwrite/Platform/Workers/Usage.php
```diff
@@ -4,22 +4,22 @@
 
 use Exception;
 use Utopia\CLI\Console;
-use Utopia\Database\Database;
-use Utopia\Database\DateTime;
 use Utopia\Database\Document;
 use Utopia\Platform\Action;
+use Appwrite\Event\UsageDump;
 use Utopia\Queue\Message;
```
christyjacob4, line 11: Add a todo that we still need to handle the case of deployments performed when usage us being aggregated.

christyjacob4: Add another Todo to refactor the names of the two usage workers

src/Appwrite/Platform/Workers/Usage.php
christyjacob4, line 19: Lets use this  variable we already have
 $interval = (int) App::getEnv('_APP_USAGE_AGGREGATION_INTERVAL', '60000');

app/controllers/api/users.php
```diff
@@ -1189,15 +1189,16 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
         // clone user object to send to workers
         $clone = clone $user;
 
-        $dbForProject->deleteDocument('users', $userId);
-
-        $queueForDeletes
-            ->setType(DELETE_TYPE_DOCUMENT)
-            ->setDocument($clone);
+        $affected = $dbForProject->deleteDocument('users', $userId);
```
christyjacob4, line 1192: Lets move this change to a separate PR, where we can apply this change to all the endpoints that require it

docker-compose.yml
```diff
@@ -84,6 +84,7 @@ services:
       - ./public:/usr/src/code/public
       - ./src:/usr/src/code/src
       - ./dev:/usr/src/code/dev
+      #- ./vendor/utopia-php/database:/usr/src/code/vendor/utopia-php/database
```
christyjacob4, line 87: lets remove this line

christyjacob4 (approved):

PR #7510, opened 2024-01-29
Add message delete route

abnegate:
## What does this PR do?

- Add delete message endpoint

## Test Plan

- Added test for delete message cases

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/messaging.php
```diff
@@ -3266,3 +3266,51 @@
...
+    ->desc('Delete a message')
+    ->groups(['api', 'messaging'])
+    ->label('audits.event', 'message.delete')
+    ->label('audits.resource', 'message/{request.route.messageId}')
+    ->label('event', 'messages.[messageId].delete')
+    ->label('scope', 'messages.write')
+    ->label('sdk.auth', [APP_AUTH_TYPE_ADMIN, APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'messaging')
+    ->label('sdk.method', 'delete')
+    ->label('sdk.description', '/docs/references/messaging/delete-message.md')
```
stnguyen90, line 3280: Shouldn't this PR include this file?

app/controllers/api/messaging.php
```diff
@@ -3266,3 +3266,51 @@
...
+        $message = $dbForProject->getDocument('messages', $messageId);
+
+        if ($message->isEmpty()) {
+            throw new Exception(Exception::MESSAGE_NOT_FOUND);
+        }
+
+        switch ($message->getAttribute('status')) {
+            case MessageStatus::PROCESSING:
+                throw new Exception(Exception::MESSAGE_ALREADY_SCHEDULED);
+            case MessageStatus::SCHEDULED:
```
stnguyen90, line 3298: Should we also check if the status is scheduled and `scheduledAt` is after now?

app/controllers/api/messaging.php
abnegate: Added

app/controllers/api/messaging.php
abnegate: Yep makes sense, added

stnguyen90 (approved):

PR #7512, opened 2024-01-29
Feat: API Tokens

Meldiron:
## What does this PR do?

Introduces API Tokens to function executions. They are automatically provided short-liked JWT tokens that developer can use instead of creating and maintaining API keys for their functions.

## Test Plan

- [x] Implement new tests

## Related PRs and Issues

x

## Checklist


app/config/collections.php
```diff
@@ -2099,7 +2099,18 @@
                 'required' => false,
                 'default' => null,
                 'filters' => [],
-            ]
+            ],
+            [
+                '$id' => ID::custom('apiTokenScopes'),
```
eldadfux, line 2104: This doesn't really fit well with the naming style of other attributes we have in this collection and others. We would usually try to avoid namespaces as much as possible and call something like this simply `scopes`.

app/controllers/api/functions.php
```diff
@@ -150,6 +150,7 @@
     ->param('logging', true, new Boolean(), 'Whether executions will be logged. When set to false, executions will not be logged, but will reduce resource used by your Appwrite project.', true)
     ->param('entrypoint', '', new Text(1028, 0), 'Entrypoint File. This path is relative to the "providerRootDirectory".', true)
     ->param('commands', '', new Text(8192, 0), 'Build Commands.', true)
+    ->param('apiTokenScopes', [], new ArrayList(new WhiteList(array_keys(Config::getParam('scopes')), true), APP_LIMIT_ARRAY_PARAMS_SIZE), 'List of scopes allowed for API Token auto-generated for every execution. Maximum of ' . APP_LIMIT_ARRAY_PARAMS_SIZE . ' scopes are allowed.', true)
```
eldadfux, line 153: Same as the above comment. I would just call this `scopes`.

app/controllers/api/functions.php
```diff
@@ -1569,6 +1573,14 @@
...
         }
 
+        $jwtExpiry = $function->getAttribute('timeout', 900);
+        $jwtObj = new JWT(App::getEnv('_APP_OPENSSL_KEY_V1'), 'HS256', $jwtExpiry, 10);
+        $apiToken = $jwtObj->encode([
+            'projectId' => $project->getId(),
+            'scopes' => $function->getAttribute('apiTokenScopes', [])
+        ]);
+
+        $headers['x-appwrite-api-token'] = $apiToken;
```
eldadfux, line 1583: api-token vs api-key is extremely confusing. This is exactly how we create complexity where two parallel terms split the user journey in two different paths and they are forced to memorize an Appwrite specific patch. 

I understand we can't create a regular API key for every function as this will create a huge clutter, but I'm curious by the idea of using the same `x-appwrite-key` header and SDK method with some prefix (so we can parse it internally and validate it differently) and let it behave exactly the same for the user.

app/controllers/general.php
```diff
@@ -531,6 +533,40 @@ function router(App $utopia, Database $dbForConsole, SwooleRequest $swooleReques
             }
         }
 
+        // API Token authentication (like API Key but JWT short-term)
```
eldadfux, line 536: Internally maybe we can call it dynamic key vs the regular key which is persistent.

eldadfux (changes requested):

app/controllers/api/functions.php
Meldiron: What about using Authentication header? 
`Authorization: <auth-scheme> <authorization-parameters>`

eldadfux: This could be nice as we move down the line, my biggest worry is to have to explain 2 very similar methods, I would love us to abstract that complexity from developers using the API or SDK with the `setKey()` method.

Meldiron: Okay, Ill use setKey and `x-appwrite-key` for both.
Would it be fine to have a bit of magic for recognizing if it's regular or dynamic key? My initial idea is to first check against API keys on project (we have project document already, so it's not network request). If we don't find it that way, then we try to parse it as JWT. If both fails, then we consider we don't have any key.

eldadfux: I think performance wise it might be better if we add some prefix or other metadata behind the base64 encoding so we can parse the key in the correct way faster.

app/controllers/shared/api.php
```diff
@@ -197,54 +199,89 @@
...
-                    'name' => $project->getAttribute('name', 'Untitled'),
-                ]);
-
-                $role = Auth::USER_ROLE_APPS;
-                $scopes = \array_merge($roles[$role]['scopes'], $key->getAttribute('scopes', []));
-
-                $expire = $key->getAttribute('expire');
-                if (!empty($expire) && $expire < DateTime::formatTz(DateTime::now())) {
-                    throw new Exception(Exception::PROJECT_KEY_EXPIRED);
+            if(str_contains($authKey, '.')) {
```
eldadfux, line 209: How does the new key looks like? Maybe we should add a more detailed docblock here to explain behavior

src/Appwrite/Platform/Workers/Functions.php
```diff
@@ -369,8 +378,13 @@ private function execute(
             $vars[$var->getAttribute('key')] = $var->getAttribute('value', '');
         }
 
+        $protocol = System::getEnv('_APP_OPTIONS_FORCE_HTTPS') == 'disabled' ? 'http' : 'https';
+        $hostname = System::getEnv('_APP_DOMAIN');
+        $endpoint = $protocol . '://' . $hostname . "/v1";
+
         // Appwrite vars
         $vars = \array_merge($vars, [
+            'APPWRITE_FUNCTION_ENDPOINT' => $endpoint,
```
eldadfux, line 387: Name might be a bit misleading, maybe add API before endpoint.
```suggestion
            'APPWRITE_FUNCTION_API_ENDPOINT' => $endpoint,
```

or

```suggestion
            'APPWRITE_FUNCTION_API' => $endpoint,
```

app/controllers/shared/api.php
Meldiron: It's either API key, or JWT token. JWT token consists of 3 (or 4) parts separated by comma. This is the simplest check I could come up with.
Has dot: Is JWT
Doesnt have dot: Is Appwrite API key (hexadecimal)

eldadfux: Should we start adding some prefix to keys for future implementations? We might have another one when we want to introduce application to use Appwrite as an OAuth adapter as one example.

Meldiron: As discussed on Discord, we now have prefixes `standard_` and `dynamic_` for all API keys.

src/Appwrite/Platform/Workers/Functions.php
```diff
@@ -369,8 +378,13 @@ private function execute(
             $vars[$var->getAttribute('key')] = $var->getAttribute('value', '');
         }
 
+        $protocol = System::getEnv('_APP_OPTIONS_FORCE_HTTPS') == 'disabled' ? 'http' : 'https';
+        $hostname = System::getEnv('_APP_DOMAIN');
+        $endpoint = $protocol . '://' . $hostname . "/v1";
+
         // Appwrite vars
         $vars = \array_merge($vars, [
+            'APPWRITE_FUNCTION_API_ENDPOINT' => $endpoint,
```
eldadfux, line 387: Let's not forget to update our docs.

.env
```diff
@@ -15,7 +15,7 @@ _APP_OPTIONS_ROUTER_PROTECTION=disabled
 _APP_OPTIONS_FORCE_HTTPS=disabled
 _APP_OPTIONS_FUNCTIONS_FORCE_HTTPS=disabled
 _APP_OPENSSL_KEY_V1=your-secret-key
-_APP_DOMAIN=localhost
+_APP_DOMAIN=traefik
```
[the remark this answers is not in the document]
Meldiron: Yes, domain goes to function as env variable. In case of Cloud that would be `cloud.appwrite.io`. We need this here so test pass - otherwise test using ENV variable for endpoint API will not pass

src/Appwrite/Platform/Workers/Functions.php
Meldiron: Made Linear task to keep it in mind

app/init.php
```diff
@@ -208,6 +208,9 @@
 const MESSAGE_TYPE_EMAIL = 'email';
 const MESSAGE_TYPE_SMS = 'sms';
 const MESSAGE_TYPE_PUSH = 'push';
+// API key types
+const API_KEY_STANDARD = 'standard';
+const API_KEY_DYNAMIC = 'dynamic';
```
christyjacob4, line 213: What is the size of the API key ? Will we hit any limits on the API key length during any stage? Here are the scenarios I can think of.
1. When storing the encrypted API key in database
2. When sending the API key through the header

christyjacob4: We can also use smaller prefixes like stripe. I believe they use `pk` for production key and `dk` for development key

src/Appwrite/Platform/Workers/Functions.php
```diff
@@ -282,6 +283,14 @@ private function execute(
 
         $runtime = $runtimes[$function->getAttribute('runtime')];
 
+        $jwtExpiry = $function->getAttribute('timeout', 900);
+        $jwtObj = new JWT(System::getEnv('_APP_OPENSSL_KEY_V1'), 'HS256', $jwtExpiry, 10);
+        $apiKey = $jwtObj->encode([
+            'projectId' => $project->getId(),
+            'scopes' => $function->getAttribute('scopes', [])
+        ]);
```
christyjacob4, line 291: Shouldn't we save this API key to the db as well? How do we validate this key?

app/init.php
Meldiron: i dont like `pk` and `dk`, and openAI's `sk`.. I never know what those mean. to me `sk` means `Slovakia` 😅
I would say going verbose isn't problem, considering there is `_` so it is consider one word when you double-click to copy

Meldiron: JWT keys arent stored in DB, should be fine.
Header can go up to KBs of size, also shouldn't be an issue. If we ever get some kind of report, I am sure we can shorten JWT key with some configuration.

Normla keys will only get few letters prefix, shouldn't be length issue.

src/Appwrite/Platform/Workers/Functions.php
Meldiron: If you are worried about using the token when it's not expired but resource is deleted, this wont occur, because you would need to delete function or a project. We don't have a way to delete a key currently.

app/controllers/shared/api.php
```diff
@@ -195,56 +197,99 @@
...
-
-                $expire = $key->getAttribute('expire');
-                if (!empty($expire) && $expire < DateTime::formatTz(DateTime::now())) {
-                    throw new Exception(Exception::PROJECT_KEY_EXPIRED);
+            if(!\str_contains($apiKey, '_')) {
+                $keyType = API_KEY_STANDARD;
+                $authKey = $apiKey;
+            } else {
+                [ $keyType, $authKey ] = \explode('_', $apiKey, 2);
+            }
```
christyjacob4, line 214: Lets add a todo to remove this code once we have deprecated the old syntax.

christyjacob4 (approved):

PR #7524, opened 2024-01-31
feat: use geosms 2

loks0n:


src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -116,4 +113,71 @@ public function action(Message $message): void
             throw new Exception('Error sending message: ' . $error->getMessage(), 500);
         }
     }
+
+    protected static function createAdapterFromDSN(DSN $dsn): SMSAdapter
+    {
+        $from = empty($dsn->getParam('from', '')) ? null : $dsn->getParam('from', '');
```
christyjacob4, line 119: I thought `from` is always fetched from the environment variable? How is it populated in this property in the DSN ? Im not sure this is documented

loks0n: By default, it is the _APP_SMS_FROM variable.

If you have ?from= set in the DSN, this overrides it.

Will document this.

PR #7533, opened 2024-02-01
Feat: Rename security phrases

Meldiron:
## What does this PR do?

Renames from `securityPhrase` to `phrase` for simplicity.

## Test Plan

Existing tests updated

## Related PRs and Issues

x

## Checklist


loks0n (approved):

app/controllers/api/account.php
```diff
@@ -990,7 +990,7 @@
...
     ->param('email', '', new Email(), 'User email.')
     ->param('url', '', fn($clients) => new Host($clients), 'URL to redirect the user back to your app from the magic URL login. Only URLs from hostnames in your project platform list are allowed. This requirement helps to prevent an [open redirect](https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html) attack against your project API.', true, ['clients'])
-    ->param('securityPhrase', false, new Boolean(), 'Toggle for security phrase. If enabled, email will be send with a randomly generated phrase and the phrase will also be included in the response. Confirming phrases match increases the security of authentication flow.', true)
+    ->param('phrase', false, new Boolean(), 'Toggle for security phrase. If enabled, email will be send with a randomly generated phrase and the phrase will also be included in the response. Confirming phrases match increases the security of authentication flow.', true)
```
eldadfux, line 993:
```suggestion
    ->param('phrase', false, new Boolean(), 'Toggle for security phrase. If enabled, email will be send with a randomly generated phrase and the phrase will also be included in the response. Confirming phrases match increases the security of the authentication flow.', true)
```

We can also say `your` instead of `the`

eldadfux (commented):
We can also change the class name for the class that generate the phrases to keep things consistent.

eldadfux (approved):

Issue #5901, opened 2023-07-28
🐛 Bug Report: When trying to resend email in Organization tab it throws AppwriteException


ZachHandley:
### 👟 Reproduction steps

Try to hit resend on an unclaimed email invite

### 👍 Expected behavior

resend the email

### 👎 Actual Behavior

Appwrite Exception 
![image](https://github.com/appwrite/appwrite/assets/16495679/3c54598b-8b6b-4bd0-9783-98ac80c1a2cb)

### 🎲 Appwrite version

Version 1.3.x

### 💻 Operating system

Linux

### 🧱 Your Environment

1.3.8, Cloudflare, Appwrite is self hosted as a reverse proxy using Nginx

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

safwanyp:
On the backend, the endpoint controller tries to create a whole new membership, which is expected from it: 
https://github.com/appwrite/appwrite/blob/06570a081a3832b94ee6e5307002d55738be079c/app/controllers/api/teams.php#L525-L535

Considering the fact that the email delivery is in the same endpoint, can't we do something like:
- Try to create a new membership
- If membership exists for given user, resend invite instead of throwing `TEAM_INVITE_ALREADY_EXISTS` exception. 

Looking at the console API calls, this endpoint would be called only when a brand new membership is being created and when trying to resend an email, so this should be a viable fix. What do you think?

safwanyp:
I've seen the code and it seems like letting the code keep running on catching a `Duplicate` exception works. 

Changed the code from this: 
```php
try { 
   $membership = Authorization::skip(fn() => $dbForProject->createDocument('memberships', $membership)); 
} catch (Duplicate $th) { 
   throw new Exception(Exception::TEAM_INVITE_ALREADY_EXISTS); 
} 
```
to this: 
```php
try {
    $membership = Authorization::skip(fn() => $dbForProject->createDocument('memberships', $membership));
} catch (Duplicate $th) {
    null;
} catch (Exception $th) {
    throw new Exception($th->getMessage());
}
```

I haven't been able to generate a case where an error actually gets thrown, so would appreciate feedback on the above approach!

PR #7543, opened 2024-02-05
Bug 5901 fix resend invitation from create invite

ZachHandley:
## What does this PR do?

Fixes bug mentioned here (https://github.com/appwrite/appwrite/issues/5901) by me -- basically it adds a new call into the `createMembership` endpoint that checks the database for an unconfirmed membership with the same userEmail. If they don't exist, it creates a membership like normal, which will throw a duplicate error like normal if the users email exists.

If they do exist, it will set the `$membership` to the found membership, and set the attribute `invited` to `DateTime::now()`, then updates the user

## Test Plan

The only possible problem this could have was in the `$dbForProject->updateDocument` (with the $collectionId or something being included like in the client SDK's was my thought) but I have looked at the other code in the `teams.php` file and they seem to be doing things the same way, so it shouldn't conflict at all.

## Related PRs and Issues

- [(Related PR or issue)](https://github.com/appwrite/appwrite/issues/5901)

## Checklist

### DIFF
--- a/app/controllers/api/teams.php
+++ b/app/controllers/api/teams.php
@@ -69,7 +69,7 @@ App::post('/v1/teams')
         $teamId = $teamId == 'unique()' ? ID::unique() : $teamId;
 
         try {
-            $team = Authorization::skip(fn() => $dbForProject->createDocument('teams', new Document([
+            $team = Authorization::skip(fn () => $dbForProject->createDocument('teams', new Document([
                 '$id' => $teamId,
                 '$permissions' => [
                     Permission::read(Role::team($teamId)),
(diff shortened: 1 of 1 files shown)


stnguyen90 (changes requested):
@ZachHandley, sorry, would you be able to fix the merge conflicts?

ZachHandley:
> @ZachHandley, sorry, would you be able to fix the merge conflicts?

No need to apologize, of course!!

ZachHandley:
@stnguyen90 https://github.com/appwrite/appwrite/pull/7790 when I merged current and my branch it closed this one, dunno why

PR #7547, opened 2024-02-06
Add health certificate validity endpoint

vermakhushboo:
## What does this PR do?

Get status of certificate for a domain to check whether it is still valid or expired.

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/controllers/api/health.php
+++ b/app/controllers/api/health.php
@@ -7,6 +7,7 @@ use Appwrite\Utopia\Response;
 use Utopia\App;
 use Utopia\Config\Config;
 use Utopia\Database\Document;
+use Utopia\Domains\Validator\PublicDomain;
 use Utopia\Pools\Group;
 use Utopia\Queue\Client;
 use Utopia\Queue\Connection;
@@ -14,7 +15,9 @@ use Utopia\Registry\Registry;
 use Utopia\Storage\Device;
 use Utopia\Storage\Device\Local;
 use Utopia\Storage\Storage;
+use Utopia\Validator\Domain;
 use Utopia\Validator\Integer;
+use Utopia\Validator\Multiple;
 use Utopia\Validator\Text;
 use Utopia\Validator\WhiteList;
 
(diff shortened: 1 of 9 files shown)


app/controllers/api/health.php
```diff
@@ -389,6 +392,56 @@
...
+    ->groups(['api', 'health'])
+    ->label('scope', 'health.read')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'health')
+    ->label('sdk.method', 'getCertificate')
+    ->label('sdk.description', '/docs/references/health/get-certificate.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_HEALTH_CERTIFICATE)
+    ->param('domain', null, new Multiple([new Domain(), new PublicDomain()]), Multiple::TYPE_STRING, 'Domain name')
```
christyjacob4, line 406: Should the default value be `null`? How does the endpoint behave when no`domain` is passed ?

composer.json
```diff
@@ -50,7 +50,7 @@
         "utopia-php/cli": "0.15.*",
         "utopia-php/config": "0.2.*",
         "utopia-php/database": "0.45.*",
-        "utopia-php/domains": "0.3.*",
+        "utopia-php/domains": "0.5.*",
```
christyjacob4, line 53: is this update expected? what did we include in the new version?

tests/e2e/Services/Health/HealthCustomServerTest.php
```diff
@@ -424,4 +424,67 @@ public function testStorageAntiVirusSuccess(): array
...
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()), []);
+
+        $this->assertEquals(200, $response['headers']['status-code']);
+
+        /**
+         * Test for FAILURE
+         */
+        $response = $this->client->call(Client::METHOD_GET, '/health/certificate?domain=localhost', array_merge([
+            'content-type' => 'application/json',
```
christyjacob4, line 468: Can we also add a test for an empty domain?

src/Appwrite/Extend/Exception.php
```diff
@@ -235,6 +235,8 @@ class Exception extends \Exception
 
     /** Health */
     public const QUEUE_SIZE_EXCEEDED                 = 'queue_size_exceeded';
```
christyjacob4, line 237: While we are at it can we also update the key here to `HEALTH_QUEUE_SIZE_EXCEEDED`

app/controllers/api/health.php
vermakhushboo: When no domain is passed, the validator throws a 400 error. Added a test.

christyjacob4 (approved):

PR #7551, opened 2024-02-07
Fix email/sms template type enums getting the same values

abnegate:
## What does this PR do?

Fix email/sms template type enums getting the same values

## Test Plan

Specs

## Related PRs and Issues

- (Related PR or issue)

## Checklist


stnguyen90 (commented):
Looking at the specs, I see the `EmailTemplateType` has a value of `magicsession`:

image

but the value in the config is `magicSession`. Is that a problem?

https://github.com/appwrite/appwrite/blob/26b2c27eb0dbdc030bd90e0264d0b5eff9311671/app/config/locale/templates.php#L6

abnegate:
> Looking at the specs, I see the `EmailTemplateType` has a value of `magicsession` but the value in the config is `magicSession`. Is that a problem?

@stnguyen90 Yep this is okay, the `WhiteList` validator is [explicitly forcing lower case values on instantiation](
https://github.com/utopia-php/http/blob/master/src/Http/Validator/WhiteList.php#L44-L48) and [checking lower case during validation](https://github.com/utopia-php/http/blob/f21fd86c049d5fda711e966e9cd926386c36df0d/src/Http/Validator/WhiteList.php#L111) unless otherwise specified

stnguyen90 (approved):

Issue #7109, opened 2023-11-09
🐛 Bug Report: User activity is empty


stnguyen90:
### 👟 Reproduction steps

1. Create a user in a project
2. Create an email session

### 👍 Expected behavior

The activity tab shows session create activity.

### 👎 Actual Behavior

It doesn't show up.

<img width="1502" alt="image" src="https://github.com/appwrite/appwrite/assets/1477010/b97090c3-eced-43bd-b32a-c165be53959c">

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

Schrolli91:
Related: https://github.com/appwrite/appwrite/discussions/7100

moshOntong-IT:
Finally, I thought I was the only one experiencing this issue.

feschaffa:
Having the same issue here, zero activity, using 1.4.11

nick2432:
can i work on this?

stnguyen90:
Looking at the network logs, I can see data is returned:

![image](https://github.com/appwrite/appwrite/assets/1477010/75476ecb-6721-4c67-9576-0a3a69dcaf32)

However:

1. total is 0
2. userId is blank
3. userEmail is blank
4. userName is blank

Compared to collection activity:

![image](https://github.com/appwrite/appwrite/assets/1477010/2c0a0b6d-bfa9-46cc-a456-0e6758cb0652)

The 0 total is probably [preventing anything from being displayed](https://github.com/appwrite/console/blob/ac50ffb6ffb56544cb57ace266ae4bbf0a1ef028/src/lib/layout/activity.svelte#L40).

stnguyen90:
@nick2432, would you like to work on this?

stnguyen90:
@nick2432, actually, I think we'll take care of this ourselves. Assigning to @TorstenDittmann to prioritize.

Souptik2001:
Hi @stnguyen90 @TorstenDittmann,
Saw that this fix has not yet been done, therefore raised a PR for this.

If it has already been fixed please ignore and close the PR. :slightly_smiling_face: 

Thanks!

PR #7559, opened 2024-02-08
Fix missing user activity logs

Souptik2001:
## What does this PR do?

Fixes two bugs -
- For a project's users the activity logs are not populated.
  - This is because the `total` count of the user is not properly calculated and 0 is returned because of that the front-end doesn't render the logs, even if the list of logs is present in the response.
  - Apart from the above issue, a few fields like `userId`, `userEmail`, and `userName`, were not populated in the response data.
- The same thing happens at the user account.
  - Here the logs are returned correctly just the `total` is returned as 0.

## Test Plan

Account (activity logs are now populated because correct `total` value is returned) -
image

Project's user (activity logs are now populated because correct `total` value is returned) -
image

## Related PRs and Issues

#7109 

## Checklist

- [ ] (N/A) If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?


fogelito (approved):
Do not forget if needed to migrate userId data to InternalId

abnegate (approved):

PR #7572, opened 2024-02-12
fix Indexes

fogelito:
Make Attributes & Indexes collection use the InternalId indexes.

## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/messaging.php
```diff
@@ -843,8 +843,12 @@
...
         }
 
-        // Get cursor document if there was a cursor query
-        $cursor = Query::getByType($queries, [Query::TYPE_CURSOR_AFTER, Query::TYPE_CURSOR_BEFORE]);
+        /**
+         * Get cursor document if there was a cursor query, we use array_filter and reset for reference $cursor to $queries
+         */
+        $cursor = \array_filter($queries, function ($query) {
+            return \in_array($query->getMethod(), [Query::TYPE_CURSOR_AFTER, Query::TYPE_CURSOR_BEFORE]);
+        });
```
abnegate, line 851: Let's add a test to make sure that multi-page requests are working as expected. We can use a limit query so we only need to add a few documents

fogelito: Done!

abnegate (approved):

PR #7591, opened 2024-02-15
Add storage health check

vermakhushboo:
## What does this PR do?

This health endpoint for storage will do 3 operations.
```
$device->write(uniquefile);
$device->read(uniquefile);
$device->delete(uniquefile);
```

If one fails, the endpoint fails, this way we test full range of operations. We are testing this for both local as well as remote devices.

## Test Plan
Added test and also tested locally.

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/controllers/api/health.php
+++ b/app/controllers/api/health.php
@@ -702,6 +702,43 @@ App::get('/v1/health/storage/local')
         $response->dynamic(new Document($output), Response::MODEL_HEALTH_STATUS);
     });
 
+App::get('/v1/health/storage')
+    ->desc('Get storage')
+    ->groups(['api', 'health'])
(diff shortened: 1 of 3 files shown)


app/controllers/api/health.php
```diff
@@ -21,6 +21,20 @@
 use Utopia\Validator\Text;
 use Utopia\Validator\WhiteList;
 
+$checkStorageHealth = function (Device $device) {
```
eldadfux, line 24: We try to avoid global functions as much as possible, also this one is called only once so I doubt we need this as a function vs just adding the code in the loop.

christyjacob4 (approved):

PR #7601, opened 2024-02-16
fix: blocked users web controllers

TorstenDittmann:
## What does this PR do?

- move user and project logic from general to api hook
- remove project_unknown error in favor of project_not_found

## Test Plan

- added general tests for hooks
- manually and existing

## Checklist

### DIFF
--- a/app/controllers/general.php
+++ b/app/controllers/general.php
@@ -3,7 +3,6 @@
 require_once __DIR__ . '/../init.php';
 
 use Utopia\App;
-use Utopia\Database\Helpers\Role;
 use Utopia\Locale\Locale;
 use Utopia\Logger\Logger;
 use Utopia\Logger\Log;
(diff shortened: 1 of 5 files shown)


app/controllers/general.php
```diff
@@ -563,25 +563,27 @@ function router(App $utopia, Database $dbForConsole, SwooleRequest $swooleReques
...
         }
 
-        if (false === $user->getAttribute('status')) { // Account is blocked
-            throw new AppwriteException(AppwriteException::USER_BLOCKED);
-        }
-
-        if ($user->getAttribute('reset')) {
-            throw new AppwriteException(AppwriteException::USER_PASSWORD_RESET_REQUIRED);
-        }
+        if (in_array('api', $route->getGroups())) {
```
eldadfux, line 566: Why do this if instead of just adding the block to an init hook we have that already runs only for this group?

PR #7605, opened 2024-02-18
Added cancel build endpoint

vermakhushboo:
## What does this PR do?

Add an endpoint to cancel a build and kill the container. 

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/functions.php
```diff
@@ -1475,6 +1475,55 @@
         $response->noContent();
     });
 
+App::patch('/v1/functions/:functionId/deployments/:deploymentId/builds/:buildId')
```
eldadfux, line 1478: Not sure about the look of the REST API here, also the description doesn't follow our usual conventions. Also, only a delete request should return `no content`, should this be a delete request?

Should we maybe have a CRUD for `builds` and delete the build documents while updating the deployment status to cancelled?

@vermakhushboo @Meldiron @christyjacob4 would love your thoughts on the DX here.

src/Appwrite/Platform/Workers/Builds.php
```diff
@@ -390,6 +390,10 @@ protected function buildDeployment(callable $getFunctionsDevice, Func $queueForF
             $response = null;
             $err = null;
 
+            if ($build->getAttribute('status') === 'cancelled') {
```
eldadfux, line 393: How are we actually letting open runtimes to stop the runtime build and save compute time?

vermakhushboo: This command stops the runtime and kills the container: https://github.com/appwrite/appwrite/blob/6ebca949c0bd86442a826cd272b7ad197b3068a2/app/controllers/api/functions.php#L1516

If the build is cancelled before the container is created, we're preventing the creation of the container here: https://github.com/appwrite/appwrite/blob/6ebca949c0bd86442a826cd272b7ad197b3068a2/src/Appwrite/Platform/Workers/Builds.php#L393

Let me know if you have any questions.

app/controllers/api/functions.php
vermakhushboo: Updated the response model and description. `deployment` doesn't have a status. I don't think we should delete the build document, since the status of the build is displayed on the console. Open for suggestions.

app/controllers/api/functions.php
```diff
@@ -1184,6 +1184,29 @@
...
 
+            $buildId = ID::unique();
+            $build = $dbForProject->createDocument('builds', new Document([
+                '$id' => $buildId,
+                '$permissions' => [],
+                'creationTime' => DateTime::now(),
+                'startTime' => null,
+                'deploymentInternalId' => $deployment->getInternalId(),
+                'deploymentId' => $deployment->getId(),
+                'status' => 'waiting',
```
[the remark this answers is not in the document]
vermakhushboo: `waiting` is a valid status for builds. We already use it at some places.

eldadfux (approved):

app/controllers/api/functions.php
```diff
@@ -1417,7 +1461,6 @@
     ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
     ->label('sdk.namespace', 'functions')
     ->label('sdk.method', 'createBuild')
-    ->label('sdk.description', '/docs/references/functions/create-build.md')
```
Meldiron, line 1420: Was this deleted on purpose?

app/controllers/api/functions.php
```diff
@@ -1475,7 +1538,57 @@
         $response->noContent();
     });
 
-App::post('/v1/functions/:functionId/executions')
+    App::patch('/v1/functions/:functionId/deployments/:deploymentId/builds/:buildId')
+    ->groups(['api', 'functions'])
+    ->desc('Update build status to cancelled')
```
Meldiron, line 1543:
```suggestion
    ->desc('Cancel a deployment')
```

app/controllers/api/functions.php
```diff
@@ -1475,7 +1538,57 @@
...
+    ->label('audits.resource', 'function/{request.functionId}')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'functions')
+    ->label('sdk.method', 'cancelBuild')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_BUILD)
+    ->param('functionId', '', new UID(), 'Function ID.')
+    ->param('deploymentId', '', new UID(), 'Deployment ID.')
+    ->param('buildId', '', new UID(), 'Build unique ID.')
```
Meldiron, line 1555: Lets remove buildId as param, I dont think other function endpoints have it. When you have deployment, you can get build. its one to one relation

docs/references/functions/create-build.md
```diff
@@ -1 +0,0 @@
-Create a new build for an Appwrite Function deployment. This endpoint can be used to retry a failed build.
```
Meldiron, line 1: Something feels off. This file deletion seems weird, and description says retry so maybe it was incorrect already

src/Appwrite/Platform/Workers/Builds.php
```diff
@@ -210,6 +194,12 @@ protected function buildDeployment(callable $getFunctionsDevice, Func $queueForF
                 $stdout = '';
                 $stderr = '';
                 Console::execute('mkdir -p /tmp/builds/' . \escapeshellcmd($buildId), '', $stdout, $stderr);
+
+                $build = $dbForProject->getDocument('builds', $buildId);
+                if ($build->getAttribute('status') === 'cancelled') {
+                    return;
+                }
```
Meldiron, line 201: I see this check above, do we need to duplicate it here?

Meldiron: I see duplication in other places too

src/Appwrite/Platform/Workers/Builds.php
```diff
@@ -456,6 +451,10 @@ protected function buildDeployment(callable $getFunctionsDevice, Func $queueForF
                 ]);
 
             if ($err) {
+                $build = $dbForProject->getDocument('builds', $buildId);
+                if ($build->getAttribute('status') === 'cancelled') {
+                    return;
```
Meldiron, line 456: Here: https://github.com/appwrite/appwrite/pull/7605/files#diff-e06b561b3950351ec356b5f1b2ecc67b6999336f9efb3cb339fae4a67797b4e0L497

I would also ass one more cancelled check - so it doesn't get overriten with "failed"

app/controllers/api/functions.php
vermakhushboo: Yes, Eldad asked to hide it from the docs so that we can make breaking changes later.

src/Appwrite/Platform/Workers/Builds.php
vermakhushboo: The duplication is intentional, since cancellation can happen at any stage and we should save compute by capturing it before time consuming tasks.

app/controllers/api/functions.php
```diff
@@ -1494,6 +1557,59 @@
         $response->noContent();
     });
 
+App::patch('/v1/functions/:functionId/deployments/:deploymentId/builds/:buildId')
```
Meldiron, line 1560: Considering deploymentId and buildId is oneToOne relationship, I don't like you need to specify both on endpoint. We could get buildId using subqueries.

Let's discuss with Christy or Eldad how this endpoint could look like, and how we can rename existing one (redeploy endpoint)

app/controllers/api/functions.php
```diff
@@ -117,6 +117,28 @@
...
     ]));
 
+    $buildId = ID::unique();
+    $build = $dbForProject->createDocument('builds', new Document([
+        '$id' => $buildId,
+        '$permissions' => [],
+        'startTime' => null,
+        'deploymentInternalId' => $deployment->getInternalId(),
+        'deploymentId' => $deployment->getId(),
+        'status' => 'waiting',
```
christyjacob4, line 127: we have to stay consistent with the available build types her https://github.com/appwrite/appwrite/blob/feat-cancel-build-deployment/src/Appwrite/Utopia/Response/Model/Build.php#L26-L29

Waiting doesn't seem to one we use..

app/console
christyjacob4, line 1: we need to remove this file from the PR

src/Appwrite/Platform/Workers/Builds.php
```diff
@@ -221,6 +209,12 @@ protected function buildDeployment(Device $deviceForFunctions, Func $queueForFun
                 $stdout = '';
                 $stderr = '';
                 Console::execute('mkdir -p /tmp/builds/' . \escapeshellcmd($buildId), '', $stdout, $stderr);
+
+                $build = $dbForProject->getDocument('builds', $buildId);
+                if ($build->getAttribute('status') === 'cancelled') {
+                    return;
```
christyjacob4, line 215: Is the return sufficient? Can we also log that the build has been cancelled and then return?

christyjacob4 (changes requested):
Some minor comments

app/controllers/api/functions.php
vermakhushboo: Updating the endpoint to `deployments/:deploymentId/build` as discussed

app/controllers/api/functions.php
vermakhushboo: Created a separate task to use enums for states

src/Appwrite/Platform/Workers/Builds.php
vermakhushboo: Added console::info

app/controllers/api/functions.php
```diff
@@ -1499,6 +1562,59 @@
         $response->noContent();
     });
 
+App::patch('/v1/functions/:functionId/deployments/:deploymentId/build')
+    ->groups(['api', 'functions'])
+    ->desc('Update build status')
```
Meldiron, line 1567:
```suggestion
    ->desc('Cancel deployment')
```

app/controllers/api/functions.php
```diff
@@ -1499,6 +1562,59 @@
...
+    ->label('scope', 'functions.write')
+    ->label('audits.event', 'deployment.update')
+    ->label('audits.resource', 'function/{request.functionId}')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'functions')
+    ->label('sdk.method', 'cancelBuild')
```
Meldiron, line 1573:
```suggestion
    ->label('sdk.method', 'updateDeploymentBuild')
```

app/controllers/api/functions.php
```diff
@@ -1499,6 +1562,59 @@
...
+
+        if ($deployment->isEmpty()) {
+            throw new Exception(Exception::DEPLOYMENT_NOT_FOUND);
+        }
+
+        $build = Authorization::skip(fn () => $dbForProject->getDocument('builds', $deployment->getAttribute('buildId', '')));
+
+        if ($build->isEmpty()) {
+            throw new Exception(Exception::BUILD_NOT_FOUND);
+        }
```
Meldiron, line 1600: I dont think we should throw here, after our most recent discussion - approach when we create build here

app/controllers/api/functions.php
```diff
@@ -1499,6 +1562,59 @@
...
+
+        if (\in_array($build->getAttribute('status'), ['ready', 'failed'])) {
+            throw new Exception(Exception::GENERAL_BAD_REQUEST, 'Build is already completed and cannot be cancelled.');
```
Meldiron, line 1603: I think this is worth a new exception type

vermakhushboo:
TODO: Check how to compute duration for cancelled build

Meldiron (approved):

app/controllers/api/functions.php
```diff
@@ -1493,6 +1491,87 @@
...
+                'startTime' => DateTime::now(),
+                'deploymentInternalId' => $deployment->getInternalId(),
+                'deploymentId' => $deployment->getId(),
+                'status' => 'canceled',
+                'path' => '',
+                'runtime' => $function->getAttribute('runtime'),
+                'source' => $deployment->getAttribute('path', ''),
+                'sourceType' => '',
+                'logs' => '',
+                'endTime' => DateTime::now(),
```
christyjacob4, line 1541: shouldnt endtime be empty when creating the build ?

loks0n: fixed

PR #7607, opened 2024-02-19
Allow execution filter attributes

vermakhushboo:
## What does this PR do?

Allow executions to be filtered by attributes like status, trigger, path, duration etc.

## Test Plan

Tested using client

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/config/collections.php
```diff
@@ -2934,6 +2934,27 @@
...
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['requestMethod'],
+                'lengths' => [128],
+                'orders' => [Database::ORDER_ASC],
+            ],
+            [
+                '$id' => ID::custom('_key_requestPath'),
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['requestPath'],
+                'lengths' => [Database::LENGTH_KEY],
```
christyjacob4, line 2948: What is the max length of the request path? Will it be under 255 characters ?

vermakhushboo: Good point. We use 2048 in the API, so changing it to that.

christyjacob4 (approved):
Looks good. I believe we can target this to the 1.6 branch

PR #7611, opened 2024-02-19
Feat email attachments

abnegate:
## What does this PR do?

Allow attaching files from storage to emails via a list of compound `bucketId:fileId`

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/worker.php
```diff
@@ -209,6 +210,11 @@
         return getDevice(APP_STORAGE_CACHE . '/app-' . $projectId);
     };
 });
+Server::setResource('getLocalCache', function () {
```
eldadfux, line 213: Why do we return a callback? best practice for a resource to return an instance. I'm pretty sure we should already have a common way for accessing storage devices.

src/Appwrite/Utopia/Database/Validator/CompoundUID.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+namespace Appwrite\Utopia\Database\Validator;
```
eldadfux, line 3: We're missing a unit test for this one.

app/worker.php
eldadfux: Also, why do we need the local cache? don't we use it just for image previews?

abnegate: We do this for the worker resources because we can't inject the project, so we need to return functions that take the `projectId` as a parameter.

We need the local cache because if the files are using a remote storage service, we need to download them locally to be able to pass the path for the attachment

eldadfux: What we would usually do with large files is use the container storage or memory to download them in chunks and not store it on one of our volumes. I would just create it locally for now and avoid the resource. We do need to fine better dependency injection for `project` in the worker content. I'm almost sure someone already worked on that. Let's check internally.

eldadfux (approved):

PR #7621, opened 2024-02-20
Fix 1.5.x Migrations

PineappleIOnic:
## What does this PR do?

This PR makes sure we don't migrate things that the `console` DB doesn't contain

## Test Plan

Create an Appwrite install of 1.4.x, switch branch to this one and run the migration script, then finally check everything looks good including usage

## Checklist


src/Appwrite/Migration/Version/V20.php
```diff
@@ -47,14 +47,17 @@ public function execute(): void
         Console::info('Migrating Collections');
         $this->migrateCollections();
 
-        Console::info('Migrating Functions');
-        $this->migrateFunctions();
+        // No need to migrate stats for console
+        if ($this->project->getInternalId() !== 'console') {
```
shimonewman, line 51: We can also move these lines to the console if condition

image

src/Appwrite/Utopia/Request/Filters/V17.php
```diff
@@ -59,6 +59,10 @@ public function parse(array $content, string $model): array
```
shimonewman, line 59: I think some of usage API request/response models were updated

abnegate (approved):

PR #7649, opened 2024-02-24
chore: remove array sdk method

loks0n:
- removes array type 'sdk.method' label
- removes shared method for updatePhoneSession and updateMagicURLSession in favour of multiple endpoints
- adds 'sdk.deprecated' label for future use in SDK generator

- replaces sdk.hideClient and sdk.hideServer with more flexible sdk.hide
- hides createOAuth2Session on server sdk


src/Appwrite/Specification/Format/OpenAPI3.php
```diff
@@ -120,18 +120,10 @@ public function parse(): array
...
-            if ($hide) {
-                continue;
-            }
-
-            $method = $route->getLabel('sdk.method', [\uniqid()]);
-            if (\is_array($method)) {
-                $method = $method[0];
-            }
 
+            $id = $route->getLabel('sdk.method', \uniqid());
```
abnegate, line 126: Let's revert this back to `$method` and keep using it throughout, same for `Swagger2`

loks0n: Done

abnegate (approved):

PR #7662, opened 2024-02-25
Fix: Empty password validation

Meldiron:
## What does this PR do?

When personal data check was enabled, users with empty password didn't create using Users API. They should. I fixed it.

## Test Plan

Implemented new test

## Related PRs and Issues

x

## Checklist


tests/e2e/Services/Account/AccountCustomClientTest.php
```diff
@@ -2251,8 +2251,11 @@ public function testPhoneVerification(array $data): array
 
         $smsRequest = $this->getLastRequest();
 
+        $message = $smsRequest['data']['message'];
+        $token = substr($message, 0, 6);
+
         return \array_merge($data, [
-            'token' => $smsRequest['data']['secret']
+            'token' => $token
```
abnegate, line 2258: Not sure if this will solve the problem entirely, seems to be a race condition between the webhook firing and the message sending. No action just noting

app/controllers/api/users.php
```diff
@@ -71,7 +71,7 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
             : ID::custom($userId);
 
         if ($project->getAttribute('auths', [])['personalDataCheck'] ?? false) {
-            $personalDataValidator = new PersonalData($userId, $email, $name, $phone);
+            $personalDataValidator = new PersonalData($userId, $email, $name, $phone, false, true);
```
abnegate, line 74: Let's use named parameters so it's clear what false and true are doing without checking the class

tests/e2e/Services/Account/AccountCustomClientTest.php
Meldiron: I did var dump and there was no secret, but there was message `XXXXXX is your code for bla bla bla`

Meldiron: Unless this is larger problem, it was just a leftover after Khushboo's change to improve copy

abnegate (approved):

PR #7668, opened 2024-02-26
Rescheduling fixes

abnegate:
## What does this PR do?

Scheduling fixes:
- Throw on update is no schedule exists
- Throw on update if no recipients and status is not draft
- Disable schedules when updating message to draft or processing
- Fix `scheduledAt` not updated on update routes

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/messaging.php
```diff
@@ -3251,19 +3251,64 @@
...
+                throw new Exception(Exception::SCHEDULE_NOT_FOUND);
+            }
+
+            if (!\is_null($scheduledAt)) {
+                $schedule
+                    ->setAttribute('resourceUpdatedAt', DateTime::now())
+                    ->setAttribute('schedule', $scheduledAt)
+                    ->setAttribute('active', $status === MessageStatus::SCHEDULED);
+            } else {
+                $schedule->setAttribute('active', $status === MessageStatus::SCHEDULED);
```
eldadfux, line 3306: Should this be the same as the above?

app/controllers/api/messaging.php
```diff
@@ -3613,38 +3679,6 @@
...
-                    ->setAttribute('resourceUpdatedAt', DateTime::now())
-                    ->setAttribute('schedule', $scheduledAt)
-                    ->setAttribute('active', $status === MessageStatus::SCHEDULED);
-
-                $dbForConsole->updateDocument('schedules', $schedule->getId(), $schedule);
-            }
-
-            $message->setAttribute('scheduleId', $schedule->getId());
-        }
-
```
eldadfux, line 3647: Can we add a test for canceling the schedule?

app/controllers/api/messaging.php
abnegate: No, if you didn't changed the scheduled at time, it should just be activated/deactivated

app/controllers/api/messaging.php
abnegate: Will do :+1:

app/controllers/api/messaging.php
```diff
@@ -3251,19 +3251,68 @@
...
                 throw new Exception(Exception::MESSAGE_ALREADY_FAILED);
+            case MessageStatus::SCHEDULED:
+                if (\is_null($scheduledAt) && \is_null($currentScheduledAt)) {
+                    throw new Exception(Exception::MESSAGE_MISSING_SCHEDULE);
+                }
+                break;
         }
 
-        if (!\is_null($message->getAttribute('scheduledAt')) && $message->getAttribute('scheduledAt') < new \DateTime()) {
+        if (!\is_null($currentScheduledAt) && $currentScheduledAt < new \DateTime()) {
```
stnguyen90, line 3274: I don't think comparing a string with a DateTime() works. 

image

app/controllers/api/messaging.php
```diff
@@ -3251,19 +3251,68 @@
...
                 throw new Exception(Exception::MESSAGE_ALREADY_PROCESSING);
             case MessageStatus::SENT:
                 throw new Exception(Exception::MESSAGE_ALREADY_SENT);
             case MessageStatus::FAILED:
                 throw new Exception(Exception::MESSAGE_ALREADY_FAILED);
+            case MessageStatus::SCHEDULED:
+                if (\is_null($scheduledAt) && \is_null($currentScheduledAt)) {
+                    throw new Exception(Exception::MESSAGE_MISSING_SCHEDULE);
+                }
+                break;
```
stnguyen90, line 3271: What if the message was draft and then I update it to be scheduled, but I don't pass `$scheduledAt`?

stnguyen90 (changes requested):

app/controllers/api/messaging.php
eldadfux: Should schedule even be a status??? I really doubt this.

abnegate: Resolved by https://github.com/appwrite/appwrite/pull/7670

eldadfux (approved):

stnguyen90:
> It was bad DX to allow explicitly passing the status, and we only needed draft as a bool and the existing sheduledAt params to determine the status, so we decided to remove it.

PR #7696, opened 2024-02-29
feat: mfa collection restructure

TorstenDittmann:
## What does this PR do?

- creates a new collection for authenticators
- allow deleting session

## Checklist


app/controllers/api/account.php
```diff
@@ -3942,26 +3964,30 @@
     ->action(function (string $challengeId, string $otp, Document $project, Response $response, Document $user, Database $dbForProject, Event $queueForEvents) {
 
         $challenge = $dbForProject->getDocument('challenges', $challengeId);
+        var_dump($challenge);
```
stnguyen90, line 3967:
```suggestion
```

app/controllers/api/account.php
```diff
@@ -3641,32 +3660,35 @@
...
     ->inject('dbForProject')
     ->inject('queueForEvents')
-    ->action(function (string $type, string $otp, ?\DateTime $requestTimestamp, Response $response, Document $user, Document $project, Database $dbForProject, Event $queueForEvents) {
+    ->action(function (string $type, string $otp, Response $response, Document $user, Document $project, Database $dbForProject, Event $queueForEvents) {
 
-        $success = match ($type) {
-            'totp' => Challenge\TOTP::verify($user, $otp),
-            default => false
+        $authenticator = match ($type) {
+            'totp' => TOTP::getAuthenticatorFromUser($user),
```
stnguyen90, line 3671: Can we use a const/enum instead of `'totp'` magic string?

stnguyen90 (changes requested):
Does app/console need to be in this PR?

app/controllers/api/account.php
```diff
@@ -3947,21 +3968,24 @@
...
             default => false
         };
 
-    if (!$success && $provider === 'totp') {
-        $backups = $user->getAttribute('totpBackup', []);
-        if (in_array($otp, $backups)) {
+    if (!$success && $type === Type::TOTP) {
+        $authenticator = TOTP::getAuthenticatorFromUser($user);
+        $data = $authenticator->getAttribute('data', []);
+        if (in_array($otp, $data['backups'])) {
```
stnguyen90, line 3983: I know it's unlikely to happen, but we should probably do some sort of null check here just in case. When static analyzer is enabled, it'll complain if we don't.

```suggestion
        $data = $authenticator?->getAttribute('data', []);
        if (in_array($otp, $data['backups'] ?? [])) {
```

src/Appwrite/Auth/MFA/Challenge/Email.php
```diff
@@ -15,8 +15,8 @@ public static function verify(Document $challenge, string $otp): bool
     public static function challenge(Document $challenge, Document $user, string $otp): bool
     {
         if (
-            $challenge->isSet('provider') &&
-            $challenge->getAttribute('provider') === 'email'
+            $challenge->isSet('type') &&
+            $challenge->getAttribute('type') === 'email'
```
stnguyen90, line 19:
```suggestion
            $challenge->getAttribute('type') === TYPE::email
```

src/Appwrite/Auth/MFA/Challenge/Phone.php
```diff
@@ -15,8 +15,8 @@ public static function verify(Document $challenge, string $otp): bool
     public static function challenge(Document $challenge, Document $user, string $otp): bool
     {
         if (
-            $challenge->isSet('provider') &&
-            $challenge->getAttribute('provider') === 'phone'
+            $challenge->isSet('type') &&
+            $challenge->getAttribute('type') === 'phone'
```
stnguyen90, line 19:
```suggestion
            $challenge->getAttribute('type') === Type::PHONE
```

stnguyen90 (changes requested):

app/controllers/api/account.php
```diff
@@ -2630,7 +2630,7 @@
 
 App::delete('/v1/account/sessions/:sessionId')
     ->desc('Delete session')
-    ->groups(['api', 'account'])
+    ->groups(['api', 'account', 'mfa'])
```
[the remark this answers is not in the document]
TorstenDittmann: there is no other way for the user to escape the session that requires mfa

stnguyen90 (approved):

PR #7709, opened 2024-03-01
Fix: MFA flows and docs

Meldiron:
## What does this PR do?

Addresses last minute flow fixes before 1.5 release

## Test Plan

- [x] Manual QA

## Related PRs and Issues

x

## Checklist

- [y] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?


Meldiron:
**QA WENT WELL ✅**

Tasks I focused on during QA:

## Endpoint to create recovery codes

Endpoint works, and created recovery codes for me 

```
{"recoveryCodes":["178452622d","a999eeb54d","2dee5b0805","52db3defe0","12394ff246","98d1a66ec8"]}%                                                             
```

I could only use it once, then I got exception.

## Recovery code to be 10 hexadecimal chars

As seen above, code is for example `52db3defe0` which follows new pattern

## Rename 2FA methods in SDK

Console is broken and I need to use new endpoints during QA for it to work.

Method names are 100% to work, as we already depend on that logic all over Appwrite

## All 2fa endpoints must accept recovery codes

Only one missing was delete authenticator. I tried to delete authenticator using recovery code and it worked.

## Separate recovery codes to custom type

After creating recovery codes, I tried them to sign in, and it worked. I had to use new type. using old TOTP type didn't accept recovery codes anymore. I only could use it once - using it second time didn't work.

app/controllers/api/account.php
```diff
@@ -3575,8 +3575,8 @@
     ->label('audits.userId', '{response.$id}')
     ->label('sdk.auth', [APP_AUTH_TYPE_SESSION, APP_AUTH_TYPE_JWT])
     ->label('sdk.namespace', 'account')
-    ->label('sdk.method', 'addAuthenticator')
-    ->label('sdk.description', '/docs/references/account/add-authenticator.md')
+    ->label('sdk.method', 'createMfaAuthenticator')
```
eldadfux, line 3578:
```suggestion
    ->label('sdk.method', 'createMfaAuthenticators')
```

app/controllers/api/account.php
```diff
@@ -3593,19 +3593,17 @@
...
         $otp->setIssuer($project->getAttribute('name'));
 
-        $backups = Type::generateBackupCodes();
-
         $authenticator = TOTP::getAuthenticatorFromUser($user);
 
         if ($authenticator) {
             if ($authenticator->getAttribute('verified')) {
-                throw new Exception(Exception::GENERAL_UNKNOWN, 'Authenticator already exists on this account.');
+                throw new Exception(Exception::GENERAL_UNKNOWN, 'Authenticator already exists on this account.'); // TODO: Type this exception
```
eldadfux, line 3606: Reminder

app/controllers/api/account.php
Meldiron: Are you sure? In documents we have `createDocument` even tho endpoint is `/documents`

app/controllers/api/account.php
Meldiron: Thanks, missed that one.

app/controllers/api/account.php
Meldiron: From discord chat, Eldad meant list documents. Note for myself to check all list documents for MFA if they arent singular.

app/controllers/api/account.php
Meldiron: All good, double-checked all list endpoints in user& account contorller

app/controllers/api/account.php
Meldiron: All good, double-checked all list endpoints in user& account contorller

app/controllers/api/account.php
Meldiron: Fixed

eldadfux (approved):

PR #7725, opened 2024-03-04
Allow users to disable APIs

vermakhushboo:
## What does this PR do?

Allow users to disable APIs 

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/projects.php
```diff
@@ -473,6 +473,71 @@
...
+
+        $apis = $project->getAttribute('apis', []);
+        $apis[$api] = $status;
+
+        $project = $dbForConsole->updateDocument('projects', $project->getId(), $project->setAttribute('apis', $apis));
+
+        $response->dynamic($project, Response::MODEL_PROJECT);
+    });
+
+App::patch('/v1/projects/:projectId/api/all')
```
eldadfux, line 507: Not sure I love this all endpoint, I see it was already there for services. Do we use it in the console? Can we remove both? Can also be addressed later.

app/controllers/general.php
```diff
@@ -92,6 +92,13 @@ function router(App $utopia, Database $dbForConsole, callable $getProjectDB, Swo
         }
     }
 
+    if (array_key_exists('proxy', $project->getAttribute('apis', []))) {
```
eldadfux, line 95: What is proxy? is this for custom domains/functions?

app/controllers/api/projects.php
vermakhushboo: Yes, we use it in the console. As discussed, we can address it later.

app/controllers/general.php
vermakhushboo: Probably used for functions. Just copied it from the existing services code.

PR #7744, opened 2024-03-07
Add event information into specs

gewenyu99:
## What does this PR do?

Generate event information into Appwrite specs in two places
- Each endpoint will have the events that it can trigger.
- Each spec will have a giant list of all possible events
- Removes unused offline support headers as a side quest.

This can be used to make documenting event easier when we have new releases.

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/console
stnguyen90, line 1: Please remove this change from the PR

src/Appwrite/Specification/Format.php
```diff
@@ -21,43 +21,56 @@ abstract class Format
...
+        "endpoint" => "https://localhost",
+        "version" => "1.0.0",
+        "terms" => "",
+        "support.email" => "",
+        "support.url" => "",
+        "contact.name" => "",
+        "contact.email" => "",
+        "contact.url" => "",
+        "license.name" => "",
+        "license.url" => "",
```
stnguyen90, line 44: We typically use single quote for strings.

stnguyen90 (changes requested):

stnguyen90 (commented):
Fyi, 3 things:

1. We may be removing events from endpoints in the future because may use the database listener to trigger events.
2. The way we have events right now limits us to 1 event per endpoint, but there are times where we need multiple events on a single endpoint.
3. We need to make sure we're not coupling our SDK generator with Appwrite

src/Appwrite/Specification/Format.php
gewenyu99: @stnguyen90 I ran the formatter lol I was hoping this would work :P

gewenyu99: Fixed

src/Appwrite/Platform/Tasks/Specs.php
```diff
@@ -299,4 +308,53 @@ public function action(string $version, string $mode, Registry $register): void
...
+            // if it's not an array of nodes, but something description or model, skip
+            if (!is_array($subNode)) {
+                continue;
+            }
+
+            // add wild card if applicable
+            if ($subNode['$resource'] ?? false) {
+                $pattern = "{$key}.*";
+            } else {
+                $pattern = "{$key}";
```
gewenyu99, line 327:
```suggestion
                $pattern = '{$key}';
```

src/Appwrite/Platform/Tasks/Specs.php
```diff
@@ -299,4 +308,53 @@ public function action(string $version, string $mode, Registry $register): void
...
+        foreach ($events as $key => $subNode) {
+            $pattern = '';
+            // if it's not an array of nodes, but something description or model, skip
+            if (!is_array($subNode)) {
+                continue;
+            }
+
+            // add wild card if applicable
+            if ($subNode['$resource'] ?? false) {
+                $pattern = "{$key}.*";
```
gewenyu99, line 325:
```suggestion
                $pattern = '{$key}.*';
```

src/Appwrite/Specification/Format.php
```diff
@@ -375,23 +323,27 @@ public function getEnumKeys(string $service, string $method, string $param): arr
                     case 'getCollectionUsage':
                     case 'getDatabaseUsage':
                         // Range Enum Keys
-                        return ['Twenty Four Hours', 'Thirty Days', 'Ninety Days'];
+                        $values  = ['Twenty Four Hours', 'Seven Days', 'Thirty Days', 'Ninety Days'];
+                        return $values;
                 }
                 break;
-            case 'functions':
+            case 'function':
```
loks0n, line 330: Is `function` correct? The rest are plural

package-lock.json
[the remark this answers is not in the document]
gewenyu99: Ooops

src/Appwrite/Platform/Tasks/Specs.php
```diff
@@ -299,4 +308,53 @@ public function action(string $version, string $mode, Registry $register): void
             }
         }
     }
+
+    private function generateEvents(array $events, $parentModel = null)
```
loks0n, line 312: For this logic, is it possible to reuse the existing code in `Events::generateEvents()`?

gewenyu99: Not sure, I don't know if we have a base patter we can pass into there. I also assumed no because in the old docs website we didn't reuse this logic :/

app/console
stnguyen90: Looks like it's still in here. You can try to remove it by running:

```
cd app/console
git checkout d75ef00fb088c909bf8fdc5b12c2fe25ed270b43
cd ../..
git commit
```

src/Appwrite/Platform/Tasks/Specs.php
stnguyen90: What about re-using logic in here: https://github.com/appwrite/appwrite/blob/main/src/Appwrite/Event/Validator/Event.php#L34

stnguyen90 (changes requested):

src/Appwrite/Platform/Tasks/Specs.php
gewenyu99: Oh this is worth taking a look.

PR #7777, opened 2024-03-10
Fix tests required for Cloud

lohanidamodar:
## What does this PR do?

- Fix tests

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/init.php
```diff
@@ -1161,7 +1161,7 @@ function (mixed $value) {
         ->setMetadata('host', \gethostname())
         ->setMetadata('project', 'console')
         ->setTimeout(APP_DATABASE_TIMEOUT_MILLISECONDS);
-
```
lohanidamodar, line 1164: Without this initial load was always getting to pool empty

christyjacob4: I looked at the PR. Im okay with all the change except this one. There can be unexpected side effects. 
You can also tweak the values of these variables only for the test in cloud 
https://github.com/appwrite/appwrite/blob/main/app/init.php#L790-L805

lohanidamodar: reverted this.

christyjacob4 (approved):

PR #7780, opened 2024-03-10
Fix git deployments

stnguyen90:
## What does this PR do?

* fix deployments stuck at processing 
  * Ensure project is passed in the event because the init hook
isn't able to detect the project.
  * Make sure the build is triggered for each function.
  * Reset the build event so the shutdown hook doesn't trigger
again.
* prevent an error with one function deployment stopping others
* reduce nested code

## Test Plan

Manually deployed via GitHub:

image

## Related PRs and Issues

None

## Checklist

### DIFF
--- a/app/controllers/api/vcs.php
+++ b/app/controllers/api/vcs.php
@@ -238,11 +238,16 @@ $createGitDeployments = function (GitHub $github, string $providerInstallationId
             $queueForBuilds
                 ->setType(BUILD_TYPE_DEPLOYMENT)
                 ->setResource($function)
-                ->setDeployment($deployment);
+                ->setDeployment($deployment)
+                ->setProject($project); // set the project because it won't be set for git deployments
+
+            $queueForBuilds->trigger(); // must trigger here so that we create a build for each function
 
             //TODO: Add event?
         }
     }
+
+    $queueForBuilds->setType(''); // prevent shutdown hook from triggering again
 };
 
 App::get('/v1/vcs/github/authorize')
--- a/app/controllers/api/vcs.php
+++ b/app/controllers/api/vcs.php
@@ -254,7 +254,7 @@ $createGitDeployments = function (GitHub $github, string $providerInstallationId
         }
     }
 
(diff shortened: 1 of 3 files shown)


app/controllers/api/vcs.php
```diff
@@ -238,11 +243,22 @@
...
+
+            $queueForBuilds->trigger(); // must trigger here so that we create a build for each function
 
             //TODO: Add event?
+        } catch (Throwable $e) {
+            $errors[] = $e->getMessage();
         }
     }
+
+    $queueForBuilds->setType(''); // prevent shutdown hook from triggering again
```
Meldiron, line 257: Seems like patch. Can we introduce methods `disable` and `enable` for this purpose?

app/controllers/api/vcs.php
```diff
@@ -238,11 +243,22 @@
...
+
+    if (!empty($errors)) {
+        throw new Exception(Exception::GENERAL_UNKNOWN, \implode("\n", $errors));
```
Meldiron, line 260: I think we can throw server error here, as its not public facing anyway.

Meldiron: Doing that also ensures we get it on Sentry

stnguyen90: This should end up in Sentry since `GENERAL_UNKNOWN` is a 500 error right?

app/controllers/api/vcs.php
stnguyen90: We trigger when type is set:

https://github.com/appwrite/appwrite/blob/33b4ee5c0f41b5919ae1dac1fb86572667062f29/app/controllers/shared/api.php#L639-L653

So, I could add a disable method that sets the type to empty string. Would that suffice?

stnguyen90: To minimize changes, I updated reset() to properly reset all private variables (like type) and am now using that instead of setType('').

Meldiron (approved):

vermakhushboo (approved):

PR #7787, opened 2024-03-11
Fix database exception wrapping

abnegate:
## What does this PR do?

Move database exception wrapping into error handler instead of per-route

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


stnguyen90:
Is this just to consolidate code so that we handle exceptions in one place?

stnguyen90 (commented):
Is this just to consolidate code so that we handle exceptions in one place?

abnegate:
> Is this just to consolidate code so that we handle exceptions in one place?

Yes

eldadfux (approved):

PR #7790, opened 2024-03-11
Bug 5901 fix resend invitation from create invite

ZachHandley:
I merged heads to resolve the merge conflicts and check something in my code to make sure I didn't see a bug, and that closed my pull request, so anyways this one says it can merge automatically @stnguyen90 sorry about the closed one

## What does this PR do?

Adds a new call into the `createMembership` endpoint that checks the database for an unconfirmed membership with the same userEmail. If they don't exist, it creates a membership like normal, which will throw a duplicate error like normal if the users email exists.

If they do exist, it will set the `$membership` to the found membership, and set the attribute `invited` to `DateTime::now()`, then updates the user and it continues onward like normal then sends the email, many success!

## Test Plan

The only possible problem this could have was in the `$dbForProject->updateDocument` (with the $collectionId or something being included like in the client SDK's was my thought) but I have looked at the other code in the `teams.php` file and they seem to be doing things the same way, so it shouldn't conflict at all.

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/5901

## Checklist


app/controllers/api/teams.php
```diff
@@ -1022,29 +1036,26 @@
 
         $dbForProject->purgeCachedDocument('users', $user->getId());
 
-        Authorization::skip(fn () => $dbForProject->increaseDocumentAttribute('teams', $team->getId(), 'total', 1));
+        $team = Authorization::skip(fn () => $dbForProject->updateDocument('teams', $team->getId(), $team->setAttribute('total', $team->getAttribute('total', 0) + 1)));
```
stnguyen90, line 1039: Why can't we use the increaseDocumentAttribute() method?

app/controllers/api/teams.php
```diff
@@ -1102,14 +1113,14 @@
         $dbForProject->purgeCachedDocument('users', $user->getId());
 
         if ($membership->getAttribute('confirm')) { // Count only confirmed members
-            Authorization::skip(fn () => $dbForProject->decreaseDocumentAttribute('teams', $team->getId(), 'total', 1, 0));
```
stnguyen90, line 1105: Why can't we use the `decreaseDocumentAttribute()` method?

stnguyen90 (changes requested):

app/controllers/api/teams.php
ZachHandley: We totally can, I think I defaulted to habit haha

ZachHandley:
@stnguyen90 swapped it, sorry for the delay

app/controllers/api/teams.php
stnguyen90: This one hasn't been updated.

stnguyen90 (changes requested):

app/controllers/api/teams.php
```diff
@@ -1103,14 +1114,13 @@
...
+            Authorization::skip(fn () => $dbForProject->increaseDocumentAttribute('teams', $team->getId(), 'total', 1));
```
stnguyen90, line 1117: shouldn't this be decrease? Best to restore it to what it was before.

stnguyen90 (changes requested):
A lot of tests are failing. Maybe your branch is behind?

app/controllers/api/teams.php
ZachHandley: @stnguyen90 I'm a bit confused about that, it says if `$membership->getAttribute('confirm')`, doesn't that mean we _should_ count that member?

ZachHandley: I'll commit it to reverse, but I was confused by that

stnguyen90: Because it was a confirmed member, the total was incremented previously. So, now in the delete membership, the total should be decremented.

stnguyen90 (changes requested):
Looks like the change has broken some tests.

Issue #7823, opened 2024-03-13
🐛 Bug Report: The Msg91 OTP isn't working after the 1.5 upgrade


rohankm:
### 👟 Reproduction steps

Hi,
I have set it up according to the docs https://appwrite.io/docs/advanced/self-hosting/sms
the OTP for login was working before the upgrade to 1.5

here are the appwrite-worker-messaging logs

```
Warning: Undefined array key "templateId" in /usr/src/code/src/Appwrite/Platform/Workers/Messaging.php on line 461
[Job] (65f19b7b27cfc0.65895609) failed to run.
[Job] (65f19b7b27cfc0.65895609) Utopia\Messaging\Adapter\SMS\Msg91::__construct(): Argument #3 ($templateId) must be of type string, null given, called in /usr/src/code/src/Appwrite/Platform/Workers/Messaging.php on line 461
[Error] Type: TypeError
[Error] Message: Utopia\Messaging\Adapter\SMS\Msg91::__construct(): Argument #3 ($templateId) must be of type string, null given, called in /usr/src/code/src/Appwrite/Platform/Workers/Messaging.php on line 461
[Error] File: /usr/src/code/vendor/utopia-php/messaging/src/Utopia/Messaging/Adapter/SMS/Msg91.php
[Error] Line: 21
```

### 👍 Expected behavior

it should send the otp

### 👎 Actual Behavior

```
Warning: Undefined array key "templateId" in /usr/src/code/src/Appwrite/Platform/Workers/Messaging.php on line 461
[Job] (65f19b7b27cfc0.65895609) failed to run.
[Job] (65f19b7b27cfc0.65895609) Utopia\Messaging\Adapter\SMS\Msg91::__construct(): Argument #3 ($templateId) must be of type string, null given, called in /usr/src/code/src/Appwrite/Platform/Workers/Messaging.php on line 461
[Error] Type: TypeError
[Error] Message: Utopia\Messaging\Adapter\SMS\Msg91::__construct(): Argument #3 ($templateId) must be of type string, null given, called in /usr/src/code/src/Appwrite/Platform/Workers/Messaging.php on line 461
[Error] File: /usr/src/code/vendor/utopia-php/messaging/src/Utopia/Messaging/Adapter/SMS/Msg91.php
[Error] Line: 21
```

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

[...]

Issue #7785, opened 2024-03-11
🐛 Bug Report: Undefined array key "templateld" msg91


rvndsngwn:
### 👟 Reproduction steps

Appwrite throws the following error when I try to use the msg91 service for messaging:-
```
   
appwrite-worker-messaging     | Warning: Undefined array key "templateId" in /usr/src/code/src/Appwrite/Platform/Workers/Messaging.php on line 461
appwrite-worker-messaging     | 
appwrite-worker-messaging     | Fatal error: Uncaught TypeError: Utopia\Messaging\Adapter\SMS\Msg91::__construct(): Argument #3 ($templateId) must be of type string, null given, called in /usr/src/code/src/Appwrite/Platform/Workers/Messaging.php on line 461 and defined in /usr/src/code/vendor/utopia-php/messaging/src/Utopia/Messaging/Adapter/SMS/Msg91.php:21
appwrite-worker-messaging     | Stack trace:
appwrite-worker-messaging     | #0 /usr/src/code/src/Appwrite/Platform/Workers/Messaging.php(461): Utopia\Messaging\Adapter\SMS\Msg91->__construct('MOHESU', '390713A5p9NDmG7...', NULL)
appwrite-worker-messaging     | #1 /usr/src/code/src/Appwrite/Platform/Workers/Messaging.php(236): Appwrite\Platform\Workers\Messaging->getSmsAdapter(Object(Utopia\Database\Document))
appwrite-worker-messaging     | #2 @swoole/library/core/Coroutine/functions.php(43): Appwrite\Platform\Workers\Messaging->Appwrite\Platform\Workers\{closure}()
appwrite-worker-messaging     | #3 [internal function]: Swoole\Coroutine\{closure}()
appwrite-worker-messaging     | #4 {main}
appwrite-worker-messaging     |   thrown in /usr/src/code/vendor/utopia-php/messaging/src/Utopia/Messaging/Adapter/SMS/Msg91.php on line 21
```

### 👍 Expected behavior

Appwrite should handle the Msg91 service for messaging without encountering any errors and should not produce warnings or fatal errors.

### 👎 Actual Behavior

The error mentioned above is thrown, indicating an issue with the handling of the Msg91 service in the messaging component of Appwrite.

![Screenshot 2024-03-11 at 3 29 00 PM](https://github.com/appwrite/appwrite/assets/59143934/40e3a0bd-5d23-4893-8551-ade5c4780c4c)

### 🎲 Appwrite version

[...]

PR #7824, opened 2024-03-13
Fix msg91 params

abnegate:
## What does this PR do?

Add `templateId` parameter and remove `from` parameter

Fixes: https://github.com/appwrite/appwrite/issues/7823

## Test Plan

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/7823
- https://github.com/appwrite/appwrite/issues/7785

## Checklist


src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -409,7 +409,8 @@ private function sendInternalSMSMessage(Document $message, Document $project, ar
                 ],
                 'msg91' => [
                     'senderId' => $user,
-                    'authKey' => $password
+                    'authKey' => $password,
+                    'templateId' => $smsDSN->getParam('templateId'),
```
stnguyen90, line 413: Previous versions used the `_APP_SMS_FROM` env var so we should fallback to that.

```suggestion
                    'templateId' => $smsDSN->getParam('templateId', $from),
```

app/controllers/api/messaging.php
```diff
@@ -313,24 +313,23 @@
...
+    ->action(function (string $providerId, string $name, string $templateId, string $senderId, string $authKey, ?bool $enabled, Event $queueForEvents, Database $dbForProject, Response $response) {
         $providerId = $providerId == 'unique()' ? ID::unique() : $providerId;
 
         $options = [];
+        $credentials = [];
 
-        if (!empty($from)) {
-            $options['from'] = $from;
+        if (!empty($templateId)) {
+            $credentials['templateId'] = $templateId;
```
stnguyen90, line 330: Should this be in $options since it doesn't need to be encrypted?

src/Appwrite/Platform/Workers/Messaging.php
christyjacob4: Okay lets apply this suggestion 👍

app/controllers/api/messaging.php
christyjacob4: Yea makes sense. lets be consistent with what we use in other places

stnguyen90: @christyjacob4, oye, I'm not sure what the convention was when this was first created. For example, `isEuRegion` (a bool) is in `$credentials`:

https://github.com/appwrite/appwrite/blob/644d078f7f7054ffd0616c35d4df028374b5cd1b/app/controllers/api/messaging.php#L82-L84

Also host and port:

https://github.com/appwrite/appwrite/blob/644d078f7f7054ffd0616c35d4df028374b5cd1b/app/controllers/api/messaging.php#L246-L254

Let's just leave it.

stnguyen90 (approved):

Issue #7682, opened 2024-02-27
🐛 Wrong error message when changing bucket settings


str0tta-salesmart:
### 👟 Reproduction steps

1. Create an account and a project if you already didn't.
2. Create a bucket.
3. Go to its settings page.
4. Change ANY setting to a different value.

### 👍 Expected behavior

The bucket settings change.

### 👎 Actual Behavior

An error message pops out saying "The file you are trying to upload is above the limit allowed for your plan. Upgrade to higher plan to increase the limit."

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

Linux

### 🧱 Your Environment

Appwrite version: 0.12.102

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

### Additional Info

Error message:
![image](https://github.com/appwrite/appwrite/assets/151525241/a9256413-2eec-4052-a024-39b5c8b4ada5)

Sniffed HTTP request:
```bash
curl 'https://cloud.appwrite.io/v1/storage/buckets/65de2e1b397683ff4f4a' \
  -X 'PUT' \
  -H 'authority: cloud.appwrite.io' \
  -H 'accept: */*' \
  -H 'accept-language: it-IT,it;q=0.9,en-US;q=0.8,en;q=0.7' \
  -H 'content-type: application/json' \
[...]

PR #7843, opened 2024-03-19
Fix file size default limit

shimonewman:
Fixes: https://github.com/appwrite/appwrite/issues/7682


composer.lock
stnguyen90, line 1: Does this change need to be included?

stnguyen90 (changes requested):

christyjacob4 (approved):

PR #7855, opened 2024-03-21
swoole custom dispatch function

lohanidamodar:
## What does this PR do?

- Fixes random function timeout

## Test Plan

- manual tests and benchmarks

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/http.php
+++ b/app/http.php
@@ -8,6 +8,7 @@ use Swoole\Http\Server;
 use Swoole\Http\Request as SwooleRequest;
 use Swoole\Http\Response as SwooleResponse;
 use Utopia\App;
+use Utopia\Cache\Cache;
 use Utopia\CLI\Console;
 use Utopia\Config\Config;
 use Utopia\Database\Helpers\ID;
@@ -434,7 +435,6 @@ function fetchDomains()
                 $time = DateTime::now();
                 $limit = 1000;
                 $sum = $limit;
-                $total = 0;
                 $latestDocument = null;
 
                 while ($sum === $limit) {
@@ -454,7 +454,6 @@ function fetchDomains()
                     }
 
                     $sum = count($results);
-                    $total = $total + $sum;
                     foreach ($results as $document) {
                         $domain = $document->getAttribute('domain');
                         if (str_ends_with($domain, App::getEnv('_APP_DOMAIN_FUNCTIONS'))) {
@@ -465,8 +464,8 @@ function fetchDomains()
                     $latestDocument = !empty(array_key_last($results)) ? $results[array_key_last($results)] : null;
                 }
                 $lastSyncUpdate = $time;
-                if ($total > 0) {
-                    Console::log("Sync domains tick: {$total} domains were updated in");
+                if ($sum > 0) {
+                    Console::log("Sync domains tick: {$sum} domains were updated in");
                 }
             } catch (Throwable $th) {
                 Console::error($th->getMessage());
--- a/app/http.php
+++ b/app/http.php
@@ -20,11 +20,17 @@ use Utopia\Database\Database;
 use Utopia\Database\Document;
 use Utopia\Swoole\Files;
 use Appwrite\Utopia\Request;
+use Swoole\Timer;
+use Utopia\Database\DateTime;
+use Utopia\Database\Query;
 use Utopia\Logger\Log;
 use Utopia\Logger\Log\User;
 use Utopia\Pools\Group;
 
+const DOMAIN_SYNC_TIMER = 30; // 30 seconds
+
 $domains = [];
+$lastSyncUpdate = null;
 
 $http = new Server("0.0.0.0", App::getEnv('PORT', 80), SWOOLE_PROCESS);
 
@@ -44,39 +50,41 @@ $http
         'buffer_output_size' => $payloadSize,
     ]);
 
-function dispatch(Server $server, $fd, $type, $data = null) {
+function dispatch(Server $server, $fd, $type, $data = null)
+{
     global $workerNumber, $domains;
-    
+
     $safeThreadsPercent = intval(App::getEnv('_APP_SAFE_THREADS_PERCENT', 80)) / 100;
 
     $safeThreads = (int) floor($workerNumber * $safeThreadsPercent);
 
     $risky = false;
-    if($data && str_contains($data, 'POST') && str_contains($data, '/executions')) {
+    if ($data && str_contains($data, 'POST') && str_contains($data, '/executions')) {
         $risky = true;
-    } else if ($data && str_contains($data, 'GET')){
+    } elseif ($data && str_contains($data, '.appwrite.global')) {
+        $risky = true;
+    } elseif ($data) {
         $lines = explode($data, '\n');
-        if(count($lines) > 2) {
+        if (count($lines) > 2) {
             $host = trim(explode('Host: ', $lines[1])[1]);
-            if(array_key_exists($host, $domains)) {
+            if (array_key_exists($host, $domains)) {
                 $risky = true;
             }
         }
     }
 
-    if($risky) {
+    if ($risky) {
         var_dump('execution request sending to unsafe thread');
-        for($j = $safeThreads; $j < $workerNumber; $j++) {
-            if($server->getWorkerStatus($j) === 2) {
+        for ($j = $safeThreads; $j < $workerNumber; $j++) {
+            if ($server->getWorkerStatus($j) === 2) {
                 return $j;
             }
         }
         return rand($safeThreads, $workerNumber - 1);
-
     }
 
-    for($i = 0; $i < $workerNumber; $i++) {
-        if($server->getWorkerStatus($i) === 2) {
+    for ($i = 0; $i < $workerNumber; $i++) {
+        if ($server->getWorkerStatus($i) === 2) {
             return $i;
         }
     }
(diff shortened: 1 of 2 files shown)


app/http.php
```diff
@@ -327,4 +378,50 @@
     }
 });
 
+function getDomains(Database $dbForConsole, &$lastSyncUpdate, &$domains, $pools)
+{
+    go(function () use ($dbForConsole, &$lastSyncUpdate, &$domains, $pools) {
+        Timer::tick(DOMAIN_SYNC_TIMER * 1000, function () use ($dbForConsole, &$domains, &$lastSyncUpdate, $pools) {
+            $time = DateTime::now();
+            $timerStart = \microtime(true);
+            var_dump($domains);
```
eldadfux, line 387: Reminder

app/http.php
```diff
@@ -36,11 +44,53 @@
...
+function dispatch(Server $server, $fd, $type, $data = null)
+{
+    global $workerNumber, $domains;
+
+    $safeThreadsPercent = intval(App::getEnv('_APP_SAFE_THREADS_PERCENT', 80)) / 100;
+
+    $safeThreads = (int) floor($workerNumber * $safeThreadsPercent);
+
+    $risky = false;
+    if ($data && str_contains($data, 'POST') && str_contains($data, '/executions')) {
```
Meldiron, line 62: $data also includes body, right? I am worried there will be false positive.
Can we split and only get first line for this check? First line should only have method and URL. Then we can also do checks for starts_with and ends_with.

app/http.php
```diff
@@ -36,11 +44,53 @@
...
+        $risky = true;
+    } elseif ($data && str_contains($data, '.appwrite.global')) {
+        $risky = true;
+    } elseif ($data) {
+        $lines = explode($data, '\n');
+        if (count($lines) > 2) {
+            $host = trim(explode('Host: ', $lines[1])[1]);
```
Meldiron, line 69: Hmm, what's most trust-worthy? Host header, or domain parsed from first line (method+url)? Maybe we can use Domains library to simplify the work.

app/http.php
```diff
@@ -36,11 +44,53 @@
...
+            if (array_key_exists($host, $domains)) {
+                $risky = true;
+            }
+        }
+    }
+
+    if ($risky) {
+        var_dump('execution request sending to unsafe thread');
```
Meldiron, line 77: Leftover

app/http.php
```diff
@@ -36,11 +44,53 @@
...
+        for ($j = $safeThreads; $j < $workerNumber; $j++) {
+            if ($server->getWorkerStatus($j) === 2) {
```
Meldiron, line 79: Let's document what 2 means, or use const

app/http.php
```diff
@@ -36,11 +44,53 @@
...
+                return $j;
+            }
+        }
+        return rand($safeThreads, $workerNumber - 1);
+    }
+
+    for ($i = 0; $i < $workerNumber; $i++) {
+        if ($server->getWorkerStatus($i) === 2) {
```
Meldiron, line 87: Let's document what 2 means, or use const if swoole has one.

app/http.php
```diff
@@ -36,11 +44,53 @@
         // 'document_root' => __DIR__.'/../public',
         // 'enable_static_handler' => true,
         'http_compression' => true,
+        'dispatch_func' => 'dispatch',
         'http_compression_level' => 6,
         'package_max_length' => $payloadSize,
         'buffer_output_size' => $payloadSize,
     ]);
 
+function dispatch(Server $server, $fd, $type, $data = null)
```
eldadfux, line 53: Add a docblock to explain our strategy and steps to achieve it.

app/http.php
```diff
@@ -36,11 +44,53 @@
...
+        }
+        return rand($safeThreads, $workerNumber - 1);
+    }
+
+    for ($i = 0; $i < $workerNumber; $i++) {
+        if ($server->getWorkerStatus($i) === 2) {
+            return $i;
+        }
+    }
+    return rand(0, $safeThreads - 1);
```
Meldiron, line 91: Can we do `rand(0, workerNumber - 1)` here instead? If I understand correctly, this is for safe requests. We can use any worker for that.

app/http.php
eldadfux, line 62: Wouldn't it be easier to first parse the payload and the set the conditions? It would also be more effective than the extensive usage in str_contains

app/http.php
eldadfux, line 79: What is worker status 2? Let's add a reference to relevant docs. Maybe use constants.

app/http.php
```diff
@@ -207,6 +257,8 @@
             $dbForConsole->createCollection('bucket_' . $bucket->getInternalId(), $attributes, $indexes);
         }
 
+        getDomains($dbForConsole, $lastSyncUpdate, $domains, $pools);
+
         $pools->reclaim();
```
Meldiron, line 262: Let's not reclaim here, as dbForConsole is sent to getDomains, and there we have infinite loop.

app/http.php
```diff
@@ -327,4 +378,50 @@
...
+                    }
+                    $domains[$domain] = true;
+                }
+                $latestDocument = !empty(array_key_last($results)) ? $results[array_key_last($results)] : null;
+            }
+
+            $lastSyncUpdate = $time;
+            $timerEnd = \microtime(true);
+
+            $pools->reclaim();
```
Meldiron, line 420: Let's not reclaim. Its fine to steal 1 connection forever, and never put it back to pool.

app/http.php
```diff
@@ -327,4 +378,50 @@
...
+
+            Console::log("Sync domains tick: {$total} domains were updated in " . ($timerEnd - $timerStart) . " seconds");
```
Meldiron, line 422: Hmm, this could end up as spam. Can we print only if there was actually some new rules?

app/http.php
lohanidamodar: We can, but I also worry about the performance?

app/http.php
```diff
@@ -327,4 +378,50 @@
     }
 });
 
+function getDomains(Database $dbForConsole, &$lastSyncUpdate, &$domains, $pools)
+{
+    go(function () use ($dbForConsole, &$lastSyncUpdate, &$domains, $pools) {
```
Meldiron, line 383: Hmm, can we try to add `sleep(10)` in here, all over the place? Just to ensure if anything goes wrong here, master doesn't freeze and can still process HTTP requests.

app/http.php
Meldiron, line 387: Leftover

app/http.php
lohanidamodar: First line only has path `/v1` and second line has the host

app/http.php
```diff
@@ -327,4 +378,50 @@
...
+                }
+                $queries[] = Query::equal('resourceType', ['function']);
+
+                $results = Authorization::skip(fn () =>  $dbForConsole->find('rules', $queries));
+
+                $sum = count($results);
+                $total = $total + $sum;
+                foreach ($results as $document) {
+                    $domain = $document->getAttribute('domain');
+                    if (str_contains($domain, '.appwrite.global')) {
```
Meldiron, line 409: str_ends_with is safer

app/http.php
lohanidamodar: no because we reach here only hwne there are no idle threads, and we don't want safe request going to unsafe thread and get blocked in the loop like before.

app/http.php
Meldiron: We can benchmark, split with limit 2 should be very fast - after first split its quick. First split is after method and path, which shouldtn be too much, max ~2k chars.

I would love to avoid this.. Scarry to have a bug that causes random timeouts when there is specific word inside a body of my request

app/http.php
Meldiron: I see.

app/http.php
christyjacob4: I looked into it, splitting by \n itself requires the code to scan through the entire $data.. I don't see a value trying to parse it because the compute we save by removing str_contains is taken up in splitting. Plus, the check the host header requires us to find the host header among all the split lines.

Meldiron: I dont think full scan is needed if we limit the split,third parameter of https://www.php.net/manual/en/function.explode.php

If we tell it to split only into 2 parts by the first occurrence, I don't want to believe PHP will continue looking for others. We can benchmark it to double check.

app/http.php
```diff
@@ -207,8 +257,6 @@
             $dbForConsole->createCollection('bucket_' . $bucket->getInternalId(), $attributes, $indexes);
         }
 
-        $pools->reclaim();
```
stnguyen90, line 210: Why was this removed?

app/http.php
```diff
@@ -36,11 +40,57 @@
...
+    global $workerNumber;
+
+    $safeThreadsPercent = intval(App::getEnv('_APP_SAFE_THREADS_PERCENT', 80)) / 100;
+
+    $safeThreads = (int) floor($workerNumber * $safeThreadsPercent);
+
+    $risky = false;
+    if ($data && str_contains($data, 'POST') && str_contains($data, '/executions')) {
+        $risky = true;
+    } elseif ($data && str_contains($data, '.appwrite.global')) {
```
stnguyen90, line 72: Can we use the function domain environment variable here instead of hard coding .appwrite.global?

ekamalov: what happens if we use custom domain?

PR #7870, opened 2024-03-27
Fix not hiding tokens for clients via realtime

abnegate:
## What does this PR do?

Fix secret being set in realtime payload

## Test Plan

Existing tests

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/account.php
```diff
@@ -1845,6 +1845,9 @@
             'team' => '',
         ];
 
+        // Hide secret for clients
+        $token->setAttribute('secret', ($isPrivilegedUser || $isAppUser) ? $tokenSecret : '');
```
stnguyen90, line 1849: This will remove the secret from other destinations like functions and webhooks, too won't it?

abnegate: Yep, will fix

stnguyen90 (changes requested):
Did the tests break?

stnguyen90 (approved):
looks good. there's just a merge conflict.

PR #7896, opened 2024-04-01
Create failed execution from worker if deployment doesn't exist

vermakhushboo:
## What does this PR do?

For a scenario where an execution is triggered through worker and deployment doesn't exist, we would create a failed execution to let user know about the failure.

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Workers/Functions.php
+++ b/src/Appwrite/Platform/Workers/Functions.php
@@ -200,6 +200,8 @@ class Functions extends Action
      * @param string $path
      * @param string $method
      * @param Document $user
+     * @param string|null $jwt
+     * @param string|null $event
      * @throws Exception
      */
     private function fail(
@@ -209,7 +211,21 @@ class Functions extends Action
         string $path,
         string $method,
         Document $user,
+        string $jwt = null,
+        string $event = null,
     ): void {
+        $headers['x-appwrite-trigger'] = $trigger;
+        $headers['x-appwrite-event'] = $event ?? '';
+        $headers['x-appwrite-user-id'] = $user->getId() ?? '';
+        $headers['x-appwrite-user-jwt'] = $jwt ?? '';
+
+        $headersFiltered = [];
+        foreach ($headers as $key => $value) {
+            if (\in_array(\strtolower($key), FUNCTION_ALLOWLIST_HEADERS_REQUEST)) {
+                $headersFiltered[] = ['name' => $key, 'value' => $value];
+            }
+        }
+
         $executionId = ID::unique();
         $execution = new Document([
             '$id' => $executionId,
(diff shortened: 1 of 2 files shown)


src/Appwrite/Platform/Workers/Functions.php
```diff
@@ -248,15 +310,17 @@ private function execute(
             $deployment = $dbForProject->getDocument('deployments', $deploymentId);
 
         if ($deployment->getAttribute('resourceId') !== $functionId) {
-            throw new Exception('Deployment not found. Create deployment before trying to execute a function');
+            $this->fail($dbForProject, $function, $trigger, $path, $method, $user, $jwt, $event);
```
eldadfux, line 313: Let's provide a verbose message that can appear in the execution logs.

src/Appwrite/Platform/Workers/Functions.php
```diff
@@ -248,15 +310,17 @@ private function execute(
...
+            return;
         }
 
-            /** Check if build has exists */
-            $build = $dbForProject->getDocument('builds', $deployment->getAttribute('buildId', ''));
+        /** Check if the build exists */
+        $build = $dbForProject->getDocument('builds', $deployment->getAttribute('buildId', ''));
         if ($build->isEmpty()) {
             throw new Exception('Build not found');
```
eldadfux, line 325: These build errors can also be propagated to devs.

eldadfux (changes requested):

src/Appwrite/Platform/Workers/Functions.php
```diff
@@ -194,6 +194,70 @@ function: $function,
...
+            'trigger' => $trigger,
+            'status' => 'failed',
+            'responseStatusCode' => 400,
+            'responseHeaders' => [],
+            'requestPath' => $path,
+            'requestMethod' => $method,
+            'requestHeaders' => $headersFiltered,
+            'errors' => $message,
+            'logs' => '',
+            'duration' => 0.0,
```
eldadfux, line 248: Can we still calculate duration? Its important because compute was still used.

src/Appwrite/Platform/Workers/Functions.php
```diff
@@ -194,6 +194,70 @@ function: $function,
...
+        $execution = new Document([
+            '$id' => $executionId,
+            '$permissions' => $user->isEmpty() ? [] : [Permission::read(Role::user($user->getId()))],
+            'functionInternalId' => $function->getInternalId(),
+            'functionId' => $function->getId(),
+            'deploymentInternalId' => '',
+            'deploymentId' => '',
+            'trigger' => $trigger,
+            'status' => 'failed',
+            'responseStatusCode' => 400,
```
eldadfux, line 241: Should we give status code like 0? After all request didn't really go through.

src/Appwrite/Platform/Workers/Functions.php
vermakhushboo: Even in the execute method, we start calculating the duration after the execution document is created. Since we throw the error just after creating the document and since we're not actually executing anything, the duration should be 0 in my opinion.

eldadfux (approved):

Issue #7914, opened 2024-04-03
🐛 Bug Report: Custom SMTP doesn't work for messaging


stnguyen90:
### 👟 Reproduction steps

1. Set up a custom SMTP provider that has SSL/TLS
2. Create an email message

### 👍 Expected behavior

The email sends.

### 👎 Actual Behavior

Messages are stuck at processing:

<img width="1171" alt="image" src="https://github.com/appwrite/appwrite/assets/1477010/91d25da8-07ff-420e-b803-47ee63690146">

Worker crashes with:

```
appwrite-worker-messaging  | [Job] Received Job (660db82b77efb7.33903010).
appwrite-worker-messaging  | [2024-04-03 20:12:27 #7.0] WARNING ProcessPool::wait(): worker#0 abnormal exit, status=0, signal=11
appwrite-worker-messaging  | A bug occurred in Swoole-v5.1.2, please report it.
appwrite-worker-messaging  | The Swoole developers probably don't know about it,
appwrite-worker-messaging  | and unless you report it, chances are it won't be fixed.
appwrite-worker-messaging  | You can read How to report a bug doc before submitting any bug reports:
appwrite-worker-messaging  | >> https://github.com/swoole/swoole-src/blob/master/.github/ISSUE.md 
appwrite-worker-messaging  | Please do not send bug reports in the mailing list or personal letters.
appwrite-worker-messaging  | The issue page is also suitable to submit feature requests.
appwrite-worker-messaging  | 
appwrite-worker-messaging  | OS: Linux 5.15.49-linuxkit #1 SMP PREEMPT Tue Sep 13 07:51:32 UTC 2022 aarch64
appwrite-worker-messaging  | GCC_VERSION: 13.2.1 20231014
appwrite-worker-messaging  | OPENSSL_VERSION: OpenSSL 3.1.4 24 Oct 2023
appwrite-worker-messaging  | PHP_VERSION : 8.3.3
appwrite-worker-messaging  | 
appwrite-worker-messaging  | [Worker] Worker 0 is ready!
appwrite-worker-messaging  | Worker messaging  started
appwrite-worker-messaging  | [Job] Received Job (660db8643afa21.09449913).
appwrite-worker-messaging  | [2024-04-03 20:13:24 #7.0] WARNING ProcessPool::wait(): worker#0 abnormal exit, status=0, signal=11
appwrite-worker-messaging  | A bug occurred in Swoole-v5.1.2, please report it.
[...]

PR #7915, opened 2024-04-03
Fix worker crash when using custom SMTP provider

stnguyen90:
## What does this PR do?

Due to a bug in the Swoole TCP hook, the worker crashes when doing the TLS handshake. This PR disables the Swoole TCP hook just like we do in the mails worker.

Fixes: https://github.com/appwrite/appwrite/issues/7914

## Test Plan

Manually created an email and the logs show it was processed successfully:

```
appwrite-worker-messaging  | [Worker] Worker 0 is ready!
appwrite-worker-messaging  | Worker messaging  started
appwrite-worker-messaging  | [Job] Received Job (660dbf5c37b395.78058781).
appwrite-worker-messaging  | [Job] (660dbf5c37b395.78058781) successfully run.
appwrite-worker-messaging  | [Job] Received Job (660dc049d8a217.47574305).
appwrite-worker-messaging  | [Job] (660dc049d8a217.47574305) successfully run.
```

The Console also shows success for the email:

image

Note: the SMS failed, but that's because of the provider returned an error message. This is still good because it proves API calls still send as expected. 

## Related PRs and Issues

* https://github.com/appwrite/appwrite/issues/7914
* https://github.com/appwrite/appwrite/pull/6966
* https://github.com/swoole/swoole-src/issues/4909

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Workers/Messaging.php
+++ b/src/Appwrite/Platform/Workers/Messaging.php
@@ -4,6 +4,7 @@ namespace Appwrite\Platform\Workers;
 
 use Appwrite\Event\Usage;
 use Appwrite\Messaging\Status as MessageStatus;
+use Swoole\Runtime;
 use Utopia\App;
 use Utopia\CLI\Console;
 use Utopia\Config\Config;
@@ -80,6 +81,7 @@ class Messaging extends Action
         Device $deviceForLocalFiles,
         Usage $queueForUsage
     ): void {
+        Runtime::setHookFlags(SWOOLE_HOOK_ALL ^ SWOOLE_HOOK_TCP);
         $payload = $message->getPayload() ?? [];
 
         if (empty($payload)) {


src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -80,6 +81,7 @@ public function action(
         Device $deviceForLocalFiles,
         Usage $queueForUsage
     ): void {
+        Runtime::setHookFlags(SWOOLE_HOOK_ALL ^ SWOOLE_HOOK_TCP);
```
abnegate, line 84: We can do this outside the class definition right? Don't think we need to it for every job

stnguyen90: My concern with putting it outside of the action is that it would get triggered in other workers. For example, I added `Console::warning('Messaging class');` outside the class and it was executed in the build worker:

image

abnegate: Ohh makes sense because they're all initialized together. All good :+1:

abnegate (approved):

PR #7933, opened 2024-04-08
prevent functions domain and subdomain to be added as custom domain

lohanidamodar:
## What does this PR do?

 prevent functions domain and subdomain to be added as custom domain

## Test Plan

- Manually verified

## Related PRs and Issues

- (Related PR or issue)

## Checklist

- [x] Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [x] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

PR #7934, opened 2024-04-08
Prevent functions domain to be used as custom domain

lohanidamodar:
## What does this PR do?

Prevent functions domain to be used as custom domain

## Test Plan

Manually tested

## Related PRs and Issues

- https://github.com/appwrite/appwrite/pull/7933

## Checklist

### DIFF
--- a/app/controllers/api/proxy.php
+++ b/app/controllers/api/proxy.php
@@ -50,7 +50,7 @@ App::post('/v1/proxy/rules')
             throw new Exception(Exception::GENERAL_ARGUMENT_INVALID, 'You cannot assign your main domain to specific resource. Please use subdomain or a different domain.');
         }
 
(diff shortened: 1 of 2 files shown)


app/controllers/api/proxy.php
```diff
@@ -49,6 +49,12 @@
         if ($domain === $mainDomain) {
             throw new Exception(Exception::GENERAL_ARGUMENT_INVALID, 'You cannot assign your main domain to specific resource. Please use subdomain or a different domain.');
         }
+
+        $functionsDomain = App::getEnv('_APP_DOMAIN_FUNCTIONS', '');
```
eldadfux, line 53: Should we do the same for regular custom domains? or this is taking care of this and the error message can be more accurate?

lohanidamodar: I don't understand what you mean?

eldadfux (approved):

PR #7950, opened 2024-04-10
Fix: recovery code removal

Meldiron:
## What does this PR do?

When recovery code gets used, its removed fro marray but indexes arent fixed.
Because of that, Appwrite then returns those as object instead of array:

CleanShot 2024-04-10 at 16 00 39@2x

I know this is issue with index 0 missing, as per test here:

CleanShot 2024-04-10 at 16 00 24@2x

Code in this PR ensires array is recreated and has proper indexing - will remain proper array everywhere. Performance isn't concern as it has only a few codes always.

## Test Plan

None

## Related PRs and Issues

x

## Checklist

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -3821,6 +3821,7 @@ App::delete('/v1/account/mfa/authenticators/:type')
             $mfaRecoveryCodes = $user->getAttribute('mfaRecoveryCodes', []);
             if (in_array($otp, $mfaRecoveryCodes)) {
                 $mfaRecoveryCodes = array_diff($mfaRecoveryCodes, [$otp]);
+                $mfaRecoveryCodes = array_values($mfaRecoveryCodes);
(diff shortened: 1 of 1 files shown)


eldadfux (approved):

app/controllers/api/account.php
```diff
@@ -3821,6 +3821,7 @@
             $mfaRecoveryCodes = $user->getAttribute('mfaRecoveryCodes', []);
             if (in_array($otp, $mfaRecoveryCodes)) {
                 $mfaRecoveryCodes = array_diff($mfaRecoveryCodes, [$otp]);
+                $mfaRecoveryCodes = array_values($mfaRecoveryCodes);
```
stnguyen90, line 3824: when/how is the used OTP removed?

Meldiron: Array_dif is removing it. It's done in 2 places. Few lines below, we are updating users document.

PR #7990, opened 2024-04-18
Read 20MBs but send response in chunks of 2MB

vermakhushboo:
## What does this PR do?

The code efficiently reads data in chunks of APP_STORAGE_READ_BUFFER size and sends it in chunks of MAX_OUTPUT_CHUNK_SIZE, optimizing the transfer process for large files.

## Test Plan

Tested using node SDK. Download time went down `4x` from 68528ms to 25188ms.

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/storage.php
```diff
@@ -1109,17 +1109,32 @@
...
-                );
+            $buffer = '';
+            $bytesRead = 0;
+
+            while ($bytesRead < $size) {
+                // Determine how many bytes to read in this iteration
+                $bytesToRead = min(APP_STORAGE_READ_BUFFER, $size - $bytesRead);
+
+                // Read data from the device and append to the buffer
+                $buffer .= $deviceFiles->read($path, $bytesRead, $bytesToRead);
```
lohanidamodar, line 1120: I don't think we should append to buffer, because this means by the end whole file in the memory, and this logic is to prevent exactly that. Just keep the recent chunk in the buffer

lohanidamodar (changes requested):
Please check the comment

app/controllers/api/storage.php
vermakhushboo: We are not appending the entire file to the buffer. If you look at the loop below, we remove each chunk from the buffer after processing. 

```
// Remove the sent chunk from the buffer
$buffer = substr($buffer, MAX_OUTPUT_CHUNK_SIZE);
```

lohanidamodar: Why do this, why not instead of `$buffer .= $deviceFiles->read($path, $bytesRead, $bytesToRead);` simply use `$buffer = $deviceFiles->read($path, $bytesRead, $bytesToRead);` and when loop ends simply `$buffer = null`. The `substr` adds more processing

vermakhushboo:
Closing it as this PR is no longer relevant

PR #8008, opened 2024-04-23
Fix server errors from invalid or outdated cookies

PineappleIOnic:
## What does this PR do?

This PR introduces an additional check before we access `factors` on the $session object to make sure it exists since invalid or outdated cookies will cause the session check to pass for factors to be null resulting in a 500 error on all requests including non-protected routes.

## Checklist

### DIFF
--- a/app/controllers/shared/api.php
+++ b/app/controllers/shared/api.php
@@ -290,7 +290,7 @@ App::init()
             $minimumFactors = ($mfaEnabled && $hasMoreFactors) ? 2 : 1;
 
             if (!in_array('mfa', $route->getGroups())) {
-                if ($session && \count($session->getAttribute('factors')) < $minimumFactors) {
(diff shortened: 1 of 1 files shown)


app/controllers/shared/api.php
```diff
@@ -290,7 +290,7 @@
             $minimumFactors = ($mfaEnabled && $hasMoreFactors) ? 2 : 1;
 
             if (!in_array('mfa', $route->getGroups())) {
-                if ($session && \count($session->getAttribute('factors')) < $minimumFactors) {
+                if ($session && !empty($session->getAttribute('factors')) && \count($session->getAttribute('factors')) < $minimumFactors) {
```
abnegate, line 293: Might be simpler to pass an empty array as the default for getAttribute inside count:

```suggestion
                if ($session && \count($session->getAttribute('factors', [])) < $minimumFactors) {
```

abnegate (approved):

Issue #5232, opened 2023-03-15
🚀 Feature:  Security Scans like SAST, DAST, FOSS, CAST in pipeline


MananShah9:
### 🔖 Feature description

In the build process, all the security scans like SAST DAST FOSS and CAST should be performed to deliver a safer product

### 🎤 Pitch

While making appwrite safer and a product of a higher quality it opens up the possibility for many MNC's to adopt the product and thereby increasing the impact and reach of the same.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

stnguyen90:
@MananShah9, thanks for creating this issue! 🙏🏼 Do you have suggestions on what to use? If you do, please also provide alternatives and the reason for choosing that tool over the alternative.

MananShah9:
So I have used IBM App scan, checkmarx, Sonatype and netsparker before... I'm not really sure if I have reasons ready to choose one over other and might need to get involved a bit more to be able to brainstorm the same.

joeyouss:
Hi @MananShah9 thanks for the suggestion. Do you mind clarifying for the community, including myself, what these scans do and why they'd be beneficial? Some of us without the background can't properly appreciate your suggestions, because we might not understand these terms.
This way, we can have more community participation in the discussion. Thanks!

MananShah9:
These scans are like security checkups for softwares.

Think of them as tools that look closely at the source code or the environment to interact with the running app to find any weak points that could be exploited. 

Here's what each of these Scan does:

SAST (Static Application Security Testing): This tool reads the source code and figures out where there might be security problems.

DAST (Dynamic Application Security Testing): This tool watches the software while it's running and finds problems that only show up when the software is actually being used (only for web apps).

CAST (Container Scanning): This tool looks inside docker containers to find any issues that might be hiding there.

FOSS Scans (Free Open-Source Software Scans): This tool helps us find weaknesses in the free software packages we use in our own software.

It's really hard for us to find these problems just by looking at the code ourselves. 

The company I work for (HSBC Technologies) uses tools like the ones you can find here:

https://www.invicti.com/
https://checkmarx.com/
https://www.sonatype.com/products/vulnerability-scanner

[...]

PR #8021, opened 2024-04-24
Add OSV Scanner for vulnerability scans

stnguyen90:
## What does this PR do?

This workflow action uses OSV Scanner, an open source vulnerability
scanner by Google. We're using OSV Scanner because it has:

* good usability - JSON output and multiple options
* good accuracy - OSV database from google and support
  for multiple languages including PHP

## Test Plan

None

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/5232

## Checklist


abnegate:
@stnguyen90 I don't see any analysis results [here](https://github.com/appwrite/appwrite/security/code-scanning?query=pr%3A8021), the comment above makes it sound like there should be results even if there were no issues detected, is that correct?

stnguyen90:
@abnegate I think it'll show up once this is merged and the scan runs on main

abnegate (approved):

PR #8026, opened 2024-04-25
getEnv on worker.php

shimonewman:


app/worker.php
```diff
@@ -215,7 +215,7 @@
 
 $pools = $register->get('pools');
 $platform = new Appwrite();
-$args = $_SERVER['argv'];
+$args = $platform->getEnv('argv');
```
stnguyen90, line 218: Does Platform have `getEnv()`? I couldn't find it in https://github.com/utopia-php/platform/blob/main/src/Platform/Platform.php

shimonewman: Now it will (:

christyjacob4: lets update composer with the new library version

christyjacob4: Lets update composer with https://github.com/utopia-php/platform/releases/tag/0.7.0

composer.lock
christyjacob4, line 1: Released a new version here https://github.com/utopia-php/platform/releases/tag/0.5.2 lets update composer

christyjacob4 (approved):
Lets update composer dependencies and then merge

christyjacob4 (approved):

PR #8055, opened 2024-05-01
Feat migration logging improvements

PineappleIOnic:
## What does this PR do?

Adds extra data to our sentry logs on migrations

## Checklist


loks0n (approved):

src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -291,21 +294,26 @@ protected function processMigration(Document $project, Document $migration, Log
             if (!empty($sourceErrors) || !empty($destinationErrors)) {
                 $migrationDocument->setAttribute('status', 'failed');
                 $migrationDocument->setAttribute('stage', 'finished');
+                $log->addTag('stage', 'finished');
 
-                $errorMessages = [];
+                $prettyMessages = [];
+                $simpleMessages = [];
```
christyjacob4, line 300: why do we need two variables here? I see they contain the same information. resourceType, resourceId & message

PR #8057, opened 2024-05-01
Migration Logging Improvements

PineappleIOnic:
## What does this PR do?

This PR adds more tags to migration logs and explicitly throws the exception under one name for easy tracking

## Checklist


src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -256,6 +257,7 @@ protected function processMigration(Document $project, Document $migration, Log
             $migrationDocument = $this->dbForProject->getDocument('migrations', $migration->getId());
             $migrationDocument->setAttribute('stage', 'processing');
             $migrationDocument->setAttribute('status', 'processing');
+            $log->addTag('stage', 'processing');
```
stnguyen90, line 260: Looks like you want to track something that changes over time and I don't think we need to search on this, right? If so, the best way to do that is probably with [breadcrumbs](https://github.com/utopia-php/logger/blob/9151b7d16eab18d4c37c34643041cc0f33ca4a6c/src/Logger/Log.php#L354). For example, in Sentry, it would look like https://docs.sentry.io/product/issues/issue-details/breadcrumbs/.

christyjacob4 (approved):

stnguyen90 (approved):

PR #8063, opened 2024-05-02
feat: add warning header

loks0n:
- Adds a new response header `X-Appwrite-Warning`

It can contain multiple warnings delimetered by `;`, for example:
```
X-Appwrite-Warning: Invalid SDK format version requested ; Wrong content-type requested
```

- Adds an error message when an SDK designed for a later version of Appwrite is used.


app/controllers/general.php
```diff
@@ -543,8 +543,16 @@ function router(App $utopia, Database $dbForConsole, callable $getProjectDB, Swo
             $response->addHeader('Strict-Transport-Security', 'max-age=' . (60 * 60 * 24 * 126)); // 126 days
         }
 
+        $responseVersion = System::getEnv('_APP_VERSION', 'UNKNOWN');
```
stnguyen90, line 546: Since this is going to be public, we should use the [`APP_VERSION_STABLE`](https://github.com/appwrite/appwrite/blob/a30d06bc585390658f2ef6677267756852c9a44e/app/init.php#L116) constant

stnguyen90 (changes requested):

stnguyen90 (approved):

app/controllers/general.php
```diff
@@ -543,8 +543,16 @@ function router(App $utopia, Database $dbForConsole, callable $getProjectDB, Swo
             $response->addHeader('Strict-Transport-Security', 'max-age=' . (60 * 60 * 24 * 126)); // 126 days
         }
 
+        $defaultResponseFormat = APP_VERSION_STABLE;
+        if ($defaultResponseFormat !== 'UNKNOWN') {
+            $defaultResponseFormat = \explode('.', $defaultResponseFormat);
+            $defaultResponseFormat[2] = '0';
```
stnguyen90, line 549: We should probably be defensive here and make sure the array has the right number of elements.

loks0n: Updated, but this is hardcoded, not user input so don't think there is huge risk here

christyjacob4 (approved):

PR #8075, opened 2024-05-05
Fix potential github action smells

ceddy4395:
## What does this PR do?
Hey! 🙂
I want to contribute the following changes to your workflow:

- Stop running workflows when there is a newer commit in PR
- Prevent running issue/PR actions on forks line
- Avoid deploying jobs on forks

(These changes are part of a research Study at TU Delft looking at GitHub Action Smells. [Find out more](https://ceddy4395.github.io/research/gha-smells.html))

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/.github/workflows/cleanup-cache.yml
+++ b/.github/workflows/cleanup-cache.yml
@@ -8,6 +8,8 @@ on:
 jobs:
   cleanup:
     runs-on: ubuntu-latest
+    permissions: 
+      actions: write
     steps:
       - name: Check out code
         uses: actions/checkout@v3
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,7 +8,7 @@ jobs:
   tests:
     name: Release
     runs-on: ubuntu-latest
-
+    if: github.repository == 'appwrite-appwrite'
     steps:
     - name: Checkout repository
       uses: actions/checkout@v3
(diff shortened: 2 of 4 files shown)


detunjiSamuel:
This is interesting, but why would you not want to run regular jobs/actions in a fork? @ceddy4395

.github/workflows/release.yml
```diff
@@ -8,7 +8,7 @@ jobs:
   tests:
     name: Release
     runs-on: ubuntu-latest
-
+    if: github.repository == 'appwrite-appwrite'
```
loks0n, line 11: Shouldnt this be appwrite/appwrite?

.github/workflows/cleanup-cache.yml
```diff
@@ -8,6 +8,8 @@ on:
 jobs:
   cleanup:
     runs-on: ubuntu-latest
+    permissions: 
+      actions: write
```
loks0n, line 12: What does this do?

.github/workflows/release.yml
ceddy4395: Ah yes you're correct! Allow me to fix that!

ceddy4395:
> This is interesting, but why would you not want to run regular jobs/actions in a fork? @ceddy4395

@detunjiSamuel The idea is not to eliminate all workflows from forks but only the ones which do some sort of releasing. In your case the fork will not have access to the correct secrets when logging into Docker Hub and will thus fail by default. Therefore, it's more ideal to just prevent the whole workflow from running.

HarshMN2345:

PR #8081, opened 2024-05-07
Don't kick user and require verification after enabling MFA

stnguyen90:
## What does this PR do?

* remove auth duration from `Auth::sessionVerify()` calls because the method doesn't require it
* ensure user isn't kicked out after enabling MFA as it is very disruptive to the user experience
* ensure session factors don't contain duplicates

## Test Plan

Manual

## Related PRs and Issues

None

## Checklist


TorstenDittmann (commented):
Can we have a test for that? :)

TorstenDittmann (approved):

PR #8136, opened 2024-05-16
Fix: executor failures

Meldiron:
## What does this PR do?

There is warning due to Swoole Table size (on executor) when hosntame of executor is too long. That lead's to activeRuntimes on executor not be up to date. When that happens, and there is maintenance on executor, it wipes function's folders. It thinks it doesn't exist anymore, but it does.

This PR fixes that problem.

## Test Plan

- [x] Manual QA. I could no longer reproduce the issue with code being randomly deleted in the middle of the build

## Related PRs and Issues

x

## Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -847,10 +847,10 @@ services:
 
   openruntimes-executor:
     container_name: openruntimes-executor
-    hostname: appwrite-executor
+    hostname: exc1
     <<: *x-logging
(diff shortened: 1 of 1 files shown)


vermakhushboo:
Related PR targeted to 1.5.x: https://github.com/appwrite/appwrite/pull/8147

Meldiron:
We are already on never version

PR #8145, opened 2024-05-17
Upgrade logger and add fallback calculation

PineappleIOnic:
## What does this PR do?

This PR upgrades the logger and adds a fallback calculation for older env-var types as to not break compatibility.

## Test Plan

Test with both new and old type, both work

## Checklist


app/init.php
```diff
@@ -732,6 +732,16 @@ function (mixed $value) {
         throw new Exception(Exception::GENERAL_SERVER_ERROR, "Logging provider not supported. Logging is disabled");
     }
 
+    // Old Sentry Format conversion
```
christyjacob4, line 735:
```suggestion
    // Old Sentry Format conversion. Fallback until the old syntax is completely deprecated. 
```

PR #8152, opened 2024-05-17
fix: improve cold start error

loks0n:
Screenshot 2024-05-17 at 18 16 37


app/controllers/general.php
```diff
@@ -288,6 +288,9 @@ function router(App $utopia, Database $dbForConsole, callable $getProjectDB, Swo
             $execution->setAttribute('logs', $executionResponse['logs']);
             $execution->setAttribute('errors', $executionResponse['errors']);
             $execution->setAttribute('duration', $executionResponse['duration']);
+
+        } catch (AppwriteException $ex) {
+            throw $ex;
```
christyjacob4, line 293: Whats the need for this ?

src/Appwrite/Extend/Exception.php
```diff
@@ -153,6 +153,7 @@ class Exception extends \Exception
     public const FUNCTION_NOT_FOUND                = 'function_not_found';
     public const FUNCTION_RUNTIME_UNSUPPORTED      = 'function_runtime_unsupported';
     public const FUNCTION_ENTRYPOINT_MISSING       = 'function_entrypoint_missing';
+    public const FUNCTION_SYNCHRONOUS_TIMEOUT      = 'function_synchronous_timeout';
```
christyjacob4, line 156: we need to add the config for this error in the errors.php file

src/Executor/Executor.php
```diff
@@ -327,8 +331,11 @@ public function call(string $method, string $path = '', array $headers = [], arr
             }
         }
 
-        if ((curl_errno($ch)/* || 200 != $responseStatus*/)) {
-            throw new Exception(curl_error($ch) . ' with status code ' . $responseStatus, $responseStatus);
+        if ($curlError) {
+            if ($curlError == CURLE_OPERATION_TIMEDOUT) {
+                throw new AppwriteException(AppwriteException::FUNCTION_SYNCHRONOUS_TIMEOUT, $curlErrorMessage, Response::STATUS_CODE_REQUEST_TIMEOUT);
```
christyjacob4, line 336: once you add this error config to errors.php, you dont need to pass in the status code here. Its picked up automatically from the config

christyjacob4: Also what is the need to handle this case specifically?

Meldiron, line 336: If we do what Christy suggested, add definition for exception, we might not need to define status code here.

christyjacob4 (commented):
lets also change the target branch to refactor-usage-sn

src/Executor/Executor.php
loks0n: Added. We handle the case specifically to throw a specific exception, in order to display a error page to the user, instead of a 500 with no body.

src/Appwrite/Extend/Exception.php
loks0n: Added

app/controllers/general.php
loks0n: The `createExecution` method is called inside this `try` block, and therefore without this the `AppwriteException` thrown within will be caught by the existing `catch` block below.

If the `AppwriteException is `caught`, we show an error page.

src/Executor/Executor.php
christyjacob4: okay cool 👍

app/controllers/general.php
Meldiron, line 293: We need to set some execution params here, like status, errors, body, ..
We also then need to save execution (createDocument) before we thorw.

loks0n: Done 👍

christyjacob4 (approved):

Meldiron (approved):

PR #8188, opened 2024-05-23
Sync 1.5.x with cloud

christyjacob4:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/config/collections.php
```diff
@@ -996,7 +996,7 @@
                 '$id' => ID::custom('_key_provider_providerUid'),
                 'type' => Database::INDEX_KEY,
                 'attributes' => ['provider', 'providerUid'],
-                'lengths' => [100, 100],
+                'lengths' => [128, 128],
```
stnguyen90, line 999: Does this mean we require some sort of migration? 🧐

app/console
stnguyen90, line 1: was this intended? The commit hash for this looks like [4.0.6](https://github.com/appwrite/console/releases/tag/4.0.6) 🧐

app/views/install/compose.phtml
```diff
@@ -727,7 +727,7 @@ services:
 
   openruntimes-executor:
     container_name: openruntimes-executor
-    hostname: appwrite-executor
+    hostname: exc1
```
stnguyen90, line 730: we need the variable updated

app/controllers/api/projects.php
```diff
@@ -73,11 +76,13 @@
     ->param('legalCity', '', new Text(256), 'Project legal City. Max length: 256 chars.', true)
     ->param('legalAddress', '', new Text(256), 'Project legal Address. Max length: 256 chars.', true)
     ->param('legalTaxId', '', new Text(256), 'Project legal Tax ID. Max length: 256 chars.', true)
+    ->inject('request')
     ->inject('response')
     ->inject('dbForConsole')
     ->inject('cache')
     ->inject('pools')
-    ->action(function (string $projectId, string $name, string $teamId, string $region, string $description, string $logo, string $url, string $legalName, string $legalCountry, string $legalState, string $legalCity, string $legalAddress, string $legalTaxId, Response $response, Database $dbForConsole, Cache $cache, Group $pools) {
+    ->inject('hooks')
```
stnguyen90, line 84: where is this resource injected?

app/views/install/compose.phtml
christyjacob4, line 730: We might need to remove this change as it can affect people upgrading 🤔

tests/e2e/Services/Databases/DatabasesBase.php
```diff
@@ -2098,16 +2128,17 @@ public function testDocumentsListQueries(array $data): array
...
-            'x-appwrite-project' => $this->getProject()['$id'],
-        ], $this->getHeaders()), [
-            'queries' => [
-                Query::search('actors', 'Tom')->toString(),
-            ],
-        ]);
-        $this->assertEquals(400, $documents['headers']['status-code']);
-        $this->assertEquals('Invalid query: Cannot query search on attribute "actors" because it is an array.', $documents['body']['message']);
+        // Todo: Disabled for CL - Uncomment after ProxyDatabase cleanup for find method
+        // $documents = $this->client->call(Client::METHOD_GET, '/databases/' . $databaseId . '/collections/' . $data['moviesId'] . '/documents', array_merge([
```
christyjacob4, line 2132: @abnegate can this be uncommented now?

app/views/install/compose.phtml
abnegate: Should be fine as long as we update env var default [here](https://github.com/appwrite/appwrite/blob/sync-1.5.x-with-cloud/app/config/variables.php#L767-L774) as Steven mentioned

tests/e2e/Services/Databases/DatabasesBase.php
abnegate: Not yet, we need this commented out until the `find` override is removed on cloud. This functionality is already tested in the db lib and won't change so not concerned here

abnegate (approved):

app/views/install/compose.phtml
christyjacob4: Yea that will fix it for new installations of 1.5.7 but not for people upgrading from 1.5.6

abnegate: Because the variable has `overwrite` set to true, the value will be updated for upgrades too

app/controllers/api/projects.php
```diff
@@ -195,35 +264,23 @@
...
-                ]);
+            try {
+                $dbForProject->createCollection($key, $attributes, $indexes);
+            } catch (Duplicate) {
+                // Collection already exists
             }
-            $dbForProject->createCollection($key, $attributes, $indexes);
         }
 
+        $hooks->trigger('afterProjectCreation', [ $project, $pools, $cache ]);
```
christyjacob4, line 282: @Meldiron is this still required

src/Appwrite/Extend/Exception.php
```diff
@@ -298,11 +301,20 @@ class Exception extends \Exception
...
     protected bool $publish;
 
-    public function __construct(string $type = Exception::GENERAL_UNKNOWN, string $message = null, int $code = null, \Throwable $previous = null)
+    public function __construct(string $type = Exception::GENERAL_UNKNOWN, string $message = null, int|string $code = null, \Throwable $previous = null)
     {
         $this->errors = Config::getParam('errors');
         $this->type = $type;
         $this->code = $code ?? $this->errors[$type]['code'];
+
+        if(\is_string($this->code)) {
```
christyjacob4, line 310: @Meldiron lets add a comment here to indicate this is done for SQL errors that are usually strings like `HY001` etc.

app/controllers/api/projects.php
Meldiron: It will be needed for next ZDT. We can safely keep this, if there is no hook for that name, it will just do nothing.

src/Appwrite/Extend/Exception.php
Meldiron: Will do, in which branch?

app/controllers/api/projects.php
christyjacob4: lets add a comment about it

src/Appwrite/Extend/Exception.php
christyjacob4: in the current branch

PR #8193, opened 2024-05-24
Feat remove random shared tables

abnegate:
## What does this PR do?

- Remove random assignment of projects to shared tables
- Use environment variable for setting shared tables db
- Run all tests using both shared and non-shared tables

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/projects.php
```diff
@@ -111,35 +111,8 @@
 
         $projectId = ($projectId == 'unique()') ? ID::unique() : $projectId;
 
-        $backups['database_db_fra1_v14x_02'] = ['from' => '03:00', 'to' => '05:00'];
-        $backups['database_db_fra1_v14x_03'] = ['from' => '00:00', 'to' => '02:00'];
-        $backups['database_db_fra1_v14x_04'] = ['from' => '00:00', 'to' => '02:00'];
-        $backups['database_db_fra1_v14x_05'] = ['from' => '00:00', 'to' => '02:00'];
-        $backups['database_db_fra1_v14x_06'] = ['from' => '00:00', 'to' => '02:00'];
-        $backups['database_db_fra1_v14x_07'] = ['from' => '00:00', 'to' => '02:00'];
```
lohanidamodar, line 119: Why are we removing these?

app/controllers/api/projects.php
```diff
@@ -152,37 +125,12 @@
...
-        }
-
-        // TODO: Allow overriding in development mode. Temporary until all projects are using shared tables.
-        if (
-            App::isDevelopment()
-            && System::getEnv('_APP_EDITION', 'self-hosted') !== 'self-hosted'
-            && $request->getHeader('x-appwrited-share-tables', false)
-        ) {
+        // TODO: Temporary until all projects are using shared tables.
+        if ($dsn === System::getEnv('_APP_DATABASE_SHARED_TABLES', '')) {
```
lohanidamodar, line 129: does this mean it will either be shared table or non-shared table all together?

lohanidamodar (commented):
some queries, rest looks good.

app/controllers/api/projects.php
abnegate: We don't have these DBs anymore, or any managed DBs, so I think it's redundant. cc @christyjacob4 to confirm

app/controllers/api/projects.php
abnegate: It will still depend on which database is assigned, so if by random assignment or by a database override the user gets the shared tables db, that project will get shared tables.

For tests, since we only have one DB, it will be either shared tables or not. This is the main reason I added an extra test job that will run all tests for all service with shared tables. Actually gives us much better coverage as well since we know all existing tests pass with both methods.

lohanidamodar: 🆒

tests/e2e/Services/Projects/ProjectsConsoleClientTest.php
```diff
@@ -3494,504 +3493,4 @@ public function testDeleteProject(): array
 
         return $data;
     }
-
-    public function testTenantIsolation(): void
```
christyjacob4, line 3498: Are we moving all these tests to cloud now ? Why did we remove them ?

.github/workflows/tests.yml
```diff
@@ -145,3 +145,6 @@ jobs:
 
       - name: Run ${{matrix.service}} Tests
         run: docker compose exec -T appwrite test /usr/src/code/tests/e2e/Services/${{matrix.service}} --debug
+
+      - name: Run ${{matrix.service}} Shared Tables Tests
+        run: _APP_DATABASE_SHARED_TABLES=database_db_main docker compose exec -T appwrite test /usr/src/code/tests/e2e/Services/${{matrix.service}} --debug
```
christyjacob4, line 150: Can we document this somewhere?

christyjacob4: As in, the syntax/ format of this variable

app/controllers/api/projects.php
christyjacob4: Yeap agreed. We needed this code to prevent DDL operations being performed on a DB undergoing backup. Its not required anymore

christyjacob4 (approved):

tests/e2e/Services/Projects/ProjectsConsoleClientTest.php
abnegate: Instead of the few single tests, we now have all tests running for both shared and non-shared table modes

.github/workflows/tests.yml
abnegate: Yep, will make more sense in cloud

PR #8207, opened 2024-05-27
Feat: users.createJWT()

Meldiron:
## What does this PR do?

Adds method to generate JWT for an user. Useful for some server-side apps, as well as preparation for new CLI features 🙊

## Test Plan

- [x] Implement tests

## Related PRs and Issues

x

## Checklist


Meldiron:
Force-push intentional, by mistake I merged main instead of 1.6.x to sync.

app/controllers/api/account.php
```diff
@@ -2369,15 +2370,13 @@
...
 
         $response
             ->setStatusCode(Response::STATUS_CODE_CREATED)
             ->dynamic(new Document(['jwt' => $jwt->encode([
-                // 'uid'    => 1,
-                // 'aud'    => 'http://site.com',
-                // 'scopes' => ['user'],
-                // 'iss'    => 'http://api.mysite.com',
+                'iat' => \time(),
+                'exp' => \intval((new \DateTime())->add(new \DateInterval('PT900S'))->format('U')),
```
stnguyen90, line 2379: the maxAge param when initializing the JWT is used for the exp. Why do we need to manually add these things in?

app/controllers/api/users.php
```diff
@@ -2091,6 +2093,61 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+        $session = new Document();
+
+        if($sessionId === 'current') {
+            // Get most recent
+            $session = \count($sessions) > 0 ? $sessions[\count($sessions) - 1] : new Document();
+        } else {
+            // Find by ID
+            foreach ($sessions as $loopSession) { /** @var Utopia\Database\Document $loopSession */
+                if ($loopSession->getId() == $sessionId) {
+                    $session = $loopSession;
```
stnguyen90, line 2130: It's a good practice to break to prevent continuing on with the rest of the array:

```suggestion
                    $session = $loopSession;
                    break;
```

app/controllers/api/account.php
```diff
@@ -2369,15 +2370,13 @@
             throw new Exception(Exception::USER_SESSION_NOT_FOUND);
         }
 
-        $jwt = new JWT(System::getEnv('_APP_OPENSSL_KEY_V1'), 'HS256', 900, 10); // Instantiate with key, algo, maxAge and leeway.
```
stnguyen90, line 2372: why are we removing the leeway?

app/controllers/api/users.php
```diff
@@ -2091,6 +2093,61 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
+    ->label('scope', 'users.write')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'users')
+    ->label('sdk.method', 'createJWT')
+    ->label('sdk.description', '/docs/references/users/create-user-jwt.md')
+    ->label('sdk.response.code', Response::STATUS_CODE_CREATED)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_JWT)
+    ->param('userId', '', new UID(), 'User ID.')
+    ->param('sessionId', 'current', new UID(), 'Session ID. Use the string \'current\' to use the most recent session. Defaults to the most recent session.', true)
```
stnguyen90, line 2108: "recent" can be recently used or recently created. Since the code looks for recently created, we should make that clear here.

docs/references/users/create-user-jwt.md
```diff
@@ -0,0 +1 @@
+Use this endpoint to create a JSON Web Token for user by its unique ID. You can use the resulting JWT to authenticate on behalf of the user. The JWT secret will become invalid if the session it uses gets deleted.
```
stnguyen90, line 1: > The JWT secret will become invalid if the session it uses gets deleted.

We actually have a bug with this: 
* https://github.com/appwrite/appwrite/issues/8000

But hopefully it will be fixed soon.

tests/e2e/Services/Users/UsersBase.php
```diff
@@ -1553,6 +1553,137 @@ public function testDeleteUser(array $data): array
         return $data;
     }
 
+    public function testUsetJWT()
```
stnguyen90, line 1556: Is the function name supposed to be:

```suggestion
    public function testUserJWT()
```

app/controllers/api/account.php
Meldiron: We now support 1-3600 seconds JWT keys. If developer decides to have JWT work for 15 seconds, it shouldn't be 25.

app/controllers/api/account.php
Meldiron: Used maxAge instead

app/controllers/api/users.php
Meldiron: Renamed to "recent"

docs/references/users/create-user-jwt.md
Meldiron: Good to know. This is going to 1.6.x, We have enough time to have that fix implement before 1.6.x

eldadfux (approved):

PR #8214, opened 2024-05-29
Remove build id from api endpoint

vermakhushboo:
## What does this PR do?

Remove `buildId` from API endpoint and fetch it from deployments document.

## Test Plan

Tested manually.

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/functions.php
```diff
@@ -1448,7 +1449,7 @@
     ->label('sdk.response.model', Response::MODEL_NONE)
     ->param('functionId', '', new UID(), 'Function ID.')
     ->param('deploymentId', '', new UID(), 'Deployment ID.')
-    ->param('buildId', '', new UID(), 'Build unique ID.')
+    ->param('buildId', '', new UID(), 'Build unique ID.', true)
```
Meldiron, line 1452: I would remove it all-together

Meldiron: Discussed, lets add comment about why its here

app/controllers/api/functions.php
```diff
@@ -1433,7 +1433,8 @@
         $response->noContent();
     });
 
-App::post('/v1/functions/:functionId/deployments/:deploymentId/builds/:buildId')
+App::post('/v1/functions/:functionId/deployments/:deploymentId/build')
+    ->alias('/v1/functions/:functionId/deployments/:deploymentId/builds/:buildId')
     ->groups(['api', 'functions'])
     ->desc('Create build')
```
Meldiron, line 1439:
```suggestion
    ->desc('Rebuild deployment')
```

app/controllers/api/functions.php
```diff
@@ -1469,6 +1470,9 @@
             throw new Exception(Exception::DEPLOYMENT_NOT_FOUND);
         }
 
+        // buildId is added as param just for backward compatibility
+        // It is fetched from the deployments collection and not from the request as the user doesn't have to know the buildId
+        $buildId = $deployment->getAttribute('buildId', '');
         $build = Authorization::skip(fn () => $dbForProject->getDocument('builds', $buildId));
 
         if ($build->isEmpty()) {
```
christyjacob4, line 1478: Why do we need this block of code? If you look at the code below, we don't use this build's ID for anything

app/controllers/api/functions.php
vermakhushboo: @Meldiron what should be the sdk method called?

app/controllers/api/functions.php
vermakhushboo: Removed the code as it wasn't being used anywhere

Meldiron (approved):

app/controllers/api/functions.php
```diff
@@ -1448,15 +1449,14 @@
     ->label('sdk.response.model', Response::MODEL_NONE)
     ->param('functionId', '', new UID(), 'Function ID.')
     ->param('deploymentId', '', new UID(), 'Deployment ID.')
-    ->param('buildId', '', new UID(), 'Build unique ID.')
+    ->param('buildId', '', new UID(), 'Build unique ID.', true) // added as optional param for backward compatibility
```
stnguyen90, line 1452: What backwards compatibility are we worried about?

This is scoped for 1.6 which means we're going to be bumping the SDKs major version so breaking changes for that should be fine.

christyjacob4: The endpoint should continue to work for people using older SDKs as well

christyjacob4 (approved):

PR #8215, opened 2024-05-29
feat: Adding cli to the function deployment type

byawitz:
## What does this PR do?

Set the function deployment type as `cli` when `x-sdk-language` matches the `cli`. This will make it possible to filter function deployments by CLI.

## Test Plan

Tested against local `1.5.5` instance

## Related

* https://github.com/appwrite/console/pull/1112

## Checklist


PineappleIOnic (approved):

app/controllers/api/functions.php
```diff
@@ -1163,6 +1163,7 @@
         }
 
         $activate = (bool) filter_var($activate, FILTER_VALIDATE_BOOLEAN);
+        $type = $request->getHeader('x-sdk-language') === 'cli' ? 'cli' : 'manual';
```
stnguyen90, line 1166: Any chance we can update a test case to check for this?

byawitz: Done

byawitz: The failing test is not relevant

christyjacob4 (approved):

PR #8217, opened 2024-05-29
Feat Sync team deletion

### LINKED ISSUE #7939: 🐛 Bug Report: Membership Deletion Queue Holds Account Deletion
### 👟 Reproduction steps

Discord thread [here](https://discord.com/channels/564160730845151244/636852860709240842/1226918113241600082).

---
Deleted all organizations & then proceeded to delete the account on Cloud, however this [check](https://github.com/appwrite/appwrite/blob/61f04d5340f0fd9afd11e09f7f3ef53278e9882a/app/controllers/api/account.php#L360) returned an error saying there was an active membership attached to the account.

### 👍 Expected behavior

Account should be deleted.

### 👎 Actual Behavior

Account wasn't deleted with the above mentioned error.

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

Linux

### 🧱 Your Environment

_No response_

byawitz:
## What does this PR do?

Changed team and organization deleting logic to the API instead of queuing it to the deleting worker

## Test Plan

Tested against local `1.5.5` instance, 
1. All of the projects were pulled using the CLI and pushed into a new Organization with different project IDs.
2. A few resources and files were added manually 
3. Delete organization, delete everything.

## Related PRs and Issues

Close #7939 

## Checklist


christyjacob4 (approved):

Meldiron (commented):
Looks good. Now that it's sync, let's update tests for sync behaviour. We either don't have test for this, or it's async (with sleep)

byawitz:
@Meldiron, checked, the test is without `sleep` 

https://github.com/appwrite/appwrite/blob/main/tests/e2e/Services/Teams/TeamsBase.php#L371-L388

src/Appwrite/Platform/Workers/Deletes.php
```diff
@@ -1125,4 +1129,16 @@ function (Document $function) use ($deleteByFunction) {
             );
         }
     }
+
+    /**
+     * @throws Exception
+     * @throws Throwable
+     */
+    public function deleteTeams($dbForConsole, $getProjectDB, $team, $project): void
```
stnguyen90, line 1137: Please add typing for these params. Also, this should be singular (`deleteTeam` instead of `deleteTeams`), right?

stnguyen90 (changes requested):

stnguyen90:
> @Meldiron, checked, the test is without `sleep`
> 
> https://github.com/appwrite/appwrite/blob/main/tests/e2e/Services/Teams/TeamsBase.php#L371-L388

@byawitz, but that tests doesn't verify the memberships are deleted synchronously. Maybe you can add a users list memberships? Or at least manually test?

byawitz:
The test returns 404, which means the teams were deleted, and there's no way to delete them in an async way. 

What should the test be different?

stnguyen90:
> The test returns 404, which means the teams were deleted, and there's no way to delete them in an async way.

Even without your change, the current test returns 404 so the current test doesn't actually test anything about memberships? 

> What should the test be different?

A test that verifies the memberships are deleted would be great. I'm not totally sure on how to test, though. Maybe [users.listMemberships()](https://appwrite.io/docs/references/cloud/server-nodejs/users#listMemberships)?

byawitz:
The tests check if the team for a given ID exists. When the endpoint returns `404`, it means the team no longer exists.

Do you mean to add a whole new test for `userMemberships`?

src/Appwrite/Platform/Workers/Deletes.php
byawitz: The method can delete more than one team

stnguyen90:
> Do you mean to add a whole new test for userMemberships

@byawitz, something like that. The goal is to test and make sure the memberships are deleted when the team is deleted.

stnguyen90 (commented):
> Do you mean to add a whole new test for userMemberships

@byawitz, something like that. The goal is to test and make sure the memberships are deleted when the team is deleted.

byawitz:
Test added [here](https://github.com/appwrite/appwrite/pull/8217/files#diff-1d6c1e5a3ea8933facab76e4532515a7ff6ed2d9eb896349cbf3e19c4f4cd65a)

src/Appwrite/Platform/Workers/Deletes.php
```diff
@@ -1158,4 +1162,24 @@ function (Document $function) use ($deleteByFunction) {
...
+     * @throws DatabaseException
+     * @throws Exception
+     * @throws Restricted
+     * @throws Structure
+     */
+    public function deleteTeam(Database $dbForConsole, callable $getProjectDB, Document $team, Document $project): void
+    {
+        $this->deleteMemberships($getProjectDB, $team, $project);
+        if ($project->getId() === 'console') {
+            $this->deleteProjectsByTeam($dbForConsole, $getProjectDB, $team);
```
stnguyen90, line 1182: Deleting projects synchronously during the API call can take some time, right? Can we avoid that?

app/controllers/api/teams.php
```diff
@@ -338,10 +339,13 @@
     ->label('sdk.response.model', Response::MODEL_NONE)
     ->param('teamId', '', new UID(), 'Team ID.')
     ->inject('response')
+    ->inject('getProjectDB')
     ->inject('dbForProject')
-    ->inject('queueForEvents')
+    ->inject('dbForConsole')
```
stnguyen90, line 344: `dbForConsole` isn't needed anymore, right?

docker-compose.yml
```diff
@@ -76,6 +76,7 @@ services:
       - appwrite-config:/storage/config:rw
       - appwrite-certificates:/storage/certificates:rw
       - appwrite-functions:/storage/functions:rw
+      - appwrite-builds:/storage/builds:rw
```
stnguyen90, line 79: Is this change needed?

app/views/install/compose.phtml
```diff
@@ -65,6 +65,7 @@ $image = $this->getParam('image', '');
       - appwrite-config:/storage/config:rw
       - appwrite-certificates:/storage/certificates:rw
       - appwrite-functions:/storage/functions:rw
+      - appwrite-builds:/storage/builds:rw
```
stnguyen90, line 68: Is this change needed?

byawitz: Not any more

docker-compose.yml
byawitz: Not any more

stnguyen90 (approved):

christyjacob4 (approved):

PR #8222, opened 2024-05-30
feat: support twilio messaging service sid

loks0n:
Allows `_APP_SMS_FROM` to be set to a Messaging Service Sid when using Twilio as SMS provider. See https://www.twilio.com/docs/messaging/services

### DIFF
--- a/src/Appwrite/Platform/Workers/Messaging.php
+++ b/src/Appwrite/Platform/Workers/Messaging.php
@@ -400,7 +400,9 @@ class Messaging extends Action
                 'twilio' => [
                     'accountSid' => $user,
                     'authToken' => $password,
-                    'messagingServiceSid' => $smsDSN->getParam('messagingServiceSid') ?? null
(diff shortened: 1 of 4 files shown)


src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -465,7 +473,7 @@ private function getSmsAdapter(Document $provider): ?SMSAdapter
 
         return match ($provider->getAttribute('provider')) {
             'mock' => new Mock('username', 'password'),
-            'twilio' => new Twilio($credentials['accountSid'], $credentials['authToken']),
+            'twilio' => new Twilio($credentials['accountSid'], $credentials['authToken'], null, isset($credentials['messagingServiceSid']) ? $credentials['messagingServiceSid'] : null),
```
christyjacob4, line 476: So `from` here is always `null` ?

christyjacob4 (approved):

PR #8223, opened 2024-05-31
Add growth events hook for cloud

PineappleIOnic:
## What does this PR do?

This PR adds the growth hooks for cloud

## Checklist


app/init.php
```diff
@@ -1071,6 +1072,9 @@ function (mixed $value) {
 App::setResource('queueForMigrations', function (Connection $queue) {
     return new Migration($queue);
 }, ['queue']);
+App::setResource('queueForGrowth', function (Connection $queue) {
+    return new Growth($queue);
```
christyjacob4, line 1076:
```suggestion
    return null;
```

PR #8224, opened 2024-05-31
Respect `activate` value for Functions Deployments

ItzNotABug:
## What does this PR do?

This PR fixes the issue where the if `activate` is false when deploying a function, it is still activated regardless.

## Test Plan

Manual, see -

https://github.com/appwrite/appwrite/assets/20625965/26e4abb6-6013-4059-b7a4-49edcca20500

## Related PRs and Issues

- NA

## Checklist


app/controllers/api/functions.php
```diff
@@ -1050,7 +1050,7 @@
     ->param('entrypoint', null, new Text(1028), 'Entrypoint File.', true)
     ->param('commands', null, new Text(8192, 0), 'Build Commands.', true)
     ->param('code', [], new File(), 'Gzip file with your code package. When used with the Appwrite CLI, pass the path to your code directory, and the CLI will automatically package your code. Use a path that is within the current directory.', skipValidation: true)
-    ->param('activate', false, new Boolean(true), 'Automatically activate the deployment when it is finished building.')
+    ->param('activate', false, new Boolean(true), 'Activate the deployment when it is finished building.', skipValidation: true)
```
Meldiron, line 1053: Is `skipValidation` nessessary addition? Does it not work without it - what's the error?

ItzNotABug: It doesn't actually. The issue is `('sdk.request.type', 'multipart/form-data')` actually which transforms the `false` value to an empty string (`''`) while the truthy ones are properly converted to `1`. This triggers a validation failure from the `new Boolean(true)`.

ItzNotABug:
The latest commit 269cdc3 properly uses the correct status-code (the only failing test), a re-run of the tests might not be necessary considering all of them run for quite a while.

ItzNotABug:
Closing in favour of - https://github.com/appwrite/appwrite/pull/8358.

PR #8232, opened 2024-06-04
Updates in multi region sync

shimonewman:


app/controllers/api/edge.php
```diff
@@ -0,0 +1,63 @@
...
+        $token = str_replace(["Bearer"," "], "", $token);
+        $jwt = new JWT(System::getEnv('_APP_OPENSSL_KEY_V1'), 'HS256', 600, 10);
+        try {
+            $payload = $jwt->decode($token);
+        } catch (JWTException $error) {
+            throw new Exception(Exception::USER_JWT_INVALID, 'Failed to verify JWT. ' . $error->getMessage());
+        }
+    });
+
+App::post('/v1/edge/sync')
```
eldadfux, line 31: Maybe edge is not the best term here. Lets try `/v1/region` for region metadata (name, etc), and `/v1/region/syncs` unless we find a better name for reporting all the keys to sync.

bin/worker-sync-out-delivery
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+php /usr/src/code/app/worker.php sync-out-delivery $@
```
eldadfux, line 3: Not sure about the name delivery, maybe we can just drop it.

bin/edge-sync
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+php /usr/src/code/app/cli.php edge-sync $@
```
eldadfux, line 3: What is this task for? let's avoid the term edge as it doesn't 100% fit

bla.tar.gz
eldadfux, line 1: leftover

src/Appwrite/Platform/Tasks/EdgeSync.php
```diff
@@ -0,0 +1,85 @@
...
+{
+    public static function getName(): string
+    {
+        return 'edge-sync';
+    }
+
+    public function __construct()
+    {
+        $this
+            ->desc('Schedules edge sync tasks')
```
eldadfux, line 25: Why do we use a scheduler? I feel this will not be good enough for production. We need to strive to keep sync latency below 300ms at worst. This intervals might not cut it. I thought we will have a long running tasks that constantly fetches logs and send them async to all regions at the same time and has some retry logic.

eldadfux (changes requested):

bin/edge-sync
shimonewman: This task loops over payloads to delete the ones that were sent successfully and re-queue the failed ones.
Renamed to region

app/controllers/api/edge.php
shimonewman: renamed to region

PR #8236, opened 2024-06-05
Fix: Don't set the target field if the existing target document is false

2002Bishwajeet:
If you create a user, set the target then delete the user and the target from console and then create a user again, you get a type error in the User Model
image

This happens cause it tries to populate the target field but the response was false

Adds a check if the document exists then set the target

## Test Plan

WIP

## Related PRs and Issues

https://discord.com/channels/564160730845151244/1062789941727338577/1247967828015972362

## Checklist

### DIFF
--- a/app/controllers/api/users.php
+++ b/app/controllers/api/users.php
@@ -138,7 +138,9 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
                 $existingTarget = $dbForProject->findOne('targets', [
                     Query::equal('identifier', [$email]),
                 ]);
-                $user->setAttribute('targets', [...$user->getAttribute('targets', []), $existingTarget]);
+                if($existingTarget) {
+                    $user->setAttribute('targets', [...$user->getAttribute('targets', []), $existingTarget]);
+                }
             }
         }
 
@@ -160,7 +162,9 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
                 $existingTarget = $dbForProject->findOne('targets', [
                     Query::equal('identifier', [$phone]),
                 ]);
-                $user->setAttribute('targets', [...$user->getAttribute('targets', []), $existingTarget]);
+                if($existingTarget) {
+                    $user->setAttribute('targets', [...$user->getAttribute('targets', []), $existingTarget]);
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


tests/e2e/Services/Account/AccountCustomClientTest.php
```diff
@@ -327,7 +327,6 @@ public function testGetAccountLogs($data): array
         $this->assertEquals('desktop', $response['body']['logs'][2]['deviceName']);
         $this->assertEquals('', $response['body']['logs'][2]['deviceBrand']);
         $this->assertEquals('', $response['body']['logs'][2]['deviceModel']);
-        $this->assertEquals($response['body']['logs'][2]['ip'], filter_var($response['body']['logs'][2]['ip'], FILTER_VALIDATE_IP));
```
stnguyen90, line 330: How come this was removed?

2002Bishwajeet: We already do the assertions check [here](https://github.com/appwrite/appwrite/pull/8236/files/485b8eeb8b2461f46a877cbc017baeaf487a9c33#diff-434428e8237205f1c4c6540329c60816d1e243383b6a7b044b844a594f4a35edL296) . Or its different? 👀

2002Bishwajeet:
I will add the unit tests for this PR by today. Had some busy days🙏

app/controllers/api/users.php
```diff
@@ -138,7 +138,9 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
                 $existingTarget = $dbForProject->findOne('targets', [
                     Query::equal('identifier', [$email]),
                 ]);
-                $user->setAttribute('targets', [...$user->getAttribute('targets', []), $existingTarget]);
+                if($existingTarget) {
+                    $user->setAttribute('targets', [...$user->getAttribute('targets', []), $existingTarget]);
```
stnguyen90, line 142: `setAttribute()` has support for appending. See the [function signature](https://github.com/utopia-php/database/blob/45def2f7c6bc5f631dbb67e5df0e8e7331af5f63/src/Database/Document.php#L215). and example:

https://github.com/appwrite/appwrite/blob/0376189dbcb791d6ea28fd69332d7df1b208f1c3/app/init.php#L1075-L1080

stnguyen90 (changes requested):

app/controllers/api/users.php
2002Bishwajeet: Done

2002Bishwajeet:
I don't think a unit test would be needed (needs too much repetitive code create user, target, delete user, target then create user to check again).

stnguyen90 (approved):
Theoretically, the existing tests should confirm everything still works as intended.

Issue #8159, opened 2024-05-19
🐛 Bug Report: Expired targets not automatically deleted


ayushpathak-48:
### 👟 Reproduction steps

Go to console 

Create  FCM provider for push messaging 

Create a token of a device and save it into appwrite

Then re create the token and again save it to appwrite

Then send notification to the web target 

Then you will get an error for the expired device token

Again after 1 day th same error occurs.

But appwrite says it should be removed after a days if it's expired

### 👍 Expected behavior

It should be removed

### 👎 Actual Behavior

It's giving same errror everyday

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

Windows

### 🧱 Your Environment

I am using node-appwrite to create push target and for creating push notification 

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

stnguyen90:
Related thread: https://discord.com/channels/564160730845151244/1241034854922063954

ItzNotABug:
@stnguyen90 is this issue open for assignment?

stnguyen90:
@ItzNotABug, sure, ya. Assigned!

PR #8239, opened 2024-06-06
Delete Expired Targets Per Project

ItzNotABug:
## What does this PR do?

Deletes the expired messaging targets per project.

Fixes https://github.com/appwrite/appwrite/issues/8159

## Test Plan

N/A.

## Related PRs and Issues

- #8159

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Tasks/Maintenance.php
+++ b/src/Appwrite/Platform/Tasks/Maintenance.php
@@ -49,6 +49,7 @@ class Maintenance extends Action
             $this->foreachProject($dbForConsole, function (Document $project) use ($queueForDeletes, $usageStatsRetentionHourly) {
                 $queueForDeletes->setProject($project);
 
+                $this->notifyDeleteTargets($queueForDeletes);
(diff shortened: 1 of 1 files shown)


stnguyen90 (commented):
Nice work! Would you be able to provide some screenshot of maybe logs or something to verify this does fix the problem?

ItzNotABug:
@stnguyen90,
Before (expired and active targets) -

Screenshot 2024-06-09 at 7 34 20 PM

&nbsp;
Ran - `docker exec appwrite maintenance`

After -

Screenshot 2024-06-09 at 7 37 06 PM

stnguyen90:
> Update - I tried with the current logic as well, that seemed to remove the expired targets even when not run per project.

@ItzNotABug you mean before your change? It could be because the project is set after the loop and so the delete expired works, but only for that last project.

ItzNotABug:
@stnguyen90 Had the same idea after I realized that there was only one project added. Thanks for the confirmation!

stnguyen90 (approved):

PR #8243, opened 2024-06-07
Feat: Scheduled executions (execute later)

Meldiron:
## What does this PR do?

Adds parameter  to `createExecution` to allow execution in the future, similarly to how messages can also be scheduled for later.

## Test Plan

- [x] Implement tests

## Related PRs and Issues

x

## Checklist


tests/e2e/Services/Functions/FunctionsCustomClientTest.php
```diff
@@ -175,6 +175,100 @@ public function testCreateExecution(): array
...
+        $execution = $this->client->call(Client::METHOD_POST, '/functions/' . $function['body']['$id'] . '/executions', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()), [
+            'async' => true,
+            'scheduledAt' => $futureTime,
+        ]);
+
+        $this->assertEquals(202, $execution['headers']['status-code']);
+        $this->assertEquals('scheduled', $execution['body']['status'], \json_encode($execution['body']));
```
Meldiron, line 259: Let's sleep here for 1.5 mins, (if future time is 1min), and let's check it was completed after that.

src/Appwrite/Platform/Tasks/ScheduleFunctions.php
```diff
@@ -40,32 +40,39 @@ protected function enqueueResources(Group $pools, Database $dbForConsole): void
...
                 continue;
             }
 
-            $total++;
-
-            $promiseStart = \time(); // in seconds
-            $executionStart = $nextDate->getTimestamp(); // in seconds
-            $delay = $executionStart - $promiseStart; // Time to wait from now until execution needs to be queued
+            $total += 1;
+            $delay = $nextDate->getTimestamp() - \time(); // Time to wait from now until execution needs to be queued
```
Meldiron, line 65: In messaigng we do something a bit different:
https://github.com/appwrite/appwrite/blob/feat-scheduled-executions/src/Appwrite/Platform/Tasks/ScheduleMessages.php#L34-L36

Sould we do the same? I think this method runs every minute.. I am worried that with this logic, if you schedule function in 5 mins, it will run 5 times. Maybe.
Also scarred if worker is down, and it gets negative number here, god knows what happens.

I think it's the best bet to copy 99% of code from messaging - messaging scheduling works great.

app/controllers/api/functions.php
```diff
@@ -1511,16 +1512,21 @@
...
     ->inject('queueForEvents')
     ->inject('queueForUsage')
-    ->inject('mode')
     ->inject('queueForFunctions')
     ->inject('geodb')
-    ->action(function (string $functionId, string $body, bool $async, string $path, string $method, array $headers, Response $response, Document $project, Database $dbForProject, Document $user, Event $queueForEvents, Usage $queueForUsage, string $mode, Func $queueForFunctions, Reader $geodb) {
+    ->action(function (string $functionId, string $body, bool $async, string $path, string $method, array $headers, ?string $scheduledAt, Response $response, Document $project, Database $dbForProject, Database $dbForConsole, Document $user, Event $queueForEvents, Usage $queueForUsage, Func $queueForFunctions, Reader $geodb) {
+
+        if(!$async && !is_null($scheduledAt)) {
+            throw new Exception(Exception::GENERAL_QUERY_INVALID, 'Scheduled executions must run asynchronously. Don\'t set scheduledAt to execute immediately, or set async to true.');
```
christyjacob4, line 1528: Might be more relevant to use the BAD_REQUEST error here

src/Appwrite/Platform/Tasks/ScheduleFunctions.php
```diff
@@ -40,32 +40,39 @@ protected function enqueueResources(Group $pools, Database $dbForConsole): void
...
 
             if (!isset($delayedExecutions[$delay])) {
                 $delayedExecutions[$delay] = [];
             }
 
-            $delayedExecutions[$delay][] = $key;
+            $delayedExecutions[$delay][] = $scheduleKey;
```
christyjacob4, line 71: Where is $scheduleKey defined? I cant seem to find it

tests/e2e/Services/Functions/FunctionsCustomClientTest.php
christyjacob4: why 1.5 minutes? Cant we schedule something sooner? Like 15-20 seconds? Tests are already slow so lets try not to introduce more delays

Meldiron: I think there is 10s sync interval, and 60s loop interval in scheduler.
Good catch, this makes function scheduled in 15 seconds from now only execute after a minute, I think. Let's check. Lets understand how quickly it can execute.

src/Appwrite/Platform/Tasks/ScheduleFunctions.php
loks0n: In the foreach above

app/controllers/api/functions.php
loks0n: done

src/Appwrite/Event/Func.php
```diff
@@ -200,6 +223,7 @@ public function trigger(): string|bool
             'project' => $this->project,
             'user' => $this->user,
             'function' => $this->function,
+            'functionId' => $this->functionId,
```
Meldiron, line 226: There is already $function, why cant we get function ID as `function->getId()`?

src/Appwrite/Platform/Workers/Functions.php
```diff
@@ -83,6 +83,7 @@ public function action(Message $message, Database $dbForProject, Func $queueForF
         $eventData = $payload['payload'] ?? '';
         $project = new Document($payload['project'] ?? []);
         $function = new Document($payload['function'] ?? []);
+        $functionId = $payload['functionId'] ?? '';
```
Meldiron, line 86: Related to previous comment, we can `$function->getId()` I think

src/Appwrite/Platform/Workers/Functions.php
```diff
@@ -92,6 +93,10 @@ public function action(Message $message, Database $dbForProject, Func $queueForF
             return;
         }
 
+        if ($function->isEmpty() && !empty($functionId)) {
+            $function = $dbForProject->getDocument('functions', $functionId);
```
Meldiron, line 97: We could do `getDocument` when we added `setFunctionId()` and do `setFunction` instead - preventing need for this

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
```diff
@@ -0,0 +1,66 @@
...
+
+use Appwrite\Event\Func;
+use Cron\CronExpression;
+use Utopia\Database\Database;
+use Utopia\Pools\Group;
+
+class ScheduleExecutions extends ScheduleBase
+{
+    public const UPDATE_TIMER = 3; // seconds
+    public const ENQUEUE_TIMER = 4; // seconds
```
Meldiron, line 13: Those numbers look very random. What about 5 and 5? 3 seems a bit too fast, considering we might have many of those worker, causing bunch of SQL queries just from here.

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
```diff
@@ -0,0 +1,66 @@
...
+
+    public static function getSupportedResource(): string
+    {
+        return 'execution';
+    }
+
+    protected function enqueueResources(Group $pools, Database $dbForConsole): void
+    {
+        foreach ($this->schedules as $schedule) {
+            if (!$schedule['active'] || CronExpression::isValidExpression($schedule['schedule'])) {
```
Meldiron, line 28: If I understand it right, we dont care about schedule cron here. Also I think it would never be valid cron in this worker - its datetime

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
```diff
@@ -0,0 +1,66 @@
...
+            }
+
+            $now = new \DateTime();
+            $scheduledAt = new \DateTime($schedule['schedule']);
+
+            if ($scheduledAt > $now) {
+                continue;
+            }
+
+            \go(function () use ($schedule, $pools, $dbForConsole) {
```
Meldiron, line 40: I think we don't need to do this in `go`. We do it for CRONs because we delay it some more with sleep. But here it's instant execution - we can do it without go.

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
```diff
@@ -0,0 +1,66 @@
...
+                $queueForFunctions = new Func($connection);
+
+                $queueForFunctions
+                    ->setType('schedule')
+                    ->setFunctionId($schedule['resource']['functionId'])
+                    ->setExecution($schedule['resource'])
+                    ->setMethod('POST')
+                    ->setPath('/')
+                    ->setProject($schedule['project'])
+                    ->trigger();
```
Meldiron, line 53: setType, setMethod and setPath doesnt sound right. Method and path should be coming from the execution itself - this worker isn't setting it.
setType might in theory stay, but behaviour is closer to HTTP execution, compared to scheduled one.

src/Appwrite/Event/Func.php
loks0n: The `resource` of the scheduler is an `execution`, not a `function`. We aren't able to pass the whole `function` resource as we only have the linked `functionId` from the `execution`.

src/Appwrite/Platform/Workers/Functions.php
loks0n: We don't have `dbForProject` injected the scheduler, only `dbForConsole`.

src/Appwrite/Platform/Workers/Functions.php
loks0n: Please see above.

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
loks0n: These numbers are copied from messaging scheduling, which works well. I don't think `5` and `3` will offer any difference?

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
loks0n: Yes - removed.

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
loks0n: OK - removed.

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
```diff
@@ -0,0 +1,66 @@
...
+            $now = new \DateTime();
+            $scheduledAt = new \DateTime($schedule['schedule']);
+
+            if ($scheduledAt > $now) {
+                continue;
+            }
+
+            \go(function () use ($schedule, $pools, $dbForConsole) {
+                $queue = $pools->get('queue')->pop();
+                $connection = $queue->getResource();
```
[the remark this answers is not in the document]
loks0n: OK - fixed.

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
loks0n: Removed:
```
                    ->setMethod('POST')
                    ->setPath('/')
```

Do you have a suggestion for the alternative to `->setType('schedule')`?

app/controllers/api/functions.php
```diff
@@ -1591,16 +1592,21 @@
...
     ->inject('queueForEvents')
     ->inject('queueForUsage')
-    ->inject('mode')
     ->inject('queueForFunctions')
     ->inject('geodb')
-    ->action(function (string $functionId, string $body, bool $async, string $path, string $method, array $headers, Response $response, Document $project, Database $dbForProject, Document $user, Event $queueForEvents, Usage $queueForUsage, string $mode, Func $queueForFunctions, Reader $geodb) {
+    ->action(function (string $functionId, string $body, bool $async, string $path, string $method, array $headers, ?string $scheduledAt, Response $response, Document $project, Database $dbForProject, Database $dbForConsole, Document $user, Event $queueForEvents, Usage $queueForUsage, Func $queueForFunctions, Reader $geodb) {
+
+        if(!$async && !is_null($scheduledAt)) {
+            throw new Exception(Exception::GENERAL_BAD_REQUEST, 'Scheduled executions must run asynchronously. Set scheduledAt to a future date, or set async to true.');
```
Meldiron, line 1608: Let's add test that expects this exception

app/controllers/api/functions.php
```diff
@@ -1705,15 +1711,21 @@
...
         $execution = new Document([
             '$id' => $executionId,
             '$permissions' => !$user->isEmpty() ? [Permission::read(Role::user($user->getId()))] : [],
             'functionInternalId' => $function->getInternalId(),
             'functionId' => $function->getId(),
             'deploymentInternalId' => $deployment->getInternalId(),
             'deploymentId' => $deployment->getId(),
-            'trigger' => 'http', // http / schedule / event
-            'status' => $async ? 'waiting' : 'processing', // waiting / processing / completed / failed
+            'trigger' => (!is_null($scheduledAt)) ? 'schedule' : 'http',
```
Meldiron, line 1727: Depending on what Eldad says, we might always want HTTP here.

app/controllers/api/functions.php
```diff
@@ -1736,20 +1748,33 @@
...
+            } else {
+                $dbForConsole->createDocument('schedules', new Document([
+                    'region' => System::getEnv('_APP_REGION', 'default'),
+                    'resourceType' => 'execution',
+                    'resourceId' => $execution->getId(),
+                    'resourceInternalId' => $execution->getInternalId(),
+                    'resourceUpdatedAt' => DateTime::now(),
+                    'projectId' => $project->getId(),
+                    'schedule' => $scheduledAt,
+                    'active' => true,
```
Meldiron, line 1775: Compared to event triggered above, we are losing details here, like: body, headers, path, method, user.
JWT we shouldn't set here (it might expire), functions worker should generate it if we pass user.

I think we should add new parameter to `schedules` collection to store additional data like those, and just forward the mto function with execution.

Meldiron: Once fixed, lets update test to ensure this. For example the function could receive body and log it - when we do getExecution, we can check it's logs and see if it includes the word.

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
```diff
@@ -0,0 +1,60 @@
...
+            $now = new \DateTime();
+            $scheduledAt = new \DateTime($schedule['schedule']);
+
+            if ($scheduledAt > $now) {
+                continue;
+            }
+
+            $queueForFunctions
+                ->setType('schedule')
+                ->setFunctionId($schedule['resource']['functionId'])
```
Meldiron, line 45: Let's add a comment here, that we only set ID because we don't have dbForProject at this point. Worker will later get the whole function.

Meldiron: I would like to use dbForPRoject here, but this shouldn' be the reason to introduce it. When we have a comment, in future, when we add dbForProject for other reasons, we will improve this too.

src/Appwrite/Platform/Tasks/ScheduleFunctions.php
```diff
@@ -84,8 +84,6 @@ protected function enqueueResources(Group $pools, Database $dbForConsole): void
                     $queueForFunctions
                         ->setType('schedule')
                         ->setFunction($schedule['resource'])
-                        ->setMethod('POST')
-                        ->setPath('/')
```
Meldiron, line 88: Scheduled functions should still have those. That is coming from CRON executions, when they can't set those. Those needs to remain.

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
```diff
@@ -0,0 +1,60 @@
...
+                ->setExecution($schedule['resource'])
+                ->setProject($schedule['project'])
+                ->trigger();
```
Meldiron, line 48: As mention in API, here we need to also set stuff like body, headers, method, path, user jwt...
To get those data here, we could add attribute to schedules as temporary storage for those info (since we delete the document right after)

src/Appwrite/Event/Func.php
```diff
@@ -11,9 +11,10 @@ class Func extends Event
     protected string $jwt = '';
     protected string $type = '';
     protected string $body = '';
-    protected string $path = '';
-    protected string $method = '';
+    protected ?string $path = null;
+    protected ?string $method = null;
```
Meldiron, line 15: Do we need this change?

tests/e2e/Services/Functions/FunctionsCustomClientTest.php
```diff
@@ -59,6 +59,7 @@ public function testCreateExecution(): array
             'execute' => [Role::user($this->getUser()['$id'])->toString()],
             'runtime' => 'php-8.0',
             'entrypoint' => 'index.php',
+            'logging' => true,
```
Meldiron, line 62: Do we need this change?

tests/e2e/Services/Functions/FunctionsCustomClientTest.php
```diff
@@ -195,6 +196,113 @@ public function testCreateExecution(): array
...
+                $deployment['headers']['status-code'] >= 400
+                || \in_array($deployment['body']['status'], ['ready', 'failed'])
+            ) {
+                break;
+            }
+
+            \sleep(1);
+        }
+
+        $this->assertEquals('ready', $deployment['body']['status'], \json_encode($deployment['body']));
```
Meldiron, line 259: Not sure if 3rd parameter here is expected

tests/e2e/Services/Functions/FunctionsCustomClientTest.php
```diff
@@ -195,6 +196,113 @@ public function testCreateExecution(): array
...
+        $execution = $this->client->call(Client::METHOD_POST, '/functions/' . $function['body']['$id'] . '/executions', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()), [
+            'async' => true,
+            'scheduledAt' => $futureTime,
+        ]);
+
+        $this->assertEquals(202, $execution['headers']['status-code']);
+        $this->assertEquals('scheduled', $execution['body']['status'], \json_encode($execution['body']));
```
Meldiron, line 281: Not sure if 3rd parameter here is expected

app/controllers/api/functions.php
loks0n: done

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
loks0n: done

tests/e2e/Services/Functions/FunctionsCustomClientTest.php
loks0n: removed

tests/e2e/Services/Functions/FunctionsCustomClientTest.php
loks0n: removed

src/Appwrite/Event/Func.php
loks0n: removed

loks0n: removed

src/Appwrite/Platform/Tasks/ScheduleFunctions.php
loks0n: reverted

app/controllers/api/functions.php
loks0n: Added `metadata` column to hold this information in the `schedules` table

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
loks0n: Went with schedule for now

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    protected function enqueueResources(Group $pools, Database $dbForConsole): void
+    {
+        $queue = $pools->get('queue')->pop();
+        $connection = $queue->getResource();
+        $queueForFunctions = new Func($connection);
+
+        foreach ($this->schedules as $schedule) {
+            if (!$schedule['active']) {
```
Meldiron, line 31: deleteDocument here too

loks0n: done

app/controllers/api/functions.php
```diff
@@ -1591,16 +1592,21 @@
     ->param('path', '/', new Text(2048), 'HTTP path of execution. Path can include query params. Default value is /', true)
     ->param('method', 'POST', new Whitelist(['GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'OPTIONS'], true), 'HTTP method of execution. Default value is GET.', true)
     ->param('headers', [], new Assoc(), 'HTTP headers of execution. Defaults to empty.', true)
+    ->param('scheduledAt', null, new DatetimeValidator(requireDateInFuture: true), 'Scheduled execution time in [ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html) format. DateTime value must be in future.', true)
```
eldadfux, line 1595: Not sure about this name, its in a past tense, but part of an action we're making now

eldadfux (changes requested):

app/controllers/api/functions.php
Meldiron: Made a task to discuss after QA is out

app/config/collections.php
```diff
@@ -4550,6 +4550,17 @@
...
+                '$id' => ID::custom('data'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 65535,
+                'signed' => true,
+                'required' => false,
+                'default' => new \stdClass(),
+                'array' => false,
+                'filters' => ['json', 'encrypt'],
+            ],
```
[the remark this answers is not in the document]
Meldiron: We need to add a column, but it doesn't require migration - patch script is enough. Like we did for some billing columns and tables.

app/controllers/api/functions.php
```diff
@@ -1736,20 +1748,42 @@
...
+                    'region' => System::getEnv('_APP_REGION', 'default'),
+                    'resourceType' => 'execution',
+                    'resourceId' => $execution->getId(),
+                    'resourceInternalId' => $execution->getInternalId(),
+                    'resourceUpdatedAt' => DateTime::now(),
+                    'projectId' => $project->getId(),
+                    'schedule' => $scheduledAt,
+                    'data' => $data,
+                    'active' => true,
+                ]));
```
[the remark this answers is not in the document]
Meldiron: In another PR, for 1.6, we are making it so logging=false still creates document. We can ignore this bug, as it wont be there once all PRs are merged.

lohanidamodar (approved):
Approving this for Q/A release, confirming that the naming changes required will be taken care of during QA week.

PR #8251, opened 2024-06-10
Implement CPU and Memory runtime controls

PineappleIOnic:
## What does this PR do?

This PR implements CPU and Memory controls for Runtimes, adding new `cpus` and `memory` attributes onto the functions document and it's relevant API endpoints. `_APP_FUNCTIONS_CPUS` and `_APP_FUNCTIONS_MEMORY` are now utilised as the max amount of memory and CPUs that can be allocated to a function.

### To be implemented
- [x] Plan Differentiators in API Parameters
- [x] Using `_APP_FUNCTIONS_CPUS` and `_APP_FUNCTIONS_MEMORY` as limiters and to reflect that in the API
- [x] Implementing the API in console
- [x] Fixing the issue we're having with TS Enums having numeric names
- [x] Implement Tests

## Test Plan
1. Create a function and execution then check it's stats
2. Change the runtime memory, rebuild and check it's stats again
3. Change the runtime CPU, rebuild and check it's stats one more time

## Related PRs and Issues

- https://github.com/appwrite/console/pull/1138

## Checklist


.env
```diff
@@ -66,8 +66,8 @@ _APP_STORAGE_PREVIEW_LIMIT=20000000
 _APP_FUNCTIONS_SIZE_LIMIT=30000000
 _APP_FUNCTIONS_TIMEOUT=900
 _APP_FUNCTIONS_BUILD_TIMEOUT=900
-_APP_FUNCTIONS_CPUS=1
-_APP_FUNCTIONS_MEMORY=1024
```
Meldiron, line 70: Lets use previous values again

app/controllers/api/functions.php
```diff
@@ -160,6 +169,8 @@
     ->param('templateOwner', '', new Text(128, 0), 'The name of the owner of the template.', true)
     ->param('templateRootDirectory', '', new Text(128, 0), 'Path to function code in the template repo.', true)
     ->param('templateBranch', '', new Text(128, 0), 'Production branch for the repo linked to the function template.', true)
+    ->param('memory', 512, fn (array $plan) => empty($plan['memoryLimit']) ? new WhiteList($filterBelowThreshold(MEMORY_VALUES, System::getEnv('_APP_FUNCTIONS_MEMORY')), false, WhiteList::TYPE_INTEGER) : new WhiteList($filterBelowThreshold(MEMORY_VALUES, $plan['memoryLimit']), false, WhiteList::TYPE_INTEGER), 'Memory in MB allocated for the function.', true, ['plan'])
```
Meldiron, line 172: I dont think $plan will be here,  it's Cloud-only.
I would suggest special Memory and CPU validators here instead. Then, on Cloud, we can override those validators and add billing logic.

app/controllers/api/functions.php
```diff
@@ -442,6 +455,26 @@
...
     });
 
+App::get('/v1/functions/specs')
+    ->groups(['api', 'functions'])
+    ->desc('Get available function specs')
+    ->label('scope', 'functions.read')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
+    ->label('sdk.namespace', 'functions')
+    ->label('sdk.method', 'getSpecs')
+    ->label('sdk.description', '/docs/references/functions/get-specs.md')
```
Meldiron, line 465: Let's write markdown file

app/controllers/api/functions.php
Meldiron: This comment applies to all places where you used $plan

src/Appwrite/Platform/Workers/Builds.php
```diff
@@ -410,6 +410,8 @@ protected function buildDeployment(Device $deviceForFunctions, Func $queueForFun
                             source: $source,
                             image: $runtime['image'],
                             version: $version,
+                            cpus: $function->getAttribute('cpus', 1),
+                            memory: $function->getAttribute('memory', 128),
```
Meldiron, line 414: Execution specs and build specs should remain different. We should either hard-code build specs (very high), or make it configurable separatelly.

src/Appwrite/Utopia/Response/Model/Func.php
```diff
@@ -145,6 +145,18 @@ public function __construct()
...
                 'example' => false,
             ])
+            ->addRule('memory', [
+                'type' => self::TYPE_INTEGER,
+                'description' => 'Function execution memory limit in MB.',
+                'default' => 512,
+                'example' => 1024,
+            ])
+            ->addRule('cpus', [
+                'type' => self::TYPE_INTEGER,
```
Meldiron, line 155: Let's confirm with team lead.. Should this be integer, or string? If it's integer I would expect slider on Console. If it was string, I would expect select dropdown.
Same for developers using API, integer indicates I might be able to set it to anything. Using ENUM here might work better

src/Appwrite/Utopia/Response/Model/Func.php
```diff
@@ -145,6 +145,18 @@ public function __construct()
...
+                'description' => 'Function execution cores.',
```
Meldiron, line 156:
```suggestion
                'description' => 'Function execution CPU cores.',
```

src/Executor/Executor.php
```diff
@@ -83,8 +85,8 @@ public function createRuntime(
             'variables' => $variables,
             'remove' => $remove,
             'command' => $command,
-            'cpus' => $this->cpus,
-            'memory' => $this->memory,
```
Meldiron, line 87: We can remove those $this variables, they shouldn' be needed anymore

src/Appwrite/Utopia/Response/Model/Func.php
PineappleIOnic: Enum will probably make more sense here, will update appropiately

src/Appwrite/Platform/Workers/Builds.php
PineappleIOnic: I hard-coded 2 cpu cores and 1024 memory, that should be enough right?

src/Appwrite/Utopia/Response/Model/Func.php
PineappleIOnic: Can we add enums to models? I can't seem to see a `self::TYPE_ENUM` in the Model class

app/controllers/api/functions.php
```diff
@@ -53,6 +53,9 @@
 
 include_once __DIR__ . '/../shared/api.php';
 
+const MEMORY_VALUES = [512, 1024, 2048, 4096, 8192, 16384];
+const CPU_VALUES = [1, 2, 4, 8, 16];
```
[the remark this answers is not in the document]
PineappleIOnic: I moved these into the new validators I just added

app/controllers/api/functions.php
PineappleIOnic: Created the new validators, please re-review

app/controllers/api/functions.php
PineappleIOnic: Added 👍

src/Executor/Executor.php
PineappleIOnic: Alright, done

docs/references/functions/get-specs.md
```diff
@@ -0,0 +1 @@
+Get allowed function and cpus values on this instance.
```
Meldiron, line 1:
```suggestion
Get allowed function memory and cpus values on this instance.
```

src/Appwrite/Utopia/Response/Model/Func.php
Meldiron: Not sure, if we never did it, we can do STRING. Check what we did for `plan` in billing, I think that behaves as enum

src/Executor/Executor.php
Meldiron: I meant those:

CleanShot 2024-07-10 at 09 16 16@2x

src/Executor/Executor.php
```diff
@@ -65,6 +65,8 @@ public function createRuntime(
         string $source,
         string $image,
         string $version,
+        int $cpus = 1,
+        int $memory = 512,
```
Meldiron, line 69: I dont see those two being used.

src/Executor/Executor.php
PineappleIOnic: Whoops, my bad. Fixing now

src/Executor/Executor.php
PineappleIOnic: Yeah since I accidentally deleted the part that used it, adding it back now

PineappleIOnic:
Closing this to retarget it to 1.5.x

PR #8258, opened 2024-06-11
Implement deploymentsStorage metric for projects API

PineappleIOnic:
## What does this PR do?

This PR implements the deploymentsStorageTotal and deploymentsStorageBreakdown metrics for the projects API.

## Test Plan

Tests have been updated to include the new metrics

## Related PRs and Issues

- [Console UI Changes](https://github.com/appwrite/console/pull/1137)

## Checklist


app/controllers/api/project.php
```diff
@@ -176,8 +193,10 @@
             'usersTotal' => $total[METRIC_USERS],
             'bucketsTotal' => $total[METRIC_BUCKETS],
             'filesStorageTotal' => $total[METRIC_FILES_STORAGE],
+            'deploymentsStorageTotal' => $total[METRIC_DEPLOYMENTS_STORAGE],
             'executionsBreakdown' => $executionsBreakdown,
-            'bucketsBreakdown' => $bucketsBreakdown
+            'bucketsBreakdown' => $bucketsBreakdown,
+            'deploymentsStorageBreakdown' => $deploymentsStorageBreakdown,
```
christyjacob4, line 199:
```suggestion
            'deploymentsBreakdown' => $deploymentsBreakdown,
```

app/controllers/api/project.php
```diff
@@ -144,6 +145,22 @@
             ];
         }, $dbForProject->find('buckets'));
 
+        $deploymentsStorageBreakdown = array_map(function ($function) use ($dbForProject) {
```
christyjacob4, line 148: not sure we need to call it `deploymentsStorageBreakdown` because `bucketsBreakdown` already captures the fact that its a breakdown of the buckets storage... So you might as well just use `deploymentsBreakdown`

src/Appwrite/Utopia/Response/Model/UsageProject.php
```diff
@@ -88,6 +94,13 @@ public function __construct()
                 'example' => [],
                 'array' => true
             ])
+            ->addRule('deploymentsStorageBreakdown', [
```
christyjacob4, line 97:
```suggestion
            ->addRule('deploymentsBreakdown', [
```
I think this should suffice

christyjacob4 (commented):
Looks good. Please address the comments

app/controllers/api/project.php
PineappleIOnic: Not sure I agree, we already have `deployments`, `deploymentsTotal`, etc... etc... I don't think it'll be clear in the code that this is for the storage specifically, in the buckets we know we're only tracking storage but in the deployments we're not sure if we're tracking storage or how many exist

fogelito (commented):
LGTM

christyjacob4 (approved):

shimonewman (approved):
LGTM

PR #8297, opened 2024-06-20
Disable validation for subquery to prevent error

stnguyen90:
## What does this PR do?

If there are more than 100 targets, the validation throws an error. This change skips the validation so no error is thrown.

Fixes https://github.com/appwrite/appwrite/issues/8299

## Test Plan

Added new test case which should pass

## Related PRs and Issues

* https://github.com/appwrite/appwrite/issues/8299

Parent PR:

* https://github.com/appwrite/appwrite/pull/8317

## Checklist


app/init.php
```diff
@@ -610,9 +610,17 @@ function (mixed $value, Document $document, Database $database) {
...
+                $database->disableValidation();
+                $targets = $database->find('targets', [
+                    Query::equal('$internalId', $targetIds)
+                ]);
+                return $targets;
+            } catch (\Throwable $th) {
+                return [];
+            } finally {
+                $database->enableValidation();
+            }
```
abnegate, line 623: I don't think we should try/catch here, it will be hard to debug if something goes wrong

stnguyen90: Updated

app/init.php
```diff
@@ -610,9 +610,15 @@ function (mixed $value, Document $document, Database $database) {
             ])
         ));
         if (\count($targetIds) > 0) {
-            return $database->find('targets', [
-                Query::equal('$internalId', $targetIds)
-            ]);
+            try {
+                $database->disableValidation();
```
abnegate, line 614: Let's use the new skip here

stnguyen90: Updated

abnegate (approved):

PR #8308, opened 2024-06-21
Fix 'Missing required attribute "expire"' on `users.createSession()`

### LINKED ISSUE #8280: 🐛 Bug Report: Invalid document structure: Missing required attribute "expire" on `users.createSession()`
### 👟 Reproduction steps

This bug happens when I try to use `createSession(userId)` from the Node SDK, but it appears to also be happening in the PHP SDK when using `$users->createSession($userId)`.

### 👍 Expected behavior

A session should be created for the user ID that is provided.

### 👎 Actual Behavior

The following errors gets thrown and logged
```
[Error] Timestamp: 2024-06-14T12:44:49+00:00
[Error] Method: POST
[Error] URL: /v1/users/:userId/sessions
[Error] Type: Utopia\Database\Exception\Structure
[Error] Message: Invalid document structure: Missing required attribute "expire"
[Error] File: /usr/src/code/vendor/utopia-php/database/src/Database/Database.php
[Error] Line: 3070
```

### 🎲 Appwrite version

Version 1.5.x

### 💻 Operating system

Linux

### 🧱 Your Environment

I have tried this on both Appwrite 1.5.4 and 1.5.7, both running on a DigitalOcean droplet. I'm using the latest version of the Node SDK

stnguyen90:
### Fix expire error when creating user session 

Before this, the Create session API call would throw:

> Invalid document structure: Missing required attribute "expire"

This is because the `expire` attribute is required, but it was omitted
from the document. This PR ensures the `expire` attribute is set when
creating the session document.

Fixes #8280 

### Update session secret to be longer

The create session endpoint created a 6 character secret which is too
short. This changes the secret to be 256 characters which is in line
with the secret for `account.createEmailPasswordSession()`.

## Test Plan

Added a test case

## Related PRs and Issues

- #8280 

Parent PR:

* https://github.com/appwrite/appwrite/pull/8317

## Checklist

### DIFF
--- a/app/controllers/api/users.php
+++ b/app/controllers/api/users.php
@@ -1801,6 +1801,7 @@ App::post('/v1/users/:userId/sessions')
                 'userAgent' => $request->getUserAgent('UNKNOWN'),
                 'ip' => $request->getIP(),
                 'countryCode' => ($record) ? \strtolower($record['country']['iso_code']) : '--',
+                'expire' => $expire,
             ],
             $detector->getOS(),
             $detector->getClient(),
@@ -1812,7 +1813,6 @@ App::post('/v1/users/:userId/sessions')
         $session = $dbForProject->createDocument('sessions', $session);
         $session
             ->setAttribute('secret', $secret)
-            ->setAttribute('expire', $expire)
             ->setAttribute('countryName', $countryName);
 
         $queueForEvents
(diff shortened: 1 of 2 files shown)


app/controllers/api/users.php
```diff
@@ -1784,7 +1784,7 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
             throw new Exception(Exception::USER_NOT_FOUND);
         }
 
-        $secret = Auth::codeGenerator();
+        $secret = Auth::tokenGenerator(Auth::TOKEN_LENGTH_SESSION);
```
stnguyen90, line 1787: @loks0n, do you remember why the secret was created using `Auth::codeGenerator();` which creates a 6 character code?

abnegate (approved):

Issue #8174, opened 2024-05-21
🐛 Bug Report: Unable to get current user session on functions using jwt


Irutom:
### 👟 Reproduction steps

```   
            import { Client} from 'node-appwrite';
            const userClient = new Client();
            const userAccount = new Account(userClient);

            userClient.setEndpoint(process.env.FUNCTIONS_END_POINT as string)
                .setProject(process.env.PROJECT_ID as string)
                .setJWT(jwtToken);

            const currentSession = await userAccount.getSession('current');
```

### 👍 Expected behavior

getSession('current') should return the current user session that is being used.

### 👎 Actual Behavior

Fails with user_session_not_found error.

`
{"code":404,"type":"user_session_not_found","response":{"message":"The current user session could not be found."...}},
`

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

Linux

### 🧱 Your Environment

1.5.5

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

nickaldwin:
Hi @stnguyen90  I'm interested in working on this issue. Can I work on it?

stnguyen90:
@nickaldwin, assigned! Thanks for your interest! 🙏🏼

nickaldwin:
> @nickaldwin, assigned! Thanks for your interest! 🙏🏼

Thank you so much

nickaldwin:
Hi @stnguyen90, Good morning and good evening! I hope you're doing great. This is my first time contributing to Appwrite. I'm having trouble replicating the issue on my end. I've followed the setup instructions and tried the reproduction steps, but I'm still facing difficulties.

Could you please provide more details on how to access and use the Appwrite Cloud function to replicate the issue?

stnguyen90:
@nickaldwin, have you looked through [these](https://appwrite.io/docs/products/functions) docs? 

That said, you don't even need to use an Appwrite function for this. You should be able to use a client SDK to

1. [Create an email password session](https://appwrite.io/docs/references/cloud/client-web/account#createEmailPasswordSession)
2. [Create a JWT token](https://appwrite.io/docs/references/cloud/client-web/account#createJWT)

[...]

PR #8313, opened 2024-06-24
Pass session data in JWT and use that to authenticate when using JWT

Souptik2001:
## What does this PR do?

As per mentioned in the attached issue, when using JWT as an authentication method, if we don't pass the session in the "get session" endpoint, then the authentication fails.
But that shouldn't be the case ideally, as JWT is meant for authentication and it should be able to authenticate even without the need of session. Or more specifically it should include the session data in itself, which should be used for authentication.

- The reason why the authentication is failing because the `Auth::$secret` value which should be equals to the `$session['secret']`, where `$session` is the decoded value of the client session string.
- But when the client session string is not passed, then the `Auth::$secret` value is not populated, and that's why later where we check if the hash of `Auth::$secret` value is equal to the `secret` of the matched session object, it is failing.
- Therefore our target is to populate `Auth::$secret` value when using JWT. But that is only possible when we get the decoded value of the client session string as mentioned in the second point.
- So, to get that data we have to store that data in the JWT and then we can easily get that data.
- Now to store that data in JWT, we have to get the client session data in the `accounts/jwt` handler function. But there is not resource currently present which provides that data.
- So, we have to create a new resource called `clientSession` in `app/init.php` and now consume that resource in the above mentioned handler function and pass the data to the JWT.

## Test Plan

We can test this in Postman -
- Create a new session using`http://localhost/v1/account/sessions/email` endpoint.
  - It will automatically set the cookies. If not then in the next step you have to pass the session value using `x-appwrite-session` header.
- Now create a new JWT using this endpoint - `http://localhost/v1/account/jwt`.
  - Copy the JWT.
- Now create a request to the get session endpoint - `http://localhost/v1/account/sessions/current`, with the JWT set and also the session set.
  - It should work.
- Now delete the cookies (session). If you were using header, then remove the header.
- Now request again and it should still work (**this step should have been failed earlier without this PR changes**)

Screenshot 2024-06-24 at 2 51 44 PM

Screenshot 2024-06-24 at 2 52 20 PM

Screenshot 2024-06-24 at 2 52 44 PM

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/8174

## Checklist

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -2358,8 +2358,9 @@ App::post('/v1/account/jwt')
     ->label('abuse-key', 'url:{url},userId:{userId}')
     ->inject('response')
     ->inject('user')
+    ->inject('clientSession')
     ->inject('dbForProject')
-    ->action(function (Response $response, Document $user, Database $dbForProject) {
+    ->action(function (Response $response, Document $user, array $clientSession, Database $dbForProject) {
 
 
         $sessions = $user->getAttribute('sessions', []);
@@ -2386,6 +2387,7 @@ App::post('/v1/account/jwt')
                 // 'iss'    => 'http://api.mysite.com',
                 'userId' => $user->getId(),
                 'sessionId' => $current->getId(),
+                'session' => json_encode($clientSession),
             ])]), Response::MODEL_JWT);
     });
 
--- a/app/init.php
+++ b/app/init.php
@@ -1214,6 +1214,7 @@ App::setResource('user', function ($mode, $project, $console, $request, $respons
 
         $jwtUserId = $payload['userId'] ?? '';
         $jwtSessionId = $payload['sessionId'] ?? '';
+        $jwtSession = json_decode($payload['session'] ?? '', true);
 
         if ($jwtUserId && $jwtSessionId) {
             $user = $dbForProject->getDocument('users', $jwtUserId);
@@ -1221,6 +1222,9 @@ App::setResource('user', function ($mode, $project, $console, $request, $respons
 
         if (empty($user->find('$id', $jwtSessionId, 'sessions'))) { // Match JWT to active token
             $user = new Document([]);
+        } else {
+            Auth::$unique = $jwtSession['id'] ?? '';
+            Auth::$secret = $jwtSession['secret'] ?? '';
         }
     }
 
@@ -1245,6 +1249,38 @@ App::setResource('project', function ($dbForConsole, $request, $console) {
 
     return $project;
 }, ['dbForConsole', 'request', 'console']);
+
+App::setResource('clientSession', function (string $mode, Document $project, Document $console, Appwrite\Utopia\Request $request) {
+    Auth::setCookieName('a_session_' . $project->getId());
+
+    if (APP_MODE_ADMIN === $mode) {
+        Auth::setCookieName('a_session_' . $console->getId());
+    }
+
+    $session = Auth::decodeSession(
+        $request->getCookie(
+            Auth::$cookieName, // Get sessions
+            $request->getCookie(Auth::$cookieName . '_legacy', '')
+        )
+    );
+
+    // Get session from header for SSR clients
+    if (empty($session['id']) && empty($session['secret'])) {
+        $sessionHeader = $request->getHeader('x-appwrite-session', '');
+
+        if (!empty($sessionHeader)) {
+            $session = Auth::decodeSession($sessionHeader);
+        }
+    }
+
+    if (empty($session['id']) && empty($session['secret'])) {
+        $fallback = $request->getHeader('x-fallback-cookies', '');
+        $fallback = \json_decode($fallback, true);
+        $session = Auth::decodeSession(((isset($fallback[Auth::$cookieName])) ? $fallback[Auth::$cookieName] : ''));
+    }
+
+    return $session;
+}, ['mode', 'project', 'console', 'request']);
 
 App::setResource('session', function (Document $user) {
     if ($user->isEmpty()) {


app/init.php
```diff
@@ -1246,6 +1250,38 @@ function (mixed $value) {
...
+
+    if (empty($session['id']) && empty($session['secret'])) {
+        $fallback = $request->getHeader('x-fallback-cookies', '');
+        $fallback = \json_decode($fallback, true);
+        $session = Auth::decodeSession(((isset($fallback[Auth::$cookieName])) ? $fallback[Auth::$cookieName] : ''));
+    }
+
+    return $session;
+}, ['mode', 'project', 'console', 'request']);
+
```
stnguyen90, line 1284: Can we avoid setting an additional resource considering we already have the `user` resource?

Also, we have a `session` resource below this.

app/init.php
```diff
@@ -1214,13 +1214,17 @@ function (mixed $value) {
 
         $jwtUserId = $payload['userId'] ?? '';
         $jwtSessionId = $payload['sessionId'] ?? '';
+        $jwtSession = json_decode($payload['session'] ?? '', true);
```
stnguyen90, line 1217: Weird... I don't think the payload has `session` It should only have `userId` and `sessionId`:

https://github.com/appwrite/appwrite/blob/e3bc1887a4abbc8de3b82d0ec7aadc5e9f46c8a1/app/controllers/api/account.php#L2380-L2389

stnguyen90 (changes requested):

app/controllers/api/account.php
```diff
@@ -2386,6 +2387,7 @@
                 // 'iss'    => 'http://api.mysite.com',
                 'userId' => $user->getId(),
                 'sessionId' => $current->getId(),
+                'session' => json_encode($clientSession),
```
stnguyen90, line 2390: I'd prefer to avoid adding this `session`.

stnguyen90 (changes requested):

stnguyen90 (changes requested):
I think we need to rely a little less on the session secret. We should validate once and then rely on the `Auth::$unique`, instead. Relying on `Auth::$secret` requires us to iterate over session and hash, which can be expensive.

Please also add a test case to verify calling get current session works with a JWT.

tests/e2e/Services/Account/AccountCustomClientTest.php
```diff
@@ -1500,6 +1500,15 @@ public function testCreateJWT(): array
...
         $this->assertEquals(200, $response['headers']['status-code']);
 
+        $response = $this->client->call(Client::METHOD_GET, '/account/sessions/current', array_merge([
+            'origin' => 'http://localhost',
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'x-appwrite-jwt' => $jwt,
+        ]));
+
+        $this->assertEquals(200, $response['headers']['status-code']);
```
Souptik2001, line 1510: This previously returned 404. But now it is correctly returning 200.

Souptik2001:
Thanks @stnguyen90! Got your point. 👍  It makes sense to use `Auth::$secret` at only one place instead of multiple places.

Therefore I have removed `Auth::$secret` check from all other places except the `user` resource callback. User resource will be the source of truth and the check will be done over there.

But the `Auth::$secret` was used at all other places to get the `session ID`, but now as we are not relying on `Auth::$secret` therefore we have to pass the `session ID` to be consumed at other places.
Therefore I have added another static variable in the `Auth` class called - `$sessionId`. And this will be set to the current session ID in the user resource callback function.

And now this `Auth::$sessionId` is used everywhere else.

How does this look now?

I have also added the test case. 👍

app/init.php
```diff
@@ -1219,8 +1221,12 @@ function (mixed $value) {
             $user = $dbForProject->getDocument('users', $jwtUserId);
         }
 
-        if (empty($user->find('$id', $jwtSessionId, 'sessions'))) { // Match JWT to active token
+        $session = $user->find('$id', $jwtSessionId, 'sessions');
+
+        if (empty($session) || Auth::isSessionExpired($session)) { // Match JWT to active token
```
Souptik2001, line 1226: @stnguyen90 This missing time check was the cause for this #8000 issue. Therefore added this session expired check over here. That should be also fixed now, as per my tests. 🤞 

Just FYI - For JWT we still don't have a secret check like we have in sessions, that's because we don't have that data in JWT. But I think that's the point of directly storing the userID, so that we don't have to depend on the session secret, right?

tests/e2e/Services/Account/AccountCustomClientTest.php
```diff
@@ -1500,6 +1500,15 @@ public function testCreateJWT(): array
...
+
```
Souptik2001, line 1511: @stnguyen90 I was thinking to add a test for checking the expiry time condition for JWT to validate the changes of #8000 I did above.

Should we add that? If yes, is there any recommended way here to change the default expiry time, or use some other session creation endpoint which allows setting the expiry time?

app/init.php
stnguyen90: > Just FYI - For JWT we still don't have a secret check like we have in sessions, that's because we don't have that data in JWT. But I think that's the point of directly storing the userID, so that we don't have to depend on the session secret, right?

Yes. We just need to check the JWT is valid (`$jwt->decode()`) and if it is, trust the value is valid.

tests/e2e/Services/Account/AccountCustomClientTest.php
stnguyen90: Can you make an API call to update the project auth settings and change the session expiry before creating the session?

Otherwise, you can try doing a manual test and add screenshot(s) as evidence.

src/Appwrite/Auth/Auth.php
```diff
@@ -110,6 +110,13 @@ class Auth
      */
     public static $secret = '';
 
+    /**
+     * User session ID.
+     *
+     * @var string
+     */
+    public static $sessionId = '';
+
```
stnguyen90, line 119: Oye..I didn't realize we don't have this yet...

@TorstenDittmann, is this okay so that we:

1. prevent finding the session using the secret multiple times as it can be expensive to iterate over every session and verify the secret hash
2. not rely on a non-existent session secret when a JWT is provided.

tests/e2e/Services/Account/AccountCustomClientTest.php
Souptik2001: @stnguyen90 Thanks for the idea of calling project auth API to change the expiry time!

Did that, and added the test to test if JWT properly respects session expiry time!

HarshMN2345:

PR #8315, opened 2024-06-24
Feat new session alert

loks0n:


tests/e2e/Services/Account/AccountCustomClientTest.php
```diff
@@ -1190,6 +1190,64 @@ public function testUpdateAccountRecovery($data): array
...
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'user-agent' => 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/126.0.0.0 Safari/537.36',
+        ]), [
+            'email' => $email,
+            'password' => $password,
+        ]);
+
+        $this->assertEquals(201, $response['headers']['status-code']);
+
```
Meldiron, line 1240: Sleep 3 or 5 seconds. check what we do elsewhere when we do getLastEmail

byawitz (approved):

app/controllers/api/projects.php
```diff
@@ -104,8 +104,10 @@
             'passwordHistory' => 0,
             'passwordDictionary' => false,
             'duration' => Auth::TOKEN_EXPIRATION_LOGIN_LONG,
-            'personalDataCheck' => false
+            'personalDataCheck' => false,
+            'sessionAlerts' => false,
```
Meldiron, line 108: I vote for `true`  as default.

Meldiron, line 108: I think we should also set it to `true` in `Console` project document.

Meldiron: Added myself

Meldiron (approved):

vermakhushboo (approved):

app/controllers/api/projects.php
```diff
@@ -603,6 +605,37 @@
...
+    ->groups(['api', 'projects'])
+    ->label('scope', 'projects.write')
+    ->label('sdk.auth', [APP_AUTH_TYPE_ADMIN])
+    ->label('sdk.namespace', 'projects')
+    ->label('sdk.method', 'updateSessionAlerts')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_PROJECT)
+    ->param('projectId', '', new UID(), 'Project unique ID.')
+    ->param('sessionAlerts', false, new Boolean(true), 'Set to true to enable session emails.')
```
abnegate, line 619: `enabled` might be a better name here, since the scope is already session alerts

app/init.php
```diff
@@ -1323,6 +1323,7 @@ function (mixed $value) {
             'invites' => System::getEnv('_APP_CONSOLE_INVITES', 'enabled') === 'enabled',
             'limit' => (System::getEnv('_APP_CONSOLE_WHITELIST_ROOT', 'enabled') === 'enabled') ? 1 : 0, // limit signup to 1 user
             'duration' => Auth::TOKEN_EXPIRATION_LOGIN_LONG, // 1 Year in seconds
+            'sessionAlerts' => true
```
abnegate, line 1326: This should be configurable via env var

app/controllers/api/projects.php
Meldiron: I looked the same direction, but in other endpoints we do this too. like `updateAuthLimit(limit)` or `updateAuthDuration(duration)` or `updateAuthStatus(status)`.
If you are against this pattern, we can make task in 1.6 QA board and discuss, but for now lets follow what is there.

With that in mind, I am changing it from `updateSessionAlerts(sessionAlerts)` to `updateSessionAlerts(alerts)`.

app/init.php
Meldiron: Added `_APP_CONSOLE_SESSION_ALERTS`

PR #8325, opened 2024-06-25
Add expiration  headers  to a missed cache response

shimonewman:
Added a set of expiration headers to a cache missed request to enforce a fresh  file response when 
a file is overridden with the same custom ID.


stnguyen90:
Closing as the change was already merged.

PR #8350, opened 2024-07-02
Tests: Ensure CI/CD works

Meldiron:
Lust a lazy way to ensure tests are passing


tests/resources/functions/php-fn/index.php
```diff
@@ -10,7 +10,7 @@
         'APPWRITE_FUNCTION_TRIGGER' => $context->req->headers['x-appwrite-trigger'] ?? '',
         'APPWRITE_FUNCTION_RUNTIME_NAME' => \getenv('APPWRITE_FUNCTION_RUNTIME_NAME') ?: '',
         'APPWRITE_FUNCTION_RUNTIME_VERSION' => \getenv('APPWRITE_FUNCTION_RUNTIME_VERSION') ?: '',
-        'APPWRITE_VERSION' => APP_VERSION_STABLE,
+        'APPWRITE_VERSION' => \getenv('APPWRITE_VERSION') ?: '',
```
byawitz, line 13: Is the `APPWRITE_VERSION` need to be returned if not being checked anymore?

byawitz (approved):

tests/resources/functions/php-fn/index.php
Meldiron: It is still checked, but in other tests. I removed it from test that tried to do the test for multiple runtimes, as there is no need for that. it still remains in 2 tests (sync and async execution,  of `php-fn`)

eldadfux (approved):

PR #8366, opened 2024-07-08
Feat project last activity

lohanidamodar:
## What does this PR do?

- Adds accessed at attribute to project that updates once every 24 hours
- Needs patch to create this attribute for migration

## Test Plan

- Same as `accessedAt` attribute in `users`
- Tested manually, as this is not included in response model

## Related PRs and Issues

- (Related PR or issue)

## Checklist


Meldiron (approved):

Meldiron (approved):

app/controllers/api/projects.php
```diff
@@ -22,6 +22,7 @@
 use Utopia\Cache\Cache;
 use Utopia\Config\Config;
 use Utopia\Database\Database;
+use Utopia\Database\DateTime as DatabaseDateTime;
```
christyjacob4, line 25:
```suggestion
use Utopia\Database\DateTime;
```
Can we just not use this ?

christyjacob4 (approved):
Looks good. Just one small comment

app/controllers/api/projects.php
lohanidamodar: Done.

PR #8367, opened 2024-07-08
Update delete authenticator to remove OTP Validation

PineappleIOnic:
## What does this PR do?

 This PR modifies the delete authenticator endpoint to no longer need a challenge verification and to instead make it a `mfaProtected` route so only recently authorised sessions can delete an authenticator.

This PR also generates the documentation for 1.6.x and adds the spec files.

## Related PRs and Issues

- https://github.com/appwrite/appwrite/pull/8282 depends on this

## Checklist


TorstenDittmann (approved):

docs/examples/1.5.x/client-rest/examples/account/delete-mfa-authenticator.md
stnguyen90, line 1: Probably shouldn't delete it for 1.5, right?

PineappleIOnic: In this case don't we need to generate all the docs for 1.6 then?

app/controllers/api/account.php
```diff
@@ -3945,7 +3945,7 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
 
 App::delete('/v1/account/mfa/authenticators/:type')
     ->desc('Delete Authenticator')
-    ->groups(['api', 'account'])
+    ->groups(['api', 'account', 'mfaProtected'])
```
abnegate, line 3948: Should there be a corresponding hook for this group?

app/controllers/api/account.php
```diff
@@ -3958,12 +3958,11 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
     ->label('sdk.response.code', Response::STATUS_CODE_NOCONTENT)
     ->label('sdk.response.model', Response::MODEL_NONE)
     ->param('type', null, new WhiteList([Type::TOTP]), 'Type of authenticator.')
-    ->param('otp', '', new Text(256), 'Valid verification token.')
```
abnegate, line 3961: We'll need a request filter for this removal

app/controllers/api/account.php
PineappleIOnic: It already exists: https://github.com/appwrite/appwrite/blob/c87a7053b4f0160ea8de2e4054f8f3cc503772de/app/controllers/shared/api/auth.php#L13-L30

app/controllers/api/account.php
PineappleIOnic: Noted :+1:

abnegate (approved):

PR #8379, opened 2024-07-10
Add APPWRITE_ENV to vars

vermakhushboo:
## What does this PR do?

When you appwrite run function locally using Appwrite CLI, variable`APPWRITE_ENV` is set to development. This is useful to have test cases separated from your main code-base, based on this variable.

In Appwrite, we don't set this variable which can lead to some undefined/nulls errors. This PR adds APPWRITE_ENV to all executions. Hard-coded it to always be production.

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


Meldiron (approved):

app/controllers/general.php
```diff
@@ -250,7 +250,8 @@ function router(App $utopia, Database $dbForConsole, callable $getProjectDB, Swo
             'APPWRITE_FUNCTION_PROJECT_ID' => $project->getId(),
             'APPWRITE_FUNCTION_RUNTIME_NAME' => $runtime['name'] ?? '',
             'APPWRITE_FUNCTION_RUNTIME_VERSION' => $runtime['version'] ?? '',
-            'APPWRITE_VERSION' => APP_VERSION_STABLE
+            'APPWRITE_VERSION' => APP_VERSION_STABLE,
+            'APPWRITE_ENV' => 'production',
```
christyjacob4, line 254: why is this hard coded? There is an environment variable for this. 
https://github.com/appwrite/appwrite/blob/main/app/config/variables.php#L11
This value should be used

vermakhushboo: I know it was hard-coded for CLI. @Meldiron can we use the variable?

vermakhushboo:
Closing this PR as we already offer the ability to pass custom headers and custom environment variables and the developer can pass any custom variable they like, for CI/CD they can send APPWRITE_ENV from their side and its not something we need to send explicitly.

PR #8381, opened 2024-07-11
Remove relationship attributes from realtime event payloads

byawitz:
## What does this PR do?

Mark all relationship attributes as sensitive so they aren't passed via realtime to prevent sending data to subscribers who shouldn't have access to the nested data.

## Test Plan

Manually 

## Related PRs and Issues

None

## Checklist


stnguyen90 (approved):
verified no relationships are included in realtime event payload for create, update, and delete:

image

PR #8384, opened 2024-07-12
Feat runtime controls

PineappleIOnic:
## What does this PR do?

This PR implements runtime controls for `cpus` and `memory` within Appwrite for 1.5.x

## Test Plan

Tests have been written to check the functions document gets correctly updated and limits work

## Checklist


app/controllers/api/functions.php
```diff
@@ -160,6 +162,8 @@
     ->param('templateOwner', '', new Text(128, 0), 'The name of the owner of the template.', true)
     ->param('templateRootDirectory', '', new Text(128, 0), 'Path to function code in the template repo.', true)
     ->param('templateBranch', '', new Text(128, 0), 'Production branch for the repo linked to the function template.', true)
+    ->param('memory', 512, new Memory(), 'Memory in MB allocated for the function.', true)
+    ->param('cpus', 1, new Cpus(), 'CPU cores allocated for the function.', true)
```
christyjacob4, line 166: as we discussed we should replace these two params with a `size` parameter which should accept constants like `s-1vcpu-512mb` etc.

app/controllers/api/functions.php
```diff
@@ -229,6 +233,8 @@
             'providerBranch' => $providerBranch,
             'providerRootDirectory' => $providerRootDirectory,
             'providerSilentMode' => $providerSilentMode,
+            'memory' => $memory,
+            'cpus' => $cpus
```
christyjacob4, line 237: The collection itself should contain both CPU and memory attributes as this allows us to be more flexible in the future. so you need to map the string `s-1vcpu-512mb` which comes in as a param to 1 CPU and 512MB memory

app/config/collections.php
```diff
@@ -3054,7 +3054,29 @@
...
-            ]
+            ],
+            [
+                'array' => false,
+                '$id' => ID::custom('memory'),
+                'type' => Database::VAR_INTEGER,
+                'format' => '',
+                'size' => 512,
+                'signed' => false,
+                'required' => true,
```
christyjacob4, line 3065: These attributes should not be required. 

Else we can't to deploy this change without a migration

app/controllers/api/functions.php
```diff
@@ -442,6 +448,25 @@
         ]), Response::MODEL_RUNTIME_LIST);
     });
 
+App::get('/v1/functions/specs')
```
christyjacob4, line 451: lets not use short forms and use `specifications` or `sizes` since the param is called `size`

app/controllers/api/functions.php
```diff
@@ -442,6 +448,25 @@
...
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_SPECS)
+    ->inject('response')
+    ->action(function (Response $response) {
+        $response->dynamic(new Document([
+            'memory' => Memory::getAllowedValues(),
+            'cpus' => Cpus::getAllowedValues()
+        ]), Response::MODEL_SPECS);
+    });
```
christyjacob4, line 468: as discussed, we should probably store the list of all sizes as a `Config` or something. And that way we can reuse the config inside the validators & the response here.

Then we can use the `plan` object to store the whitelist of allowed sizes for each tier. 
https://github.com/appwrite/appwrite/blob/main/app/init.php#L1698-L1700

app/config/collections.php
```diff
@@ -3054,7 +3054,40 @@
...
+                'array' => false,
+                '$id' => ID::custom('size'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 128,
+                'signed' => false,
+                'required' => false,
+                'default' => 's-1vcpu-512mb',
+                'filters' => [],
+            ],
```
christyjacob4, line 3090: we don't need a size attribute here. `size` is only for the API

app/config/runtimes/enum.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+enum Sizes: string
+{
+    case S_1VCPU_512MB = 's-1vcpu-512mb';
+    case S_1VCPU_1GB = 's-1vcpu-1gb';
+    case S_2VCPU_1GB = 's-2vcpu-1gb';
+    case S_2VCPU_4GB = 's-2vcpu-4gb';
```
christyjacob4, line 8:
```suggestion
    case S_2VCPU_2GB = 's-2vcpu-2gb';
    case S_2VCPU_4GB = 's-2vcpu-4gb';
```

christyjacob4: might be a better choice

app/config/runtimes/sizes.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+require_once __DIR__ . '/enum.php';
```
christyjacob4, line 3: Why not an import?

app/config/runtimes/sizes.php
```diff
@@ -0,0 +1,22 @@
...
+return [
+    Sizes::S_1VCPU_512MB->value => [
+        'memory' => 512,
+        'cpus' => 1
+    ],
+    Sizes::S_1VCPU_1GB->value => [
+        'memory' => 1024,
+        'cpus' => 1
+    ],
+    Sizes::S_2VCPU_1GB->value => [
```
christyjacob4, line 14: lets update this value

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -386,6 +386,71 @@ public function testUpdate($data): array
...
+        $response1 = $this->client->call(Client::METHOD_PUT, '/functions/' . $data['functionId'], array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()), [
+            'name' => 'Test1',
+            'events' => [
+                'users.*.update.name',
+                'users.*.update.email',
+            ],
+            'schedule' => '0 0 1 1 *',
```
christyjacob4, line 403: lets not create a scheduled function here .. you can remove this param

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -386,6 +386,71 @@ public function testUpdate($data): array
...
+            'timeout' => 15,
+            'runtime' => 'php-8.0',
+            'entrypoint' => 'index.php',
+            'size' => 's-2vcpu-512mb', // Invalid size
+        ]);
+
+        $this->assertEquals(400, $response2['headers']['status-code']);
+
```
christyjacob4, line 447: lets also validate the error message

src/Executor/Executor.php
```diff
@@ -24,19 +24,13 @@ class Executor
...
-
     public function __construct(string $endpoint)
     {
         if (!filter_var($endpoint, FILTER_VALIDATE_URL)) {
             throw new Exception('Unsupported endpoint');
         }
 
         $this->endpoint = $endpoint;
-        $this->cpus = \intval(System::getEnv('_APP_FUNCTIONS_CPUS', '1'));
-        $this->memory = \intval(System::getEnv('_APP_FUNCTIONS_MEMORY', '512'));
```
christyjacob4, line 39: are we deprecating these variables now? Are they used anywhere?

src/Appwrite/Specification/Format/Swagger2.php
```diff
@@ -428,7 +428,7 @@ public function parse(): array
                             }
                         }
 
-                        if ($allowed) {
+                        if ($allowed && $validator->getType() === 'string') {
```
christyjacob4, line 431: whats this change supposed to be?

src/Appwrite/Platform/Workers/Builds.php
```diff
@@ -410,6 +410,8 @@ protected function buildDeployment(Device $deviceForFunctions, Func $queueForFun
                             source: $source,
                             image: $runtime['image'],
                             version: $version,
+                            cpus: 2,
+                            memory: 1024,
```
christyjacob4, line 414: why is this hard coded here?

src/Appwrite/Platform/Workers/Functions.php
```diff
@@ -507,6 +509,8 @@ private function execute(
                 ->addMetric(str_replace('{functionInternalId}', $function->getInternalId(), METRIC_FUNCTION_ID_EXECUTIONS), 1)
                 ->addMetric(METRIC_EXECUTIONS_COMPUTE, (int)($execution->getAttribute('duration') * 1000))// per project
                 ->addMetric(str_replace('{functionInternalId}', $function->getInternalId(), METRIC_FUNCTION_ID_EXECUTIONS_COMPUTE), (int)($execution->getAttribute('duration') * 1000))
+                ->addMetric(METRIC_EXECUTIONS_MB_SECONDS, $function->getAttribute('memory') * $execution->getAttribute('duration', 0) * $function->getAttribute('cpus', 1))
```
christyjacob4, line 512: why is there no default value for `memory` ?

src/Appwrite/Platform/Workers/Builds.php
```diff
@@ -540,9 +542,11 @@ protected function buildDeployment(Device $deviceForFunctions, Func $queueForFun
                 ->addMetric(METRIC_BUILDS, 1) // per project
                 ->addMetric(METRIC_BUILDS_STORAGE, $build->getAttribute('size', 0))
                 ->addMetric(METRIC_BUILDS_COMPUTE, (int)$build->getAttribute('duration', 0) * 1000)
+                ->addMetric(METRIC_BUILDS_MB_SECONDS, $function->getAttribute('memory') * $build->getAttribute('duration', 0) * $function->getAttribute('cpus', 1))
```
christyjacob4, line 545: why is there no default value for `memory` ?

src/Appwrite/Platform/Workers/Builds.php
```diff
@@ -540,9 +542,11 @@ protected function buildDeployment(Device $deviceForFunctions, Func $queueForFun
...
                 ->addMetric(str_replace('{functionInternalId}', $function->getInternalId(), METRIC_FUNCTION_ID_BUILDS), 1) // per function
                 ->addMetric(str_replace('{functionInternalId}', $function->getInternalId(), METRIC_FUNCTION_ID_BUILDS_STORAGE), $build->getAttribute('size', 0))
                 ->addMetric(str_replace('{functionInternalId}', $function->getInternalId(), METRIC_FUNCTION_ID_BUILDS_COMPUTE), (int)$build->getAttribute('duration', 0) * 1000)
+                ->addMetric(str_replace('{functionInternalId}', $function->getInternalId(), METRIC_FUNCTION_ID_BUILDS_MB_SECONDS), $function->getAttribute('memory') * $build->getAttribute('duration', 0) * $function->getAttribute('cpus', 1))
```
christyjacob4, line 549: why is there no default value for `memory` ?

app/config/collections.php
PineappleIOnic: but if we don't know the machine type, we're going to have to work backwards and recalculate it on the console and work backwards from the CPU's and Memory to show the values allowed

app/config/runtimes/sizes.php
PineappleIOnic: Updated to use a namespace import

app/config/runtimes/sizes.php
PineappleIOnic: Updated

src/Executor/Executor.php
PineappleIOnic: No, these are now enforced as limits within the validators

src/Appwrite/Specification/Format/Swagger2.php
PineappleIOnic: This was for SDK Enums, as having non-string enums wasn't supported in most languages. We can removed it but I'm leaning towards keeping it just incase

src/Appwrite/Platform/Workers/Builds.php
PineappleIOnic: We decided on using the functions limits right? reverting

app/config/runtimes/sizes.php
PineappleIOnic: Nothing in `/app` is in the autoloader so we need to either add /app to the autoloader for this one file or require the fil

src/Appwrite/Platform/Workers/Functions.php
PineappleIOnic: Added a default of 512mb

src/Appwrite/Platform/Workers/Builds.php
PineappleIOnic: Added a default of 512mb

src/Appwrite/Platform/Workers/Builds.php
PineappleIOnic: Added a default of 512mb

app/config/collections.php
Meldiron: Can we only store size, and then get cpus/memory from config, when needed?

app/controllers/api/functions.php
```diff
@@ -160,6 +161,7 @@
     ->param('templateOwner', '', new Text(128, 0), 'The name of the owner of the template.', true)
     ->param('templateRootDirectory', '', new Text(128, 0), 'Path to function code in the template repo.', true)
     ->param('templateBranch', '', new Text(128, 0), 'Production branch for the repo linked to the function template.', true)
+    ->param('size', 's-1vcpu-512mb', fn (array $plan) => new RuntimeSize($plan), 'Runtime size for the function.', true, ['plan'])
```
Meldiron, line 164: In description I would mention it applyies both for builds and excution

app/controllers/api/functions.php
```diff
@@ -229,6 +237,9 @@
             'providerBranch' => $providerBranch,
             'providerRootDirectory' => $providerRootDirectory,
             'providerSilentMode' => $providerSilentMode,
+            'memory' => $spec['memory'],
+            'cpus' => $spec['cpus'],
+            'size' => $size
```
Meldiron, line 242: We dont need all 3 in db. Either memory&cpu, or size.

app/controllers/api/functions.php
```diff
@@ -795,6 +840,23 @@
...
+        if (empty($spec) || empty($spec['memory']) || empty($spec['cpus'])) {
+            throw new Exception(Exception::FUNCTION_INVALID_RUNTIME_SIZE);
+        }
+
+        // Enforce Cold Start if spec limits change.
+        if (($function->getAttribute('cpus') !== $spec['cpus'] ||
+            $function->getAttribute('memory') !== $spec['memory']) && !empty($function->getAttribute('deployment'))) {
+            $executor = new Executor(App::getEnv('_APP_EXECUTOR_HOST'));
+            try {
+                $executor->deleteRuntime($project->getId(), $function->getAttribute('deployment'));
```
Meldiron, line 854: Lets make task in backlog, this should be deploymentId. We should address it in 1.7

app/config/collections.php
PineappleIOnic: Yeah, just updated that

app/controllers/api/functions.php
PineappleIOnic: Done

app/controllers/api/functions.php
```diff
@@ -202,6 +204,12 @@
             throw new Exception(Exception::GENERAL_ARGUMENT_INVALID, 'When connecting to VCS (Version Control System), you need to provide "installationId" and "providerBranch".');
         }
 
+        $spec = Config::getParam('runtime-specifications')[$specification] ?? [];
+
+        if (empty($spec) || empty($spec['memory']) || empty($spec['cpus'])) {
+            throw new Exception(Exception::FUNCTION_INVALID_RUNTIME_SPECIFICATION);
+        }
+
```
Meldiron, line 212: This should not be needed, ideally. Ideally validator checks all of this

app/controllers/api/functions.php
```diff
@@ -442,6 +451,57 @@
         ]), Response::MODEL_RUNTIME_LIST);
     });
 
+App::get('/v1/functions/specifications')
+    ->groups(['api', 'functions'])
+    ->desc('Get available function runtime specifications')
+    ->label('scope', 'functions.read')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY])
```
Meldiron, line 458: I think this can be public endpoint.
If not agreed upon, then at least `APP_AUTH_TYPE_ADMIN` alongside API key

app/controllers/api/functions.php
```diff
@@ -442,6 +451,57 @@
...
+            } else {
+                // Show plan names on specs
+                foreach ($plans as $billingPlan) {
+                    if (key_exists('runtimeSpecifications', $billingPlan) && in_array($key, $billingPlan['runtimeSpecifications'])) {
+                        if ($billingPlan['$id'] !== 'tier-0') {
+                            $spec['plan'] = $billingPlan['name'];
+                        }
+                        break;
+                    }
+                }
```
Meldiron, line 485: Can this be abstracted to cloud repository, by using triggers? Like we used to use for ZDT.

app/controllers/api/functions.php
```diff
@@ -1536,6 +1625,10 @@
 
         $version = $function->getAttribute('version', 'v2');
         $runtimes = Config::getParam($version === 'v2' ? 'runtimes-v2' : 'runtimes', []);
+        $spec = Config::getParam('runtime-specifications')[$function->getAttribute('specification', 's-1vcpu-512mb')];
+
+        var_dump($function);
+        var_dump($spec);
```
Meldiron, line 1631: Leftover

src/Appwrite/Platform/Workers/Builds.php
```diff
@@ -413,6 +416,8 @@ protected function buildDeployment(Device $deviceForFunctions, Func $queueForFun
                             source: $source,
                             image: $runtime['image'],
                             version: $version,
+                            cpus: $spec['cpus'] ?? 1,
+                            memory: $spec['memory'] ?? 512,
```
Meldiron, line 420: We shouldn't limit build container to this. We either need ability to configure different specs for build, or we need some high build imits hard-coded.

src/Appwrite/Utopia/Response/Model/Specification.php
```diff
@@ -0,0 +1,64 @@
...
+            'description' => 'Size slug.',
+            'default' => '',
+            'example' => 's-1vcpu-512mb'
+        ]);
+        $this->addRule('plan', [
+            'type' => self::TYPE_STRING,
+            'description' => 'Size plan.',
+            'default' => '',
+            'example' => 'starter'
+        ]);
```
Meldiron, line 42: I dont think this belong to CE version

tests/e2e/General/UsageTest.php
```diff
@@ -795,15 +795,17 @@ public function testFunctionsStats(array $data): void
...
         $this->assertEquals($response['body']['range'], '30d');
         $this->assertIsArray($response['body']['functions']);
         $this->assertIsArray($response['body']['deployments']);
         $this->assertIsArray($response['body']['deploymentsStorage']);
         $this->assertIsArray($response['body']['builds']);
         $this->assertIsArray($response['body']['buildsTime']);
+        $this->assertIsArray($response['body']['buildsMbSeconds']);
         $this->assertIsArray($response['body']['executions']);
         $this->assertIsArray($response['body']['executionsTime']);
+        $this->assertIsArray($response['body']['executionsMbSeconds']);
```
Meldiron, line 808: We increased count by 6, but only added 2 new assertions

src/Appwrite/Utopia/Response/Model/Specification.php
PineappleIOnic: How do we separate this correctly on the cloud and generate a correct SDK?

app/controllers/api/functions.php
PineappleIOnic: Do you know how and where we create hooks in Cloud/

app/controllers/api/functions.php
PineappleIOnic: Added `APP_AUTH_TYPE_ADMIN`, I used only APP_AUTH_TYPE_KEY as that's what we do in the listRuntimes endpoint. Do we want to consider modifying that?

app/controllers/api/functions.php
PineappleIOnic: Remove

tests/e2e/General/UsageTest.php
PineappleIOnic: The other two are for mbSeconds

src/Appwrite/Platform/Workers/Builds.php
PineappleIOnic: Not sure what to do here, make 1024 the minimum or make s-1vcpu-1gb the minimum specification?

app/controllers/api/functions.php
```diff
@@ -807,6 +848,22 @@
             $live = false;
         }
 
+        $spec = Config::getParam('runtime-specifications')[$specification] ?? [];
+
+        if (empty($spec) || empty($spec['memory']) || empty($spec['cpus'])) {
+            throw new Exception(Exception::FUNCTION_INVALID_RUNTIME_SPECIFICATION);
+        }
```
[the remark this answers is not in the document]
PineappleIOnic: We don't actually really need this; the RuntimeSpecification validator should handle this

app/config/errors.php
```diff
@@ -525,6 +525,12 @@
         'code' => 408,
     ],
 
+    Exception::FUNCTION_INVALID_RUNTIME_SPECIFICATION => [
+        'name' => Exception::FUNCTION_INVALID_RUNTIME_SPECIFICATION,
+        'description' => "The requested runtime specification is either unsupported or isn't included in your plan. If your running community edition check the value of the _APP_FUNCTIONS_MEMORY and _APP_FUNCTIONS_CPUS environment variable.",
+        'code' => 400,
+    ],
+
```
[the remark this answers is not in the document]
PineappleIOnic: Just removed them, we aren't using them anyway

app/config/runtimes/specifications.php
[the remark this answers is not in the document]
PineappleIOnic: Added some more specs

app/controllers/api/functions.php
```diff
@@ -442,6 +446,43 @@
...
+    ->inject('plan')
+    ->inject('hooks')
+    ->action(function (Response $response, array $plan, Hooks $hooks) {
+        $allRuntimeSpecs = Config::getParam('runtime-specifications', []);
+
+        $runtimeSpecs = [];
+        foreach ($allRuntimeSpecs as $spec) {
+            $spec['enabled'] = true;
+
+            if (key_exists('runtimeSpecifications', $plan)) {
```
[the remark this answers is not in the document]
PineappleIOnic: Done

app/controllers/api/functions.php
```diff
@@ -807,6 +848,22 @@
...
+        }
+
+        // Enforce Cold Start if spec limits change.
+        if ($function->getAttribute('specification') !== $specification && !empty($function->getAttribute('deployment'))) {
+            $executor = new Executor(App::getEnv('_APP_EXECUTOR_HOST'));
+            try {
+                $executor->deleteRuntime($project->getId(), $function->getAttribute('deployment'));
+            } catch (\Throwable $th) {
+                // Don't throw if the deployment doesn't exist
+            }
```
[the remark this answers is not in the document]
PineappleIOnic: Added a $e->code === 404 check here

app/controllers/api/functions.php
```diff
@@ -1548,6 +1606,7 @@
 
         $version = $function->getAttribute('version', 'v2');
         $runtimes = Config::getParam($version === 'v2' ? 'runtimes-v2' : 'runtimes', []);
+        $spec = Config::getParam('runtime-specifications')[$function->getAttribute('specification', 's-1vcpu-512mb')];
```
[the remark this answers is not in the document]
PineappleIOnic: Done 👍 added `APP_FUNCTION_BASE_SPECIFICATION`, `APP_FUNCTION_BASE_SPECIFICATION_CPUS` and `APP_FUNCTION_BASE_SPECIFICATION_MEMORY`

src/Appwrite/Extend/Exception.php
```diff
@@ -154,6 +154,7 @@ class Exception extends \Exception
     public const FUNCTION_RUNTIME_UNSUPPORTED      = 'function_runtime_unsupported';
     public const FUNCTION_ENTRYPOINT_MISSING       = 'function_entrypoint_missing';
     public const FUNCTION_SYNCHRONOUS_TIMEOUT      = 'function_synchronous_timeout';
+    public const FUNCTION_INVALID_RUNTIME_SPECIFICATION     = 'function_invalid_runtime_specification';
```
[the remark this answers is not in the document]
PineappleIOnic: Removed this error

src/Appwrite/Utopia/Response/Model/Func.php
```diff
@@ -145,6 +145,12 @@ public function __construct()
                 'default' => false,
                 'example' => false,
             ])
+            ->addRule('specification', [
+                'type' => self::TYPE_STRING,
+                'description' => 'Function execution and builds machine specification.',
```
christyjacob4, line 150:
```suggestion
                'description' => 'Machine specification for builds and executions.',
```

src/Appwrite/Platform/Workers/Builds.php
```diff
@@ -400,6 +401,13 @@ protected function buildDeployment(Device $deviceForFunctions, Func $queueForFun
                 $vars[$var->getAttribute('key')] = $var->getAttribute('value', '');
             }
 
+            $cpus = $spec['cpus'] ?? APP_FUNCTION_BASE_SPECIFICATION_CPUS;
+            $memory = max($spec['memory'] ?? APP_FUNCTION_BASE_SPECIFICATION_MEMORY, 1024);
+
+            if ($memory < 1024) {
+                $memory = 1024;
+            }
```
christyjacob4, line 409:
```suggestion
```
not required since you have the max check above

app/controllers/api/functions.php
```diff
@@ -442,6 +446,43 @@
...
+            $spec['enabled'] = true;
+
+            if (key_exists('runtimeSpecifications', $plan)) {
+                $spec['enabled'] = in_array($spec['slug'], $plan['runtimeSpecifications']);
+            }
+
+            // Only add specs that are within the limits set by environment variables
+            if ($spec['cpus'] <= System::getEnv('_APP_FUNCTIONS_CPUS', 1) && $spec['memory'] <= System::getEnv('_APP_FUNCTIONS_MEMORY', 512)) {
+                $runtimeSpecs[] = $spec;
+            }
```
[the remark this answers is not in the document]
abnegate: I think this is consistent with how we handle the runtime list so all good :+1:

app/controllers/api/functions.php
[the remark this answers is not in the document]
abnegate: Let's use `APP_FUNCTION_SPECIFICATION_DEFAULT`, `APP_FUNCTION_CPU_DEFAULT` and `APP_FUNCTION_MEMORY_DEFAULT`   to be consistent with other constants

app/controllers/api/functions.php
```diff
@@ -474,6 +482,42 @@
         ]), Response::MODEL_RUNTIME_LIST);
     });
 
+App::get('/v1/functions/specifications')
+    ->groups(['api', 'functions'])
+    ->desc('Get available function runtime specifications')
+    ->label('scope', 'functions.read')
+    ->label('sdk.auth', [APP_AUTH_TYPE_KEY, APP_AUTH_TYPE_ADMIN])
+    ->label('sdk.namespace', 'functions')
+    ->label('sdk.method', 'getSpecifications')
```
abnegate, line 491: This should be `listSpecifications` for consistency, same for the description

app/controllers/api/functions.php
```diff
@@ -840,6 +889,21 @@
             $live = false;
         }
 
+        $spec = Config::getParam('runtime-specifications')[$specification] ?? [];
+
+        // Enforce Cold Start if spec limits change.
+        if ($function->getAttribute('specification') !== $specification && !empty($function->getAttribute('deployment'))) {
+            $executor = new Executor(App::getEnv('_APP_EXECUTOR_HOST'));
```
abnegate, line 896: Do we have a dependency for this we can inject instead of instantiating inline?

src/Appwrite/Functions/Validator/RuntimeSpecification.php
```diff
@@ -0,0 +1,112 @@
...
+     *
+     * @return array
+     */
+    public function getAllowedSpecifications(): array
+    {
+        $allowedSpecifications = [];
+
+        foreach ($this->specifications as $size => $values) {
+            if ($values['cpus'] <= $this->maxCpus && $values['memory'] <= $this->maxMemory) {
+                if (!empty($this->plan) && key_exists('runtimeSpecifications', $this->plan)) {
```
abnegate, line 38: Let's use `array_key_exists` instead of `key_exists`

abnegate (commented):
Let's also handle the new validator explicitly in 3 places:

- The spec generator `Swagger2.php`
- The spec generator `OpenAPI3.php`
- The GraphQL schema generator `Mapper.php`

src/Appwrite/Functions/Specification.php
```diff
@@ -0,0 +1,16 @@
...
+{
+    public const S_05VCPU_512MB = 's-0.5vcpu-512mb';
+    public const S_1VCPU_512MB = 's-1vcpu-512mb';
+    public const S_1VCPU_1GB = 's-1vcpu-1gb';
+    public const S_2VCPU_2GB = 's-2vcpu-2gb';
+    public const S_2VCPU_4GB = 's-2vcpu-4gb';
+    public const S_4VCPU_4GB = 's-4vcpu-4gb';
+    public const S_4VCPU_8GB = 's-4vcpu-8gb';
+    public const S_8VCPU_4GB = 's-8vcpu-4gb';
+    public const S_8VCPU_8GB = 's-8vcpu-8gb';
```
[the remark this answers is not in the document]
PineappleIOnic: No, not really. We don't plan on offering anything higher than 2 cores and 4GB of ram on cloud. This is more for self-hosted

app/controllers/api/functions.php
PineappleIOnic: No, we do not currently have a dependency for this

app/controllers/api/functions.php
PineappleIOnic: Updated 👍

src/Appwrite/Utopia/Response/Model/Specification.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+
+        $this->addRule('memory', [
+            'type' => self::TYPE_INTEGER,
+            'description' => 'Memory size in MB.',
+            'default' => 0,
+            'example' => 512
+        ]);
+        $this->addRule('cpus', [
+            'type' => self::TYPE_INTEGER,
```
abnegate, line 20: This should be a float now too, let's add a test covering it too

```suggestion
            'type' => self::TYPE_FLOAT,
```

abnegate (approved):

PR #8389, opened 2024-07-13
Add oauth2 details in session when creating session from oauth2 token

### LINKED ISSUE #8206: 🐛 Bug Report: `createOAuth2Token` when using with Github doesn't return `providerUid` or any other provider information
### 👟 Reproduction steps

Calling `createOAuth2Token` on the server side in `Nuxt`

```ts
export default defineEventHandler(async (event) => {
  const config = useRuntimeConfig(event);
  const { account } = useAppwriteAdminClient(event);

  try {
    const redirectUrl = await account.createOAuth2Token(
      'github',
      `${config.public.api.url}/oauth`,
      `${config.public.api.url}/login`,
      ['user']
    );

    return redirectUrl;
  } catch (error) {
    console.log('OAUTH ERROR', error);
  }
});
```

### 👍 Expected behavior

returns `oauth` provider information like `providerUid` like in the `web-sdk`

### 👎 Actual Behavior

Example response:

Screenshot 2024-03-19 at 14 00 10

Related threads:

* https://discord.com/channels/564160730845151244/1244250557213053069
* https://discord.com/channels/564160730845151244/1219632310547382353/1219632310547382353

### 🎲 Appwrite version

### 💻 Operating system

Linux

### 🧱 Your Environment

_No response_

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

Souptik2001:
## What does this PR do?

This PR adds the required oauth2 data, i.e -
- `provider`
- `providerUid`
- `providerAccessToken`
- `providerRefreshToken`
- `providerAccessTokenExpiry`

to the session created using oauth2 token.

When we use the `/v1/account/sessions/oauth2/:provider` endpoint to create a session using Oauth2 login, these details are properly added to the session.

But when we -
- Create a oauth2 token using `/v1/account/tokens/oauth2/:provider`.
- And use the token (i.e basically the secret and userId received) to create a session using `/v1/account/sessions/token` API, these data are not added to the session.

## Test Plan

**Important note: There is a new field added in the `token` collection. So, the below steps will fail if you try these steps on an existing project. You need to create new project. I have added the necessity of this new field in the comments.**

- Create a new project.
- Enable Github Oauth2 provider for your App, and configure with details.
- Now in your browser visit - http://localhost/v1/account/tokens/oauth2/github (please make sure you provide proper `X-Appwrite-Project` request header. _P.S: You can use some browser extension to add this header to your request._
- This will do the oauth2 login (considering you have configured everything correctly) and finally redirect it to `http://localhost/auth/oauth2/success?secret=<secret>&userId=<user_id>#`
- Now in postman create a POST request to - `http://localhost/v1/account/sessions/token` with body with `userId` and `secret` set. And you should see all the provider related details in the response (**which you previously didn't get**) -
Screenshot 2024-07-13 at 9 41 08 AM
(P.S: User tokens revoked)

## Related PRs and Issues

Closes #8206 

## Checklist

### DIFF
--- a/app/config/collections.php
+++ b/app/config/collections.php
@@ -480,6 +480,17 @@ $commonCollections = [
                 'filters' => [],
             ],
             [
+                '$id' => ID::custom('provider'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 128,
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
                 '$id' => ID::custom('type'),
                 'type' => Database::VAR_INTEGER,
                 'format' => '',
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -76,6 +76,11 @@ $createSession = function (string $userId, string $secret, Request $request, Res
         throw new Exception(Exception::USER_INVALID_TOKEN);
     }
 
+    $identity = $dbForProject->findOne('identities', [
+        Query::equal('userInternalId', [$userFromRequest->getInternalId()]),
+        Query::equal('provider', [$verifiedToken->getAttribute('provider')]),
+    ]);
+
     $user->setAttributes($userFromRequest->getArrayCopy());
 
     $duration = $project->getAttribute('auths', [])['duration'] ?? Auth::TOKEN_EXPIRATION_LOGIN_LONG;
@@ -109,6 +114,14 @@ $createSession = function (string $userId, string $secret, Request $request, Res
         $detector->getClient(),
         $detector->getDevice()
     ));
+
+    if ($identity !== false && !$identity->isEmpty()) {
+        $session->setAttribute('provider', $identity->getAttribute('provider'));
+        $session->setAttribute('providerUid', $identity->getAttribute('providerUid'));
+        $session->setAttribute('providerAccessToken', $identity->getAttribute('providerAccessToken'));
+        $session->setAttribute('providerRefreshToken', $identity->getAttribute('providerRefreshToken'));
+        $session->setAttribute('providerAccessTokenExpiry', $identity->getAttribute('providerAccessTokenExpiry'));
+    }
 
     Authorization::setRole(Role::user($user->getId())->toString());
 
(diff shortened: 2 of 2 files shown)


app/config/collections.php
```diff
@@ -479,6 +479,17 @@
...
+                '$id' => ID::custom('provider'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 128,
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
```
Souptik2001, line 492: @stnguyen90 We had to create a new field for the tokens collection named `provider`. I tried to do it without creating a new field but this seems inevitable.

So, the reason for this is -

Here - https://github.com/appwrite/appwrite/pull/8389/files#diff-93771e9bb8e57cbe7d0c0e91e7aea5d85238a362f51caf957049f1d2ccc4d312R81, where we are fetching the identity of this particular user for this particular provider, we currently didn't have the oauth2 provider name.
So, we have multiple identities for a user. So, how would we get the identity for that particular oauth2 provider?
Because of that only we needed to have that information and as per I think it will make sense if this data comes from the token itself.

But I am unsure how the existing projects will adapt this, because else it will give `DOCUMENT_INVALID_STRUCTURE` at the early stage when creating the oauth2 token only.
Is there some mechanism implemented to adapt these collection changes to already created projects/databases?

stnguyen90: @Souptik2001, how about we pull the provider* values from the identity?

https://github.com/appwrite/appwrite/blob/09664c7e60f22ecaa980d3fe4cac612544db0923/app/controllers/api/account.php#L1430-L1451

stnguyen90: @Souptik2001, how about we pull the provider* values from the identity?

https://github.com/appwrite/appwrite/blob/09664c7e60f22ecaa980d3fe4cac612544db0923/app/controllers/api/account.php#L1430-L1451

stnguyen90 (commented):
See https://github.com/appwrite/appwrite/pull/8389#discussion_r1732115437

stnguyen90:
Closing stale PR

PR #8396, opened 2024-07-15
fix: build size limits

loks0n:
- introduces a new environment variable `_APP_FUNCTIONS_BUILD_SIZE_LIMIT`
- adds size limit check in Builds.php and throws an exception if the built deployment size exceeds the limit


app/config/variables.php
```diff
@@ -718,6 +718,15 @@
                 'question' => '',
                 'filter' => ''
             ],
+            [
+                'name' => '_APP_FUNCTIONS_SIZE_LIMIT',
```
christyjacob4, line 722: we already have the same variable above?

loks0n: whoops, fixed

app/config/variables.php
```diff
@@ -711,13 +711,22 @@
...
                 'default' => '30000000',
                 'required' => false,
                 'question' => '',
                 'filter' => ''
             ],
+            [
+                'name' => '_APP_FUNCTIONS_BUILD_SIZE_LIMIT',
+                'description' => 'The maximum size of a built deployment in bytes. The default value is 2GB.',
+                'introduction' => '1.6.0',
+                'default' => '2000000000',
```
stnguyen90, line 725: If this is higher than builds.size:

https://github.com/appwrite/appwrite/blob/4b1bf59091b747c5855c355b6f48a5346b1cbe0c/app/config/collections.php#L3559-L3569

there will still be an error, right?

I think we need to update the collection to make the size bigger than 8 or unsigned.

byawitz: The value of `builds.size` is up to [4294967295](https://github.com/utopia-php/database/blob/c712d1f6c8ec37886a7a1ad4d60a8cd75dec00aa/src/Database/Adapter/SQL.php#L185), so we're good on that part.

stnguyen90: What if someone tries to increase this env var to bigger than that?

loks0n: I've added this as a note to the description of the variable. Users changing this value can refer to this, I don't think there's any more value in doing validation here.

byawitz (approved):

christyjacob4 (approved):

PR #8398, opened 2024-07-16
Handle false appended to string

vermakhushboo:
## What does this PR do?

If you append true to a PHP string, it changes to 1 like 
`"hello" . true = "hello1"`
But if you append false, it just disappears 
`"hello" . false = "hello"`

So we need to handle it separately

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


loks0n (approved):

src/Appwrite/Utopia/Fetch/BodyMultipart.php
```diff
@@ -136,7 +136,7 @@ public function exportBody(): string
             }
 
             $query .= $eol . $eol;
-            $query .= $value . $eol;
+            $query .= $value == false ? 0 . $eol : $value . $eol;
```
loks0n, line 139: Should we do `===` instead of `==`? Not sure if other falsy values like empty array or string should be `0`

docker-compose.yml
```diff
@@ -851,7 +851,7 @@ services:
     hostname: exc1
     <<: *x-logging
     stop_signal: SIGINT
-    image: openruntimes/executor:0.6.1
+    image: openruntimes/executor:0.6.2
```
loks0n, line 854: I think this also needs updating in the docker-compose phtml

vermakhushboo: Nice catch

christyjacob4 (approved):

PR #8403, opened 2024-07-17
Feat 1.6.x migrations and filters

PineappleIOnic:
## What does this PR do?

This PR adds migrations for 1.6.x as well as request and response filters

## Test Plan

Tests have been added for the new request filters and I have locally tested the migrations work as intended.

## Checklist


app/controllers/general.php
```diff
@@ -11,9 +11,11 @@
 use Appwrite\Utopia\Request;
 use Appwrite\Utopia\Request\Filters\V16 as RequestV16;
 use Appwrite\Utopia\Request\Filters\V17 as RequestV17;
+use Appwrite\Utopia\Request\Filters\V18 as RequestV18;
```
Meldiron, line 14: We need to do something about v3 and v4 in functions

Meldiron (approved):

app/controllers/general.php
PineappleIOnic: I added some code to create execution that will automatically update v3 runtimes into v4, we can remove it after we do a migration

abnegate (approved):

src/Appwrite/Migration/Version/V21.php
```diff
@@ -57,16 +62,92 @@ private function migrateCollections(): void
...
+
+                // Add accessedAt attribute
+                $document->setAttribute('accessedAt', DateTime::now());
+                break;
+            case 'functions':
+                // Add scopes attribute
+                $document->setAttribute('scopes', []);
+
+                // Add size attribute
+                $document->setAttribute('size', 's-1vcpu-512m');
```
Meldiron, line 148: Just making sure, this will work fine on self-hosting, right?

src/Appwrite/Utopia/Request/Filters/V18.php
```diff
@@ -0,0 +1,21 @@
...
+use Appwrite\Utopia\Request\Filter;
+
+class V18 extends Filter
+{
+    // Convert 1.5 params to 1.6
+    public function parse(array $content, string $model): array
+    {
+        switch ($model) {
+            case 'account.deleteMfaAuthenticator':
+                unset($content['otp']);
```
Meldiron, line 14: I think there is more.. For example, functions accept scopes now.
Please do full comparison between 1.6.x and 1.5.8 release, and look at changes in all response models, and controllers (->param part)

src/Executor/Executor.php
```diff
@@ -188,6 +194,13 @@ public function createExecution(
 
         $runtimeId = "$projectId-$deploymentId";
         $route = '/runtimes/' . $runtimeId . '/execution';
+
+
+        // Remove after migration
```
Meldiron, line 199: Let's add this comment here too:

CleanShot 2024-07-22 at 12 51 38@2x


Its also something that can go away after migration

src/Appwrite/Utopia/Request/Filters/V18.php
Meldiron: Ignore that, they are optional

Meldiron (approved):

PR #8416, opened 2024-07-19
Fix realtime permission change

stnguyen90:
## What does this PR do?

1. add userId to queueForEvents so that the [permissionsChanged](https://github.com/appwrite/appwrite/blob/483ef3c00bcbda150102070e3a15c01692b21626/app/realtime.php#L371-L389) handler is triggered
2. unsubscribe before resubscribing on permission change to ensure old subscriptions are removed

## Test Plan

Manually tested by:

1. granting a team access to a collection
2. adding a user to the team
3. subscribing
4. updating the document to verify the user receives the event
5. removing the user from the team
6. updating hte document again to verify the user doesn't receive the event

image

## Related PRs and Issues

- None

## Checklist

### DIFF
--- a/app/realtime.php
+++ b/app/realtime.php
@@ -380,8 +380,10 @@ $server->onWorkerStart(function (int $workerId) use ($server, $register, $stats,
                         $user = $database->getDocument('users', $userId);
 
                         $roles = Auth::getRoles($user);
+                        $channels = $realtime->connections[$connection]['channels'];
 
-                        $realtime->subscribe($projectId, $connection, $roles, $realtime->connections[$connection]['channels']);
+                        $realtime->unsubscribe($connection);
+                        $realtime->subscribe($projectId, $connection, $roles, $channels);
 
                         $register->get('pools')->reclaim();
                     }
--- a/app/controllers/api/teams.php
+++ b/app/controllers/api/teams.php
@@ -668,6 +668,7 @@ App::post('/v1/teams/:teamId/memberships')
         }
 
         $queueForEvents
+            ->setParam('userId', $invitee->getId())
             ->setParam('teamId', $team->getId())
             ->setParam('membershipId', $membership->getId())
         ;
@@ -901,6 +902,7 @@ App::patch('/v1/teams/:teamId/memberships/:membershipId')
         $dbForProject->purgeCachedDocument('users', $profile->getId());
 
         $queueForEvents
+            ->setParam('userId', $profile->getId())
             ->setParam('teamId', $team->getId())
             ->setParam('membershipId', $membership->getId());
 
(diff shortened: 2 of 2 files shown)


app/realtime.php
```diff
@@ -380,8 +380,10 @@ function getCache(): Cache
                         $user = $database->getDocument('users', $userId);
 
                         $roles = Auth::getRoles($user);
+                        $channels = $realtime->connections[$connection]['channels'];
 
-                        $realtime->subscribe($projectId, $connection, $roles, $realtime->connections[$connection]['channels']);
+                        $realtime->unsubscribe($connection);
+                        $realtime->subscribe($projectId, $connection, $roles, $channels);
```
TorstenDittmann, line 386: would it makes sense to call `unsubscribe($connection)` inside of `subscribe($projectId, $connection, $roles, $channels)` that way we can prevent such issues in the first place

stnguyen90: 1. I'm wary about unsubscribing for every subscribe when it's not needed as it would add extra processing
2. The `subscribe()` method might get messy trying to handle old and new channels:

```diff
diff --git a/src/Appwrite/Messaging/Adapter/Realtime.php b/src/Appwrite/Messaging/Adapter/Realtime.php
index 55d8db292..fe89061d2 100644
--- a/src/Appwrite/Messaging/Adapter/Realtime.php
+++ b/src/Appwrite/Messaging/Adapter/Realtime.php
@@ -45,10 +45,14 @@ class Realtime extends Adapter
      * @param array $channels
      * @return void
      */
-    public function subscribe(string $projectId, mixed $identifier, array $roles, array $channels): void
+    public function subscribe(string $projectId, mixed $identifier, array $roles, array $newChannels): void
     {
         if (!isset($this->subscriptions[$projectId])) { // Init Project
             $this->subscriptions[$projectId] = [];
+        } else {
+            $oldChannels = $this->connections[$identifier]['channels'] ?? [];
+            $channels = array_unique(array_merge($oldChannels, $newChannels));
+            $this->unsubscribe($identifier);
         }
 
         foreach ($roles as $role) {
```

TorstenDittmann: I mean we can keep the `unsubscribe` method - but call it basically. If not calling unsubscribe before subscribing again, I dont think it will hurt.

stnguyen90: I wasn't suggesting removing the `unsubscribe` method. I was worried about adding extra processing of iterating over every role/channel when it's not needed considering realtime needs to be super fast.

Also, it's messy to handle both old and new channels in the subscribe method. Please see the diff above.

TorstenDittmann (approved):

PR #8422, opened 2024-07-21
feat: address review comments

christyjacob4:
## What does this PR do?

* Check for duplicate mock numbers
* Ensure OTP for mock numbers can only consist of digits

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


src/Appwrite/Auth/Validator/MockNumber.php
```diff
@@ -47,8 +47,8 @@ public function isValid($value): bool
         }
 
         $otp = new Text(6, 6);
-        if (!$otp->isValid($value['otp'])) {
-            $this->message = 'OTP must be a valid string and exactly 6 characters.';
+        if (!$otp->isValid($value['otp']) || !\ctype_digit($value['otp'])) {
+            $this->message = 'Invalid OTP. Please make sure the OTP is a 6 digit number';
```
Meldiron, line 51: Third parameter in text validator is `array $allowList = [])`, you can pass `Text::NUMBERS` to only allow numbers.

app/controllers/api/projects.php
```diff
@@ -882,6 +882,14 @@
             throw new Exception(Exception::PROJECT_NOT_FOUND);
         }
 
+        $uniqueNumbers = [];
+        foreach ($numbers as $number) {
+            if (isset($uniqueNumbers[$number['phone']])) {
+                throw new Exception(Exception::GENERAL_BAD_REQUEST, 'Duplicate phone numbers are not allowed.');
```
Meldiron, line 888: I think it would be better to have exception defined for this. By doing that we give it custom exception type to which developers can then react properly.

Meldiron (commented):
LGTM, jsut a few code-quality comments.

app/controllers/api/projects.php
christyjacob4: This is actually something the validator should have caught. But our validator system doesn't support this. 
And if an error is thrown in the validator we usually throw a general 400 Bad Request error.

Meldiron (approved):

PR #8434, opened 2024-07-25
Fix: function redeployments

Meldiron:
## What does this PR do?

Fixes following flows:

1. Create deployment A -> redeploy (B) -> delete A -> redeploy B (fails, no file found in build logs)
2. Create deployment A -> redeploy (B) -> delete A -> redeploy B -> delete B (server error, cannot delete file)
3. Push to git (deployment A) -> download A (fails, deployment not found)

## Test Plan

- [x] Manual QA

CleanShot 2024-07-25 at 10 38 14@2x

CleanShot 2024-07-25 at 10 38 37@2x

## Related PRs and Issues

x

## Checklist

### DIFF
--- a/app/controllers/api/functions.php
+++ b/app/controllers/api/functions.php
@@ -1459,7 +1459,8 @@ App::post('/v1/functions/:functionId/deployments/:deploymentId/build')
     ->inject('project')
     ->inject('queueForEvents')
     ->inject('queueForBuilds')
-    ->action(function (string $functionId, string $deploymentId, string $buildId, Request $request, Response $response, Database $dbForProject, Document $project, Event $queueForEvents, Build $queueForBuilds) {
+    ->inject('deviceForFunctions')
+    ->action(function (string $functionId, string $deploymentId, string $buildId, Request $request, Response $response, Database $dbForProject, Document $project, Event $queueForEvents, Build $queueForBuilds, Device $deviceForFunctions) {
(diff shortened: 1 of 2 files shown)


vermakhushboo (approved):

app/controllers/api/functions.php
```diff
@@ -1471,13 +1472,23 @@
...
 
+        $path = $deployment->getAttribute('path');
+        if(empty($path) || !$deviceForFunctions->exists($path)) {
+            throw new Exception(Exception::DEPLOYMENT_NOT_FOUND);
+        }
+
         $deploymentId = ID::unique();
 
+        $destination = $deviceForFunctions->getPath($deploymentId . '.' . \pathinfo('code.tar.gz', PATHINFO_EXTENSION));
+        $deviceForFunctions->transfer($path, $destination, $deviceForFunctions);
```
christyjacob4, line 1483: PR looks okay.. But I need to understand why we use transfer and if it can be done without the overhead of moving files around thereby slowing down the endpoint significantly depending on the size of the deployment

Meldiron: We need to copy deployment, so each has different path for their source. Because if they share a source, and then you delete one of the deployments, then the other one doesn't have source (you cannot download, or redeploy anymore)

christyjacob4 (approved):

PR #8445, opened 2024-07-29
Fix: template tags

Meldiron:
## What does this PR do?

Allows to use tags (0.1.x) to allow non-breaking changes to templates in future.

This is temporary fix and should not be merged to 1.6.x. In 1.6.x, only tag names will be supported. This is only a patch for 1.5.x version to prevent breaking changes.

## Test Plan

- [x] Manual QA

With "Main" branch

CleanShot 2024-07-29 at 20 11 13@2x

With "0.1.0" branch:

CleanShot 2024-07-29 at 20 12 55@2x

With "0.1.*" branch:
CleanShot 2024-07-29 at 21 00 59@2x

## Related PRs and Issues

x

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -69,7 +69,7 @@
         "utopia-php/storage": "0.18.*",
         "utopia-php/swoole": "0.8.*",
         "utopia-php/system": "0.8.*",
-        "utopia-php/vcs": "0.8.*",
+        "utopia-php/vcs": "dev-fix-wildcard-tags as 0.8.99",
         "utopia-php/websocket": "0.1.*",
         "matomo/device-detector": "6.1.*",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -69,7 +69,7 @@
         "utopia-php/storage": "0.18.*",
         "utopia-php/swoole": "0.8.*",
         "utopia-php/system": "0.8.*",
-        "utopia-php/vcs": "0.6.*",
+        "utopia-php/vcs": "dev-fix-wildcard-tags as 0.8.99",
```
christyjacob4, line 72: reminder to update the tag

christyjacob4 (approved):

PR #8450, opened 2024-07-30
Db project by region

shimonewman:


app/controllers/api/projects.php
```diff
@@ -119,6 +119,11 @@
         if ($index !== false) {
             $dsn = $databases[$index];
         } else {
+
+            $databases = array_filter($databases, function($value, $region) {
+                return str_contains($value, $region);
+            });
```
christyjacob4, line 125: this will be a problem on self hosted, because the region there is `default` and I dont think the database connection string in self hosted has `default` in its name?

app/init.php
```diff
@@ -1443,23 +1423,28 @@ function (mixed $value) {
...
-}, ['project']);
+App::setResource('deviceForBuilds', function ($project, $connectionStorage) {
+    return getDevice(APP_STORAGE_BUILDS.'/app-'.$project->getId(), $connectionStorage);
+}, ['project', 'connectionStorage']);
 
-function getDevice($root): Device
-{
-    $connection = System::getEnv('_APP_CONNECTIONS_STORAGE', '');
+App::setResource('connectionStorage', function () {
+    return '';
```
christyjacob4, line 1439: return System::getEnv('_APP_CONNECTIONS_STORAGE', '');

app/init.php
```diff
@@ -1443,23 +1423,28 @@ function (mixed $value) {
...
+});
 
-    if (!empty($connection)) {
+function getDevice(string $root, string $connectionStorage = ''): Device
+{
+    $connectionStorage = ! empty($connectionStorage) ? $connectionStorage : System::getEnv('_APP_CONNECTIONS_STORAGE', '');
```
christyjacob4, line 1444:
```suggestion
```

app/init.php
```diff
@@ -1443,23 +1423,28 @@ function (mixed $value) {
...
-    return getDevice(APP_STORAGE_BUILDS . '/app-' . $project->getId());
-}, ['project']);
+App::setResource('deviceForBuilds', function ($project, $connectionStorage) {
+    return getDevice(APP_STORAGE_BUILDS.'/app-'.$project->getId(), $connectionStorage);
+}, ['project', 'connectionStorage']);
 
-function getDevice($root): Device
-{
-    $connection = System::getEnv('_APP_CONNECTIONS_STORAGE', '');
+App::setResource('connectionStorage', function () {
```
christyjacob4, line 1438:
```suggestion
App::setResource('connectionString', function () {
```

PR #8460, opened 2024-08-02
Implement MFA Emails for authenticator creation and deletion

PineappleIOnic:
## What does this PR do?

This PR implements emails when authenticators are created and deleted in Appwrite. This has been tested on both the console and the user-facing API to ensure the templates work.

## Test Plan

No tests have been added for this

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/config/locale/templates/email-mfa-altered.tpl
stnguyen90, line 1: Don't we still need the hello, thanks, and signature?

app/controllers/api/account.php
```diff
@@ -3703,6 +3706,99 @@
...
+        $customTemplate = $project->getAttribute('templates', [])['email.mfaCreated-' . $locale->default] ?? [];
+
+        // Appwrite theming for console
+        if ($project->getInternalId() === 'console') {
+            $message = Template::fromFile(__DIR__ . '/../../config/locale/templates/email-base-styled.tpl');
+            $body = Template::fromFile(__DIR__ . '/../../config/locale/templates/email-mfa-altered.tpl');
+
+            $body
+                ->setParam('{{buttonText}}', $locale->getText("emails.mfaCreated.buttonText"))
+                ->setParam('{{redirect}}', 'http://' . App::getEnv('_APP_DOMAIN') . '/console/account')
```
stnguyen90, line 3720: Might be good to be consistent and use the same approach we do in other places like:

https://github.com/appwrite/appwrite/blob/5ddc705099a7751b12432f37d1b0f0381c9aecfa/app/controllers/api/account.php#L1072-L1073

app/config/locale/translations/en.json
```diff
@@ -31,6 +31,20 @@
     "emails.mfaChallenge.clientInfo": "This verification code was requested using {{b}}{{agentClient}}{{/b}} on {{b}}{{agentDevice}}{{/b}} {{b}}{{agentOs}}{{/b}}. If you didn't request the verification code, you can safely ignore this email.",
     "emails.mfaChallenge.thanks": "Thanks,",
     "emails.mfaChallenge.signature": "{{project}} team",
+    "emails.mfaCreated.subject": "Two-factor authentication is enabled in your {{project}} account",
+    "emails.mfaCreated.hello": "Hello {{user}}",
+    "emails.mfaCreated.body": "Two-factor authentication has been successfully enabled for your {{project}} account.<br/><br/>Please make sure to store your security codes safely, so you can use them if you lose or can’t access your mobile device. Each recovery code can only be used once, but you can re-generate a new set of 6 codes at any time.",
```
stnguyen90, line 36: I'm not sure about adding the recovery code part because those are separate API calls that happen after. Also, if you delete and add, you may not go through the flow of generating the codes again.

app/config/locale/translations/en.json
```diff
@@ -31,6 +31,20 @@
...
+    "emails.mfaCreated.footer": "If you didn’t enable two-factor authentication, please contact us immediately.",
+    "emails.mfaCreated.thanks": "Thanks,",
```
stnguyen90, line 38: Probably best to leave the comma out.

app/config/locale/translations/en.json
```diff
@@ -31,6 +31,20 @@
...
+    "emails.mfaCreated.signature": "{{project}} team",
+    "emails.mfaCreated.buttonText": "Visit your account",
+    "emails.mfaRemoved.subject": "Two-factor authentication is disabled in your {{project}} account",
```
stnguyen90, line 41: Two factor auth can still be enabled if they have other factors/methods enabled.

app/controllers/api/account.php
```diff
@@ -3703,6 +3706,99 @@
...
+                ->setSmtpSenderEmail($senderEmail)
+                ->setSmtpSenderName($senderName);
+        }
+
+        $emailVariables = [
+            'owner' => $user->getAttribute('name'),
+            'direction' => $locale->getText('settings.direction'),
+            'user' => $user->getAttribute('name'),
+            'project' => $project->getAttribute('name')
+        ];
```
stnguyen90, line 3791: redirect is required in email-inner-base.tpl 😬

PineappleIOnic:
@stnguyen90 Going to move this to the update MFA endpoint and maybe alter the emails to trigger on
- MFA Enabled
- MFA Disabled
- MFA Authenticator Added
- MFA Authenticator Remove
- MFA Email Enabled
- MFA Email Disabled

app/controllers/api/account.php
PineappleIOnic: We shouldn't be using redirect should we? We don't know where the user wants to redirect. It's up to them to specify it in their template otherwise it'll look like so:

image

app/config/locale/translations/en.json
PineappleIOnic: Changed it to use the updateMFA endpoint instead of the create/delete authenticator ones

app/config/locale/translations/en.json
PineappleIOnic: Update

app/config/locale/translations/en.json
PineappleIOnic: Not sure what to do here, this was the copy that was provided for cloud. Should we possibly split these into console/user facing?

app/controllers/api/account.php
PineappleIOnic: Updated 👍

app/config/locale/templates/email-mfa-altered.tpl
PineappleIOnic: In the Figma file I was sent it didn't have a hello, thanks or a signature. I could add them if you like

app/controllers/api/account.php
```diff
@@ -166,6 +166,100 @@
     $response->dynamic($session, Response::MODEL_SESSION);
 };
 
+$sendMFAAlert = function (Document $project, Document $user, Locale $locale, Mail $queueForMails, string $type, string $redirect) {
```
abnegate, line 169: If we only do this in one route, let's do it inline instead of declaring a global function var

app/controllers/api/account.php
```diff
@@ -166,6 +166,100 @@
...
+            }
+
+            $body = $customTemplate['message'] ?? '';
+            $subject = $customTemplate['subject'] ?? $subject;
+        }
+
+        $queueForMails
+            ->setSmtpReplyTo($replyTo)
+            ->setSmtpSenderEmail($senderEmail)
+            ->setSmtpSenderName($senderName);
```
abnegate, line 243: Let use null coalescing here to avoid warnings if the variables are undefined

abnegate (approved):

app/config/locale/translations/en.json
stnguyen90: Probably, yes.

app/config/locale/translations/en.json
stnguyen90: What do you mean? The email code still looksl ike it's in the delete authenticator endpoint.

app/controllers/api/account.php
```diff
@@ -3876,6 +3975,99 @@
...
+        $customTemplate = $project->getAttribute('templates', [])['email.mfaRemoved-' . $locale->default] ?? [];
+
+        // Appwrite theming for console
+        if ($project->getInternalId() === 'console') {
+            $message = Template::fromFile(__DIR__ . '/../../config/locale/templates/email-base-styled.tpl');
+            $body = Template::fromFile(__DIR__ . '/../../config/locale/templates/email-mfa-altered.tpl');
+
+            $body
+                ->setParam('{{buttonText}}', $locale->getText("emails.mfaRemoved.buttonText"))
+                ->setParam('{{redirect}}', 'http://' . App::getEnv('_APP_DOMAIN') . '/console/account')
```
stnguyen90: We typically use the hostname from the request instead of the `_APP_DOMAIN` env var. For example:

https://github.com/appwrite/appwrite/blob/90bc303c7cbb683c92ae499a54954cea5a4b43fc/app/controllers/api/account.php#L1011

Should we do the same here?

stnguyen90 (commented):
Can you add screenshots in the PR description of the emails for both created and deleted authenticators for:

1. console user
2. project user with default template
3. project user with custom template

Also, does the console need to be updated for the custom template?

app/config/locale/translations/en.json
```diff
@@ -40,6 +40,20 @@
     "emails.mfaChallenge.clientInfo": "This verification code was requested using {{b}}{{agentClient}}{{/b}} on {{b}}{{agentDevice}}{{/b}} {{b}}{{agentOs}}{{/b}}. If you didn't request the verification code, you can safely ignore this email.",
     "emails.mfaChallenge.thanks": "Thanks,",
     "emails.mfaChallenge.signature": "{{project}} team",
+    "emails.mfaEnabled.subject": "Two-factor authentication is enabled for your {{project}} account",
+    "emails.mfaEnabled.hello": "Hello {{user}}",
+    "emails.mfaEnabled.body": "Two-factor authentication has been successfully enabled for your {{project}} account.<br/><br/>Please make sure to store your security codes safely, so you can use them if you lose or can’t access your mobile device. Each recovery code can only be used once, but you can re-generate a new set of 6 codes at any time.",
```
stnguyen90, line 45: I'm not sure about adding the recovery code part because those are separate API calls that happen after. Also, if you delete and add, you may not go through the flow of generating the codes again.

app/controllers/api/account.php
stnguyen90: > We shouldn't be using redirect should we? We don't know where the user wants to redirect.

Correct.

> It's up to them to specify it in their template

But developers on the free plan can't set a custom template.

PineappleIOnic: Split them :+1:

app/config/locale/translations/en.json
PineappleIOnic: Split them into it's own file, which should add the bedrock for future console specific translations as well

stnguyen90:
> Can you add screenshots in the PR description of the emails for both created and deleted authenticators for:
> 
> 1. console user
> 2. project user with default template
> 3. project user with custom template

@PineappleIOnic, looks like this only sends emails when MFA is enabled or disabled, so please test and add screenshots for

1. console user
2. project user with default template
3. project user with custom template

for both

1. default locale
2. some other locale

app/controllers/api/account.php
```diff
@@ -3648,12 +3648,119 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
+                    if (!empty($customTemplate['senderName'])) {
+                        $senderName = $customTemplate['senderName'];
+                    }
+                    if (!empty($customTemplate['replyTo'])) {
+                        $replyTo = $customTemplate['replyTo'];
+                    }
+
+                    $body = $customTemplate['message'] ?? '';
+                    $subject = $customTemplate['subject'] ?? $subject;
+                }
```
stnguyen90, line 3740: When a developer tries to customize the template for their project, the Console makes a `GET /v1/projects/acme-app/templates/email/mfaEnabled/en` API call to fetch the default template. It looks like this is returning the Console template and not the Project template:

```json
{
    "type": "mfaEnabled",
    "locale": "en",
    "message": "<p>Hello {{user}},</p>\n<p>Two-factor authentication has been successfully enabled for your {{project}} account.</p>\n<p><a href=\"{{redirect}}\" target=\"_blank\">{{redirect}}</a></p>\n<p>If you didn’t enable two-factor authentication, please contact us immediately.</p>\n<p style=\"margin-bottom: 32px\">\n    Thanks,\n    <br/>\n    {{project}} team\n</p>",
    "senderName": "",
    "senderEmail": "",
    "replyTo": "",
    "subject": "Two-factor authentication is enabled for your {{project}} account"
}
```

Also, getting a non-english locale like `GET /v1/projects/acme-app/templates/email/mfaEnabled/es` returns a template some of the variables:

```json
{
    "type": "mfaEnabled",
    "locale": "es",
    "message": "<p>{{emails.mfaEnabled.hello}},</p>\n<p>{{emails.mfaEnabled.body}}</p>\n<p><a href=\"{{redirect}}\" target=\"_blank\">{{redirect}}</a></p>\n<p>{{emails.mfaEnabled.footer}}</p>\n<p style=\"margin-bottom: 32px\">\n    {{emails.mfaEnabled.thanks}},\n    <br/>\n    {{emails.mfaEnabled.signature}}\n</p>",
    "senderName": "",
    "senderEmail": "",
    "replyTo": "",
    "subject": "{{emails.mfaEnabled.subject}}"
}
```

I don't think we need to work on this now, but it would be good to create an issue or something to track the work.

stnguyen90 (changes requested):

app/controllers/api/account.php
```diff
@@ -3648,12 +3648,119 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
+
+            $subject = $locale->getText("emails.{$type}.subject");
+            $customTemplate = $project->getAttribute('templates', [])["email.{$type}-" . $locale->default] ?? [];
+
+            // Appwrite Theming for Console
+            if ($project->getInternalId() === 'console') {
+                $message = Template::fromFile(__DIR__ . '/../../config/locale/templates/email-base-styled.tpl');
+                $body = Template::fromFile(__DIR__ . '/../../config/locale/templates/email-mfa-altered-console.tpl');
+
+                $buttonText = $locale->getText("emails.{$type}.buttonText", 'console-'.$locale->default) ?? $locale->getText("emails.{$type}.buttonText");
```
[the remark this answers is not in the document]
PineappleIOnic: Ah I see, I got confused because we have $default and $this->default, that second one should probably be `$this->defaultLanguage` to remove confusion or even `$this->currentLanguage`

app/controllers/api/account.php
```diff
@@ -3648,12 +3648,119 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
+                }
+
+                $queueForMails
+                    ->setSmtpReplyTo($replyTo ?? '')
+                    ->setSmtpSenderEmail($senderEmail ?? '')
+                    ->setSmtpSenderName($senderName ?? '');
+            }
+
+            $emailVariables = [
+                'owner' => $user->getAttribute('name'),
```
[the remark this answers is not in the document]
PineappleIOnic: There isn't one. Removing

PineappleIOnic:
@stnguyen90 

Console:
image
image

Stock, No custom template, user-facing:
image
image

Custom Template:
image
image

PineappleIOnic:
The redirection is within the default template, we need to figure out what exactly we're planning on doing with that to move forward

stnguyen90:
Closing stale PR

Issue #8339, opened 2024-06-28
🐛 Bug Report: Executions count and bandwidth not incremented when executing function domain


cn0ss:
### 👟 Reproduction steps

When I trigger a function using the custom generated url: `http://667ddc33e6ea2cfd945c.{url}/` it is **not** getting registered as usage. Not inside the function, not on the dashboard. But when triggering `https://{url}/v1/functions/667ddc33002502a9b799/executions` it is.

### 👍 Expected behavior

It should register the usage of the function both ways. Ideally if custom domain is connected the other way can be blocked (minimizing attack surface and reducing entry points).

### 👎 Actual Behavior

It actually only registers when triggered this: `https://{url}/v1/functions/667ddc33002502a9b799/executions` way.

### 🎲 Appwrite version

Version 1.5.x

### 💻 Operating system

Linux

### 🧱 Your Environment

Cloudflare is used for the domain. It is hosted on Hetzner VPS CPX21.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

stnguyen90:
@cn0ss, thanks for creating this issue! 🙏🏼 

Related discord thread: https://discord.com/channels/564160730845151244/1254814557353017344

stnguyen90:
Reopening because it seems bandwidth and request count still aren't incrementing.

PR #8463, opened 2024-08-02
Fix queueForUsage not triggering for domain executions

PineappleIOnic:
## What does this PR do?

Fix domain executions not being tracked in metrics because the event wasn't being triggered.

Fixes https://github.com/appwrite/appwrite/issues/8339

## Checklist


abnegate (commented):
Are we able to add a test for it?

abnegate (approved):

Issue #8328, opened 2024-06-26
📚 Documentation: All vars `console.log` has to be `result` instead of `response`


xuelink:
### 💭 Description

Check the following pages
- https://appwrite.io/docs/references/cloud/client-web/storage#listFiles
- https://appwrite.io/docs/references/cloud/client-web/storage#createFile
- https://appwrite.io/docs/references/cloud/client-web/storage#getFile
- https://appwrite.io/docs/references/cloud/client-web/storage#updateFile
- https://appwrite.io/docs/references/cloud/client-web/storage#deleteFile
- https://appwrite.io/docs/references/cloud/client-react-native/storage#listFiles
- https://appwrite.io/docs/references/cloud/client-react-native/storage#createFile
- https://appwrite.io/docs/references/cloud/client-react-native/storage#getFile
- https://appwrite.io/docs/references/cloud/client-react-native/storage#updateFile
- https://appwrite.io/docs/references/cloud/client-react-native/storage#deleteFile

In these examples that i found in docs, variables has to be updated. 

<img width="627" alt="Screenshot 2024-06-26 at 3 22 36 AM" src="https://github.com/appwrite/appwrite/assets/3280475/75396a72-298b-4845-891b-6ff52672fbe2">

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

stnguyen90:
@xuelink, thanks for raising this! 🙏🏼 Looks like we need to fix our templates:

* https://github.com/appwrite/sdk-generator/blob/d382c1f14607cef8819b70d4a1fb3cf6b0880255/templates/web/docs/example.md.twig#L25
* https://github.com/appwrite/sdk-generator/blob/d382c1f14607cef8819b70d4a1fb3cf6b0880255/templates/react-native/docs/example.md.twig#L25

kunalArya1:
Hi @stnguyen90,

Thank you for identifying the issue and providing the necessary information. I want to contribute to fixing the templates for the Appwrite documentation.

I have reviewed the relevant template files:

[Web Docs Template](https://github.com/appwrite/sdk-generator/blob/d382c1f14607cef8819b70d4a1fb3cf6b0880255/templates/web/docs/example.md.twig#L25)
[...]

PR #8475, opened 2024-08-04
Fix create.md

Buzut:
logging the actual API return res rather than an undefined variable. Note that this error is present all around the other exemples.


stnguyen90:
Thanks for raising this PR, but this is being handled as part of https://github.com/appwrite/appwrite/issues/8328

PR #8477, opened 2024-08-05
Add multipart support

vermakhushboo:
## What does this PR do?

Support and test binary request and response 

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/functions.php
```diff
@@ -1614,12 +1618,33 @@
...
+        foreach ($assocParams as $assocParam) {
+            if (!empty('headers') && !is_array($$assocParam)) {
+                $$assocParam = \json_decode($$assocParam, true);
+            }
+        }
+
+        // 'body' validator
+        if (\strlen($body) > 20971520) {
+            throw new Exception("Parameter body can be only up to 20MB in size.");
+        }
```
Meldiron, line 1640: No need

app/controllers/api/functions.php
```diff
@@ -1603,9 +1606,10 @@
     ->param('async', false, new Boolean(), 'Execute code in the background. Default value is false.', true)
```
Meldiron, line 1606: Above this we have `body` param, unlimited. I think we should limit it to 20MB

app/controllers/api/functions.php
```diff
@@ -1614,12 +1618,33 @@
...
+    ->action(function (string $functionId, string $body, bool $async, string $path, string $method, mixed $headers, ?string $scheduledAt, Response $response, Request $request, Document $project, Database $dbForProject, Database $dbForConsole, Document $user, Event $queueForEvents, Usage $queueForUsage, Func $queueForFunctions, Reader $geodb) {
 
         if(!$async && !is_null($scheduledAt)) {
             throw new Exception(Exception::GENERAL_BAD_REQUEST, 'Scheduled executions must run asynchronously. Set scheduledAt to a future date, or set async to true.');
         }
 
+        /**
+         * @var array<string, mixed> $headers
+         */
+        $assocParams = ['headers'];
```
Meldiron, line 1630: We are missing other types of params - boolean for async.

app/controllers/api/functions.php
```diff
@@ -1926,9 +1951,38 @@
         $execution->setAttribute('responseBody', $executionResponse['body'] ?? '');
         $execution->setAttribute('responseHeaders', $headers);
 
-        $response
-            ->setStatusCode(Response::STATUS_CODE_CREATED)
-            ->dynamic($execution, Response::MODEL_EXECUTION);
+        $acceptTypes = \explode(', ', $request->getHeader('accept', 'multipart/form-data'));
+        $isJson = false;
```
Meldiron, line 1955: As I said before, here we need opposite logic. Notice this is using JSON only if you provided JSON accept header. And anything else is multipart.

We need JSON by default, and only if you specified multipart, then we do multipart.

app/controllers/api/functions.php
```diff
@@ -1926,9 +1951,38 @@
...
+            if (\str_starts_with($acceptType, 'application/json') || \str_starts_with($acceptType, 'application/*')) {
+                $isJson = true;
+                break;
+            }
+        }
+
+        if ($isJson) {
+            $executionString = \json_encode($execution, JSON_UNESCAPED_UNICODE);
+            if (!$executionString) {
+                throw new Exception('Execution resulted in binary response, but JSON response does not allow binaries. Use "Accept: multipart/form-data" header to support binaries.', 400);
```
Meldiron, line 1967: Let's have that ensures this specific exception message.

app/controllers/api/functions.php
```diff
@@ -1926,9 +1951,38 @@
...
+            }
+
+            $response
+                ->setStatusCode(Response::STATUS_CODE_OK)
+                ->addHeader('content-type', 'application/json')
+                ->send($executionString);
```
Meldiron, line 1973: We cant do this, because response filters don't apply anymore. We need to keep using response filters.

app/controllers/api/functions.php
```diff
@@ -1926,9 +1951,38 @@
...
+            // Multipart form data response
+            $multipart = new BodyMultipart();
+            foreach ($execution as $key => $value) {
+                $multipart->setPart($key, $value);
+            }
+
+            $response
+                ->setStatusCode(Response::STATUS_CODE_CREATED)
+                ->addHeader('content-type', $multipart->exportHeader())
+                ->send($multipart->exportBody());
```
Meldiron, line 1984: We cant do this, because response filters don't apply anymore. We need to keep using response filters.

src/Appwrite/Functions/Validator/Headers.php
```diff
@@ -0,0 +1,90 @@
...
+namespace Appwrite\Functions\Validator;
+
+use Utopia\Validator;
+
+/**
+ * Headers.
+ *
+ * Validates user provided headers
+ */
+class Headers extends Validator
```
Meldiron, line 12: Let's have unit tests for this validator

src/Appwrite/Functions/Validator/Headers.php
```diff
@@ -0,0 +1,90 @@
...
+                        return false;
+                    }
+                    // Check for x-appwrite- prefix
+                    if (0 === strpos($key, 'x-appwrite-')) {
+                        return false;
+                    }
+                }
+            }
+        }
+        return \json_last_error() == JSON_ERROR_NONE;
```
Meldiron, line 64: Notice this, and like 50. This code doesn't make sense. Both cant be true at the same time

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -1316,6 +1316,197 @@ public function testCreateCustomExecution(string $folder, string $name, string $
...
+
+        // Wait a little for activation to finish
+        sleep(5);
+
+        $execution = $this->client->call(Client::METHOD_POST, '/functions/' . $functionId . '/executions', array_merge([
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'accept' => 'multipart/form-data',
+        ], $this->getHeaders()), [
+            'body' => null,
+        ]);
```
Meldiron, line 1386: Let's ensure sresponse content-type is multipart

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -1316,6 +1316,197 @@ public function testCreateCustomExecution(string $folder, string $name, string $
...
+        $this->assertEquals(255, $bytes['byte3']);
+
+        $execution = $this->client->call(Client::METHOD_POST, '/functions/' . $functionId . '/executions', array_merge([
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'accept' => 'application/json',
+        ], $this->getHeaders()), [
+            'body' => null,
+        ]);
+
+        $this->assertEquals(500, $execution['headers']['status-code']);
```
Meldiron, line 1401: This should be 400. otherwise developer never sees the error message

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -1316,6 +1316,197 @@ public function testCreateCustomExecution(string $folder, string $name, string $
...
+            'content-type' => 'multipart/form-data',
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'accept' => 'application/json',
+        ], $this->getHeaders()), [
+            'body' => $bytes,
+        ], false);
+
+        $executionBody = json_decode($execution['body'], true);
+
+        $this->assertEquals(200, $execution['headers']['status-code']);
```
Meldiron, line 1497: Let's ensure response content-type header is json

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -1316,6 +1316,197 @@ public function testCreateCustomExecution(string $folder, string $name, string $
...
+        $this->assertEquals(\md5($bytes), $executionBody['responseBody']);
+
```
Meldiron, line 1499: Let's add test where we try to send bytes with JSON content-type. We should expect some kind of error (either 4XX, or md5 mismatch)

app/controllers/api/functions.php
vermakhushboo: Added a test to check for this

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
vermakhushboo: md5 mismatch is happening, added a test for it

src/Appwrite/Utopia/Response.php
```diff
@@ -556,7 +559,11 @@ public function dynamic(Document $document, string $model): void
 
         switch ($this->getContentType()) {
             case self::CONTENT_TYPE_JSON:
-                $this->json(!empty($output) ? $output : new \stdClass());
+                try {
+                    $this->json(!empty($output) ? $output : new \stdClass());
+                } catch (JsonException $e) {
+                    throw new Exception('Failed to parse binary response: ' . $e->getMessage(), 400);
```
vermakhushboo, line 565: What should be the error message for this be called @christyjacob4? And is it okay for it to be a 400 error?

christyjacob4: The exception can arise in case of any malformed JSON . So the error message shouldn't talk about binary response specifically. 
400 is fine.

app/controllers/api/functions.php
```diff
@@ -1599,13 +1601,14 @@
     ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
     ->label('sdk.response.model', Response::MODEL_EXECUTION)
     ->param('functionId', '', new UID(), 'Function ID.')
-    ->param('body', '', new Text(0, 0), 'HTTP body of execution. Default value is empty string.', true)
+    ->param('body', '', new Text(10485760, 0), 'HTTP body of execution. Default value is empty string.', true)
```
Meldiron, line 1604: Why 10MB, and not 20MB like on executor?

src/Appwrite/Functions/Validator/Headers.php
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        if (\is_string($value)) {
+            $value = \json_decode($value, true);
+        }
+
+        if (\json_last_error() == JSON_ERROR_NONE) {
+            if (\is_array($value)) {
+                foreach ($value as $key => $val) {
+                    // Check for invalid characters in key and value
```
Meldiron, line 53:
```suggestion
                    // Check for invalid characters in key
```

src/Appwrite/Functions/Validator/Headers.php
```diff
@@ -0,0 +1,90 @@
...
+                    if (!preg_match('/^[a-zA-Z0-9_-]+$/', $key)) {
+                        return false;
+                    }
+                    // Check for x-appwrite- prefix
+                    if (0 === strpos($key, 'x-appwrite-')) {
```
Meldiron, line 58: We can use this: https://www.php.net/manual/en/function.str-starts-with.php

src/Appwrite/Functions/Validator/Headers.php
```diff
@@ -0,0 +1,90 @@
...
+
+        if (\is_string($value)) {
+            $value = \json_decode($value, true);
+        }
+
+        if (\json_last_error() == JSON_ERROR_NONE) {
+            if (\is_array($value)) {
+                foreach ($value as $key => $val) {
+                    // Check for invalid characters in key and value
+                    if (!preg_match('/^[a-zA-Z0-9_-]+$/', $key)) {
```
Meldiron, line 54: Regex is considerred bad practice in most scenarios. Let's check other validators and see how we do similar checks there. If does similar validation yet, lets ask Christy about this. I think we should have whitelist approach here instead (I think Text validator uses whitelist approach)

src/Appwrite/Functions/Validator/Headers.php
```diff
@@ -0,0 +1,90 @@
...
+                        return false;
+                    }
+                    // Check for x-appwrite- prefix
+                    if (0 === strpos($key, 'x-appwrite-')) {
+                        return false;
+                    }
+                }
+            }
+        }
+        return true;
```
Meldiron, line 64: I might be missunderstanding, but I think if there is json error, then validator will say it's valid. but it isn't, because its not valid json
CleanShot 2024-08-09 at 18 06 24@2x

tests/unit/Functions/Validator/HeadersTest.php
```diff
@@ -0,0 +1,48 @@
...
+class HeadersTest extends TestCase
+{
+    protected ?Headers $object = null;
+
+    public function setUp(): void
+    {
+        $this->object = new Headers();
+    }
+
+    public function testValues(): void
```
Meldiron, line 17: Validator accepts `param mixed $value`. Lets try to pass some invaid value like string, number, boolean or null. Ensure they return false too, as they are not valid

src/Appwrite/Utopia/Response.php
```diff
@@ -697,6 +708,49 @@ public function yaml(array $data): void
...
+     *
+     * @see http://en.wikipedia.org/wiki/JSON
+     *
+     * @param  mixed  $data
+     * @return void
+     */
+    public function json($data): void
+    {
+        if (!is_array($data) && !$data instanceof \stdClass) {
+            throw new \Exception('Invalid JSON input var');
```
Meldiron, line 746:
```suggestion
            throw new \Exception('Response body is not a valid JSON object.');
```

tests/e2e/General/HTTPTest.php
```diff
@@ -131,7 +131,6 @@ public function testSpecs()
                 'content-type' => 'application/json',
             ], json_decode(file_get_contents($directory . $file), true));
 
-            $response['body'] = json_decode($response['body'], true);
```
Meldiron, line 134: This removal is a bit concerning. What test is failing without it? Something doesn't seem right.

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -1316,6 +1316,217 @@ public function testCreateCustomExecution(string $folder, string $name, string $
...
+            'entrypoint' => 'index.php',
+            'code' => new CURLFile($code, 'application/x-gzip', basename($code)),
+            'activate' => true
+        ]);
+
+        $deploymentId = $deployment['body']['$id'] ?? '';
+        $this->assertEquals(202, $deployment['headers']['status-code']);
+
+        // Poll until deployment is built
+        while (true) {
```
Meldiron, line 1354: Check other tests, I think I saw some "waitUntil" methods - waiting until deployment is ready. It could simplify this test a little.

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -1316,6 +1316,217 @@ public function testCreateCustomExecution(string $folder, string $name, string $
...
+
+        $this->assertEquals(201, $function['headers']['status-code']);
+
+        $variable = $this->client->call(Client::METHOD_POST, '/functions/' . $functionId . '/variables', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()), [
+            'key' => 'CUSTOM_VARIABLE',
+            'value' => 'variable',
+        ]);
```
Meldiron, line 1447: Does this test need this variable?

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -1316,6 +1316,217 @@ public function testCreateCustomExecution(string $folder, string $name, string $
...
+        ], $this->getHeaders()), [
+            'entrypoint' => 'index.php',
+            'code' => new CURLFile($code, 'application/x-gzip', basename($code)),
+            'activate' => true
+        ]);
+
+        $deploymentId = $deployment['body']['$id'] ?? '';
+        $this->assertEquals(202, $deployment['headers']['status-code']);
+
+        while (true) {
```
Meldiron, line 1463: Same as above with waitUntil helper

app/controllers/api/functions.php
Meldiron: Re-opening this based on recent discussion on Discord.

app/controllers/api/functions.php
vermakhushboo: The max payload size in Appwrite is 12 MB - so made it 10 MB here to keep some margin for metadata.
https://github.com/appwrite/appwrite/blob/e297371aa4c87a2347fb57e0510e65095a18c95f/app/http.php#L35C2-L35C13

tests/unit/Functions/Validator/HeadersTest.php
vermakhushboo: Is `1234 => 'value',` a valid header?

src/Appwrite/Functions/Validator/Headers.php
vermakhushboo: @christyjacob4 what do you recommend using here?

src/Appwrite/Functions/Validator/Headers.php
vermakhushboo: Updated

tests/e2e/General/HTTPTest.php
vermakhushboo: It is not related to my changes, I just updated it to fix the failing test.

This is what the `body` looks like. 
```
["body"]=>
  array(1) {
    ["schemaValidationMessages"]=>
    array(0) {
    }
  }
  ```
  
  It is not a string. It is an array and the `json_decode` fails.

Comment (Meldiron) [reply tests/e2e/General/HTTPTest.php:134]: Can you check if this test is passing fine in 1.6.x without removal? Would also be best to double check with Christy to find which PR started to fail this test, to be 100% sure we are not missing anything. Changes to general HTTP test are scarry.

File: tests/unit/Functions/Validator/HeadersTest.php (hunk 20 above)
Comment (Meldiron) [reply tests/unit/Functions/Validator/HeadersTest.php:17]: It is not, we can have assertion for that. But I meant `$headres = 123` or `$headers = false`. When header isn't array.

File: src/Appwrite/Functions/Validator/Headers.php [hunk 26]
```diff
@@ -0,0 +1,102 @@
...
+            return true;
+        }
+
+        if (\is_string($value)) {
+            $value = \json_decode($value, true);
+        }
+
+        if (json_last_error() !== JSON_ERROR_NONE) {
+            return false;
+        } else {
```
Meldiron, line 52: This can be simplified, `else` isnt required.

src/Appwrite/Functions/Validator/Headers.php
```diff
@@ -0,0 +1,102 @@
...
+     */
+    public function isValid($value): bool
+    {
+        if ($this->allowEmpty && empty($value)) {
+            return true;
+        }
+
+        if (\is_string($value)) {
+            $value = \json_decode($value, true);
+        }
```
Meldiron, line 48: I think this valudator shouldnt encode. It could lead to validator passing, but then having string value in the action causing trouble.

app/controllers/api/functions.php
Meldiron: It wasnt addressed yet.

tests/e2e/General/HTTPTest.php
vermakhushboo: Yes, it is passing in 1.6.x before this PR. In this PR, we have added this line in `Client.php`: 
`$strpos = \is_bool($strpos) ? \strlen($responseType) : $strpos;`
If we don't add it, the test passes for `HttpTest.php`, however, I believe that code isn't correct because in case a `;` is not present, it doesn't consider the `responseType` as `application/json`. Would love for you to take a look here @christyjacob4 and @lohanidamodar (since you updated that test file last).

app/controllers/api/functions.php
vermakhushboo: Done

src/Appwrite/Functions/Validator/Headers.php
```diff
@@ -0,0 +1,103 @@
...
+
+                // Check if the key is a single character and ensure it is an alphabetic character
+                if (\strlen($key) === 1 && !preg_match('/^[a-zA-Z]$/', $key)) {
+                    return false;
+                }
+
+                // Check for invalid characters in keys longer than one character
+                if (\strlen($key) > 1 && !preg_match('/^[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9]$/', $key)) {
+                    return false;
+                }
```
Meldiron, line 69: My only remaining concern is with using regex here.

vermakhushboo: Replaced it with `ctype_alnum` method

tests/e2e/General/HTTPTest.php
vermakhushboo: Marking it as resolved as the logic was verified

src/Appwrite/Functions/Validator/Headers.php
```diff
@@ -0,0 +1,107 @@
...
+            return true;
+        }
+
+        if (!\is_array($value)) {
+            return false;
+        }
+
+        if (json_last_error() !== JSON_ERROR_NONE) {
+            return false;
+        }
```
christyjacob4, line 52: what does this check do? doesn't seem to be relevant

christyjacob4 (approved):

PR #8493, opened 2024-08-08
Fix: 4XX execution status

Meldiron:
## What does this PR do?

Marks 4XX executions as successful

## Test Plan

- [x] New tests

## Related PRs and Issues

x

## Checklist


tests/resources/functions/php-4xx/index.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+return function ($context) {
+    return $context->res->text('Invalid input.', 400);
+};
```
christyjacob4, line 5: Each new function adds a lot of time to the tests. Why dont we just update an existing test to return a 4xx code instead ?  
Screenshot 2024-08-10 at 1 05 59 PM

Meldiron: I agree function tests need a bigger rework. They are also flaky. But it's not in the scope of this PR. We can address that separately.

christyjacob4: Yea but for your specific case you can just modify the existing function in `php` itself and make it return a 400 status

Meldiron: It will still be making new function, and new deployment. tarring a code takes a few ms. waiting for the build is slow. A bigger refactor is needed to actually make it faster.

Alternatively I could make this a dependent test but that will create inconsistency - not sure if worth it before proper refactor of function tests.

christyjacob4: In this function 
https://github.com/appwrite/appwrite/blob/fix-4xx-executions-status/tests/resources/functions/php/index.php

1. Pass the expected status code in the body. 
2. The function then returns the same status code as passed in the parameter. So from the FunctionsCustomClientTest.php you send 200 and expect a 200. You send a 400 and expect a 400 back. 

And you will be just reusing this same test. 
https://github.com/appwrite/appwrite/blob/fix-4xx-executions-status/tests/e2e/Services/Functions/FunctionsCustomClientTest.php#L166 

just create 1 more execution passing 200 and 400 to test both scenarios. No need to rebuild or redeploy

Meldiron: Updated

christyjacob4 (approved):

PR #8525, opened 2024-08-12
Permissions + Comment Fix

EVDOG4LIFE:
## What does this PR do?

Fixes permissions for non contributors and allows workflow to edit comments instead of spamming. 

## Test Plan

See test run in https://github.com/EVDOG4LIFE/appwrite_wfTest/pull/21

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/.github/workflows/pr-scan.yml
+++ b/.github/workflows/pr-scan.yml
@@ -1,15 +1,19 @@
 name: PR Security Scan
-on:
+on: 
   pull_request_target:
-    branches: ['**']
-
+    types: [opened, synchronize, reopened]
+ 
 jobs:
   scan:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
     steps:
-    - name: Check out code
+    - name: Check out code 
       uses: actions/checkout@v4
       with:
+        ref: ${{ github.event.pull_request.head.sha }}
         fetch-depth: 0
         submodules: 'recursive'
 
@@ -42,7 +46,6 @@ jobs:
       id: process-results
       uses: actions/github-script@v7
       with:
-        github-token: ${{ secrets.GITHUB_TOKEN }}
         script: |
           const fs = require('fs');
           let commentBody = '## Security Scan Results for PR\n\n';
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-scan.yml
```diff
@@ -1,8 +1,8 @@
 name: PR Security Scan
 on:
-  pull_request:
-    types: [opened, synchronize, reopened]
-  workflow_dispatch:
+  pull_request_target:
+    branches: ['**']
```
TorstenDittmann, line 4: this allows to do lots of permissions that might not be there 🤔 

I have not found a good way on how to handle these scenarios :-/

.github/workflows/pr-scan.yml
```diff
@@ -1,8 +1,8 @@
...
+
```
stnguyen90, line 5: Let's limit the access of the GITHUB_TOKEN by specifying [permissions](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token).

.github/workflows/pr-scan.yml
```diff
@@ -79,9 +84,22 @@ jobs:
...
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        issue-number: ${{ github.event.pull_request.number }}
+        comment-author: 'github-actions[bot]'
+        body-includes: Security Scan Results for PR
+
+    - name: Create or update comment
+      uses: peter-evans/create-or-update-comment@v3
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
```
stnguyen90, line 101: Let's remove setting the token for all of these because they automatically pick up the GITHUB_TOKEN by default.

.github/workflows/pr-scan.yml
stnguyen90: jobs triggered from pull_request_target are executed in the target branch. Since will be using the workflow file in the target branch, it won't be possible for contributors to modify the workflow to do something malicious.

Source: https://stackoverflow.com/questions/74957218/what-is-the-difference-between-pull-request-and-pull-request-target-event-in-git

If anything, there may be concern with secrets being in repo being dumped, but we will need to mitigate that by reading the code and making sure nothing nefarious is being done.

Besides that, we can limit the permissions on the `GITHUB_TOKEN` and move secrets to a separate environment to be used only when we deploy.

stnguyen90 (changes requested):

stnguyen90 (approved):

PR #8544, opened 2024-08-16
Support renaming attributes

PineappleIOnic:
## What does this PR do?

This PR implements the ability to rename attributes in the API; it also adds tests for this, including for renaming relationship attributes to ensure they do not break when doing this.

## Test Plan

Tests have been added for both normal attributes and relationships (all types) to confirm nothing breaks.

## Checklist


abnegate (approved):
Looks good, will keep the PR open until the console side is sorted in case we need to change target branch :+1:

Issue #6061, opened 2023-08-30
🚀 Feature: Delete all user sessions on blocking a user


prechILLEd:
### 🔖 Feature description

When deleting a user in the appwrite backend the sessions are still stored and not deleted. 

Therefore a different user using the same device wont be able to sign in because the request is completely blocked.

### 🎤 Pitch

In my use-case it would make more sense to delete all sessions when a user is blocked since other users might not be able to sign in because the requests get rejected due to the blocked users sessions.

Steps:
1. Sign in with user A
2. Block user A in the appwrite backend
3. Restart App/Reload Website
4. Try to sign in with User B
5. Getting "AppwriteException: user_blocked, The current user has been blocked. You can unblock the user from the Appwrite console. (401)"

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

JorensM:
I think an even better solution would be to automatically end the previous session if a new session is being created. I'm not sure about the security concerns of this though

prechILLEd:
> 

For that use-case it is probably best to set the session limit per user to 1 in the Auth Security tab.

You might want a user to be logged in on multiple devices at the same time and your thought would deny that.

joeyouss:
Hi, thank you for opening this issue, but would it make sense for you to maybe set the limit once and then try it out? https://appwrite.io/docs/authentication-security#limits

I thinj, there are some potential downsides to doing this. In some cases, retaining the session might be necessary for auditing or investigative purposes, especially in applications that require high security. Also, automatically deleting sessions could result in data loss if the user was in the middle of an unsaved task. There might be scenarios where an account is temporarily blocked due to some system or admin error. In such cases, deleting all sessions would force the user to log back into all devices, which could be an inconvenience.

Can you let me know if setting the limit does the work for you and what are your thoughts?

prechILLEd:
Thanks @joeyouss for the detailed answer.

You are totally right. You do not want to delete all of the users sessions for that purposes.
For my in-house app it does not really matter if data is lost.

Even with a session limit set to 1 my issue would still be not fixed - beside the fact it is not my goal to limit the user to have only 1 active session.

As a workaround I wrote a function that deletes all sessions for every user that is blocked. 
[...]

PR #8545, opened 2024-08-16
Delete all user sessions on blocking a user

SophieNguyen113:
## What does this PR do?

This PR introduces updates to the Update user status API to include the functionality of deleting all sessions when a user is blocked. Previously, user sessions were not deleted, which caused issues when a different user attempted to sign in on the same device. The changes ensure that when a user's status is updated to blocked, all their active sessions are deleted, allowing a different user to sign in.

Fixing https://github.com/appwrite/appwrite/issues/6061

## Test Plan

1. **Verify User Existence**
   - Send a GET request to `/users/{userId}` and check for a 200 status code.

2. **Activate User**
   - PATCH `/users/{userId}/status` with `status: true`.
   - Verify response: status code 200 and body contains `status: true`.

3. **Block User**
   - PATCH `/users/{userId}/status` with `status: false`.
   - Verify response: status code 500.

4. **Check Block Status**
   - GET `/users/{userId}`.
   - Verify response: status code 200 and body contains `status: false`.

5. **Reactivate User**
   - PATCH `/users/{userId}/status` with `status: true`.
   - Verify response: status code 200 and body contains `status: true`.

## Checklist


SophieNguyen113:
@stnguyen90  Hello, could you please give me permission for the workflows? Thanks.

app/controllers/api/users.php
```diff
@@ -962,6 +962,7 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
     ->label('audits.event', 'user.update')
     ->label('audits.resource', 'user/{response.$id}')
     ->label('audits.userId', '{response.$id}')
+    ->label('usage.metric', 'users.{scope}.requests.update')
```
stnguyen90, line 965: You can remove this. It's not used anymore.

composer.lock
stnguyen90, line 1: I don't think it was necessary to modify the composer.lock, right? If not, it would be best to leave it untouched.

docs/references/users/update-user-status.md
```diff
@@ -1 +1 @@
-Update the user status by its unique ID. Use this endpoint as an alternative to deleting a user if you want to keep user's ID reserved.
\ No newline at end of file
+Update the user status by its unique ID. This endpoint now includes functionality to delete all active sessions when a user is blocked, resolving issues when a different user attempts to sign in on the same device.
```
stnguyen90, line 1: > This endpoint now includes functionality

No need to say it "now includes functionality". Just state it as a fact.

stnguyen90 (changes requested):
Tests seem to be failing. Have you tried running the tests locally?

SophieNguyen113:
@stnguyen90  Hello, all the test cases were fixed. Please review. Thanks

SophieNguyen113:
Since I accidentally messed up this PR, I would close it and open a new one. Thanks.

PR #8546, opened 2024-08-16
Feat: Add seconds precision to scheduledAt

Meldiron:
## What does this PR do?

Ensures we don't accept schedule datetime we cannot achieve.

## Test Plan

- [x] Update existing success tests
- [x] Add new failure test

Passed locally
CleanShot 2024-08-16 at 11 17 00@2x

## Related PRs and Issues

- https://github.com/utopia-php/database/pull/438

## Checklist


composer.json
```diff
@@ -50,7 +50,7 @@
         "utopia-php/cache": "0.10.*",
         "utopia-php/cli": "0.15.*",
         "utopia-php/config": "0.2.*",
-        "utopia-php/database": "0.50.*",
+        "utopia-php/database": "dev-feat-datetime-validator-precision as 0.50.99",
```
Meldiron, line 53: Release

PR #8553, opened 2024-08-19
Sync 1.6.x Filters and Migrations with latest

PineappleIOnic:
## What does this PR do?

- Added migration for `files` metadata size increase
- Disabled `specification` migration until variable runtime specs are ready
- Unset `key` for runtimes in response filters
- Set the correct attribute `scopes` to be optional, reverting it for key and setting it for functions.

## Test Plan

- Added test for removing key from runtimes

## Checklist


src/Appwrite/Migration/Version/V21.php
```diff
@@ -177,7 +188,7 @@ protected function fixDocument(Document $document): Document
                 $document->setAttribute('scopes', []);
 
                 // Add size attribute
-                $document->setAttribute('specification', APP_FUNCTION_BASE_SPECIFICATION);
+                // $document->setAttribute('specification', APP_FUNCTION_BASE_SPECIFICATION); Uncomment after we merge variable runtimes specifications.
```
Meldiron, line 191: Double check we also unset it in request filter in the other PR

Meldiron (approved):

src/Appwrite/Migration/Version/V21.php
```diff
@@ -148,6 +148,17 @@ private function migrateCollections(): void
                     } catch (\Throwable $th) {
                         Console::warning("'scheduleId' from {$id}: {$th->getMessage()}");
                     }
+                    break;
+                case 'files':
+                    try {
+                        /**
+                         * Increase metadata size to 7500
+                         */
+                        $this->projectDB->updateAttribute($id, 'metadata', size: 75000);
```
Meldiron, line 157: Size above 65535 will cause breaking change. For simplicity of 1.6 migration, we should avoid that

src/Appwrite/Migration/Version/V21.php
PineappleIOnic: Good catch, added it

christyjacob4 (approved):

PR #8556, opened 2024-08-19
Feat: Separate deployment sizes

Meldiron:
## What does this PR do?

Separate deployment and build size to 2 attributes, and allow querying one of them.

## Test Plan

- [x] Update tests

## Related PRs and Issues

x

## Checklist


tests/e2e/Services/Account/AccountBase.php
```diff
@@ -13,6 +13,7 @@ public function testCreateAccount(): array
         $email = uniqid() . 'user@localhost.test';
         $password = 'password';
         $name = 'User Name';
+        \var_dump($email);
```
christyjacob4, line 16:
```suggestion
```

src/Appwrite/Platform/Workers/Builds.php
```diff
@@ -244,13 +244,7 @@ protected function buildDeployment(Device $deviceForFunctions, Func $queueForFun
                         $tmpTemplateDirectory .= '/';
                     }
 
-                    $directorySize = $localDevice->getDirectorySize($tmpTemplateDirectory);
-                    $functionsSizeLimit = (int)System::getEnv('_APP_FUNCTIONS_SIZE_LIMIT', '30000000');
-                    if ($directorySize > $functionsSizeLimit) {
-                        throw new \Exception('Repository directory size should be less than ' . number_format($functionsSizeLimit / 1048576, 2) . ' MBs.');
-                    }
```
christyjacob4, line 251: why was this check removed?

Meldiron: template is whitelist, we have full control over size and can be sure no abuse occurs here. Also, it didn't work - it was always saying 0.
Also, size wasn't being set right.. When you upload deployment, it uses tar.gz file size. but here it was using folder size before compression. It wasn't giving correct value. Getting cize from .tar.gz file instead makes implementation consistent with existing flows.

christyjacob4 (approved):

PR #8561, opened 2024-08-20
Feat: Headers validator benchmark

Meldiron:
## What does this PR do?

Adds PHPBench, and benchmark for headers validator

## Test Plan

- [x] Add benchmark CI/CD
- [x] Manual run:

image

## Related PRs and Issues

x

## Checklist


.github/workflows/bench.yml
```diff
@@ -0,0 +1,16 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    - name: Run Benchmark
+      run: |
+        docker run --rm -v $PWD:/app composer sh -c \
+        "composer install --profile --ignore-platform-reqs && git config --global --add safe.directory /app && composer bench -- --progress=plain"
```
christyjacob4, line 16: lets run this as a job in one of the existing workflows 
https://github.com/appwrite/appwrite/blob/main/.github/workflows/tests.yml#L102

Meldiron: tests.yml runs for every service. This is perfectly fine to run only once. Separation also allows concurrency, I don't see downside in that. Plus it's copy&paste from utopia library, so it follows consistency to have it separated.

Meldiron: I could see same discussion about linter - linter could be in tests.yml, but is separate.

Meldiron: Removed bench CI/CD

christyjacob4 (approved):

PR #8597, opened 2024-08-30
Feat string attribute resizing

PineappleIOnic:
## What does this PR do?

This PR introduces the ability to resize string attributes as well as the relevant tests

## Test Plan

Tests have been added to ensure the new functionality is working as expected.

## Checklist


tests/e2e/Services/Databases/DatabasesCustomServerTest.php
```diff
@@ -3242,6 +3242,168 @@ public function testAttributeUpdateEnum(array $data)
...
+            'x-appwrite-key' => $this->getProject()['apiKey']
+        ]), [
+            'data' => [
+                'string' => str_repeat('a', 10)
+            ]
+        ]);
+
+        $this->assertEquals(200, $document['headers']['status-code']);
+        $this->assertEquals(10, strlen($document['body']['string']));
+    }
```
abnegate, line 3405: Let's add a check that we get an exception when trying to create a document with data bigger the new size

app/controllers/api/databases.php
```diff
@@ -1859,21 +1871,31 @@ function updateAttribute(
...
+                required: $required,
+                size: $size
+            );
+        } catch (DatabaseException $e) {
+            if ($e->getMessage() === "Resize would result in data truncation") {
+                throw new Exception(Exception::ATTRIBUTE_INVALID_RESIZE);
+            } else {
+                throw $e;
+            }
+        }
```
abnegate, line 1898: Let's add the try/catch directly around the `$dbForProject->updateAttribute` call inside the `updateAttribute` function to keep the scope narrow

abnegate (approved):

PR #8627, opened 2024-09-05
Fix: logging behaviour for Functions

Meldiron:
## What does this PR do?

Ensures that function with logging=false still always makes execution, just doesn't have logs and errors

## Test Plan

- [x] New tests
CleanShot 2024-09-05 at 15 21 57@2x

## Related PRs and Issues

x

## Checklist

### DIFF
--- a/tests/e2e/Services/Functions/FunctionsCustomServerTest.php
+++ b/tests/e2e/Services/Functions/FunctionsCustomServerTest.php
@@ -2666,4 +2666,147 @@ class FunctionsCustomServerTest extends Scope
 
         $this->assertEquals(204, $response['headers']['status-code']);
     }
+
+    public function testFunctionLogging()
+    {
+        // Preparations: Create Function
+        $folder = 'node';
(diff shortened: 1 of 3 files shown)


tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -2666,4 +2666,147 @@ public function testCreateFunctionWithResponseFormatHeader()
 
         $this->assertEquals(204, $response['headers']['status-code']);
     }
+
+    public function testFunctionLogging()
+    {
+        // Preparations: Create Function
+        $folder = 'node';
+        $code = realpath(__DIR__ . '/../../../resources/functions') . "/$folder/code.tar.gz";
+        $this->packageCode($folder);
```
christyjacob4, line 2675: lets make use of an existing test to just add the logging attribute. adding a whole new test for this will keep adding overhead to the test times. functions tests already takes the longest

PR #8667, opened 2024-09-15
Multi region support

shimonewman:


src/Appwrite/Platform/Tasks/Maintenance.php
```diff
@@ -47,7 +47,10 @@ public function action(Database $dbForPlatform, Document $console, Certificate $
 
             Console::info("[{$time}] Notifying workers with maintenance tasks every {$interval} seconds");
 
-            $dbForPlatform->foreach('projects', function (Document $project) use ($queueForDeletes, $usageStatsRetentionHourly) {
+            $dbForPlatform->foreach('projects', [
+                    Query::equal('region', System::getEnv('_APP_REGION', 'default'))
```
christyjacob4, line 51: do we have an index on this attribute?

docker-compose.yml
```diff
@@ -198,6 +198,8 @@ services:
       - _APP_DATABASE_SHARED_TABLES_V1
       - _APP_DATABASE_SHARED_NAMESPACE
       - _APP_FUNCTIONS_CREATION_ABUSE_LIMIT
+      - _APP_DATABASE_KEYS
```
christyjacob4, line 201: is this variable needed here? I dont see it used anywhere

app/controllers/api/databases.php
```diff
@@ -563,7 +563,8 @@ function updateAttribute(
     ->param('search', '', new Text(256), 'Search term to filter your list results. Max length: 256 chars.', true)
     ->inject('response')
     ->inject('dbForProject')
-    ->action(function (array $queries, string $search, Response $response, Database $dbForProject) {
+    ->inject('project')
```
christyjacob4, line 566: why do we inject project here? I dont see it being used

app/controllers/api/projects.php
```diff
@@ -138,6 +138,14 @@
 
         $databases = Config::getParam('pools-database', []);
 
+        if ($region !== 'default') {
```
christyjacob4, line 141: I think we need a more robust check here. The console sends the `region` attribute right? 

if the console stops sending the region or sends a different string, this check will break .

app/controllers/api/proxy.php
```diff
@@ -55,14 +55,19 @@
...
-        $functionsDomain = System::getEnv('_APP_DOMAIN_FUNCTIONS', '');
-        if ($functionsDomain != '' && str_ends_with($domain, $functionsDomain)) {
-            throw new Exception(Exception::GENERAL_ARGUMENT_INVALID, 'You cannot assign your functions domain or it\'s subdomain to specific resource. Please use different domain.');
+        $denyListString = System::getEnv('_APP_CUSTOM_DOMAIN_DENY_LIST', '');
+        $deniedDomains = array_map('trim', explode(',', $denyListString));
+
+        foreach ($deniedDomains as $deniedDomain) {
+            if (str_ends_with($domain, $deniedDomain)) {
+                throw new Exception(Exception::GENERAL_ARGUMENT_INVALID, 'You cannot assign your functions domain or its subdomain to a specific resource. Please use a different domain.');
+            }
```
christyjacob4, line 70: for backwards compatibility lets merge the two variables `_APP_DOMAIN_FUNCTIONS` and `_APP_CUSTOM_DOMAIN_DENY_LIST` so that we dont accidentally forget it

docker-compose.yml
shimonewman: it is used in projects.php controller

app/controllers/api/proxy.php
shimonewman: Add _APP_CUSTOM_DOMAIN_DENY_LIST vars to variables.php and  inject it to the appwrite container

docker-compose.yml
shimonewman: It is used only for cloud, so we can remove t

src/Appwrite/Platform/Tasks/Maintenance.php
shimonewman: yes

app/controllers/api/projects.php
shimonewman: We should test what happens  if region is empty and if its default

app/controllers/api/proxy.php
```diff
@@ -55,14 +55,25 @@
...
 
-        $functionsDomain = System::getEnv('_APP_DOMAIN_FUNCTIONS', '');
-        if ($functionsDomain != '' && str_ends_with($domain, $functionsDomain)) {
-            throw new Exception(Exception::GENERAL_ARGUMENT_INVALID, 'You cannot assign your functions domain or it\'s subdomain to specific resource. Please use different domain.');
+        $functionsDomain = System::getEnv('_APP_DOMAIN_FUNCTIONS');
+        $denyListString = System::getEnv('_APP_CUSTOM_DOMAIN_DENY_LIST');
+
+        if (!empty($functionsDomain)) {
+            $denyListString .= ',' . $functionsDomain;
+        }
```
christyjacob4, line 69: it should be the other way. 

if !empty(denyListString) {
     functionsDomain + denyListString
}

christyjacob4 (approved):

PR #8668, opened 2024-09-16
fix: function storage metrics

TorstenDittmann:
## What does this PR do?

- breaks down the function storage metrics to individual metrics
- fix docker build warnings using lowercase `as`

## Test Plan

- fixed tests

## Checklist


app/config/specs/open-api3-1.6.x-console.json
```diff
@@ -9451,9 +9451,6 @@
                                             "python-3.11",
                                             "python-3.12",
                                             "python-ml-3.11",
-                                            "deno-1.21",
-                                            "deno-1.24",
-                                            "deno-1.35",
                                             "deno-1.40",
```
Meldiron, line 9454: Are those intentional? If not, lets upgrade php-runtimes library to latest. Or check if it was removed there.

Meldiron (approved):
Changes looks good, 1 small comment (I saw similar problem in SDKs after 1.6)

app/config/specs/open-api3-1.6.x-console.json
TorstenDittmann: its latest version, looks like generating specs was skipped when the update happened 👍🏻

PR #8680, opened 2024-09-18
Update Database lib

PineappleIOnic:
## What does this PR do?

This PR updates the database library to version 0.55.x, is also updates a bunch of instances where we expect false from find() when a document isn't found for compatibility with 0.55.x.

As 0.55.x of database lib will now return an empty document instead of false when we have a miss.

## Test Plan

Tests will not be modified and should pass as normal.

## Related PRs and Issues

- https://github.com/utopia-php/database/pull/449

## Checklist


src/Appwrite/Platform/Workers/Certificates.php
```diff
@@ -126,7 +126,7 @@ private function execute(Domain $domain, Database $dbForConsole, Mail $queueForM
         $certificate = $dbForConsole->findOne('certificates', [Query::equal('domain', [$domain->get()])]);
 
         // If we don't have certificate for domain yet, let's create new document. At the end we save it
-        if (!$certificate) {
+        if (empty($certificate) || $certificate->isEmpty()) {
```
stnguyen90, line 129: `$certificate` will always be a `Document`, right? So, we just need the `$certificate->isEmpty()` check?

stnguyen90 (commented):
Wasn't the whole point of https://github.com/utopia-php/database/pull/449 to simplify `findOne()` so that it always returns a `Document`? And since it always returns a Document, we only need to use `$doc->isEmpty()` right?

abnegate (approved):

PR #8714, opened 2024-09-24
fix(assistant): make Appwrite Assistant hostname configurable

basert:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/config/variables.php
```diff
@@ -1119,6 +1119,15 @@
         'category' => 'Assistant',
         'description' => '',
         'variables' => [
+            [
+                'name' => '_APP_ASSISTANT_HOST',
+                'description' => 'Host and port of the assistant container.',
+                'introduction' => '1.4.0',
```
TorstenDittmann, line 1125:
```suggestion
                'introduction' => '1.4.0',
```
this version probably need to be updated before a release 👍🏻

app/controllers/api/console.php
```diff
@@ -71,7 +71,7 @@
     ->param('prompt', '', new Text(2000), 'Prompt. A string containing questions asked to the AI assistant.')
     ->inject('response')
     ->action(function (string $prompt, Response $response) {
-        $ch = curl_init('http://appwrite-assistant:3003/');
+        $ch = curl_init(System::getEnv("_APP_ASSISTANT_HOST"));
```
TorstenDittmann, line 74: we might wanna check if this env var is actually set, might be important for self-hosted instances.

There's also `_APP_ASSISTANT_ENABLED` that tells the console if the assistant is enabled, we might wanna check for this one too here 👍🏻 

We have a similar check here https://github.com/appwrite/appwrite/blob/main/app/controllers/api/account.php#L1789-L1791

PR #8718, opened 2024-09-25
fix: trigger create user event

loks0n:
Fixes https://github.com/appwrite/appwrite/issues/2406

Moves `users.[userId].create` event triggering to a database listener.

This fixes several endpoints that create a user, but didn't previously trigger the event.


app/controllers/api/account.php
```diff
@@ -1498,6 +1498,12 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
                         'providerType' => MESSAGE_TYPE_EMAIL,
                         'identifier' => $email,
                     ]));
+
+                    $queueForEvents
+                        ->setEvent('users.[userId].create')
+                        ->setParam('userId', $user->getId())
+                        ->trigger();
```
Meldiron, line 1505: I think we also need to set payload here.

Endpoints to it by default - HTTP response body becomes the payload of the event.

You can then validate this by making function for this event, and printing `context.req.bodyText`. It should include entire user document (as JSON)

app/controllers/api/account.php
```diff
@@ -1498,6 +1498,12 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
                         'providerType' => MESSAGE_TYPE_EMAIL,
                         'identifier' => $email,
                     ]));
+
+                    $queueForEvents
```
Meldiron, line 1502: Just thinking out loud, will we have same problem in Magic URL, email OTP, or SMS logins? When they create new user, is the event triggered properly?

app/controllers/api/account.php
loks0n: Done

app/controllers/api/account.php
loks0n: We can apply the same fix across them, or perhaps rethink events a little bit. In general we have a problem where many endpoints can now trigger multiple events depending on the scenerio

loks0n:
fixes: https://github.com/appwrite/appwrite/issues/2406

app/controllers/api/account.php
```diff
@@ -1498,6 +1498,13 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
+                        ->setEvent('users.[userId].create')
```
stnguyen90, line 1503: Is this going to mess up the event for the endpoint itself? since this replaces the event set during init, right?

https://github.com/appwrite/appwrite/blob/fdd39a29c3667069d97af3fc40bd5098bd32c41c/app/controllers/shared/api.php#L437

app/controllers/shared/api.php
```diff
@@ -57,8 +57,72 @@
     return $label;
 };
 
-$databaseListener = function (string $event, Document $document, Document $project, Usage $queueForUsage, Database $dbForProject) {
+$eventDatabaseListener = function (string $event, Document $document, Event $queueForEvents, Response $response, Func $queueForFunctions, Document $project) {
+    // Set event empty to avoid triggering events by default
+    $queueForEvents->setEvent('');
 
+    if ($document->getCollection() === 'users' && $event === Database::EVENT_DOCUMENT_CREATE) {
```
christyjacob4, line 64: lets use a switch case here, and we can also remove the second condition right? Since this listener is only called on document creates?

christyjacob4 (approved):

PR #8720, opened 2024-09-25
feat: payload response type

loks0n:


src/Appwrite/Specification/Format/OpenAPI3.php
```diff
@@ -576,6 +576,10 @@ public function parse(): array
                         $type = 'boolean';
                         break;
 
+                    case 'payload':
+                        $type = 'payload';
```
byawitz, line 580: We don't use the constant from the `src/Appwrite/Utopia/Response/Model.php` file?

byawitz (approved):

src/Appwrite/Specification/Format/OpenAPI3.php
loks0n: Should we? We don't elsewhere in this function

Meldiron (approved):

Meldiron (approved):

christyjacob4 (approved):

PR #8722, opened 2024-09-25
Add domain and force HTTPS env vars to mail worker

stnguyen90:
## What does this PR do?

The worker uses the env vars

https://github.com/appwrite/appwrite/blob/fdd39a29c3667069d97af3fc40bd5098bd32c41c/src/Appwrite/Platform/Workers/Mails.php#L68-L69

So we need to include it in the compose file to ensure the variable passes into the container.
## Test Plan

None

## Related PRs and Issues

None

## Checklist

### DIFF
--- a/app/views/install/compose.phtml
+++ b/app/views/install/compose.phtml
@@ -531,6 +531,8 @@ $image = $this->getParam('image', '');
       - _APP_SMTP_USERNAME
       - _APP_SMTP_PASSWORD
       - _APP_LOGGING_CONFIG
+      - _APP_DOMAIN
+      - _APP_OPTIONS_FORCE_HTTPS
 
   appwrite-worker-messaging:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -588,6 +588,8 @@ services:
       - _APP_DOMAIN
       - _APP_OPTIONS_FORCE_HTTPS
       - _APP_DATABASE_SHARED_TABLES
+      - _APP_DOMAIN
```
byawitz, line 591: Those two variables are in rows [588-589](https://github.com/appwrite/appwrite/pull/8722/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R588-R589)

stnguyen90: Oops! Updated!

byawitz (approved):

PR #8728, opened 2024-09-27
Ignore junction tables

fogelito:
When deleting a project with many2many relations attributes with junctions tables inside _metadata collection.
While iterating _metadata collections which includes junction tables. Trying to delete them will trigger an Exception  Collection not found which will terminate the script leaving orphan collections not deleted and orphan collections inside _metadata.
This happens because if the collection or related collection holding the many2many attribute is deleted, the junction table is also deleted on the fly.


fogelito:
[

](url)

christyjacob4 (approved):

PR #8747, opened 2024-10-02
Feat docker geo

lohanidamodar:
## What does this PR do?

- Add Docker GEO support

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/init.php
```diff
@@ -1622,6 +1624,76 @@ function getDevice($root): Device
...
+        $record = $geodbClient->fetch("/ips/{$ip}", Client::METHOD_GET);
+        $record = $record->json();
+    } catch (Throwable $th) {
+        Console::error($th->getMessage());
+        Console::error($th->getTraceAsString());
+        $record = null;
+    }
+
+    return $record;
+});
```
Meldiron, line 1650: Missing injection `['ip', 'geodbClient']`

app/init.php
```diff
@@ -1622,6 +1624,76 @@ function getDevice($root): Device
...
+}, ['request']);
+
+App::setResource('geodbClient', function () {
+    $client = new Client();
+    $client->addHeader('Authorization', 'Bearer ' . System::getEnv('_APP_GEO_SECRET'));
+    $client->setBaseUrl('http://appwrite-geo/v1');
+    return $client;
+});
+
+App::setResource('geoRecordDocker', function ($ip, $geodbClient) {
```
Meldiron, line 1638: Let's incude types, likely `string $ip, Client $geodbClient`

app/init.php
```diff
@@ -1622,6 +1624,76 @@ function getDevice($root): Device
...
+    $record = null;
+    try {
+        $record = $geodbClient->fetch("/ips/{$ip}", Client::METHOD_GET);
+        $record = $record->json();
+    } catch (Throwable $th) {
```
Meldiron, line 1643: Do we have logger in geoDB HTTP server? If not, I think we should. Logs like these will get lost, so best if geoDB container directly reports to Sentry

app/init.php
Meldiron, line 1638: The name of this resource is weird to me. Can it be `geodbRecord` instead?

app/init.php
```diff
@@ -1622,6 +1624,76 @@ function getDevice($root): Device
...
+        Console::error($th->getMessage());
+        Console::error($th->getTraceAsString());
+        $record = null;
+    }
+
+    return $record;
+});
+
+App::setResource('geoRecordEmbeded', function ($ip, $geodb) {
```
Meldiron, line 1652: Docker will always be there, right? Can we not have Embeded version anymore?

app/init.php
lohanidamodar: Yes, geoDB server has logger integrated.

app/init.php
lohanidamodar: We want to fallback in case the docker fails due to any issues like network, connection or other unknown issues.

app/init.php
lohanidamodar: updated.

app/init.php
lohanidamodar: updated

app/init.php
```diff
@@ -1629,6 +1631,76 @@ function getDevice(string $root, string $connection = ''): Device
...
+    } catch (Throwable $th) {
+        Console::error($th->getMessage());
+        Console::error($th->getTraceAsString());
+        $record = null;
+    }
+
+    return $record;
+}, ['ip', 'geodbClient']);
+
+App::setResource('geoRecordEmbeded', function (string $ip, Reader $geodb) {
```
eldadfux, line 1659: I'm a bit confused between the roles of the different objects. I'm pretty sure everything can be much simpler.

src/Appwrite/Utopia/Fetch/Client.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace Appwrite\Utopia\Fetch;
+
+use Utopia\Fetch\Client as UtopiaClient;
+use Utopia\Fetch\Response;
+
+class Client extends UtopiaClient
```
eldadfux, line 8: Can we use utopia/fetch instead? I think @Meldiron created a new improved version of it.

eldadfux (changes requested):

src/Appwrite/Utopia/Fetch/Client.php
lohanidamodar: I'll see if we can set base URL etc

eldadfux: Isn't this possible in utopia/fetch?? Maybe we should add it? Not a blocker, but do ask @Meldiron about it.

lohanidamodar:
closed in favor of
https://github.com/appwrite/appwrite/pull/10824

PR #8766, opened 2024-10-06
remove audits deletion

shimonewman:


eldadfux (approved):

stnguyen90 (commented):
For completeness, we should probably also remove:

https://github.com/appwrite/appwrite/blob/a49c3a33f0fd831423afa7a0b53df2c5d709fc2b/src/Appwrite/Platform/Workers/Deletes.php#L123-L125

and 

https://github.com/appwrite/appwrite/blob/a49c3a33f0fd831423afa7a0b53df2c5d709fc2b/src/Appwrite/Platform/Workers/Deletes.php#L726-L733

stnguyen90 (approved):

PR #8787, opened 2024-10-10
Reduce builds usage storage

shimonewman:
Added reduction of build storage values on deployments deletion, from inf builds usage metrics, via usage aggregation worker.


vermakhushboo:
@shimonewman is it possible to add some tests for this flow?

PR #8790, opened 2024-10-11
fix: don't run disabled functions in schedules

basert:
Skip disabled functions for getting executed on schedules or events.

### DIFF
--- a/src/Appwrite/Platform/Workers/Functions.php
+++ b/src/Appwrite/Platform/Workers/Functions.php
@@ -130,6 +130,11 @@ class Functions extends Action
                     if (!array_intersect($events, $function->getAttribute('events', []))) {
                         continue;
                     }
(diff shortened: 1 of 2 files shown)


src/Appwrite/Platform/Workers/Functions.php
```diff
@@ -130,6 +130,11 @@ public function action(Message $message, Database $dbForProject, Func $queueForF
                     if (!array_intersect($events, $function->getAttribute('events', []))) {
                         continue;
                     }
+                    // filter disabled functions from getting executed.
+                    if (!$function->getAttribute('enabled')) {
```
christyjacob4, line 134:
```suggestion
                    if ($function->getAttribute('enabled', true) === false) {
```
this would be a more strict check.

PR #8797, opened 2024-10-14
Feat migration

abnegate:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/http.php
```diff
@@ -93,7 +94,7 @@
         try {
             Console::success('[Setup] - Creating database: appwrite...');
             $dbForConsole->create();
-        } catch (\Throwable $e) {
```
christyjacob4, line 96: why make it specific to Duplicate? What if there are some other kinds of exceptions? Its gonna be uncaught right?

abnegate: Yep but if it's something else, we shouldn't be swallowing the exception and just printing that the table exists

christyjacob4 (approved):

PR #8802, opened 2024-10-15
Fix max queries values

fogelito:
Add a global const for the number of values in the Query values .

We will set this in the Database layer so that the same value can be used in Utopia validators and Appwrite as well.

Api Params in Base.php and later on in the find Method validation as well.


app/init.php
```diff
@@ -130,6 +130,7 @@
 const APP_DATABASE_ATTRIBUTE_FLOAT_RANGE = 'floatRange';
 const APP_DATABASE_ATTRIBUTE_STRING_MAX_LENGTH = 1_073_741_824; // 2^32 bits / 4 bits per char
 const APP_DATABASE_TIMEOUT_MILLISECONDS = 15_000;
+const APP_DATABASE_QUERY_VALUES = 500;
```
abnegate, line 133: Let's move both of these to env vars

abnegate, line 133: Let's name this `APP_DATABASE_QUERY_MAX_VALUES`

composer.json
```diff
@@ -51,7 +51,7 @@
         "utopia-php/cache": "0.10.*",
         "utopia-php/cli": "0.15.*",
         "utopia-php/config": "0.2.*",
-        "utopia-php/database": "0.53.6",
+        "utopia-php/database": "dev-0.53.x-max-query-size as 0.53.6",
```
abnegate, line 54: Released 0.53.7 now

PR #8812, opened 2024-10-17
feat: allow blocking based on resource attributes

basert:


app/controllers/api/databases.php
christyjacob4, line 1: We might want to add a resourceType `project` also in the projects.php file and project.php file.
Since when we access the /projects API - `X-appwrite-project` would be set to console and we would be unable to enforce blocks on console usage 🤔

app/controllers/api/messaging.php
```diff
@@ -143,6 +144,7 @@
     ->label('audits.resource', 'provider/{response.$id}')
     ->label('event', 'providers.[providerId].create')
     ->label('scope', 'providers.write')
+    ->label('resourceType', 'messaging')
```
christyjacob4, line 147: It would be useful to have more granularity for messaging. I think the resource types here need to be `providers` , `topics`, `subscribers` and `messages` ?

app/init.php
christyjacob4, line 1: Do we also need to inject this resource into the workers and CLI tasks as well ?

christyjacob4 (approved):

TorstenDittmann (approved):

PR #8826, opened 2024-10-20
Memcached for `/view` endpoint

ItzNotABug:
## What does this PR do?

Helps quite a bit if we already have a medium to big file in cache so we don't have to read the file contents again and again. The logic might not be 100% correct but gives a gist of how we could improve the file delivery.

**Note**: Its possible that speeds differ also due to the localhost fetch, but overall still a good perf. difference.

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


ItzNotABug:
Update: I found that [MemoryCacheAdapter](https://github.com/utopia-php/cache/blob/master/src/Cache/Adapter/Memory.php) is much more performant but not sure if its really scalable on Cloud though.

PR #8838, opened 2024-10-22
feat: move certificate generation to Adapter

basert:
Move certificate generation to Adapter pattern. This allows to inject different adapters and thus allows for different certificate providers.

Move all certificate handling to adapter and cleanup custom logic in delete worker.


src/Appwrite/Certificates/Adapter/LetsEncrypt.php
```diff
@@ -0,0 +1,130 @@
...
+class LetsEncrypt implements Adapter
+{
+    private string $email;
+
+    public function __construct()
+    {
+        $this->email = System::getEnv('_APP_EMAIL_CERTIFICATES', System::getEnv('_APP_SYSTEM_SECURITY_EMAIL_ADDRESS'));
+        if (empty($this->email)) {
+            throw new Exception('You must set a valid security email address (_APP_EMAIL_CERTIFICATES) to issue an SSL certificate.');
+        }
```
christyjacob4, line 22: typically we don't use environment variables inside the constructor. Since this is like a library and a library should be agnostic of application specifics like this env variable. This also makes it harder to decouple the library later on.  

It's be better if we pass these as arguments to the constructor.

src/Appwrite/Platform/Workers/Certificates.php
```diff
@@ -48,7 +48,11 @@ public function __construct()
             ->inject('queueForEvents')
             ->inject('queueForFunctions')
             ->inject('log')
-            ->callback(fn (Message $message, Database $dbForConsole, Mail $queueForMails, Event $queueForEvents, Func $queueForFunctions, Log $log) => $this->action($message, $dbForConsole, $queueForMails, $queueForEvents, $queueForFunctions, $log));
+            ->inject('adapterForCertificates')
+            ->callback(
+                fn (Message $message, Database $dbForConsole, Mail $queueForMails, Event $queueForEvents, Func $queueForFunctions, Log $log, AdapterProvider $adapterForCertificates) =>
```
christyjacob4, line 53: Why do we need an AdapterProvider wrapper? The type of `$adapterForCertificates` can just be `Appwrite\Certificates\Adapter` right? Just like we do for [`Database`](https://github.com/utopia-php/database/blob/main/src/Database/Database.php)

src/Appwrite/Certificates/Adapter/LetsEncrypt.php
```diff
@@ -0,0 +1,126 @@
...
+
+    public function deleteCertificate(string $domain): void
+    {
+        $directory = APP_STORAGE_CERTIFICATES . '/' . $domain;
+        $checkTraversal = realpath($directory) === $directory;
+
+        if ($checkTraversal && is_dir($directory)) {
+            // Delete files, so Traefik is aware of change
+            array_map('unlink', glob($directory . '/*.*'));
+            rmdir($directory);
```
eldadfux, line 120: This is a good chance to start using Utopia/Storage for all file interactions to better follow our conventions. Missing functions can be added.

src/Appwrite/Certificates/AdapterProvider.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace Appwrite\Certificates;
+
+class AdapterProvider
```
eldadfux, line 5: Our naming should be consistent with our patterns. Let's always prefer a single word name where possible and avoid redundant parts in names. AdapterProvider can just be Adapter or Provider.

app/init.php
```diff
@@ -1511,6 +1513,19 @@ function (mixed $value) {
     return new Cache(new Sharding($adapters));
 }, ['pools']);
 
+App::setResource('letsEncryptCertificateAdapter', function () {
```
eldadfux, line 1516: Very complex naming conventions: letsEncryptCertificateAdapter, adapterForCertificates. If we don't use letsEncryptCertificateAdapter resource anywhere directly, no point in having it as a separate resource, unnecessary noise. `adapterForCertificates` can just be `certificates`

src/Appwrite/Certificates/AdapterProvider.php
```diff
@@ -0,0 +1,19 @@
...
+{
+    private $decisionMaker;
+
+    public function __construct(callable $decisionMaker)
```
eldadfux, line 9: Not sure why the constructor receive a callback in the class you shared, API/class interface seems very loose this way.

src/Appwrite/Certificates/Adapter.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Appwrite\Certificates;
+
+use Utopia\Logger\Log;
+
+interface Adapter
```
eldadfux, line 7: Not clear why we need both adapter class and adapter provider class, seems over complex.

eldadfux (dismissed):

christyjacob4 (approved):

Issue #8785, opened 2024-10-10
🐛 Bug Report: All Users.CreateUser... endpoints return a vague server error is null name is provided


pingu2k4:
### 👟 Reproduction steps

POST to /v1/users or any of the "create user with ___ password" endpoints
Create body with a defined UserId, and a `name: null`.
Observe the vague server error

### 👍 Expected behavior

No server error, should continue creating the user with no name. Name is optional anyway, so specifying it as null should not create this error.

### 👎 Actual Behavior

![Image](https://github.com/user-attachments/assets/835e2c3f-dd8b-453d-8d61-2c6a85a0d1ec)

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

Windows

### 🧱 Your Environment

I am developing my own SDK, hence running into this issue. I can work around it of course, but there shouldn't be a reason for this to occur in the first place. I have replicated it using postman to eliminate any possible issues at a deeper level within my own SDK.

Wasn't sure if you would care about this sort of issue, but received no answer on Discord as to whether to file this or not.

### 👀 Have you spent some time to check if this issue has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

stnguyen90:
Looks like the error in the backend is:

```
{closure}(): Argument #5 ($name) must be of type string, null given, called in /usr/src/code/vendor/utopia-php/framework/src/App.php on line 531
```

because `$name` is supposed to be a string and not null here:

https://github.com/appwrite/appwrite/blob/a49c3a33f0fd831423afa7a0b53df2c5d709fc2b/app/controllers/api/users.php#L207

One really odd thing is why the Text validator here:

https://github.com/appwrite/appwrite/blob/a49c3a33f0fd831423afa7a0b53df2c5d709fc2b/app/controllers/api/users.php#L201

doesn't prevent null 🧐

vaifai:
Hey, I am new to open source would really like to contribute to this issue. I have worked with PHP before and I could leverage my experience to get a hang about the issue.

pingu2k4:
Have just noticed the same issue is also present on [Create User Target](https://appwrite.io/docs/references/1.6.x/server-rest/users#createTarget) also - However this effects both the `name` property as well as the `providerId`. If either are sent as null the same thing happens.

[...]

PR #8850, opened 2024-10-24
fix: parse null value as empty string instead of throwing error

Shaik-Sirajuddin:
## What does this PR do?

Name is optional while creating user ( as per api docs ) , 
This PR makes user to be created with default name '' ( empty string ) , instead of throwing a generic error 
when null is passed for name

Applicable for all create user with ___ password enpoints 

## Test Plan

- Make post request one of create user endpoint with name as null
- User is created with name = ''
- A generic_error was being thrown before fix

- image

Before Fix : 
image

After Fix : 
Screenshot from 2024-10-24 15-16-59

## Related PRs and Issues
- #8785 
- 
## Checklist

### DIFF
--- a/app/controllers/api/users.php
+++ b/app/controllers/api/users.php
@@ -50,12 +50,16 @@ use Utopia\Validator\Text;
 use Utopia\Validator\WhiteList;
 
 /** TODO: Remove function when we move to using utopia/platform */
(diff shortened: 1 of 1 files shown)


HarshMN2345:

PR #8859, opened 2024-10-25
Feat sites

vermakhushboo:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added full support for "Sites" with APIs for site deployments, duplication, downloads, status updates, and usage metrics.
  - Extended custom domain management with proxy rules for sites, APIs, functions, and redirects including multi-record DNS verification.
  - Introduced modular HTTP APIs for functions, sites, deployments, executions, variables, runtimes, specifications, and templates.
  - Added preview authorization and multi-resource routing with enhanced error and 404 pages.
  - Added static analysis workflow and PHPStan configuration for code quality.

- **Improvements**
  - Renamed and restructured environment variables for compute and sites; expanded domain and runtime configurations.
  - Updated database schemas for detailed deployment/build metadata, site management, and optimized indexing.
  - Enhanced role and scope permissions to cover sites and logging.
  - Improved metrics and usage tracking for both functions and sites.
  - Upgraded Docker images, service definitions, and persistent storage volumes for new features.
  - Refined domain validation with flexible multi-record DNS checks.

- **Bug Fixes**
  - Fixed domain validation and DNS record checks for custom domains.
  - Corrected documentation and API descriptions for clarity.

- **Documentation**
  - Expanded contributing guidelines with new module structure and coding conventions.
  - Added and updated reference documentation for new and existing endpoints.

- **Chores**
  - Added configuration files for frameworks, templates, and runtimes.
  - Updated dependencies and scripts to support static analysis and new features.

- **Style**
  - Improved consistency in naming, error handling, and API response formats.

- **Tests**
  - Enhanced CI workflows for improved service readiness and test coverage.

This release significantly enhances web app hosting and deployment capabilities by introducing sites support, advanced domain management, and improved developer experience.


app/config/function-templates.php
eldadfux, line 1: Let's just have a templates directory with functions.php and sites.php. Even if we do want to keep this flat our naming conventions would suggest to use a prefix so they're still grouped together.

app/config/collections/projects.php
```diff
@@ -797,6 +797,364 @@
...
+                'format' => '',
+                'size' => 2048,
+                'signed' => true,
+                'required' => true,
+                'default' => '',
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('adapter'), // ssr or static
```
eldadfux, line 1073: is adapter really the best name here? Should we consider `rendering`?

app/config/site-templates.php
eldadfux, line 1: Same as the comment for the functions template file. Why do we use all capitals for the keys? we don't do that anywhere else.

app/config/template-runtimes.php
eldadfux, line 1: Same comment about the templates directory,

eldadfux: Where is this file even needed? what are runtime templates vs functions templates?

app/controllers/shared/api.php
```diff
@@ -165,6 +165,15 @@
...
                 ->addMetric(METRIC_FUNCTIONS, $value); // per project
 
+            if ($event === Database::EVENT_DOCUMENT_DELETE) {
+                $queueForUsage
+                    ->addReduce($document);
+            }
+            break;
+        case $document->getCollection() === 'sites':
+            $queueForUsage
+                ->addMetric(METRIC_SITES, $value); // per project
```
eldadfux, line 175: Keep in mind we'll be deprecating this event listener and start counting resources with a DB count method instead of during runtime.

app/views/install/compose.phtml
```diff
@@ -86,7 +86,7 @@ $image = $this->getParam('image', '');
       - _APP_OPTIONS_ABUSE
       - _APP_OPTIONS_ROUTER_PROTECTION
       - _APP_OPTIONS_FORCE_HTTPS
-      - _APP_OPTIONS_FUNCTIONS_FORCE_HTTPS
+      - _APP_OPTIONS_COMPUTE_FORCE_HTTPS
```
eldadfux, line 89: This shouldn't have anything to do with compute. Sounds more like a proxy/router/edge feature.

app/init.php
```diff
@@ -1587,6 +1609,10 @@ function (mixed $value) {
     return getDevice(APP_STORAGE_UPLOADS . '/app-' . $project->getId());
 }, ['project']);
 
+App::setResource('deviceForSites', function ($project) {
+    return getDevice(APP_STORAGE_SITES . '/app-' . $project->getId());
```
eldadfux, line 1613: Should we store sites build on a separate storage? Might make thing unnecessarily complex for the edge. If deployments and build are using the same collection for sites and functions so we can use the same storage for all compute.

docker-compose.yml
```diff
@@ -110,7 +111,7 @@ services:
       - _APP_OPTIONS_ABUSE
       - _APP_OPTIONS_ROUTER_PROTECTION
       - _APP_OPTIONS_FORCE_HTTPS
-      - _APP_OPTIONS_FUNCTIONS_FORCE_HTTPS
+      - _APP_OPTIONS_COMPUTE_FORCE_HTTPS
```
eldadfux, line 114: Same comment about the namespace here.

src/Appwrite/Event/Event.php
```diff
@@ -24,6 +24,9 @@ class Event
     public const FUNCTIONS_QUEUE_NAME = 'v1-functions';
     public const FUNCTIONS_CLASS_NAME = 'FunctionsV1';
 
+    public const SITES_QUEUE_NAME = 'v1-sites';
```
eldadfux, line 27: why do we need a queue for sites?

src/Appwrite/Extend/Exception.php
```diff
@@ -151,6 +151,11 @@ class Exception extends \Exception
...
     public const GENERAL_PROVIDER_FAILURE          = 'general_provider_failure';
 
+    /** Sites */
+    public const SITE_NOT_FOUND                    = 'site_not_found';
+    public const SITE_FRAMEWORK_UNSUPPORTED        = 'site_framework_unsupported';
+    public const SITE_TEMPLATE_NOT_FOUND           = 'site_template_not_found';
+
     /** Functions */
     public const FUNCTION_NOT_FOUND                = 'function_not_found';
     public const FUNCTION_RUNTIME_UNSUPPORTED      = 'function_runtime_unsupported';
```
eldadfux, line 161:
```suggestion
    public const FUNCTION_RUNTIME_UNSUPPORTED      = 'function_runtime_unsupported';
```

Should this actually be not found?

src/Appwrite/Platform/Modules/Compute/Base.php
```diff
@@ -0,0 +1,191 @@
...
+use Utopia\Database\Validator\Authorization;
+use Utopia\Platform\Action;
+use Utopia\Swoole\Request;
+use Utopia\System\System;
+use Utopia\VCS\Adapter\Git\GitHub;
+use Utopia\VCS\Exception\RepositoryNotFound;
+
+class Base extends Action
+{
+    public function redeployVcsFunction(Request $request, Document $function, Document $project, Document $installation, Database $dbForProject, Build $queueForBuilds, Document $template, GitHub $github)
```
eldadfux, line 22: By all logic, this shouldn't be:

1. A functions in a controller
2. Used as base Action for *all* other actions
3. A public method

True to both `redeployVcsFunction ` and `redeployVcsSite` also this defeats the all purpose of the abstraction, both methods here don't make sense at all.

eldadfux (changes requested):

app/config/function-templates.php
Meldiron: Following same pattern as `collections.php`, fully understandable.

app/config/frameworks/specifications.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+use Appwrite\Sites\Specification;
+
+return [
+    Specification::S_05VCPU_512MB => [
```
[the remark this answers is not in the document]
Meldiron: We only have 1 specs now

app/config/collections/projects.php
Meldiron: A framework can support multiple of adapter, for example SSR and static.

Due to that, we need a key for it `frameworks[0].adapters[0].buildCommand`.

Notice we dont use `adapter` there, but `adapters`, since it's array.

I couldn't find a word better than adapter that would have good plural.

`renderings` sounds bad, more technically correct is `strategies`. But then `site.adapter` sounds much better than `site.strategy`. Strategy in this sense (to me) sounds more like deployment term - strategy to decide when to activate deployment on production

For those reasons I prefer adapter. If you have more suggestions, open to discuss

app/config/function-templates.php
Meldiron: Done, we now have `templates` folder with `site.php` and `function.php`

app/config/site-templates.php
Meldiron: Those uppercase keys are internal only, a helper to make this file much shorter. Uppercase here means to feel like const. We do same in functions `$templateRuntimes['GO']` ( https://github.com/appwrite/appwrite/blob/9893f77e6dcc268fbb73dea57d3fa1bfc6e0c94f/app/config/templates/function.php#L43 )

app/config/template-runtimes.php
Meldiron: Let's talk about this on call

app/controllers/shared/api.php
Meldiron: Damodar mentioned it's already implemented in statsResources too, so it's okay.
He also takes lead in removing some of those later on, and is aware of sites addition.

Marking this as close as we have proper vision on it, and doesnt require immidiate actioning

app/views/install/compose.phtml
Meldiron: Renamed to `_APP_OPTIONS_ROUTER_FORCE_HTTPS` (router, because we already got `_APP_OPTIONS_ROUTER_PROTECTION`)

docker-compose.yml
Meldiron: Fixed too, `_APP_OPTIONS_ROUTER_FORCE_HTTPS`

src/Appwrite/Event/Event.php
Meldiron: To confuse enemy

src/Appwrite/Extend/Exception.php
Meldiron: Agree, but we do it in 16 errors already. I dont want to change error types, as that would be breaking change for those doing custom messages for exceptions.
We dont currently have system in-place for exception type backwards compatibility.

I agree to have project to improve this and add backwards compatibility, but separate from Sites G1

app/init.php
Meldiron: Notice devices are same, just use different paths. Thanks to that, we faced no issues on executor. I expect the same worry-free support on edge.

src/Appwrite/Platform/Modules/Compute/Base.php
Meldiron: Agree, same here: https://github.com/appwrite/appwrite/blob/9893f77e6dcc268fbb73dea57d3fa1bfc6e0c94f/app/controllers/api/vcs.php#L63

We also never addressed the fact that site deployment creation endpoint and function deployment creation are 99% same. They should also be much much simpler.

I dont think we have time to focus on this for 1.7 release. I vote to improve this later. Having structure that allows us to fix it is good first step - better than what we had before.

app/config/template-runtimes.php
Meldiron: It's helper object used by functions templates, and framework. Wit this, we can just say "NODE" and it automatically knows to mention all Node versions in the config.

Agree this is not the nicest solution; not sure what would be the best. Mini-library? Class in config?
We might need discussion and decision before I can improve this.

app/init.php
Meldiron: Luke mentioned edge already knows about site and function, so he is not worired.

app/config/services.php
```diff
@@ -173,12 +173,25 @@
         'optional' => false,
         'icon' => '',
     ],
+    'sites' => [
+        'key' => 'sites',
+        'name' => 'Sites',
+        'subtitle' => 'The Sites Service allows you view, create and manage your web applications.',
+        'description' => '/docs/services/sites.md',
```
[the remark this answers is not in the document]
Meldiron: Adytia will do in separate PR

eldadfux (approved):

PR #8891, opened 2024-10-29
Fix XDebug Extension not being removed

PineappleIOnic:
XDebug's Zend API version changed with 8.3 so the xdebug.so file was not being correctly removed. This PR fixes the Zend API folder version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -92,10 +92,10 @@ RUN chmod +x /usr/local/bin/doctor && \
 RUN mkdir -p /etc/letsencrypt/live/ && chmod -Rf 755 /etc/letsencrypt/live/
 
 # Enable Extensions
-RUN if [ "$DEBUG" == "true" ]; then cp /usr/src/code/dev/xdebug.ini /usr/local/etc/php/conf.d/xdebug.ini; fi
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -94,8 +94,8 @@ RUN mkdir -p /etc/letsencrypt/live/ && chmod -Rf 755 /etc/letsencrypt/live/
 # Enable Extensions
 RUN if [ "$DEBUG" == "true" ]; then cp /usr/src/code/dev/xdebug.ini /usr/local/etc/php/conf.d/xdebug.ini; fi
 RUN if [ "$DEBUG" == "true" ]; then mkdir -p /tmp/xdebug; fi
-RUN if [ "$DEBUG" = "false" ]; then rm -rf /usr/src/code/dev; fi
-RUN if [ "$DEBUG" = "false" ]; then rm -f /usr/local/lib/php/extensions/no-debug-non-zts-20220829/xdebug.so; fi
+RUN if [ "$DEBUG" == "false" ]; then rm -rf /usr/src/code/dev; fi
```
abnegate, line 97: Dockerfile RUN commands use sh, in which `==` is undefined; these should all be single equals instead of double. Double equals is a bash extension

abnegate (approved):

Issue #2365, opened 2021-10-28
🐛 Bug Report: OAuth2 does not add port number to redirect uri


dro-1:
### 👟 Reproduction steps

When I started my appwrite server, I set it to run on port 3000 for http and left all other values as their defaults.
I added Google OAuth by following all the required steps:
- I created the Client ID and set its redirect URI as http://localhost:3000/v1/account/sessions/oauth2/callback/google/[PROJECT_ID]
- I also added the App ID and App Secret to Appwrite
- I was building a React App and it's endpoint was 'http://localhost:3001/' as it was running on port 3001
- The react app had the appwrite sdk installed so I called the function for creating a new Oauth Session passing 'google' as the provider and the success and failure links

### 👍 Expected behavior

I expected the redirect URI to have port 3000

### 👎 Actual Behavior

I get an error saying:
 
![image](https://user-images.githubusercontent.com/45892107/139344396-92177787-5a60-4baa-82ad-1b691966566c.png)

and I noticed that the redirect_uri didn't have the port 3000 on it.

One of the maintainers helped track down the issue to line 292 in [this file](https://github.com/appwrite/appwrite/blob/master/app/controllers/api/account.php#L292)

### 🎲 Appwrite version

Different version (specify in environment)

### 💻 Operating system

Windows

### 🧱 Your Environment

I'm using version 0.11.0

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/appwrite/blob/HEAD/CODE_OF_CONDUCT.md)

eldadfux:
I have assigned you to this issue, let me know if you need any help.

dro-1:
I already opened a PR on it but couldn't write the proper tests as I'm a PHP noob.
Check the PR #2368

skyface753:
Hello, I have the same problem. Is there any solution?

stnguyen90:
@skyface753, for now, I recommend exposing Appwrite on the standard ports. If you need other apps on the same server on standard ports, you can out a reverse proxy in front.

stnguyen90:
@skyface753, for now, I recommend exposing Appwrite on the standard ports. If you need other apps on the same server on standard ports, you can out a reverse proxy in front.

kilroyjones:
[...]

PR #8896, opened 2024-10-29
Fix 2365 Oauth2 does not add port number to uri

AryanSwaroop:
## What does this PR do?

This pr if detects a port then adds the port number to the redirrect uri's so it wouldn't cause any mismatch when using third party services like google auth. Previously , uri was only localhost but after this it will be localhost:8080.

## Test Plan

1) Previously , it was showing uri mismatch and error while using google OAuth.
2) After changing the code , it redirects to success link without any problem.

## Related PRs and Issues

- #2365

## Checklist

- [ yes ] Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [ yes ] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?


stnguyen90:
Sorry we had to fix another issue and ended up fixing this at the same time.

PR #8913, opened 2024-11-01
feat: list memberships as client

loks0n:


app/controllers/api/teams.php
```diff
@@ -790,27 +791,30 @@
...
-                $totpEnabled = $totp && $totp->getAttribute('verified', false);
-                $emailEnabled = $user->getAttribute('email', false) && $user->getAttribute('emailVerification', false);
-                $phoneEnabled = $user->getAttribute('phone', false) && $user->getAttribute('phoneVerification', false);
+        $hideSensitiveFields = $project->getAttribute('auths', [])['teamHideSensitiveFields'] ?? false;
 
-                if (!$totpEnabled && !$emailEnabled && !$phoneEnabled) {
-                    $mfa = false;
+        $memberships = array_map(function ($membership) use ($dbForProject, $team, $hideSensitiveFields) {
+            $user = $dbForProject->getDocument('users', $membership->getAttribute('userId'));
+            $membership->setAttribute('teamName', $team->getAttribute('name'));
```
christyjacob4, line 798: The DB library has a a select query , so we should use that instead of fetching all attributes and filtering right? 🤔 
https://github.com/utopia-php/database/blob/main/src/Database/Query.php#L449-L452

app/controllers/api/projects.php
```diff
@@ -648,6 +649,37 @@
         $response->dynamic($project, Response::MODEL_PROJECT);
     });
 
+App::patch('/v1/projects/:projectId/auth/session-alerts')
```
christyjacob4, line 652: the endpoint needs to be updated

app/controllers/api/projects.php
```diff
@@ -111,6 +111,7 @@
             'personalDataCheck' => false,
             'mockNumbers' => [],
             'sessionAlerts' => false,
+            'teamHideSensitiveFields' => false,
```
christyjacob4, line 114: there's also the project response model where this needs to be updated

christyjacob4, line 114:
```suggestion
              'teamsShowSensitiveFields' => true,
```
To make it easier logically, it's better to avoid inverse logic.

app/controllers/api/projects.php
```diff
@@ -648,6 +649,37 @@
...
+    ->action(function (string $projectId, bool $alerts, Response $response, Database $dbForConsole) {
+
+        $project = $dbForConsole->getDocument('projects', $projectId);
+
+        if ($project->isEmpty()) {
+            throw new Exception(Exception::PROJECT_NOT_FOUND);
+        }
+
+        $auths = $project->getAttribute('auths', []);
+        $auths['teamShowSensitiveFields'] = $alerts;
```
christyjacob4, line 675: we need other update this `$alerts` variable

tests/e2e/Services/Teams/TeamsBaseClient.php
```diff
@@ -130,6 +130,67 @@ public function testGetTeamMemberships($data): array
         $this->assertEmpty($response['body']['memberships']);
         $this->assertEquals(0, $response['body']['total']);
 
+        /**
+         * Update project settings to hide sensitive fields
+         */
+        $projectId = $this->getProject()['$id'];
+        $response = $this->client->call(Client::METHOD_PATCH, '/projects/' . $projectId . '/auth/session-alerts', array_merge([
```
christyjacob4: lets update the URL as well

app/controllers/api/projects.php
loks0n: Sure although usage becomes awkward in cases where it's null

app/controllers/api/projects.php
```diff
@@ -648,6 +649,37 @@
...
+    ->groups(['api', 'projects'])
+    ->label('scope', 'projects.write')
+    ->label('sdk.auth', [APP_AUTH_TYPE_ADMIN])
+    ->label('sdk.namespace', 'projects')
+    ->label('sdk.method', 'updateTeamsShowSensitiveFields')
+    ->label('sdk.response.code', Response::STATUS_CODE_OK)
+    ->label('sdk.response.type', Response::CONTENT_TYPE_JSON)
+    ->label('sdk.response.model', Response::MODEL_PROJECT)
+    ->param('projectId', '', new UID(), 'Project unique ID.')
+    ->param('teamsShowSensitiveFields', true, new Boolean(true), 'Set to true to hide sensitive fields from team members.')
```
christyjacob4, line 663:
```suggestion
    ->param('enabled', true, new Boolean(true), 'Set to true to hide sensitive fields from team members.')
```

christyjacob4 (approved):

PR #8922, opened 2024-11-05
Improve compression param checks

lohanidamodar:
## What does this PR do?
- Updates create bucket and update bucket endpoint
- Make compression param validator strict
- makes compression param nullable string `?string` in action
- Assign default value for compression if null inside action

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/storage.php
```diff
@@ -68,19 +68,19 @@
...
     ->param('allowedFileExtensions', [], new ArrayList(new Text(64), APP_LIMIT_ARRAY_PARAMS_SIZE), 'Allowed file extensions. Maximum of ' . APP_LIMIT_ARRAY_PARAMS_SIZE . ' extensions are allowed, each 64 characters long.', true)
-    ->param('compression', Compression::NONE, new WhiteList([Compression::NONE, Compression::GZIP, Compression::ZSTD]), 'Compression algorithm choosen for compression. Can be one of ' . Compression::NONE . ',  [' . Compression::GZIP . '](https://en.wikipedia.org/wiki/Gzip), or [' . Compression::ZSTD . '](https://en.wikipedia.org/wiki/Zstd), For file size above ' . Storage::human(APP_STORAGE_READ_BUFFER, 0) . ' compression is skipped even if it\'s enabled', true)
+    ->param('compression', Compression::NONE, new WhiteList([Compression::NONE, Compression::GZIP, Compression::ZSTD], true), 'Compression algorithm choosen for compression. Can be one of ' . Compression::NONE . ',  [' . Compression::GZIP . '](https://en.wikipedia.org/wiki/Gzip), or [' . Compression::ZSTD . '](https://en.wikipedia.org/wiki/Zstd), For file size above ' . Storage::human(APP_STORAGE_READ_BUFFER, 0) . ' compression is skipped even if it\'s enabled', true)
```
stnguyen90, line 71: How come this validator doesn't prevent nulls from being passed?

lohanidamodar: because the parameter is optional

stnguyen90: There are 3 different cases:

1. compression is not in the request
2. compression is in the request and set to null
3. compression is in the request and set to one of the values

If compression is not in the request, I would expect `$compression` would end up being the default value because `$optional` is `true`. Otherwise, it would have thrown an error.

In the case that `$compression` is in the request and is `null`, I expect the validation to run based on [this](https://github.com/utopia-php/http/blob/a7f577540a25cb90896fef2b64767bf8d700f3c5/src/App.php#L618-L620) code.

stnguyen90: Ah, I see. It's because of [this](https://github.com/utopia-php/servers/blob/fd5c8d32778f265256c1936372a071b944f5ba8a/src/Servers/Base.php#L293-L295). Honestly, I don't think that should be there.

stnguyen90 (approved):

PR #8939, opened 2024-11-06
Delete rules when site or deployment is deleted

vermakhushboo:
## What does this PR do?

Delete rules, deployments, builds, variables etc when deleting a site.
Delete rule when deleting a deployment.

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -1984,4 +1985,25 @@ public function testFunctionLogging()
 
         $this->cleanupFunction($functionId);
     }
+
+    /**
+     * @depends testCreateFunction
+     */
+    public function testUniqueSubdomain(array $data): void
```
Meldiron, line 1992: We can keep this logic in `testCreateFunction`, we can make section thefer for:
```
        /**
         * Test for FAILURE
         */
```

Its pattern we already use in tests

tests/e2e/Services/Sites/SitesCustomServerTest.php
```diff
@@ -0,0 +1,155 @@
...
+        $this->assertEquals(201, $variable2['headers']['status-code']);
+        $this->assertEquals(201, $variable3['headers']['status-code']);
+
+        return [
+            'siteId' => $siteId,
+        ];
+    }
+
+    /**
+     * @depends testCreateSite
```
Meldiron, line 72: Lets rewrite tests in this file to have no dependencies. 
If you need to list, or get, or delete, inside the test you can make new site.

this way we prevent need to use @depends

vermakhushboo: In that case, we'd also need to delete the resources we create in each test at the end of it so that it doesn't cause conflicts with other tests.
For example, while testing `listSites`, if I am creating sites for it inside the test itself, it should not list sites from other tests. Do you agree each test should also perform cleanup then?

christyjacob4 (approved):

PR #8959, opened 2024-11-12
Send migration errors individually

PineappleIOnic:
## What does this PR do?

This PR makes it so that migration errors get fired individually against the logger

## Test Plan

I added purposeful exceptions into migration lib locally and tested this works as intended.

## Checklist


src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -403,4 +412,38 @@ function () use ($migration, $transfer, $projectDocument) {
...
+        $log->setVersion(System::getEnv('_APP_VERSION', 'UNKNOWN'));
+        $log->setType(Log::TYPE_ERROR);
+        $log->setMessage($error->getMessage());
+        $log->setAction('appwrite-queue-' . self::getName());
+        $log->addTag('verboseType', get_class($error));
+        $log->addTag('projectId', $this->project->getId() ?? 'n/a');
+        $log->addExtra('file', $error->getFile());
+        $log->addExtra('line', $error->getLine());
+        $log->addExtra('trace', $error->getTraceAsString());
+        $log->addExtra('migrationId', $migration->getId() ?? 'n/a');
```
abnegate, line 435: Let's use empty string instead of n/a

src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -53,13 +57,14 @@ public function __construct()
             ->inject('dbForProject')
             ->inject('dbForConsole')
             ->inject('log')
-            ->callback(fn (Message $message, Database $dbForProject, Database $dbForConsole, Log $log) => $this->action($message, $dbForProject, $dbForConsole, $log));
+            ->inject('logger')
+            ->callback(fn (Message $message, Database $dbForProject, Database $dbForConsole, Log $log, ?Logger $logger) => $this->action($message, $dbForProject, $dbForConsole, $log, $logger));
```
abnegate, line 61: We can remove the $log injection now

abnegate (commented):
Does this stop the "migration failed" logs being pushed as well?

abnegate (commented):
Just saw in CLI scope we have a `logError` resource so that the logging function can be shared between tasks, let's add the same in worker scope and use that here

src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -394,7 +387,23 @@ function () use ($migration, $transfer, $projectDocument) {
                 $destination->error();
                 $source->error();
 
-                throw new Exception('Migration failed');
+                foreach ($source->getErrors() as $error) {
```
abnegate, line 390: Let's only send exceptions that either had a code of 0 or >500

abnegate (approved):

PR #8968, opened 2024-11-13
Feat: Bulk Create Documents

PineappleIOnic:
## What does this PR do?

This PR modifies the existing create document route to allow multiple documents to be created in one API call by adding a 'documents' param.

## Test Plan

Old document create tests will stay in place to ensure we don't break functionality, and new create documents tests will be added to ensure added functionality works as expected around the various special DB features such as Relationships and Permissions.

## Checklist


app/controllers/api/databases.php
```diff
@@ -3068,28 +3105,42 @@ function updateAttribute(
...
 
-        $processDocument($collection, $document);
+        foreach ($documents as $document) {
+            $processDocument($collection, $document);
 
-        $response
-            ->setStatusCode(Response::STATUS_CODE_CREATED)
-            ->dynamic($document, Response::MODEL_DOCUMENT);
+            $document->setAttribute('$databaseId', $database->getId());
+            $document->setAttribute('$collectionId', $collection->getId());
```
abnegate, line 3112: This should be done by processDocument

app/controllers/api/databases.php
```diff
@@ -3035,8 +3065,15 @@ function updateAttribute(
...
+        $queueForEvents
+            ->setParam('databaseId', $databaseId)
+            ->setParam('collectionId', $collection->getId())
+            ->setContext('collection', $collection)
+            ->setContext('database', $database);
+
         // Add $collectionId and $databaseId for all documents
-        $processDocument = function (Document $collection, Document $document) use (&$processDocument, $dbForProject, $database) {
+        $processDocument = function (Document $collection, Document $document) use (&$processDocument, $dbForProject, $database, $queueForEvents) {
+            $document->removeAttribute('$collection');
```
abnegate, line 3076: This should be done by the document response model filter

app/controllers/api/databases.php
```diff
@@ -3068,28 +3105,42 @@ function updateAttribute(
...
+            $response
+                ->setStatusCode(Response::STATUS_CODE_CREATED)
+                ->dynamic(new Document([
+                    'total' => count($documents),
+                    'documents' => $documents,
+                ]), Response::MODEL_DOCUMENT_LIST);
+        } else {
 
+            $document = $documents[0];
+            $processDocument($collection, $document);
```
abnegate, line 3138: Shouldn't need this call here since we called it in the loop already

app/controllers/api/databases.php
```diff
@@ -3023,10 +3032,29 @@ function updateAttribute(
             }
         };
 
-        $checkPermissions($collection, $document, Database::PERMISSION_CREATE);
+        $documents = array_map(function ($document) use ($collection, $permissions, $checkPermissions, $isBulk, $documentId) {
+            $document['$collection'] = $collection->getId();
+            $document['$permissions'] = $permissions;
```
abnegate, line 3037: We should only do this for single documents, for bulk, $permissions should be inside the documents like $id

tests/e2e/Services/Databases/DatabasesBase.php
```diff
@@ -4800,6 +4800,278 @@ public function testTimeout(array $data): void
...
+            'documents' => [
+                [
+                    'number' => 1,
+                ]
+            ],
+        ]);
+
+        $this->assertEquals(400, $response['headers']['status-code']);
+        $this->assertEquals('$id is required inside documents when creating bulk documents', $response['body']['message']);
+    }
```
abnegate, line 4936: Let's test some more failure cases like multiple documents where one has invalid structure

abnegate (approved):

PR #9014, opened 2024-11-21
Fix Spanish email template translations

DH-555:
Again...

As reported on discord, there is an additional dot in the end of some sentences that is not needed.

Also, the template had {{name}} placeholders instead of {{user}}
Changed them so the name is being show

https://discord.com/channels/564160730845151244/1308562407299485767


DH-555:
Closing this as it's integrated in the latest changes: https://github.com/appwrite/appwrite/pull/9072 in order to prevent merge conflicts

PR #9016, opened 2024-11-22
Feat: Safe workers

Meldiron:
## What does this PR do?

Adds safe-worker logic to API servers, separating functions and API request handling to ensure zombie worker cannot occur.

_Zombie worker used to be when a worker was handling a synchronous function execution, but from inside function API request was sent and picked up by same worker. This created deadlock only released by function timeout._

## Test Plan

Existing tests must pass

## Related PRs and Issues

x

## Checklist


stnguyen90:
Tested the node starter with 10K requests with 1K concurrency. A lot of the requests were slow, but at least they all completed successfully
image

stnguyen90 (approved):

PR #9019, opened 2024-11-22
Fix users create session secret

### LINKED ISSUE #8569: 🐛 Bug Report: "Users.createSession()" returns invalid session secret
### 👟 Reproduction steps

1. Create a session via "Users.createSession()"
2. Set the session secret in "Client.setSession()"
3. Call "Accounts.get()"

### 👍 Expected behavior

No error.
Get a valid user object response back.

### 👎 Actual Behavior

**Error => "User (role: guests) missing scope (account)"**

Session secret when creating one  via "Users.createSession()"
image

Session secret when creating one via "Account.createEmailPasswordSession()"
image

The one from "Users.createSession()" is not properly encoded and therefore not returned as JWT.

Users.createSession()
https://github.com/appwrite/appwrite/blob/9301bddd8b161e6f4207e4b86d9908cacbdf9ef0/app/controllers/api/users.php#L1815

Account.createEmailPasswordSession()
https://github.com/appwrite/appwrite/blob/9301bddd8b161e6f4207e4b86d9908cacbdf9ef0/app/controllers/api/account.php#L808

### 🎲 Appwrite version

Version 1.5.x

### 💻 Operating system

Linux

### 🧱 Your Environment

_No response_

stnguyen90:
## What does this PR do?

1. Include at least 1 factor because the minumum number of factors
   required when mfa is disabled is 1.
2. Purge the cached user document to ensure the new session is included
   in subsequent requests for the user.
3. Fix the encoding of the secret to match other parts of the codebase.

Fixes #8569 

## Test Plan

Updated test case

## Related PRs and Issues

- #8569

## Checklist

### DIFF
--- a/app/controllers/api/users.php
+++ b/app/controllers/api/users.php
@@ -1804,6 +1804,7 @@ App::post('/v1/users/:userId/sessions')
                 'provider' => Auth::SESSION_PROVIDER_SERVER,
                 'secret' => Auth::hash($secret), // One way hash encryption to protect DB leak
                 'userAgent' => $request->getUserAgent('UNKNOWN'),
+                'factors' => ['server'],
                 'ip' => $request->getIP(),
                 'countryCode' => ($record) ? \strtolower($record['country']['iso_code']) : '--',
                 'expire' => $expire,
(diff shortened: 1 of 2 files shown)


loks0n (approved):

app/controllers/api/users.php
```diff
@@ -1816,8 +1817,11 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
         $countryName = $locale->getText('countries.' . strtolower($session->getAttribute('countryCode')), $locale->getText('locale.country.unknown'));
 
         $session = $dbForProject->createDocument('sessions', $session);
+
+        $dbForProject->purgeCachedDocument('users', $user->getId());
+
         $session
-            ->setAttribute('secret', $secret)
```
loks0n, line 1820: We have to document this right, it's sort of breaking? Some people could be relying on it

stnguyen90: I don't think anyone is using it because it doesn't work.

PR #9032, opened 2024-11-27
Remove failed attribute

fogelito:
To delete failed attributes in _metadata and the adapter.

## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/databases.php
```diff
@@ -290,15 +290,9 @@ function updateAttribute(
         $attribute->setAttribute('size', $size);
     }
 
-    $formatOptions = $attribute->getAttribute('formatOptions');
-
     switch ($attribute->getAttribute('format')) {
         case APP_DATABASE_ATTRIBUTE_INT_RANGE:
         case APP_DATABASE_ATTRIBUTE_FLOAT_RANGE:
-            if ($min === $formatOptions['min'] && $max === $formatOptions['max']) {
```
fogelito, line 298: - I removed this for 2 reasons.
1. $options is never null , so will always update in the adapter updateAttribute method
2. In case no change in values, it will fix the value in _metadata, which currently is overwritten to empty array; having inconsistency between metadata and Appwrite;

abnegate (approved):

PR #9039, opened 2024-11-28
feat: use environment variable to check rules format

christyjacob4:
## What does this PR do?
Use an environment variable instead of the version to decide the mode for fetching and creating domains ( rules ) 

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/general.php
```diff
@@ -528,10 +528,10 @@ function router(App $utopia, Database $dbForConsole, callable $getProjectDB, Swo
                     $mainDomain = $envDomain;
                 } else {
                     // TODO: @christyjacob remove once we migrate the rules in 1.7.x
-                    if (version_compare(APP_VERSION_STABLE, '1.7.0', '<')) {
-                        $domainDocument = $dbForConsole->findOne('rules', [Query::orderAsc('$id')]);
-                    } else {
+                    if (System::getEnv('_APP_RULES_FORMAT', null) === 'md5') {
```
lohanidamodar, line 531: do we need `null` as default value here? it's already `null` by default right?

lohanidamodar (commented):
Rest looks good. Let's make sure we search everywhere for `version_compare` to not miss anything

lohanidamodar (commented):
Rest looks good. Let's make sure we search everywhere for `version_compare` to not miss anything

app/controllers/general.php
christyjacob4: Got you. Yeap, will revert it

christyjacob4:
@lohanidamodar checked, looks good 👍

lohanidamodar (approved):

PR #9108, opened 2024-12-16
Feat new sdk label

PineappleIOnic:
## What does this PR do?

This PR Implements an entirely new way of us defining SDK Metadata meant to be more type safe and less prone to errors.

## Test Plan

Regenerated Specs should be very close to their previous versions apart from the few times where bugs were actually found and corrected within the spec from this refactor work.

## Checklist


lohanidamodar (commented):
Haven't gone line by line, but it's a great improvement. With this we should also think about support for multiple responses in endpoint (open API spec already does) so that later we can improve the spec with all the possible responses success and error as well. Mostly I see that should be possible by changing few attributes in Method definition.

PineappleIOnic:
Took me a while but I've implemented the SDKResponse class which should allow us to achieve what you've asked for @lohanidamodar

app/controllers/api/avatars.php
```diff
@@ -164,13 +169,20 @@
...
+        description: '/docs/references/avatars/get-credit-card.md',
+        auth: [AuthType::SESSION, AuthType::KEY, AuthType::JWT],
+        methodType: MethodType::LOCATION,
+        responses: [
+            new SDKResponse(
+                code: Response::STATUS_CODE_OK,
+                model: Response::MODEL_NONE,
+            )
+        ],
+        responseType: ResponseType::IMAGE_PNG
```
abnegate, line 184: If responses can be an array, the response type should probably be a property of `SDKResponse` so that each can define it's own type if needed

src/Appwrite/SDK/Method.php
```diff
@@ -0,0 +1,256 @@
...
+        foreach ($responses as $response) {
+            /** @var \Appwrite\SDK\Response $response */
+            $this->validateResponseModel($response->getModel());
+
+            // No content check
+            if ($response->getCode() === 204) {
+                if ($response->getModel() !== Response::MODEL_NONE) {
+                    self::$exceptions[] = "Error with {$this->getDebugName()} method: Response code 204 must have response model 'none'";
+                }
+            }
```
abnegate, line 104: Let's move this logic to it's own validate function

src/Appwrite/SDK/Method.php
```diff
@@ -0,0 +1,256 @@
...
+}
+
+class Method
+{
+    public static array $knownMethods = [];
+
+    //Tmp
+    public static array $knownMissingDescriptions = [];
+
+    public static array $exceptions = [];
```
abnegate, line 43: Let's call this errors since we're storing strings not exceptions

src/Appwrite/SDK/Method.php
```diff
@@ -0,0 +1,256 @@
...
+{
+    case NONE = '';
+    case JSON = 'application/json';
+    case IMAGE = 'image/*';
+    case IMAGE_PNG = 'image/png';
+    case MULTIPART = 'multipart/form-data';
+    case HTML = 'text/html';
+    case TEXT = 'text/plain';
+    case ANY = '*/*';
+}
```
abnegate, line 34: These should be in their own files

src/Appwrite/SDK/Method.php
```diff
@@ -0,0 +1,256 @@
...
+
+class Method
+{
+    public static array $knownMethods = [];
+
+    //Tmp
+    public static array $knownMissingDescriptions = [];
```
abnegate, line 41: Let's remove this if it's temporary

app/controllers/api/databases.php
```diff
@@ -2852,15 +3094,38 @@ function updateAttribute(
...
+                    responseModel: Response::MODEL_DOCUMENT,
+                ),
+                new Multiplex(
+                    name: 'createDocuments',
+                    parameters: ['documents'],
+                    required: ['documents'],
+                    responseModel: Response::MODEL_DOCUMENT_LIST,
+                ),
+            ]
+        )
```
abnegate, line 3127: It would make more sense here if `sdk` has an array with two instances of `Method`

src/Appwrite/GraphQL/Schema.php
```diff
@@ -98,14 +98,17 @@ protected static function api(App $utopia, callable $complexity): array
...
+                $sdk = $route->getLabel('sdk', false);
 
-                if (empty($name)) {
+                if (empty($sdk)) {
                     continue;
                 }
 
+                $namespace = $sdk->getNamespace();
+                $method = $sdk->getMethodName();
+                $name = $namespace . \ucfirst($method);
```
abnegate, line 110: We'll need to handle multiple methods here as well, if there a multiple SDK methods, there should be multiple GraphQL queries too

src/Appwrite/SDK/Method.php
```diff
@@ -0,0 +1,256 @@
...
+            // No content check
+            if ($response->getCode() === 204) {
+                if ($response->getModel() !== Response::MODEL_NONE) {
+                    self::$exceptions[] = "Error with {$this->getDebugName()} method: Response code 204 must have response model 'none'";
+                }
+            }
+        }
+    }
+
+    private function getDebugName(): string
```
abnegate, line 108: Let's use a better name if it's not only used for debug, otherwise remove

src/Appwrite/SDK/Method.php
```diff
@@ -0,0 +1,256 @@
...
+        return $this->multiplex;
+    }
+
+    // Throw any errors that were found during initialization
+    static function finaliseInitialization(): void
+    {
+        if (!empty(self::$exceptions)) {
+            throw new \Exception('Errors found during SDK initialization:' . PHP_EOL . implode(PHP_EOL, self::$exceptions));
+        }
+    }
```
abnegate, line 255: This is hacky, let's find a better way to do it. Should be an init check after all controllers are loaded

src/Appwrite/SDK/Multiplex.php
```diff
@@ -0,0 +1,40 @@
...
+     * @param string $name
+     * @param array<string> $parameters
+     * @param array<string> $required
+     * @param string $responseModel
+     */
+    public function __construct(
+        private string $name,
+        private array $parameters,
+        private array $required,
+        private string $responseModel
```
abnegate, line 17: We can merge with `Method` and just use one model

app/config/specs/swagger2-latest-client.json
```diff
@@ -367,7 +367,9 @@
                 "consumes": [
                     "application\/json"
                 ],
-                "produces": [],
+                "produces": [
+                    "application\/json"
+                ],
```
abnegate, line 372: I think this should still be empty array

app/controllers/api/databases.php
```diff
@@ -2852,15 +3093,49 @@ function updateAttribute(
...
+                responses: [
+                    new SDKResponse(
+                        code: Response::STATUS_CODE_CREATED,
+                        model: Response::MODEL_DOCUMENT,
+                    )
+                ],
+                responseType: ResponseType::JSON,
+                offlineKey: '{documentId}',
+                offlineModel: '/databases/{databaseId}/collections/{collectionId}/documents',
+                parameters: [
```
abnegate, line 3113: Do we need default/validator/desc etc. here? Could we pull them from the actual parameters during spec generation instead?

app/config/specs/swagger2-latest-client.json
PineappleIOnic: Just set all the delete routes to use ResponseType::NONE

app/controllers/api/databases.php
PineappleIOnic: I just split parameters and additionalParameters

app/controllers/api/avatars.php
abnegate, line 184: Let's rename this `contentType`

app/controllers/api/databases.php
```diff
@@ -2852,15 +3093,49 @@ function updateAttribute(
...
+                    'documentId' => ['optional' => false],
+                    'data' => ['optional' => false],
```
abnegate, line 3115: Why do we need optional? We can get it from the actual param as well

app/controllers/api/databases.php
```diff
@@ -2852,15 +3093,49 @@ function updateAttribute(
...
+                auth: [AuthType::SESSION, AuthType::KEY, AuthType::JWT],
+                responses: [
+                    new SDKResponse(
+                        code: Response::STATUS_CODE_CREATED,
+                        model: Response::MODEL_DOCUMENT,
+                    )
+                ],
+                responseType: ResponseType::JSON,
+                offlineKey: '{documentId}',
+                offlineModel: '/databases/{databaseId}/collections/{collectionId}/documents',
```
abnegate, line 3132: Let's remove these keys from everywhere, they're not in use

app/controllers/api/databases.php
```diff
@@ -2852,15 +3093,49 @@ function updateAttribute(
...
+            ),
+            new Method(
+                namespace: 'databases',
+                name: 'createDocuments',
+                description: '/docs/references/databases/create-document.md',
+                auth: [AuthType::SESSION, AuthType::KEY, AuthType::JWT],
+                responses: [
+                    new SDKResponse(
+                        code: Response::STATUS_CODE_CREATED,
+                        model: Response::MODEL_DOCUMENT,
```
abnegate, line 3127: This should be document list

app/controllers/api/databases.php
PineappleIOnic: We can't get it from the API because it's different, in the API we need to have `documentId, data and documents` all optional so we can allow either bulk or single creation. As a result we have to define optional elsewhere

src/Appwrite/SDK/Method.php
```diff
@@ -0,0 +1,284 @@
...
+    ) {
+        $this->validateMethod($name, $namespace);
+        $this->validateAuthTypes($auth);
+        //$this->validateDesc($description);
+
+        foreach ($responses as $response) {
+            /** @var SDKResponse $response */
+            $this->validateResponseModel($response->getModel());
+
+            // No content check
```
abnegate, line 57: This code is clear enough without the comment, let's remove it

src/Appwrite/SDK/Method.php
```diff
@@ -0,0 +1,284 @@
...
+
+namespace Appwrite\SDK;
+
+use Appwrite\SDK\Response as SDKResponse;
+use Appwrite\Utopia\Response;
+use Swoole\Http\Response as HttpResponse;
+
+class Method
+{
+    public static array $knownMethods = [];
```
abnegate, line 11: Let's name this `processed`

src/Appwrite/SDK/Method.php
```diff
@@ -0,0 +1,284 @@
...
+        return $this->description;
+    }
+
+    public function getAuth(): array
+    {
+        return $this->auth;
+    }
+
+    /**
+     * @return Array<SDKResponse>
```
abnegate, line 153: Lowercase `array`

src/Appwrite/SDK/Method.php
```diff
@@ -0,0 +1,284 @@
...
+            }
+
+            return;
+        }
+
+        try {
+            $response->getModel($responseModel);
+        } catch (\Exception $e) {
+            self::$errors[] = "Error with {$this->getRouteName()} method: Invalid response model, make sure the model has been defined in Response.php";
+        }
```
abnegate, line 120: Let's put a single item in an array here so we can remove the duplicate logic

src/Appwrite/Utopia/Request.php
```diff
@@ -28,8 +28,14 @@ public function getParams(): array
...
-            $method = self::getRoute()->getLabel('sdk.method', 'unknown');
-            $endpointIdentifier = self::getRoute()->getLabel('sdk.namespace', 'unknown') . '.' . $method;
+            $method = self::getRoute()->getLabel('sdk', null);
+            /** @var \Appwrite\SDK\Method $method */
+
+            if (empty($method)) {
+                $endpointIdentifier = 'unknown.unknown';
+            } else {
+                $endpointIdentifier = $method->getNamespace() . '.' . $method->getMethodName();
+            }
```
abnegate, line 39: This needs to handle when method is an array

app/config/specs/open-api3-latest-console.json
```diff
@@ -23272,11 +22469,11 @@
                 "description": "",
                 "responses": {
                     "200": {
-                        "description": "Project",
+                        "description": "EmailTemplate",
                         "content": {
                             "application\/json": {
                                 "schema": {
-                                    "$ref": "#\/components\/schemas\/project"
+                                    "$ref": "#\/components\/schemas\/emailTemplate"
```
abnegate, line 22476: This should still be project

src/Appwrite/GraphQL/Schema.php
```diff
@@ -98,14 +98,25 @@ protected static function api(App $utopia, callable $complexity): array
...
+                    $mainSdk = array_shift($sdk);
+                    $additionalMethods = $sdk;
+
+                    $sdk = $mainSdk;
+                }
+
+                $namespace = $sdk->getNamespace();
+                $method = $sdk->getMethodName();
+                $name = $namespace . \ucfirst($method);
+
```
abnegate, line 119: This will hide any additional methods from graphql, we'll need to make sure both methods are available there too

PineappleIOnic: Should be fixed now: 👍

src/Appwrite/GraphQL/Schema.php
```diff
@@ -98,27 +98,36 @@ protected static function api(App $utopia, callable $complexity): array
...
+                if (!\is_array($sdk)) {
+                    $sdk = [$sdk];
+                }
+
+                foreach ($sdk as $method) {
+                    $namespace = $method->getNamespace();
+                    $methodName = $method->getMethodName();
+                    $name = $namespace . \ucfirst($methodName);
+
+                    foreach (Mapper::route($utopia, $route, $complexity) as $field) {
```
abnegate, line 117: We should pass $method into the `route` call, and use $methods->getParameters() to match against the route parameters (instead of iterating route parameters only on line 128). Otherwise both createDocument and createDocuments will have **all** route params in graphql

abnegate (approved):

PR #9138, opened 2024-12-27
Fix: VCS identity collision

Meldiron:
## What does this PR do?

Fixes a problem when personal GitHub installation could only be created on Appwrite account with same email as github email.

VCS no longer uses identities, preventing this problem. Nessessary tokens are now stored as part of installation document. These tokens are only needed when creating new repository.

## Test Plan

- [x] Manual QA

1. Create account by "Sign in with GitHub"

CleanShot 2024-12-27 at 13 22 40@2x

2. Sign out, and create new account `matej@appwrite.io` (different email than GitHub)

CleanShot 2024-12-27 at 13 24 01@2x

3. Create organization, and a project. Add installation successfully

CleanShot 2024-12-27 at 13 30 20@2x

- [x] Backwards-compatibility check:

1. Registered, made project, and linked GitHub with `1.6.x` branch
2. Switched to this branch, and created repository:

CleanShot 2024-12-30 at 14 57 06@2x

3. Ensured installations don't have tokens columns, and identities do ✅

## Related PRs and Issues

x

## Checklist


app/config/collections.php
```diff
@@ -5417,6 +5417,39 @@
                 'default' => false,
                 'array' => false,
             ],
+            [
+                '$id' => ID::custom('personalAccessToken'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 16384,
```
loks0n, line 5424: Could this be smaller? `1024?`

app/config/collections.php
```diff
@@ -5417,6 +5417,39 @@
...
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => ['datetime'],
+            ],
+            [
+                '$id' => ID::custom('personalRefreshToken'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 16384,
```
loks0n, line 5446: Same here - could this be smaller? `1024?`

loks0n (approved):

app/config/collections.php
Meldiron: I copied it from identities.
Looking at GH docs, doesn't seem to be mentioned. From what I recall it was only like 16 or 32 chars - setting to 128 should be good enough.
cc @christyjacob4 do we want to try and lower this?

christyjacob4: Any value from 1 to 16381 will result [in VARCHAR](https://github.com/utopia-php/database/blob/main/src/Database/Adapter/MariaDB.php#L2236-L2240) And the current value `16384` will be treated as TEXT

So lets go with varchar and set it to 256 for some future proofing
Screenshot 2025-01-08 at 3 05 08 PM

loks0n (approved):

christyjacob4 (approved):

PR #9142, opened 2024-12-29
Usage databases api read writes addition

shimonewman:


app/init.php
```diff
@@ -232,6 +232,8 @@
 // Usage metrics
 const METRIC_TEAMS = 'teams';
 const METRIC_USERS = 'users';
+const METRIC_DATABASE_API_READ  = 'database.api.events.read';
```
eldadfux, line 235: 1. Lets also have per database metrics
2. We don't really need either `api` (all stats are technically from the API, not need to be explicit about it) or `events` (used at Appwrite all for triggering functions and webhooks)
3. In other places we use `databases` and not `database`

Maybe we can instead do: `databases.operations.reads` and `databases.operations.writes`

eldadfux: We can also move this lines closer to other database related stuff we have on this list.

app/controllers/api/databases.php
```diff
@@ -2885,7 +2885,11 @@ function updateAttribute(
             throw new Exception(Exception::DOCUMENT_INVALID_STRUCTURE, '$id is not allowed for creating new documents, try update instead');
         }
 
-        $database = Authorization::skip(fn () => $dbForProject->getDocument('databases', $databaseId));
+        $database = Authorization::skip(function () use ($queueForUsage, $dbForProject, $databaseId) {
+            $queueForUsage->addMetric(METRIC_DATABASE_API_READ, 1);
+            return $dbForProject->getDocument('databases', $databaseId);
```
eldadfux, line 2890: We need to have 1 count per endpoint, we will reflect all costs inside, we don't need to actually check for every under the hood operation, this will make usage stats extremely unpredictable. 

For example:

createDocument = 1 write, if it has relationships, maybe multiple writes in the metric value
getDocument = 1 read, , if it has relationships, maybe multiple reads in the metric value
listDocuments = 1 read , if it has relationships, maybe multiple reads in the metric value
etc

eldadfux (changes requested):

eldadfux (approved):

PR #9144, opened 2024-12-30
Remove firebase OAuth API endpoints

PineappleIOnic:
## What does this PR do?

The firebase migration code had the ability to one click migrate using Sign in with Google and OAuth2 but Google blocked our application from accessing the required scopes to achieve this so this functionality was never used. This PR removes those OAuth API endpoints.

## Test Plan

All functionality should work as previously expected due to these API's never being used.

## Checklist


eldadfux (approved):
Thank you Google, but not thank you. PR approved.

PR #9146, opened 2024-12-30
Set base specification CPUs to 0.5 again

PineappleIOnic:
## What does this PR do?

This PR rolls back a temporary change that set the base specification's CPU to 1 instead of 0.5

## Checklist

### DIFF
--- a/app/config/runtimes/specifications.php
+++ b/app/config/runtimes/specifications.php
@@ -6,7 +6,7 @@ return [
     Specification::S_05VCPU_512MB => [
         'slug' => Specification::S_05VCPU_512MB,
         'memory' => 512,
-        'cpus' => 1 // TODO: revert this, it's a temporary change to test function performance.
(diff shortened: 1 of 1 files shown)


app/config/runtimes/specifications.php
```diff
@@ -6,7 +6,7 @@
     Specification::S_05VCPU_512MB => [
         'slug' => Specification::S_05VCPU_512MB,
         'memory' => 512,
-        'cpus' => 1 // TODO: revert this, it's a temporary change to test function performance.
+        'cpus' => 0.5
```
Meldiron, line 9: I think it was increased due to waterfall effect on Cloud.
Let's make sure with Christy that everyone was migrated to `S_1VCPU_512MB` before releasing this.

Meldiron (approved):

abnegate (approved):

PR #9147, opened 2024-12-30
Feat: Add executionId and client IP to function headers

JoshiJoshiJoshi:
## What does this PR do?

This PR adds the executionId and the client IP to the function headers so that they are accessible within the function execution.

Fixes https://github.com/appwrite/appwrite/issues/9316

## Test Plan
Manual testing (see screenshots)
CleanShot 2024-12-30 at 18 49 22@2x
CleanShot 2024-12-30 at 18 53 02@2x
CleanShot 2024-12-30 at 18 57 49@2x
image
CleanShot 2024-12-30 at 19 11 40@2x

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/9316

## Checklist

* [x]  Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

### DIFF
--- a/src/Appwrite/Platform/Modules/Functions/Http/Executions/Create.php
+++ b/src/Appwrite/Platform/Modules/Functions/Http/Executions/Create.php
@@ -80,7 +80,7 @@ class Create extends Base
             ->param('body', '', new Text(10485760, 0), 'HTTP body of execution. Default value is empty string.', true)
             ->param('async', false, new Boolean(true), 'Execute code in the background. Default value is false.', true)
             ->param('path', '/', new Text(2048), 'HTTP path of execution. Path can include query params. Default value is /', true)
-            ->param('method', 'POST', new Whitelist(['GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'OPTIONS'], true), 'HTTP method of execution. Default value is GET.', true)
+            ->param('method', 'POST', new Whitelist(['GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'OPTIONS'], true), 'HTTP method of execution. Default value is POST.', true)
             ->param('headers', [], new AnyOf([new Assoc(), new Text(65535)], AnyOf::TYPE_MIXED), 'HTTP headers of execution. Defaults to empty.', true)
             ->param('scheduledAt', null, new Text(100), 'Scheduled execution time in [ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html) format. DateTime value must be in future with precision in minutes.', true)
             ->inject('response')


Meldiron:
> Unsure how to proceed with the following lines:
> Should they be added to v2 runtimes as well?

V2 is backwards compatibility for older functions syntax. It's for when functions used to get `req, res` instead of `context`. V2 functions don't allow new build, and support only exists to keep supporting existing functions.

**No need to implement any of the new features into `v2` statements**

tests/resources/functions/swift/index.swift
```diff
@@ -12,6 +12,8 @@ func main(req: RequestValue, res: RequestResponse) throws -> RequestResponse {
         "APPWRITE_FUNCTION_USER_ID": req.variables["APPWRITE_FUNCTION_USER_ID"],
         "APPWRITE_FUNCTION_JWT": req.variables["APPWRITE_FUNCTION_JWT"],
         "APPWRITE_FUNCTION_PROJECT_ID": req.variables["APPWRITE_FUNCTION_PROJECT_ID"],
+        "APPWRITE_FUNCTION_EXECUTION_ID": req.variables["APPWRITE_FUNCTION_EXECUTION_ID"],
+        "APPWRITE_FUNCTION_CLIENT_IP": req.variables["APPWRITE_FUNCTION_CLIENT_IP"],
```
Meldiron, line 16: Amazing job updating tests too 🔥
Let's also update test themselves in `FunctionsCustomClientTest.php` and `FunctionsCustomServerTest.php`. You can look for tests for `APPWRITE_FUNCTION_USER_ID`, and in those places, add test for `APPWRITE_FUNCTION_EXECUTION_ID` and `APPWRITE_FUNCTION_CLIENT_IP`.

src/Appwrite/Platform/Workers/Functions.php
```diff
@@ -380,6 +381,7 @@ private function execute(
             'scopes' => $function->getAttribute('scopes', [])
         ]);
 
+        $headers['x-appwrite-execution-id'] = $executionId ?? '';
```
Meldiron, line 384: @christyjacob4 @eldadfux To confirm header naming is okay

app/controllers/general.php
```diff
@@ -202,13 +202,16 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
...
+        $executionId = ID::unique();
+        $headers['x-appwrite-execution-id'] = $executionId ?? '';
         $headers['x-appwrite-key'] = API_KEY_DYNAMIC . '_' . $apiKey;
         $headers['x-appwrite-trigger'] = 'http';
         $headers['x-appwrite-user-id'] = '';
         $headers['x-appwrite-user-jwt'] = '';
         $headers['x-appwrite-country-code'] = '';
         $headers['x-appwrite-continent-code'] = '';
         $headers['x-appwrite-continent-eu'] = 'false';
+        $headers['x-appwrite-client-ip'] = $request->getIP();
```
Meldiron, line 214: @christyjacob4 @eldadfux To confirm header naming is okay

src/Appwrite/Platform/Workers/Functions.php
Meldiron, line 1: In worker, when we dont have the IP, let's add header anyway as empty string. It improves developer experience in strictly typed languages when they can safely cast it into string type

app/controllers/general.php
eldadfux: We use user almost everywhere, if we start using client, won't we need to change a few more things?

Meldiron: We dont have Appwrite User, it could be a guest execution.
I think we need new namespace anyway, and `client` meaning "internet client" makes sense to me.

Hanibachi:
Cannot wait for this to be implemented

watzon:
Any movement on this? I need a way to get the execution id from within my function call.

app/controllers/general.php
```diff
@@ -356,10 +356,13 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
         }
 
         $headers = \array_merge([], $requestHeaders);
+        $executionId = ID::unique();
+        $headers['x-appwrite-execution-id'] = $executionId ?? '';
         $headers['x-appwrite-user-id'] = '';
         $headers['x-appwrite-country-code'] = '';
         $headers['x-appwrite-continent-code'] = '';
         $headers['x-appwrite-continent-eu'] = 'false';
+        $headers['x-appwrite-client-ip'] = $request->getIP();
```
Meldiron: Good point by CodeRabbit here, let's add client IP header to whitelist of request headers, so it can be seen in Console by developer.

It's not privacy issue as tracking IP is allowed for security reasons, and we anyway already store IP of sessions, so our compliance already cover it.

src/Appwrite/Platform/Modules/Functions/Http/Executions/Create.php
```diff
@@ -221,13 +221,16 @@ public function action(
...
         $headers['x-appwrite-key'] = API_KEY_DYNAMIC . '_' . $apiKey;
         $headers['x-appwrite-trigger'] = 'http';
         $headers['x-appwrite-user-id'] = $user->getId() ?? '';
         $headers['x-appwrite-user-jwt'] = $jwt ?? '';
         $headers['x-appwrite-country-code'] = '';
         $headers['x-appwrite-continent-code'] = '';
         $headers['x-appwrite-continent-eu'] = 'false';
+        $headers['x-appwrite-client-ip'] = $request->getIP();
 
         $ip = $headers['x-real-ip'] ?? '';
```
Meldiron: Hmm this seems wrong now, I think we should use $request->getIP() here too.

Let's check where else we use this x-real-ip directly, and consider replacing it there as well

src/Appwrite/Platform/Workers/Functions.php
```diff
@@ -264,10 +264,13 @@ private function fail(
         string $jwt = null,
         string $event = null,
     ): void {
+        $executionId = ID::unique();
+        $headers['x-appwrite-execution-id'] = $executionId ?? '';
         $headers['x-appwrite-trigger'] = $trigger;
         $headers['x-appwrite-event'] = $event ?? '';
         $headers['x-appwrite-user-id'] = $user->getId() ?? '';
         $headers['x-appwrite-user-jwt'] = $jwt ?? '';
+        $headers['x-appwrite-client-ip'] = '';
```
Meldiron: Agree with CodeRabbit here - if headers already have client IP, we shouldn't replace it. Only set it to empty string if we dont have it already

src/Appwrite/Platform/Workers/Functions.php
```diff
@@ -405,18 +408,20 @@ private function execute(
         $headers['x-appwrite-country-code'] = '';
         $headers['x-appwrite-continent-code'] = '';
         $headers['x-appwrite-continent-eu'] = 'false';
+        $headers['x-appwrite-client-ip'] = '';
```
Meldiron: Same here, let's only set to empty if its not already there

tests/e2e/Services/Functions/FunctionsCustomClientTest.php
```diff
@@ -159,6 +159,11 @@ public function testCreateCustomExecution(): array
         $this->assertNotEmpty($output['APPWRITE_FUNCTION_JWT']);
         $this->assertEquals($this->getProject()['$id'], $output['APPWRITE_FUNCTION_PROJECT_ID']);
 
+        $executionId = $execution['body']['$id'] ?? '';
+        $this->assertNotEmpty($output['APPWRITE_FUNCTION_EXECUTION_ID']);
+        $this->assertEquals($executionId, $output['APPWRITE_FUNCTION_EXECUTION_ID']);
```
Meldiron, line 177: I would do 1 more assertion, in async flow. async execution, or event trigger, or schedule, anywhere.

There we should assert execution ID exists, but client IP is empty

PR #9153, opened 2025-01-02
Split the collection definitions

ItzNotABug:
## What does this PR do?

This PR splits the extended collection additions from a single file to proper groups like `platform`/`console`, `project`, `databases`, buckets, etc.

## Test Plan

Manual.

## Related PRs and Issues

N/A.

## Checklist


app/config/collections.php
```diff
@@ -1,12 +1,15 @@
 <?php
 
-use Appwrite\Auth\Auth;
 use Utopia\Config\Config;
-use Utopia\Database\Database;
-use Utopia\Database\Helpers\ID;
 
-$providers = Config::getParam('oAuthProviders', []);
+$common = include __DIR__ . '/collections/common.php';
+$bucket = include __DIR__ . '/collections/bucket.php';
```
christyjacob4, line 6:
```suggestion
$buckets = include __DIR__ . '/collections/buckets.php';
```
Lets also update `databases` and `projects` - both the file name and the variable names

app/config/collections/buckets.php
Meldiron, line 1: Do we need to separate buckets? 🤔
IMO, it could be part of common. We shouldn't need dedicated custom buckets like we need for databases.

app/config/collections.php
```diff
@@ -1,12 +1,15 @@
...
 
-$providers = Config::getParam('oAuthProviders', []);
+$common = include __DIR__ . '/collections/common.php';
+$buckets = include __DIR__ . '/collections/buckets.php';
+$projects = include __DIR__ . '/collections/projects.php';
+$databases = include __DIR__ . '/collections/databases.php';
+$platform = include __DIR__ . '/collections/platform.php';
+
 $auth = Config::getParam('auth', []);
+$providers = Config::getParam('oAuthProviders', []);
```
Meldiron, line 12: I dont see those being used. Do we need them? Did we have them before - do we need to move to some of the new files?

app/config/collections.php
Meldiron, line 1: In this file, we still use `console` terminology. Lets mention this to team lead, so we have task to rename console -> platform everywhere

app/config/collections.php
ItzNotABug: We had it before, didn't find its usage but kept it as is, just in case...

app/config/collections.php
ItzNotABug: we could use `platform` for variables, but changing the collection names, I am not sure of that one.

app/config/collections/buckets.php
ItzNotABug: better to keep it separate instead of clogging the main collections imo.
also helps in separating logic.better to keep it separate instead of clogging the main collections imo. also helps in separating logic.

Meldiron: Why should we separate buckets, but not users? In theory we could have separate DB for each service, question is how we decide when to use common, and when to make separate one

app/config/collections.php
Meldiron: Let's try to remove and see if tests pass

app/config/collections.php
Meldiron: If we can switch to platform terminology yet everywhere, it might be better to stick to console for now, and address that in separate PR

app/config/collections/buckets.php
eldadfux: I think Matej has a good point

ItzNotABug: Will move buckets to `commons`.

ItzNotABug: Done.

app/config/collections/platform.php
```diff
@@ -0,0 +1,1534 @@
...
+        ],
+    ],
+
+    'vcsCommentLocks' => [
+        '$collection' => ID::custom(Database::METADATA),
+        '$id' => ID::custom('vcsCommentLocks'),
+        'name' => 'vcsCommentLocks',
+        'attributes' => [],
+        'indexes' => []
+    ]
```
christyjacob4, line 1533: are you sure this collection has no attributes?

ItzNotABug: Doesn't seem like - https://github.com/appwrite/appwrite/blob/7ecba8c0f282567911904de266ec1f0b4e0546a6/app/config/collections.php#L5701C5-L5707C7

Not sure if this is a leftover from previous implementations.

PR #9168, opened 2025-01-04
Optimise webhook events

christyjacob4:
## What does this PR do?
Perviously all endpoints with an event, triggered webhooks events which resulted in the a lot of spurious events in the webhooks queue and causing redis memory usage to increase. 

In this PR we check if the project has webhooks enabled and only then trigger the event.

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/shared/api.php
```diff
@@ -74,17 +74,20 @@
...
+    if ($project->getAttribute('webhooks', []) !== []) {
+        $queueForWebhooks
+            ->from($queueForEvents)
+            ->trigger();
     }
 
-    $queueForRealtime
-        ->from($queueForEvents)
-        ->trigger();
+    /** Trigger realtime events only for non console events */
```
eldadfux, line 85: but the console also uses realtime events, or those are all associated with a specific project?

app/controllers/shared/api.php
```diff
@@ -74,17 +74,20 @@
...
         ->trigger();
 
-    $queueForWebhooks
-        ->from($queueForEvents)
-        ->trigger();
 
-    if ($queueForEvents->getProject()->getId() === 'console') {
-        return;
+    /** Trigger webhooks events only if a project has them enabled */
+    if ($project->getAttribute('webhooks', []) !== []) {
```
eldadfux, line 79: Are you sure this condition works? wouldn't something like array !empty() would be safer here?

eldadfux, line 79: Do we want to go deeper and check if the specific even is listened by the webhooks? we should have the list of affected events in this context.

eldadfux (changes requested):

app/controllers/shared/api.php
christyjacob4: Screenshot 2025-01-07 at 10 27 53 AM

I have added a todo here, for additional optimisation. As it needs some benchmarks etc to understand the added latency in a request.

christyjacob4: Fixed 👍

app/controllers/shared/api.php
christyjacob4: We don't fire realtime events for DOCUMENT_CREATE events fired from the console. This is the existing behaviour. I have not changed it. 

https://github.com/appwrite/appwrite/blob/1.6.x/app/controllers/shared/api.php#L81-L88

https://github.com/appwrite/appwrite/blob/1.6.x/app/controllers/shared/api.php#L529-L536

loks0n (approved):

PR #9181, opened 2025-01-07
Make 0 equal unlimited for functions environment variables

PineappleIOnic:
## What does this PR do?

This PR makes it so if `_APP_FUNCTIONS_CPUS` and `_APP_FUNCTIONS_MEMORY` are set to 0 it will mean unlimited. This fits how these variables worked before variable runtime specifications.

## Test Plan

All existing tests should work as intended and I have manually tested the API's and Console to ensure everything is functional.
## Checklist

### DIFF
--- a/app/controllers/api/functions.php
+++ b/app/controllers/api/functions.php
@@ -550,7 +550,9 @@ App::get('/v1/functions/specifications')
             }
 
             // Only add specs that are within the limits set by environment variables
-            if ($spec['cpus'] <= System::getEnv('_APP_FUNCTIONS_CPUS', 1) && $spec['memory'] <= System::getEnv('_APP_FUNCTIONS_MEMORY', 512)) {
+            if (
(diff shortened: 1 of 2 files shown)


app/controllers/api/functions.php
```diff
@@ -550,7 +550,9 @@
             }
 
             // Only add specs that are within the limits set by environment variables
-            if ($spec['cpus'] <= System::getEnv('_APP_FUNCTIONS_CPUS', 1) && $spec['memory'] <= System::getEnv('_APP_FUNCTIONS_MEMORY', 512)) {
+            if (
+                ($spec['cpus'] <= System::getEnv('_APP_FUNCTIONS_CPUS', 1) || System::getEnv('_APP_FUNCTIONS_CPUS', 1) === 0) &&
+                ($spec['memory'] <= System::getEnv('_APP_FUNCTIONS_MEMORY', 512)) || System::getEnv('_APP_FUNCTIONS_MEMORY', 512) === 0) {
```
christyjacob4, line 555: Why don't we create a dummy specification with unlimited CPU and RAM and handle it like the other specs instead of these if conditions?

PR #9190, opened 2025-01-08
Fix dead connections

abnegate:
## What does this PR do?

When a PDO or database exception is thrown, and the exception was due to a lost connection, mark a pool connection as unhealthy, so that on next reclaim it will be re-created.

Introduces a `connectionForProject` resource that intercepts the project DB so that the connection can be used directly when needed for marking as unhealthy.

## Test Plan

Manual tests + library tests

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/controllers/general.php
+++ b/app/controllers/general.php
@@ -21,6 +21,7 @@ use Appwrite\Utopia\Response\Filters\V18 as ResponseV18;
 use Appwrite\Utopia\View;
 use Executor\Executor;
 use MaxMind\Db\Reader;
+use Swoole\Database\DetectsLostConnections;
 use Swoole\Http\Request as SwooleRequest;
 use Utopia\App;
 use Utopia\CLI\Console;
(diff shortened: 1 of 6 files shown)


app/controllers/general.php
```diff
@@ -746,7 +749,16 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
     ->inject('logger')
     ->inject('log')
     ->inject('queueForUsage')
-    ->action(function (Throwable $error, App $utopia, Request $request, Response $response, Document $project, ?Logger $logger, Log $log, Usage $queueForUsage) {
+    ->inject('connectionForProject')
+    ->action(function (Throwable $error, App $utopia, Request $request, Response $response, Document $project, ?Logger $logger, Log $log, Usage $queueForUsage, Connection $connectionForProject) {
+        if (
```
christyjacob4, line 754: for anything uncaught by this, we should also have a similar check in the error handler for the http server here https://github.com/appwrite/appwrite/blob/1.6.x/app/http.php#L348

abnegate: Added the check there too

christyjacob4 (approved):

PR #9191, opened 2025-01-08
chore: shifted authphone usage tracking to api calls

ChiragAgg5k:
## What does this PR do?

- Shifts usage tracking for MESSAGE_SEND_TYPE_INTERNAL from message worker to api calls: phone, mfa, team and verification.
- uses `authPhone` in plan object to set abuse checks if within free limit dont report usage for message
- a small comment fix

## Checklist

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -2473,11 +2473,11 @@ App::post('/v1/account/tokens/phone')
                         $queueForUsage
                             ->addMetric(str_replace('{countryCode}', $countryCode, METRIC_AUTH_METHOD_PHONE_COUNTRY_CODE), 1);
                     }
-                    $queueForUsage
-                        ->addMetric(METRIC_AUTH_METHOD_PHONE, 1)
-                        ->setProject($project)
-                        ->trigger();
                 }
+                $queueForUsage
+                    ->addMetric(METRIC_AUTH_METHOD_PHONE, 1)
+                    ->setProject($project)
+                    ->trigger();
             }
         }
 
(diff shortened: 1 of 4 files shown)


app/controllers/api/account.php
```diff
@@ -2452,6 +2458,27 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
+                    $countryCode = $helper->parse($phone)->getCountryCode();
+
+                    if (!empty($countryCode)) {
+                        $queueForUsage
+                            ->addMetric(str_replace('{countryCode}', $countryCode, METRIC_AUTH_METHOD_PHONE_COUNTRY_CODE), 1);
+                    }
+                    $queueForUsage
+                        ->addMetric(METRIC_AUTH_METHOD_PHONE, 1)
+                        ->setProject($project)
+                        ->trigger();
```
loks0n, line 2479: Can we move this outside the abuse check condition?
This should allow us to track the usage of the free SMS

src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -96,7 +96,7 @@ public function action(
                 $message = new Document($payload['message'] ?? []);
                 $recipients = $payload['recipients'] ?? [];
 
-                $this->sendInternalSMSMessage($message, $project, $recipients, $queueForUsage, $log);
```
christyjacob4, line 99: if we dont use this resource, can we remove the injection as well ?

ChiragAgg5k: https://github.com/ChiragAgg5k/appwrite/blob/16b32a1069c3cd2fb655058fa632bfd538c44fcf/src/Appwrite/Platform/Workers/Messaging.php#L104

loks0n (approved):

christyjacob4 (approved):

PR #9197, opened 2025-01-09
Feat: Deprecated runtims

Meldiron:
## What does this PR do?

Introduces deprecated rutimes that cannot be used to create a new function, or update existing function to deprecated runtime version

## Test Plan

- [x] new tests

## Related PRs and Issues

x

## Checklist


app/controllers/api/functions.php
```diff
@@ -151,7 +151,7 @@
     ->label('sdk.response.model', Response::MODEL_FUNCTION)
     ->param('functionId', '', new CustomId(), 'Function ID. Choose a custom ID or generate a random ID with `ID.unique()`. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.')
     ->param('name', '', new Text(128), 'Function name. Max length: 128 chars.')
-    ->param('runtime', '', new WhiteList(array_keys(Config::getParam('runtimes')), true), 'Execution runtime.')
+    ->param('runtime', '', new WhiteList(array_keys(Config::getParam('runtimes-active')), true), 'Execution runtime.')
```
christyjacob4, line 154: how do we ensure existing functions using the old runtime can continue to build?

app/config/runtimes-active.php
```diff
@@ -0,0 +1,9 @@
+<?php
```
christyjacob4, line 1: also why do we need a new config? can we just not use a single `runtime` config that has only the active ones?

HarshMN2345:
Closing as this targets a branch that is no longer maintained.

PR #9198, opened 2025-01-09
feat project sms usage

loks0n:


app/init.php
```diff
@@ -1844,6 +1844,10 @@ function getDevice(string $root, string $connection = ''): Device
     return [];
 });
 
+App::setResource('getSmsPrice', function () {
```
eldadfux, line 1847: I would usually vote against using resources this way as kind of a global memory placeholder. Can we avoid it? It can indicate a bad design pattern, the proper usage of resources is to hold and instance of an object.

src/Appwrite/Utopia/Response/Model/MetricBreakdown.php
```diff
@@ -27,6 +28,13 @@ public function __construct()
                 'description' => 'The value of this metric at the timestamp.',
                 'default' => 0,
                 'example' => 1,
+            ])
+            ->addRule('estimate', [
```
eldadfux, line 32: Will this be useful in other cases? Is this breakdown used only for SMS messages?

app/init.php
loks0n: Moved towards a static generated object instead: `smsRates`

eldadfux (approved):

christyjacob4 (approved):

PR #9204, opened 2025-01-10
chore: abstract token secret hiding to response model

ChiragAgg5k:
## What does this PR do?

This PR abstracts the secret hiding logic for the Token responses across various places by introduces a new attribute called `sensitive` to Response Model, which when set checks if user is appUser or priviledgedUser, and if not it sets the key as empty.

## Test Plan

I tested it for createEmailToken endpoint and it works like it did before.

test code:
```js
const client = new Client()
    .setEndpoint(endpoint)  
    .setProject(projectId)
    .setKey(apiKey);

const token = await account.createEmailToken(ID.unique(), 'test6@test.com', true);
console.log(token);
```

without apikey:
Screenshot 2025-01-10 at 7 00 37 PM

with apikey:
Screenshot 2025-01-10 at 7 01 02 PM

## Checklist

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -1979,16 +1979,10 @@ App::post('/v1/account/tokens/magic-url')
             ->setRecipient($email)
             ->trigger();
 
-        // Set to unhashed secret for events and server responses
         $token->setAttribute('secret', $tokenSecret);
 
         $queueForEvents
             ->setPayload($response->output($token, Response::MODEL_TOKEN), sensitive: ['secret']);
-
-        // Hide secret for clients
-        if (!$isPrivilegedUser && !$isAppUser) {
-            $token->setAttribute('secret', '');
-        }
 
         if (!empty($phrase)) {
             $token->setAttribute('phrase', $phrase);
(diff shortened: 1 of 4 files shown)


app/controllers/api/account.php
```diff
@@ -2481,13 +2469,12 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
 
-        // Set to unhashed secret for events and server responses
         $token->setAttribute('secret', $secret);
 
         $queueForEvents
             ->setPayload($response->output($token, Response::MODEL_TOKEN), sensitive: ['secret']);
 
-        // Hide secret for clients
+        // Encode secret for clients
         $token->setAttribute('secret', ($isPrivilegedUser || $isAppUser) ? Auth::encodeSession($user->getId(), $secret) : '');
```
loks0n, line 2478: Can't we remove the hide logic now?
```suggestion
        $token->setAttribute('secret', Auth::encodeSession($user->getId(), $secret));
```

src/Appwrite/Utopia/Response/Model/Token.php
```diff
@@ -33,6 +33,7 @@ public function __construct()
                 'description' => 'Token secret key. This will return an empty string unless the response is returned using an API key or as part of a webhook payload.',
                 'default' => '',
                 'example' => '',
+                'sensitive' => true,
```
loks0n, line 36: Are there some other other secrets we can use this on? 
What about the session secret for example?

loks0n (changes requested):

src/Appwrite/Utopia/Response/Model/Token.php
ChiragAgg5k: yess that abstracted a lot more things, thanks

ChiragAgg5k:
Screenshot 2025-01-11 at 6 44 34 PM

not sure why tests are failing in workflow... they are passing locally

PR #9211, opened 2025-01-12
Update audits to include user type

ItzNotABug:
## What does this PR do?

Adds the ability to log `user` types & `key` as name to Audits.

## Test Plan

Manual.

Console usage -
audit console

API usage -
audit api

Anonymous usage -
anonymous api

## Related PRs and Issues

N/A.

## Checklist


src/Appwrite/Auth/Auth.php
```diff
@@ -43,6 +43,12 @@ class Auth
     public const USER_ROLE_APPS = 'apps';
     public const USER_ROLE_SYSTEM = 'system';
 
+    /**
+     * Audit User Types.
+     */
+    public const AUDIT_TYPE_APP = 'app';
```
eldadfux, line 49: Not audit type, but user type, its always best to try and decouple libraries from knowing too much about each other and let the controller/action layer connect them. This help with better maintainability long terms and helps us keep better domain driven design.

eldadfux: Also, with new DB we should lean to start stop using the term audit, and use activity. Eventually, we will also deprecate the audit library.

eldadfux (changes requested):

src/Appwrite/Auth/Auth.php
ItzNotABug: Renaming to `ACTIVITY_TYPE_*` makes sense. Using `USER_TYPE_*` wouldn’t align with cases like `USER_TYPE_USER`.

app/controllers/shared/api.php
```diff
@@ -243,16 +244,20 @@
                     $user = new Document([
                         '$id' => '',
                         'status' => true,
+                        'type' => Auth::ACTIVITY_TYPE_APP,
                         'email' => 'app.' . $project->getId() . '@service.' . $request->getHostname(),
                         'password' => '',
-                        'name' => $project->getAttribute('name', 'Untitled'),
+                        'name' => 'Dynamic Key',
```
stnguyen90, line 250: It'll be a bit of a change but I wonder if it'll be useful to include the function name here. I'm thinking we'd have to include it in the encoded JWT.

app/controllers/shared/api.php
```diff
@@ -709,10 +723,11 @@
         }
 
         if (!$user->isEmpty()) {
+            $user->setAttribute('type', Auth::ACTIVITY_TYPE_USER);
             $queueForAudits->setUser($user);
         }
 
-        if (!empty($queueForAudits->getResource()) && !empty($queueForAudits->getUser()->getId())) {
+        if (!empty($queueForAudits->getResource()) && !$queueForAudits->getUser()->isEmpty()) {
```
stnguyen90, line 730: For cases where the request is not made by API key or user, the document from `getUser()` will be empty. That means we won't track it, right? Should we?

ItzNotABug: Yes. Added an `Anonymous` type to audits.

app/controllers/shared/api.php
ItzNotABug: what type of function name? call site name?
we could add maybe, that goes out of scope for this one.

eldadfux (approved):

app/controllers/shared/api.php
```diff
@@ -304,6 +310,9 @@
                             $dbForPlatform->purgeCachedDocument('projects', $project->getId());
                         }
                     }
+
+                    // standard api key user
```
christyjacob4, line 314:
```suggestion
```

app/controllers/shared/api.php
```diff
@@ -246,16 +247,20 @@
...
+                        'name' => 'Dynamic Key',
                     ]);
 
                     $role = Auth::USER_ROLE_APPS;
                     $scopes = \array_merge($roles[$role]['scopes'], $tokenScopes);
 
                     Authorization::setRole(Auth::USER_ROLE_APPS);
                     Authorization::setDefaultStatus(false);  // Cancel security segmentation for API keys.
+
+                    // dynamic api key user
```
christyjacob4, line 262:
```suggestion
```

app/controllers/shared/api.php
```diff
@@ -511,8 +518,16 @@
             ->setIP($request->getIP())
             ->setHostname($request->getHostname())
             ->setEvent($route->getLabel('audits.event', ''))
-            ->setProject($project)
-            ->setUser($user);
+            ->setProject($project);
+
+        // check first,
+        // as api key user might already exists
```
christyjacob4, line 524:
```suggestion
        /* If a session exists, use the user associated with the session */
```

app/controllers/shared/api.php
```diff
@@ -511,8 +518,15 @@
             ->setIP($request->getIP())
             ->setHostname($request->getHostname())
             ->setEvent($route->getLabel('audits.event', ''))
-            ->setProject($project)
-            ->setUser($user);
+            ->setProject($project);
+
+        /* If a session exists, use the user associated with the session */
+        if (!$user->isEmpty()) {
+            $typedUser = clone $user;
```
christyjacob4, line 525:
```suggestion
            $userClone = clone $user;
```
please update the var name everywhere

christyjacob4 (approved):

Issue #8825, opened 2024-10-19
🐛 Bug Report: CreateTeamMembership and providing an email to an anonymous user, calls them by the incorrect name


pingu2k4:
### 👟 Reproduction steps

- Create a user on your Appwrite App
- With said user, create a team
- With said user, call the [CreateTeamMembership endpoint](https://appwrite.io/docs/references/1.6.x/client-rest/teams#createMembership), passing an email address and a name for the invited user
- Inspect said email
- Note that the email begins with "Hello Inviter," (where Inviter is the person who invited, not the one receiving the email)

### 👍 Expected behavior

It should greet the person the email has been sent to, not the person who invited them.

### 👎 Actual Behavior

![Image](https://github.com/user-attachments/assets/8d82b995-6540-4f0c-82e9-c2d79748b53a)
In the image here, you can see that this was sent to my email address, and my name there was set as "Your Name" which is what I provided. Inviter was the name of the user that did the inviting, yet is greeted in the email. It then also states that Inviter did the inviting.

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

Windows

### 👀 Have you spent some time to check if this issue has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

Kraftsman1:
Hello @stnguyen90 what's the status of this issue?

I'd love to pick it up and work on.

Kraftsman1:
Thanks @Meldiron

asharan2buff:
Can I take this up if no one is working on it?

Kraftsman1:
@asharan2buff I'm working on it atm. I'll be a making a PR soon

ChiragAgg5k:
@Kraftsman1 are you still working on it?

stnguyen90:
Related community thread: https://discord.com/channels/564160730845151244/1327754291468832880

callmetushar123:
Hi, can you please locate me the code where the changes are expected, I am new to this repo. Will help a ton

ChiragAgg5k:
Here is the code that should be changed:

https://github.com/appwrite/appwrite/blob/7ecba8c0f282567911904de266ec1f0b4e0546a6/app/controllers/api/teams.php#L635

It should instead be `$name`, which if empty is defaulted to email above.

```php
$emailVariables = [
[...]

PR #8531, opened 2024-08-13
Bug 5901 fix resend invitation from create invite

ZachHandley:
## What does this PR do?

Fixes #5901 

## Test Plan

(Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified your changes work. Screenshots may also be helpful.)

## Related PRs and Issues

- #5901 

## Checklist

- [x] Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [x] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

PR #9218, opened 2025-01-13
fix: resend invitation

### LINKED ISSUE #5901: 🐛 Bug Report: When trying to resend email in Organization tab it throws AppwriteException

ChiragAgg5k:
## What does this PR do?

Fixes #5901

## Test Plan

## Related PRs and Issues

- #8531

## Checklist


app/controllers/api/teams.php
```diff
@@ -546,45 +546,57 @@
...
+                'teamInternalId' => $team->getInternalId(),
+                'roles' => $roles,
+                'invited' => DateTime::now(),
+                'joined' => ($isPrivilegedUser || $isAppUser) ? DateTime::now() : null,
+                'confirm' => ($isPrivilegedUser || $isAppUser),
+                'secret' => Auth::hash($secret),
+                'search' => implode(' ', [$membershipId, $invitee->getId()])
+            ]);
+
+            $createdMembership = true;
```
loks0n, line 580: is it possible to avoid this `createdMembership` var? and just group the side effects together?

loks0n: something like:
```
$membership = findOne(...)
if ($membership->isEmpty()) {
    $membership = new Document (...)
    $dbForProject->createDocument($membership)
    Authorization::skip(fn () => $dbForProject->increaseDocumentAttribute('teams', $team->getId(), 'total', 1));
 else {
    $dbForProject->updateDocument('memberships', $membership->getId(), $membership);
 }
 ```

**Review by loks0n — APPROVED** [review]

File: app/controllers/api/teams.php (hunk 1 above)
Comment (ChiragAgg5k) [reply app/controllers/api/teams.php]: done!

File: tests/e2e/Services/Teams/TeamsBaseClient.php [hunk 2]
```diff
@@ -292,9 +292,9 @@ public function testCreateTeamMembership($data): array
         $this->assertEquals('Invitation to ' . $teamName . ' Team at ' . $this->getProject()['name'], $lastEmail['subject']);
 
         /**
-         * Test for FAILURE
+         * Test for resend
+         * SUCCESS
          */
-
```
stnguyen90, line 297: No need to make a test for resend success block here. you can just add a 1 line comment saying this tests resending the invite.

app/controllers/api/teams.php
```diff
@@ -546,47 +546,51 @@
...
-            'teamInternalId' => $team->getInternalId(),
-            'roles' => $roles,
-            'invited' => DateTime::now(),
-            'joined' => ($isPrivilegedUser || $isAppUser) ? DateTime::now() : null,
-            'confirm' => ($isPrivilegedUser || $isAppUser),
-            'secret' => Auth::hash($secret),
-            'search' => implode(' ', [$membershipId, $invitee->getId()])
+        $membership = $dbForProject->findOne('memberships', [
+            Query::equal('userId', [$invitee->getId()]),
+            Query::equal('teamId', [$team->getId()]),
```
stnguyen90, line 551: We should always be using the internal iDs instead of the public IDs.

app/controllers/api/teams.php
```diff
@@ -546,47 +546,51 @@
...
-            $dbForProject->purgeCachedDocument('users', $invitee->getId());
         } else {
-            try {
-                $membership = $dbForProject->createDocument('memberships', $membership);
-            } catch (Duplicate $th) {
-                throw new Exception(Exception::TEAM_INVITE_ALREADY_EXISTS);
-            }
+            $membership = ($isPrivilegedUser || $isAppUser) ?
+                Authorization::skip(fn () => $dbForProject->updateDocument('memberships', $membership->getId(), $membership)) :
+                $dbForProject->updateDocument('memberships', $membership->getId(), $membership);
```
stnguyen90, line 587: What would this be updating?

app/controllers/api/teams.php
```diff
@@ -546,47 +546,51 @@
...
+        }
+
 
+        if ($isPrivilegedUser || $isAppUser) {
+            $dbForProject->purgeCachedDocument('users', $invitee->getId());
```
stnguyen90, line 592: Why would this be needed?

stnguyen90 (commented):
Want to fix https://github.com/appwrite/appwrite/issues/8825 too since you're modifying the same area of code?

app/controllers/api/teams.php
ChiragAgg5k: I guess it updates the `invited` field with the current timestamp. maybe i can switch it to just update that field instead of the entire document

app/controllers/api/teams.php
ChiragAgg5k: got itt

app/controllers/api/teams.php
ChiragAgg5k: @stnguyen90  It was present in the previous code... and i guess its there because we don't send verification email if the user is privileged? not entirely sure as well

app/controllers/api/teams.php
stnguyen90: Where is the invited timestamp being updated?

Also, this should probably update joined and confirmed for privileged and app users.

ChiragAgg5k: @stnguyen90 it does so here -
https://github.com/ChiragAgg5k/appwrite/blob/8dc3b4ca6032c337e7f0e9a0f6b078d90a38e83f/app/controllers/api/teams.php#L573-L574

ChiragAgg5k: ohh right, ig i skipped it, fixing

app/controllers/api/teams.php
```diff
@@ -546,47 +546,57 @@
...
             Authorization::skip(fn () => $dbForProject->increaseDocumentAttribute('teams', $team->getId(), 'total', 1));
 
-            $dbForProject->purgeCachedDocument('users', $invitee->getId());
         } else {
-            try {
-                $membership = $dbForProject->createDocument('memberships', $membership);
-            } catch (Duplicate $th) {
-                throw new Exception(Exception::TEAM_INVITE_ALREADY_EXISTS);
-            }
+            $membership = new Document([
```
stnguyen90, line 585: What happens to the rest of the attributes if you initialize a new document like this?

stnguyen90 (changes requested):

app/controllers/api/teams.php
ChiragAgg5k: from my understanding `updateDocument` will overlap these attribute with already existing one in dbForProject, and hence `joined` and `invited` will be updated

And `id` is just there so it doesnt need to be stored in separate variable.

Screenshot 2025-01-17 at 10 36 54 PM

stnguyen90: For consistency, let's update the document that exists in memory instead. Like:

https://github.com/appwrite/appwrite/blob/7ecba8c0f282567911904de266ec1f0b4e0546a6/app/controllers/api/users.php#L1232

Remember: 

1. $membership is already the existing document because it was fetched some lines above
2. Joined and confirm should only be updated if the request is from a privileged use or API key
3. What about invited timestamp?

stnguyen90 (changes requested):

app/controllers/api/teams.php
ChiragAgg5k: Ohhh okay, got it 👍

stnguyen90 (approved):

PR #9228, opened 2025-01-15
fix: slack oauth scopes, and updated to v2

ChiragAgg5k:
## What does this PR do?

- Fixes adding of scopes to the API call.
- Updates api to use the v2 version by slack

slack docs: https://api.slack.com/authentication/oauth-v2

## Checklist


loks0n (approved):

ChiragAgg5k:
Screenshot 2025-01-17 at 12 36 10 PM

working ^

src/Appwrite/Auth/OAuth2/Slack.php
```diff
@@ -20,10 +20,7 @@ class Slack extends OAuth2
      * @var array
      */
     protected array $scopes = [
-        'identity.avatar',
-        'identity.basic',
-        'identity.email',
-        'identity.team'
+        'users:read',
```
stnguyen90, line 23: hmm im worried this is more than what we actually need and it grants access to read all users in a workspace.

maybe all we need is `email`?

ChiragAgg5k: oops yeah, updated it to:
- openid for userId
- email 
- profile for user name

christyjacob4 (approved):

src/Appwrite/Auth/OAuth2/Slack.php
```diff
@@ -56,16 +56,15 @@ public function getLoginURL(): string
...
                 'GET',
-                'https://slack.com/api/oauth.access?' . \http_build_query([
+                'https://slack.com/api/oauth.v2.access?' . \http_build_query([
                     'client_id' => $this->appID,
                     'client_secret' => $this->appSecret,
                     'code' => $code,
                     'redirect_uri' => $this->callback
                 ])
-            ), true);
+            ), true)['authed_user'];
```
stnguyen90, line 67: Accessing like this is dangerous and can throw an exception leading to a 500 error. Let's be a bit more defensive and use `??`

stnguyen90 (changes requested):

stnguyen90 (approved):

PR #9237, opened 2025-01-17
Implement new SDK Class on 1.6.x

PineappleIOnic:
## What does this PR do?

This PR implements the new SDK Class for creating SDK Methods. The generated specifications should look the same apart from the removal of offline labels and me fixing some invalid SDK labels here and there.

### SDK Label Fixes:
- Create MFA challenge (confirmation) in the SDK Now correctly returns session instead of no content
- Update custom email templates now correctly returns Email Template instead of Project
- Delete authenticator now returns No Content instead of User

## Test Plan

All existing tests should function as normal.

## Checklist


src/Appwrite/SDK/Method.php
```diff
@@ -0,0 +1,276 @@
...
+        protected bool $deprecated = false,
+        protected array|bool $hide = false,
+        protected bool $packaging = false,
+        protected string $requestType = 'application/json',
+        protected array $parameters = [],
+        protected array $additionalParameters = []
+    ) {
+        $this->validateMethod($name, $namespace);
+        $this->validateAuthTypes($auth);
+        //$this->validateDesc($description);
```
abnegate, line 51: Let's enabled this now and add any missing descriptions

abnegate (approved):

Issue #8261, opened 2024-06-12
🐛 Bug Report: Email from other targets/users are sent to all targets by default


JoshiJoshiJoshi:
### 👟 Reproduction steps

1. Add two or more users to a messaging topic.
2. Send an email either via console or api and select that topic.
OR
1. Send an email either via console or api and select two or more different users as target.

### 👍 Expected behavior

The users receives the email AND the email DOES NOT show the email from the other targets or users in the same topic.
Other users should not know the email from the other targets. This can be considered a data breach and maybe not GDPR compliant.

### 👎 Actual Behavior

By default Appwrite shows the user/target to all other targets as well.

### 🎲 Appwrite version

Version 1.5.x

### 💻 Operating system

Linux

### 🧱 Your Environment

Appwrite v1.5.7

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

stnguyen90:
Related discord discussion: https://discord.com/channels/564160730845151244/564160731327758347/1250438121733095444

stnguyen90:
Another thread: https://discord.com/channels/564160730845151244/1298192741678710814

alangautham:
![Image](https://github.com/user-attachments/assets/5ed1c04b-4fbb-4265-9157-0059c41ccf43)

Sent email to my app users (around 300) using topics from the appwrite console. Users received mail with other's mail ids exposed. Please refer to the attached screenshot. This is clearly a data breach and not GDPR compliant. Please fix it soon guys!

7oaker:
Hi, would be great if this gets on the bucket list for 2025 ;) 
Would save some trouble, also for data protection laws.

BR

ChiragAgg5k:
@7oaker the team has started working on this, thanks your patience 🙏

7oaker:
Thanks a lot for your efforts @ChiragAgg5k 
i just tested but unfortunately it is the same:

![Image](https://github.com/user-attachments/assets/ce138fcc-035f-4892-8b18-bcae3e0429a3)

Beside that it would be nice to have a option to send an email to all users (i think you call it targets). Otherwise it can be very time consuming to go through all pages if you have lots of users.

The Message feature would be great, but without this fixes you cannot use it in praxis. :( 

Best Regards

[...]

PR #9243, opened 2025-01-19
Fix: email (smtp) to multiple recipients

ChiragAgg5k:
## What does this PR do?

According to issue mentioned below, sending emails to multiple recipients can lead to data privacy issues.
This PR makes the messaging worker send a separate mail for each recipient if the provider is **SMTP** as for sendgrid and mailgun there are api endpoints for _batch-sending_

> Note: If users want to send a single email, they can use `cc`, and don't want to expose emails to all recipients, they can use `bcc`

## Related PRs and Issues

* #8261 

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -59,7 +59,7 @@
         "utopia-php/image": "0.8.*",
         "utopia-php/locale": "0.4.*",
         "utopia-php/logger": "0.6.*",
-        "utopia-php/messaging": "0.14.*",
+        "utopia-php/messaging": "0.16.*",
         "utopia-php/migration": "0.6.*",
         "utopia-php/orchestration": "0.9.*",
         "utopia-php/platform": "0.7.3",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "b1d26d441fe1fb6a26978d99ef01985f",
+    "content-hash": "5e45619efebf755a1753eb91535bf387",
     "packages": [
         {
             "name": "adhocore/jwt",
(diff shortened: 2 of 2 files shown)


src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -245,10 +245,17 @@ private function sendExternalMessage(
...
                 };
 
-                $batches = \array_chunk(
-                    \array_keys($identifiersForProvider),
-                    $adapter->getMaxMessagesPerRequest()
-                );
+                if ($provider->getAttribute('provider') === 'smtp') {
+                    $batches = array_map(
+                        fn ($identifier) => [$identifier],
+                        array_keys($identifiersForProvider)
```
loks0n, line 251: The performance impact is probably negligible, but array_keys is o(n) so preferable to run it once:
e.g.
```
$keys = array_keys($identifiersForProvider);

$batches = $provider->getAttribute('provider') === 'smtp'
    ? array_map(fn($id) => [$id], $keys)
    : array_chunk($keys, $adapter->getMaxMessagesPerRequest());
```

loks0n (approved):

src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -245,10 +245,11 @@ private function sendExternalMessage(
...
 
-                $batches = \array_chunk(
-                    \array_keys($identifiersForProvider),
-                    $adapter->getMaxMessagesPerRequest()
-                );
+                $keys = array_keys($identifiersForProvider);
+
+                $batches = $provider->getAttribute('provider') === 'smtp'
+                    ? array_map(fn ($id) => [$id], $keys)
+                    : array_chunk($keys, $adapter->getMaxMessagesPerRequest());
```
abnegate, line 252: Let's do a benchmark to compare how long it take to send a message with 1000 recipients with an SMTP provider before and after this change

abnegate (approved):

PR #9246, opened 2025-01-20
fix: phone number parsing exception handling

ChiragAgg5k:
## What does this PR do?

When a invalid phone number is passed, the `libphonenumber\PhoneNumberUtil` library would throw a error which since wasn't handled properly would throw a 500 error.

This PR adds proper handling for this error, throwing a 400 error instead.

## Checklist

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -28,6 +28,7 @@ use Appwrite\Utopia\Database\Validator\CustomId;
 use Appwrite\Utopia\Database\Validator\Queries\Identities;
 use Appwrite\Utopia\Request;
 use Appwrite\Utopia\Response;
+use libphonenumber\NumberParseException;
 use libphonenumber\PhoneNumberUtil;
 use MaxMind\Db\Reader;
 use Utopia\Abuse\Abuse;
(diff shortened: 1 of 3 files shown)


app/controllers/api/account.php
```diff
@@ -2467,7 +2468,12 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
                 $abuse = new Abuse($timelimit);
                 if ($abuse->check() && System::getEnv('_APP_OPTIONS_ABUSE', 'enabled') === 'enabled') {
                     $helper = PhoneNumberUtil::getInstance();
-                    $countryCode = $helper->parse($phone)->getCountryCode();
+
+                    try {
+                        $countryCode = $helper->parse($phone)->getCountryCode();
```
christyjacob4, line 2473: Can we try and understand what inputs lead to this exception?

This should then be handled in the Phone() validator and we should have tests for these cases

app/controllers/api/account.php
```diff
@@ -3587,7 +3593,12 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
```
christyjacob4, line 3598: Also this phone number comes from the user object, when creating a user we use a phone validator as well. We need to fix it there too

christyjacob4 (approved):

PR #9250, opened 2025-01-21
chore: skipped job instead of throwing error

ChiragAgg5k:
## What does this PR do?

When renew wasn't required in Certificate.php, it would throw a 500 error that was not required. Instead we just want the job to be skipped.

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Workers/Certificates.php
+++ b/src/Appwrite/Platform/Workers/Certificates.php
@@ -147,7 +147,8 @@ class Certificates extends Action
 
                 // If certificate exists already, double-check expiry date. Skip if job is forced
                 if (!$certificates->isRenewRequired($domain->get(), $log)) {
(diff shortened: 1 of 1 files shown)


src/Appwrite/Platform/Workers/Certificates.php
```diff
@@ -149,7 +149,7 @@ private function execute(Domain $domain, Database $dbForConsole, Mail $queueForM
 
             // If certificate exists already, double-check expiry date. Skip if job is forced
             if (!$skipRenewCheck && !$this->isRenewRequired($domain->get(), $log)) {
-                throw new Exception('Renew isn\'t required.');
```
christyjacob4, line 152: lets do a console::info here and then return

christyjacob4 (changes requested):
Also change the target branch to 1.6.x

christyjacob4 (approved):

PR #9255, opened 2025-01-22
Feat batch usage dump

abnegate:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


src/Appwrite/Platform/Workers/UsageDump.php
```diff
@@ -83,109 +84,73 @@ public function action(Message $message, callable $getProjectDB): void
...
                         } catch (\Exception $e) {
-                            console::error('[' . DateTime::now() . '] failed to calculate database storage for key [' . $key . '] ' . $e->getMessage());
+                            Console::error('[' . DateTime::now() . '] failed to calculate database storage for key [' . $key . '] ' . $e->getMessage());
                         }
                         continue;
                     }
 
+                    $documents = [];
+
                     foreach ($this->periods as $period => $format) {
```
christyjacob4, line 94: I would swap the order of the loops. 
First we loop over the keys in the stats array and for each key we iterate through the periods. Periods are fixed, `1h`, `1d` and `inf`

So in the current scenario if you have 1000 keys, you are going to make 1000 DB calls. the batching here is happening only for the periods. 

I would like us to swap the forloops, so that you are able to batch 1000 documents for each period separately and instead make 3 calls to the DB

abnegate: Couldn't swap the loops because $periods loop depends on $key from the outer loop, but I've moved the DB call to outside both loops, so there's only one write per project.

I've also added caching for the database/collection/collection size calls because they can be re-used within the same projects loop

src/Appwrite/Platform/Workers/UsageDump.php
```diff
@@ -57,230 +55,235 @@ public function action(Message $message, callable $getProjectDB): void
...
-
-            try {
+        try {
+            foreach ($payload['stats'] ?? [] as $stats) {
+                $project = new Document($stats['project'] ?? []);
                 $dbForProject = $getProjectDB($project);
+                $projectDocuments = [];
+                $databaseCache = [];
+                $collectionSizeCache = [];
+
```
christyjacob4, line 65: `console::log('['.DateTime::now().'] Id: '.$project->getId(). ' InternalId: '.$project->getInternalId(). ' Db: '.$project->getAttribute('database').' ReceivedAt: '.$receivedAt. ' Keys: '.$numberOfKeys);`

lets add a log here to indicate the job for a project has been picked up

PR #9262, opened 2025-01-23
feat: custom app schemes

loks0n:
- Replaces the Host validator with a new Redirect validator that supports schemes
- Allows `exp://` and `appwrite-callback-{projectId}://` schemes by default


app/config/collections/platform.php
```diff
@@ -564,6 +564,17 @@
                 'default' => null,
                 'array' => false,
                 'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('scheme'),
```
christyjacob4, line 569: do we have a migration ready for this? In which case the patch needs to be applied before we can release

app/controllers/api/projects.php
```diff
@@ -1768,7 +1769,8 @@
             'name' => $name,
             'key' => $key,
             'store' => $store,
-            'hostname' => $hostname
+            'hostname' => $hostname,
+            'scheme' => $scheme,
```
christyjacob4, line 1773: we also need to update the response model for platform to include the scheme

app/init.php
```diff
@@ -1187,58 +1187,57 @@ function (mixed $value) {
...
-        'hostname' => $request->getHostname(),
-    ], Document::SET_TYPE_APPEND);
-
-    $hostnames = explode(',', System::getEnv('_APP_CONSOLE_HOSTNAMES', ''));
+App::setResource('platforms', function (Document $project, Document $console) {
+    $platforms = [];
+    $platforms[] = $project->getAttribute('platforms', []);
+    $platforms[] = $console->getAttribute('platforms', []);
+    return $platforms;
+}, ['project', 'console']);
```
christyjacob4, line 1195: lets add a line break here for readability

app/init.php
```diff
@@ -1187,58 +1187,57 @@ function (mixed $value) {
...
+        ])) {
+            $hostnames[] = $platform['hostname'];
         }
     }
 
-    return \array_unique($clients);
-}, ['request', 'console', 'project']);
+    return \array_unique($hostnames);
+}, ['platforms']);
+App::setResource('schemes', function (array $platforms) {
```
christyjacob4, line 1222: same here

src/Appwrite/Network/Validator/Redirect.php
```diff
@@ -0,0 +1,81 @@
...
+     */
+    public function isValid($value): bool
+    {
+        if (empty($value)) {
+            return false;
+        }
+
+        $url = parse_url($value);
+        if (!isset($url["scheme"])) {
+            // `parse_url` does not URLs without hostname.
```
christyjacob4, line 57: I think there is a typo here `does not URLs`

src/Appwrite/Network/Validator/Redirect.php
```diff
@@ -0,0 +1,81 @@
...
+            // We handle this scenario with regex
+            if (!preg_match('/^([a-z][a-z0-9+\.-]*):\/+$/i', $value, $matches)) {
```
christyjacob4, line 59: Can we also add an explanation for this regex? And is there a better way to do it without a regex 😬

app/init.php
```diff
@@ -1187,58 +1187,57 @@ function (mixed $value) {
...
-    $hostnames = explode(',', System::getEnv('_APP_CONSOLE_HOSTNAMES', ''));
+App::setResource('platforms', function (Document $project, Document $console) {
+    $platforms = [];
+    $platforms[] = $project->getAttribute('platforms', []);
+    $platforms[] = $console->getAttribute('platforms', []);
+    return $platforms;
+}, ['project', 'console']);
+App::setResource('hostnames', function (array $platforms) {
+    // Always allow console hostname
+    $hostnames = [System::getEnv('_APP_CONSOLE_HOSTNAME', 'console')];
```
christyjacob4, line 1198: I believe this should be `$hostnames = ['console'];` ? 
Because we get and split the `_APP_CONSOLE_HOSTNAME` in the line below

app/init.php
```diff
@@ -1187,58 +1187,57 @@ function (mixed $value) {
 App::setResource('queueForMigrations', function (Connection $queue) {
     return new Migration($queue);
 }, ['queue']);
-App::setResource('clients', function ($request, $console, $project) {
-    $console->setAttribute('platforms', [ // Always allow current host
-        '$collection' => ID::custom('platforms'),
-        'name' => 'Current Host',
-        'type' => Origin::CLIENT_TYPE_WEB,
-        'hostname' => $request->getHostname(),
```
christyjacob4, line 1195: I also see that the current hostname which was previously obtained from `getHostname()` is not used anymore

app/config/collections/platform.php
loks0n: I thought we handled this automatically?

It's just an optional extra attribute

app/controllers/api/projects.php
loks0n: Added

tests/unit/Network/Validators/RedirectTest.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    public function redirectsProvider(): array
+    {
+        return [
+            "localhost" => [["localhost"], [], "http://localhost", true],
+            "localhost-no-scheme" => [["localhost"], [], "localhost", false],
+            "expo scheme" => [[], ["exp"], "exp://192.168.0.1", true],
+            "custom scheme" => [[], ["myapp"], "myapp://", true],
+            "custom scheme triple slash" => [[], ["myapp"], "myapp:///", true],
+            "scheme with special chars" => [[], ["my-app+custom.123"], "my-app+custom.123://", true],
```
stnguyen90, line 18: these aren't allowed for schemes but i guess it's besides the point? 🤷🏼

stnguyen90 (approved):

PR #9263, opened 2025-01-24
feat: add fast2SMS adapter

christyjacob4:
## What does this PR do?
Adds a GEO adapter to send SMS based on the geo location

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -56,6 +62,21 @@ public static function getName(): string
...
+        }
+
+        $providers = System::getEnv('_APP_SMS_PROVIDER', '');
+
+        if (!empty($providers)) {
+            $providers = explode(',', $providers);
+            foreach ($providers as $provider) {
+                $this->dsns[] = new DSN($provider);
+            }
+        }
```
abnegate, line 76: This should still be in `sendInternalSMSMessage`, otherwise it will stop the messaging service from working unless these vars are set, which may not be the case for self-hosted

christyjacob4: Updated thanks 👍

src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -56,6 +62,9 @@ public static function getName(): string
      */
     public function __construct()
     {
+
+        $this->adapter = $this->createInternalSMSAdapter($this->dsns);
```
abnegate, line 66: We still do the same check inside this method

abnegate (approved):

PR #9269, opened 2025-01-25
chore: added teamName to membership redirect url

ChiragAgg5k:
## What does this PR do?

When creating a team invite page using the redirect url passed, a user might want the created team's name as well.Currently, they are only provided the teamId, but they can't fetch as the new user isn't logged in yet. 

This PR adds `teamName` to the redirect url as well.

## Checklist

### DIFF
--- a/app/controllers/api/teams.php
+++ b/app/controllers/api/teams.php
@@ -638,7 +638,7 @@ App::post('/v1/teams/:teamId/memberships')
             $dbForProject->purgeCachedDocument('users', $invitee->getId());
         } else {
             $url = Template::parseURL($url);
-            $url['query'] = Template::mergeQuery(((isset($url['query'])) ? $url['query'] : ''), ['membershipId' => $membership->getId(), 'userId' => $invitee->getId(), 'secret' => $secret, 'teamId' => $teamId]);
+            $url['query'] = Template::mergeQuery(((isset($url['query'])) ? $url['query'] : ''), ['membershipId' => $membership->getId(), 'userId' => $invitee->getId(), 'secret' => $secret, 'teamId' => $teamId, 'teamName' => $team->getAttribute('name')]);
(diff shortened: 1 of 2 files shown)


app/controllers/api/teams.php
```diff
@@ -638,7 +638,7 @@
             $dbForProject->purgeCachedDocument('users', $invitee->getId());
         } else {
             $url = Template::parseURL($url);
-            $url['query'] = Template::mergeQuery(((isset($url['query'])) ? $url['query'] : ''), ['membershipId' => $membership->getId(), 'userId' => $invitee->getId(), 'secret' => $secret, 'teamId' => $teamId]);
+            $url['query'] = Template::mergeQuery(((isset($url['query'])) ? $url['query'] : ''), ['membershipId' => $membership->getId(), 'userId' => $invitee->getId(), 'secret' => $secret, 'teamId' => $teamId, 'teamName' => $team->getAttribute('name')]);
```
christyjacob4, line 641: we probably have an existing test to check for the remaining attributes. Can we add an additional check to also check for name? 

I believe we also need a change in the console to support this right?

ChiragAgg5k: i don't think we have any tests for the other attributes as well since the tests just check the `response`, meanwhile the attributes are added to the `redirect` url sent via email

also the change is just for the end user to create their own team invitation page, i dont think it requires a console change?

ChiragAgg5k: https://github.com/appwrite/appwrite/blob/505a072bdd48c31e6553e2758f3d070407df0036/app/controllers/api/teams.php#L458

according to this currently we don't have our own built-in confirm page so no change required in the console

loks0n (approved):

christyjacob4 (approved):

PR #9271, opened 2025-01-26
chore: override getUserAgent function to set forwarded user agent

ChiragAgg5k:
## What does this PR do?

According to [docs](https://appwrite.io/docs/products/auth/server-side-rendering#forwarding-user-agent), using `client.setForwardedUserAgent()` or setting `X-Forwarded-User-Agent` manually should set the User-Agent with whatever user wants on the backend IF "priviledged to" i.e is using an API key,

but that doesn't seem to be happening atm.

This PR overrides the `getUserAgent` function in the Swoole Request to fix this behaviour.

## Test Plan

| Without using API Key | While using API Key |
|---------------------|------------------------|
| With API Key | Without API Key |

## Checklist


loks0n (changes requested):
Can we add a test for it?

You can assert the created session contains this user agent.

Maybe somewhere in account PHP, where we test SSR patterns

loks0n (approved):

src/Appwrite/Utopia/Request.php
```diff
@@ -180,4 +182,28 @@ public function getHeader(string $key, string $default = ''): string
...
+    *
+    * @param  string  $default
+    * @return string
+    */
+    public function getUserAgent(string $default = ''): string
+    {
+        $forwardedUserAgent = $this->getHeader('x-forwarded-user-agent');
+        if (!empty($forwardedUserAgent)) {
+            $roles = Authorization::getRoles();
+            $isPrivilegedUser = Auth::isPrivilegedUser($roles);
```
stnguyen90, line 199: A privileged user is a Console user. We shouldn't allow console users to override the user agent because then they can spoof requests to the Console.

stnguyen90 (approved):

PR #9314, opened 2025-02-06
Feat: SPA support

Meldiron:
## What does this PR do?

Integrates SPA with appwrite-branded 404, and ability to customize behaviour

## Test Plan

New tests coverage both cases - appwrite-branded 404, and custom 404

- [x] Add Appwrite-branded 404 tst

## Related PRs and Issues

x

## Checklist


tests/e2e/Services/Sites/SitesCustomServerTest.php
```diff
@@ -939,5 +904,142 @@ public function testGetFrameworks(): void
...
+        $this->assertEquals(1, $rules['body']['total']);
+        $this->assertCount(1, $rules['body']['rules']);
+        $this->assertNotEmpty($rules['body']['rules'][0]['domain']);
+
+        $domain = $rules['body']['rules'][0]['domain'];
+
+        $proxyClient = new Client();
+        $proxyClient->setEndpoint('http://' . $domain);
+
+        \var_dump($domain);
```
Meldiron, line 1014:
```suggestion
```

src/Appwrite/Platform/Modules/Sites/Http/Sites/Create.php
```diff
@@ -76,7 +76,7 @@ public function __construct()
             ->param('outputDirectory', '', new Text(8192, 0), 'Output Directory for site.', true)
             ->param('subdomain', '', new CustomId(), 'Unique custom sub-domain. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.', true)
             ->param('buildRuntime', '', new WhiteList(array_keys(Config::getParam('runtimes')), true), 'Runtime to use during build step.')
-            ->param('adapter', '', new Text(8192, 0), 'Framework adapter. Allows: static, ssr', true)
+            ->param('adapter', '', new WhiteList(['static', 'ssr']), 'Framework adapter defining rendering strategy. Allowed values are: static, ssr', true)
```
vermakhushboo, line 79: We call it `ssg` in detection library. Do we wish to call it `static` or `ssg` here?

Meldiron: I recall discussions calling it something else than adapter too.

While we call it adapter, static and ssr is accurate.

Once we rename it to strategy/rendering, then we should replace whitelist too

tests/e2e/Services/Sites/SitesCustomServerTest.php
```diff
@@ -1010,41 +1010,6 @@ public function testGetDeployment(): void
         $this->cleanupSite($siteId);
     }
 
-    // public function testLoadSite(): void
```
eldadfux, line 1013: leftovers

app/controllers/general.php
```diff
@@ -418,6 +423,13 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
                 requestTimeout: 30
             );
 
+            // Branded 404 override
+            if ($executionResponse['statusCode'] === 404 && $resource->getAttribute('adapter', '') === 'static') {
```
eldadfux, line 427: When do we use this file vs the user own 404 file?

app/controllers/general.php
```diff
@@ -360,6 +360,11 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
             'APPWRITE_VCS_ROOT_DIRECTORY' => $deployment->getAttribute('providerRootDirectory', ''),
         ]);
 
+        // SPA fallbackFile override
+        if ($resource->getAttribute('adapter', '') === 'static' && $resource->getAttribute('fallbackFile', '') !== '') {
```
eldadfux, line 364: Instead of having a special env var here, can we just overwrite the path we sent to the runtime? Won't it make the rest of the flow simpler?

app/views/general/404.phtml
```diff
@@ -0,0 +1,185 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
```
eldadfux, line 4: Can you share a screenshot? Did we test this across screen sizes / different browsers?

tests/e2e/Services/Sites/SitesCustomServerTest.php
Meldiron: It was leftover, this is removing it. it was just piece of comment as "template" for tests, before they were implemented. Now we already got dozens of tests with sites, including flow with router. This was just a leftover

Meldiron: We use ours only if site is static, and site is not SPA

app/controllers/general.php
Meldiron: We use ours only if site is static, and site is not SPA

app/views/general/404.phtml
Meldiron: Chen approved it already

app/controllers/general.php
Meldiron: What we are setting is just a path, inside build output directory of developer.
Static server inside runtime then recieves this env var, so it can do proper SPA routing

eldadfux: This is valid only to the SPA runtime, right?

eldadfux (approved):

PR #9315, opened 2025-02-06
Integrate detection library with Appwrite

vermakhushboo:
## What does this PR do?

Detect runtime and framework in `detect` and `listRepo` endpoints in `VCS` controller
Detect rendering strategy and fallback file in `builds` worker

## Test Plan

new tests added

## Related PRs and Issues

x

## Checklist


app/controllers/api/vcs.php
```diff
@@ -556,18 +570,22 @@
...
+            new SDKResponse(
+                code: Response::STATUS_CODE_OK,
+                model: Response::MODEL_FRAMEWORK_DETECTION,
             )
         ]
     ))
     ->param('installationId', '', new Text(256), 'Installation Id')
     ->param('providerRepositoryId', '', new Text(256), 'Repository Id')
     ->param('providerRootDirectory', '', new Text(256, 0), 'Path to Root Directory', true)
+    ->param('type', '', new WhiteList(['runtime', 'framework']), 'Detector type. Must be one of the following: runtime, framework', true)
```
Meldiron, line 584:
```suggestion
    ->param('type', '', new WhiteList(['runtime', 'framework']), 'Detector type. Must be one of the following: runtime, framework')
```
The `true` there means it is optional. I think it should be required, otherwise we don't know what kind of detection they want

app/controllers/api/vcs.php
```diff
@@ -544,8 +557,9 @@
         ]), Response::MODEL_VCS_CONTENT_LIST);
     });
 
-App::post('/v1/vcs/github/installations/:installationId/providerRepositories/:providerRepositoryId/detection')
-    ->desc('Detect runtime settings from source code')
+App::post('/v1/vcs/github/installations/:installationId/providerRepositories/:providerRepositoryId/detections')
```
Meldiron, line 560: `providerRepositories` -> `provider-repositories`

Meldiron: Alternative is to have just a simple

```
/v1/vcs/github/installations/:installationId/detections
```

And accept providerRepositoryId as parameter. Let's suggest it, see which approach is more liked by Eldad/Christy

app/controllers/api/vcs.php
```diff
@@ -593,32 +611,100 @@
...
-            ->addDetector(new PHP())
-            ->addDetector(new Python())
-            ->addDetector(new Dart())
-            ->addDetector(new Swift())
-            ->addDetector(new Ruby())
-            ->addDetector(new Java())
-            ->addDetector(new CPP())
-            ->addDetector(new Deno())
-            ->addDetector(new Dotnet());
+        $packagerName = $detectedPackager ? $detectedPackager->getName() : 'npm';
```
Meldiron, line 621:
```suggestion
        $packagerName = !\is_null($detectedPackager) ? $detectedPackager->getName() : 'npm';
```

Using is_null specifically is a bit more exact

app/controllers/api/vcs.php
```diff
@@ -593,32 +611,100 @@
...
 
-        $runtime = $detectorFactory->detect();
+        $detection = [];
```
Meldiron, line 623: I think we dont need this line

app/controllers/api/vcs.php
```diff
@@ -593,32 +611,100 @@
...
-        $runtimeDetail = \array_reverse(\array_filter(\array_keys($runtimes), function ($key) use ($runtime, $runtimes) {
-            return $runtimes[$key]['key'] === $runtime;
-        }))[0] ?? '';
+        if ($type === 'framework') {
+            $detection = [
+                'framework' => '',
+                'installCommand' => '',
+                'buildCommand' => '',
+                'outputDirectory' => '',
+            ];
```
Meldiron, line 631: We can use `new Document([...])` instead of named array

app/controllers/api/vcs.php
```diff
@@ -593,32 +611,100 @@
...
+                ->addOption(new SvelteKit())
+                ->addOption(new NextJs());
+
+            $detectedFramework = $frameworkDetector->detect();
+
+            if ($detectedFramework) {
+                $framework = $detectedFramework->getName();
+                $detection['installCommand'] = $detectedFramework->getInstallCommand();
+                $detection['buildCommand'] = $detectedFramework->getBuildCommand();
+                $detection['outputDirectory'] = $detectedFramework->getOutputDirectory();
```
Meldiron, line 648: We can use `setAttribute`

app/controllers/api/vcs.php
```diff
@@ -593,32 +611,100 @@
...
+                ->addOption(new Flutter())
+                ->addOption(new Nuxt())
+                ->addOption(new Astro())
+                ->addOption(new Remix())
+                ->addOption(new SvelteKit())
+                ->addOption(new NextJs());
+
+            $detectedFramework = $frameworkDetector->detect();
+
+            if ($detectedFramework) {
```
Meldiron, line 644: If we cannot detect, let's throw error (some 4xx)

app/controllers/api/vcs.php
```diff
@@ -593,32 +611,100 @@
...
+                $framework = $detectedFramework->getName();
+                $detection['installCommand'] = $detectedFramework->getInstallCommand();
+                $detection['buildCommand'] = $detectedFramework->getBuildCommand();
+                $detection['outputDirectory'] = $detectedFramework->getOutputDirectory();
+            }
 
-        $detection = [];
-        $detection['runtime'] = $runtimeDetail;
+            if (!empty($framework)) {
```
Meldiron, line 651: I dont think there is scenario when this will be empty

app/controllers/api/vcs.php
```diff
@@ -593,32 +611,100 @@
...
+                $frameworks = Config::getParam('frameworks');
+                $frameworkDetail = \array_reverse(\array_filter(\array_keys($frameworks), function ($key) use ($framework, $frameworks) {
+                    return $frameworks[$key]['key'] === $framework;
+                }))[0] ?? '';
+                $detection['framework'] = $frameworkDetail;
```
Meldiron, line 656: Lets also throw exception here, when we cant find framework

Meldiron: You can probably trigger this on purpose by changing .env, and removing framework from list of allowed frameworks

vermakhushboo:
Return `other` when framework is not detected. For runtime, in `detection` endpoint, throw an exception if runtime is not detection, where as in `listRepos`, return null or empty string if runtime is not detected, so that it doesn't affect the detection of other repos.

app/controllers/api/vcs.php
vermakhushboo: Christy approved this one: `/v1/vcs/github/installations/:installationId/detections`

app/controllers/api/vcs.php
vermakhushboo: It would be when it is `other` framework

src/Appwrite/Utopia/Response/Model/ProviderRepository.php
```diff
@@ -47,6 +47,12 @@ public function __construct()
                 'default' => '',
                 'example' => 'node',
             ])
+            ->addRule('framework', [
```
vermakhushboo, line 50: Create separate response models

src/Appwrite/Utopia/Response/Model/RuntimeProviderRepository.php
```diff
@@ -5,7 +5,7 @@
 use Appwrite\Utopia\Response;
 use Appwrite\Utopia\Response\Model;
 
-class ProviderRepository extends Model
+class RuntimeProviderRepository extends Model
```
Meldiron, line 8: Lets make this extend ProviderRepository too, like comment above

app/controllers/api/vcs.php
```diff
@@ -593,32 +611,103 @@
...
+            $detection->setAttribute('framework', $frameworkDetail);
 
-        $runtime = $detectorFactory->detect();
+            $response->dynamic($detection, Response::MODEL_FRAMEWORK_DETECTION);
+        } else {
+            $detection = new Document([
+                'runtime' => '',
+                'commands' => '',
+                'entrypoint' => '',
+            ]);
```
Meldiron, line 666: Duplicate code, can be common above if/else

app/controllers/api/vcs.php
```diff
@@ -593,32 +611,103 @@
...
+            if ($detectedFramework) {
+                $framework = $detectedFramework->getName();
+                $detection->setAttribute('installCommand', $detectedFramework->getInstallCommand());
+                $detection->setAttribute('buildCommand', $detectedFramework->getBuildCommand());
+                $detection->setAttribute('outputDirectory', $detectedFramework->getOutputDirectory());
+            } else {
+                $framework = 'other';
+                $detection->setAttribute('installCommand', '');
+                $detection->setAttribute('buildCommand', '');
+                $detection->setAttribute('outputDirectory', './');
```
Meldiron, line 651:
```suggestion
                $detection->setAttribute('outputDirectory', '');
```

app/controllers/api/vcs.php
```diff
@@ -593,32 +611,103 @@
...
-            ->addDetector(new Ruby())
-            ->addDetector(new Java())
-            ->addDetector(new CPP())
-            ->addDetector(new Deno())
-            ->addDetector(new Dotnet());
+            $frameworks = Config::getParam('frameworks');
+            $frameworkDetail = \array_reverse(\array_filter(\array_keys($frameworks), function ($key) use ($framework, $frameworks) {
+                return $frameworks[$key]['key'] === $framework;
+            }))[0] ?? '';
+            $detection->setAttribute('framework', $frameworkDetail);
```
Meldiron, line 658: Can you please share with me what value of such $frameworkDetail is?  If it's just string, let's explore if we can simplify this code, I am almost certain it can be done.

app/controllers/api/vcs.php
```diff
@@ -593,32 +611,103 @@
...
+            }
 
-        $detection = [];
-        $detection['runtime'] = $runtimeDetail;
+            if (!empty($runtime)) {
+                $runtimes = Config::getParam('runtimes');
+                $runtimeDetail = \array_reverse(\array_filter(\array_keys($runtimes), function ($key) use ($runtime, $runtimes) {
+                    return $runtimes[$key]['key'] === $runtime;
+                }))[0] ?? '';
+                $detection->setAttribute('runtime', $runtimeDetail);
```
Meldiron, line 704: Same as comment above, I think this can be simplified.

app/controllers/api/vcs.php
```diff
@@ -593,32 +611,103 @@
...
+            } else {
+                throw new Exception(Exception::GENERAL_BAD_REQUEST, 'Runtime not detected.');
```
Meldiron, line 706: Let's have proper error type for this, avoid using gneric GENERAL_BAD_REQUEST as rule of thumb.

app/controllers/api/vcs.php
```diff
@@ -884,7 +1019,7 @@
         $repository['organization'] = $installation->getAttribute('organization', '');
         $repository['provider'] = $installation->getAttribute('provider', '');
 
-        $response->dynamic(new Document($repository), Response::MODEL_PROVIDER_REPOSITORY);
+        $response->dynamic(new Document($repository), Response::MODEL_RUNTIME_PROVIDER_REPOSITORY);
```
Meldiron, line 1022: This endpoint doesnt seem to be setting runtime. Let's check with Arman to understand when this is beign used, so we can decide if we want type parameter here too.

app/controllers/api/vcs.php
```diff
@@ -830,7 +965,7 @@
...
```
Meldiron, line 968: Let's have generic provider repository (without runtime or framework) for this scenario, since we are not doing any detection.

This is nice follow-up to my first comments when I suggsted abstract base class to extend - this is good reas to make it non-abstract and actually use it in responses

app/controllers/api/vcs.php
```diff
@@ -673,39 +766,81 @@
...
-
-                    $repo['runtime'] = $runtimeDetail;
-                } catch (Throwable $error) {
-                    $repo['runtime'] = "";
-                    Console::warning("Runtime not detected for " . $repo['organization'] . "/" . $repo['name']);
+        $repos = batch(\array_map(function ($repo) use ($type, $github) {
+            return function () use ($repo, $type, $github) {
+                $files = $github->listRepositoryContents($repo['organization'], $repo['name'], '');
+                $files = \array_column($files, 'name');
+                $languages = $github->listRepositoryLanguages($repo['organization'], $repo['name']);
```
Meldiron, line 773: We only need languages for runtime detection, doing this up here means we also do this for frameworks which isnt nssessary

app/controllers/api/vcs.php
```diff
@@ -673,39 +766,81 @@
...
+                        if ($detectedFramework) {
+                            $framework = $detectedFramework->getName();
+                        } else {
+                            $framework = 'other';
+                        }
+
+                        $frameworks = Config::getParam('frameworks');
+                        $frameworkDetail = \array_reverse(\array_filter(\array_keys($frameworks), function ($key) use ($framework, $frameworks) {
+                            return $frameworks[$key]['key'] === $framework;
+                        }))[0] ?? '';
```
Meldiron, line 796: Try to simplify

app/controllers/api/vcs.php
```diff
@@ -673,39 +766,81 @@
...
+                        $repo['framework'] = $frameworkDetail;
+                    } catch (Throwable $error) {
```
Meldiron, line 798: Let's try to avoid this try/catch block. It should either not be needed at all, or only have a few lines of code in it

app/controllers/api/vcs.php
```diff
@@ -673,39 +766,81 @@
...
+                } else {
+                    try {
+                        $strategies = [
+                            new Strategy(Strategy::FILEMATCH),
+                            new Strategy(Strategy::LANGUAGES),
+                            new Strategy(Strategy::EXTENSION),
+                        ];
+
+                        foreach ($strategies as $strategy) {
+                            $runtimeDetector = new Runtime($strategy === Strategy::LANGUAGES ? $languages : $files, $strategy, 'npm');
```
Meldiron, line 811: Let's not assume NPM here, do detection for package manager

app/controllers/api/vcs.php
```diff
@@ -673,39 +766,81 @@
...
-                    $repo['runtime'] = "";
-                    Console::warning("Runtime not detected for " . $repo['organization'] . "/" . $repo['name']);
+        $repos = batch(\array_map(function ($repo) use ($type, $github) {
+            return function () use ($repo, $type, $github) {
+                $files = $github->listRepositoryContents($repo['organization'], $repo['name'], '');
+                $files = \array_column($files, 'name');
+                $languages = $github->listRepositoryLanguages($repo['organization'], $repo['name']);
+                if ($type === 'framework') {
+                    try {
+                        $frameworkDetector = new Framework($files);
```
Meldiron, line 776: Let's detect package manager too, and pass it here.

app/controllers/api/vcs.php
```diff
@@ -673,39 +766,81 @@
...
+                                break;
+                            }
+                        }
+
+                        if (!empty($runtime)) {
+                            $runtimes = Config::getParam('runtimes');
+                            $runtimeDetail = \array_reverse(\array_filter(\array_keys($runtimes), function ($key) use ($runtime, $runtimes) {
+                                return $runtimes[$key]['key'] === $runtime;
+                            }))[0] ?? '';
+                            $repo['runtime'] = $runtimeDetail;
```
Meldiron, line 838: Let's see if we can simplify

app/controllers/api/vcs.php
```diff
@@ -673,39 +766,81 @@
...
+                        }
+                    } catch (Throwable $error) {
+                        $repo['runtime'] = "";
+                        Console::warning("Runtime not detected for " . $repo['organization'] . "/" . $repo['name']);
```
Meldiron, line 842: Lets not use warnings; Check with Arman if empty string is correct default here

docker-compose.yml
```diff
@@ -962,7 +962,7 @@ services:
     hostname: exc1
     <<: *x-logging
     stop_signal: SIGINT
-    image: openruntimes/executor:0.7.4
+    image: openruntimes-executor-2
```
vermakhushboo, line 965: Revert this and update with released version of executor

app/controllers/api/vcs.php
vermakhushboo: No, it cannot be common, both runtime and framework have different fields

app/controllers/api/vcs.php
vermakhushboo: Can we use abstract providerRepository model instead?

app/controllers/api/vcs.php
vermakhushboo: Packager is not required here since we are just detecting framework/ runtime and not any commands.

app/controllers/api/vcs.php
```diff
@@ -721,39 +818,95 @@
...
+                                if ($config['key'] === $runtime) {
+                                    $runtimeDetail = $key;
+                                    break;
+                                }
+                            }
+                            $repo['runtime'] = $runtimeDetail;
+                        }
+                    } catch (Throwable $error) {
+                        $repo['runtime'] = "";
+                        throw new Exception(Exception::FUNCTION_RUNTIME_NOT_DETECTED, "Runtime not detected for " . $repo['organization'] . "/" . $repo['name']);
```
Meldiron, line 908: If this is listX endpoint, do not throw

composer.json
```diff
@@ -104,5 +105,11 @@
             "php-http/discovery": true,
             "tbachert/spi": true
         }
-    }
+    },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/utopia-php/detector"
```
Meldiron, line 112: Remove before merge

composer.json
```diff
@@ -53,6 +53,7 @@
         "utopia-php/cli": "0.15.*",
         "utopia-php/config": "0.2.*",
         "utopia-php/database": "0.59.0",
+        "utopia-php/detector": "dev-feat-pseudocode-draft2 as 0.1.99",
```
Meldiron, line 56: Release before merge

tests/e2e/Services/Sites/SitesCustomServerTest.php
```diff
@@ -324,6 +324,40 @@ public function testVariablesE2E(): void
         $this->cleanupSite($siteId);
     }
 
+    public function testRenderingDetectionSSR(): void
```
Meldiron, line 327: Lets have similar test but for functions, one for node and one for PHP

Meldiron: Nevermind, not relevant.

Meldiron (approved):

app/config/site-templates.php
eldadfux, line 1: Maybe we can have templates/functions.php and templates/sites.php?

docker-compose.yml
```diff
@@ -976,7 +976,7 @@ services:
       # It's not possible to share mount file between 2 containers without host mount (copying is too slow)
       - /tmp:/tmp:rw
     environment:
-      - OPR_EXECUTOR_IMAGE_PULL=enabled
+      - OPR_EXECUTOR_IMAGE_PULL=disabled
```
eldadfux, line 979: Is this intentional?

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
```diff
@@ -705,6 +708,33 @@ protected function buildDeployment(Device $deviceForFunctions, Device $deviceFor
                 throw new \Exception('Build size should be less than ' . number_format($buildSizeLimit / 1048576, 2) . ' MBs.');
             }
 
+            if ($resource->getCollection() === 'sites' && empty($resource->getAttribute('adapter'))) {
```
eldadfux, line 711: Instead of checking for sites or functions, we should have feature flags for the build worker that we can pass true or false from the API when triggering the job to control how we use them. Each worker, API container should be designed as general purpose service for more flexible design. In this case the job should have a boolean attribute called `detection`.

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
```diff
@@ -705,6 +708,33 @@ protected function buildDeployment(Device $deviceForFunctions, Device $deviceFor
...
+                $listFilesCommand = "cd /usr/local/build && cd " . \escapeshellarg($resource->getAttribute('outputDirectory')) . " && find . -name 'node_modules' -prune -o -type f -print";
```
eldadfux, line 712: Todo to replace this with a properly structured execution command later on.

src/Appwrite/Platform/Modules/Sites/Http/Sites/Update.php
```diff
@@ -98,7 +98,7 @@ public function __construct()
             ->callback([$this, 'action']);
     }
 
-    public function action(string $siteId, string $name, string $framework, bool $enabled, int $timeout, string $installCommand, string $buildCommand, string $outputDirectory, string $buildRuntime, string $adapter, ?string $fallbackFile, string $installationId, ?string $providerRepositoryId, string $providerBranch, bool $providerSilentMode, string $providerRootDirectory, string $specification, Request $request, Response $response, Database $dbForProject, Document $project, Event $queueForEvents, Build $queueForBuilds, Database $dbForPlatform, GitHub $github)
+    public function action(string $siteId, string $name, string $framework, bool $enabled, int $timeout, string $installCommand, string $buildCommand, string $outputDirectory, string $buildRuntime, string $adapter, string $fallbackFile, string $installationId, ?string $providerRepositoryId, string $providerBranch, bool $providerSilentMode, string $providerRootDirectory, string $specification, Request $request, Response $response, Database $dbForProject, Document $project, Event $queueForEvents, Build $queueForBuilds, Database $dbForPlatform, GitHub $github)
```
eldadfux, line 101: Now when we have class per endpoint, I think its a better convention to break lines here.

src/Appwrite/Utopia/Response/Model/ProviderRepository.php
```diff
@@ -41,12 +41,6 @@ public function __construct()
                 'default' => false,
                 'example' => true,
             ])
-            ->addRule('runtime', [
-                'type' => self::TYPE_STRING,
```
eldadfux, line 45: What do we do instead?

src/Executor/Executor.php
```diff
@@ -249,6 +249,31 @@ public function createExecution(
         return $response['body'];
     }
 
+    public function createCommand(
```
eldadfux, line 252: where do we use this? is it related to detection?

app/controllers/api/vcs.php
```diff
@@ -721,39 +818,86 @@
...
-                    $repo['runtime'] = "";
-                    Console::warning("Runtime not detected for " . $repo['organization'] . "/" . $repo['name']);
+                    $strategies = [
+                        new Strategy(Strategy::FILEMATCH),
+                        new Strategy(Strategy::LANGUAGES),
+                        new Strategy(Strategy::EXTENSION),
+                    ];
+
+                    foreach ($strategies as $strategy) {
+                        $redector = new Runtime($strategy === Strategy::LANGUAGES ? $languages : $files, $strategy, $packager);
```
eldadfux, line 868: redector?

docker-compose.yml
Meldiron: No, thanks.

app/config/site-templates.php
Meldiron: Agree

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
Meldiron: Do you agree to ignore it in this PR and have task for next week? I would like to separate feature and code-quality, so we can do QA sooner

Meldiron: (plus this is already problem in many many places of worker, and Appwrite router)

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
Meldiron: added

src/Appwrite/Utopia/Response/Model/ProviderRepository.php
Meldiron: ProviderRepository is (as of right now) only used as response model for endpoint that creates repository on GitHub.

Actual list of repositories either gives ProviderRepositoryRuntime, or ProviderRepositoryFramework - those extend this model and add the specific attributes

src/Executor/Executor.php
Meldiron: yes, in builds worker.
1. Start build container, remove=false
2. Execute command in container to get list of output files
3. Detect SSR or static based on files in output directory

app/controllers/api/vcs.php
Meldiron: Detector. Welp, AI would not make that mistake 🙈

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
eldadfux: Will be taken care of in a separate PR

eldadfux (approved):

PR #9325, opened 2025-02-08
refactor: migrate Realtime::send calls to queueForRealtime

ChiragAgg5k:
## What does this PR do?

Replace direct usage of `Realtime::send` with the new `queueForRealtime` resource across workers and other parts of the codebase for consistent realtime message handling.

## Checklist


src/Appwrite/Platform/Workers/Certificates.php
```diff
@@ -367,50 +371,34 @@ private function updateDomainDocuments(string $certificateId, string $domain, bo
...
+                ->trigger();
 
             /** Trigger Functions */
             $queueForFunctions
-                ->setProject($project)
-                ->setEvent('rules.[ruleId].update')
-                ->setParam('ruleId', $rule->getId())
-                ->setPayload($rule->getArrayCopy(array_keys($ruleModel->getRules())))
+                ->from($queueForEvents)
                 ->trigger();
```
loks0n, line 390: This is slightly confusing, should we call queueForFunctions->from before the calls to `setQueue` and `setClass`?

src/Appwrite/Platform/Workers/Certificates.php
```diff
@@ -367,50 +371,34 @@ private function updateDomainDocuments(string $certificateId, string $domain, bo
...
-                channels: $target['channels'],
-                roles: $target['roles']
-            );
-            Realtime::send(
-                projectId: $project->getId(),
-                payload: $rule->getArrayCopy(),
-                events: $allEvents,
-                channels: $target['channels'],
-                roles: $target['roles']
-            );
```
loks0n, line 413: What we have is okayish, I wonder if there's a better solution to cover this scenario?

With both console and project events.

src/Appwrite/Platform/Workers/Certificates.php
ChiragAgg5k: well i can move webhooks trigger to last, but that will modify the order of how we call these triggers throughout the repo, which is webhooks > functions > realtime.

in my opinion what we had before was more confusing and this is a bit better. best would be to i guess inject queueForWebhooks as well. should i do that?

src/Appwrite/Platform/Workers/Certificates.php
ChiragAgg5k: i don't follow... do you want both `console` and `project->getId()`'s events be triggered in a single call?

ChiragAgg5k:
Testing 👍

Screenshot 2025-02-18 at 5 15 35 PM

src/Appwrite/Event/Realtime.php
```diff
@@ -53,17 +82,33 @@ public function trigger(): string|bool
...
-            payload: $this->getRealtimePayload(),
-            events: $allEvents,
-            channels: $target['channels'],
-            roles: $target['roles'],
-            options: [
-                'permissionsChanged' => $target['permissionsChanged'],
-                'userId' => $this->getParam('userId')
-            ]
-        );
+        if (!empty($this->getTargets())) {
```
loks0n, line 85: Small doubt about the name of this property, but can't think of something better.

loks0n (approved):

src/Appwrite/Event/Realtime.php
```diff
@@ -24,6 +31,28 @@ public function getRealtimePayload(): array
         return $payload;
     }
 
+    /**
+     * Set targets for this realtime event.
```
abnegate, line 35: Let's call this `subscribers` instead of `targets`

src/Appwrite/Platform/Workers/Databases.php
```diff
@@ -598,31 +581,42 @@ protected function deleteByGroup(string $collection, array $queries, Database $d
...
-            ]
-        );
+        $queueForRealtime
+            ->setProject($project)
+            ->setTargets(['console'])
+            ->setEvent($event)
+            ->setParam('databaseId', $database->getId())
+            ->setParam('collectionId', $collection->getId());
+
+        if ($attribute !== null) {
```
abnegate, line 609: We should also check !attribute->isEmpty and same for the index

abnegate (approved):

PR #9326, opened 2025-02-08
Feat: preview branding

Meldiron:
## What does this PR do?

Adds Appwrit banner to preview deployments

## Test Plan

New tests added

## Related PRs and Issues

x

## Checklist


src/Appwrite/Platform/Modules/Sites/Transformers/Preview.php
eldadfux, line 1: This would not be the place to handle this. We need a library to handle all possible proxy's router parser/transformation/filters. This library will end up living on the edge (punch not intended).

We will need here a strong interface that will later allow us to transform images, css, js files and more.

eldadfux (changes requested):

src/Appwrite/Platform/Modules/Sites/Transformers/Preview.php
Meldiron: Updated

app/controllers/general.php
```diff
@@ -446,6 +448,20 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
...
             );
 
+
+            // Branded banner for previews
+            $transformation = new Transformation(new Preview($executionResponse['body']));
+            if ($type === 'deployment' && $transformation->isValid($executionResponse['headers'])) {
+                $transformation->transform();
+                $executionResponse['body'] = $transformation->getOutput();
+
+                foreach ($executionResponse['headers'] as $key => $value) {
```
eldadfux, line 458: Why are we doing this instead of just setting the content-length directly?

app/controllers/general.php
```diff
@@ -446,6 +448,20 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
                 requestTimeout: 30
             );
 
+
+            // Branded banner for previews
+            $transformation = new Transformation(new Preview($executionResponse['body']));
+            if ($type === 'deployment' && $transformation->isValid($executionResponse['headers'])) {
```
eldadfux, line 454: I would expect the `isValid` method to be protected and used inside the transform method. After all, if the transformation is not valid you shouldn't allow it in any way. This way you can't enforce dev will write the code this way every time they need to transform something.

src/Appwrite/Transformation/Transformation.php
```diff
@@ -0,0 +1,28 @@
...
+class Transformation
+{
+    public function __construct(protected Adapter $adapter)
+    {
+    }
+
+    /**
+     * @param array<mixed> $traits
+     */
+    public function isValid(array $traits): bool
```
eldadfux, line 14: This should be the protected method who always triggers before `return $this->adapter->transform();`

src/Appwrite/Transformation/Transformation.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace Appwrite\Transformation;
+
+class Transformation
+{
+    public function __construct(protected Adapter $adapter)
```
eldadfux, line 7: I would expect this class to accept an array of transformers in specific order and allow them to run one after the other. $transformation->add(Adapter $adapter) kind of thing.

eldadfux (changes requested):

app/controllers/general.php
Meldiron: This way I can do `tolowercase` during check, and ensure it works even if response from runtime has uppercase letters - headers as per RFC are case-insensitive

eldadfux (approved):

PR #9329, opened 2025-02-09
Catch invalid index

fogelito:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/databases.php
```diff
@@ -393,6 +394,8 @@ function updateAttribute(
             throw new Exception(Exception::ATTRIBUTE_NOT_FOUND);
         } catch (LimitException) {
             throw new Exception(Exception::ATTRIBUTE_LIMIT_EXCEEDED);
+        } catch (indexException $e) {
```
abnegate, line 397: Let's fix casing here

app/controllers/general.php
```diff
@@ -837,6 +837,9 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
             case 'Utopia\Database\Exception\Dependency':
                 $error = new AppwriteException(AppwriteException::INDEX_DEPENDENCY, null, previous: $error);
                 break;
+            case 'Utopia\Database\Exception\Index':
+                $error = new AppwriteException(AppwriteException::INDEX_INVALID, $error->getMessage(), previous: $error);
+                break;
```
abnegate, line 842: We want to move away from catching expected exceptions in the general handler, let's add this only where needed in the databases controller/worker

fogelito: 100% removed 👍

app/controllers/api/databases.php
fogelito: Done!

abnegate (approved):

PR #9332, opened 2025-02-10
feat(builds): check if function is blocked before building

basert:
Before building a function, check if it's blocked. This ensures we don't keep building blocked functions due to e.g. webhooks.

### DIFF
--- a/src/Appwrite/Platform/Workers/Builds.php
+++ b/src/Appwrite/Platform/Workers/Builds.php
@@ -129,11 +129,11 @@ class Builds extends Action
 
         $function = $dbForProject->getDocument('functions', $functionId);
         if ($function->isEmpty()) {
-            throw new \Exception('Function not found', 404);
+            throw new \Exception('Function not found');
         }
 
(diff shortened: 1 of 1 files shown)


src/Appwrite/Platform/Workers/Builds.php
```diff
@@ -130,6 +132,10 @@ protected function buildDeployment(Device $deviceForFunctions, Func $queueForFun
             throw new \Exception('Function not found', 404);
         }
 
+        if ($isResourceBlocked($project, RESOURCE_TYPE_FUNCTIONS, $functionId)) {
+            throw new \Exception('Function blocked', 403);
```
christyjacob4, line 136:
```suggestion
            throw new \Exception('Function blocked');
```
Since this is a worker, the exception shouldn't use a HTTP error code. The exception on L132 is also incorrect.

christyjacob4 (changes requested):
Please address the comment

christyjacob4 (approved):

PR #9336, opened 2025-02-11
Feat key segmented usage

abnegate:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/general.php
```diff
@@ -770,8 +771,9 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
     ->inject('project')
     ->inject('logger')
     ->inject('log')
+    ->inject('apiKey')
     ->inject('queueForStatsUsage')
-    ->action(function (Throwable $error, App $utopia, Request $request, Response $response, Document $project, ?Logger $logger, Log $log, StatsUsage $queueForStatsUsage) {
```
Meldiron, line 774: Leftover, not used

app/controllers/api/databases.php
```diff
@@ -3136,10 +3137,10 @@ function updateAttribute(
     ->inject('response')
     ->inject('dbForProject')
     ->inject('user')
+    ->inject('apiKey')
     ->inject('queueForEvents')
     ->inject('queueForStatsUsage')
-    ->inject('mode')
-    ->action(function (string $databaseId, string $documentId, string $collectionId, string|array $data, ?array $permissions, Response $response, Database $dbForProject, Document $user, Event $queueForEvents, StatsUsage $queueForStatsUsage, string $mode) {
+    ->action(function (string $databaseId, string $documentId, string $collectionId, string|array $data, ?array $permissions, Response $response, Database $dbForProject, Document $user, ?AuthKey $apiKey, Event $queueForEvents, StatsUsage $queueForStatsUsage) {
```
Meldiron, line 3143: Leftover? We inject, but dont use

Meldiron: Seems to be the case with all changes in this file; databases.php

app/init.php
```diff
@@ -1942,3 +1943,13 @@ function getDevice(string $root, string $connection = ''): Device
 
     return '';
 }, ['request']);
+
+App::setResource('apiKey', function (Request $request, Document $project): ?Key {
+    $key = $request->getHeader('x-appwrite-key');
+
+    if (empty($key)) {
+        return null;
+    }
```
Meldiron, line 1952: Let's make `Key::decode` handle this case, so it returns guestKey instad. This prevents `null` from ever hapenning, and makes later use much safer

src/Appwrite/Auth/Key.php
```diff
@@ -0,0 +1,158 @@
...
+                    key: System::getEnv('_APP_OPENSSL_KEY_V1'),
+                    algo: 'HS256',
+                    maxAge: 86400,
+                    leeway: 0
+                );
+
+                try {
+                    $payload = $jwtObj->decode($secret);
+                } catch (JWTException) {
+                    $expired = true;
```
Meldiron, line 106: Could be more reasons than just expired. Should we rename this to `$valid`?

src/Appwrite/Auth/Key.php
```diff
@@ -0,0 +1,158 @@
...
+        if (\str_contains($key, '_')) {
+            [$type, $secret] = \explode('_', $key, 2);
+        } else {
+            $type = API_KEY_STANDARD;
+            $secret = $key;
+        }
+
+        $role = Auth::USER_ROLE_APPS;
+        $roles = Config::getParam('roles', []);
+        $scopes = $roles[Auth::USER_ROLE_APPS]['scopes'] ?? [];
```
Meldiron, line 83: This on root level feels like by default, scopes of app rule will be used.

But it seems to only be used in dynamic key. Can we move it inside it's switch/case?

tests/e2e/Services/Migrations/MigrationsBase.php
```diff
@@ -579,6 +589,21 @@ public function testAppwriteMigrationDatabasesDocument(array $data)
...
+
+        $finalStats = $this->client->call(Client::METHOD_GET, '/project/usage', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()), [
+            'startDate' => UsageTest::getYesterday(),
+            'endDate' => UsageTest::getTomorrow(),
+        ]);
+
+        // Ensure database reads/writes did not change
```
Meldiron, line 603: Is any usage stat affected by this? If so, let's have check that it assertNotEquals. That will give us asurance that usage dump cycle ran.

app/controllers/shared/api.php
```diff
@@ -703,7 +656,7 @@
     ->inject('queueForWebhooks')
     ->inject('queueForRealtime')
     ->inject('dbForProject')
-    ->action(function (App $utopia, Request $request, Response $response, Document $project, Document $user, Event $queueForEvents, Audit $queueForAudits, StatsUsage $queueForStatsUsage, Delete $queueForDeletes, EventDatabase $queueForDatabase, Build $queueForBuilds, Messaging $queueForMessaging, Func $queueForFunctions, Event $queueForWebhooks, Realtime $queueForRealtime, Database $dbForProject) use ($parseLabel) {
+    ->action(function (App $utopia, Request $request, Response $response, Document $project, Document $user, ?Key $apiKey, Event $queueForEvents, Audit $queueForAudits, StatsUsage $queueForStatsUsage, Delete $queueForDeletes, EventDatabase $queueForDatabase, Build $queueForBuilds, Messaging $queueForMessaging, Func $queueForFunctions, Event $queueForWebhooks, Realtime $queueForRealtime, Database $dbForProject) use ($parseLabel) {
```
Meldiron, line 659: Leftover probably, I cant see it being used inside

app/init.php
abnegate: I don't think it makes sense here, `null` is the absence of a key, which is different to the guest key. This way, later checks are just "is there a key" vs "is the key a guest key"

src/Appwrite/Auth/Key.php
abnegate: This is used for both key types, it gets merged with the standard key's scopes too

Meldiron (approved):

PR #9347, opened 2025-02-12
feat: batch create abuse logs

christyjacob4:
## What does this PR do?

In the audits worker we invoked `createDocument()` for each job and this tends to affect the database performance under higher loads. Here we do a small refactor to aggregate the audit logs in memory and batch create the logs in one request when it exceeds a threshold. The threshold can be either a certain no. of jobs ( 5_000 ) in this case or the elapsing of a time interval like 60s in this case

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Workers/Audits.php
+++ b/src/Appwrite/Platform/Workers/Audits.php
@@ -22,7 +22,9 @@ class Audits extends Action
     private const BATCH_SIZE_PRODUCTION = 5_000;
     private const BATCH_AGGREGATION_INTERVAL = 60; // in seconds
 
-    private static array $logs = [];
+    private int $lastTriggeredTime = 0;
+
+    private array $logs = [];
 
     private function getBatchSize(): int
     {
@@ -46,6 +48,8 @@ class Audits extends Action
             ->inject('message')
             ->inject('dbForProject')
             ->callback(fn ($message, $dbForProject) => $this->action($message, $dbForProject));
+
+        $this->lastTriggeredTime = time();
     }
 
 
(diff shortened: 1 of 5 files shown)


src/Appwrite/Platform/Workers/Audits.php
```diff
@@ -63,23 +80,58 @@ public function action(Message $message, Database $dbForProject): void
...
+                    'userId' => $event['userId'],
+                    'event' => $event['event'],
+                    'resource' => $event['resource'],
+                    'userAgent' => $event['userAgent'],
+                    'ip' => $event['ip'],
+                    'location' => $event['location'],
+                    'data' => $event['data'],
+                    'timestamp' => $event['timestamp']
+                ];
+            }, self::$logs);
```
abnegate, line 127: Do we need array_map here?

src/Appwrite/Platform/Workers/Audits.php
```diff
@@ -63,23 +80,58 @@ public function action(Message $message, Database $dbForProject): void
...
+
+            $audit->logBatch($batchEvents);
+
+            // Clear the pending events after successful batch processing
+            self::$logs = [];
```
abnegate, line 132: If the log call fails, we won't clear the batch, is that expected?

src/Appwrite/Platform/Workers/Audits.php
christyjacob4: you're right. removed 👍

src/Appwrite/Platform/Workers/Audits.php
christyjacob4: Im not sure whats the best approach here. should we hold onto the jobs in case of failure and try them again in the next batch? 🤔  Or just clear them upon first sign of failure?

abnegate: I think it's better to clear on error here:
- We'll still have all the data in the failed queue
- Not retrying automatically is consistent with other workers
- We avoid potential for filling memory constantly in the case where we have consistent failures

PR #9348, opened 2025-02-12
Fix resend team invite if already accepted

ChiragAgg5k:
## What does this PR do?

Prevents resending of team invite if already accepted. It also fixes generation of a new `$secret` when resending the invite.

### Steps to reproduce:

1. Invite a team member to your organisation.
2. Accept the invite.
3. Send the invite again.

Expected:

- It should not send an invite again since the user already accepted it.

What's happening:

- Its sending the invite again. This is likely due to the recent fix that allows resending of invite but we should only do that when invitation is pending, if it's not then we should throw an error.

## Test Plan

Screenshot 2025-02-12 at 3 23 09 PM

## Checklist

### DIFF
--- a/src/Appwrite/Extend/Exception.php
+++ b/src/Appwrite/Extend/Exception.php
@@ -114,7 +114,6 @@ class Exception extends \Exception
     public const TEAM_NOT_FOUND                    = 'team_not_found';
     public const TEAM_INVITE_NOT_FOUND             = 'team_invite_not_found';
     public const TEAM_INVALID_SECRET               = 'team_invalid_secret';
-    public const TEAM_MEMBERSHIP_ALREADY_EXISTS    = 'team_membership_already_exists';
     public const TEAM_MEMBERSHIP_MISMATCH          = 'team_membership_mismatch';
     public const TEAM_INVITE_MISMATCH              = 'team_invite_mismatch';
     public const TEAM_ALREADY_EXISTS               = 'team_already_exists';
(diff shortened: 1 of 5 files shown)


src/Appwrite/Extend/Exception.php
```diff
@@ -114,6 +114,7 @@ class Exception extends \Exception
     public const TEAM_NOT_FOUND                    = 'team_not_found';
     public const TEAM_INVITE_NOT_FOUND             = 'team_invite_not_found';
     public const TEAM_INVALID_SECRET               = 'team_invalid_secret';
+    public const TEAM_MEMBERSHIP_ALREADY_EXISTS    = 'team_membership_already_exists';
```
christyjacob4, line 117: lets move this to the Memberships section and call it `MEMBERSHIP_ALREADY_EXISTS`

ChiragAgg5k: oh right, and there is already an exception for it `MEMBERSHIP_ALREADY_CONFIRMED`

tests/e2e/Services/Teams/TeamsBaseServer.php
```diff
@@ -175,17 +175,10 @@ public function testCreateTeamMembership($data): array
...
-        //     'x-appwrite-project' => $this->getProject()['$id'],
-        // ], $this->getHeaders()), []);
-
-        // $this->assertEquals($userUid, $response['body']['$id']);
-        // $this->assertContains('team:'.$teamUid, $response['body']['roles']);
-        // $this->assertContains('team:'.$teamUid.'/admin', $response['body']['roles']);
-        // $this->assertContains('team:'.$teamUid.'/editor', $response['body']['roles']);
+        /**
+         * Test for FAILURE
+         */
```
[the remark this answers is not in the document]
ChiragAgg5k: i can add more tests if you want, but i believe these two together handles both the cases well

christyjacob4 (approved):

PR #9353, opened 2025-02-13
Feat messaging metrics

lohanidamodar:
## What does this PR do?

- Count targets (targets)
- Count targets by provider type (email.targets, sms.targets, push.targets)

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


src/Appwrite/Platform/Workers/StatsResources.php
```diff
@@ -129,9 +129,20 @@ protected function countForProject(Database $dbForPlatform, callable $getLogsDB,
...
+            $targets = $dbForProject->count('targets');
+            $emailTargets = $dbForProject->count('targets', [
+                Query::equal('providerType', [MESSAGE_TYPE_EMAIL])
+            ]);
+            $pushTargets = $dbForProject->count('targets', [
+                Query::equal('providerType', [MESSAGE_TYPE_PUSH])
+            ]);
+            $smsTargets = $dbForProject->count('targets', [
+                Query::equal('providerType', [MESSAGE_TYPE_SMS])
+            ]);
```
abnegate, line 145: Would be much quicker (and potentially more consistent) to use array_filter on the PHP side than to re-count for each type

lohanidamodar: @abnegate I'm not sure what you mean? We need to count for each type, get count by type, do we support groupBy? If not how will it work?

abnegate (approved):

abnegate (commented):
Let's add a check in the tests

lohanidamodar:
> Let's add a check in the tests

We don't have a test for these yet as we have not yet exposed logs DB via API. I'm tracking it in separate issue.

PR #9357, opened 2025-02-13
Feat: Authorized previews

Meldiron:
## What does this PR do?

Makes preview unaccessable to guests - only authorized Console users can visit previews

## Test Plan

- [x] Implement tests
- [x] Address todos

## Related PRs and Issues

x

## Checklist


app/controllers/general.php
```diff
@@ -173,6 +174,78 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
             $path .= '?' . $query;
         }
 
+        $protocol = $request->getProtocol();
+
+        // Preview authorization (configure)
+        if (\str_starts_with($path, '/_appwrite/authorize')) {
```
Meldiron, line 180: Move into endpoint

app/controllers/general.php
```diff
@@ -155,6 +156,70 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
             $path .= '?' . $query;
         }
 
+        $protocol = $request->getProtocol();
+
+        // Ensure preview authorization
+        $requirePreview = \is_null($apiKey) || !$apiKey->isPreviewAuthDisabled();
+        if ($isPreview && $requirePreview) {
```
eldadfux, line 163: Since this is a rather long sequence, let's add a docblock to describe the flow.

app/controllers/general.php
```diff
@@ -155,6 +156,70 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
...
+            $cookie = $request->getCookie(Auth::$cookieNamePreview, '');
+            $authorized = false;
+
+            // Security checks to mark authorized true
+            if (!empty($cookie)) {
+                $jwt = new JWT(System::getEnv('_APP_OPENSSL_KEY_V1'), 'HS256', 3600, 0);
```
eldadfux, line 169: Since this is going to be part of the edge, with no access to the database, can we instead redirect this to another Appwrite endpoint that will check the DB, issue the token, and redirect back so we can set it as a local cookie?

app/controllers/general.php
```diff
@@ -155,6 +156,70 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
...
+
+                $payload = [];
+                try {
+                    $payload = $jwt->decode($cookie);
+                } catch (JWTException $error) {
+                    // Authorized remains false
```
eldadfux, line 175: Shouldn't we throw here an exception?

app/controllers/general.php
```diff
@@ -155,6 +156,70 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
...
+                }
+
+                $userExists = false;
+                $userId = $payload['userId'] ?? '';
+                if (!empty($userId)) {
+                    $user = Authorization::skip(fn () => $dbForPlatform->getDocument('users', $userId));
```
eldadfux, line 181: I would add another check to confirm current user is same as the user in JWT payload. For extra safety.

eldadfux, line 181: if this will happen on an Appwrite domain, we also don't need to do this skip, which will make the validation even stronger. I would also add a check to ensure this is possible only if project is console to minimize the attack surface.

app/controllers/general.php
```diff
@@ -155,6 +156,70 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
...
+                }
+
+                $sessionExists = false;
+                $jwtSessionId = $payload['sessionId'] ?? '';
+                if (!empty($jwtSessionId) && !empty($user->find('$id', $jwtSessionId, 'sessions'))) {
+                    $sessionExists = true;
+                }
+
+                $membershipExists = false;
+                $project = Authorization::skip(fn () => $dbForPlatform->getDocument('projects', $projectId));
```
eldadfux, line 194: I would prevent the auth skip for project, if the user is part of the team, they will have access to this.

eldadfux (changes requested):

app/controllers/general.php
Meldiron: Agree, let's address later when edge is finished

app/controllers/general.php
Meldiron: No need, `$payload` got default value (empty array), so it will simply end up with `$authorized=false`.

If we throw, we show error page. But instead we want logic for when $authorized is false, which is redirecting to Console to get new JWT

app/controllers/general.php
Meldiron: This user is taken from userId, which is form JWT payload. It's always the same.

if by _current_ you mean getResource('user'), that will be empty, because we don't have project ID header, we don't have user's session cookie, and so on.

Meldiron: We need to skip - we dont have project ID, user's session cookie, and similar. This is considered guest request by Appwrite APIs

app/controllers/general.php
Meldiron: Same as above - we need to skip. I tripple-checked, it is handled securely.

eldadfux (approved):

PR #9366, opened 2025-02-15
Feat: Site screenshots

Meldiron:
## What does this PR do?

Makes screenshot for site builds

## Test Plan

- [x] Add tests

## Related PRs and Issues

x

## Checklist


app/config/collections/projects.php
```diff
@@ -1528,7 +1528,18 @@
                 'default' => false,
                 'array' => false,
                 'filters' => [],
-            ]
+            ],
+            [
+                '$id' => ID::custom('screenshot'), // File ID from 'screenshots' Console bucket
```
Meldiron, line 1533: One more for dark theme

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
```diff
@@ -700,6 +709,85 @@ protected function buildDeployment(Device $deviceForFunctions, Func $queueForFun
...
+                        Query::equal("resourceInternalId", [$deployment->getInternalId()])
+                    ]));
+
+                    if($rule->isEmpty()) {
+                        throw new \Exception("Rule for build not found");
+                    }
+
+
+                    $client = new FetchClient();
+                    $client->addHeader('Authorization', 'Bearer ' . App::getEnv('_APP_OPENSSL_KEY_V1', ''));
```
eldadfux, line 726: I would avoid using this key again, we over use it. I think we can do same as we do for the assistant container.

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
```diff
@@ -700,6 +709,85 @@ protected function buildDeployment(Device $deviceForFunctions, Func $queueForFun
...
+                    if(!$success) {
+                        throw new \Exception("Screenshot failed to save");
+                    }
+
+                    $teamId = $project->getAttribute('teamId', '');
+                    $file = new Document([
+                        '$id' => $fileId,
+                        '$permissions' => [
+                            Permission::read(Role::team(ID::custom($teamId))),
+                            Permission::update(Role::team(ID::custom($teamId), 'owner')),
```
eldadfux, line 756: Not sure if write permissions are needed here.

app/config/collections/projects.php
Meldiron, line 1533: update deletes worker to remove screnshot files with build/deployments/sites/proejcts/organizations

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
Meldiron: Changed to use new env var, `_APP_BROWSER_SECRET`

src/Appwrite/Utopia/Response/Model/Deployment.php
```diff
@@ -76,6 +76,12 @@ public function __construct()
                 'default' => false,
                 'example' => true,
             ])
+            ->addRule('screenshot', [
```
Meldiron, line 79: Add tests

app/init.php
```diff
@@ -1911,8 +1911,9 @@ function getDevice(string $root, string $connection = ''): Device
 );
 
 App::setResource('previewHostname', function (Request $request) {
+    // TODO: @Meldiron Allow in production too for internal communication (authorized with secret)
```
Meldiron, line 1914: Address todo

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
Meldiron: Removed secret all-together

docker-compose.yml
```diff
@@ -880,6 +881,14 @@ services:
     environment:
       - _APP_ASSISTANT_OPENAI_API_KEY
 
+  appwrite-browser:
+    container_name: appwrite-browser
+    image: appwrite/browser:0.1.0
```
Meldiron, line 886: Release

app/http.php
```diff
@@ -302,6 +303,60 @@ function createDatabase(App $app, string $resourceKey, string $dbName, array $co
 
                 $dbForPlatform->createCollection('bucket_' . $bucket->getInternalId(), $attributes, $indexes);
             }
+
+            if ($dbForPlatform->getDocument('buckets', 'screenshots')->isEmpty() &&
+                !$dbForPlatform->exists($dbForPlatform->getDatabase(), 'bucket_2')) {
+                Console::info("    └── Creating screenshots bucket...");
```
vermakhushboo, line 309: Why does it have space in the start?

app/http.php
```diff
@@ -302,6 +303,60 @@ function createDatabase(App $app, string $resourceKey, string $dbName, array $co
 
                 $dbForPlatform->createCollection('bucket_' . $bucket->getInternalId(), $attributes, $indexes);
             }
+
+            if ($dbForPlatform->getDocument('buckets', 'screenshots')->isEmpty() &&
+                !$dbForPlatform->exists($dbForPlatform->getDatabase(), 'bucket_2')) {
```
vermakhushboo, line 308: Why is `bucket_2` for?

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
```diff
@@ -713,6 +720,107 @@ protected function buildDeployment(Device $deviceForFunctions, Func $queueForFun
...
+
+                    $client = new FetchClient();
+                    $client->addHeader('content-type', FetchClient::CONTENT_TYPE_APPLICATION_JSON);
+
+                    $bucket = $dbForPlatform->getDocument('buckets', 'screenshots');
+
+                    $configs = [
+                        'screenshot' => [
+                            'headers' => [ 'x-appwrite-hostname' => $rule->getAttribute('domain') ],
+                            'url' => 'http://traefik/?appwrite-preview=1&appwrite-theme=light',
```
vermakhushboo, line 743: Would it remain as traefik? Should we use an env var instead of hardcoding?

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
```diff
@@ -713,6 +720,107 @@ protected function buildDeployment(Device $deviceForFunctions, Func $queueForFun
...
+                            'color' => 'light'
```
vermakhushboo, line 744: Should it be called `theme` here as well?

src/Appwrite/Platform/Workers/Deletes.php
```diff
@@ -749,7 +750,7 @@ private function deleteAuditLogs(Document $project, callable $getProjectDB, stri
      * @return void
      * @throws Exception
      */
-    private function deleteSite(Database $dbForPlatform, callable $getProjectDB, Device $deviceForSites, Device $deviceForFunctions, Device $deviceForBuilds, Document $document, CertificatesAdapter $certificates, Document $project): void
+    private function deleteSite(Database $dbForPlatform, callable $getProjectDB, Device $deviceForSites, Device $deviceForFunctions, Device $deviceForBuilds, Device $deviceForFiles, Document $document, CertificatesAdapter $certificates, Document $project): void
```
vermakhushboo, line 753: Do we also need to do this when a single deployment is deleted?

tests/e2e/Services/Sites/SitesCustomServerTest.php
```diff
@@ -1430,10 +1412,7 @@ public function testSitePreviewBranding(): void
         $proxyClient = new Client();
         $proxyClient->setEndpoint('http://' . $previewDomain);
 
-        $response = $proxyClient->call(Client::METHOD_GET, '/', array_merge([
-            'content-type' => 'application/json',
```
vermakhushboo, line 1434: Why did we remove the headers from tests?

app/controllers/general.php
```diff
@@ -585,7 +585,7 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
 */
 
 App::init()
-    ->groups(['database', 'functions', 'storage', 'messaging'])
```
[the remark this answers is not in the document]
Meldiron: It's list of services that cant be using using `console` as project, and we now need to use Storage on Console with it's `screenshots` bucket

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
Meldiron: It's internal complxity, it should never be changed by instance of server on which its running, so I would say it's fine to hard-code. It's a hostname that appwirte-browser can use to talk to Appwrite API. Those names wont change unless we decide to remove Traefik, which would be pretty rare. Tests would also fail so indication would be good enough to update it here

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
Meldiron: Hmm, it's called `color` on Appwrite Browser because internally it sets `colorScheme`, and to simplify, just `color`.

But I agree with you, in contect of Browser API params, `theme` would make more sense.

src/Appwrite/Platform/Workers/Deletes.php
Meldiron: Yes, nice catch

tests/e2e/Services/Sites/SitesCustomServerTest.php
Meldiron: They were not nessessary, visits from browser on site domains dont usually have those, so best to test with same scenarios

app/http.php
eldadfux, line 308: Can we do something better here?

app/http.php
```diff
@@ -302,6 +303,60 @@ function createDatabase(App $app, string $resourceKey, string $dbName, array $co
...
+                Console::info("    └── Creating screenshots bucket...");
+                $dbForPlatform->createDocument('buckets', new Document([
+                    '$id' => ID::custom('screenshots'),
+                    '$collection' => ID::custom('buckets'),
+                    'name' => 'Screenshots',
+                    'maximumFileSize' => (int) System::getEnv('_APP_STORAGE_LIMIT', 0), // 10MB
```
eldadfux, line 314: Not sure I would use the server default here, can cause weird side effects if too small

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
eldadfux: Would it behave the same on k8s as on swarm/compose?

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
```diff
@@ -713,6 +720,107 @@ protected function buildDeployment(Device $deviceForFunctions, Func $queueForFun
...
+                        throw new \Exception("Rule for build not found");
+                    }
+
+                    $client = new FetchClient();
+                    $client->addHeader('content-type', FetchClient::CONTENT_TYPE_APPLICATION_JSON);
+
+                    $bucket = $dbForPlatform->getDocument('buckets', 'screenshots');
+
+                    $configs = [
+                        'screenshot' => [
```
eldadfux, line 741:
```suggestion
                        'screenshotLight' => [
```

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
```diff
@@ -713,6 +720,107 @@ protected function buildDeployment(Device $deviceForFunctions, Func $queueForFun
...
+                        if ($response->getStatusCode() >= 400) {
+                            throw new \Exception("Screenshot failed to generate: " . $response->getBody());
+                        }
+
+                        $screenshot = $response->getBody();
+
+                        $fileId = ID::unique();
+                        $fileName = $fileId . '.png';
+                        $path = $deviceForFiles->getPath($fileName);
+                        $path = str_ireplace($deviceForFiles->getRoot(), $deviceForFiles->getRoot() . DIRECTORY_SEPARATOR . $bucket->getId(), $path); // Add bucket id to path after root
```
eldadfux, line 779: Why do we have to do this? it feels like this should be simpler, or abstracted somehow, no?

app/http.php
Meldiron: I do same as lines above.

In my opinion first check is enough, so if you are good with it, I can remove second part of if from both here, and default bucket above

src/Appwrite/Platform/Modules/Compute/Base.php
```diff
@@ -172,14 +172,10 @@ public function redeployVcsSite(Request $request, Document $site, Document $proj
             'activate' => $activate,
         ]));
 
-        // Preview deployments for sites
-        $projectId = $project->getId();
-
         $sitesDomain = System::getEnv('_APP_DOMAIN_SITES', '');
-        $domain = "{$deploymentId}-{$projectId}.{$sitesDomain}";
+        $domain = ID::unique() . "." . $sitesDomain;
```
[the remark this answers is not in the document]
Meldiron: It's something we discussed with Christy, and it was small so I addressed it in this PR. We dont want to do multi-ID domains, we want fully random for preview domains too. Another PR will introduce per-branch and per-commti domains, which will have some predistability based on details from GitHub.

app/http.php
Meldiron: Will hard-code to 5MB, should be more than enough.

Meldiron:
> Would it behave the same on k8s as on swarm/compose?

Hmm maybe not, will make it into env variable.

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
Meldiron: We do this in storage endpoint for createFile. I just copied it. We can improve it, agree, but let's do it in different scope from Sites.

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
Meldiron: I checked what migration is doing, its doing `appwrite`. So switching from `traefik` to `appwrite`, and keeping it hard-coded.

src/Appwrite/Platform/Modules/Compute/Base.php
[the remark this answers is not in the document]
eldadfux: This still doesn't make sense to me, but not a blocker for this specific change. Would love to to discuss further, I also don't see any strong need to use separated domains for Sites and Functions, both should be able to use `.run` and `.network` each for its own purpose (regional vs global serving).

app/http.php
eldadfux: I referring to the use of the ID this way.

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
eldadfux: See if we can have something in storage backlog, and we can give it some love later. I would be very specific about the task as it will be hard to recall later what we wanted to achieve here.

app/http.php
Meldiron: Me too. To allow us to merge, removing this intenralID check from both existing scenario, and my new one

src/Appwrite/Platform/Modules/Compute/Base.php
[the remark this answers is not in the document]
Meldiron: Noted down, I see missalignment between us and Christy too; will open up to discussion in near future

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
Meldiron: Noted down

PR #9387, opened 2025-02-20
chore: make min/max params optional for attribute update

ChiragAgg5k:
## What does this PR do?

Even though min/max parameters for attributes integer and float are optional, when updating them they were not. This PR fixes this and ensures they are in default range when null.

We will need a subsequent SDK change after the merge.

## Related PRs and Issues

- https://github.com/appwrite/sdk-for-dart/issues/48

## Checklist


app/controllers/api/databases.php
```diff
@@ -2339,14 +2339,19 @@ function updateAttribute(
...
+    ->param('max', null, new Integer(), 'Maximum value to enforce on new documents', true)
     ->param('default', null, new Nullable(new Integer()), 'Default value for attribute when not provided. Cannot be set when attribute is required.')
     ->param('newKey', null, new Key(), 'New attribute key.', true)
     ->inject('response')
     ->inject('dbForProject')
     ->inject('queueForEvents')
     ->action(function (string $databaseId, string $collectionId, string $key, ?bool $required, ?int $min, ?int $max, ?int $default, ?string $newKey, Response $response, Database $dbForProject, Event $queueForEvents) {
+
+        // Ensure attribute default is within range
+        $min = \is_null($min) ? PHP_INT_MIN : $min;
```
christyjacob4, line 2352:
```suggestion
        $min ??= PHP_INT_MIN;
```

app/controllers/api/databases.php
```diff
@@ -2398,14 +2403,19 @@ function updateAttribute(
...
+    ->param('max', null, new FloatValidator(), 'Maximum value to enforce on new documents', true)
     ->param('default', null, new Nullable(new FloatValidator()), 'Default value for attribute when not provided. Cannot be set when attribute is required.')
     ->param('newKey', null, new Key(), 'New attribute key.', true)
     ->inject('response')
     ->inject('dbForProject')
     ->inject('queueForEvents')
     ->action(function (string $databaseId, string $collectionId, string $key, ?bool $required, ?float $min, ?float $max, ?float $default, ?string $newKey, Response $response, Database $dbForProject, Event $queueForEvents) {
+
+        // Ensure attribute default is within range
+        $min = \is_null($min) ? -PHP_FLOAT_MAX : $min;
```
christyjacob4, line 2416:
```suggestion
        $min ??= PHP_INT_MIN;
```

app/controllers/api/databases.php
```diff
@@ -2339,14 +2339,19 @@ function updateAttribute(
...
     ->param('default', null, new Nullable(new Integer()), 'Default value for attribute when not provided. Cannot be set when attribute is required.')
     ->param('newKey', null, new Key(), 'New attribute key.', true)
     ->inject('response')
     ->inject('dbForProject')
     ->inject('queueForEvents')
     ->action(function (string $databaseId, string $collectionId, string $key, ?bool $required, ?int $min, ?int $max, ?int $default, ?string $newKey, Response $response, Database $dbForProject, Event $queueForEvents) {
+
+        // Ensure attribute default is within range
+        $min = \is_null($min) ? PHP_INT_MIN : $min;
+        $max = \is_null($max) ? PHP_INT_MAX : $max;
```
christyjacob4, line 2353:
```suggestion
        $max ??= PHP_INT_MAX;
```

app/controllers/api/databases.php
christyjacob4: is PHP_INT_MIN a negative number ?

app/controllers/api/databases.php
christyjacob4: is there no float_min? 
if not I think the minimum value would be `-PHP_FLOAT_MAX + 1` 🤔

app/controllers/api/databases.php
ChiragAgg5k: yep seems so - 

Screenshot 2025-02-26 at 8 26 22 PM

loks0n (approved):

app/controllers/api/databases.php
```diff
@@ -2339,8 +2342,8 @@ function updateAttribute(
     ->param('collectionId', '', new UID(), 'Collection ID. You can create a new collection using the Database service [server integration](https://appwrite.io/docs/server/databases#databasesCreateCollection).')
     ->param('key', '', new Key(), 'Attribute Key.')
     ->param('required', null, new Boolean(), 'Is attribute required?')
-    ->param('min', null, new Integer(), 'Minimum value to enforce on new documents')
-    ->param('max', null, new Integer(), 'Maximum value to enforce on new documents')
+    ->param('min', null, new Integer(), 'Minimum value to enforce on new documents', true)
+    ->param('max', null, new Integer(), 'Maximum value to enforce on new documents', true)
```
abnegate, line 2346: Why only make min and max optional but keep required and default as required?

ChiragAgg5k: true should update them as well!

ChiragAgg5k:
@abnegate since "default" attribute can also be null as a valid value, we cant make it optional as that would make it hard to differentiate overriding it in update endpoints, something like this:

image

so we should skip making "default" optional for now

Hardanish-Singh (approved):

app/controllers/api/databases.php
```diff
@@ -2089,7 +2093,7 @@ function updateAttribute(
     ->param('databaseId', '', new UID(), 'Database ID.')
     ->param('collectionId', '', new UID(), 'Collection ID. You can create a new collection using the Database service [server integration](https://appwrite.io/docs/server/databases#databasesCreateCollection).')
     ->param('key', '', new Key(), 'Attribute Key.')
-    ->param('required', null, new Boolean(), 'Is attribute required?')
+    ->param('required', null, new Boolean(), 'Is attribute required?', true)
```
abnegate, line 2096: Let's keep required as required if not changing default

ChiragAgg5k: reverted 👍

abnegate (approved):

PR #9391, opened 2025-02-20
chore: fix null errors on team invite

ChiragAgg5k:
## What does this PR do?

An invitee's name may be `null` on rare cases, leading to errors. The PR fixes this by replacing the value with `$name` which can either be the optional name param set by user or the invitee's email, or if the invitee's account exist then the account's name (that too if not empty).

Priority if all 3 exist:

```
Invitee's account name > Optional Name set by inviter > Invitee's email
```

## Checklist


ChiragAgg5k:
@stnguyen90 Updated the PR, now `$name` is only used in the "email body".

so the email will start with "Hi user@gmail.com" if both account name and optional name params don't exist.

christyjacob4 (approved):

PR #9408, opened 2025-02-25
Feat dynamic regions

eldadfux:
## What does this PR do?

Prep PR to allow Cloud to sort regions by proximity to user coordinates. For this to work properly we need city level coordinates and geo location of each user. On country level some mistakes are possible when countries are closer to each other and city X can be closer to country Y than to the coordinates of the center of its own country. 

## Test Plan

Relying on existing tests here, and new unit tests on the cloud repo. 

## Related PRs and Issues

- (Related PR or issue)

## Checklist

Yes!

N/A


app/config/locale/continents.php
```diff
@@ -1,11 +1,46 @@
 <?php
 
+/**
```
Meldiron, line 3: Might be nice to also mention for what purposes coordinates are, I am guessing to recommend best region.

Meldiron (approved):

app/config/locale/continents.php
Meldiron, line 1: Contitents and countries are source for responses in locale APIs, I believe.

By default they wont be exposed, which is good, but do we want to consider to add coordinates to response models too?

PR #9414, opened 2025-02-27
chore: bump sdk

ChiragAgg5k:
## What does this PR do?

Bumps SDKs for Flutter, Dart, Appwrite, Swift, Node, Web and CLI.

## Checklist


app/config/platforms.php
```diff
@@ -59,7 +59,7 @@
             [
                 'key' => 'flutter',
                 'name' => 'Flutter',
-                'version' => '13.1.1',
+                'version' => '14.0.1',
```
stnguyen90, line 62: if we're changing the type of the AppwriteExcpetion response, this should be a major bump

app/config/platforms.php
```diff
@@ -371,7 +371,7 @@
             [
                 'key' => 'dart',
                 'name' => 'Dart',
-                'version' => '12.2.0',
+                'version' => '13.0.1',
```
stnguyen90, line 374: major bump if we're changing the type

app/config/platforms.php
ChiragAgg5k: this release does not contain that change yet

app/config/platforms.php
ChiragAgg5k: this release does not contain that change yet

stnguyen90 (approved):

loks0n (approved):

christyjacob4 (approved):

PR #9433, opened 2025-03-04
Truncate commit message to 252 characters

vermakhushboo:
## What does this PR do?

If commit message is longer than 255 characters, truncate it to `252` characters and add leading zeroes

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/vcs.php
```diff
@@ -224,7 +224,7 @@
                 'providerCommitHash' => $providerCommitHash,
                 'providerCommitAuthorUrl' => $providerCommitAuthorUrl,
                 'providerCommitAuthor' => $providerCommitAuthor,
-                'providerCommitMessage' => $providerCommitMessage,
+                'providerCommitMessage' => mb_strimwidth($providerCommitMessage, 0, 255, '...'),
```
Meldiron, line 227: Let's double check, but I think we should not need this here.

Meldiron (approved):

PR #9435, opened 2025-03-04
feat: allow exp and appwrite-callback-${projectId} scheme URIs in auth redirects

loks0n:
**Problem:** 
React Native OAuth 2 authentication flows rely on custom schemes to redirect the user from the mobile browser (where they authenticate with an OAuth2 provider), back to the app. We don't currently accept URIs with schemes other than `http` and `https`.

**Current implementation:** 
We have two validators: 
- `Origin` validator is used in the init hook for all API requests. It's designed to take a list of platforms from the database. 
- `Host` is the validator used for specific endpoint parameters. It takes a simple allowlist of hostnames (which is derived from the platforms collection)

**Suggested solution:**
Add `exp://*` and `appwrite-callback-${projectId}://*` to the allowed patterns within the `Host` validator. We probably need to extend it, maybe it could be called Redirect.

Note: `parse_url` php builtin function returns `false` when the hostname value is empty, but a scheme exists:
```php
parse_url('appwrite-callback-my-project://') === false
```


app/init.php
```diff
@@ -1206,57 +1206,43 @@ function (mixed $value) {
 App::setResource('queueForMigrations', function (Queue\Publisher $publisher) {
     return new Migration($publisher);
 }, ['publisher']);
-App::setResource('clients', function ($request, $console, $project) {
-    $console->setAttribute('platforms', [ // Always allow current host
-        '$collection' => ID::custom('platforms'),
-        'name' => 'Current Host',
-        'type' => Origin::CLIENT_TYPE_WEB,
-        'hostname' => $request->getHostname(),
-    ], Document::SET_TYPE_APPEND);
```
loks0n, line 1215: I've intentionally removed this. It's a historical piece of code that allows any `Host` as an Origin for console.

We now have `_APP_CONSOLE_HOSTNAMES` and don't want to allow custom domains to serve cloud.

src/Appwrite/Network/Validator/Redirect.php
```diff
@@ -0,0 +1,102 @@
...
+
+        // `parse_url` couldn't handle the URL, try extracting scheme with regex
+        if (preg_match('/^([a-zA-Z][a-zA-Z0-9+.-]*):\/\//', $value, $matches)) {
+            $scheme = $matches[1];
+            if (\in_array($scheme, $this->schemes)) {
+                return true;
+            }
+        }
+
+        return false;
```
loks0n, line 76: `parse_url` fails when the scheme is present but no hostname. This is a common pattern with custom schemes on mobile.

For example:
```php
parse_url('appwrite-callback-my-project://') === false
```

tests/unit/Network/Validators/RedirectTest.php
```diff
@@ -0,0 +1,103 @@
...
+        $this->assertEquals(true, $validator->isValid('https://appwrite.io/callback'));
+        $this->assertEquals(true, $validator->isValid('https://localhost'));
+        $this->assertEquals(true, $validator->isValid('http://localhost:3000'));
+        $this->assertEquals(true, $validator->isValid('https://example.com/auth/callback?token=123'));
+
+        // Invalid hostnames
+        $this->assertEquals(false, $validator->isValid('https://unauthorized.com'));
+        $this->assertEquals(false, $validator->isValid('http://subdomain.appwrite.io')); // subdomain not in allowed list
+        $this->assertEquals(false, $validator->isValid('https://app-write.io')); // hyphenated variant not in allowed list
+        $this->assertEquals(false, $validator->isValid('ftp://appwrite.io')); // valid hostname but not http/https
```
loks0n, line 30: This may be an important case to consider. We currently **allow** any scheme if the hostname is a match. I've changed this behaviour to only allow if the hostname is a match **AND** the scheme is `http` or `https`.

Any thoughts?

tests/e2e/Scopes/ProjectCustom.php
```diff
@@ -52,6 +53,21 @@ public function getProject(bool $fresh = false): array
...
+            'cookie' => 'a_session_console=' . $this->getRoot()['session'],
+            'x-appwrite-project' => 'console',
+        ], [
+            'name' => 'Local development',
+            'hostname' => 'localhost',
+            'type' => Origin::CLIENT_TYPE_WEB,
+        ]);
+
+        $this->assertEquals(201, $platform['headers']['status-code']);
+        $this->assertNotEmpty($platform['body']);
```
loks0n, line 69: Another important consideration. `localhost` will need to be added manually.

app/init.php
stnguyen90: This can break things. For example, mobile apps (flutter, ios, android) call `createOAuth2Sesssion()` without any success or failure URLs and then and then Appwrite redirects them to the default [success](https://github.com/appwrite/console/blob/main/src/routes/(public)/auth/oauth2/success/%2Bpage.svelte) and [failure](https://github.com/appwrite/console/blob/main/src/routes/(public)/auth/oauth2/failure/%2Bpage.svelte) URLs. If they're using a custom domain, these pages won't load.

We prevented accessing the Console via custom domains before and had to revert because it broke people's apps. If we want to make this change how things work and give people notice.

stnguyen90 (changes requested):

app/init.php
loks0n: Ok - I've added `$request->getHostname()` back to the allow list. We can consider removing it seperately.

stnguyen90 (commented):
Do tests need to be updated?

src/Appwrite/Network/Validator/Origin.php
```diff
@@ -2,145 +2,98 @@
...
-    public const CLIENT_TYPE_FLUTTER_LINUX = 'flutter-linux';
-    public const CLIENT_TYPE_FLUTTER_WEB = 'flutter-web';
-    public const CLIENT_TYPE_APPLE_IOS = 'apple-ios';
-    public const CLIENT_TYPE_APPLE_MACOS = 'apple-macos';
-    public const CLIENT_TYPE_APPLE_WATCHOS = 'apple-watchos';
-    public const CLIENT_TYPE_APPLE_TVOS = 'apple-tvos';
-    public const CLIENT_TYPE_ANDROID = 'android';
-    public const CLIENT_TYPE_UNITY = 'unity';
-    public const CLIENT_TYPE_REACT_NATIVE_IOS = 'react-native-ios';
-    public const CLIENT_TYPE_REACT_NATIVE_ANDROID = 'react-native-android';
```
loks0n, line 25: These have been moved to Client.php, I don't really think our validators should be coupled to this DB whitelist.

src/Appwrite/Network/Validator/Origin.php
```diff
@@ -2,145 +2,98 @@
...
-    public function getDescription(): string
-    {
-        if (!\array_key_exists($this->client, $this->platforms)) {
-            return 'Unsupported platform';
+        if (!empty($this->schemes)) {
+            $messages[] = 'URL scheme must be one of: ' . \implode(', ', $this->schemes);
         }
 
-        return 'Invalid Origin. Register your new client (' . $this->host . ') as a new '
-            . $this->platforms[$this->client] . ' platform on your project console dashboard';
```
loks0n, line 112: As far I can tell, while this error looks good, I think it only worked for web hostnames?

tests/e2e/Services/Realtime/RealtimeBase.php
```diff
@@ -1,28 +1,18 @@
...
             $projectId = $this->getProject()['$id'];
         }
 
-        $headers = array_merge(
-            [
-                "Origin" => "appwrite.test",
-            ],
-            $headers
-        );
-
```
loks0n, line 25: This origin was never added as a platform. We should have been blocking this!

stnguyen90 (commented):
https://github.com/appwrite/appwrite/pull/9435/files#r1985426694

stnguyen90 (commented):
https://github.com/appwrite/appwrite/pull/9435/files#r1985426694

202420505:
@loks0n Why is this closed?

PR #9436, opened 2025-03-04
Chore: Rename deployment activate endpoint

Meldiron:
## What does this PR do?

Renames activation endpoint for better consistency with upcoming rename of deployment cancelation

## Test Plan

New tests implemented to ensure this

## Related PRs and Issues

x

## Checklist


composer.json
```diff
@@ -55,7 +55,7 @@
         "utopia-php/database": "0.59.0",
         "utopia-php/domains": "0.5.*",
         "utopia-php/dsn": "0.2.1",
-        "utopia-php/framework": "0.33.*",
+        "utopia-php/framework": "dev-fix-alias-params as 0.33.99",
```
Meldiron, line 58: Since framework lib needs extensive reviews, we can merge with this in-here.
We can update later when HTTP gets released. Itwill be blocker for merging feat-sites to main

vermakhushboo, line 58: Let's add a todo to change it later

src/Appwrite/Platform/Modules/Sites/Http/Deployments/Download/Get.php
```diff
@@ -31,7 +31,6 @@ public function __construct()
         $this
             ->setHttpMethod(Action::HTTP_REQUEST_METHOD_GET)
             ->setHttpPath('/v1/sites/:siteId/deployments/:deploymentId/download')
-            ->httpAlias('/v1/sites/:functionId/deployments/:deploymentId/build/download', [ 'type' => 'output' ])
```
vermakhushboo, line 34: This was added yesterday since the default value was `source`. Why did we remove it?

composer.json
Meldiron: Added linear task, framework needs extensive reviews so best not to be blocked by it's release

src/Appwrite/Platform/Modules/Sites/Http/Deployments/Download/Get.php
vermakhushboo: Ah, right

vermakhushboo (approved):

PR #9442, opened 2025-03-05
fix: model for image transformations in usage project

ChiragAgg5k:
## What does this PR do?

Makes `imageTransformations` an array instead of integer

## Checklist

### DIFF
--- a/src/Appwrite/Utopia/Response/Model/UsageProject.php
+++ b/src/Appwrite/Utopia/Response/Model/UsageProject.php
@@ -206,7 +206,7 @@ class UsageProject extends Model
             ])
             ->addRule('imageTransformationsTotal', [
                 'type' => self::TYPE_INTEGER,
-                'description' => 'An array of aggregated number of image transformations.',
+                'description' => 'Total aggregated number of image transformations.',
                 'default' => 0,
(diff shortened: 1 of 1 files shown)


src/Appwrite/Utopia/Response/Model/UsageProject.php
```diff
@@ -197,15 +197,16 @@ public function __construct()
...
-                'example' => 0,
+                'default' => [],
+                'example' => [],
+                'array' => true
             ])
-            ->addRule('imageTransformations', [
+            ->addRule('imageTransformationsTotal', [
                 'type' => self::TYPE_INTEGER,
-                'description' => 'Total aggregated number of image transformations.',
+                'description' => 'An array of aggregated number of image transformations.',
```
ItzNotABug, line 209: not array of aggregation, but a total.
```
Total aggregated number of image transformations.
```

christyjacob4, line 209: this is supposed to be an integer right?

ItzNotABug (changes requested):

ItzNotABug (approved):

PR #9462, opened 2025-03-07
chore: added logsdb for deletes worker

ChiragAgg5k:
## What does this PR do?

Adds logsDB to delete worker.

## Test Plan

before:
Screenshot 2025-03-07 at 7 18 07 PM

after:
Screenshot 2025-03-07 at 7 28 57 PM

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Workers/Deletes.php
+++ b/src/Appwrite/Platform/Workers/Deletes.php
@@ -47,7 +47,7 @@ class Deletes extends Action
             ->inject('project')
             ->inject('dbForPlatform')
             ->inject('getProjectDB')
-            ->inject('timelimit')
+            ->inject('getLogsDB')
             ->inject('deviceForFiles')
             ->inject('deviceForFunctions')
             ->inject('deviceForBuilds')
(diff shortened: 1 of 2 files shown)


src/Appwrite/Platform/Workers/Deletes.php
```diff
@@ -47,7 +47,7 @@ public function __construct()
             ->inject('project')
             ->inject('dbForPlatform')
             ->inject('getProjectDB')
-            ->inject('timelimit')
+            ->inject('getLogsDB')
```
christyjacob4, line 50: I believe luke added a new resource useLogsDB which uses pools more safely. Can we use that instead?

ChiragAgg5k: sure

ChiragAgg5k: thats only in cloud rn, wanna init it in CE as well?

src/Appwrite/Platform/Workers/Deletes.php
```diff
@@ -412,14 +412,22 @@ function (Document $document) use ($cache, $projectId) {
...
+        $dbForLogs = $getLogsDB($project);
+
+        // Delete Usage stats from projectDB
         $this->deleteByGroup('stats', [
             Query::lessThan('time', $hourlyUsageRetentionDatetime),
             Query::equal('period', ['1h']),
         ], $dbForProject);
+
+        // Delete Usage stats from logsDB
+        $this->deleteByGroup('stats', [
```
christyjacob4, line 427: okay this is usage stats. We also need to delete the audit logs ( activities ) on the new logs DB

christyjacob4 (approved):

PR #9477, opened 2025-03-09
Feat apps module

eldadfux:
## What does this PR do?

Adding OAuth2 support

## Test Plan

Using existing e2e, unit and new tests. 

## Related PRs and Issues

- N/A

## Checklist


src/Appwrite/Migration/Version/V23.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Appwrite\Migration\Version;
+
+use Appwrite\Migration\Migration;
+use Exception;
+use Throwable;
+
+class V23 extends Migration
```
Meldiron, line 9: We can remove for now. We can do this later when all is merged in 1.7.x

src/Appwrite/Migration/Migration.php
```diff
@@ -93,6 +93,7 @@ abstract class Migration
         '1.6.0' => 'V21',
         '1.6.1' => 'V21',
         '1.6.2' => 'V22',
+        '1.7.0' => 'V23',
```
Meldiron, line 96: We can remove for now. We can do this later when all is merged in 1.7.x

tests/e2e/Services/Account/AccountCustomServerTest.php
```diff
@@ -218,7 +218,7 @@ public function testCreateMagicUrl(): array
         $this->assertEquals($email, $lastEmail['to'][0]['address']);
         $this->assertEquals($this->getProject()['name'] . ' Login', $lastEmail['subject']);
 
-        $token = substr($lastEmail['text'], strpos($lastEmail['text'], '&secret=', 0) + 8, 64);
+        $token = substr($lastEmail['text'], strpos($lastEmail['text'], '&secret=', 0) + 8, 256);
```
Meldiron, line 221: Just want to give this attention; is this not breaking? Did token use to be shorter, and now is longer? Can current logic still work with shorter tokens?

tests/e2e/Scopes/ProjectCustom.php
```diff
@@ -149,6 +149,7 @@ public function getProject(bool $fresh = false): array
         $project = [
             '$id' => $project['body']['$id'],
             'name' => $project['body']['name'],
+            'teamId' => $team['body']['$id'],
```
Meldiron, line 152: Is this change needed, or leftover?

app/controllers/api/users.php
```diff
@@ -92,7 +102,18 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
         }
 
-        $password = (!empty($password)) ? ($hash === 'plaintext' ? Auth::passwordHash($password, $hash, $hashOptionsObject) : $password) : null;
+        $hashedPassword = null;
+
+        if (!empty($password)) {
+            if ($hash instanceof Plaintext) { // Password was never hashed, hash it with the default hash
+                $defaultHash = new ProofsPassword();
+                $hashedPassword = $defaultHash->hash($password);
+                $hash = $defaultHash->getHash();
```
Meldiron, line 111: This is probably okay for now, but if shutdown hook wanted to look at hash, he would not see those changes; might become problem for coroutines (original value in resource does not get overriden)

Lets add todo at least, so we can find this place easily in future

Meldiron: alternatively, methods like setAlgo, setOptions, ... could work

Or maybe like we have in events, functon to from(), reset()

src/Appwrite/Migration/Version/V23.php
eldadfux: Tests are breaking without it.

src/Appwrite/Migration/Migration.php
eldadfux: Tests are breaking without it.

tests/e2e/Services/Account/AccountCustomServerTest.php
eldadfux: Only for a few seconds during deployment. We can announce a maintenance window, impact should be minimal.

tests/e2e/Scopes/ProjectCustom.php
eldadfux: Pretty sure a test was failing.

lohanidamodar:
Closed in favor of
https://github.com/appwrite/appwrite/pull/10468

PR #9479, opened 2025-03-10
chore: queue console project for maintenance delete

ChiragAgg5k:
## What does this PR do?

Till now console project was not scheduled for deletes, specifically for DELETE_TYPE_MAINTENANCE as the loop would just list projects inside dbForPlatform, that won't contain console project.

This PR initializes console project for `CLI` and queues it for delete as well.

## Test Plan

Screenshot 2025-03-10 at 2 23 28 PM

## Checklist

### DIFF
--- a/app/cli.php
+++ b/app/cli.php
@@ -2,11 +2,13 @@
 
 require_once __DIR__ . '/init.php';
 
+use Appwrite\Auth\Auth;
 use Appwrite\Event\Certificate;
 use Appwrite\Event\Delete;
 use Appwrite\Event\Func;
 use Appwrite\Event\StatsResources;
 use Appwrite\Event\StatsUsage;
+use Appwrite\Network\Validator\Origin;
 use Appwrite\Platform\Appwrite;
 use Appwrite\Runtimes\Runtimes;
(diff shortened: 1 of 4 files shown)


app/cli.php
```diff
@@ -99,6 +102,48 @@
     return $dbForPlatform;
 }, ['pools', 'cache']);
 
+CLI::setResource('console', function () {
```
christyjacob4, line 105: Eldad suggested moving the console resource to an associative array inside https://github.com/appwrite/appwrite/tree/1.6.x/app/config/console.php and reusing that array across init.php ,cli.php , in workers and in other places in the future.

app/cli.php
```diff
@@ -99,6 +99,10 @@
...
+    return new Document(Config::getParam('console'));
```
christyjacob4, line 103: do the workers use this resource? Any change required in the workers.php file?

christyjacob4: And did we verify the deletion is actually happening?

ChiragAgg5k: nope workers don't use it, no changes required there. it was only initialized in `init.php`

Issue #9449, opened 2025-03-05
🚀 Enhancement: Filter membership by role


stnguyen90:
### 🔖 Enhancement description

Update:

1. [teams.listMemberships()](https://appwrite.io/docs/references/cloud/client-web/teams#listMemberships)
2. [users.listMemberships()](https://appwrite.io/docs/references/cloud/server-nodejs/users#listMemberships)

to support querying/filtering on roles.

### 🎤 Pitch

At the moment, in order to find a user who has a particular role, all the memberships need to be fetched and then filtered client-side. It would be helpful to be able to pass a query with the API call to filter the data server side so that I don't need to add more code client-side.

### 👀 Have you spent some time to check if this issue has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

PR #9503, opened 2025-03-13
feat: add query by role to memberships

ChiragAgg5k:
## What does this PR do?

Allows querying memberships by `roles` attribute and updates `users.listMemberships` to accept query and search params.

## Test Plan

Screenshot 2025-03-13 at 2 15 58 PM

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/9449

## Checklist


src/Appwrite/Migration/Version/V22.php
```diff
@@ -75,6 +75,14 @@ private function migrateCollections(): void
...
                     }
                     break;
+                case 'memberships':
+                    // Create roles index
+                    try {
+                        $this->createIndexFromCollection($this->projectDB, $id, '_key_roles');
+                    } catch (Throwable $th) {
+                        Console::warning("'_key_roles' from {$id}: {$th->getMessage()}");
+                    }
+                    break;
```
stnguyen90, line 85: we can't require a migration in 1.6.x. This may need to be in 1.7.x if we need to add a index.

stnguyen90 (changes requested):
we also need to update the method on the teams api

src/Appwrite/Migration/Version/V22.php
ChiragAgg5k: @stnguyen90 we already support queries in teams api, its just `roles` was previously not an allowed attribute to be queried.

no changes are required on that side. check my screenshot i have tested its working.

stnguyen90 (approved):

stnguyen90 (commented):
Can we get at least 1 test case added? Maybe for the `users.listMemberships()`

stnguyen90 (commented):
listMemberships

stnguyen90 (approved):

PR #9527, opened 2025-03-18
chore: remove runtime specifications

ChiragAgg5k:
## What does this PR do?

Remove runtime specifications and use the default values from env.

## Checklist


ChiragAgg5k:
here is how /specifications endpoint will look like:

Screenshot 2025-03-18 at 10 47 58 AM

app/controllers/api/functions.php
```diff
@@ -183,7 +183,13 @@
...
     ->param('specification', APP_FUNCTION_SPECIFICATION_DEFAULT, fn (array $plan) => new RuntimeSpecification(
         $plan,
-        Config::getParam('runtime-specifications', []),
+        Config::getParam('runtime-specifications', [
+            APP_FUNCTION_SPECIFICATION_DEFAULT => [
+                'slug ' => APP_FUNCTION_SPECIFICATION_DEFAULT,
+                'memory' => APP_FUNCTION_MEMORY_DEFAULT,
+                'cpus' => APP_FUNCTION_CPUS_DEFAULT
+            ]
+        ]),
```
Meldiron, line 192: I think we should fully remove specification from open source, and only relay on values in env variables (for cpus and memory).
That will also mean removing params for it, removing it from collections, response filters, ...

PR #9547, opened 2025-03-20
chore: fix console audits delete

ChiragAgg5k:
## What does this PR do?

Uses dbForPlatform for deleting console audits, and adds new retention for it. default is 6 months.

## Test Plan

Screenshot 2025-03-20 at 2 47 46 PM

## Checklist


app/worker.php
```diff
@@ -215,15 +215,19 @@
...
-    return time() - (int) System::getEnv('_APP_MAINTENANCE_RETENTION_ABUSE', 86400);
+    return time() - (int) System::getEnv('_APP_MAINTENANCE_RETENTION_ABUSE', 86400); // 1 day
 });
 
 Server::setResource('auditRetention', function () {
-    return DateTime::addSeconds(new \DateTime(), -1 * System::getEnv('_APP_MAINTENANCE_RETENTION_AUDIT', 1209600));
+    return DateTime::addSeconds(new \DateTime(), -1 * System::getEnv('_APP_MAINTENANCE_RETENTION_AUDIT', 1209600)); // 14 days
+});
+
+Server::setResource('auditRetentionConsole', function () {
```
abnegate, line 225: Instead of a new resource, from `auditRetention`, let's return a map with keys 'project' and 'console'

.env
```diff
@@ -85,7 +85,7 @@ _APP_MAINTENANCE_DELAY=
 _APP_MAINTENANCE_RETENTION_CACHE=2592000
 _APP_MAINTENANCE_RETENTION_EXECUTION=1209600
 _APP_MAINTENANCE_RETENTION_ABUSE=86400
-_APP_MAINTENANCE_RETENTION_AUDIT=1209600
+_APP_MAINTENANCE_RETENTION_AUDIT=project=1209600,console=15778800
```
abnegate, line 88: Let's still use separate env vars and just fetch them both in the `auditRetention` resource

docker-compose.yml
```diff
@@ -389,6 +390,8 @@ services:
       - _APP_DATABASE_SHARED_TABLES
       - _APP_DATABASE_SHARED_TABLES_V1
       - _APP_EMAIL_CERTIFICATES
+      - _APP_MAINTENANCE_RETENTION_AUDIT
+      - _APP_MAINTENANCE_RETENTION_AUDIT_CONSOLE
```
christyjacob4, line 394: we only need to add this to the maintenance task and deletes worker right?

abnegate (approved):

christyjacob4 (approved):

app/worker.php
```diff
@@ -215,15 +215,18 @@
...
+    return time() - (int) System::getEnv('_APP_MAINTENANCE_RETENTION_ABUSE', 86400); // 1 day
 });
 
 Server::setResource('auditRetention', function () {
-    return DateTime::addSeconds(new \DateTime(), -1 * System::getEnv('_APP_MAINTENANCE_RETENTION_AUDIT', 1209600));
+    return [
+        'project' => DateTime::addSeconds(new \DateTime(), -1 * System::getEnv('_APP_MAINTENANCE_RETENTION_AUDIT', 1209600)), // 14 days
+        'console' => DateTime::addSeconds(new \DateTime(), -1 * System::getEnv('_APP_MAINTENANCE_RETENTION_AUDIT_CONSOLE', 15778800)) // 6 months
+    ];
 });
```
stnguyen90, line 226: Isn't the typical way we do this is to depend on the project? So if the project is console, it'd be one value. if the project is not, it would be another value.

Also, this should probably depend on the org because the retention should really be different based on the plan.

PR #9561, opened 2025-03-22
chore: update gitpod.yml config

ChiragAgg5k:
## What does this PR do?

- removes prebuilds from gitpod config as they have been deprecated
Screenshot 2025-03-22 at 7 24 29 PM

- add `git config --global --add safe.directory /app` as a init step

## Checklist

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,6 +1,7 @@
 tasks:
   - name: Run Appwrite Docker Stack
     init: |
+      git config --global --add safe.directory /app
       git submodule update --init
       docker compose build
       docker compose pull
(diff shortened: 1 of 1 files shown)


.gitpod.yml
```diff
@@ -1,6 +1,7 @@
 tasks:
   - name: Run Appwrite Docker Stack
     init: |
+      git config --global --add safe.directory /app
```
stnguyen90, line 4: Why this?

ChiragAgg5k: its to fix an error always showed whenever i use gitpod

Screenshot 2025-04-04 at 9 37 32 AM

stnguyen90: Looks like this happens because the composer install runs as root, but the owner of the files is gitpod. Let's make the compose install run as gitpod instead of adding this global safe config:

```shell
      docker run --rm --interactive --tty \
      --user "$(id -u):$(id -g)" \
      --volume $PWD:/app \
      composer install \
      --ignore-platform-reqs \
      --optimize-autoloader \
      --no-plugins \
      --no-scripts \
      --prefer-dist
```

ChiragAgg5k: that results in this error:

Screenshot 2025-04-05 at 8 09 49 PM

stnguyen90 (approved):

PR #9574, opened 2025-03-25
chore: abstract token secret hiding to response model

ChiragAgg5k:
## What does this PR do?

This PR abstracts the secret hiding logic for the responses across various places by introduces a new attribute called `sensitive` to Response Model, which when set checks if user is appUser or priviledgedUser, and if not it sets the key as empty.

Attributes abstracted - 
1. session secret
2. token secret
3. function logs
4. function errors

## Test Plan

I tested it for createEmailToken endpoint and it works like it did before.

test code:
```js
const client = new Client()
    .setEndpoint(endpoint)  
    .setProject(projectId)
    .setKey(apiKey);

const token = await account.createEmailToken(ID.unique(), 'test6@test.com', true);
console.log(token);
```

without apikey:
Screenshot 2025-01-10 at 7 00 37 PM

with apikey:
Screenshot 2025-01-10 at 7 01 02 PM

## Checklist


src/Appwrite/Utopia/Response/Model.php
```diff
@@ -90,7 +90,8 @@ protected function addRule(string $key, array $options): self
             'required' => true,
             'array' => false,
             'description' => '',
-            'example' => ''
+            'example' => '',
+            'sensitive' => false
```
christyjacob4, line 94: lets check if there are any other response models needing this.

app/controllers/api/account.php
```diff
@@ -3407,7 +3355,7 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
         $queueForEvents
             ->setParam('userId', $profile->getId())
             ->setParam('tokenId', $recoveryDocument->getId())
-        ;
+            ->setPayload($response->output($recoveryDocument, Response::MODEL_TOKEN, true), sensitive: ['secret']);
```
christyjacob4, line 3358: why do we need to show the sensitive attribute here?

app/controllers/api/functions.php
```diff
@@ -2394,14 +2374,6 @@
             throw new Exception(Exception::EXECUTION_NOT_FOUND);
         }
 
-        $roles = Authorization::getRoles();
-        $isPrivilegedUser = Auth::isPrivilegedUser($roles);
-        $isAppUser = Auth::isAppUser($roles);
-        if (!$isPrivilegedUser && !$isAppUser) {
-            $execution->setAttribute('logs', '');
-            $execution->setAttribute('errors', '');
```
christyjacob4, line 2402: lets ensure all the affected endpoints are tested well

app/controllers/api/account.php
ChiragAgg5k: because here it's being queued for webhook event, and we need to show the response in webhooks.

earlier we didnt need to set it because we hid the secret after queuing it.

app/controllers/api/functions.php
ChiragAgg5k: sure

loks0n (approved):

christyjacob4 (approved):

PR #9590, opened 2025-03-28
Fix template to use button

ItzNotABug:
## What does this PR do?

Use buttons instead of a plain url text.

## Test Plan

Manual.

image

## Related PRs and Issues

N/A.

## Checklist


eldadfux (commented):
what was the motivation behind this one? we didn't do it originally to keep the email templates neutral and now they will all be pink.

sara-k-48:
> what was the motivation behind this one? we didn't do it originally to keep the email templates neutral and now they will all be pink.

@ItzNotABug let's use the neutral color?

app/config/locale/templates/email-inner-base.tpl
```diff
@@ -1,6 +1,6 @@
 <p>{{hello}}</p>
 <p>{{body}}</p>
-<p><a href="{{redirect}}" target="_blank">{{redirect}}</a></p>
+<p><a href="{{redirect}}" target="_blank" style="font-size: 14px; font-family: Inter; color: #ffffff; text-decoration: none; background-color: #2D2D31; border-radius: 8px; padding: 9px 14px; border: 1px solid #414146; display: inline-block; text-align:center; box-sizing: border-box;">{{buttonText}}</a></p>
```
stnguyen90: I have dark mode in my email client. How will this look there?

stnguyen90 (approved):

app/config/locale/templates/email-inner-base.tpl
ItzNotABug: shared the relevant screenshots. marking as resolved.

PR #9596, opened 2025-03-31
feat: sdk group attribute

ChiragAgg5k:
## What does this PR do?

Introduces a new attribute to the SDK method called the group that helps in grouping the attributes further within a namespace.

For eg. methods inside `account.php` can be further grouped into `Account` (update, create, delete etc.), `Auth` (create session, anonymous session etc.), `MFA` for all the mfa related methods and more.

## Related PRs and Issues

- https://github.com/appwrite/website/pull/1885

## Checklist


app/controllers/api/vcs.php
```diff
@@ -275,6 +275,7 @@
     ->label('error', __DIR__ . '/../../views/general/error.phtml')
     ->label('sdk', new Method(
         namespace: 'vcs',
+        group: 'auth',
```
stnguyen90, line 278: this should probably be grouped with installations

app/controllers/api/locale.php
```diff
@@ -17,6 +17,7 @@
     ->label('scope', 'locale.read')
     ->label('sdk', new Method(
         namespace: 'locale',
+        group: 'locale',
```
eldadfux, line 20: No need to split forcefully, the split should be based on actual resources as they're represented in Appwrite. Same for the avatar service. Locale can stay locale for all, or just no label as no split is needed, same for avatars

app/controllers/api/migrations.php
```diff
@@ -37,6 +37,7 @@
     ->label('audits.event', 'migration.create')
     ->label('sdk', new Method(
         namespace: 'migrations',
+        group: 'appwrite',
```
eldadfux, line 40: Same for migrations

app/controllers/api/project.php
```diff
@@ -26,6 +26,7 @@
     ->label('scope', 'projects.read')
     ->label('sdk', new Method(
         namespace: 'project',
+        group: 'usage',
```
eldadfux, line 29: This API is not public, let's skip it.

app/controllers/api/proxy.php
```diff
@@ -35,6 +35,7 @@
     ->label('audits.resource', 'rule/{response.$id}')
```
eldadfux, line 35: No need to split the proxy at this point,

app/controllers/api/storage.php
```diff
@@ -1396,9 +1407,21 @@
     ->groups(['api', 'storage'])
     ->label('scope', 'public')
     ->label('resourceType', RESOURCE_TYPE_BUCKETS)
-    ->label('sdk.response.code', Response::STATUS_CODE_OK)
-    ->label('sdk.response.type', '*/*')
-    ->label('sdk.methodType', 'location')
+    ->label('sdk', new Method(
```
eldadfux, line 1410: Pretty sure this method should be hidden or don't exists at all. Checking with Jake, please don't add it unless I mentioned otherwise.

app/controllers/api/storage.php
```diff
@@ -1772,6 +1797,7 @@
     ->label('resourceType', RESOURCE_TYPE_BUCKETS)
     ->label('sdk', new Method(
         namespace: 'storage',
+        group: 'usage',
```
eldadfux, line 1800: All usage endpoints should be hidden from the docs and public SDKs, let's skip them on all services.

app/controllers/api/teams.php
```diff
@@ -62,6 +62,7 @@
     ->label('audits.resource', 'team/{response.$id}')
```
eldadfux, line 62: This should be:

- Teams
- Memberships

app/controllers/api/users.php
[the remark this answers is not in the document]
eldadfux: @ChiragAgg5k auth should be sessions. Not sure if we want to group tokens as well, please ask Matej on Discord for feedback

eldadfux (changes requested):

app/controllers/api/storage.php
eldadfux: Yes, confirmed, this should stay hidden.

ChiragAgg5k: okkk i was wondering why it was still following old sdk labels, will remove it 👍

app/controllers/api/users.php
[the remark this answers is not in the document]
ChiragAgg5k: alright 👍🏻

app/controllers/api/vcs.php
ChiragAgg5k: yep, makes sense. done

app/controllers/api/storage.php
ChiragAgg5k: hidden 👍🏻

app/controllers/api/teams.php
ChiragAgg5k: done

app/controllers/api/proxy.php
ChiragAgg5k: done

app/controllers/api/project.php
ChiragAgg5k: removed

app/controllers/api/migrations.php
ChiragAgg5k: done

app/controllers/api/locale.php
ChiragAgg5k: made group to be nullable

app/controllers/api/account.php
```diff
@@ -513,6 +516,7 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
     ->label('scope', 'account')
     ->label('sdk', new Method(
         namespace: 'account',
+        group: 'auth',
```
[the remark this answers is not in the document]
ChiragAgg5k: done

app/controllers/api/databases.php
```diff
@@ -2919,6 +2955,7 @@ function updateAttribute(
     ->label('resourceType', RESOURCE_TYPE_DATABASES)
     ->label('sdk', new Method(
         namespace: 'databases',
+        group: 'indexes',
```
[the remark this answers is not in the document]
ChiragAgg5k: 🤔

christyjacob4 (approved):

PR #9608, opened 2025-04-03
Fix: image transformation cache

lohanidamodar:
## What does this PR do?

- Fix image transformation cache

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/controllers/shared/api.php
+++ b/app/controllers/shared/api.php
@@ -390,7 +390,8 @@ App::init()
     ->inject('timelimit')
     ->inject('mode')
     ->inject('apiKey')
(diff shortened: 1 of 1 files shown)


app/controllers/shared/api.php
```diff
@@ -573,8 +578,10 @@
                 $response
                     ->addHeader('Cache-Control', sprintf('private, max-age=%d', $timestamp))
                     ->addHeader('X-Appwrite-Cache', 'hit')
-                    ->setContentType($cacheLog->getAttribute('mimeType'))
-                    ->send($data);
+                    ->setContentType($cacheLog->getAttribute('mimeType'));
+                if (!$isImageTransformation || !$isDisabled) {
```
abnegate, line 582: What happens if $isImageTransformation and $isDisabled are both true, where does the response get sent?

lohanidamodar: it goes to the endpoint as cache doesn't send the response.

abnegate (approved):

PR #9620, opened 2025-04-07
feat: inforu adapter

ChiragAgg5k:
## What does this PR do?

Updates utopia/messaging to add the new Inforu SMS adapter.

## Checklist


src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -780,6 +785,10 @@ private function createProviderFromDSN(DSN $dsn): Document
                     'messageId' => $dsn->getParam('messageId'),
                     'useDLT' => $dsn->getParam('useDLT'),
                 ],
+                'inforu' => [
+                    'senderId' => $dsn->getParam('senderId', $from),
```
christyjacob4, line 789:
```suggestion
                    'senderId' => $user,
```
we can just pass the senderID as the user in the DSN right? 

`sms://senderId:password@inforu`

ChiragAgg5k: yeah, updated

PR #9622, opened 2025-04-08
Feat: CSV Import

ItzNotABug:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/migrations.php
```diff
@@ -290,6 +300,136 @@
...
+        responses: [
+            new SDKResponse(
+                code: Response::STATUS_CODE_ACCEPTED,
+                model: Response::MODEL_MIGRATION,
+            )
+        ]
+    ))
+    ->param('bucketId', '', new UID(), 'Storage bucket unique ID. You can create a new storage bucket using the Storage service [server integration](https://appwrite.io/docs/server/storage#createBucket).')
+    ->param('fileId', '', new UID(), 'File ID.')
+    ->param('resourceId', null, new Text(75), 'Composite ID in the format {databaseId:collectionId}, identifying a collection within a database.')
```
abnegate, line 323: We have a `CompoundUID` validator for this case

src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -222,8 +237,23 @@ protected function processMigration(Document $migration, Realtime $queueForRealt
         $projectDocument = $this->dbForPlatform->getDocument('projects', $project->getId());
         $tempAPIKey = $this->generateAPIKey($projectDocument);
 
+        $importDocument = null;
         $transfer = $source = $destination = null;
 
+        if ($migration->getAttribute('source') === Csv::getName()) {
+            $fileSize = $migration->getAttribute('credentials', [])['size'] ?? 0;
```
abnegate, line 244: Let's not pass this in credentials

app/controllers/api/migrations.php
```diff
@@ -290,6 +300,136 @@
...
+                $file->getAttribute('openSSLCipher'),
+                System::getEnv('_APP_OPENSSL_KEY_V' . $file->getAttribute('openSSLVersion')),
+                0,
+                \hex2bin($file->getAttribute('openSSLIV')),
+                \hex2bin($file->getAttribute('openSSLTag'))
+            );
+        }
+
+        // decompress
+        switch ($file->getAttribute('algorithm', Compression::NONE)) {
```
abnegate, line 385: We can stipulate that CSV files need to be unencrypted and uncompressed, to avoid needing to read the whole file into memory here. cc @eldadfux

app/controllers/api/migrations.php
```diff
@@ -290,6 +300,136 @@
...
+
+        $migration = $dbForProject->createDocument('migrations', new Document([
+            '$id' => $migrationId,
+            'status' => 'pending',
+            'stage' => 'init',
+            'source' => Csv::getName(),
+            'destination' => Appwrite::class::getName(),
+            'resources' => $resources,
+            'resourceId' => $resourceId,
+            'resourceType' => Resource::TYPE_DATABASE,
```
abnegate, line 411: I think this needs be collection, since it can only be one per file

app/controllers/api/migrations.php
```diff
@@ -290,6 +300,136 @@
...
+            'statusCounters' => [],
+            'resourceData' => [],
+            'errors' => [],
+            'credentials' => [
+                'path' => $path,
+                'size' => $fileSize,
+            ],
```
abnegate, line 418: Do we have `options` that we can use for this instead?

app/controllers/api/migrations.php
```diff
@@ -290,6 +300,136 @@
...
+                $compressor = new GZIP();
+                $source = $compressor->decompress($source);
+                break;
+        }
+
+        // copy to temporary folder
+        $migrationId = ID::unique();
+        $path = $deviceForCsvImports->getRoot() . '/' . $migrationId . '_' . $fileId . '.csv';
+        $deviceForCsvImports->write($path, $source, 'text/csv');
+        $fileSize = $deviceForCsvImports->getFileSize($path);
```
abnegate, line 400: There's a helper method like Device::transfer you can use to move files between `Devices`, we can use it here once not reading into memory

src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -222,8 +237,23 @@ protected function processMigration(Document $migration, Realtime $queueForRealt
...
+            $importDocument = new Document([
+                '$id' => ID::unique(),
+                'size' => $fileSize, // uncompressed and decrypted file size
+                'startedAt' => DateTime::now(),
+                'migrationId' => $migration->getId(),
+                'migrationInternalId' => $migration->getInternalId(),
+                'resourceId' => $migration->getAttribute('resourceId', ''),
+                'resourceType' => $migration->getAttribute('resourceType', ''),
+                'errors' => [],
+            ]);
```
abnegate, line 254: Maybe we don't need this collection at all :thinking: if we have `options` on migrations, we can add size/path etc there, and we can differentiate what will be in options based on source/destination combo

app/controllers/api/migrations.php
```diff
@@ -290,6 +300,136 @@
...
+        $isAPIKey = Auth::isAppUser(Authorization::getRoles());
+        $isPrivilegedUser = Auth::isPrivilegedUser(Authorization::getRoles());
+
+        // Check if migration/import is already in progress!
+        // if for some reason the worker crashes, the stage will always be `init`, what do we do?
+        $isInProgress = Authorization::skip(function () use ($dbForProject, $resourceId) {
+            $exists = $dbForProject->findOne(
+                'migrations',
+                [
+                    Query::notEqual('stage', 'finished'),
```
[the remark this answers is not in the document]
ItzNotABug: removed in-order to allow concurrent imports.

app/controllers/api/migrations.php
ItzNotABug: Its tricky because it requires the bucket to have those settings changed by the user before uploading the file. thats why this block checks what the encryption and decompression status is on a given file. I agree loading while file in memory should be avoided but then it requires a change on a bucket and then upload a file.

app/controllers/api/migrations.php
ItzNotABug: I think it'd throw an error - `Resource type must be one of bucket, database, function, user, team`. Only the top layer of the services is expected here.

app/controllers/api/migrations.php
ItzNotABug: We don't have `options` for something similar to `credentials`. I'd avoid adding a new set of attribute but fine if it makes more sense. Not sure if we'd need a migration patch though 🤔.

app/controllers/api/migrations.php
ItzNotABug: Answered above. This is needed due to decrypted/uncompressed file after a fetch from bucket.

src/Appwrite/Platform/Workers/Migrations.php
ItzNotABug: Makes sense to me. But then its an `option` attribute set only for `csv` led migration [atleast atm]. wdyt?

app/controllers/api/migrations.php
eldadfux: Compression shouldn't be an issues for us. Appwrite limit compressions to files up to size X to avoid loading huge files to memory even if the bucket has encryption enabled. Please verify the exact details with @lohanidamodar.

Until today, we avoided handling file uploads in other endpoints as much as possible to avoid increasing the complexity surface they introduce and lack of proper ways to extends endpoints.

Today, we have better ways to extend endpoints using our class based approach, this will require refactoring the createFile endpoint and carefully abstracting some of its complexities to allow us to have some base class for routes that require this.

That said, this might be a bigger take, and considering encryption, or compressions shouldn't be a massive problem, it may not be needed right now.

ItzNotABug:
TODO: Add `migration`.

app/controllers/api/migrations.php
abnegate: We can avoid duplicating the logic altogether here, I think CSV import flows best as 2 parts: upload, then import. From console the operation can still be one action, but make 2 API calls. Since we can handle without any changes to storage, we'll avoid the module refactor in this scope

Dockerfile
```diff
@@ -44,12 +44,14 @@ COPY ./dev /usr/src/code/dev
...
 # Set Volumes
 RUN mkdir -p /storage/uploads && \
+    mkdir -p /storage/csv-imports && \
     mkdir -p /storage/cache && \
     mkdir -p /storage/config && \
     mkdir -p /storage/certificates && \
     mkdir -p /storage/functions && \
     mkdir -p /storage/debug && \
     chown -Rf www-data.www-data /storage/uploads && chmod -Rf 0755 /storage/uploads && \
+    chown -Rf www-data.www-data /storage/csv-imports && chmod -Rf 0755 /storage/csv-imports && \
```
abnegate, line 54: Let's name this simply `storage/imports`

app/controllers/api/migrations.php
```diff
@@ -290,6 +297,98 @@
...
+                'size' => $fileSize,
+            ],
+        ]));
+
+        $queueForEvents->setParam('migrationId', $migration->getId());
+
+        $queueForMigrations
+            ->setMigration($migration)
+            ->setProject($project)
+            ->trigger();
```
abnegate, line 385: Do we need the trigger? Is the migrations worker not auto-triggered in the shutdown hooks?

app/controllers/api/migrations.php
```diff
@@ -290,6 +297,98 @@
...
+
+        // copy to temporary folder
+        $migrationId = ID::unique();
+        $newPath = $deviceForCsvImports->getPath('/' . $migrationId . '_' . $fileId . '.csv');
+        if (!$deviceForFiles->transfer($path, $newPath, $deviceForCsvImports)) {
+            throw new \Exception("Unable to copy file");
+        }
+
+        $fileSize = $deviceForCsvImports->getFileSize($path);
+        $resources = Transfer::extractServices([Transfer::GROUP_DATABASES]);
```
abnegate, line 360: I think this should only be `[Resource::TYPE_DOCUMENT]`

app/controllers/api/migrations.php
```diff
@@ -290,6 +297,98 @@
...
+
+        $migration = $dbForProject->createDocument('migrations', new Document([
+            '$id' => $migrationId,
+            'status' => 'pending',
+            'stage' => 'init',
+            'source' => Csv::getName(),
+            'destination' => Appwrite::class::getName(),
```
abnegate, line 367: I don't think we need ::class here

app/controllers/api/migrations.php
ItzNotABug: No, I don't think so. We `trigger` each time.

app/controllers/api/migrations.php
ItzNotABug: IDE auto-completion 😅.

app/controllers/api/migrations.php
ItzNotABug: No, we need a `root resource`.
See [here](https://github.com/utopia-php/migration/blob/main/src/Migration/Transfer.php#L235) and [here](https://github.com/utopia-php/migration/blob/main/src/Migration/Transfer.php#L316C5-L332C6). We seem to be revolving more around `ROOT_*` on the public front of the library.

app/config/collections/projects.php
```diff
@@ -1848,7 +1859,29 @@
                 'default' => null,
                 'array' => false,
                 'filters' => [],
-            ]
+            ],
+            [
+                '$id' => ID::custom('resourceId'),
```
eldadfux, line 1864: What resourceId and resourceType are we referring to here? How come is this a 1:1 relation if migration can in theory migrate endless resources?

app/init/constants.php
```diff
@@ -49,6 +49,7 @@
 const APP_STORAGE_FUNCTIONS = '/storage/functions';
 const APP_STORAGE_BUILDS = '/storage/builds';
 const APP_STORAGE_CACHE = '/storage/cache';
+const APP_STORAGE_CSV_IMPORTS = '/storage/imports'; // Temporary storage for csv imports
```
eldadfux, line 52: Can we simply call this imports? We don't really need to couple this with CSV, we can reuse it in the future for other formats.

ItzNotABug: We can.
I was just worried that out of a random look, it shouldn't look unrelated to CSV. But I'll change and keep the comment anyways.

app/config/collections/projects.php
ItzNotABug: these already exists on current migrations. The only new is the `options` attribute.
1. `resourceID` for csv is a `CompoundID` of `databaseId:collectionId`,
2. `resourceType` is Database Group afaik.

eldadfux: @abnegate we never use this pattern, let's stick with what we have. I think Matej used a parent resource approach, or we can use a reference approach with having only resource.

abnegate: @eldadfux I can see future issues with the `parentResource` pattern for things like documents/attributes/indexes, which have a 3 level hierarchy: database -> collection -> document, if we use a single attribute but compound as many IDs as we need, it's very flexible. We use it a couple of places in messaging, for referencing bucket/file in a single param. What do you think?

eldadfux: I understand, but IIRC we do have a pattern for it already if we want to go with the compound approach, I just want us to be consistent. Please verify how it works with @Meldiron

Meldiron: @eldadfux We ended up avoiding "parent" in rules.

What we did, is that rule has:

- `deploymentId` - only set when `type` is `deployment`
- `deploymentResourceType` - can be `function` or `site`
- `deploymentResourceId` - ID of site/function to which deployment belongs

Meldiron: After discussion with Darshan, importantto mention this was just addition moved from Cloud. Those already exist on cloud; we just added them to OSS now for this feature

tests/e2e/Services/Migrations/MigrationsBase.php
```diff
@@ -896,4 +898,228 @@ public function testAppwriteMigrationFunction(): void
...
+                Query::limit(150)->toString()
+            ]
+        ]);
+
+        $this->assertEquals(200, $documents['headers']['status-code']);
+        $this->assertIsArray($documents['body']['documents']);
+        $this->assertIsNumeric($documents['body']['total']);
+        $this->assertEquals($documentsCountInCSV, $documents['body']['total']);
+    }
+
```
abnegate, line 1116: Let's check more failure cases: 
- Invalid attribute in header row
- Missing required attribute in header row
- Missing attribute in document row
- Invalid attribute in document row

The first two should throw a 400 from the API. The other two should fail the job in the worker

ItzNotABug: All need to come from the Worker.
Added the tests with respect to these changes - https://github.com/utopia-php/migration/pull/75!

app/controllers/api/migrations.php
abnegate: Let's update 👌

docker-compose.yml
```diff
@@ -672,6 +673,8 @@ services:
       - ./app:/usr/src/code/app
       - ./src:/usr/src/code/src
       - ./tests:/usr/src/code/tests
+      # for csv import access
+      - appwrite-imports:/storage/imports:rw
```
abnegate, line 677: Let's move this to the top of the list and remove the comment

app/controllers/api/migrations.php
ItzNotABug: I think I did before, came back somehow. Done 👍

abnegate (approved):

PR #9643, opened 2025-04-14
Update database

abnegate:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Workers/Databases.php
+++ b/src/Appwrite/Platform/Workers/Databases.php
@@ -563,22 +563,19 @@ class Databases extends Action
         $start = \microtime(true);
 
         try {
-            $documents = $database->deleteDocuments($collectionId, $queries);
+            $count = $database->deleteDocuments(
+                $collectionId,
+                $queries,
+                Database::DELETE_BATCH_SIZE,
+                $callback
+            );
         } catch (\Throwable $th) {
             $tenant = $database->getSharedTables() ? 'Tenant:'.$database->getTenant() : '';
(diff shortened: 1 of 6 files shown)


src/Appwrite/Platform/Workers/Databases.php
```diff
@@ -563,22 +563,19 @@ protected function deleteByGroup(string $collectionId, array $queries, Database
...
-        if (\is_callable($callback)) {
-            foreach ($documents as $document) {
-                $callback($document);
-            }
-        }
-
         $end = \microtime(true);
-        $count = \count($documents);
-
         Console::info("Deleted {$count} documents by group in " . ($end - $start) . " seconds");
```
lohanidamodar, line 579: as we have return inside catch block we never reach this block if catch is called, so this can safely be moved inside try block?

src/Appwrite/Platform/Workers/Deletes.php
```diff
@@ -1050,24 +1050,20 @@ protected function deleteByGroup(
...
-        if (\is_callable($callback)) {
-            foreach ($documents as $document) {
-                $callback($document);
-            }
-        }
-
         $end = \microtime(true);
-        $count = \count($documents);
-
         Console::info("Deleted {$count} documents by group in " . ($end - $start) . " seconds");
```
lohanidamodar, line 1067: same as above?

src/Appwrite/Platform/Workers/Databases.php
abnegate: I think it's better to keep the try/catch scope as small as possible - the timing/logging code isn't exceptional

lohanidamodar (approved):

PR #9648, opened 2025-04-14
Add configurable function and build size

vermakhushboo:
## What does this PR do?

Read function size limit value from the plan

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/controllers/api/functions.php
+++ b/app/controllers/api/functions.php
@@ -1312,14 +1312,14 @@ App::post('/v1/functions/:functionId/deployments')
             throw new Exception(Exception::STORAGE_FILE_EMPTY, 'No file sent');
         }
 
-        $functionFileSize = (System::getEnv('_APP_FUNCTIONS_SIZE_LIMIT') ?: 30000000);
+        $functionSizeLimit = System::getEnv('_APP_FUNCTIONS_SIZE_LIMIT', '30000000');
 
         if (isset($plan['functionSize'])) {
-            $functionFileSize = $plan['functionSize'] * 1000 * 1000;
+            $functionSizeLimit = $plan['functionSize'] * 1000 * 1000;
         }
 
         $fileExt = new FileExt([FileExt::TYPE_GZIP]);
-        $fileSizeValidator = new FileSize($functionFileSize);
+        $fileSizeValidator = new FileSize($functionSizeLimit);
         $upload = new Upload();
 
         // Make sure we handle a single file and multiple files the same way
--- a/app/controllers/api/functions.php
+++ b/app/controllers/api/functions.php
@@ -1312,7 +1312,12 @@ App::post('/v1/functions/:functionId/deployments')
             throw new Exception(Exception::STORAGE_FILE_EMPTY, 'No file sent');
         }
 
-        $functionFileSize = isset($plan['functionSize']) ? $plan['functionSize'] * 1000 * 1000 : (System::getEnv('_APP_FUNCTIONS_SIZE_LIMIT') ?: 30000000);
+        $functionFileSize = (System::getEnv('_APP_FUNCTIONS_SIZE_LIMIT') ?: 30000000);
+
+        if (isset($plan['functionSize'])) {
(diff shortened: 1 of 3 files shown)


src/Appwrite/Platform/Workers/Builds.php
```diff
@@ -401,7 +404,11 @@ protected function buildDeployment(Device $deviceForFunctions, Webhook $queueFor
                 }
 
                 $directorySize = $localDevice->getDirectorySize($tmpDirectory);
-                $functionsSizeLimit = (int)System::getEnv('_APP_FUNCTIONS_SIZE_LIMIT', '30000000');
+
+                $functionsSizeLimit = empty($plan['functionSize'])
+                    ? (int) System::getEnv('_APP_FUNCTIONS_SIZE_LIMIT', 30000000)
+                    : (int) $plan['functionSize'] * 1000 * 1000;
+
```
Meldiron, line 411: Same as before. Also notice here you did `empty`, but above `isset`. Let's stay consistent - same implementation in both places.

app/controllers/api/functions.php
```diff
@@ -1311,8 +1312,9 @@
             throw new Exception(Exception::STORAGE_FILE_EMPTY, 'No file sent');
         }
 
+        $functionFileSize = isset($plan['functionSize']) ? $plan['functionSize'] * 1000 * 1000 : (System::getEnv('_APP_FUNCTIONS_SIZE_LIMIT') ?: 30000000);
```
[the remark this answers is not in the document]
Meldiron: Let's revisit this, (int) cast seems to be missing in new code.
Also lets use 2nd paraam in getEnv for default value, it's what we typically do, so lets stay consistent.

src/Appwrite/Platform/Workers/Builds.php
```diff
@@ -401,7 +404,13 @@ protected function buildDeployment(Device $deviceForFunctions, Webhook $queueFor
                 }
 
                 $directorySize = $localDevice->getDirectorySize($tmpDirectory);
-                $functionsSizeLimit = (int)System::getEnv('_APP_FUNCTIONS_SIZE_LIMIT', '30000000');
+
+                $functionsSizeLimit = (int) (System::getEnv('_APP_FUNCTIONS_SIZE_LIMIT') ?: 30000000);
```
Meldiron, line 408: A small detail, one is called functionFileSize, the other functionsSizeLimit. For simplicity we might just do `maxSize` or `sizeLimit`. but same in both

src/Appwrite/Platform/Workers/Builds.php
vermakhushboo: `int` casting was inconsistent in the existing code as well as in one case, we use a validator

app/controllers/api/functions.php
```diff
@@ -1311,8 +1312,14 @@
             throw new Exception(Exception::STORAGE_FILE_EMPTY, 'No file sent');
         }
 
+        $functionSizeLimit = System::getEnv('_APP_FUNCTIONS_SIZE_LIMIT', '30000000');
```
Meldiron, line 1315:
```suggestion
        $functionSizeLimit = (int) System::getEnv('_APP_FUNCTIONS_SIZE_LIMIT', '30000000');
```

src/Appwrite/Platform/Workers/Builds.php
```diff
@@ -620,8 +629,11 @@ protected function buildDeployment(Device $deviceForFunctions, Webhook $queueFor
             $durationEnd = \microtime(true);
 
             $buildSizeLimit = (int)System::getEnv('_APP_FUNCTIONS_BUILD_SIZE_LIMIT', '2000000000');
+            if (isset($plan['buildSize'])) {
+                $buildSizeLimit = $plan['buildSize'] * 1000 * 1000 * 1000;
```
Meldiron, line 633: Let's keep it as MB too,

```suggestion
                $buildSizeLimit = $plan['buildSize'] * 1000 * 1000;
```

Meldiron (approved):

PR #9650, opened 2025-04-14
feat: exp scheme

loks0n:
To facilitate React Native OAuth development, where App callback links are `exp://`, I've added `exp` and `appwrite-callback-${projectId}` origins by default.

- A new client type `Platform::TYPE_SCHEME` has been added.
- The `Origin` validator now permits any hostname if there is a `Platform::TYPE_SCHEME` match.
- A new `Redirect` validator extends `Origin` validator, simply overriding the error mesage.
- Realtime tests were broken, fixed.


app/init/resources.php
```diff
@@ -133,39 +133,32 @@
...
+            'type' => Platform::TYPE_SCHEME,
+            'name' => 'Expo',
+            'key' => 'exp',
+        ], Document::SET_TYPE_APPEND);
+        $project->setAttribute('platforms', [
+            '$collection' => ID::custom('platforms'),
+            'type' => Platform::TYPE_SCHEME,
+            'name' => 'Appwrite Callback',
+            'key' => 'appwrite-callback-' . $project->getId(),
+        ], Document::SET_TYPE_APPEND);
```
loks0n, line 155: Here is where we add `exp` and `appwrite-callback-projectId` schemes to the platform config.

stnguyen90 (approved):

rentplay:
Hi Folks, 

This is causing iOS appwrite auth to fail. please merge in asap!

.github/workflows/tests.yml
```diff
@@ -283,8 +283,6 @@ jobs:
     name: E2E Service Test (Dev Keys)
     runs-on: ubuntu-latest
     needs: setup
-    strategy:
-      fail-fast: false
```
loks0n, line 287: Is this intentional?

src/Appwrite/Network/Validator/Redirect.php
```diff
@@ -0,0 +1,25 @@
...
+    {
+        $platform = $this->scheme ? Platform::getNameByScheme($this->scheme) : null;
+        $host = $this->host ? '(' . $this->host . ')' : '';
+
+        if (empty($this->host) && empty($this->scheme)) {
+            return 'Invalid URI.';
+        }
+
+        return 'Invalid URI. Register your new client ' . $host . ' as a new '
+            . $platform . ' platform on your project console dashboard';
```
loks0n, line 23: I think we need to handle when platform is null here

src/Appwrite/Network/Platform.php
```diff
@@ -0,0 +1,117 @@
...
+    public static function getSchemes(array $platforms): array
+    {
+        $schemes = [];
+        foreach ($platforms as $platform) {
+            $type = $platform['type'] ?? self::TYPE_UNKNOWN;
+            $scheme = $platform['key'] ?? '';
+
+            switch ($type) {
+                case self::TYPE_SCHEME:
+                    $schemes[] = $scheme;
```
loks0n: One small thing, should we do strtolower here, just to be defensive?

loks0n: I guess we could also consider validating the scheme to ensure it doesn't have random chars?

.github/workflows/tests.yml
ChiragAgg5k: Yes this was, according to my IDE its not required

src/Appwrite/Network/Validator/Redirect.php
ChiragAgg5k: done, now `getNameByScheme` will return empty string and get caught by the empty check

src/Appwrite/Network/Platform.php
ChiragAgg5k: added both 👍

TorstenDittmann:
Bildschirmfoto 2025-07-04 um 16 09 29

PR #9657, opened 2025-04-16
Fix/throwing 400 for null order attributes 

ArnabChatterjee20k:
## What does this PR do?

Updated the database release and started using the OrderException in find methods where queries are coming from users and throwing error along with the reason of why cursor is giving error

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/tests/e2e/Services/Databases/DatabasesBase.php
+++ b/tests/e2e/Services/Databases/DatabasesBase.php
@@ -1641,20 +1641,7 @@ trait DatabasesBase
         $this->assertEquals(2019, $documents['body']['documents'][0]['releaseYear']);
         $this->assertCount(3, $documents['body']['documents']);
 
-
-
-        $response = $this->client->call(Client::METHOD_GET, '/databases/' . $databaseId . '/collections/' . $data['moviesId'] . '/attributes', array_merge([
-            'content-type' => 'application/json',
-            'x-appwrite-project' => $this->getProject()['$id'],
-            'x-appwrite-key' => $this->getProject()['apiKey'],
-        ]), [
-            'queries' => [
-                Query::equal('type', ['string'])->toString(),
-                Query::cursorAfter(new Document(['$id' => 'title']))->toString()
-            ],
-        ]);
-
-
+        // changing description attribute to be null by default instead of empty string
         $patchNull = $this->client->call(Client::METHOD_PATCH, '/databases/' . $databaseId . '/collections/' . $data['moviesId'] . '/attributes/string/description', array_merge([
             'content-type' => 'application/json',
             'x-appwrite-project' => $this->getProject()['$id'],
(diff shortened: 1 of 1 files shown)


app/controllers/api/databases.php
```diff
@@ -596,10 +597,15 @@ function updateAttribute(
...
-        ]), Response::MODEL_DATABASE_LIST);
+        try {
+            $response->dynamic(new Document([
+                'databases' => $dbForProject->find('databases', $queries),
+                'total' => $dbForProject->count('databases', $filterQueries, APP_LIMIT_COUNT),
+            ]), Response::MODEL_DATABASE_LIST);
+        } catch (OrderException $e) {
+            $message = "The order attribute '{$e->getAttribute()}' had a null value. Cursor pagination requires all documents order attribute values are non-null.";
+            throw new Exception(Exception::DATABASE_QUERY_ORDER_NULL, $message);
+        }
```
abnegate, line 608: Let's move the response->dynamic call after the try/catch so only the find/count calls are inside, for all changes

app/controllers/api/databases.php
```diff
@@ -596,10 +597,15 @@ function updateAttribute(
...
-            'total' => $dbForProject->count('databases', $filterQueries, APP_LIMIT_COUNT),
-        ]), Response::MODEL_DATABASE_LIST);
+        try {
+            $response->dynamic(new Document([
+                'databases' => $dbForProject->find('databases', $queries),
+                'total' => $dbForProject->count('databases', $filterQueries, APP_LIMIT_COUNT),
+            ]), Response::MODEL_DATABASE_LIST);
+        } catch (OrderException $e) {
+            $message = "The order attribute '{$e->getAttribute()}' had a null value. Cursor pagination requires all documents order attribute values are non-null.";
+            throw new Exception(Exception::DATABASE_QUERY_ORDER_NULL, $message);
```
abnegate, line 607: Let's just pass the message directly to the constructor, no need for the variable since we don't need it anywhere else, for all changes

tests/e2e/Services/Databases/DatabasesBase.php
```diff
@@ -1641,9 +1641,63 @@ public function testListDocuments(array $data): array
...
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()), [
+            'queries' => [
+                Query::orderAsc('dummy')->toString(),
+                Query::cursorAfter(new Document(['$id' => $document1['body']['$id']]))->toString()
+            ],
+        ]);
+
+        $this->assertEquals(400, $documentsPaginated['headers']['status-code']);
```
abnegate, line 1692: Let's add a comment here about why this is 400

abnegate (approved):

PR #9678, opened 2025-04-23
Fix: FileTokens

ItzNotABug:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced token-based access control for storage files, including tracking of last access time and improved permission handling.
  - Added support for new resource types in tokens, such as files, sites, functions, and databases.
  - Implemented new constants for resource token access duration and resource types.

- **Bug Fixes**
  - Improved validation and handling of token expiry, including support for infinite expiry.

- **Tests**
  - Added comprehensive end-to-end tests for token-based access to storage files, including preview, view, and download endpoints.

- **Chores**
  - Updated internal module structure and naming for improved maintainability and clarity.
  - Extended automated test coverage to include the new Tokens service.


src/Appwrite/Platform/Modules/Storage/Http/Tokens/Buckets/Files/XList.php
lohanidamodar, line 1: Did we decide on the module here? since it's `/v1/tokens` I believe it should be `Tokens` module not the Storage module as tokens though currently we have endpoint for storage, tokens itself should be separate module?

app/controllers/api/storage.php
lohanidamodar, line 1: Now that I think, if we add to view/preview then yes download also makes sense.

lohanidamodar (changes requested):

lohanidamodar (approved):
Looks good.

app/controllers/api/storage.php
```diff
@@ -1293,10 +1300,11 @@
             throw new Exception(Exception::STORAGE_BUCKET_NOT_FOUND);
         }
 
+        $isToken = !$resourceToken->isEmpty() && $resourceToken->getAttribute('bucketInternalId') === $bucket->getInternalId();
```
eldadfux, line 1303: The PR looks good, but I do have some doubts regarding the maintenance scalability of this security check. It feels very complex and very easy to get wrong, also, if we want to support more resources it will be very hard to confidently replicate. 

I wonder if we can abstract this in a higher level like DB authorization check or one of the init hooks as a fallback option.

eldadfux (changes requested):

app/controllers/api/storage.php
ItzNotABug: @eldadfux 
We could do this check beforehand around here - https://github.com/appwrite/appwrite/blob/fix-missing-filter/app/init/resources.php#L932C9-L953C6

That way it throws an error before we even hit the controller. WDYT?

ItzNotABug: @eldadfux I think the checks needs to be present per endpoint, a helper method like logic could help. but I don't think using a hook would work because even if the token is valid, we still need to verify with an additional check like we do now to not throw the exceptions.

ItzNotABug: cc @Meldiron @eldadfux as per internal conversations, planned for later.

PR #9680, opened 2025-04-23
feat: allow non-critical events to ignore exceptions when enqueuing the event

basert:


src/Appwrite/Messaging/Adapter/Realtime.php
```diff
@@ -129,7 +138,7 @@ public function hasSubscriber(string $projectId, string $role, string $channel =
      * @param array $options
      * @return void
      */
-    public static function send(string $projectId, array $payload, array $events, array $channels, array $roles, array $options = []): void
+    public function send(string $projectId, array $payload, array $events, array $channels, array $roles, array $options = []): void
```
christyjacob4, line 141: we have static usages of this function in a couple of places. 
https://github.com/search?q=repo%3Aappwrite%2Fappwrite%20%3A%3ASend(&type=code

So not sure this will work as expected

src/Appwrite/Event/StatsResources.php
```diff
@@ -6,6 +6,8 @@
 
 class StatsResources extends Event
 {
+    protected bool $critical = false;
```
christyjacob4, line 9: I would also make the audits event non critical

christyjacob4 (approved):

PR #9687, opened 2025-04-24
feat(dispatch): add contextual dispatch logic

basert:
Swoole `dispatch_func` gets called multiple times during a request lifecycle:
* on connection establish, with the initial HTTP request data (up to first 8k bytes of request)
* on connection close, without request data

The underlying TCP connection is shared between all Swoole processes and it's not guaranteed that the `dispatch_func` is called on the same process all the time.
The previous implementation returned a random workerId when there was no data as part of the request. This happens during connection close phase. Debugging the trace here showed the `dispatch_func` getting invoked over and over again. This caused really high CPU load.

It seems Swoole expects the same workerId during connection close that was initially assigned during the connection establish phase. This means we need to bind the assigned workerId to the current `$fd` and use `DISPATCH_UIDMOD`.

### DIFF
--- a/app/http.php
+++ b/app/http.php
@@ -43,29 +43,6 @@ $http = new Server(
 $payloadSize = 12 * (1024 * 1024); // 12MB - adding slight buffer for headers and other data that might be sent with the payload - update later with valid testing
 $totalWorkers = intval(System::getEnv('_APP_CPU_NUM', swoole_cpu_num())) * intval(System::getEnv('_APP_WORKER_PER_CORE', 6));
 
-$http
-    ->set([
-        'worker_num' => $totalWorkers,
-        'dispatch_func' => 'dispatch',
-        'open_http2_protocol' => true,
-        'http_compression' => false,
-        'package_max_length' => $payloadSize,
-        'buffer_output_size' => $payloadSize,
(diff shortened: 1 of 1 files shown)


app/http.php
```diff
@@ -43,10 +43,40 @@
...
+        $risky = false;
+        if (str_starts_with($request, 'POST') && str_contains($request, '/executions')) {
+            $risky = true;
+        } elseif (str_ends_with($domain, System::getEnv('_APP_DOMAIN_FUNCTIONS'))) {
+            $risky = true;
+        } elseif ($this->domains->get(md5($domain), 'value') === 1) {
+            // executions request coming from custom domain
+            $risky = true;
+        }
+        return $risky;
```
TorstenDittmann, line 72:
```suggestion
        if (str_starts_with($request, 'POST') && str_contains($request, '/executions')) {
            return true;
        } elseif (str_ends_with($domain, System::getEnv('_APP_DOMAIN_FUNCTIONS'))) {
            return true;
        } elseif ($this->domains->get(md5($domain), 'value') === 1) {
            // executions request coming from custom domain
            return true;
        }
        return false;
```

we can probably skip the var 👍🏻

christyjacob4 (approved):

PR #9689, opened 2025-04-25
feat(users): synchronously delete identities and targets in controller

basert:


app/controllers/api/users.php
```diff
@@ -2275,6 +2275,8 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
         $clone = clone $user;
 
         $dbForProject->deleteDocument('users', $userId);
+        \Appwrite\Deletes\Identities::delete($dbForProject, Query::equal('userInternalId', [$user->getInternalId()]));
```
christyjacob4, line 2278: lets import \Appwrite\Deletes and use `Identities::delete` in the worker as well

christyjacob4 (approved):

PR #9693, opened 2025-04-26
Database aliases, new terminologies and a migration to modules

ItzNotABug:
## What does this PR do?

This PR does the following -
1. Moves the `database` controller to a module structure
2. Introduces a new `Table` API, similar to Collections with new `Column` & `Row` APIs
3. Refactor the Realtime events to send both `Collection` and `Table` events until we fully deprecate `Collection` APIs
3. Updates the tests to integrate the `Table` APIs alongside `Collection` APIs for improved coverage to make sure everything is consistent across versions!
4. Also adds the new scopes to the new APIs along with the existing ones

## Test Plan

Unit & E2E Tests!

## Related PRs and Issues

N/A.

## Checklist


ItzNotABug:
TODOs: 
1. event system, 
2. response filtering, @abnegate won't this increase the payload when sending the data back while including both the keys [legacy and new]? Or do we just do it on `1.6.x`, that would make more sense.

app/config/events.php
```diff
@@ -95,22 +95,22 @@
         '$model' => Response::MODEL_DATABASE,
         '$resource' => true,
         '$description' => 'This event triggers on any database event.',
-        'collections' => [
+        'tables' => [
```
abnegate: We should have both types of events: collections API will use collections events, tables API will use tables events

src/Appwrite/Platform/Modules/Databases/Http/Columns/Boolean/Update.php
```diff
@@ -0,0 +1,86 @@
...
+                    )
+                ],
+                contentType: ContentType::JSON
+            ))
+            ->param('databaseId', '', new UID(), 'Database ID.')
+            ->param('tableId', '', new UID(), 'Table ID. You can create a new table using the Database service [server integration](https://appwrite.io/docs/server/databases#databasesCreateCollection).')
+            ->param('key', '', new Key(), 'Column Key.')
+            ->param('required', null, new Boolean(), 'Is column required?')
+            ->param('default', null, new Nullable(new Boolean()), 'Default value for column when not provided. Cannot be set when column is required.')
+            ->param('newKey', null, new Key(), 'New Column Key.', true)
```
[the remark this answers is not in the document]
ItzNotABug: Its correct as per `1.6.x` source - https://github.com/appwrite/appwrite/blob/1.6.x/app/controllers/api/databases.php#L2518

src/Appwrite/Platform/Modules/Databases/Http/Indexes/Get.php
```diff
@@ -0,0 +1,80 @@
...
+                    new SDKResponse(
+                        code: SwooleResponse::STATUS_CODE_OK,
+                        model: UtopiaResponse::MODEL_INDEX,
+                    )
+                ],
+                contentType: ContentType::JSON
+            ))
+            ->param('databaseId', '', new UID(), 'Database ID.')
+            ->param('tableId', '', new UID(), 'Table ID. You can create a new table using the Database service [server integration](https://appwrite.io/docs/server/databases#databasesCreateCollection).')
+            ->param('key', null, new Key(), 'Index Key.')
```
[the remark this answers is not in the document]
ItzNotABug: Still correct from `1.6.x` - https://github.com/appwrite/appwrite/blob/1.6.x/app/controllers/api/databases.php#L3071

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Delete.php
```diff
@@ -0,0 +1,105 @@
...
+
+        $queueForDatabase
+            ->setType(DATABASE_TYPE_DELETE_COLLECTION)
+            ->setDatabase($database);
+
+        if ($this->isCollectionsAPI()) {
+            $queueForDatabase->setCollection($collection);
+        } else {
+            $queueForDatabase->setTable($collection);
+        }
```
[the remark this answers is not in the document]
ItzNotABug: should be fine since its for the workers.

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Indexes/Delete.php
```diff
@@ -0,0 +1,126 @@
...
+
+        if ($db->isEmpty()) {
+            throw new Exception(Exception::DATABASE_NOT_FOUND);
+        }
+        $collection = $dbForProject->getDocument('database_' . $db->getInternalId(), $collectionId);
+
+        if ($collection->isEmpty()) {
+            // table or collection.
+            throw new Exception($this->getGrantParentNotFoundException());
+        }
```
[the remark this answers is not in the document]
ItzNotABug: there's a `$this->getGrantParentNotFoundException()`.

app/config/errors.php
```diff
@@ -71,7 +71,7 @@
     ],
     Exception::GENERAL_QUERY_LIMIT_EXCEEDED => [
         'name' => Exception::GENERAL_QUERY_LIMIT_EXCEEDED,
-        'description' => 'Query limit exceeded for the current attribute. Usage of more than 100 query values on a single attribute is prohibited.',
+        'description' => 'Query limit exceeded for the current attribute/column. Usage of more than 100 query values on a single attribute/column is prohibited.',
```
abnegate: Can we split the messages here? Also 100 shouldn't be hard-coded, this is dynamic depending on $db->getMaxQueryValues()

app/config/errors.php
```diff
@@ -757,13 +811,67 @@
         'description' => 'The attribute type is invalid.',
         'code' => 400,
     ],
+    Exception::ATTRIBUTE_INVALID_RESIZE => [
+        'name' => Exception::ATTRIBUTE_INVALID_RESIZE,
+        'description' => "Existing data is too large for new size, truncate your existing data then try again.",
```
abnegate: Let's single quote where not using interpolation

app/config/events.php
```diff
@@ -95,6 +95,56 @@
...
+                'delete' => [
+                    '$description' => 'This event triggers when an index is deleted.'
+                ]
+            ],
+            'columns' => [
+                '$model' => Response::MODEL_COLUMN,
+                '$resource' => true,
+                '$description' => 'This event triggers on any columns event.',
+                'create' => [
+                    '$description' => 'This event triggers when an column is created.',
```
abnegate: Should be `a column` vs `an column`

app/config/services.php
```diff
@@ -58,7 +58,7 @@
         'name' => 'Databases',
         'subtitle' => 'The Databases service allows you to create structured collections of documents, query and filter lists of documents',
         'description' => '/docs/services/databases.md',
-        'controller' => 'api/databases.php',
+        'controller' => '', // Uses modules
```
abnegate, line 61: What effect does missing the file here have? Is it only used to `include` the file?

app/controllers/api/project.php
```diff
@@ -354,6 +354,7 @@
             'executionsMbSecondsTotal' => $total[METRIC_EXECUTIONS_MB_SECONDS],
             'buildsMbSecondsTotal' => $total[METRIC_BUILDS_MB_SECONDS],
             'documentsTotal' => $total[METRIC_DOCUMENTS],
+            'rowsTotal' => $total[METRIC_DOCUMENTS],
```
abnegate, line 357: Do we need `tablesTotal` here too?

app/controllers/general.php
```diff
@@ -1161,7 +1164,12 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
                 }
                 break;
             case 'Utopia\Database\Exception\Conflict':
-                $error = new AppwriteException(AppwriteException::DOCUMENT_UPDATE_CONFLICT, previous: $error);
+                $error = new AppwriteException(
+                    $isTablesAPI
+                        ? AppwriteException::ROW_UPDATE_CONFLICT
+                        : AppwriteException::DOCUMENT_UPDATE_CONFLICT,
+                    previous: $error
+                );
```
abnegate: We should move away from this global catch, and instead catch these exceptions specifically in each route that can throw it

app/controllers/general.php
```diff
@@ -1596,6 +1630,20 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
...
+    ->inject('request')
+    ->inject('dbForProject')
+    ->action(function (Request $request, Database $dbForProject) {
+        $timeout = \intval($request->getHeader('x-appwrite-timeout'));
+
+        if (!empty($timeout) && App::isDevelopment()) {
+            $dbForProject->setTimeout($timeout);
+        }
+    });
+
```
abnegate: Does module style not have it's own way to handle init hooks?

app/init/constants.php
```diff
@@ -265,3 +265,15 @@
...
+// Context constants for database
+
+const DATABASE_ROWS_CONTEXT = 'row';
+const DATABASE_TABLES_CONTEXT = 'table';
+const DATABASE_COLUMNS_CONTEXT = 'column';
+const DATABASE_INDEX_CONTEXT = 'index';
+
+const DATABASE_DOCUMENTS_CONTEXT = 'document';
+const DATABASE_ATTRIBUTES_CONTEXT = 'attribute';
+const DATABASE_COLLECTIONS_CONTEXT = 'collection';
```
abnegate: What are these for? Let's move them into the module unless we really need them globally

app/init/resources.php
```diff
@@ -707,6 +707,7 @@ function getDevice(string $root, string $connection = ''): Device
         },
     ];
 
+    // NOTE: `params` and `urls` are not used internally in the `Schema::build` function below!
```
abnegate, line 687: What does this mean?

src/Appwrite/Event/Realtime.php
```diff
@@ -72,20 +72,26 @@ public function trigger(): string|bool
...
+        $events = $this->mirrorCollectionEvents($firstEvent, $events);
+
         $payload = new Document($this->getPayload());
 
         $db = $this->getContext('database');
-        $collection = $this->getContext('collection');
         $bucket = $this->getContext('bucket');
 
+        // Can be Tables API or Collections API; generated channels include both!
+        $tableOrCollection = $this->getContext('table') ?? $this->getContext('collection');
```
abnegate, line 83: Do we need to include both? Shouldn't collections events should fire for collections API, and tables events for tables API?

src/Appwrite/Event/Realtime.php
```diff
@@ -72,20 +72,26 @@ public function trigger(): string|bool
             return false;
         }
 
-        $allEvents = Event::generateEvents($this->getEvent(), $this->getParams());
+        $events = Event::generateEvents($this->getEvent(), $this->getParams());
+        $firstEvent = $events[0]; // most verbose event pattern
+
+        // generate and merge all collection and tables api events.
+        $events = $this->mirrorCollectionEvents($firstEvent, $events);
```
abnegate: If we need to do this (not sure we do), it should be responsibility of the `generateEvents` function

src/Appwrite/Extend/Exception.php
```diff
@@ -196,6 +196,11 @@ class Exception extends \Exception
     public const COLLECTION_ALREADY_EXISTS         = 'collection_already_exists';
     public const COLLECTION_LIMIT_EXCEEDED         = 'collection_limit_exceeded';
 
+    /** Tables */
+    public const TABLE_NOT_FOUND              = 'table_not_found';
+    public const TABLE_ALREADY_EXISTS         = 'table_already_exists';
+    public const TABLE_LIMIT_EXCEEDED         = 'table_limit_exceeded';
+
```
abnegate: Let's keep columns aligned in this file

src/Appwrite/Messaging/Adapter/Realtime.php
```diff
@@ -297,19 +298,26 @@ public static function fromPayload(string $event, Document $payload, Document $p
...
+                        throw new \Exception('Collection or the Table needs to be passed to Realtime for Document/Row events in the Database.');
                     }
 
+                    // 1.7.x - Tables API
+                    $channels[] = 'rows';
+                    $channels[] = 'databases.' . $database->getId() .  '.tables.' . $payload->getAttribute('$tableId') . '.rows';
+                    $channels[] = 'databases.' . $database->getId() . '.tables.' . $payload->getAttribute('$tableId') . '.rows.' . $payload->getId();
+
+                    // 1.6.x - Collections API
                     $channels[] = 'documents';
```
abnegate: Let's remove versions from the comments

src/Appwrite/Platform/Services/Workers.php
```diff
@@ -23,7 +22,6 @@ public function __construct()
         $this
             ->addAction(Audits::getName(), new Audits())
             ->addAction(Certificates::getName(), new Certificates())
-            ->addAction(Databases::getName(), new Databases())
```
abnegate, line 26: Should this be removed?

app/config/events.php
ItzNotABug: yepp, already below tables.

app/config/errors.php
ItzNotABug: I don't see this exception being used in the codebase at all 🤔. But anyways, I've made a split for column and attribute.

app/config/services.php
ItzNotABug: yepp. It only points towards the file which handles the legacy controller registration. Module structures no longer require these changes. See - https://github.com/appwrite/appwrite/blob/1.7.x/app/config/services.php#L173C5-L198C7

app/controllers/api/project.php
ItzNotABug: No, we don't add collections/tables in the project's usage model. Only DB and Rows

app/controllers/general.php
ItzNotABug: Agreed.
But do we want to address this in this same PR?
Because this has been a legacy code style from the start. Changing this requires a little bit of refactor around the `POST`, `PATCH` and `DELETE` endpoints for all Row/Document, Index API calls.

app/controllers/general.php
ItzNotABug: Not at the moment, no. Init hooks still need to be registered here.

app/init/constants.php
ItzNotABug: These give a context in our base Actions to know what the api is, Legacy or the New terminology one. We can move them to modules.

src/Appwrite/Event/Realtime.php
ItzNotABug: As far as I remember, we decided to maintain backwards compatibility. So even though the terminology is changed, technically, the underlying structure is still `collections`. We'd start using the Tables API on Console after release but we should still trigger events for legacy types.

src/Appwrite/Event/Realtime.php
ItzNotABug: We need to mirror to generate the events for both types based on one given type. Moving to Events class.

src/Appwrite/Platform/Services/Workers.php
ItzNotABug: Yes, because we use workers in the module structure now and we should register those there. See - https://github.com/appwrite/appwrite/blob/database-aliases/src/Appwrite/Platform/Modules/Databases/Services/Workers.php#L13

app/init/resources.php
ItzNotABug: It means that even though we are building `$params` and the `$urls`, those are not utilized in the `Schema::build` below for GraphQL. Its just a harmless note.

app/controllers/general.php
ItzNotABug: I think these seem to be already moved to respective handlers. will update accordingly.

ItzNotABug: Done.

src/Appwrite/Platform/Modules/Databases/Http/Databases/Tables/Rows/Bulk/Delete.php
```diff
@@ -0,0 +1,71 @@
...
+        $this
+            ->setHttpMethod(self::HTTP_REQUEST_METHOD_DELETE)
+            ->setHttpPath('/v1/databases/:databaseId/tables/:collectionId/rows')
+            ->desc('Delete rows')
+            ->groups(['api', 'database'])
+            ->label('scope', 'documents.write')
+            ->label('resourceType', RESOURCE_TYPE_DATABASES)
+            ->label('audits.event', 'rows.delete')
+            ->label('audits.resource', 'database/{request.databaseId}/table/{request.tableId}')
+            ->label('abuse-key', 'ip:{ip},method:{method},url:{url},userId:{userId}')
```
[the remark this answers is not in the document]
ItzNotABug: fixed. thanks!

tests/e2e/Services/GraphQL/Base.php
```diff
@@ -45,18 +54,74 @@ trait Base
...
+    public static string $GET_INDEXES = 'get_attribute_indexes';
+    public static string $GET_INDEX = 'get_attribute_index';
+    public static string $DELETE_INDEX = 'delete_attribute_index';
+
+    // Column Indexes
+    public static string $CREATE_COLUMN_INDEX = 'create_column_index';
+    public static string $GET_COLUMN_INDEXES = 'get_column_indexes';
+    public static string $GET_COLUMN_INDEX = 'get_column_index';
+    public static string $DELETE_COLUMN_INDEX = 'delete_column_index';
+
```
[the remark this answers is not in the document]
ItzNotABug: Its fine because the index creation is not split into 2. cc @abnegate.
We use `createIndex` for `SDK.Method#name` for both Collection & Tables API.

.github/workflows/tests.yml
```diff
@@ -145,7 +145,8 @@ jobs:
           Account,
           Avatars,
           Console,
-          Databases,
+          Databases/Collections,
```
abnegate: Let's go with `Legacy` instead of `Collections` for the old API, since we may use `Collections` as the name for the new document-based DB API

app/controllers/general.php
abnegate: Looks like you can call `setType(Action::TYPE_INIT)` to register an init hook, check the billing module on cloud

app/init/resources.php
abnegate: Oh got you, yeah those are for building DB graphql schema which didn't scale well so is disabled

src/Appwrite/Platform/Modules/Databases/Context.php
```diff
@@ -0,0 +1,26 @@
...
+    public const DATABASE_ROWS = 'row';
+    public const DATABASE_TABLES = 'table';
+    public const DATABASE_COLUMNS = 'column';
+    public const DATABASE_COLUMN_INDEX = 'columnIndex';
+
+    public const DATABASE_INDEX = 'index';
+    public const DATABASE_DOCUMENTS = 'document';
+    public const DATABASE_ATTRIBUTES = 'attribute';
+    public const DATABASE_COLLECTIONS = 'collection';
+}
```
abnegate: We don't need a class here, we can just define the constants in the namespace. We can also drop the `DATABASE_` prefix because we're in the Databases namespace (and the consts are scoped to it)

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Attributes/Action.php
```diff
@@ -0,0 +1,652 @@
...
+    }
+
+    /**
+     * Get the appropriate parent level not found exception.
+     */
+    final protected function getParentNotFoundException(): string
+    {
+        return $this->isCollectionsAPI()
+            ? Exception::COLLECTION_NOT_FOUND
+            : Exception::TABLE_NOT_FOUND;
```
abnegate, line 90: For flexibility in future, it'll be easier to make all these checks switches instead of if's

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Attribute/Increment.php
```diff
@@ -0,0 +1,126 @@
...
+                id: $documentId,
+                attribute: $attribute,
+                value: $value,
+                max: $max
+            );
+        } catch (ConflictException) {
+            throw new Exception($this->getConflictException());
+        } catch (NotFoundException) {
+            // todo: @itznotabug what do we name this exception now?
+            throw new Exception($this->getStructureNotFoundException());
```
abnegate: Do we need a change here?

src/Appwrite/Platform/Modules/Databases/Context.php
abnegate, line 1: We can name the file `Constants`

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Attributes/Action.php
```diff
@@ -0,0 +1,652 @@
...
+     */
+    final protected function getCollectionsEventsContext(): string
+    {
+        return $this->isCollectionsAPI() ? 'collection' : 'table';
+    }
+
+    /**
+     *  Get the proper column/attribute type based on set context.
+     */
+    final protected function getCorrectModel(string $type, string $format): string
```
abnegate: Can we name this `getModel`?

src/Appwrite/SDK/Specification/Format/OpenAPI3.php
abnegate, line 1: We'll need the same change in Swagger2

tests/e2e/Services/GraphQL/Base.php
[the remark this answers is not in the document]
abnegate: @ItzNotABug It's fine, the queries/mutations are prefixed with the SDK namespace so we'll have `databasesCreateIndex` and `tablesCreateIndex`

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Attributes/Action.php
```diff
@@ -0,0 +1,652 @@
...
+    {
+        return $this->isCollectionsAPI() ? 'attributes' : 'columns';
+    }
+
+    /**
+     * Get the SDK namespace for the current action.
+     */
+    final protected function getSdkNamespace(): string
+    {
+        return $this->isCollectionsAPI() ? 'collections' : 'tables';
```
abnegate: Should be `databases` or `tables` here

tests/e2e/Services/GraphQL/Base.php
```diff
@@ -438,9 +579,54 @@ public function getQuery(string $name): string
                         status
                     }
                 }';
+            case self::$GET_TABLE:
+                return 'query getTable($databaseId: String!, $tableId: String!) {
+                    databasesGetTable(databaseId: $databaseId, tableId: $tableId) {
```
abnegate: These should be `tablesGet` etc. Since tests are passing, we must not run these anywhere, please update

tests/e2e/Services/GraphQL/Base.php
```diff
@@ -438,9 +579,54 @@ public function getQuery(string $name): string
...
+            case self::$DELETE_TABLE:
+                return 'mutation deleteTable($databaseId: String!, $tableId: String!) {
+                    databasesDeleteTable(databaseId: $databaseId, tableId: $tableId) {
+                        status
+                    }
+                }';
             case self::$CREATE_STRING_ATTRIBUTE:
                 return 'mutation createStringAttribute($databaseId: String!, $collectionId: String!, $key: String!, $size: Int!, $required: Boolean!, $default: String, $array: Boolean){
-                    databasesCreateStringAttribute(databaseId: $databaseId, collectionId: $collectionId, key: $key, size: $size, required: $required, default: $default, array: $array) {
+                    collectionsCreateStringAttribute(databaseId: $databaseId, collectionId: $collectionId, key: $key, size: $size, required: $required, default: $default, array: $array) {
```
abnegate: These should stay as `databases`

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Attributes/Action.php
ItzNotABug: For future cases, sure. atm, its only two so keeping it as is. Change should be easy when we decide to add more.

src/Appwrite/Platform/Modules/Databases/Context.php
ItzNotABug: I might as well ditch the manual calling of `setContext` I think. I'll check and report.

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Attribute/Increment.php
ItzNotABug: leftover. removed.

src/Appwrite/SDK/Specification/Format/OpenAPI3.php
ItzNotABug: Don't see any existing references to Attributes or Collection in `Swagger2`.

ItzNotABug: And `Utopia\Database\Validator\Queries\Document(s)` exists on both but don't think those are used in the source.

src/Appwrite/Platform/Modules/Databases/Http/Databases/Tables/Update.php
```diff
@@ -0,0 +1,66 @@
...
+                        model: UtopiaResponse::MODEL_TABLE,
+                    )
+                ],
+                contentType: ContentType::JSON
+            ))
+            ->param('databaseId', '', new UID(), 'Database ID.')
+            ->param('tableId', '', new UID(), 'Table ID.')
+            ->param('name', null, new Text(128), 'Collection name. Max length: 128 chars.')
+            ->param('permissions', null, new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE), 'An array of permission strings. By default, the current permissions are inherited. [Learn more about permissions](https://appwrite.io/docs/permissions).', true)
+            ->param('rowSecurity', false, new Boolean(true), 'Enables configuring permissions for individual rows. A user needs one of row or table level permissions to access a document. [Learn more about permissions](https://appwrite.io/docs/permissions).', true)
```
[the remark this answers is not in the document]
ItzNotABug: Fixed, thanks!

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/XList.php
```diff
@@ -0,0 +1,127 @@
...
+            ->inject('dbForProject')
+            ->callback($this->action(...));
+    }
+
+    public function action(string $databaseId, array $queries, string $search, UtopiaResponse $response, Database $dbForProject): void
+    {
+        print_r(json_encode([
+            'context' => $this->getContext(),
+            'model' => $this->getResponseModel()
+        ], JSON_PRETTY_PRINT));
```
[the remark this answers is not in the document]
ItzNotABug: yikes! leftover, removed.

tests/e2e/Services/GraphQL/Base.php
ItzNotABug: Done. Will let the tests run one more time.

tests/e2e/Services/GraphQL/Base.php
```diff
@@ -2488,6 +3064,249 @@ functionsCreateDuplicateDeployment(functionId: $functionId, deploymentId: $deplo
                         }
                     }
                 }' . PHP_EOL . self::$FRAGMENT_ATTRIBUTES;
+            case self::$COMPLEX_QUERY_TABLE:
+                return 'mutation complex($databaseId: String!, $databaseName: String!, $tableId: String!, $tableName: String!, $rowSecurity: Boolean!, $tablePermissions: [String!]!) {
+                    databasesCreate(databaseId: $databaseId, name: $databaseName) {
+                        _id
+                        name
+                    }
+                    databasesCreateTable(databaseId: $databaseId, tableId: $tableId, name: $tableName, rowSecurity: $rowSecurity, permissions: $tablePermissions) {
```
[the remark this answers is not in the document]
ItzNotABug: Seems like a valid point but I am wondering, how did the test pass 🤔.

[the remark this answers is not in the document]
ItzNotABug: The env `_APP_OPTIONS_ABUSE` for the instance is never used so the tests are always skipped. Checking manually!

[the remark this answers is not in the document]
ItzNotABug: Screenshot 2025-06-19 at 6 29 28 PM

That fixed it. cc @abnegate

src/Appwrite/SDK/Specification/Format/OpenAPI3.php
abnegate: I mean we add to add the new validators that we added here. All changes in OpenAPI spec generation should be mirrored in Swagger2

src/Appwrite/Platform/Modules/Databases/Http/Databases/Tables/Columns/URL/Create.php
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public function __construct()
+    {
+        $this
+            ->setHttpMethod(self::HTTP_REQUEST_METHOD_POST)
+            ->setHttpPath('/v1/databases/:databaseId/tables/:tableId/columns/url')
+            ->desc('Create URL column')
+            ->groups(['api', 'database', 'schema'])
+            ->label('scope', 'collections.write')
```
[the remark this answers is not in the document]
abnegate: @ItzNotABug This is valid, tables should have it's own scopes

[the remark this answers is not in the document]
ItzNotABug: donee

src/Appwrite/Platform/Modules/Databases/Http/Databases/Tables/Columns/String/Update.php
```diff
@@ -0,0 +1,70 @@
...
+                responses: [
+                    new SDKResponse(
+                        code: SwooleResponse::STATUS_CODE_OK,
+                        model: $this->getResponseModel(),
+                    )
+                ],
+                contentType: ContentType::JSON
+            ))
+            ->param('databaseId', '', new UID(), 'Database ID.')
+            ->param('tableId', '', new UID(), 'Table ID. You can create a new table using the Database service [server integration](https://appwrite.io/docs/server/databases#databasesCreateCollection).')
```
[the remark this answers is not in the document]
ItzNotABug: Handled in a different PR - #10030

src/Appwrite/Platform/Modules/Databases/Http/Databases/Tables/Indexes/Delete.php
```diff
@@ -0,0 +1,66 @@
...
+    public function __construct()
+    {
+        $this
+            ->setHttpMethod(self::HTTP_REQUEST_METHOD_DELETE)
+            ->setHttpPath('/v1/databases/:databaseId/tables/:tableId/indexes/:key')
+            ->desc('Delete index')
+            ->groups(['api', 'database'])
+            ->label('scope', 'tables.write')
+            ->label('resourceType', RESOURCE_TYPE_DATABASES)
+            ->label('event', 'databases.[databaseId].tables.[tableId].indexes.[indexId].update')
```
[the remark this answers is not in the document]
ItzNotABug: intentional - https://github.com/appwrite/appwrite/blob/1.8.x/app/controllers/api/databases.php#L3137

abnegate (approved):

PR #9697, opened 2025-04-28
Tls Cert failed msg override

shimonewman:


src/Appwrite/Platform/Workers/Certificates.php
```diff
@@ -171,10 +171,7 @@ private function execute(Domain $domain, Database $dbForPlatform, Mail $queueFor
             $certificate->setAttribute('issueDate', DateTime::now());
             $success = true;
         } catch (Throwable $e) {
-            $logs = $e->getMessage();
-
-            // Set exception as log in certificate document
-            $certificate->setAttribute('logs', \mb_strcut($logs, 0, 1000000));// Limit to 1MB
+            $certificate->setAttribute('logs', 'Failed to create custom domain');
```
stnguyen90, line 174: 1. this isn't always for custom domain.
2. we should still capture logs to help us investigate the issue.

stnguyen90 (changes requested):

src/Appwrite/Platform/Workers/Certificates.php
shimonewman: 1. Cool, Could you phrase it in a more general way?
2. For that we have Sentry, no?

PR #9721, opened 2025-05-05
feat(scheduling): add telemetry for scheduler tasks

basert:


src/Appwrite/Platform/Tasks/ScheduleBase.php
```diff
@@ -110,105 +154,59 @@ public function action(Group $pools, Database $dbForPlatform, callable $getProje
...
                     }
-
-                    $latestDocument = \end($results);
                 }
+            }
 
-                $lastSyncUpdate = $time;
-                $timerEnd = \microtime(true);
-
-                $pools->reclaim();
```
christyjacob4, line 201: the getProjectDB() call consumes one connection from the pool and does not return it back. So it will lead to the pool is empty errors.

christyjacob4 (approved):

christyjacob4 (approved):

Issue #9510, opened 2025-03-15
🐛 Bug Report: Execution stuck at processing


stnguyen90:
### 👟 Reproduction steps

1. Create a function that logs 1000001 chars:

```javascript
export default async ({ req, res, log, error }) => {
  // Create a string that is 1000001 characters long
  const longString = 'a'.repeat(1000001);

  log(longString);

  return res.json({
    motto: 'Build like a team of hundreds_',
    learn: 'https://appwrite.io/docs',
    connect: 'https://appwrite.io/discord',
    getInspired: 'https://builtwith.appwrite.io',
  });
};
```
2. Execute the function

### 👍 Expected behavior

It runs successfully but has a warning about the logs being truncated.

### 👎 Actual Behavior

The execution status stays at processing:

![Image](https://github.com/user-attachments/assets/4c58fd75-3a96-441a-aa3b-11fa1d79159a)

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

stnguyen90:
If the function is executed via `functions.createExecution()` with async set to false, they'll get an error:

```
{
    "message": "Invalid document structure: Attribute \"logs\" has invalid type. Value must be a valid string and no longer than 1000000 chars",
    "code": 400,
    "type": "document_invalid_structure",
    "version": "1.6.2"
}
```

If they execute via the function domain, the function will execute successfully, but an execution record won't be saved in the executions list.

samikshaaagarwal:
Hi, I'm able to reproduce the issue! Taking a look 👀

samikshaaagarwal:
Regarding the original issue, 

Utopia database isn't letting create document with attribute logs which breaches the set limit. 
`[Error] Message: Invalid document structure: Attribute "logs" has invalid type. Value must be a valid string and no longer than 1000000 chars`
I have solved that in the draft PR linked. 

Checking if the other reported issues in the comment are also solved by this 🤔

samikshaaagarwal:
Have fixed the other bits too. @stnguyen90 Please do take a look at the PR!

stnguyen90:
[...]

PR #9773, opened 2025-05-16
Truncate logs in function worker

samikshaaagarwal:
Fixes https://github.com/appwrite/appwrite/issues/9510

## Checklist

## Testing Method 

Following curls were utilised: 

1. Via /executions api with async = true 

Curl derived from console
```
curl 'http://localhost/v1/functions/6825588000018d78a2be/executions' \
  -H 'Accept: */*' \
  -H 'Accept-Language: en-GB,en-US;q=0.9,en;q=0.8' \
  -H 'Connection: keep-alive' \
  -b 'a_session_console_legacy=TOKEN' \
  -H 'DNT: 1' \
  -H 'Origin: http://localhost' \
  -H 'Referer: http://localhost/console/project-6821d9471ced049f3ed2/functions/function-6825588000018d78a2be/executions/execute-function' \
  -H 'Sec-Fetch-Dest: empty' \
  -H 'Sec-Fetch-Mode: cors' \
  -H 'Sec-Fetch-Site: same-origin' \
  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36' \
  -H 'X-Appwrite-Mode: admin' \
  -H 'X-Appwrite-Project: 6821d9471ced049f3ed2' \
  -H 'X-Appwrite-Response-Format: 1.6.0' \
  -H 'X-Fallback-Cookies: {"a_session_console":"TOKEN"}' \
  -H 'content-type: application/json' \
  -H 'sec-ch-ua: "Not.A/Brand";v="99", "Chromium";v="136"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  -H 'x-sdk-language: web' \
  -H 'x-sdk-name: Console' \
  -H 'x-sdk-platform: console' \
  -H 'x-sdk-version: 1.1.0' \
  --data-raw '{"body":"","async":true,"path":"/","method":"GET","headers":{"null":""}}'
```

2. Via /executions api with async = false 

```
Same as above with async = false
```

3. Via domains 

```
curl 'http://682558800a0d98f1150c.functions.localhost' \
  -H 'Accept: */*' \
  -H 'Accept-Language: en-GB,en-US;q=0.9,en;q=0.8' \
  -H 'Connection: keep-alive' \
  -X POST \
  --data-raw '{"body":"","async":false,"path":"/","method":"GET","headers":{"null":""}}' \
  --output output.txt \
  --silent --show-error \
  2>> output.txt
```


samikshaaagarwal:
@stnguyen90 can you please take a look at this before the branch goes very stale. Thanks.

vermakhushboo:
> @stnguyen90 can you please take a look at this before the branch goes very stale. Thanks.

@samikshaaagarwal could you change the base to 1.7.x?

app/controllers/api/functions.php
```diff
@@ -2222,13 +2222,22 @@
                 }
             }
 
+            $maxLogLength = 100000;
+            $logs = $executionResponse['logs'] ?? '';
+            $logsTruncated = false;
+
+            if (\is_string($logs) && \strlen($logs) > $maxLogLength) {
+                $logs = \substr($logs, 0, $maxLogLength) . "\n[WARNING] Logs truncated. The output exceeded {$maxLogLength} characters.";
```
vermakhushboo: @samikshaaagarwal the issue was that the attribute could only save upto `100000` chars. If you truncate to `100000` and append the `warning..` string, it would again surpass the `100000` limit. So make sure that log length never crosses `100000` including the warning.

vermakhushboo: Also, let's add relevant tests

app/controllers/general.php
```diff
@@ -615,10 +615,21 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
...
+                $maxContentLength = $maxLogLength - $warningLength;
+                $logs = \substr($logs, 0, $maxContentLength) . $warningMessage;
+            }
+
             /** Update execution status */
             $status = $executionResponse['statusCode'] >= 500 ? 'failed' : 'completed';
             $execution->setAttribute('status', $status);
-            $execution->setAttribute('logs', $executionResponse['logs']);
+            $execution->setAttribute('logs', $logs);
             $execution->setAttribute('errors', $executionResponse['errors']);
```
Meldiron: @vermakhushboo  We should do same truncate logic for errors too

app/controllers/general.php
```diff
@@ -615,10 +615,21 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
                 }
             }
 
+            // Truncate logs if they exceed the limit
+            $maxLogLength = APP_FUNCTION_LOG_LENGTH_LIMIT;
```
Meldiron, line 619: @vermakhushboo Since this is change to Appwrite Router, we need a PR for Edge as well

tests/resources/functions/log-truncation/index.js
```diff
@@ -0,0 +1,13 @@
+module.exports = async(context) => {
+    // Create a string that is 1000001 characters long (exceeds the 1000000 limit)
+    const longString = 'a'.repeat(1000001);
```
Meldiron: @vermakhushboo  Let's improve the test tiiiny-bit. Let's have the extra character be something else (b), so in test, we can then ensure that this last character is missing.

This ensures that we truncate from end, and not from behinning, which is important decision we made for the flow.

Meldiron: Actually, small change of plans.

Let's remove oldest logs, from the beginning, as that's expected behaviour when it comes to log rotation.

And warning message should then be at the beginning of logs - a place from where logs were truncated

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -2282,4 +2282,70 @@ public function testErrorPagesEmptyBody(): void
...
+        $this->assertLessThanOrEqual(APP_FUNCTION_LOG_LENGTH_LIMIT, strlen($logs));
+        $this->assertStringStartsWith('[WARNING] Logs truncated', $logs);
+
+        $this->assertStringNotContainsString('z', $logs);
+        $this->assertStringContainsString('a', $logs);
+
+        $this->cleanupFunction($functionId);
+    }
+
+    public function testErrorTruncation(): void
```
Meldiron: Let's merge these tests into one, so we dont need to do 2 deployments. they slow down CI/CD by around 10 seconds

tests/resources/functions/error-truncation/index.js
```diff
@@ -0,0 +1,7 @@
+module.exports = async(context) => {
+    // Create a string that is 1000001 characters long (exceeds the 1000000 limit)
+    const longString = 'z' + 'a'.repeat(1000000);
```
Meldiron: Let's re-use the other truncation test. we can do both context.log and context.error there

Meldiron (approved):

Issue #9818, opened 2025-05-19
🐛 Bug Report: Migration failed v1.6.2 -> v1.7.0


PROTAG0N1ST:
### 👟 Reproduction steps

When I migrate to v1.6.2 -> v1.7.0.
## LOGS
[migrate-logs.txt](https://github.com/user-attachments/files/20296885/migrate-logs.txt)

### 👍 Expected behavior

It should migrate to v1.7.0.

### 👎 Actual Behavior

It failed, and the console goes into an infinite loading state.

### 🎲 Appwrite version

Version 1.6.x

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

jmondragon:
I was excited to see the new Sites feature, so I might have jumped the gun on this, but I upgraded from 1.6.2 to 1.7.1 and have the same result.

Console shows spinner but never finishes loading.  Console logs in dev tools shows it looping and uncaught exception here:

B07S-UXP.js:1 Uncaught 
```
function Rn() {
    throw new Error("https://svelte.dev/e/effect_update_depth_exceeded")
}
```

Shadowfita:
Same issue here. Seems to be hanging on:

```Migrating documents for collection "executions"```

**Comment by AapicoDev** [comment]
Same issue here. B07S-UXP.js:1 Uncaught

`function Rn() {
    throw new Error("https://svelte.dev/e/effect_update_depth_exceeded")
}`

**Comment by stnguyen90** [comment]
@PROTAG0N1ST thanks for raising this issue! 🙏🏼 The two relevant log lines I see are:

## Functions

```
Migrating collection "functions"
Failed to create attributes "deploymentId, deploymentCreatedAt, latestDeploymentId, latestDeploymentInternalId, latestDeploymentCreatedAt, latestDeploymentStatus" in collection functions: Row width limit reached. Cannot create new attribute.
Failed to create index "_key_deploymentId" from functions: SQLSTATE[42000]: Syntax error or access violation: 1072 Key column 'deploymentId' doesn't exist in table
```

This one is odd, especially since it doesn't happen for the other projects. Would you please check the first project in your `console_projects` table and note the projects internal ID (`_id` column).

Is the project important?

Then, get details on the table:

[...]

PR #9856, opened 2025-05-21
fix(migration): set owner and region while migrating rules

stnguyen90:
## What does this PR do?

These attributes were added to the rules collection in 1.6.2, but the migration script was not updated to set them. Because they're empty, the 1.7 migration fails while updating rules documents because region is required.

## Test Plan

Tested by setting region and owner to null:

image

Then, I ran the migration:

image

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/9818

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Rule documents now automatically include a "region" attribute based on the associated project's region or a default value.
  - Rule documents now set the "owner" attribute to "Appwrite" when the domain matches specific suffixes.

- **Bug Fixes**
  - Improved handling of missing project information by assigning a default region.

### DIFF
--- a/src/Appwrite/Migration/Version/V22.php
+++ b/src/Appwrite/Migration/Version/V22.php
@@ -12,6 +12,7 @@ use Utopia\Database\Exception\Conflict;
 use Utopia\Database\Exception\Structure;
 use Utopia\Database\Exception\Timeout;
 use Utopia\Database\Query;
+use Utopia\System\System;
 
 class V22 extends Migration
 {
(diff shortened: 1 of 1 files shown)


src/Appwrite/Migration/Version/V22.php
```diff
@@ -346,14 +349,29 @@ private function migrateDocument(Document $document): Document
...
+                $functionsDomain = System::getEnv('_APP_DOMAIN_FUNCTIONS', '');
+                $sitesDomain = System::getEnv('_APP_DOMAIN_SITES', '');
+                $owner = '';
+                if (
+                    (!empty($functionsDomain) && \str_ends_with($domain, $functionsDomain)) ||
+                    (!empty($sitesDomain) && \str_ends_with($domain, $sitesDomain))
+                ) {
+                    $owner = 'Appwrite';
+                }
+                $document->setAttribute('owner', $document->getAttribute('owner', $owner));
```
TorstenDittmann, line 372:
```suggestion
                if (
                    (!empty($functionsDomain) && \str_ends_with($domain, $functionsDomain)) ||
                    (!empty($sitesDomain) && \str_ends_with($domain, $sitesDomain))
                ) {
                    $document->setAttribute('owner', 'Appwrite');
                }
```

should be the same right?

TorstenDittmann (approved):
just that nitpick, was a bit hard to follow 👍🏻

src/Appwrite/Migration/Version/V22.php
stnguyen90: We want to keep whatever value is in the attribute if one is set already. Otherwise, set it to Appwrite if that if resolves to true.

stnguyen90: How about:

```php
      $domain = $document->getAttribute('domain', '');
      $functionsDomain = System::getEnv('_APP_DOMAIN_FUNCTIONS', '');
      $sitesDomain = System::getEnv('_APP_DOMAIN_SITES', '');
      $owner = $document->getAttribute('owner', '');
      if (
          empty($owner) &&
          (!empty($functionsDomain) && \str_ends_with($domain, $functionsDomain)) ||
          (!empty($sitesDomain) && \str_ends_with($domain, $sitesDomain))
      ) {
          $document->setAttribute('owner', 'Appwrite');
      }
```

PR #9859, opened 2025-05-22
fix: send deploymentResourceType in rules verification

basert:
## Summary by CodeRabbit

- **New Features**
  - Enhanced certificate issuance process to dynamically determine domain type based on rule attributes, improving accuracy and flexibility in certificate handling during domain verification updates.

### DIFF
--- a/src/Appwrite/Platform/Modules/Proxy/Http/Rules/Function/Create.php
+++ b/src/Appwrite/Platform/Modules/Proxy/Http/Rules/Function/Create.php
@@ -192,7 +192,7 @@ class Create extends Action
             $queueForCertificates
                 ->setDomain(new Document([
                     'domain' => $rule->getAttribute('domain'),
-                    'domainType' => 'function',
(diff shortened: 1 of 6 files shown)


src/Appwrite/Platform/Modules/Proxy/Http/Rules/Function/Create.php
```diff
@@ -192,7 +192,7 @@ public function action(string $domain, string $functionId, string $branch, Respo
             $queueForCertificates
                 ->setDomain(new Document([
                     'domain' => $rule->getAttribute('domain'),
-                    'domainType' => 'function',
+                    'domainType' => $rule->getAttribute('deploymentResourceType', $rule->getAttribute('type')),
```
christyjacob4, line 195: For all existing rules, `type` and  `deploymentResourceType` are empty. Dont think we can use the domainType from those attributes

christyjacob4 (approved):

PR #9874, opened 2025-05-23
Update: `z-index` to be the highest

ItzNotABug:
## What does this PR do?

Available on all the latest browsers after 2022.
https://caniuse.com/mdn-css_types_calc-keyword_infinity

## Test Plan

Manual.

## Related PRs and Issues

N/A.

## Checklist

### DIFF
--- a/src/Appwrite/Transformation/Adapter/Preview.php
+++ b/src/Appwrite/Transformation/Adapter/Preview.php
@@ -47,7 +47,7 @@ class Preview extends Adapter
                 position: fixed;
                 right: 16px;
                 bottom: 16px;
-                z-index: 1;
+                z-index: calc(infinity);
                 border-radius: var(--border-radius-S, 8px);
                 border: var(--border-width-S, 1px) solid var(--color-border-neutral, #EDEDF0);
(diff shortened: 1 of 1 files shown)


src/Appwrite/Transformation/Adapter/Preview.php
```diff
@@ -47,7 +47,7 @@ public function transform(): void
                 position: fixed;
                 right: 16px;
                 bottom: 16px;
-                z-index: 1;
+                z-index: calc(infinity);
```
Meldiron, line 50: !important too, maybe? 🤔

Meldiron: try to have custom css on your site, and do
```
#appwrite-preview {
  z-index: -1 !important;
  }
```

By adding important to ours, we prevent others from changing it easily. We should probably do same for all params

ItzNotABug: that I understood but not quite the point.
This isn't a general usecase and something common. I can still add it but someone can also hide the whole banner with JS as well 😉.

atharvadeosthale (approved):

PR #9985, opened 2025-06-10
1.8.x

abnegate:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/databases.php
```diff
@@ -1347,7 +1347,17 @@ function updateAttribute(
...
     ->inject('queueForEvents')
-    ->action(function (string $databaseId, string $collectionId, string $key, ?int $size, ?bool $required, ?string $default, bool $array, bool $encrypt, Response $response, Database $dbForProject, EventDatabase $queueForDatabase, Event $queueForEvents) {
+    ->inject('plan')
+    ->action(function (string $databaseId, string $collectionId, string $key, ?int $size, ?bool $required, ?string $default, bool $array, bool $encrypt, Response $response, Database $dbForProject, EventDatabase $queueForDatabase, Event $queueForEvents, array $plan) {
+        if ($encrypt && !empty($plan) && !($plan['databasesAllowEncrypt'] ?? false)) {
+            throw new Exception(Exception::GENERAL_BAD_REQUEST, 'Encrypted string attributes are not available on your plan. Please upgrade to create encrypted string attributes.');
+        }
+        if ($encrypt && $size < APP_DATABASE_ENCRYPT_SIZE_MIN) {
+            throw new Exception(
+                Exception::GENERAL_BAD_REQUEST,
```
eldadfux, line 1359: Why do we need a minimum size here? is this a must?

app/controllers/api/storage.php
```diff
@@ -1345,6 +1345,15 @@
             $contentType = $file->getAttribute('mimeType');
         }
 
+        $response
```
eldadfux: Why did we make this changes?

app/init/database/filters.php
```diff
@@ -71,18 +71,27 @@ function (mixed $value) {
...
+                    $options = $attribute->getAttribute('options');
+                    foreach ($options as $key => $value) {
+                        $attribute->setAttribute($key, $value);
+                    }
+                    $attribute->removeAttribute('options');
+                    break;
+
+                case Database::VAR_STRING:
+                    $filters = $attribute->getAttribute('filters', []);
+                    $attribute->setAttribute('encrypt', in_array('encrypt', $filters));
```
eldadfux, line 93: Why do we set this "virtual" attribute is this just a convenience or we have other reasons for it?

src/Appwrite/Utopia/Database/Validator/Queries/Base.php
```diff
@@ -71,8 +71,8 @@ public function __construct(string $collection, array $allowedAttributes)
             'array' => false,
         ]);
 
-        $internalId = new Document([
-            'key' => '$internalId',
+        $sequence = new Document([
+            'key' => '$sequence',
             'type' => Database::VAR_STRING,
             'array' => false,
         ]);
```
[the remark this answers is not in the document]
eldadfux: @abnegate this comment make sense.

src/Appwrite/Utopia/Response/Model/Document.php
```diff
@@ -36,6 +36,12 @@ public function __construct()
                 'default' => '',
                 'example' => '5e5ea5c16897e',
             ])
+            ->addRule('$sequence', [
+                'type' => self::TYPE_INTEGER,
+                'description' => 'Document automatically incrementing ID.',
```
eldadfux, line 41: We might want to have a small section for this on the docs. Let's discuss with DevRel.

eldadfux (commented):
@abnegate question about sequences, today, when we migrate a project internally with  one of our ZDT migration scripts from one db to another do we preserve them with same values? What happens also on the public migration / backup services?

app/controllers/api/databases.php
abnegate: Encrypting 1 char results in ~150 chars in the DB, so if you set size smaller than that, you'll always get a validation failure. We have an inline validation on the field on console though, so you don't have to hit the backend to get the error first

app/controllers/api/storage.php
abnegate: Looks like a sync error, will check

app/init/database/filters.php
abnegate: Just for convenience

src/Appwrite/Utopia/Database/Validator/Queries/Base.php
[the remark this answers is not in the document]
abnegate: Addressed in increment PR

src/Appwrite/Utopia/Response/Model/Document.php
abnegate: Yep @adityaoberai is handling this

abnegate:
> @abnegate question about sequences, today, when we migrate a project internally with one of our ZDT migration scripts from one db to another do we preserve them with same values? What happens also on the public migration / backup services?

@eldadfux Yep we preserve across ZDT. For migrations and backups, we preserve for only the database service (since we still use the API for all other services)

PR #9996, opened 2025-06-11
feat after build hook

loks0n:
## Summary by CodeRabbit

- **New Features**
  - Added an internal extension point to enable future enhancements after a successful build process. No visible changes to end-users at this time.

### DIFF
--- a/src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
+++ b/src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
@@ -877,6 +877,10 @@ class Builds extends Action
                 }
             }
 
+            $deployment->setAttribute('buildLogs', $logs);
+
+            $this->afterBuildSuccess($dbForProject, $deployment);
+
             $deployment = $dbForProject->updateDocument('deployments', $deployment->getId(), $deployment);
 
             $queueForRealtime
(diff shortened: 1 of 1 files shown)


src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
```diff
@@ -1313,6 +1317,19 @@ protected function sendUsage(Document $resource, Document $deployment, Document
...
     }
 
+    /**
+     * Hook to run after build success
+     *
+     * @param Database $dbForProject
+     * @param Document $deployment
+     * @return void
+     */
+    protected function afterBuildSuccess(Database $dbForProject, Document &$deployment): void
```
loks0n, line 1327: Deployment has to be passed by ref due to how to write logs

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
```diff
@@ -1313,6 +1317,19 @@ protected function sendUsage(Document $resource, Document $deployment, Document
...
+    {
+        assert($dbForProject instanceof Database);
+        assert($deployment instanceof Document);
+    }
```
christyjacob4, line 1331: do we need an assertion here? The function can be empty also right?

christyjacob4 (approved):

PR #9998, opened 2025-06-12
Fix all vcs urls missing region

vermakhushboo:
## What does this PR do?

Add `region` to all `/console/project-` endpoints for VCS flows

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/controllers/api/vcs.php
+++ b/app/controllers/api/vcs.php
@@ -363,6 +363,7 @@ $createGitDeployments = function (GitHub $github, string $providerInstallationId
             if (!empty($providerCommitHash) && $resource->getAttribute('providerSilentMode', false) === false) {
                 $resourceName = $resource->getAttribute('name');
                 $projectName = $project->getAttribute('name');
+                $region = $project->getAttribute('region', 'default');
                 $name = "{$resourceName} ({$projectName})";
                 $message = 'Starting...';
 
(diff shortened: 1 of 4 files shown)


app/controllers/general.php
```diff
@@ -345,7 +346,7 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
             $permissions = $resource->getAttribute('execute');
             if (!(\in_array('any', $permissions)) && !(\in_array('guests', $permissions))) {
                 $exception = new AppwriteException(AppwriteException::FUNCTION_EXECUTE_PERMISSION_MISSING, view: $errorView);
-                $exception->addCTA('View settings', $url . '/console/project-' . $project->getId() . '/functions/function-' . $resource->getId() . '/settings');
+                $exception->addCTA('View settings', $url . '/console/project-' . $project->getAttribute('region', 'default') . '-' . $project->getId() . '/functions/function-' . $resource->getId() . '/settings');
```
Meldiron, line 349: What $url is this using? I think we need to use CONSOLE env var here too

https://github.com/appwrite/appwrite/blob/fix-github-check-urls/app/controllers/general.php#L265

app/controllers/api/vcs.php
```diff
@@ -377,7 +378,7 @@
                 }
                 $owner = $github->getOwnerName($providerInstallationId);
 
-                $providerTargetUrl = $request->getProtocol() . '://' . $request->getHostname() . "/console/project-$projectId/$resourceCollection/$resourceType-$resourceId";
+                $providerTargetUrl = $request->getProtocol() . '://' . $request->getHostname() . "/console/project-$region-$projectId/$resourceCollection/$resourceType-$resourceId";
```
Meldiron, line 381: Let's ignore current requests's protocol and hostname. Instead, do what we do in worker - env var for http/https, and console env var for hostname

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
```diff
@@ -1444,10 +1444,11 @@ protected function runGitAction(
             $hostname = System::getEnv('_APP_DOMAIN');
```
Meldiron, line 1444: I think here we need to use console hostname env var

app/controllers/api/vcs.php
vermakhushboo: Not related to this PR, updating it in the PR for env vars

christyjacob4 (approved):

PR #9999, opened 2025-06-12
Use _APP_CONSOLE_DOMAIN, if not found, then use _APP_DOMAIN

vermakhushboo:
## What does this PR do?

This would help prevent `region.` prefix from domains.

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/config/templates/site.php
```diff
@@ -7,7 +7,7 @@
  */
 
 $protocol = System::getEnv('_APP_OPTIONS_FORCE_HTTPS') == 'disabled' ? 'http' : 'https';
-$hostname = System::getEnv('_APP_DOMAIN');
+$hostname = System::getEnv('_APP_CONSOLE_DOMAIN', System::getEnv('_APP_DOMAIN', ''));
```
Meldiron, line 10: With this, let's set _APP_CONSOLE_DOMAIN=localhost in .env for local developemt, and tests.
Then, we can remove development override just below, and it should still work. Please test, by visitng Console and making sure templates have screenshots properly

app/controllers/api/console.php
```diff
@@ -43,8 +43,8 @@
     ))
     ->inject('response')
     ->action(function (Response $response) {
-        $validator = new Domain(System::getEnv('_APP_DOMAIN'));
-        $isDomainValid = !empty(System::getEnv('_APP_DOMAIN', '')) && $validator->isKnown() && !$validator->isTest();
+        $validator = new Domain(System::getEnv('_APP_CONSOLE_DOMAIN', System::getEnv('_APP_DOMAIN', '')));
+        $isDomainValid = !empty(System::getEnv('_APP_CONSOLE_DOMAIN', System::getEnv('_APP_DOMAIN', ''))) && $validator->isKnown() && !$validator->isTest();
```
Meldiron: We shouldnt replace all the places, only what's relevant - what's for frontend. Domain validation is very likely for Proxy and DNS.
If we need Console for something on frontend, we can expose that as separate variable

app/controllers/general.php
```diff
@@ -448,7 +448,7 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
         }
 
         $protocol = System::getEnv('_APP_OPTIONS_FORCE_HTTPS') == 'disabled' ? 'http' : 'https';
-        $hostname = System::getEnv('_APP_DOMAIN');
+        $hostname = System::getEnv('_APP_CONSOLE_DOMAIN', System::getEnv('_APP_DOMAIN', ''));
```
Meldiron: This is endpoint for functions, so they can conncet server SDK. Those SDKs are expected to use region endpoint

app/controllers/general.php
```diff
@@ -809,7 +809,7 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
         * Appwrite Router
         */
         $host = $request->getHostname() ?? '';
-        $mainDomain = System::getEnv('_APP_DOMAIN', '');
+        $mainDomain = System::getEnv('_APP_CONSOLE_DOMAIN', System::getEnv('_APP_DOMAIN', ''));
```
Meldiron: I am worried this would break when router runs and when not. Since this already works, lets not touch it.

vermakhushboo:
Replace `$request->getProtocol()` and `$request->getHostname()` with env vars

app/controllers/api/vcs.php
```diff
@@ -116,8 +116,10 @@
             }
 
             $commentStatus = $isAuthorized ? 'waiting' : 'failed';
+            $protocol = System::getEnv('_APP_OPTIONS_FORCE_HTTPS') == 'disabled' ? 'http' : 'https';
```
christyjacob4:
```suggestion
            $protocol = System::getEnv('_APP_OPTIONS_FORCE_HTTPS') === 'disabled' ? 'http' : 'https';
```
use strict type checking

app/controllers/api/vcs.php
```diff
@@ -437,14 +439,16 @@
         ]);
 
         $appName = System::getEnv('_APP_VCS_GITHUB_APP_NAME');
+        $protocol = System::getEnv('_APP_OPTIONS_FORCE_HTTPS') == 'disabled' ? 'http' : 'https';
```
christyjacob4: same here

src/Appwrite/Vcs/Comment.php
```diff
@@ -87,7 +87,7 @@ public function generateComment(): string
         $i = 0;
         foreach ($projects as $projectId => $project) {
             $protocol = System::getEnv('_APP_OPTIONS_FORCE_HTTPS') == 'disabled' ? 'http' : 'https';
```
christyjacob4: same here

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
```diff
@@ -594,7 +594,7 @@ protected function buildDeployment(
             ]);
 
             $protocol = System::getEnv('_APP_OPTIONS_FORCE_HTTPS') == 'disabled' ? 'http' : 'https';
-            $hostname = System::getEnv('_APP_DOMAIN');
+            $hostname = System::getEnv('_APP_CONSOLE_DOMAIN', System::getEnv('_APP_DOMAIN', ''));
```
Meldiron: Possible problem

Meldiron (approved):

PR #10011, opened 2025-06-15
Override project via custom domains log

shimonewman:
## Summary by CodeRabbit

- **Chores**
  - Improved internal handling of project data during custom domain function execution. No visible changes to end-user features.


app/controllers/general.php
```diff
@@ -1267,7 +1273,12 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
             $log->addTag('url', $request->getURI());
             $log->addTag('verboseType', get_class($error));
             $log->addTag('code', $error->getCode());
-            $log->addTag('projectId', $project->getId());
+
+            $tags = $log->getTags();
+            if (isset($tags['projectId'])) {
```
stnguyen90: Shouldn't this be:

```suggestion
            if (!isset($tags['projectId'])) {
```

PR #10019, opened 2025-06-17
Lazy load relationships

ItzNotABug:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/tests/e2e/Services/Databases/Tables/DatabasesBase.php
+++ b/tests/e2e/Services/Databases/Tables/DatabasesBase.php
@@ -4572,7 +4572,11 @@ trait DatabasesBase
         $response = $this->client->call(Client::METHOD_GET, '/databases/' . $databaseId . '/tables/' . $personCollection . '/rows/' . $person2['body']['$id'], array_merge([
             'content-type' => 'application/json',
             'x-appwrite-project' => $this->getProject()['$id'],
-        ], $this->getHeaders()));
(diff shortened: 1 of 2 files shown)


app/controllers/api/databases.php
```diff
@@ -3649,8 +3649,20 @@ function updateAttribute(
...
         try {
-            $documents = $dbForProject->find('database_' . $database->getSequence() . '_collection_' . $collection->getSequence(), $queries);
+            $selectQueries = Query::groupByType($queries)['selections'] ?? [];
+
+            if (! empty($selectQueries)) {
+                // has selects, allow relationship on documents!
+                $documents = $dbForProject->find('database_' . $database->getSequence() . '_collection_' . $collection->getSequence(), $queries);
+            } else {
+                // has no selects, disable relationship looping on documents!
+                $documents = $dbForProject->skipRelationships(fn () => $dbForProject->find('database_' . $database->getSequence() . '_collection_' . $collection->getSequence(), $queries));
```
abnegate: I think we still can't merge because of this, relationships won't be fetched without passing select queries

ItzNotABug: @abnegate lets mark this PR in draft then? removing this check basically makes the PR of not much use anyways. We can merge it in the end since the structure would be the module by the end anyways. Wdyt?

PR #10023, opened 2025-06-18
Feat txn

abnegate:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Modules/Databases/Http/Databases/Transactions/Update.php
+++ b/src/Appwrite/Platform/Modules/Databases/Http/Databases/Transactions/Update.php
@@ -403,10 +403,11 @@ class Update extends Action
             $data['$id'] = $documentId;
         }
         $dbForProject->withRequestTimestamp($createdAt, function () use ($dbForProject, $collectionId, $data, &$state) {
-            $state[$collectionId][$data['$id']] = $dbForProject->createDocument(
+            $doc = $dbForProject->createDocument(
                 $collectionId,
                 new Document($data),
             );
+            $state[$collectionId][$doc->getId()] = $doc;
         });
     }
 
@@ -493,11 +494,12 @@ class Update extends Action
             return;
         }
 
-        $dbForProject->withRequestTimestamp($createdAt, function () use ($dbForProject, $collectionId, $documentId, $data, &$state) {
-            $state[$collectionId][$documentId] = $dbForProject->upsertDocument(
+        $dbForProject->withRequestTimestamp($createdAt, function () use ($dbForProject, $collectionId, $data, &$state) {
+            $doc = $dbForProject->upsertDocument(
                 $collectionId,
                 new Document($data),
             );
+            $state[$collectionId][$doc->getId()] = $doc;
         });
     }
 
--- a/tests/e2e/Services/Databases/TablesDB/Transactions/TransactionsBase.php
+++ b/tests/e2e/Services/Databases/TablesDB/Transactions/TransactionsBase.php
@@ -5176,4 +5176,143 @@ trait TransactionsBase
             $this->assertEquals('processed', $response['body']['flag'], 'Bulk update should have matched document created in same transaction');
         }
     }
+
+    /**
+     * Test upsert with auto-generated ID followed by update
+     * This tests that the transaction state properly stores the document under its actual ID,
+     * not under null when the ID is auto-generated
+     */
+    public function testUpsertAutoIdThenUpdate(): void
+    {
+        // Create database and table
+        $database = $this->client->call(Client::METHOD_POST, '/tablesdb', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'x-appwrite-key' => $this->getProject()['apiKey']
+        ]), [
+            'databaseId' => ID::unique(),
+            'name' => 'UpsertAutoIDTestDB'
+        ]);
+
+        $databaseId = $database['body']['$id'];
+
+        $table = $this->client->call(Client::METHOD_POST, "/tablesdb/{$databaseId}/tables", array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'x-appwrite-key' => $this->getProject()['apiKey']
+        ]), [
+            'tableId' => ID::unique(),
+            'name' => 'TestCollection',
+            'permissions' => [
+                Permission::create(Role::any()),
+                Permission::read(Role::any()),
+                Permission::update(Role::any()),
+            ],
+        ]);
+
+        $tableId = $table['body']['$id'];
+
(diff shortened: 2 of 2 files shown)


abnegate:
@coderabbitai full review

app/config/collections/projects.php
```diff
@@ -2521,4 +2521,128 @@
...
+    ],
+
+    'transactionLogs' => [
+        '$collection' => ID::custom(Database::METADATA),
+        '$id' => ID::custom('transactionLogs'),
+        'name' => 'Transaction Logs',
+        'attributes' => [
+            [
+                '$id' => ID::custom('transactionInternalId'),
+                'type' => Database::VAR_STRING,
```
fogelito, line 2579: VAR_ID

app/config/collections/projects.php
```diff
@@ -2521,4 +2521,128 @@
...
+                'size' => 5_000_000, // Allow large payloads for bulk operations
+                'signed' => false,
+                'required' => true,
+                'default' => null,
+                'array' => false,
+                'filters' => ['json'],
+            ],
+        ],
+        'indexes' => [
```
fogelito, line 2638: Do we need indexes for databaseInternalId or maybe collectionInternalId

abnegate: Initially had them, but removed since we don't actually filter by them, always by `transactionInternalId`

app/config/collections/projects.php
```diff
@@ -2521,4 +2521,128 @@
...
+            [
+                '$id' => ID::custom('_key_transaction'),
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['transactionInternalId'],
```
fogelito, line 2642: We can add $createdAt as the second part of the index based on this?

$operations = $this->dbForProject->find('transactionLogs', [
            Query::equal('transactionInternalId', [$transaction->getSequence()]),
            Query::orderAsc('$createdAt'), // Ensure operations are processed in order
        ]);

app/config/collections/projects.php
```diff
@@ -2521,4 +2521,128 @@
...
+                'signed' => true,
+                'required' => true,
+                'default' => null,
+                'array' => false,
+                'filters' => ['datetime'],
+            ],
+        ],
+        'indexes' => [
+            [
+                '$id' => ID::custom('_key_expiresAt'),
```
fogelito, line 2563: Do we need a cleanup in maintenance task?

app/config/collections/projects.php
```diff
@@ -2521,4 +2521,128 @@
...
         ],
     ],
+
+    'transactions' => [
+        '$collection' => ID::custom(Database::METADATA),
+        '$id' => ID::custom('transactions'),
+        'name' => 'Transactions',
+        'attributes' => [
+            [
+                '$id' => ID::custom('status'),
```
fogelito, line 2531: Constants INT if possible :)

docs/references/databases/create-operations.md
eldadfux, line 1: Maybe we can add more references to the docs in places like this.

src/Appwrite/Databases/TransactionState.php
```diff
@@ -0,0 +1,231 @@
+<?php
```
eldadfux, line 1: What's the role of this class?

app/config/collections/projects.php
abnegate: I will actually remove the $createdAt, since we should use default _id ordering instead

app/config/collections/projects.php
abnegate: Yep as part of maintenance action in deletes worker we clean up: https://github.com/appwrite/appwrite/blob/614285f1ae1137ad086c171b8a709b30c06f6663/src/Appwrite/Platform/Workers/Deletes.php#L188

src/Appwrite/Databases/TransactionState.php
abnegate: Provides transaction-aware versions of get/list documents, so that if in transaction scope, you can read documents including any pending changes from the previous operations in the transaction. E.g. updateDocument needs to know about any previous updates in the transaction, so it doesn't blow away pending changes

app/config/collections/projects.php
fogelito: Better to add an explicit order by query for $sequence , not to rely on the code logic defaults. 
We want to remove the default order if the user does not add a query

abnegate:
@eldadfux 

> How do we handle preventing table locks?

The operations we support will only row lock, no table locks

> Do we need to account for any deadlock scenarios?

- If it's a true deadlock, one commit will fail, the other will succeed
- If it's a lock wait, waiter will wait up to 15 seconds (API timeout), if the locker succeeds in that time, it will run as well 

> Are we worried from schema changes during a transaction?

We'll throw a 400 at commit time if the structure doesn't match the staged operations

> Are we measuring db operations for transactions? (btw, did we do this on bulk operations?)

Yep we're tracking usage for reads/writes, same for bulk operations

app/config/collections/projects.php
abnegate: For $id? I don't think we have any yet

src/Appwrite/Platform/Modules/Databases/Http/Databases/Transactions/Update.php
```diff
@@ -0,0 +1,662 @@
...
+            $databaseOperations = [];
+
+            $dbForProject->withTransaction(function () use ($dbForProject, $queueForDeletes, $transactionId, &$transaction, &$operations, &$totalOperations, &$databaseOperations, $queueForEvents, $queueForStatsUsage, $queueForRealtime, $queueForFunctions, $queueForWebhooks) {
+                $dbForProject->updateDocument('transactions', $transactionId, new Document([
+                    'status' => 'committing',
+                ]));
+
+                // Fetch operations ordered by sequence by default to replay operations in exact order they were created
+                $operations = $dbForProject->find('transactionLogs', [
+                    Query::equal('transactionInternalId', [$transaction->getSequence()]),
```
fogelito: Do we need to order by?

fogelito: Do we need query limit since default limit is 25?

src/Appwrite/Databases/TransactionState.php
```diff
@@ -0,0 +1,230 @@
...
+                    } elseif ($docState['action'] === 'upsert') {
+                        // Upsert created a new document
+                        $documentMap[$docId] = $docState['document'];
+                    }
+                }
+            }
+        }
+
+        return array_values($documentMap);
+    }
```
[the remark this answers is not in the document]
abnegate: I don't think they should be reapplied, the queries are meant to be applied to the initial query only, not the final result as well

PR #10033, opened 2025-06-20
fix:  Users.CreateUser... vague error into optional name parameter.

Abiji-2020:
## What does this PR do?
In the api/users.php the function signature requires string for name parameter even though its an optional one and displays a vague message over it. 
I have modified the signature and added a null check and made the name as empty string if the parameter is given as null. 

## Test Plan
 
image
After updating the code, I have created a user with the null values using the api call using `Thunder client VS code extension` (similar to postman)

## Related PRs and Issues

- #8785 

## Checklist

## Summary by CodeRabbit

- **New Features**
  - User creation and target creation endpoints now allow the user name field to be optional. If not provided, it will default to an empty value.

- **Bug Fixes**
  - Improved handling of missing or null user name fields during user and target creation, preventing errors when the name is omitted.

### DIFF
--- a/app/controllers/api/users.php
+++ b/app/controllers/api/users.php
@@ -59,7 +59,7 @@ use Utopia\Validator\Text;
 use Utopia\Validator\WhiteList;
 
 /** TODO: Remove function when we move to using utopia/platform */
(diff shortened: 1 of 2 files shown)


HarshMN2345:

PR #10048, opened 2025-06-24
Fix: Build duration calculation

Meldiron:
## What does this PR do?

Currently just a draft, showcasing PoC of adding init, shutdown and resources to workers.

Works like a charm:

CleanShot 2025-06-24 at 10 54 10@2x

Subimtted PR so we can discuss if structure is good, or needs improvement.

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Modules/Functions/Workers/Builds/Actions/Shutdown.php
+++ b/src/Appwrite/Platform/Modules/Functions/Workers/Builds/Actions/Shutdown.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace Appwrite\Platform\Modules\Functions\Workers\Builds\Actions;
+
+use Exception;
+use Utopia\Platform\Action;
+
+class Shutdown extends Action
+{
+    public static function getName(): string
+    {
+        return 'builds-shutdown';
+    }
+
+    /**
+     * @throws Exception
+     */
+    public function __construct()
+    {
+        $this->setType(self::TYPE_SHUTDOWN);
+        $this
+            ->desc('Builds worker shutdown hook')
+            ->groups(['builds'])
+            ->inject('buildStartTime')
+            ->callback($this->action(...));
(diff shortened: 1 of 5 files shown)


src/Appwrite/Platform/Modules/Functions/Workers/Builds/Builds.php
```diff
@@ -0,0 +1,23 @@
...
+
+class Builds
+{
+    public function __construct(Workers $worker)
+    {
+        $worker->addAction(BuildsInit::getName(), new BuildsInit());
+        $worker->addAction(BuildsHandler::getName(), new BuildsHandler());
+        $worker->addAction(BuildsShutdown::getName(), new BuildsShutdown());
+
+        Server::setResource('buildStartTime', function () {
```
Meldiron, line 19: Not sure I like this is not separate file like `init/resources.php`, but that would no longer follow classes approach.
Cant think of where else to put worker-specific resources.

src/Appwrite/Platform/Modules/Functions/Workers/Builds/Actions/Shutdown.php
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * @throws Exception
+     */
+    public function __construct()
+    {
+        $this->setType(self::TYPE_SHUTDOWN);
+        $this
+            ->desc('Builds worker shutdown hook')
+            ->groups(['builds'])
+            ->inject('buildStartTime')
```
Meldiron, line 24: Not big fan of prefixing resources with worker name. But might be nessessary for resources that are specific to a single worker. startTime could be generic resource, but let's use this as opportunity to find a solution to this problem

Meldiron:
Too complex solution right now. Deserves attention, but later-on

PR #10049, opened 2025-06-24
Fix: Certificate deletion race condition

Meldiron:
## What does this PR do?

Ensures delete worker skips certificate deletion if there is new role for it already.

## Test Plan

- [ ] Manual QA

## Related PRs and Issues

x

## Checklist


Meldiron:
Meh, not needed. it never happens anyway - deleting rule is always part of deleting something else (deployment, site, project..), and there it lists rules as part of job -  so it's accurate.

PR #10057, opened 2025-06-25
fix: inject missing $log

stnguyen90:
## What does this PR do?

$log was used in the action, but it was never injected resulting in an error like:

router(): Argument #7 ($log) must be of type Utopia\Logger\Log, null given

This commit injects the missing $log dependency into the action.

## Test Plan

None

## Related PRs and Issues

- https://github.com/appwrite/appwrite/pull/10011

## Checklist


stnguyen90:
Fixed as part of:

* https://github.com/appwrite/appwrite/pull/10061
* https://github.com/appwrite/appwrite/pull/10067

PR #10066, opened 2025-06-26
Fix: flakyness of account tests

Meldiron:
## What does this PR do?

Account tests were expecting last request to be secret from SMS mock adapter.

Sadly that was overriden by webhook requests - one worker was sometimes faster than the other one.

Separating to 2 contianers prevents the conflict.

Another solution would be to getLastRequestByPath(), but this method doesnt request in request cather, so 2 containers is simplest solution

## Test Plan

tests must pass

## Related PRs and Issues

x

## Checklist


tests/e2e/Scopes/Scope.php
```diff
@@ -76,10 +79,16 @@ protected function extractQueryParamsFromEmailLink(string $html): array
...
     }
 
-    protected function assertLastRequest(callable $probe, $timeoutMs = 20_000, $waitMs = 500): array
+    protected function assertLastRequest(callable $probe, string $type, $timeoutMs = 20_000, $waitMs = 500): array
     {
-        $this->assertEventually(function () use (&$request, $probe) {
-            $request = json_decode(file_get_contents('http://request-catcher:5000/__last_request__'), true);
+        $hostname = match ($type) {
+            'webhook' => 'request-catcher-webhook',
+            'api' => 'request-catcher-api',
```
Meldiron:
```suggestion
            'sms' => 'request-catcher-sms',
```

christyjacob4 (approved):

PR #10096, opened 2025-07-03
Realtime support for bulk api

ArnabChatterjee20k:
## What does this PR do?
-> Event working for bulk api

* They'll fire on existing channels only

* Clients will receive an individual event for each document

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
+++ b/src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
@@ -293,8 +293,16 @@ abstract class Action extends UtopiaAction
      * @param Event $queueForWebhooks
      * @return void
      */
-    final protected function triggerQueuesForBulkDocuments(string $event, Document $database, Document $collection, array $documents, Event $queueForEvents, Event $queueForRealtime, Event $queueForFunctions, Event $queueForWebhooks)
-    {
+    protected function triggerBulk(
+        string $event,
+        Document $database,
+        Document $collection,
+        array $documents,
+        Event $queueForEvents,
+        Event $queueForRealtime,
+        Event $queueForFunctions,
+        Event $queueForWebhooks
+    ): void {
         $queueForEvents
             ->setEvent($event)
             ->setParam('databaseId', $database->getId())
--- a/src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
+++ b/src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
@@ -295,17 +295,19 @@ abstract class Action extends UtopiaAction
      */
     final protected function triggerQueuesForBulkDocuments(string $event, Document $database, Document $collection, array $documents, Event $queueForEvents, Event $queueForRealtime, Event $queueForFunctions, Event $queueForWebhooks)
     {
+        $queueForEvents
+            ->setEvent($event)
+            ->setParam('databaseId', $database->getId())
+            ->setContext('database', $database)
+            ->setParam('collectionId', $collection->getId())
+            ->setParam('tableId', $collection->getId())
+            ->setContext($this->getCollectionsEventsContext(), $collection);
+
         foreach ($documents as $document) {
             $queueForEvents
-                ->setEvent($event)
-                ->setParam('databaseId', $database->getId())
-                ->setContext('database', $database)
-                ->setParam('collectionId', $collection->getId())
-                ->setParam('tableId', $collection->getId())
                 ->setParam('documentId', $document->getId())
                 ->setParam('rowId', $document->getId())
-                ->setPayload($document->getArrayCopy())
-                ->setContext($this->getCollectionsEventsContext(), $collection);
+                ->setPayload($document->getArrayCopy());
 
             $queueForRealtime
                 ->from($queueForEvents)
--- a/src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
+++ b/src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
@@ -2,6 +2,7 @@
 
 namespace Appwrite\Platform\Modules\Databases\Http\Databases\Collections\Documents;
 
+use Appwrite\Event\Event;
 use Appwrite\Extend\Exception;
 use Utopia\Database\Database;
 use Utopia\Database\Document;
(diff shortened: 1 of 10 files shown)


app/controllers/api/databases.php
```diff
@@ -3309,9 +3309,6 @@ function updateAttribute(
         $isAPIKey = Auth::isAppUser(Authorization::getRoles());
         $isPrivilegedUser = Auth::isPrivilegedUser(Authorization::getRoles());
 
-        if ($isBulk && !$isAPIKey && !$isPrivilegedUser) {
-            throw new Exception(Exception::GENERAL_UNAUTHORIZED_SCOPE);
-        }
```
abnegate: This should still be here

app/controllers/api/databases.php
```diff
@@ -4659,12 +4657,12 @@ function updateAttribute(
             throw new Exception(Exception::DOCUMENT_MISSING_PAYLOAD);
         }
 
-        $database = $dbForProject->getDocument('databases', $databaseId);
+        $database = Authorization::skip(fn () => $dbForProject->getDocument('databases', $databaseId));
```
abnegate: We don't need to skip auth in update/delete documents because the routes are server side only, and API keys always disable auth

app/controllers/api/databases.php
```diff
@@ -4721,6 +4719,13 @@ function updateAttribute(
             ->addMetric(METRIC_DATABASES_OPERATIONS_WRITES, \max(1, $modified))
             ->addMetric(str_replace('{databaseInternalId}', $database->getSequence(), METRIC_DATABASE_ID_OPERATIONS_WRITES), \max(1, $modified));
 
+        $queueForEvents
+            ->setEvent('databases.[databaseId].collections.[collectionId].documents.update')
```
abnegate: Since we only have one event for this route, we can set it directly with a route label. We only need to set it manually in createDocuments(s), because it could be single or bulk in the same route

ugurcany:
@ArnabChatterjee20k shouldn't it also trigger the individual document events?

- databases.[databaseId].collections.[collectionId].documents.[docId].create
- databases.[databaseId].collections.[collectionId].documents.[docId].update
- databases.[databaseId].collections.[collectionId].documents.[docId].delete

src/Appwrite/Messaging/Adapter/Realtime.php
```diff
@@ -311,11 +311,16 @@ public static function fromPayload(string $event, Document $payload, Document $p
...
-                    $channels[] = 'databases.' . $database->getId() . '.collections.' . $payload->getAttribute('$collectionId') . '.documents.' . $payload->getId();
-
+                    $channels[] = 'databases.' . $database->getId() .  '.collections.' . $payload[0]->getAttribute('$collectionId') . '.documents';
+                    if (count($payload) === 1) {
+                        $channels[] = 'databases.' . $database->getId() . '.collections.' . $payload[0]->getAttribute('$collectionId') . '.documents.' . $payload[0]->getId();
+                    }
+                    $payloadReads = [];
+                    foreach ($payload as $document) {
+                        $payloadReads = \array_merge($payloadReads, $document->getRead());
+                    }
```
abnegate: I dont' think we can do it like this 🤔 we need to make we only send document that a user can read, on that users connection. So here we probably need $roles to be a map of $documentId -> (collection + document permissions) then on the sending side, we need to check the permissions of each document against the user

src/Appwrite/Messaging/Adapter/Realtime.php
```diff
@@ -337,6 +343,7 @@ public static function fromPayload(string $event, Document $payload, Document $p
 
             case 'functions':
                 if ($parts[2] === 'executions') {
+                    $payload = $payload[0];
```
abnegate: Won't we need this in other switch cases too?

src/Appwrite/Messaging/Adapter/Realtime.php
ArnabChatterjee20k: @abnegate is it due to the fact that bulk api is server side only? Cause the current realtime working by checking the roles by checking the read permission of the document only?

src/Appwrite/Messaging/Adapter/Realtime.php
ArnabChatterjee20k: Made new changes for this , now the checking only happens only inside the databases only case only

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
```diff
@@ -279,4 +280,51 @@ final protected function processDocument(
...
+     * @param Document $database
+     * @param Document $collection
+     * @param Document[] $documents
+     * @param Event $queueForEvents
+     * @param Event $queueForRealtime
+     * @param Event $queueForFunctions
+     * @param Event $queueForWebhooks
+     * @return void
+     */
+    final protected function triggerQueuesForBulkDocuments(string $event, Document $database, Document $collection, array $documents, Event $queueForEvents, Event $queueForRealtime, Event $queueForFunctions, Event $queueForWebhooks)
```
abnegate: No need to make it final, we might want to extend in a child class later

abnegate:
```suggestion
   protected function triggerBulk(
        string $event,
        Document $database,
        Document $collection,
        array $documents,
        Event $queueForEvents,
        Event $queueForRealtime,
        Event $queueForFunctions,
        Event $queueForWebhooks
    )
```

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Bulk/Update.php
```diff
@@ -77,10 +78,14 @@ public function __construct()
             ->inject('dbForProject')
             ->inject('queueForStatsUsage')
             ->inject('plan')
+            ->inject('queueForEvents')
+            ->inject('queueForRealtime')
+            ->inject('queueForFunctions')
+            ->inject('queueForWebhooks')
```
abnegate: Let's move plan to the end so all queue injections are grouped

tests/e2e/Scopes/Scope.php
```diff
@@ -182,9 +182,9 @@ public function getRoot(): array
     /**
      * @return array
      */
-    public function getUser(): array
+    public function getUser(bool $override = false): array
```
abnegate: Let's call this param `$fresh` for consistency with the same pattern elsewhere

tests/e2e/Services/Realtime/RealtimeCustomClientTest.php
abnegate, line 1: These checks are quite verbose and a lot looks duplicated, can we receive in a loop for each event type (create, update, delete) instead?

ArnabChatterjee20k: Here it is intentional since the number of documents getting received are different for each client, so asserting separately and verbosely

abnegate (approved):

ArnabChatterjee20k:
> @ArnabChatterjee20k shouldn't it also trigger the individual document events?
> 
> * databases.[databaseId].collections.[collectionId].documents.[docId].create
> * databases.[databaseId].collections.[collectionId].documents.[docId].update
> * databases.[databaseId].collections.[collectionId].documents.[docId].delete

yes

PR #10098, opened 2025-07-03
fix: JWT authentication database selection for admin mode

arielweinberger:
## What does this PR do?

Use dbForPlatform for JWT user lookup in admin mode, instead of dbForProject

## Test Plan

Tested with @TorstenDittmann and works

## Related PRs and Issues

_None_

## Checklist

### DIFF
--- a/app/init/resources.php
+++ b/app/init/resources.php
@@ -222,7 +222,9 @@ App::setResource('user', function ($mode, $project, $console, $request, $respons
     Auth::$unique = $session['id'] ?? '';
     Auth::$secret = $session['secret'] ?? '';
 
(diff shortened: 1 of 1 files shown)


app/init/resources.php
```diff
@@ -264,7 +264,11 @@
 
         $jwtUserId = $payload['userId'] ?? '';
         if (!empty($jwtUserId)) {
-            $user = $dbForProject->getDocument('users', $jwtUserId);
+            if (APP_MODE_ADMIN !== $mode) {
```
TorstenDittmann: Maybe just a nitpick, can we invert the condition and check for "if(mode === admin)?

PR #10102, opened 2025-07-04
Only delete schedule document if execution is successful

vermakhushboo:
## What does this PR do?

If task-scheduler goes down for sometime and comes back up later, it was never executing schedules that were supposed to be executed during downtime. This PR fixes that and ensures that `schedule` document is only deleted if execution is successful.

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Tasks/ScheduleExecutions.php
+++ b/src/Appwrite/Platform/Tasks/ScheduleExecutions.php
@@ -62,7 +62,7 @@ class ScheduleExecutions extends ScheduleBase
                 $executionDelay = $executedAt->getTimestamp() - $scheduledAt->getTimestamp();
 
                 $headers = $data['headers'] ?? [];
-                $headers['x-appwrite-schedule-delay'] = (string)$executionDelay;
+                $headers['x-appwrite-execution-delay'] = (string)$executionDelay;
                 $headers['x-appwrite-scheduled-at'] = $scheduledAt->format('Y-m-d\TH:i:s.v\Z');
                 $headers['x-appwrite-executed-at'] = $executedAt->format('Y-m-d\TH:i:s.v\Z');
 
--- a/tests/e2e/Services/FunctionsSchedule/FunctionsScheduleTest.php
+++ b/tests/e2e/Services/FunctionsSchedule/FunctionsScheduleTest.php
@@ -136,30 +136,17 @@ class FunctionsScheduleTest extends Scope
             $this->assertStringContainsString('path-is-/custom-path', $execution['body']['logs']);
             $this->assertStringContainsString('user-is-' . $this->getUser()['$id'], $execution['body']['logs']);
             $this->assertStringContainsString('jwt-is-valid', $execution['body']['logs']);
+            $this->assertStringContainsString('execution-delay-is-valid', $execution['body']['logs']);
+            $this->assertStringContainsString('scheduled-at-is-valid', $execution['body']['logs']);
+            $this->assertStringContainsString('executed-at-is-valid', $execution['body']['logs']);
             $this->assertGreaterThan(0, $execution['body']['duration']);
 
-            $logs = explode("\n", $execution['body']['logs']);
-            $scheduleDelay = null;
-            $scheduledAt = null;
-            $executedAt = null;
-
-            foreach ($logs as $line) {
-                $line = trim($line);
-
-                if (str_starts_with($line, 'schedule-delay-is-')) {
-                    $scheduleDelay = (int) substr($line, strlen('schedule-delay-is-'));
-                } elseif (str_starts_with($line, 'scheduled-at-is-')) {
-                    $scheduledAt = substr($line, strlen('scheduled-at-is-'));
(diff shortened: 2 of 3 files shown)


src/Appwrite/Platform/Tasks/ScheduleExecutions.php
```diff
@@ -65,10 +60,10 @@ protected function enqueueResources(Group $pools, Database $dbForPlatform, calla
```
Meldiron, line 55: $delay = max_max($deleay, 0)

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
```diff
@@ -52,42 +50,54 @@ protected function enqueueResources(Group $pools, Database $dbForPlatform, calla
...
+            \go(function () use ($queueForFunctions, $schedule, $scheduledAt, $delay, $data, $dbForPlatform) {
                 Co::sleep($delay);
 
-                $queueForFunctions->setType('schedule')
+                $executedAt = new \DateTime();
+                $actualDelay = $executedAt->getTimestamp() - $scheduledAt->getTimestamp();
+
+                // Add headers for delayed executions
+                $headers = $data['headers'] ?? [];
+                if ($actualDelay > 0) {
```
Meldiron: Let's remove this `if`, let's always have the new headers

Meldiron: Let's also ensure with no delay, scheduled-at and executed-at headers are same, and delay is 0.

For this, let's also update test. test function code would log values of these 3 headers, and in test we then look into execution logs, and ensure expected values (when no delay happens)

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
```diff
@@ -52,42 +50,54 @@ protected function enqueueResources(Group $pools, Database $dbForPlatform, calla
...
 
             $this->updateProjectAccess($schedule['project'], $dbForPlatform);
 
-            \go(function () use ($queueForFunctions, $schedule, $scheduledAt, $delay, $data) {
+            \go(function () use ($queueForFunctions, $schedule, $scheduledAt, $delay, $data, $dbForPlatform) {
                 Co::sleep($delay);
 
-                $queueForFunctions->setType('schedule')
+                $executedAt = new \DateTime();
+                $actualDelay = $executedAt->getTimestamp() - $scheduledAt->getTimestamp();
```
Meldiron: Let's rename actualDelay to something else. We might want to call them relativeDelay and absoluteDelay. relative for coroutines, absolute for execution.
Or coroutineDelay and executionDelay

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
```diff
@@ -52,42 +50,51 @@ protected function enqueueResources(Group $pools, Database $dbForPlatform, calla
...
 
-            $delay = $scheduledAt->getTimestamp() - (new \DateTime())->getTimestamp();
+            $delay = floor($scheduledAt->getTimestamp() - (new \DateTime())->getTimestamp());
+            $delay = max($delay, 0);
 
             $this->updateProjectAccess($schedule['project'], $dbForPlatform);
 
-            \go(function () use ($queueForFunctions, $schedule, $scheduledAt, $delay, $data) {
+            \go(function () use ($queueForFunctions, $schedule, $scheduledAt, $delay, $data, $dbForPlatform) {
                 Co::sleep($delay);
```
Meldiron, line 59: Might be good to wrap this in `if delay > 0`

tests/e2e/Services/FunctionsSchedule/FunctionsScheduleTest.php
```diff
@@ -137,6 +137,29 @@ public function testCreateScheduledAtExecution(): void
...
+                $line = trim($line);
+
+                if (str_starts_with($line, 'schedule-delay-is-')) {
+                    $scheduleDelay = (int) substr($line, strlen('schedule-delay-is-'));
+                } elseif (str_starts_with($line, 'scheduled-at-is-')) {
+                    $scheduledAt = substr($line, strlen('scheduled-at-is-'));
+                } elseif (str_starts_with($line, 'executed-at-is-')) {
+                    $executedAt = substr($line, strlen('executed-at-is-'));
+                }
+            }
```
Meldiron: Instead of parsing, I would write tests like these:

1. Read request headers, ensure there are 3 headers
1.1. scheduled and executed at header to be valid datetime (parsing with datetine will not be false)
2. build strings in test, how function would build it, using value from request headers
2. Assert logs contain exact string, somethling like `"executed-at-is-$executedAt"`

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
vermakhushboo: I don't think we should do that as the flow needs to be executed even when delay is 0. And just above this, we anyway handle negative delays.

PR #10103, opened 2025-07-04
Removed $sequence from getting returned from response

ArnabChatterjee20k:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Utopia/Response/Model/Document.php
+++ b/src/Appwrite/Utopia/Response/Model/Document.php
@@ -36,12 +36,6 @@ class Document extends Any
                 'default' => '',
                 'example' => '5e5ea5c16897e',
             ])
-            ->addRule('$sequence', [
-                'type' => self::TYPE_INTEGER,
-                'description' => 'Document automatically incrementing ID.',
-                'default' => 0,
-                'example' => 1,
-            ])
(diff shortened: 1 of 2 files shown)


abnegate:
We can add a new response filter like this: https://github.com/appwrite/appwrite/blob/1.7.x/src/Appwrite/Utopia/Response/Filters/V19.php called V20, and remove $sequence from responses there.

Then, we should register it to only run on versions less than 1.8.0, over here: https://github.com/appwrite/appwrite/blob/1.7.x/app/controllers/general.php#L975

ArnabChatterjee20k:
> We can add a new response filter like this: https://github.com/appwrite/appwrite/blob/1.7.x/src/Appwrite/Utopia/Response/Filters/V19.php called V20, and remove $sequence from responses there.
> 
> Then, we should register it to only run on versions less than 1.8.0, over here: https://github.com/appwrite/appwrite/blob/1.7.x/app/controllers/general.php#L975

added the new pr here cause this one causing much merge conflicts due 1.7.x, https://github.com/appwrite/appwrite/pull/10115

ArnabChatterjee20k:
pr -> https://github.com/appwrite/appwrite/pull/10115

PR #10129, opened 2025-07-09
Changed the default permission param of upsert document

ArnabChatterjee20k:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/controllers/api/databases.php
+++ b/app/controllers/api/databases.php
@@ -4254,7 +4254,7 @@ App::put('/v1/databases/:databaseId/collections/:collectionId/documents/:documen
     ->param('collectionId', '', new UID(), 'Collection ID.')
     ->param('documentId', '', new CustomId(), 'Document ID.')
     ->param('data', [], new JSON(), 'Document data as JSON object. Include all required attributes of the document to be created or updated.')
-    ->param('permissions', [], new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE, [Database::PERMISSION_READ, Database::PERMISSION_UPDATE, Database::PERMISSION_DELETE, Database::PERMISSION_WRITE]), 'An array of permissions strings. By default, the current permissions are inherited. [Learn more about permissions](https://appwrite.io/docs/permissions).', true)
+    ->param('permissions', null, new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE, [Database::PERMISSION_READ, Database::PERMISSION_UPDATE, Database::PERMISSION_DELETE, Database::PERMISSION_WRITE]), 'An array of permissions strings. By default, the current permissions are inherited. [Learn more about permissions](https://appwrite.io/docs/permissions).', true)
     ->inject('requestTimestamp')
     ->inject('response')
     ->inject('dbForProject')
@@ -4309,7 +4309,7 @@ App::put('/v1/databases/:databaseId/collections/:collectionId/documents/:documen
         }
 
         $data['$id'] = $documentId;
-        $data['$permissions'] = $permissions;
+        $data['$permissions'] = $permissions ?? [];
         $newDocument = new Document($data);
 
         $operations = 0;
--- a/app/controllers/api/databases.php
+++ b/app/controllers/api/databases.php
@@ -4254,7 +4254,7 @@ App::put('/v1/databases/:databaseId/collections/:collectionId/documents/:documen
     ->param('collectionId', '', new UID(), 'Collection ID.')
     ->param('documentId', '', new CustomId(), 'Document ID.')
     ->param('data', [], new JSON(), 'Document data as JSON object. Include all required attributes of the document to be created or updated.')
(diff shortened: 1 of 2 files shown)


ItzNotABug (approved):

Meldiron (approved):

app/controllers/api/databases.php
```diff
@@ -4254,7 +4254,7 @@ function updateAttribute(
     ->param('collectionId', '', new UID(), 'Collection ID.')
     ->param('documentId', '', new CustomId(), 'Document ID.')
     ->param('data', [], new JSON(), 'Document data as JSON object. Include all required attributes of the document to be created or updated.')
-    ->param('permissions', null, new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE, [Database::PERMISSION_READ, Database::PERMISSION_UPDATE, Database::PERMISSION_DELETE, Database::PERMISSION_WRITE]), 'An array of permissions strings. By default, the current permissions are inherited. [Learn more about permissions](https://appwrite.io/docs/permissions).', true)
+    ->param('permissions', [], new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE, [Database::PERMISSION_READ, Database::PERMISSION_UPDATE, Database::PERMISSION_DELETE, Database::PERMISSION_WRITE]), 'An array of permissions strings. By default, the current permissions are inherited. [Learn more about permissions](https://appwrite.io/docs/permissions).', true)
```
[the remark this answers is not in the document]
ItzNotABug: @ArnabChatterjee20k does this makes sense here? Can you verify?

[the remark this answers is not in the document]
stnguyen90: Ya...I'm not sure about this change. We need to be able to differentiate between:

1. developer passing `[]` because they want to remove permissions 
2. developer omitting permissions because they want the default

If we change the default to `[]`, how do we differentiate the two?

tests/e2e/Services/Databases/DatabasesBase.php
```diff
@@ -1996,6 +1996,48 @@ public function testUpsertDocument(array $data): void
...
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+        ], $this->getHeaders()));
+
+        if (isset($document['body']['$permissions'])) {
+            $this->assertCount(3, $document['body']['$permissions']);
+            $this->assertContains(Permission::read(Role::users()), $document['body']['$permissions']);
+            $this->assertContains(Permission::update(Role::users()), $document['body']['$permissions']);
+            $this->assertContains(Permission::delete(Role::users()), $document['body']['$permissions']);
+        }
```
stnguyen90: Please fix the PR so the tests pass.

app/controllers/api/databases.php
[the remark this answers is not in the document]
ArnabChatterjee20k: @stnguyen90 as told in the thread in discord this would help developers doesn't getting error if not passing permission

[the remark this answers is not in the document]
ArnabChatterjee20k: > @ArnabChatterjee20k does this makes sense here? Can you verify?

@ItzNotABug as mentioned in the thread for the current implemention of the lib its not possible to differentiate between create and update. So the solution for the current scope was to omit null to [] so that alteast devs doesn't get error if making request without permissions which is a bad as now user just lost the docs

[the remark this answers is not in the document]
stnguyen90: > @stnguyen90 as told in the thread in discord this would help developers doesn't getting error if not passing permission

@ArnabChatterjee20k is there another way to handle the problem without changing the API signature?

tests/e2e/Services/Databases/DatabasesBase.php
stnguyen90: We still want a test to make sure upserting w/o permissions works. We just don't need to check the permissions right now.

ArnabChatterjee20k: @stnguyen90 but that will break the test as it would fail testCreateUnique index as it has some dependencies on the collection. And for retaining the state we have to delete the w/o permissions doc but that is not possible as we dont have the permissions

stnguyen90: @ArnabChatterjee20k you can either:

1. make document that won't break the index
2. use an api key to delete

The important part is having a test. We have this problem now because we didn't have a test.

app/controllers/api/databases.php
```diff
@@ -4309,7 +4309,7 @@ function updateAttribute(
         }
 
         $data['$id'] = $documentId;
-        $data['$permissions'] = $permissions;
+        $data['$permissions'] = $permissions ?? null;
```
stnguyen90: So what happens if someone doesn't pass `$permissions`?

ArnabChatterjee20k: document becomes unaccessible as mentioned earlier in the thread

ArnabChatterjee20k: thanks for pointing, dumps pushed while reverting the tests, reverting it out

ArnabChatterjee20k: @stnguyen90 updated this along with the tests,
if no permissions -> no error
expected behaviour -> document becomes unaccessible as mentioned earlier in the thread
reason -> no current ways to reconcile/decide which permission to use

tests/e2e/Services/Databases/DatabasesBase.php
ArnabChatterjee20k: updated

PR #10164, opened 2025-07-18
chore: improve invalid scheme error in origin check

ChiragAgg5k:
## What does this PR do?

when an unsupported scheme is used, we show this error
Screenshot 2025-07-18 at 10 09 30 PM

this lacks the platform user should be adding since the scheme itself is not supported. so if platform is not present, we throw an improved error showing that scheme is unsupported

## Checklist

### DIFF
--- a/tests/unit/Network/Validators/OriginTest.php
+++ b/tests/unit/Network/Validators/OriginTest.php
@@ -108,6 +108,6 @@ class OriginTest extends TestCase
         $this->assertEquals('Invalid Origin. Register your new client (com.company.appname) as a new Web (Edge Extension) platform on your project console dashboard', $validator->getDescription());
 
         $this->assertEquals(false, $validator->isValid('random-scheme://localhost'));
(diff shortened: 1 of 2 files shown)


tests/unit/Network/Validators/OriginTest.php
```diff
@@ -106,5 +106,8 @@ public function testValues(): void
 
         $this->assertEquals(false, $validator->isValid('ms-browser-extension://com.company.appname'));
         $this->assertEquals('Invalid Origin. Register your new client (com.company.appname) as a new Web (Edge Extension) platform on your project console dashboard', $validator->getDescription());
+
+        $this->assertEquals(false, $validator->isValid('random-scheme://localhost'));
+        $this->assertEquals('Invalid Scheme. The scheme used ( random-scheme ) in the Origin ( random-scheme://localhost ) is not supported. If you are using a custom scheme, please change it to `appwrite-callback-<PROJECT_ID>`', $validator->getDescription());
```
stnguyen90: Let's remove the space padding to match the existing format.

```suggestion
        $this->assertEquals('Invalid Scheme. The scheme used (random-scheme) in the Origin (random-scheme://localhost) is not supported. If you are using a custom scheme, please change it to `appwrite-callback-<PROJECT_ID>`', $validator->getDescription());
```

PR #10174, opened 2025-07-21
fix: sanitize headers to prevent 'null' key in scheduled executions

HarshMN2345:
## What does this PR do?

This PR fixes an issue where scheduled function executions could include a header with the key "null" (or other invalid keys) in the request headers. It does this by sanitizing the headers before passing them to the execution, ensuring only valid, non-empty, string keys are included. Additionally, a test is added to verify that no "null" or empty header keys are present in scheduled executions.

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


Meldiron:
Fixed on frontend, closing this one to allow developers to intentionally set `null` header (not ideal, but not our decision to block it)

PR #10181, opened 2025-07-21
fix: Use direct source for file-preview when empty

hmacr:
## What does this PR do?

Fixes a bug in `file-preview` API where the requested `screenshots` preview couldn't be loaded

## Test Plan

Verified the API returns the requested image + format

## Related PRs and Issues

Pre-requisite for https://github.com/appwrite/console/pull/2140


app/controllers/api/storage.php
```diff
@@ -1072,6 +1072,10 @@
 
         $decompressionTime = \microtime(true) - $startTime - $downloadTime - $decryptionTime;
 
+        if (empty($source)) {
+            $source = $deviceForFiles->read($path);
+        }
+
```
Meldiron: A bit scared to do storage API change due to sites screenshot bug.
I would guess this scenario can never happen, otherwise we would have already fixed it here - it must be sites specific problem. So I would guess sites just dont create file document properly.

Please look into Builds.php (worker), and see how screenshot document is created. My guess would be, that there we dont set 'source' parameter. Compare that to `createFile` endpoint, here in storage.php, to see what expected value is. With that knowledge you should be able to have fix be done in Builds.php worker, and so we can then remove this fix in getFilePreview.

^ If this works, I would prefer that. if it doesn't, let's keep this fix, and I will ask other leads for review, to make sure this is safe to do.

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
```diff
@@ -1005,10 +1007,10 @@ protected function buildDeployment(
                             'name' => $fileName,
                             'path' => $path,
                             'signature' => $deviceForFiles->getFileHash($path),
-                            'mimeType' => $deviceForFiles->getFileMimeType($path),
+                            'mimeType' => $mimeType,
                             'sizeOriginal' => \strlen($screenshot),
                             'sizeActual' => $deviceForFiles->getFileSize($path),
-                            'algorithm' => Compression::GZIP,
+                            'algorithm' => Compression::NONE,
```
Meldiron, line 1013: Update tests to also ensure /preview works (and also fails without permission)

hmacr: Tests have been added

Meldiron (approved):

PR #10187, opened 2025-07-22
Skip deployment when commit is created by us

hmacr:
## What does this PR do?

Skip triggering a deployment in webhooks flow when commit is created by us.

## Test Plan

- Verified double deployments aren't triggered.
- Ensured new commits are created by the same email and user.

## Related PRs and Issues

- https://github.com/utopia-php/vcs/pull/45

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -73,7 +73,7 @@
         "utopia-php/swoole": "0.8.*",
         "utopia-php/system": "0.9.*",
         "utopia-php/telemetry": "0.1.*",
-        "utopia-php/vcs": "0.10.*",
+        "utopia-php/vcs": "0.11.*",
         "utopia-php/websocket": "0.3.*",
         "matomo/device-detector": "6.1.*",
         "dragonmantank/cron-expression": "3.3.*",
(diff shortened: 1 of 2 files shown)


Meldiron (commented):
LGTM, we just need to update version in composer.json once we release DB lib version

app/controllers/api/vcs.php
Meldiron, line 1: Composer.json will need 0.11 of vcs

app/controllers/api/vcs.php
```diff
@@ -1234,7 +1234,8 @@ function (GitHub $github, Request $request, Response $response, Database $dbForP
                 $providerRepositoryUrl = $parsedPayload["repositoryUrl"] ?? '';
                 $providerCommitHash = $parsedPayload["commitHash"] ?? '';
                 $providerRepositoryOwner = $parsedPayload["owner"] ?? '';
-                $providerCommitAuthor = $parsedPayload["headCommitAuthor"] ?? '';
```
Meldiron, line 1237: Double-check we dont have `headCommitAuthor` anywhere else.

hmacr: Verified 👍

app/controllers/api/vcs.php
hmacr: Composer has been updated.

PR #10199, opened 2025-07-23
Suppress git-action exception in deployment worker

hmacr:
## What does this PR do?

Ensure failed git actions don't fail the whole deployment

## Test Plan

It's almost impossible to simulate an organic flow, so did a manual test by throwing an ad-hoc exception in `runGitAction` function and ensured warning logs are added + deployment continues.

### DIFF
--- a/src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
+++ b/src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
@@ -516,7 +516,7 @@ class Builds extends Action
                     ->setPayload($deployment->getArrayCopy())
                     ->trigger();
 
-                $this->runGitAction('processing', $github, $providerCommitHash, $owner, $repositoryName, $project, $resource, $deployment->getId(), $dbForProject, $dbForPlatform);
+                $this->runGitAction('processing', $github, $providerCommitHash, $owner, $repositoryName, $project, $resource, $deployment->getId(), $dbForProject, $dbForPlatform, $queueForRealtime);
             }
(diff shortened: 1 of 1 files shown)


src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
```diff
@@ -1441,96 +1441,109 @@ protected function runGitAction(
...
 
-                $comment = new Comment();
-                $comment->parseComment($github->getComment($owner, $repositoryName, $commentId));
-                $comment->addBuild($project, $resource, $resourceType, $status, $deployment->getId(), ['type' => 'logs'], $previewUrl);
-                $github->updateComment($owner, $repositoryName, $commentId, $comment->generateComment());
-            } finally {
-                $dbForPlatform->deleteDocument('vcsCommentLocks', $commentId);
-            }
+            $deployment->setAttribute('buildLogs', $logs);
+            $deployment = $dbForProject->updateDocument('deployments', $deployment->getId(), $deployment);
```
Meldiron, line 1547: 1 tiny detail, we need to trigger realtime here, because logs changed. like here https://github.com/appwrite/appwrite/blob/cd39bd65125238564360d56c297cce8e6010c27e/src/Appwrite/Platform/Modules/Functions/Workers/Builds.php#L1035-L1037

hmacr: Updated the PR to send the logs to queue.

PR #10227, opened 2025-07-29
chore: filter certificates renewal task in maintenance by region

ChiragAgg5k:
## What does this PR do?

certificates should be handled by the worker of the region where its issued

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Tasks/Maintenance.php
+++ b/src/Appwrite/Platform/Tasks/Maintenance.php
@@ -122,6 +122,15 @@ class Maintenance extends Action
             Console::info("[{$time}] Found " . \count($certificates) . " certificates for renewal, scheduling jobs.");
 
             foreach ($certificates as $certificate) {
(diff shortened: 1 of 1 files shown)


src/Appwrite/Platform/Tasks/Maintenance.php
```diff
@@ -122,6 +122,14 @@ private function renewCertificates(Database $dbForPlatform, Certificate $queueFo
             Console::info("[{$time}] Found " . \count($certificates) . " certificates for renewal, scheduling jobs.");
 
             foreach ($certificates as $certificate) {
+                $rule = $dbForPlatform->findOne('rules', [
+                    Query::equal('domain', [$certificate->getAttribute('domain')]),
```
loks0n: I think rules can be queried by md5 hash of domain also in case of some env var, can you confirm?

abnegate (approved):

Issue #10225, opened 2025-07-29
🐛 Bug Report: appwrite build args


sb-ranjha:
### 👟 Reproduction steps

Services.appwrite.build.args contain false, which is an invalid type 

### 👍 Expected behavior

```
build:
  context: .
  args :
    DEBUG: "false"
    TESTING: "true"
    VERSION: "dev"
```

### 👎 Actual Behavior

```
build:
  context: .
  args:
    DEBUG: false
    TESTING: true
    VERSION: dev
```

### 🎲 Appwrite version

Version 1.6.x

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

Adityakk9031:
I would like to work on this issue

sb-ranjha:
1. Hi Aditya,

Thanks for your interest in contributing. We appreciate you offering to
help out.

Best,
Shoaib

eldadfux:
Stale issue, closing down.

PR #10238, opened 2025-07-31
Fix: Convert Docker build args from boolean to string values

imkuldeepahlawat:
## What does this PR do?

This PR fixes the issue where Docker build args were defined as boolean values instead of strings in YAML files, which caused invalid type errors during Docker builds.

**Problem:**
- Docker build args in YAML must be strings, not boolean values
- This was causing `Services.appwrite.build.args contain false, which is an invalid type` errors

**Solution:**
- Convert boolean values to strings in Docker Compose files
- Ensure consistency with GitHub workflow files which already had correct format

**Files Modified:**
- `docker-compose.yml` - Fixed build args format
- `tests/resources/docker/docker-compose.yml` - Fixed build args format

**Before:**
```yaml
args:
  DEBUG: false      # ❌ Boolean value
  TESTING: true     # ❌ Boolean value
  VERSION: dev      # ❌ Missing quotes
```

**After:**
```yaml
args:
  DEBUG: "false"    # ✅ String value
  TESTING: "true"   # ✅ String value
  VERSION: "dev"    # ✅ String value
```

## Test Plan

1. **Validation Test:** Run `docker-compose config` to verify YAML syntax is valid
2. **Build Test:** Docker builds now complete successfully without type errors
3. **Consistency Check:** Build args format now matches GitHub workflow files

## Related PRs and Issues

- Fixes issue #10225: `Services.appwrite.build.args contain false, which is an invalid type`

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,4 @@
 
-> We just announced Auto-increment support for Appwrite Databases - [Learn more](https://appwrite.io/blog/post/announcing-auto-increment-support)
-
 > [Get started with Appwrite](https://apwr.dev/appcloud)
 
 <br />


HarshMN2345:

PR #10249, opened 2025-08-01
feat: add model examples + additonal examples to specs

ChiragAgg5k:
## What does this PR do?

Adds models examples as per swagger/openapi specs.

## Test Plan
Screenshot 2025-08-01 at 1 18 17 PM

## Checklist

### DIFF
--- a/src/Appwrite/Utopia/Response/Model/Any.php
+++ b/src/Appwrite/Utopia/Response/Model/Any.php
@@ -31,4 +31,14 @@ class Any extends Model
     {
         return Response::MODEL_ANY;
     }
+
+    /**
+     * Get "any" examples
+     *
+     * @return array
+     */
+    public function getAnyExamples(): array
+    {
+        return [];
+    }
 }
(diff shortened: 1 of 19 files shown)


src/Appwrite/Utopia/Response/Model/Any.php
```diff
@@ -31,4 +31,14 @@ public function getType(): string
     {
         return Response::MODEL_ANY;
     }
+
+    /**
+     * Get "any" examples
+     *
+     * @return array
+     */
+    public function getAnyExamples(): array
```
abnegate: Let's call this `getSampleData`

src/Appwrite/Utopia/Response/Model/Document.php
```diff
@@ -95,4 +95,15 @@ public function filter(DatabaseDocument $document): DatabaseDocument
 
         return $document;
     }
+
+    public function getAnyExamples(): array
+    {
```
abnegate: Let's implement this in the `Preferences` model too

christyjacob4 (approved):

PR #10255, opened 2025-08-03
chore: add support for svg favicons

ChiragAgg5k:
## What does this PR do?

adds missing svg support to favicons endpoint

## Test Plan

Screenshot 2025-08-03 at 1 25 01 PM

Screenshot 2025-08-03 at 1 25 09 PM

## Checklist

### DIFF
--- a/app/controllers/api/avatars.php
+++ b/app/controllers/api/avatars.php
@@ -362,7 +362,8 @@ App::get('/v1/avatars/favicon')
         $client = new Client();
         try {
             $res = $client
-                ->setAllowRedirects(false)
+                ->setAllowRedirects(true)
+                ->setMaxRedirects(5)
                 ->setUserAgent(\sprintf(
                     APP_USERAGENT,
                     System::getEnv('_APP_VERSION', 'UNKNOWN'),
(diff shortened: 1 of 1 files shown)


tests/e2e/Services/Avatars/AvatarsBase.php
```diff
@@ -284,7 +284,17 @@ public function testGetFavicon(): array
         ]);
 
         $this->assertEquals(200, $response['headers']['status-code']);
-        $this->assertEquals('image/x-icon', $response['headers']['content-type']);
+        $this->assertEquals('image/svg+xml', $response['headers']['content-type']);
```
ChiragAgg5k, line 287: This had to be changed since for "github.com" actual icon is in svg format, but since we did not support it we had a fallback check for `/favicon.ico`

app/controllers/api/avatars.php
```diff
@@ -449,14 +458,32 @@
...
             $response
                 ->addHeader('Cache-Control', 'private, max-age=2592000') // 30 days
                 ->setContentType('image/x-icon')
                 ->file($data);
+            return;
+        }
+
+        if ('svg' === $outputExt) { // Skip crop, Imagick isn\'t supporting svg files
+            $sanitizer = new SvgSanitizer();
+            $cleanSvg = $sanitizer->sanitize($data);
```
loks0n, line 479: Should we `->minify(true)` also?

ChiragAgg5k: yeah makes sense, added it 👌

PR #10258, opened 2025-08-03
Check CAA record in DNSTest

vermakhushboo:
## What does this PR do?
Verify domain has expected CAA record

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


composer.json
```diff
@@ -55,6 +55,7 @@
         "utopia-php/database": "0.71.*",
         "utopia-php/detector": "0.1.*",
         "utopia-php/domains": "0.8.*",
+        "utopia-php/dns": "dev-feat-add-CAA-to-client as 0.2.99",
```
vermakhushboo: Update after release

src/Appwrite/Network/Validator/DNS.php
```diff
@@ -69,6 +69,20 @@ public function isValid($value): bool
         }
 
         foreach ($query as $record) {
+            // CAA validation only needs to ensure domain
+            if ($this->type === self::RECORD_CAA) {
+                // Original: 255 issuewild "certainly.com;validationmethods=tls-alpn-01;retrytimeout=3600"
+                // Extracted: certainly.com
+                $rdata = $record->getRdata();
+                $rdata = \explode(' ', $rdata, 3)[2] ?? '';
```
vermakhushboo: @Meldiron can we add some comments to explain this?

app/config/variables.php
```diff
@@ -151,6 +151,24 @@
                 'question' => '',
                 'filter' => ''
             ],
+            [
+                'name' => '_APP_DOMAIN_TARGET_CAA',
+                'description' => 'A CAA record domain that can be used to validate custom domains. Value should be domain\'s hostname.',
+                'introduction' => '',
+                'default' => '',
```
vermakhushboo, line 158: We should add an example here

src/Appwrite/Network/Validator/DNS.php
```diff
@@ -42,46 +45,64 @@ public function getLogs(): mixed
...
             return false;
         }
 
-        foreach ($records as $record) {
-            if (isset($record[$dnsKey]) && $record[$dnsKey] === $this->target) {
+        $caaCount = 0;
+
+        foreach ($query as $record) {
+            if ($record->getTypeName() === self::RECORD_CAA) {
+                $caaCount++;
```
vermakhushboo: Didn't understand why we need count

app/config/variables.php
Meldiron: I couldnt find example attribute in other env vars, so I explained it here `Value should be domain\'s hostname.`. That should be clear enough, considering every certificate issuer anyway provide theirs values

src/Appwrite/Network/Validator/DNS.php
```diff
@@ -42,42 +45,49 @@ public function getLogs(): mixed
...
+                return true;
+            }
+
             return false;
         }
 
-        foreach ($records as $record) {
-            if (isset($record[$dnsKey]) && $record[$dnsKey] === $this->target) {
+        foreach ($query as $record) {
+            // CAA validation only needs to ensure domain
```
vermakhushboo, line 77: This comment should be inside if condition

src/Appwrite/Network/Validator/DNS.php
```diff
@@ -42,42 +45,49 @@ public function getLogs(): mixed
...
+                $rdata = \explode(' ', $rdata, 3)[2] ?? ''; // "certainly.com;validationmethods=tls-alpn-01;retrytimeout=3600"
+                $rdata = \trim($rdata, '"'); // certainly.com;validationmethods=tls-alpn-01;retrytimeout=3600
+                $rdata = \explode(';', $rdata, 2)[0] ?? ''; // certainly.com
+
+                if ($rdata === $this->target) {
+                    return true;
+                }
+            }
+
+            if ($record->getRdata() === $this->target) {
```
vermakhushboo, line 90: Redundant

Meldiron: This one is intentional, test also covers it.
It's for when env var for CAA doesnt only have domain, but exact value including flags and type. Domain match will fail, but exact match will pass

Meldiron: having this supported allows us to switch on cloud if we ever want to. maybe we realise type `issue` as compared to `issuewild` is important, so we include it in check.

Meldiron (approved):

PR #10277, opened 2025-08-07
Fix 810 return type for download deployment is wrong

MicahHeneveld:
## What does this PR do?

This PR resolves this [issue](https://github.com/appwrite/website/issues/810) by updating OpenAPI.php to generate specs(open api) that include a content block for 200 responses, but leaves out the content block for 204 responses.  The issue is in appwrite/website but our fix is in appwrite/appwrite which is why our PR is to appwrite/appwrite.

## Test Plan

Either use the latest specs included in this PR, or regenerate them through OpenAPI.php using this Docker command:
docker run --rm -v ${PWD}:/usr/src/code -w /usr/src/code appwrite/appwrite:1.7.4 php app/cli.php specs
Then replace the latest specs in appwrite/website with the correctly generated ones; this will fix the behavior described in the issue.

## Related PRs and Issues

This PR will fix the documentation bug described in issue 810, which happens in the functions page [here](https://appwrite.io/docs/references/cloud/server-deno/functions).  But the updated specs are not applied to the sites page [here](https://appwrite.io/docs/references/cloud/server-deno/sites), which shares the same Deployment Download response behavior as the functions page.  This is because the sites page pulls from versions 1.7.x specs instead of 'latest'.  Our team has come up with a solution to this behavior as well, but that fix is in appwrite/website, not appwrite/appwrite like this fix is.  Do you want us to open up an issue for this and then submit a PR to fix the sites behavior as well?

## Checklist

Yes

Includes updated API specs, not sure what 'example docs' refers to.

### DIFF
--- a/app/config/specs/open-api3-latest-client.json
+++ b/app/config/specs/open-api3-latest-client.json
@@ -8,7 +8,7 @@
         "contact": {
             "name": "Appwrite Team",
             "url": "https:\/\/appwrite.io\/support",
-            "email": "team@appwrite.io"
+            "email": "team@localhost.test"
         },
         "license": {
             "name": "BSD-3-Clause",
@@ -165,7 +165,7 @@
                 "tags": [
                     "account"
                 ],
-                "description": "Update currently logged in user account email address. After changing user address, the user confirmation status will get reset. A new confirmation email is not sent automatically however you can use the send confirmation email endpoint again to send the confirmation email. For security measures, user password is required to complete this request.\nThis endpoint can also be used to convert an anonymous account to a normal one, by passing an email address and a new password.\n",
+                "description": "Update currently logged in user account email address. After changing user address, the user confirmation status will get reset. A new confirmation email is not sent automatically however you can use the send confirmation email endpoint again to send the confirmation email. For security measures, user password is required to complete this request.\r\nThis endpoint can also be used to convert an anonymous account to a normal one, by passing an email address and a new password.\r\n",
                 "responses": {
                     "200": {
                         "description": "User",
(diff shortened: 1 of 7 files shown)


MicahHeneveld:
Upon further testing and learning the codebase, I found that this [Get.php](https://github.com/appwrite/appwrite/blob/main/src/Appwrite/Platform/Modules/Sites/Http/Deployments/Download/Get.php#L53) maps ContentType to ANY for the deployment download response.  From [ContentType](https://github.com/appwrite/appwrite/blob/main/src/Appwrite/SDK/ContentType.php#L14) ANY maps to "*/*" instead of the desired "application/gzip".  [Fix 2283](https://github.com/appwrite/website/pull/2697) addresses [issue 2283](https://github.com/appwrite/website/issues/2283) and includes a lightweight function that coverts the "*/*" to the desired"application/gzip" in the frontend.
A fix in the backend can be implemented instead of the workaround by adding GZIP = `application/gzip` to [ContentType](https://github.com/appwrite/appwrite/blob/main/src/Appwrite/SDK/ContentType.php#L14) and then mapping correctly in [Get.php](https://github.com/appwrite/appwrite/blob/main/src/Appwrite/Platform/Modules/Sites/Http/Deployments/Download/Get.php#L53) or possibly by changing [Get.php](https://github.com/appwrite/appwrite/blob/main/src/Appwrite/Platform/Modules/Sites/Http/Deployments/Download/Get.php#L53) to map to MULTIPART.

I would love to know what you think @stnguyen90 !!

HarshMN2345:

PR #10304, opened 2025-08-12
feat: allow HEAD requests and add test for it

ChiragAgg5k:
## What does this PR do?

updates executor to allow head requests and adds tests for it

## Test Plan

before - 
image

after -
image

## Checklist

### DIFF
--- a/tests/e2e/Client.php
+++ b/tests/e2e/Client.php
@@ -216,6 +216,19 @@ class Client
             return $len;
         });
 
+        // Special handling for HEAD requests
+        if ($method === self::METHOD_HEAD) {
+            curl_setopt($ch, CURLOPT_NOBODY, true); // This is crucial for HEAD requests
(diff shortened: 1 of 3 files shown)


tests/e2e/Client.php
```diff
@@ -216,7 +216,15 @@ public function call(string $method, string $path = '', array $headers = [], mix
...
 
-        if ($method != self::METHOD_GET) {
+
+        if ($method === self::METHOD_HEAD) {
+            curl_setopt($ch, CURLOPT_NOBODY, true); // This is crucial for HEAD requests
+            curl_setopt($ch, CURLOPT_HEADER, false);
+        } else {
+            curl_setopt($ch, CURLOPT_NOBODY, false);
+        }
+
```
loks0n, line 226: just fyi we might also need to add this into the internal CURL request within Edge

loks0n (approved):

PR #10308, opened 2025-08-12
Fix: variables sharing across resources

Meldiron:
## What does this PR do?

On dedicated tables, fixes scenario when project has 1 site and 1 function, and they unintentionally share variables.

## Test Plan

Wanted to do automated, but it would be tricky, or need to be skipepd on Cloud shared DB. so I did manual QA

Before (function having variables when I set none; they are coming from site where I set them)

CleanShot 2025-08-12 at 16 56 02@2x

After (fix + cache wipe makes function not see variables)

CleanShot 2025-08-12 at 17 00 48@2x

## Related PRs and Issues

x

## Checklist

### DIFF
--- a/app/init/database/filters.php
+++ b/app/init/database/filters.php
@@ -245,10 +245,16 @@ Database::addFilter(
         return;
     },
     function (mixed $value, Document $document, Database $database) {
+        $resourceType = match ($document->getCollection()) {
+            'functions' => ['function'],
+            'sites' => ['site'],
+            default => ['function', 'site']
+        };
+
         return $database
             ->find('variables', [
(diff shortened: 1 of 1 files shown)


app/init/database/filters.php
```diff
@@ -245,10 +245,16 @@ function (mixed $value) {
         return;
     },
     function (mixed $value, Document $document, Database $database) {
+        $resourceType = match ($document->getCollection()) {
+            'functions' => ['function'],
+            'sites' => ['site'],
+            default => ['function', 'site']
```
Meldiron, line 251: Cant decide - both here, or throw exception?

eldadfux (approved):

PR #10338, opened 2025-08-19
Chore: Upgrade HTTP

Meldiron:
## What does this PR do?

- fixes double slash 5xx errors
- improves domain validator

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/controllers/api/health.php
+++ b/app/controllers/api/health.php
@@ -23,10 +23,12 @@ use Utopia\Storage\Device;
 use Utopia\Storage\Device\Local;
 use Utopia\Storage\Storage;
 use Utopia\System\System;
+use Utopia\Validator\AnyOf;
 use Utopia\Validator\Domain;
 use Utopia\Validator\Integer;
 use Utopia\Validator\Multiple;
 use Utopia\Validator\Text;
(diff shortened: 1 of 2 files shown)


app/controllers/api/health.php
```diff
@@ -397,7 +399,7 @@
         ],
         contentType: ContentType::JSON
     ))
-    ->param('domain', null, new Multiple([new Domain(), new PublicDomain()]), Multiple::TYPE_STRING, 'Domain name')
+    ->param('domain', null, new Multiple([new AnyOf([new URL(), new Domain()]), new PublicDomain()]), Multiple::TYPE_STRING, 'Domain name')
```
abnegate, line 402: Do we need `Multiple` if we have `AnyOf`? Is it breaking to remove `PublicDomain`?

Meldiron: I would do as little changes as possible.
I think this makes it so public domain must pass, and then either url or domain needs to

Meldiron: Multiple is needed here, to do publicDOmain validator, which ensures you dont pass localhost to it, for example (we got test for it)

PR #10363, opened 2025-08-24
Fix Upsert No change

fogelito:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/databases.php
```diff
@@ -4435,6 +4435,10 @@ function updateAttribute(
             throw new Exception(Exception::DOCUMENT_INVALID_STRUCTURE, $e->getMessage());
         }
 
+        if (empty($upserted[0])) {
```
fogelito, line 4438: I think we should return a header for $modified 
that was returned from 
$modified = $dbForProject->createOrUpdateDocuments
Otherwise will not know if upserted was skipped.
Like now in this test had to add all attributes

abnegate (approved):

PR #10379, opened 2025-08-26
Add execution id and log id to response headers

vermakhushboo:
## What does this PR do?

Add `x-appwrite-execution-id` to function response headers and `x-appwrite-log-id` for sites

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Modules/Functions/Http/Executions/Create.php
+++ b/src/Appwrite/Platform/Modules/Functions/Http/Executions/Create.php
@@ -486,6 +486,8 @@ class Create extends Base
             $headers[] = ['name' => $key, 'value' => $value];
         }
 
+        $headers[] = ['name' => 'x-appwrite-execution-id', 'value' => $execution->getId()];
+
         $execution->setAttribute('responseBody', $executionResponse['body'] ?? '');
         $execution->setAttribute('responseHeaders', $headers);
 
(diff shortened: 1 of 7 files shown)


src/Appwrite/Platform/Workers/Functions.php
```diff
@@ -549,6 +549,7 @@ private function execute(
                     $headersFiltered[] = [ 'name' => $key, 'value' => $value ];
                 }
             }
+            $headersFiltered[] = ['name' => 'x-appwrite-execution-id', 'value' => $execution->getId()];
```
Meldiron: I would add it to `$executionResponse['headers']` instead, so it doesnt have weird side-effects. We want it to behave like any other header

src/Appwrite/Platform/Modules/Functions/Http/Executions/Create.php
```diff
@@ -486,6 +486,8 @@ public function action(
             $headers[] = ['name' => $key, 'value' => $value];
         }
 
+        $headers[] = ['name' => 'x-appwrite-execution-id', 'value' => $execution->getId()];
```
Meldiron: I would add it to `$executionResponse['headers']` instead, so it doesnt have weird side-effects. We want it to behave like any other header

tests/e2e/Services/Functions/FunctionsCustomClientTest.php
Meldiron, line 1: I would expect one test in functions to look similar to site's one - checking actual http response header, from domain execution

PR #10380, opened 2025-08-26
Feat: Deployment select queries

Meldiron:
## What does this PR do?

adds support for query:.seelct() to deployments, making it more optimal on Console where list doesnt need logs (logs can get large, 1MB, for some SSR builds)

## Test Plan

- [x] new tests

- [x] Quick QA with COnsole

CleanShot 2025-08-26 at 15 33 25@2x

CleanShot 2025-08-26 at 15 33 21@2x

(previously this was 4xx query error, select not supported)

## Related PRs and Issues

x

## Checklist


src/Appwrite/Utopia/Database/Validator/Queries/Base.php
```diff
@@ -84,6 +96,15 @@ public function __construct(string $collection, array $allowedAttributes)
...
+        if ($this->isSelectQueryAllowed()) {
+            $validators[] = new Select($allAttributes);
+        }
+
         parent::__construct($validators);
     }
+
+    public function isSelectQueryAllowed(): bool
+    {
+        return true;
```
abnegate: Let's return false here and override for specific validators

src/Appwrite/Utopia/Database/Validator/Queries/Base.php
```diff
@@ -48,33 +50,43 @@ public function __construct(string $collection, array $allowedAttributes)
...
+            new Document([
+                'key' => '$sequence',
+                'type' => Database::VAR_INTEGER,
+                'array' => false,
+            ])
+        ];
+
+        foreach ($internalAttributes as $attribute) {
+            $attributes[] = $attribute;
+            $allAttributes[] = $attribute;
```
abnegate, line 87: What's the difference between $attributes and $allAttributes?

Meldiron: attributes only got ones allowed for filter queries (order, equals, ...)

I needed separate list that got it all.
Could be rewritten to first do allAttributes and then filter, but didnt see much benefit in it

src/Appwrite/Utopia/Database/Validator/Queries/Base.php
Meldiron: OH GOD, nice catch. Also adding 1 test in some random place, to make sure select queries dont work there

src/Appwrite/Utopia/Database/Validator/Queries/Base.php
abnegate: Where do we add to `$allAttributes` but not `$attributes`? Both loops I see add to both

Meldiron: Ah, got it. there wasforeach loop `continue`, but that was affecting allAttributes too. so nice catch.
Fixed, and also added test for select query of attribute not supported by filter.

src/Appwrite/Platform/Modules/Compute/Base.php
```diff
@@ -19,6 +21,47 @@
...
+     * Helper to apply select queries
+     *
+     * Method to respect request select queries in response format,
+     * to prevent default rule values from being applied on not-selected attributes
+     *
+     * @param Request $request
+     * @param Document $document
+     * @return void
+     */
+    public function applySelectQueries(Request $request, Response $response, string $model): void
```
abnegate: Can we put this in the base action class itself? Then we can use from any API

PR #10384, opened 2025-08-26
Fix relationship backwards compatibility

stnguyen90:
## What does this PR do?

Before 1.8, relationships would automatically resolve up to 3 levels deep. To ensure old apps don't break, we need to make sure requests from older SDKs behave the same.

The problem is our [request filter](https://github.com/appwrite/appwrite/blob/1.8.x/src/Appwrite/Utopia/Request/Filters/V20.php) doesn't work because:

1. it doesn't get 3 levels deep
2. [databaseId and collectionId are empty](https://github.com/appwrite/appwrite/blob/9fc39ddfce73597db58335f7571891f9c3e83701/src/Appwrite/Utopia/Request/Filters/V20.php#L107-L108) (probably because they're path params) so it can't fetch the related keys

This PR moves the logic to controller where we have access to the path params and adds a method to recursively fetch related keys to ensure we get related documents 3 levels deep.

## Test Plan

Tested locally using an app with list documents:

image

## Related PRs and Issues

None

## Checklist


eldadfux (approved):

Meldiron:
As per Jake's comment, this was already fixed by: https://github.com/appwrite/appwrite/pull/10389

Closing this PR now

PR #10418, opened 2025-09-01
Allow running tests with specific response format

abnegate:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,7 +8,15 @@ env:
   IMAGE: appwrite-dev
   CACHE_KEY: appwrite-dev-${{ github.event.pull_request.head.sha }}
 
-on: [ pull_request ]
+on:
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      response_format:
+        description: 'Response format version to test (e.g., 1.5.0, 1.4.0)'
+        required: false
+        type: string
+        default: ''
 
 jobs:
   check_database_changes:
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -8,7 +8,15 @@ env:
...
-on: [ pull_request ]
+on:
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      response_format:
+        description: 'Response format version to test (e.g., 1.5.0, 1.4.0)'
+        required: false
+        type: string
+        default: ''
```
ItzNotABug, line 19: lets add a guard somewhere to fail right before the tests start if the version is wrong.

tests/e2e/Scopes/Scope.php
```diff
@@ -23,6 +24,11 @@ protected function setUp(): void
     {
         $this->client = new Client();
         $this->client->setEndpoint($this->endpoint);
+
+        $format = System::getEnv('_APP_E2E_RESPONSE_FORMAT');
```
ItzNotABug, line 28: lets add the env on the .env example file as well. better discoverability.

.github/workflows/tests.yml
abnegate: Addressed

tests/e2e/Scopes/Scope.php
abnegate: It's only relevant for us so no need

ItzNotABug (approved):

PR #10447, opened 2025-09-05
Chore: Upgrade platform lib to have older queue lib

Meldiron:
## What does this PR do?

Newer platform lib, for older queue lib, for stability (0.11 of queue is battle-tested)

## Test Plan

Current tests

## Related PRs and Issues

x

## Checklist


composer.json
```diff
@@ -65,10 +65,10 @@
         "utopia-php/messaging": "0.18.*",
         "utopia-php/migration": "1.*",
         "utopia-php/orchestration": "0.9.*",
-        "utopia-php/platform": "0.7.*",
+        "utopia-php/platform": "0.7.12",
```
abnegate: I think we can keep as 0.7.* right?

PR #10458, opened 2025-09-08
Auth email redesign

hmacr:
Redesigned auth related emails to include heading and footers


stnguyen90 (commented):
Wouldn't this also change the email template for emails in projects as well?

app/controllers/api/teams.php
```diff
@@ -738,11 +738,17 @@
                     'user' => $name,
                     'team' => $team->getAttribute('name'),
                     'redirect' => $url,
-                    'project' => $projectName
+                    'project' => $projectName,
+                    'twitterUrl' => APP_SOCIAL_TWITTER,
+                    'discordUrl' => APP_SOCIAL_DISCORD,
+                    'githubUrl' => APP_SOCIAL_GITHUB_APPWRITE,
+                    'termsUrl' => APP_EMAIL_TERMS_URL,
+                    'privacyUrl' => APP_EMAIL_PRIVACY_URL,
```
Meldiron, line 746: To prevent the need to set this in every flow, let's please try and see if we could do this in Emails worker instead, just once.

We can also prefix keys with Appwrite, `{{appwriteDiscordUrl}}`, to lower chance of conflict in future

app/config/locale/templates/email-auth.tpl
Meldiron, line 1: This file name is a bit confusing. Isn't this generic branded base?

Meldiron: If so, let's see how we named unbranded base, and follow similar naming pattern

app/controllers/api/account.php
```diff
@@ -2123,11 +2131,17 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
+            'discordUrl' => APP_SOCIAL_DISCORD,
+            'githubUrl' => APP_SOCIAL_GITHUB_APPWRITE,
+            'termsUrl' => APP_EMAIL_TERMS_URL,
+            'privacyUrl' => APP_EMAIL_PRIVACY_URL,
         ];
 
         $queueForMails
             ->setSubject($subject)
             ->setPreview($preview)
+            ->setBodyTemplate(__DIR__ . '/../../config/locale/templates/email-auth.tpl')
```
Meldiron, line 2144: We cannot apply branded base everywhere; we should only do this for Console emails. For Console, when we need branded emails, let's use SMTP templates (overrides) like in this PR: https://github.com/appwrite/appwrite/pull/10501

Aglowharmony (approved):
Merge it and close the PR please

PR #8102, opened 2024-05-08
Fix email used for name when user is created via Apple OAuth2

stnguyen90:
## What does this PR do?

Fixes #7268

## Test Plan

Manually tested Apple OAuth2 and got my name to appear:

image

## Related PRs and Issues

- #7268

## Checklist

- [x] Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [x] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

PR #10468, opened 2025-09-10
Feat apps module dl

lohanidamodar:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


composer.json
```diff
@@ -46,6 +46,7 @@
         "ext-sockets": "*",
         "appwrite/php-runtimes": "0.19.*",
         "appwrite/php-clamav": "2.0.*",
+        "utopia-php/auth": "0.4.0",
```
Meldiron:
```suggestion
        "utopia-php/auth": "0.4.*",
```

app/realtime.php
```diff
@@ -671,15 +674,24 @@ function getTelemetry(int $workerId): Utopia\Telemetry\Adapter
...
+                $user = $database->getDocument('users', $store->getProperty('id', ''));
+
+                /**
+                 * TODO:
+                 * Moving forward, we should try to use our dependency injection container
+                 * to inject the proof for token.
+                 * This way we will have one source of truth for the proof for token.
+                 */
+                $proofForToken = new Token();
+                $proofForToken->setHash(new Sha());
```
Meldiron, line 697: I am probably not seeing whole picture, but something feels off here.

1. Why do we decode without giving credentials like secret key?
2. Why do we create proof for token without any link to Store, to original user?

src/Appwrite/Platform/Modules/Functions/Http/Executions/Create.php
```diff
@@ -199,7 +205,7 @@ public function action(
 
             foreach ($sessions as $session) {
                 /** @var Utopia\Database\Document $session */
-                if ($session->getAttribute('secret') == Auth::hash(Auth::$secret)) { // If current session delete the cookies too
+                if ($proofForToken->verify($store->getProperty('secret', ''), $session->getAttribute('secret'))) { // If current session delete the cookies too
```
Meldiron:
```suggestion
                if ($proofForToken->verify($store->getProperty('secret', ''), $session->getAttribute('secret'))) { // Find most recent active session for user ID and JWT headers
```

tests/e2e/Services/Account/AccountCustomClientTest.php
```diff
@@ -2698,7 +2698,7 @@ public function testCreateMagicUrl(): array
         $this->assertStringContainsStringIgnoringCase('Sign in to '. $this->getProject()['name'] . ' with your secure link. Expires in 1 hour.', $lastEmail['text']);
         $this->assertStringNotContainsStringIgnoringCase('security phrase', $lastEmail['text']);
 
-        $token = substr($lastEmail['text'], strpos($lastEmail['text'], '&secret=', 0) + 8, 64);
+        $token = substr($lastEmail['text'], strpos($lastEmail['text'], '&secret=', 0) + 8, 256);
```
Meldiron: Did we make this longer intentionally? I recall discussion with design team saying urls are too long and ugly.

Can we please make screenshot what this email looks now with longer URL, and confirm with Design if the increase is okay?

app/controllers/api/users.php
```diff
@@ -1354,8 +1402,12 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
             ->setAttribute('passwordHistory', $history)
             ->setAttribute('passwordUpdate', DateTime::now())
-            ->setAttribute('hash', Auth::DEFAULT_ALGO)
-            ->setAttribute('hashOptions', Auth::DEFAULT_ALGO_OPTIONS);
+            ->setAttribute('hash', 'argon2')
+            ->setAttribute('hashOptions', [
+                'memoryCost' => 65536,
+                'timeCost' => 4,
+                'threads' => 3
+            ]);
```
Meldiron: We should reuse `$hasher` here, take values form there (aroud like 1380)

app/controllers/api/account.php
```diff
@@ -938,11 +960,11 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
         Authorization::setRole(Role::user($user->getId())->toString());
 
         // Re-hash if not using recommended algo
-        if ($user->getAttribute('hash') !== Auth::DEFAULT_ALGO) {
+        if ($user->getAttribute('hash') !== $proofForPassword->getHash()->getName()) {
             $user
-                ->setAttribute('password', Auth::passwordHash($password, Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS))
-                ->setAttribute('hash', Auth::DEFAULT_ALGO)
-                ->setAttribute('hashOptions', Auth::DEFAULT_ALGO_OPTIONS);
+                ->setAttribute('password', (new ProofsPassword())->hash($password))
```
Meldiron: Do we need to make new instance here?

app/controllers/api/account.php
```diff
@@ -1470,7 +1501,7 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
 
         $name = '';
         $nameOAuth = $oauth2->getUserName($accessToken);
-        $userParam = $request->getParam('user');
+        $userParam = \json_decode($request->getParam('user', '{}'), true); // only valid for Apple OAuth2 which returns a user param in the request
```
Meldiron: Is this intentional change in this PR?

app/controllers/api/teams.php
```diff
@@ -566,6 +571,7 @@
 
             try {
                 $userId = ID::unique();
+                $hash = $proofForPassword->hash($proofForPassword->generate());
```
Meldiron, line 573: We shouldnt auto-generate password anymore here. we have so many auth flows, password is no longer a requirement.

Might be worth checking all other flows too. I am pretty sure, in account.php, we already leave password at `null` for passwordless flows

app/controllers/api/teams.php
```diff
@@ -474,7 +477,7 @@
             ;
             $roles = array_keys(Config::getParam('roles', []));
             array_filter($roles, function ($role) {
-                return !in_array($role, [Auth::USER_ROLE_APPS, Auth::USER_ROLE_GUESTS, Auth::USER_ROLE_USERS]);
+                return !in_array($role, [USER_ROLE_APPS, USER_ROLE_GUESTS, USER_ROLE_USERS]);
             });
             return new ArrayList(new WhiteList($roles), APP_LIMIT_ARRAY_PARAMS_SIZE);
         }
```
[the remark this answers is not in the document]
stnguyen90: @lohanidamodar

app/controllers/api/account.php
lohanidamodar: @eldadfux can you help understand this?

app/init/resources.php
```diff
@@ -196,72 +202,89 @@
...
-    if (empty($session['id']) && empty($session['secret'])) {
+    if (empty($store->getProperty('id', '')) && empty($store->getProperty('secret', ''))) {
         if ($response) {
             $response->addHeader('X-Debug-Fallback', 'true');
         }
         $fallback = $request->getHeader('x-fallback-cookies', '');
         $fallback = \json_decode($fallback, true);
-        $session = Auth::decodeSession(((isset($fallback[Auth::$cookieName])) ? $fallback[Auth::$cookieName] : ''));
+        $store->decode(((isset($fallback[$store->getKey()])) ? $fallback[$store->getKey()] : ''));
     }
```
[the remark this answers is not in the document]
stnguyen90: @lohanidamodar, we should probably implement this suggestion.

app/init/resources.php
```diff
@@ -196,72 +202,89 @@
...
-                $user = $dbForPlatform->getDocument('users', Auth::$unique);
+                $user = $dbForPlatform->getDocument('users', $store->getProperty('id', ''));
             } else {
-                $user = $dbForProject->getDocument('users', Auth::$unique);
+                $user = $dbForProject->getDocument('users', $store->getProperty('id', ''));
             }
         }
+    } else {
+        $user = $dbForPlatform->getDocument('users', $store->getProperty('id', ''));
     }
```
stnguyen90: How come the logic was refactored? Having the `if ($mode === APP_MODE_ADMIN)` before keeps the smaller code block at the top so that you're more likely to see the if/else. It's not too big of a deal for this block since the code blog isn't as big, but for bigger code blocks, putting the smaller one first is generally a good idea.

src/Appwrite/Platform/Tasks/Install.php
```diff
@@ -157,12 +158,12 @@ public function action(string $httpPort, string $httpsPort, string $organization
...
 
                 if ($var['filter'] === 'token') {
-                    $input[$var['name']] = Auth::tokenGenerator();
+                    $input[$var['name']] = (new Token())->generate();
                     continue;
                 }
 
                 if ($var['filter'] === 'password') {
-                    $input[$var['name']] = Auth::passwordGenerator();
+                    $input[$var['name']] = (new Password())->generate();
```
stnguyen90: probably best to initialize Token and Password outside of the foreach.

src/Appwrite/Migration/Version/V17.php
```diff
@@ -270,7 +269,7 @@ protected function fixDocument(Document $document)
                 * Set hashOptions type
                 */
                 $document->setAttribute('hashOptions', array_merge($document->getAttribute('hashOptions', []), [
-                    'type' => $document->getAttribute('hash', Auth::DEFAULT_ALGO)
+                    'type' => $document->getAttribute('hash', 'argon2')
```
stnguyen90: Can we do `Password::ARGON2` instead? Or maybe `Password()->hash->getName()`?

app/controllers/api/teams.php
```diff
@@ -1265,22 +1273,27 @@
...
                     domain: Config::getParam('cookieDomain'),
                     secure: ('https' === $protocol),
                     httponly: true
                 )
                 ->addCookie(
-                    name: Auth::$cookieName,
-                    value: Auth::encodeSession($user->getId(), $secret),
+                    name: $store->getKey(),
+                    value: $encoded,
                     expire: (new \DateTime($expire))->getTimestamp(),
```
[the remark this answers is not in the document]
stnguyen90: @lohanidamodar please address this.

app/controllers/shared/api.php
```diff
@@ -222,38 +222,93 @@
...
             if ($apiKey->isExpired()) {
                 throw new Exception(Exception::PROJECT_KEY_EXPIRED);
             }
 
+            // Set role and scopes from API key
             $role = $apiKey->getRole();
             $scopes = $apiKey->getScopes();
 
             // Disable authorization checks for API keys
             Authorization::setDefaultStatus(false);
```
stnguyen90: FYI: https://github.com/appwrite/appwrite/pull/10489

app/controllers/api/users.php
lohanidamodar: done.

app/controllers/api/account.php
lohanidamodar: yes because the above checks is making sure that the existing proof's hash doesn't match the new hash set in the user so we need to update with new hash.

app/controllers/api/teams.php
lohanidamodar: There's this from previous as well
```
// TODO: Set password empty?

  /**
   * Set the password update time to 0 for users created using
   * team invite and OAuth to allow password updates without an
   * */
```

So I'm not modifying it as a part of this PR

app/controllers/general.php
```diff
@@ -1185,17 +1185,29 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
...
+            foreach ($trace as $index => $entry) {
+                $traceFile = $entry['file'] ?? 'unknown';
+                $traceLine = $entry['line'] ?? 0;
+                $traceFunction = $entry['function'] ?? 'unknown';
+                $traceClass = $entry['class'] ?? '';
+                $traceType = $entry['type'] ?? '';
+                Console::$logLevel("  #{$index} {$traceFile}({$traceLine}): {$traceClass}{$traceType}{$traceFunction}()");
+            }
+            Console::$logLevel('');
         }
```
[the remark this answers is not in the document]
stnguyen90: @lohanidamodar why was this resolved? It seems legitimate.

app/realtime.php
lohanidamodar: I think it's same as before, in realtime service Auth static also has these value empty or default and here too we  are using the default, which is the same. cc @eldadfux am I correct?

app/controllers/api/users.php
```diff
@@ -2247,27 +2302,30 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
...
         $user = $dbForProject->getDocument('users', $userId);
 
         if ($user->isEmpty()) {
             throw new Exception(Exception::USER_NOT_FOUND);
         }
 
-        $secret = Auth::tokenGenerator($length);
+        $secret = $proofForToken
+            ->setLength($length)
+            ->generate();
```
stnguyen90: This will change the `$proofForToken` resource. Do we really want to do that? What if something else uses it later and expects the length to be 256?

tests/e2e/Services/Account/AccountCustomClientTest.php
lohanidamodar: Fixed, it was not intended.

app/controllers/api/account.php
stnguyen90: But why can't you use `$proofForPassword`. You use it for the hash and hashOptions below.

app/controllers/api/account.php
```diff
@@ -1361,16 +1396,12 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
-        } elseif ($protocol === 'http' && $port !== '80') {
-            $callbackBase .= ':' . $port;
-        }
-        $callback = $callbackBase . '/v1/account/sessions/oauth2/callback/' . $provider . '/' . $project->getId();
+    ->inject('store')
+    ->inject('proofForPassword')
+    ->inject('proofForToken')
+    ->action(function (string $provider, string $code, string $state, string $error, string $error_description, Request $request, Response $response, Document $project, array $platforms, Document $devKey, Document $user, Database $dbForProject, Reader $geodb, Event $queueForEvents, Store $store, ProofsPassword $proofForPassword, ProofsToken $proofForToken) use ($oauthDefaultSuccess) {
+        $protocol = $request->getProtocol();
+        $callback = $protocol . '://' . $request->getHostname() . '/v1/account/sessions/oauth2/callback/' . $provider . '/' . $project->getId();
```
stnguyen90: Why was this changed?

app/controllers/api/account.php
stnguyen90: This should be something like:

https://github.com/appwrite/appwrite/blob/a703fcc8e94b8d766cfd868b76ba08646ec4fdfa/app/controllers/api/account.php#L1460-L1468

Refer to the latest 1.7.x code.

Related PRs:
* https://github.com/appwrite/appwrite/pull/8102
* https://github.com/appwrite/appwrite/pull/10130
* https://github.com/appwrite/appwrite/pull/10150

app/controllers/api/account.php
```diff
@@ -1995,15 +2034,17 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
             Authorization::skip(fn () => $dbForProject->createDocument('users', $user));
         }
 
-        $tokenSecret = Auth::tokenGenerator(Auth::TOKEN_LENGTH_MAGIC_URL);
-        $expire = DateTime::formatTz(DateTime::addSeconds(new \DateTime(), Auth::TOKEN_EXPIRATION_CONFIRM));
+        $proofForToken->setLength(TOKEN_LENGTH_MAGIC_URL);
```
stnguyen90: This will change the length of the resource so any time `generate()` is called after this (even outside of this action), it will have this length instead of the default length. Are we sure we ant ot do this rather than creating a new proof for this?

stnguyen90 (changes requested):

app/controllers/api/account.php
eldadfux: Could be that the original logic changed after the PR was created. @lohanidamodar I don't recall any intentional change, let's sync with the most up-to-date logic.

app/realtime.php
eldadfux: @Meldiron not sure what is the problem you notice here. cc @lohanidamodar

app/controllers/api/teams.php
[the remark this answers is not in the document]
lohanidamodar: @stnguyen90 I think this is intended for console we filter roles to supported roles only. Also in this PR we are only changing the constant and nothing else.

app/init/resources.php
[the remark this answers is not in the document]
lohanidamodar: updated the check.

app/init/resources.php
lohanidamodar: I'm not sure I unserstand what you mean? @eldadfux can you share your logic behind this change?

src/Appwrite/Platform/Tasks/Install.php
lohanidamodar: done.

src/Appwrite/Migration/Version/V17.php
lohanidamodar: done.

app/controllers/shared/api.php
lohanidamodar: already fixed.

app/controllers/general.php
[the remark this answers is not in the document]
lohanidamodar: Resovled because it worked. But seems it was a legacy way and still worked. Updated to latest way as suggested.

app/controllers/api/account.php
lohanidamodar: Yes, forgot to update below, fixed

app/controllers/api/account.php
lohanidamodar: fixed

app/controllers/api/account.php
lohanidamodar: fixed.

app/controllers/api/users.php
lohanidamodar: fixed.

app/controllers/api/account.php
lohanidamodar: reset

app/controllers/api/teams.php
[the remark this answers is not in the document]
stnguyen90: @lohanidamodar this array_filter code does nothing because array_filter does not modify the array passed in. Read code rabbit's comment and the docs for [array_filter](https://www.php.net/manual/en/function.array-filter.php).

src/Appwrite/Platform/Tasks/Install.php
```diff
@@ -157,12 +158,14 @@ public function action(string $httpPort, string $httpsPort, string $organization
...
-                    $input[$var['name']] = Auth::tokenGenerator();
+                    $token = new Token();
+                    $input[$var['name']] = $token->generate();
                     continue;
                 }
 
                 if ($var['filter'] === 'password') {
-                    $input[$var['name']] = Auth::passwordGenerator();
+                    $password = new Password();
+                    $input[$var['name']] = $password->generate();
```
stnguyen90: I said outside of the `foreach` because this is creating a new instance each time.

app/init/resources.php
stnguyen90: I mean this is good:

```
if ($condition) {
  // small code block
} else {
  // really
  // big
  // code
  // block
}
```

and this is bad:

```
if ($condition) {
  // really
  // big
  // code
  // block
} else {
  // small code block
}
```

Because if you put the small code block later, you forget what the condition is by the time you've finished reading the big code block.

app/controllers/api/account.php
```diff
@@ -938,11 +960,11 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
-        if ($user->getAttribute('hash') !== Auth::DEFAULT_ALGO) {
+        if ($user->getAttribute('hash') !== $proofForPassword->getHash()->getName()) {
             $user
-                ->setAttribute('password', Auth::passwordHash($password, Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS))
-                ->setAttribute('hash', Auth::DEFAULT_ALGO)
-                ->setAttribute('hashOptions', Auth::DEFAULT_ALGO_OPTIONS);
+                ->setAttribute('password', (new ProofsPassword())->hash($password))
+                ->setAttribute('hash', $proofForPassword->getHash()->getName())
+                ->setAttribute('hashOptions', $proofForPassword->getHash()->getOptions());
             $dbForProject->updateDocument('users', $user->getId(), $user);
```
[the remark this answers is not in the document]
stnguyen90: @lohanidamodar coderabbit makes a good point. If we want the password to use the hash in `$proofForPassword` why not use `$proofForPassword` for these instead of initializing a new instance?

stnguyen90 (approved):
Just minor suggestions.

app/init/resources.php
lohanidamodar: Got it. Updated.

app/controllers/api/account.php
```diff
@@ -167,8 +171,7 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
         ->trigger();
 };
 
-
-$createSession = function (string $userId, string $secret, Request $request, Response $response, Document $user, Database $dbForProject, Document $project, Locale $locale, Reader $geodb, Event $queueForEvents, Mail $queueForMails) {
+$createSession = function (string $userId, string $secret, Request $request, Response $response, Document $user, Database $dbForProject, Document $project, Locale $locale, Reader $geodb, Event $queueForEvents, Mail $queueForMails, Store $store, ProofsToken $proofForToken) {
```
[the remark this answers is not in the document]
abnegate: @lohanidamodar I think this is valid

app/controllers/api/account.php
```diff
@@ -372,7 +381,9 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
         $hooks->trigger('passwordValidator', [$dbForProject, $project, $password, &$user, true]);
 
         $passwordHistory = $project->getAttribute('auths', [])['passwordHistory'] ?? 0;
-        $password = Auth::passwordHash($password, Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS);
+        $proof = new ProofsPassword();
```
abnegate, line 385: Let's use an injection of `proofForPassword`

app/controllers/api/account.php
```diff
@@ -542,12 +553,13 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
-    ->inject('project')
-    ->action(function (Response $response, Document $user, Locale $locale, Document $project) {
+    ->inject('store')
+    ->inject('proofForToken')
+    ->action(function (Response $response, Document $user, Locale $locale, Store $store, ProofsToken $proofForToken) {
 
 
         $sessions = $user->getAttribute('sessions', []);
-        $current = Auth::sessionVerify($sessions, Auth::$secret);
+        $current = Auth::sessionVerify($sessions, $store->getProperty('secret', ''), $proofForToken);
```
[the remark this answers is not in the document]
abnegate: @lohanidamodar Is this valid?

app/init/resources.php
```diff
@@ -226,72 +232,89 @@
     ];
 }, ['request', 'console', 'project', 'dbForPlatform']);
 
-App::setResource('user', function ($mode, $project, $console, $request, $response, $dbForProject, $dbForPlatform) {
+App::setResource('user', function ($mode, $project, $console, $request, $response, $dbForProject, $dbForPlatform, Store $store, Token $proofForToken) {
```
abnegate: Let's add types for other params here instead of phpdoc

app/init/resources.php
```diff
@@ -226,72 +232,89 @@
...
+    } else {
+        if ($project->isEmpty()) {
+            $user = new Document([]);
+        } else {
             if ($project->getId() === 'console') {
-                $user = $dbForPlatform->getDocument('users', Auth::$unique);
+                $user = $dbForPlatform->getDocument('users', $store->getProperty('id', ''));
             } else {
-                $user = $dbForProject->getDocument('users', Auth::$unique);
+                $user = $dbForProject->getDocument('users', $store->getProperty('id', ''));
```
abnegate: Do we miss a wrapping `if (!empty($store->getProperty))` since we had `if (!empty(Auth::$unique))` before?

app/init/resources.php
```diff
@@ -947,6 +970,37 @@ function getDevice(string $root, string $connection = ''): Device
...
+        ->setMemoryCost(2048)
+        ->setTimeCost(4)
+        ->setThreads(3);
+
+    $password = new Password();
+    $password
+        ->setHash($hash);
+
+    return $password;
+});
```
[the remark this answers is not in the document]
abnegate: @lohanidamodar Let's address

app/controllers/api/account.php
```diff
@@ -894,15 +912,20 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
+    ->action(function (string $email, string $password, Request $request, Response $response, Document $user, Database $dbForProject, Document $project, Locale $locale, Reader $geodb, Event $queueForEvents, Mail $queueForMails, Hooks $hooks, Store $store, ProofsPassword $proofForPassword, ProofsToken $proofForToken) {
         $email = \strtolower($email);
         $protocol = $request->getProtocol();
 
         $profile = $dbForProject->findOne('users', [
             Query::equal('email', [$email]),
         ]);
 
-        if ($profile->isEmpty() || empty($profile->getAttribute('passwordUpdate')) || !Auth::passwordVerify($password, $profile->getAttribute('password'), $profile->getAttribute('hash'), $profile->getAttribute('hashOptions'))) {
+        $userProofForPassword = ProofsPassword::createHash($profile->getAttribute('hash', $proofForPassword->getHash()->getName()), $profile->getAttribute('hashOptions', $proofForPassword->getHash()->getOptions()));
```
abnegate, line 927: Don't we have the instance method `hash` on the password proof already?

app/controllers/api/account.php
```diff
@@ -1736,7 +1768,7 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
                 'providerAccessToken' => $accessToken,
                 'providerRefreshToken' => $refreshToken,
                 'providerAccessTokenExpiry' => DateTime::addSeconds(new \DateTime(), (int)$accessTokenExpiry),
-                'secret' => Auth::hash($secret), // One way hash encryption to protect DB leak
+                'secret' => $proofForToken->hash($secret), // One way hash encryption to protect DB leak
                 'userAgent' => $request->getUserAgent('UNKNOWN'),
                 'ip' => $request->getIP(),
                 'factors' => [TYPE::EMAIL, 'oauth2'], // include a special oauth2 factor to bypass MFA checks
```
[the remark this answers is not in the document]
abnegate: @lohanidamodar Is this valid?

app/controllers/api/teams.php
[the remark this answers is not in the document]
abnegate: @lohanidamodar We need to assign $roles = array_filter(...)

app/controllers/api/users.php
```diff
@@ -97,7 +107,18 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
             }
         }
 
-        $password = (!empty($password)) ? ($hash === 'plaintext' ? Auth::passwordHash($password, $hash, $hashOptionsObject) : $password) : null;
+        $hashedPassword = null;
+
+        if (!empty($password)) {
+            if ($hash instanceof Plaintext) { // Password was never hashed, hash it with the default hash
+                $defaultHash = new ProofsPassword();
```
abnegate, line 114: Can we inject the resource?

app/controllers/api/users.php
```diff
@@ -1336,12 +1376,20 @@ function createUser(string $hash, mixed $hashOptions, string $userId, ?string $e
 
         $hooks->trigger('passwordValidator', [$dbForProject, $project, $password, &$user, true]);
 
-        $newPassword = Auth::passwordHash($password, Auth::DEFAULT_ALGO, Auth::DEFAULT_ALGO_OPTIONS);
+        // Create Argon2 hasher with default settings
+        $hasher = new Argon2();
+        $hasher->setMemoryCost(2048);
+        $hasher->setTimeCost(4);
+        $hasher->setThreads(3);
```
abnegate: Let's chain setters for consistency

app/init/resources.php
```diff
@@ -226,72 +232,89 @@
...
+
+    /**
+     * Handles user authentication and session validation.
+     *
+     * This function follows a series of steps to determine the appropriate user session
+     * based on cookies, headers, and JWT tokens.
+     *
+     * Process:
+     * 1. Checks the cookie based on mode:
+     *    - If in admin mode, redirects to the console.
```
abnegate: I don't think this line is correct

src/Appwrite/Platform/Tasks/Install.php
abnegate: @lohanidamodar Agree, let's instantiate Password + Token before the loop

app/init/resources.php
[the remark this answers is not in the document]
lohanidamodar: https://github.com/utopia-php/auth/blob/02415e1a89cdbc14e3e16a7856ecf7f868869449/src/Auth/Hashes/Argon2.php#L12
Noted will need to update in the library

app/controllers/api/account.php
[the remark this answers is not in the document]
lohanidamodar: The sessionVerify is doing the same thing, this just repeats the code everywhere, we will need to find a better pattern later.

app/controllers/api/account.php
lohanidamodar: No we don't, the hash method resides on each hashing algorithm only and password proof has this helper method.

app/controllers/api/users.php
lohanidamodar: I think it's better this way? only one endpoint sends plaintext to this function, and if we inject, we will have to inject in all the endpoints to create user using various different hash.

app/init/resources.php
lohanidamodar: updated

app/init/resources.php
lohanidamodar: added the check as it prevents one database query if id is empty.

app/init/resources.php
```diff
@@ -226,72 +233,91 @@
...
+                    $user = $dbForProject->getDocument('users', $store->getProperty('id', ''));
+                }
             }
         }
     }
 
     if (
         $user->isEmpty() // Check a document has been found in the DB
-        || !Auth::sessionVerify($user->getAttribute('sessions', []), Auth::$secret)
+        || !Auth::sessionVerify($user->getAttribute('sessions', []), $store->getProperty('secret', ''), $proofForToken)
```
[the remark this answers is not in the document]
lohanidamodar: this is known limitation at the moment and not critical

app/controllers/api/account.php
abnegate: Wouldn't it make more sense to set the hash options then use the instance method?

abnegate: Oh never mind, that's pretty much what we're doing

app/controllers/api/account.php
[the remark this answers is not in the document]
abnegate: @lohanidamodar I think this is valid, at least the const

app/controllers/api/account.php
```diff
@@ -2580,6 +2646,8 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
                 'status' => true,
                 'password' => null,
                 'passwordUpdate' => null,
+                'hash' => $proofForPassword->getHash()->getName(),
+                'hashOptions' => $proofForPassword->getHash()->getOptions(),
```
abnegate: I don't think this is valid, there's no password

app/controllers/api/account.php
```diff
@@ -3810,7 +3899,7 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
         }
 
         $tokens = $profile->getAttribute('tokens', []);
-        $verifiedToken = Auth::tokenVerify($tokens, Auth::TOKEN_TYPE_VERIFICATION, $secret);
+        $verifiedToken = Auth::tokenVerify($tokens, TOKEN_TYPE_VERIFICATION, $secret, $proofForToken);
```
abnegate, line 3899: Seems wrong that we still use a few static method on `Auth`, now we have a split between these and the new lib

app/init/resources.php
```diff
@@ -947,6 +974,37 @@ function getDevice(string $root, string $connection = ''): Device
...
+App::setResource('store', function (): Store {
+    return new Store();
+});
+
+App::setResource('proofForPassword', function (): Password {
+    $hash = new Argon2();
+    $hash
+        ->setMemoryCost(7168)
+        ->setTimeCost(5)
+        ->setThreads(1);
```
abnegate, line 987: There are a couple of other isntances of Argon2, lets set the same values for all

app/controllers/api/account.php
lohanidamodar: yes, but the static methods are not the ones that will not break anything when we move to coroutine. We only use static methods no more static properties. These are just functions that return values based on provided params.

lohanidamodar: removed

app/controllers/api/account.php
lohanidamodar: updated.

PR #10472, opened 2025-09-10
Stats resources and usage sorting by unique field

fogelito:
When we do INSERT ... ON DUPLICATE KEY UPDATE in the upsert method
- It Looks up the rows that matches our unique key
- Locks the index entries in the order of the unique index definition.

If we have many concurrent workers inserting/updating the same table, they may lock different parts of the unique index in different orders.
That can create the classic deadlock

	

## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Workers/StatsUsage.php
+++ b/src/Appwrite/Platform/Workers/StatsUsage.php
@@ -504,7 +504,7 @@ class StatsUsage extends Action
 
             usort($this->statDocuments, function ($a, $b) {
                 // Tenant ASC
-                $cmp = $a['_tenant'] <=> $b['_tenant'];
+                $cmp = $a['$tenant'] <=> $b['$tenant'];
                 if ($cmp !== 0) {
                     return $cmp;
                 }


src/Appwrite/Platform/Workers/StatsResources.php
```diff
@@ -434,18 +434,32 @@ protected function writeDocuments(Database $dbForLogs, Document $project): void
...
+                $this->documents,
+                10 // See if this make an effect
             );
 
             Console::success($message . ' | Documents: ' . count($this->documents));
         } catch (\Throwable $e) {
             Console::error('Error: ' . $message . ' | Exception: ' . $e->getMessage());
             throw $e;
-        } finally {
-            $this->documents = [];
```
fogelito, line 448: Reverting since it is set in the beginning of the script.
Was added in the previous PR , which was merged

loks0n (approved):

src/Appwrite/Platform/Workers/StatsUsage.php
```diff
@@ -424,12 +424,40 @@ public function commitToDb(callable $getProjectDB): void
...
+                        return ($b['time'] === null) ? 0 : -1;
+                    }
+                    if ($b['time'] === null) {
+                        return 1;
+                    }
+
+                    return strcmp($a['time'], $b['time']);
+                });
+
                 $dbForProject->createOrUpdateDocumentsWithIncrease('stats', 'value', $projectStats['stats']);
```
fogelito, line 455: Was this function name changes ?

src/Appwrite/Platform/Workers/StatsUsage.php
```diff
@@ -424,12 +424,40 @@ public function commitToDb(callable $getProjectDB): void
...
                 Console::success('Batch successfully written to DB');
-
-                unset($this->projects[$sequence]);
             } catch (Throwable $e) {
                 Console::error('Error processing stats: ' . $e->getMessage());
+            } finally {
+                unset($this->projects[$sequence]);
```
lohanidamodar, line 460: we shouldn't clear if the insert fails, or we will loose the data.

fogelito: We have 3 retry attempts in the Utopia lib in case of failure.

PR #10478, opened 2025-09-11
Improve: UX of dns flows

Meldiron:
## What does this PR do?

- Stores last verification error message in attribute
- Improves message of DNS validator when it fails
- Moves all DNS verification to single re-usable method
- Support new action in certificate worker (verification) alongside current action (cert generation)
- Adds ability to override verification domains in certificate worker for easy cross-region support
- Instantly upsert certificate from worker, to quickly show changes to logs in UI
- comma-separated list support for dns client, for multi-level verification
- Re-orders validation to look for CAA issues first
- Rule's $updatedAt to have certificate's updatedAt if its later
- Add constants for rule status, and document each status

## Test Plan

- new proxy tests implemented for e2e validation flows (with coredns)

## Related PRs and Issues

x

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Tasks/Maintenance.php
+++ b/src/Appwrite/Platform/Tasks/Maintenance.php
@@ -2,7 +2,6 @@
 
 namespace Appwrite\Platform\Tasks;
 
-use Appwrite\Event\Certificate;
 use Appwrite\Event\Delete;
 use DateInterval;
 use DateTime;
@@ -27,18 +26,16 @@ class Maintenance extends Action
             ->desc('Schedules maintenance tasks and publishes them to our queues')
             ->inject('dbForPlatform')
             ->inject('console')
-            ->inject('queueForCertificates')
             ->inject('queueForDeletes')
             ->callback($this->action(...));
     }
 
-    public function action(Database $dbForPlatform, Document $console, Certificate $queueForCertificates, Delete $queueForDeletes): void
+    public function action(Database $dbForPlatform, Document $console, Delete $queueForDeletes): void
     {
         Console::title('Maintenance V1');
         Console::success(APP_NAME . ' maintenance process v1 has started');
 
         $interval = (int) System::getEnv('_APP_MAINTENANCE_INTERVAL', '86400'); // 1 day
-        $intervalRuleVerification = (int) System::getEnv('_APP_MAINTENANCE_INTERVAL_RULE_VERIFICATION', '60'); // 1 minute
 
         $usageStatsRetentionHourly = (int) System::getEnv('_APP_MAINTENANCE_RETENTION_USAGE_HOURLY', '8640000'); //100 days
         $cacheRetention = (int) System::getEnv('_APP_MAINTENANCE_RETENTION_CACHE', '2592000'); // 30 days
@@ -61,8 +58,8 @@ class Maintenance extends Action
 
         Console::info('Setting loop start time to ' . $next->format("Y-m-d H:i:s.v") . '. Delaying for ' . $delay . ' seconds.');
 
-        \go(function () use ($interval, $cacheRetention, $schedulesDeletionRetention, $usageStatsRetentionHourly, $dbForPlatform, $console, $queueForDeletes, $queueForCertificates, $delay) {
-            Console::loop(function () use ($interval, $cacheRetention, $schedulesDeletionRetention, $usageStatsRetentionHourly, $dbForPlatform, $console, $queueForDeletes, $queueForCertificates) {
+        \go(function () use ($interval, $cacheRetention, $schedulesDeletionRetention, $usageStatsRetentionHourly, $dbForPlatform, $console, $queueForDeletes, $delay) {
+            Console::loop(function () use ($interval, $cacheRetention, $schedulesDeletionRetention, $usageStatsRetentionHourly, $dbForPlatform, $console, $queueForDeletes) {
                 $time = DatabaseDateTime::now();
 
                 Console::info("[{$time}] Notifying workers with maintenance tasks every {$interval} seconds");
@@ -95,17 +92,10 @@ class Maintenance extends Action
                     ->trigger();
 
                 $this->notifyDeleteConnections($queueForDeletes);
-                $this->renewCertificates($dbForPlatform, $queueForCertificates);
                 $this->notifyDeleteCache($cacheRetention, $queueForDeletes);
                 $this->notifyDeleteSchedules($schedulesDeletionRetention, $queueForDeletes);
                 $this->notifyDeleteCSVExports($queueForDeletes);
             }, $interval, $delay);
-        });
-
-        \go(function () use ($dbForPlatform, $queueForCertificates, $intervalRuleVerification) {
-            Console::loop(function () use ($dbForPlatform, $queueForCertificates) {
-                $this->checkRuleVerification($dbForPlatform, $queueForCertificates);
-            }, $intervalRuleVerification);
         });
     }
 
@@ -117,85 +107,11 @@ class Maintenance extends Action
             ->trigger();
     }
 
-    private function checkRuleVerification(Database $dbForPlatform, Certificate $queueForCertificate): void
-    {
-        $time = DatabaseDateTime::now();
-
-        $oldestToCheck = new DateTime('-3 days');
-
-        $rules = $dbForPlatform->find('rules', [
-            Query::createdAfter(DatabaseDateTime::format($oldestToCheck)), // max 3 days old
-            Query::equal('status', [RULE_STATUS_VERIFICATION_FAILED]), // not verified yet
-            Query::orderAsc('$updatedAt'), // Pick the ones waiting for another attempt for longest
-            Query::equal('region', [System::getEnv('_APP_REGION', 'default')]), // Only current region
-            Query::limit(30), // Reasonable pagination limit, processable within a minute
-        ]);
-
-        if (\count($rules) > 0) {
-            Console::info("[{$time}] Found " . \count($rules) . " rules for verification, scheduling jobs.");
-
-            foreach ($rules as $rule) {
-                $queueForCertificate
-                    ->setDomain(new Document([
-                        'domain' => $rule->getAttribute('domain'),
-                        'domainType' => $rule->getAttribute('deploymentResourceType', $rule->getAttribute('type')),
-                    ]))
-                    ->setAction(Certificate::ACTION_VERIFICATION)
-                    ->trigger();
-            }
-        } else {
-            // Silenced because interval makes it too often
-            // Console::log("[{$time}] No rules for checking verification status.");
-        }
-    }
-
     private function notifyDeleteCSVExports(Delete $queueForDeletes): void
     {
         $queueForDeletes
             ->setType(DELETE_TYPE_CSV_EXPORTS)
             ->trigger();
-    }
-
-    private function renewCertificates(Database $dbForPlatform, Certificate $queueForCertificate): void
-    {
-        $time = DatabaseDateTime::now();
-
-        $certificates = $dbForPlatform->find('certificates', [
-            Query::lessThan('attempts', 5), // Maximum 5 attempts
-            Query::isNotNull('renewDate'),
-            Query::lessThanEqual('renewDate', $time), // includes 60 days cooldown (we have 30 days to renew)
-            Query::limit(200), // Limit 200 comes from LetsEncrypt (300 orders per 3 hours, keeping some for new domains)
-        ]);
-
-
-        if (\count($certificates) > 0) {
-            Console::info("[{$time}] Found " . \count($certificates) . " certificates for renewal, scheduling jobs.");
-
-            foreach ($certificates as $certificate) {
-                $domain = $certificate->getAttribute('domain');
-                if (System::getEnv('_APP_RULES_FORMAT') === 'md5') {
-                    $rule = $dbForPlatform->getDocument('rules', md5($domain));
-                } else {
-                    $rule = $dbForPlatform->findOne('rules', [
-                        Query::equal('domain', [$domain]),
-                    ]);
-                }
-
-                if ($rule->isEmpty() || $rule->getAttribute('region') !== System::getEnv('_APP_REGION', 'default')) {
-                    continue;
-                }
-
-                $queueForCertificate
-                    ->setDomain(new Document([
-                        'domain' => $rule->getAttribute('domain'),
-                        'domainType' => $rule->getAttribute('deploymentResourceType', $rule->getAttribute('type')),
-                    ]))
-                    ->setAction(Certificate::ACTION_GENERATION)
-                    ->trigger();
-            }
-        } else {
-            Console::info("[{$time}] No certificates for renewal.");
-        }
     }
 
     private function notifyDeleteCache($interval, Delete $queueForDeletes): void
--- a/src/Appwrite/Platform/Services/Tasks.php
+++ b/src/Appwrite/Platform/Services/Tasks.php
@@ -4,6 +4,7 @@ namespace Appwrite\Platform\Services;
 
 use Appwrite\Platform\Tasks\Doctor;
 use Appwrite\Platform\Tasks\Install;
+use Appwrite\Platform\Tasks\Interval;
 use Appwrite\Platform\Tasks\Maintenance;
 use Appwrite\Platform\Tasks\Migrate;
 use Appwrite\Platform\Tasks\QueueRetry;
@@ -28,6 +29,7 @@ class Tasks extends Service
         $this
             ->addAction(Doctor::getName(), new Doctor())
             ->addAction(Install::getName(), new Install())
+            ->addAction(Interval::getName(), new Interval())
             ->addAction(Maintenance::getName(), new Maintenance())
             ->addAction(Migrate::getName(), new Migrate())
             ->addAction(QueueRetry::getName(), new QueueRetry())
--- a/src/Appwrite/Platform/Tasks/Interval.php
+++ b/src/Appwrite/Platform/Tasks/Interval.php
@@ -1,0 +1,124 @@
+<?php
+
+namespace Appwrite\Platform\Tasks;
+
+use Appwrite\Event\Certificate;
+use DateTime;
+use Utopia\CLI\Console;
+use Utopia\Database\Database;
+use Utopia\Database\DateTime as DatabaseDateTime;
+use Utopia\Database\Document;
+use Utopia\Database\Query;
+use Utopia\Platform\Action;
+use Utopia\System\System;
+
+class Interval extends Action
+{
+    public static function getName(): string
+    {
+        return 'interval';
+    }
+
+    public function __construct()
+    {
+        $this
+            ->desc('Schedules interval tasks for rules verification and certificate renewal')
+            ->inject('dbForPlatform')
+            ->inject('queueForCertificates')
+            ->callback($this->action(...));
+    }
+
+    public function action(Database $dbForPlatform, Certificate $queueForCertificates): void
+    {
+        Console::title('Interval V1');
+        Console::success(APP_NAME . ' interval process v1 has started');
+
+        $intervalRuleVerification = (int) System::getEnv('_APP_MAINTENANCE_INTERVAL_RULE_VERIFICATION', '60'); // 1 minute
+        $intervalCertificateRenewal = (int) System::getEnv('_APP_MAINTENANCE_INTERVAL', '86400'); // 1 day
+
+        \go(function () use ($dbForPlatform, $queueForCertificates, $intervalRuleVerification) {
+            Console::loop(function () use ($dbForPlatform, $queueForCertificates) {
+                $this->checkRuleVerification($dbForPlatform, $queueForCertificates);
+            }, $intervalRuleVerification);
+        });
+
+        \go(function () use ($dbForPlatform, $queueForCertificates, $intervalCertificateRenewal) {
+            Console::loop(function () use ($dbForPlatform, $queueForCertificates) {
+                $this->renewCertificates($dbForPlatform, $queueForCertificates);
+            }, $intervalCertificateRenewal);
+        });
+    }
+
+    private function checkRuleVerification(Database $dbForPlatform, Certificate $queueForCertificate): void
+    {
+        $time = DatabaseDateTime::now();
+
+        $oldestToCheck = new DateTime('-3 days');
+
+        $rules = $dbForPlatform->find('rules', [
+            Query::createdAfter(DatabaseDateTime::format($oldestToCheck)), // max 3 days old
+            Query::equal('status', [RULE_STATUS_VERIFICATION_FAILED]), // not verified yet
+            Query::orderAsc('$updatedAt'), // Pick the ones waiting for another attempt for longest
+            Query::equal('region', [System::getEnv('_APP_REGION', 'default')]), // Only current region
+            Query::limit(30), // Reasonable pagination limit, processable within a minute
+        ]);
+
+        if (\count($rules) > 0) {
+            Console::info("[{$time}] Found " . \count($rules) . " rules for verification, scheduling jobs.");
+
+            foreach ($rules as $rule) {
+                $queueForCertificate
+                    ->setDomain(new Document([
+                        'domain' => $rule->getAttribute('domain'),
+                        'domainType' => $rule->getAttribute('deploymentResourceType', $rule->getAttribute('type')),
+                    ]))
+                    ->setAction(Certificate::ACTION_VERIFICATION)
+                    ->trigger();
+            }
+        } else {
+            // Silenced because interval makes it too often
+            // Console::log("[{$time}] No rules for checking verification status.");
+        }
+    }
+
+    private function renewCertificates(Database $dbForPlatform, Certificate $queueForCertificate): void
+    {
+        $time = DatabaseDateTime::now();
+
+        $certificates = $dbForPlatform->find('certificates', [
+            Query::lessThan('attempts', 5), // Maximum 5 attempts
+            Query::isNotNull('renewDate'),
+            Query::lessThanEqual('renewDate', $time), // includes 60 days cooldown (we have 30 days to renew)
+            Query::limit(200), // Limit 200 comes from LetsEncrypt (300 orders per 3 hours, keeping some for new domains)
+        ]);
+
+        if (\count($certificates) > 0) {
+            Console::info("[{$time}] Found " . \count($certificates) . " certificates for renewal, scheduling jobs.");
+
+            foreach ($certificates as $certificate) {
+                $domain = $certificate->getAttribute('domain');
+                if (System::getEnv('_APP_RULES_FORMAT') === 'md5') {
+                    $rule = $dbForPlatform->getDocument('rules', md5($domain));
+                } else {
+                    $rule = $dbForPlatform->findOne('rules', [
+                        Query::equal('domain', [$domain]),
+                    ]);
+                }
+
+                if ($rule->isEmpty() || $rule->getAttribute('region') !== System::getEnv('_APP_REGION', 'default')) {
+                    continue;
+                }
+
+                $queueForCertificate
+                    ->setDomain(new Document([
+                        'domain' => $rule->getAttribute('domain'),
+                        'domainType' => $rule->getAttribute('deploymentResourceType', $rule->getAttribute('type')),
+                    ]))
+                    ->setAction(Certificate::ACTION_GENERATION)
+                    ->trigger();
+            }
+        } else {
+            Console::info("[{$time}] No certificates for renewal.");
+        }
+    }
+}


.env
```diff
@@ -21,12 +21,12 @@ _APP_OPTIONS_ROUTER_PROTECTION=disabled
...
 _APP_OPTIONS_ROUTER_FORCE_HTTPS=disabled
 _APP_OPENSSL_KEY_V1=your-secret-key
-_APP_DNS=8.8.8.8
+_APP_DNS=8.8.8.8,1.1.1.1 
 _APP_DOMAIN=traefik
 _APP_CONSOLE_DOMAIN=localhost
 _APP_DOMAIN_FUNCTIONS=functions.localhost
 _APP_DOMAIN_SITES=sites.localhost
-_APP_DOMAIN_TARGET_CNAME=test.localhost
+_APP_DOMAIN_TARGET_CNAME=test.appwrite.org
```
Meldiron: Intentional; without this testing locally is not possible, worker throws when seing non-public hostname, blocking most of flow

app/config/collections/platform.php
```diff
@@ -1185,6 +1185,17 @@
                 'array' => false,
                 'filters' => [],
             ],
+            [
+                '$id' => ID::custom('verificationLogs'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 2048,
```
abnegate: What kind of logs? Possible that it'll ever exceed 2K chars? Safer to go with huge sizes, no real downside

src/Appwrite/Event/Certificate.php
```diff
@@ -7,9 +7,14 @@
...
 class Certificate extends Event
 {
+    public const ACTION_VERIFICATION = 'verification';
+    public const ACTION_GENERATION = 'generation';
+
     protected bool $skipRenewCheck = false;
     protected ?Document $domain = null;
-    protected ?string $validationDomain = null;
+    protected ?string $verificationDomainFunction = null; // For example: fra.cloud.appwrite.io
+    protected ?string $verificationDomainApi = null; // For example: fra.appwrite.run
```
abnegate: Let's use uppercase for API (contrary to what we have done in the past as that was an error)

src/Appwrite/Platform/Modules/Proxy/Base.php
```diff
@@ -0,0 +1,130 @@
...
+use Appwrite\Network\Validator\DNS;
+use Appwrite\Platform\Action;
+use Utopia\Database\Document;
+use Utopia\Domains\Domain;
+use Utopia\Logger\Log;
+use Utopia\System\System;
+use Utopia\Validator\AnyOf;
+use Utopia\Validator\IP;
+
+class Base extends Action
```
abnegate: Pattern we've used elsewhere is to still call this `Action` and import alias the Appwrite one as `AppwriteAction`, then in extending classes you only need to change the import to change the base class

src/Appwrite/Platform/Modules/Proxy/Http/Rules/API/Create.php
```diff
@@ -66,11 +65,19 @@ public function __construct()
             ->inject('queueForCertificates')
             ->inject('queueForEvents')
             ->inject('dbForPlatform')
+            ->inject('log')
```
abnegate, line 69: Do we need this? I don't see it used

src/Appwrite/Platform/Modules/Proxy/Http/Rules/Get.php
```diff
@@ -64,9 +64,13 @@ public function action(
             throw new Exception(Exception::RULE_NOT_FOUND);
         }
 
+        // Fill response model
         $certificate = $dbForPlatform->getDocument('certificates', $rule->getAttribute('certificateId', ''));
         $rule->setAttribute('logs', $certificate->getAttribute('logs', ''));
         $rule->setAttribute('renewAt', $certificate->getAttribute('renewDate', ''));
+        if (new \DateTime($certificate->getUpdatedAt()) > new \DateTime($rule->getUpdatedAt())) {
+            $rule->setAttribute('$updatedAt', $certificate->getUpdatedAt());
```
abnegate: Why do we do this? I don't see a call to updateDocument that will persist the change

src/Appwrite/Platform/Modules/Proxy/Http/Rules/Verification/Update.php
```diff
@@ -81,92 +78,22 @@ public function action(
...
-            $error = \implode("\n", $errors);
-            $log->addExtra('dnsResponse', \is_array($error) ? \json_encode($error) : \strval($error));
-
-            throw new Exception(Exception::RULE_VERIFICATION_FAILED);
+        try {
+            self::verifyRule($rule, $log);
+            $rule = $rule->setAttribute('verificationLogs', '');
+        } catch (Exception $err) {
+            $rule = $rule->setAttribute('verificationLogs', $err->getMessage());
+            $dbForPlatform->updateDocument('rules', $rule->getId(), $rule);
```
abnegate: Let's use partial updates throughout

src/Appwrite/Platform/Modules/Proxy/Http/Rules/Verification/Update.php
```diff
@@ -178,8 +105,13 @@ public function action(
 
         $queueForEvents->setParam('ruleId', $rule->getId());
 
+        // Fill response model
         $certificate = $dbForPlatform->getDocument('certificates', $rule->getAttribute('certificateId', ''));
         $rule->setAttribute('logs', $certificate->getAttribute('logs', ''));
+        $rule->setAttribute('renewAt', $certificate->getAttribute('renewDate', ''));
+        if (new \DateTime($certificate->getUpdatedAt()) > new \DateTime($rule->getUpdatedAt())) {
+            $rule->setAttribute('$updatedAt', $certificate->getUpdatedAt());
+        }
```
abnegate: Same here, no update call after changing

src/Appwrite/Platform/Tasks/Maintenance.php
```diff
@@ -59,43 +61,51 @@ public function action(Database $dbForPlatform, Document $console, Certificate $
...
-                ->setUsageRetentionHourlyDateTime(DatabaseDateTime::addSeconds(new \DateTime(), -1 * $usageStatsRetentionHourly))
-                ->trigger();
-
-            $this->notifyDeleteConnections($queueForDeletes);
-            $this->renewCertificates($dbForPlatform, $queueForCertificates);
-            $this->notifyDeleteCache($cacheRetention, $queueForDeletes);
-            $this->notifyDeleteSchedules($schedulesDeletionRetention, $queueForDeletes);
-        }, $interval, $delay);
+        \go(function () use ($interval, $cacheRetention, $schedulesDeletionRetention, $usageStatsRetentionHourly, $dbForPlatform, $console, $queueForDeletes, $queueForCertificates, $delay) {
+            Console::loop(function () use ($interval, $cacheRetention, $schedulesDeletionRetention, $usageStatsRetentionHourly, $dbForPlatform, $console, $queueForDeletes, $queueForCertificates) {
```
abnegate: Not sure if this work as expected, Console::loop probably uses raw sleep, not \Co::sleep. Raw sleep will block _all_ coroutines, not just this one

src/Appwrite/Platform/Tasks/Maintenance.php
```diff
@@ -106,6 +116,42 @@ private function notifyDeleteConnections(Delete $queueForDeletes): void
             ->trigger();
     }
 
+    private function checkRuleVerification(Database $dbForPlatform, Certificate $queueForCertificate): void
+    {
+        $time = DatabaseDateTime::now();
+
+        $oldestToCheck = new DateTime();
+        $oldestToCheck->sub(new DateInterval('PT72H')); // -3 days
```
abnegate: A little bit magic but you could just do `$oldestToCheck = new DateTime('-3 days');` here

app/config/collections/platform.php
Meldiron: Exception from DNS validator. It would only exceed if developer had like 10 records.

Setting to 65k, I dont see value in higher.

src/Appwrite/Platform/Modules/Proxy/Base.php
Meldiron: Refactored proxy and compute (sites&functions)

src/Appwrite/Platform/Modules/Proxy/Http/Rules/API/Create.php
Meldiron: It's used in `self::verifyRule`, a helper method for verifying rule's DNS settings. Inside we track some details like how long it took, or some errors. useful in sentry to debug a problem

src/Appwrite/Platform/Modules/Proxy/Http/Rules/Get.php
Meldiron: Like we add logs and renewAt from certificate, we also want to override $updatedAt.

During verification, rule document gets updates.
During generation, certificate document gets updated.

In UI, we want dates like these

CleanShot 2025-09-12 at 16 15 19@2x

Updating responses to have most recent updatedAt of either rule or it's certificate allows the date to get updated properly during certificate generation too.

Ideally, in future, we remove certificates table and just use rules. Like we deleted builds and we have all in deployments

src/Appwrite/Platform/Tasks/Maintenance.php
Meldiron: If swoole hooks are enabled (they are), sleep is overriden to only block current coroutine.

I did QA locally and both loops run okay. Only possible worry would be memory cleanup issues, but those are hard to guess with corutines anyway. we can react to those when we see them.

.env
```diff
@@ -21,12 +21,12 @@ _APP_OPTIONS_ROUTER_PROTECTION=disabled
...
 _APP_OPTIONS_ROUTER_FORCE_HTTPS=disabled
 _APP_OPENSSL_KEY_V1=your-secret-key
-_APP_DNS=8.8.8.8
+_APP_DNS=8.8.8.8,1.1.1.1
 _APP_DOMAIN=traefik
 _APP_CONSOLE_DOMAIN=localhost
 _APP_DOMAIN_FUNCTIONS=functions.localhost
 _APP_DOMAIN_SITES=sites.localhost
-_APP_DOMAIN_TARGET_CNAME=test.localhost
+_APP_DOMAIN_TARGET_CNAME=test.l
```
vermakhushboo: Intended?

src/Appwrite/Network/Validator/DNS.php
```diff
@@ -14,34 +15,86 @@ class DNS extends Validator
...
+
+        if ($this->count === 1) {
+            $messages[] = "Domain {$this->domain} has incorrect {$this->type} value '{$recordValuesVerbose}'";
+        } else {
+            // Two or more
+            $messages[] = "Domain {$this->domain} has {$countVerbose} incompatible {$this->type} records: '{$recordValuesVerbose}'";
+        }
+
+        if ($this->type === self::RECORD_CAA) {
+            $messages[] = 'Add new CAA record, or removing all other CAA records';
```
vermakhushboo: Should this be `or remove` instead?

src/Appwrite/Platform/Workers/Certificates.php
```diff
@@ -80,26 +82,107 @@ public function action(
...
-        $domainType = $document->getAttribute('domainType');
+        switch ($action) {
+            case Certificate::ACTION_GENERATION:
+                $this->executeGeneration($domain, $domainType, $dbForPlatform, $queueForMails, $queueForEvents, $queueForWebhooks, $queueForFunctions, $queueForRealtime, $log, $certificates, $skipRenewCheck, $plan, $verificationDomainFunction, $verificationDomainAPI);
+                break;
+            case Certificate::ACTION_VERIFICATION:
+                $this->executeVerification($domain, $dbForPlatform, $log, $queueForCertificates, $verificationDomainFunction, $verificationDomainAPI);
+                break;
+            default:
+                throw new Exception('Invalid action . ' . $action);
```
vermakhushboo:
```suggestion
                throw new Exception('Invalid action. ' . $action);
```

src/Appwrite/Platform/Workers/Certificates.php
```diff
@@ -80,26 +82,107 @@ public function action(
...
+        $domain = new Domain($document->getAttribute('domain', ''));
+        $domainType = $document->getAttribute('domainType');
+
         $skipRenewCheck = $payload['skipRenewCheck'] ?? false;
-        $validationDomain = $payload['validationDomain'] ?? null;
+        $action = $payload['action'] ?? Certificate::ACTION_GENERATION;
+        $verificationDomainFunction = $payload['verificationDomainFunction'] ?? null;
+        $verificationDomainAPI = $payload['verificationDomainAPI'] ?? null;
+
+        Console::log('Recieved ' . $action . ' action for ' . $domain->get() . ' domain');
```
vermakhushboo:
```suggestion
        Console::log('Received ' . $action . ' action for ' . $domain->get() . ' domain');
```

.env
Meldiron: Mistake, fixed

src/Appwrite/Network/Validator/DNS.php
Meldiron: nice catch

Aglowharmony (approved):
Works ###

docker-compose.yml
```diff
@@ -219,7 +222,7 @@ services:
   appwrite-console:
     <<: *x-logging
     container_name: appwrite-console
-    image: appwrite/console:7.0.2
+    image: appwrite/console:7.1.8
```
vermakhushboo: Need to update this to latest

src/Appwrite/Event/Certificate.php
```diff
@@ -7,9 +7,14 @@
 
 class Certificate extends Event
 {
+    public const ACTION_VERIFICATION = 'verification';
+    public const ACTION_GENERATION = 'generation';
+
     protected bool $skipRenewCheck = false;
     protected ?Document $domain = null;
-    protected ?string $validationDomain = null;
+    protected ?string $verificationDomainFunction = null; // For example: fra.cloud.appwrite.io
```
vermakhushboo, line 15: Do we also need for sites?

src/Appwrite/Platform/Modules/Proxy/Action.php
```diff
@@ -0,0 +1,144 @@
...
+        $targetCNAME = null;
+        $ruleType = $rule->getAttribute('type', '');
+        $resourceType = $rule->getAttribute('deploymentResourceType', '');
+
+        if ($resourceType === 'function') {
+            // For example: fra.appwrite.run
+            $targetCNAME = new Domain($verificationDomainFunction);
+        } elseif ($resourceType === 'site') {
+            // For example: appwrite.network
+            $targetCNAME = new Domain(System::getEnv('_APP_DOMAIN_SITES', ''));
```
vermakhushboo, line 80: Why not create a variable like functions and API?

src/Appwrite/Event/Certificate.php
```diff
@@ -7,9 +7,14 @@
...
+    protected ?string $verificationDomainAPI = null; // For example: fra.appwrite.run
```
vermakhushboo, line 16: Incorrect values in examples, need to swap

tests/resources/coredns/webapp.com.zone
```diff
@@ -0,0 +1,23 @@
+; Nessessary setup
```
vermakhushboo:
```suggestion
; Necessary setup
```

tests/resources/coredns/wrong-a-webapp.com.zone
```diff
@@ -0,0 +1,8 @@
+; Nessessary setup
```
vermakhushboo:
```suggestion
; Necessary setup
```

src/Appwrite/Utopia/Response/Model/Rule.php
```diff
@@ -84,11 +84,16 @@ public function __construct()
...
-                'description' => 'Domain verification status. Possible values are "created", "verifying", "verified" and "unverified"',
-                'default' => 'created',
-                'example' => 'verified',
-                'enum' => ['created', 'verifying', 'verified', 'unverified'],
+                'type' => self::TYPE_STRING,
+                'description' => 'Domain verification status. Possible values are "' . RULE_STATUS_VERIFICATION_FAILED . '", "' . RULE_STATUS_GENERATING_CERTIFICATE . '", "' . RULE_STATUS_SUCCESSFUL . '" and "' . RULE_STATUS_GENERATION_FAILED . '"',
+                'default' => '',
+                'example' => RULE_STATUS_SUCCESSFUL,
+            ])
+            ->addRule('verificationLogs', [
```
eldadfux: Can this be just simply `logs`?

src/Appwrite/Platform/Tasks/Maintenance.php
```diff
@@ -106,6 +116,38 @@ private function notifyDeleteConnections(Delete $queueForDeletes): void
             ->trigger();
     }
 
+    private function checkRuleVerification(Database $dbForPlatform, Certificate $queueForCertificate): void
+    {
+        $time = DatabaseDateTime::now();
+
+        $oldestToCheck = new DateTime('-3 days');
+
+        $rules = $dbForPlatform->find('rules', [
```
eldadfux: Can we explain what situation this covers?

src/Appwrite/Platform/Tasks/Maintenance.php
```diff
@@ -59,43 +61,51 @@ public function action(Database $dbForPlatform, Document $console, Certificate $
...
-                ->setProject($console)
-                ->setUsageRetentionHourlyDateTime(DatabaseDateTime::addSeconds(new \DateTime(), -1 * $usageStatsRetentionHourly))
-                ->trigger();
-
-            $this->notifyDeleteConnections($queueForDeletes);
-            $this->renewCertificates($dbForPlatform, $queueForCertificates);
-            $this->notifyDeleteCache($cacheRetention, $queueForDeletes);
-            $this->notifyDeleteSchedules($schedulesDeletionRetention, $queueForDeletes);
-        }, $interval, $delay);
+        \go(function () use ($interval, $cacheRetention, $schedulesDeletionRetention, $usageStatsRetentionHourly, $dbForPlatform, $console, $queueForDeletes, $queueForCertificates, $delay) {
```
eldadfux: Not sure this is a great implementation. Should we have multiple containers for maintenance each running x jobs at y interval. This structure is a prep for disaster. In microservices/container, each service/container needs to be one process.

app/init/constants.php
```diff
@@ -138,7 +138,11 @@
 const DELETE_TYPE_EXPIRED_TARGETS = 'invalid_targets';
 const DELETE_TYPE_SESSION_TARGETS = 'session_targets';
 const DELETE_TYPE_MAINTENANCE = 'maintenance';
-
+// Rule types
+const RULE_STATUS_VERIFICATION_FAILED = 'created';
```
eldadfux, line 144: The fact that the constant and values has different names is not great and very confusing.

src/Appwrite/Network/Validator/DNS.php
```diff
@@ -14,34 +15,86 @@ class DNS extends Validator
...
+    public string $resolver = '';
+    public array $recordValues = [];
+    public int $count = 0;
+    public string $reason = '';
 
     /**
      * @param string $target
      */
-    public function __construct(protected string $target, protected string $type = self::RECORD_CNAME, string $dnsServer = '')
+    public function __construct(protected string $target, protected string $type = self::RECORD_CNAME, string $dnsServers = '')
```
eldadfux: This validator needs to be in our DNS library and be unit tested. We shouldn't couple validators with Swoole specific features (WaitGroup), we can do that from inside the app context (route/worker/task/etc), would be great to abstract that as well, but this can be a separate problem.

src/Appwrite/Platform/Modules/Proxy/Http/Rules/API/Create.php
```diff
@@ -66,11 +66,19 @@ public function __construct()
...
-    public function action(string $domain, Response $response, Document $project, Certificate $queueForCertificates, Event $queueForEvents, Database $dbForPlatform)
-    {
+    public function action(
+        string $domain,
+        Response $response,
+        Document $project,
+        Certificate $queueForCertificates,
+        Event $queueForEvents,
+        Database $dbForPlatform,
+        Log $log
```
eldadfux, line 80: What is $log, why do we need it here?

src/Appwrite/Platform/Modules/Proxy/Http/Rules/API/Create.php
```diff
@@ -176,13 +162,22 @@ public function action(string $domain, Response $response, Document $project, Ce
             'region' => $project->getAttribute('region')
         ]);
 
+        if ($rule->getAttribute('status', '') === RULE_STATUS_VERIFICATION_FAILED) {
+            try {
+                $this->verifyRule($rule, $log);
```
eldadfux, line 167: Would be cool if this could have been a validator, maybe even a generic enough to be on utopia

src/Appwrite/Platform/Modules/Proxy/Http/Rules/Get.php
```diff
@@ -64,10 +64,23 @@ public function action(
             throw new Exception(Exception::RULE_NOT_FOUND);
         }
 
+        // Fill response model
         $certificate = $dbForPlatform->getDocument('certificates', $rule->getAttribute('certificateId', ''));
         $rule->setAttribute('logs', $certificate->getAttribute('logs', ''));
         $rule->setAttribute('renewAt', $certificate->getAttribute('renewDate', ''));
 
+        $certificateHasUpdatedAt = $certificate->getUpdatedAt() !== null;
```
eldadfux: Why are we doing this?!

src/Appwrite/Platform/Modules/Proxy/Http/Rules/XList.php
```diff
@@ -104,10 +104,24 @@ public function action(
...
 
         $rules = $dbForPlatform->find('rules', $queries);
+
+        // Fill response model
         foreach ($rules as $rule) {
             $certificate = $dbForPlatform->getDocument('certificates', $rule->getAttribute('certificateId', ''));
             $rule->setAttribute('logs', $certificate->getAttribute('logs', ''));
             $rule->setAttribute('renewAt', $certificate->getAttribute('renewDate', ''));
+
+            $certificateHasUpdatedAt = $certificate->getUpdatedAt() !== null;
```
eldadfux: Why are we doing this?!

src/Appwrite/Platform/Modules/Proxy/Action.php
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * Verify or re-verify a rule
+     *
+     * @param Document $rule Rule to verify
+     * @param Log|null $log Log instance to add timings to
+     * @param string|null $verificationDomainAPI Override for expected API rule value during verification
+     * @param string|null $verificationDomainFunction Override for expected Function rule value during verification
+     * @return void
+     */
+    public function verifyRule(Document $rule, ?Log $log = null, ?string $verificationDomainAPI = null, ?string $verificationDomainFunction = null): void
```
eldadfux, line 31: Classic as a rule that can also be unit tested.

src/Appwrite/Platform/Workers/Certificates.php
```diff
@@ -41,8 +41,10 @@ public static function getName(): string
     /**
      * @throws Exception
      */
-    public function __construct()
```
eldadfux, line 45: We need a docblock here to explain our flows.

eldadfux (changes requested):

src/Appwrite/Utopia/Response/Model/Rule.php
Meldiron: We use logs for certificate logs. but yes, technically you dont ever see certificate logs before verification finishes. so we could re-use same attribute 👍

src/Appwrite/Platform/Tasks/Maintenance.php
Meldiron: This runs validation atuomatically in backgroud. Let's say you add domain, verification fails. you add dns records - it stillfails. All you need to do now is wait.

So while you wait, we keep re-trying in backgrounf, for up to 3 days. Benefits:
- Once verification passes in background, we automatically go to certiifcate generation. Allows you to leave, and it will eventuelly be done with no more input from you
- If you stay in Console, realtime you see "last checked 1 minute ago" updating. So if it's in 2nd tab, you come back after 10 mins, you see we are still checking it for you
- In future, we could send email when it finishes successfully (including certificate), or send failre email if after 3 days we still couldnt verify.

app/init/constants.php
Meldiron: agree but I really dont want to do ZDT migration to change status right now. I want this live asap, const with proper name was best middle-ground to make status understandable when reading code

src/Appwrite/Network/Validator/DNS.php
Meldiron: for now, let's move it to library, and dont support concurrency there.
then extend it in Appwrite, adding concurrency support for multiple dns servers instead of just one

src/Appwrite/Platform/Modules/Proxy/Http/Rules/API/Create.php
Meldiron: We log some timings. I think luke added it couple of months ago, probably to debug some timeouts.
Let's check with Luke, see if we still need it. If not, lets remove from verywhere

src/Appwrite/Platform/Modules/Proxy/Http/Rules/Get.php
Meldiron: we used to have 2 tables - deployments and builds. we merged them because it made no sense to separate.

Same here - certificate and rule are one-to-one, no benefit to separation.
Problem is if we update certificate (status or logs), nothing changes in rule. so when we listRules, it wont see updates to certiifcates.

Doing this ensures we show most up-to-date time. Otherwise you will not see any updates to updatedAt during certificate generation (when we keep re-checking status with CDN)

src/Appwrite/Event/Certificate.php
Meldiron: Sites is same across all of appwrite ,functions and API are region specific.
We dont need one for sites

src/Appwrite/Platform/Modules/Proxy/Action.php
Meldiron: Should not be needed, we only have one domain for sites. function/api is per-region

src/Appwrite/Platform/Modules/Proxy/Action.php
```diff
@@ -0,0 +1,147 @@
...
+     * Parse DNS servers from environment variable
+     *
+     * @return array<string>
+     */
+    protected function getDNSServers(): array
+    {
+        $dnsEnv = System::getEnv('_APP_DNS', '8.8.8.8');
+        $servers = \array_map('trim', \explode(',', $dnsEnv));
+        return \array_filter($servers, fn ($server) => !empty($server));
+    }
```
Meldiron: I think this is only used in one place, we dont need helper function for it - we can do it directly in code and have comment saying what it does.

I also think we can simplify - we dont need to filter out empty.

src/Appwrite/Network/Validator/DNS.php
```diff
@@ -2,115 +2,93 @@
...
+    protected array $dnsServers = [];
+
+    /**
+     * @param string $target Expected value for the DNS record
+     * @param int $type Type of DNS record to validate
+     *  For value, use const from Record, such as Record::TYPE_A
+     *  When using CAA type, you can provide exact match, or just issuer domain as $target
+     * @param string|array<string> $dnsServers DNS server IP(s) or domain(s) to use for validation
+     */
+    public function __construct(string $target, int $type = Record::TYPE_CNAME, string|array $dnsServers = self::DEFAULT_DNS_SERVER)
```
Meldiron: I would only make it accept array, for simplicity

src/Appwrite/Network/Validator/DNS.php
```diff
@@ -2,115 +2,93 @@
...
 
-    private function extractCAAValue(string $rdata): string
+    /**
+     * Validate with a specific DNS server
+     *
+     * @param mixed $value
+     * @param string $dnsServer
+     * @return bool
+     */
+    protected function isValidWithDNSServer(mixed $value, string $dnsServer): bool
```
Meldiron: Same as before, seems like helper method only used once. We dont need to abstract here.

src/Appwrite/Network/Validator/DNS.php
```diff
@@ -2,115 +2,93 @@
...
-        $client = new Client($this->server);
-        try {
-            $response = $client->query(Message::query(
-                new Question($value, $this->type)
-            ));
-        } catch (\Throwable) {
-            return false;
+        if (\count($this->dnsServers) === 1) {
+            return $this->isValidWithDNSServer($value, $this->dnsServers[0]);
         }
```
Meldiron: We might not need this, technically, our coroutine approach should work fine and quick even for just 1 dns server

src/Appwrite/Platform/Modules/Proxy/Http/Rules/Get.php
vermakhushboo: We now store latest `updatedAt` from rules and certificates in rules collection.

src/Appwrite/Platform/Modules/Proxy/Http/Rules/API/Create.php
```diff
@@ -188,6 +188,22 @@ public function action(
...
+        $certificateLogs = $certificate->getAttribute('logs', '');
+        $logs = '';
+        if (!empty($certificateLogs) && !empty($ruleLogs)) {
+            $logs = $certificateLogs; // Certificate logs have priority
+        } elseif (!empty($certificateLogs)) {
+            $logs = $certificateLogs;
+        } elseif (!empty($ruleLogs)) {
+            $logs = $ruleLogs;
+        }
+        $rule->setAttribute('logs', $logs);
```
Meldiron, line 204: Can be simplified to: `$logs = !empty($certificateLogs) ? $certificateLogs :  $ruleLogs` in all places

src/Appwrite/Platform/Modules/Proxy/Http/Rules/Redirect/Create.php
```diff
@@ -215,6 +215,22 @@ public function action(
...
+        $certificateLogs = $certificate->getAttribute('logs', '');
+        $logs = '';
+        if (!empty($certificateLogs) && !empty($ruleLogs)) {
+            $logs = $certificateLogs; // Certificate logs have priority
+        } elseif (!empty($certificateLogs)) {
+            $logs = $certificateLogs;
+        } elseif (!empty($ruleLogs)) {
+            $logs = $ruleLogs;
+        }
+        $rule->setAttribute('logs', $logs);
```
Meldiron, line 231: In create endpoints we can skip this - certificate is only created after rule, so at this moment, it will always be empty

src/Appwrite/Platform/Workers/Certificates.php
```diff
@@ -508,7 +508,8 @@ private function updateDomainDocuments(
 
             $updates
                 ->setAttribute('certificateId', $certificateId)
-                ->setAttribute('status', $status);
+                ->setAttribute('status', $status)
+                ->setAttribute('$updatedAt', DateTime::now());
```
Meldiron, line 512: We can double check with DB team, but I believe updatedAt is automatically increased when we updateDocument.

I think we can remove this - and if we have problem with non-updated time, it might be coming from another place.

PR #10486, opened 2025-09-13
Add branch-only deployments to Sites without ref tags

atharvadeosthale:
Added new route - POST `/v1/sites/direct` which accepts `owner`, `repository`, `branch`

### DIFF
--- a/src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
+++ b/src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
@@ -311,19 +311,27 @@ class Builds extends Action
                 $templateRepositoryName = $template->getAttribute('repositoryName', '');
                 $templateOwnerName = $template->getAttribute('ownerName', '');
                 $templateVersion = $template->getAttribute('version', '');
+                $templateBranch = $template->getAttribute('branch', '');
 
                 $templateRootDirectory = $template->getAttribute('rootDirectory', '');
                 $templateRootDirectory = \rtrim($templateRootDirectory, '/');
                 $templateRootDirectory = \ltrim($templateRootDirectory, '.');
                 $templateRootDirectory = \ltrim($templateRootDirectory, '/');
 
-                if (!empty($templateRepositoryName) && !empty($templateOwnerName) && !empty($templateVersion)) {
+                if (!empty($templateRepositoryName) && !empty($templateOwnerName) && (!empty($templateVersion) || !empty($templateBranch))) {
                     $stdout = '';
                     $stderr = '';
 
                     // Clone template repo
                     $tmpTemplateDirectory = '/tmp/builds/' . $deploymentId . '-template';
-                    $gitCloneCommandForTemplate = $github->generateCloneCommand($templateOwnerName, $templateRepositoryName, $templateVersion, GitHub::CLONE_TYPE_TAG, $tmpTemplateDirectory, $templateRootDirectory);
+                    
+                    if(empty($templateVersion)) {
+                        $gitCloneCommandForTemplate = $github->generateCloneCommand($templateOwnerName, $templateRepositoryName, $templateBranch, GitHub::CLONE_TYPE_BRANCH, $tmpTemplateDirectory, $templateRootDirectory);
+                    } else {
+                        // True template
+                        $gitCloneCommandForTemplate = $github->generateCloneCommand($templateOwnerName, $templateRepositoryName, $templateVersion, GitHub::CLONE_TYPE_TAG, $tmpTemplateDirectory, $templateRootDirectory);
+                    }
+
                     $exit = Console::execute($gitCloneCommandForTemplate, '', $stdout, $stderr);
 
                     if ($exit !== 0) {
--- a/app/config/specs/open-api3-latest-console.json
+++ b/app/config/specs/open-api3-latest-console.json
@@ -4888,7 +4888,7 @@
                 "x-appwrite": {
                     "method": "getResource",
                     "group": null,
-                    "weight": 496,
+                    "weight": 497,
                     "cookies": false,
                     "type": "",
                     "demo": "console\/get-resource.md",
@@ -28209,7 +28209,7 @@
                 "x-appwrite": {
                     "method": "listRules",
                     "group": null,
-                    "weight": 502,
+                    "weight": 503,
                     "cookies": false,
                     "type": "",
                     "demo": "proxy\/list-rules.md",
@@ -28283,7 +28283,7 @@
                 "x-appwrite": {
                     "method": "createAPIRule",
                     "group": null,
-                    "weight": 497,
+                    "weight": 498,
                     "cookies": false,
                     "type": "",
                     "demo": "proxy\/create-api-rule.md",
@@ -28350,7 +28350,7 @@
                 "x-appwrite": {
                     "method": "createFunctionRule",
                     "group": null,
-                    "weight": 499,
+                    "weight": 500,
                     "cookies": false,
                     "type": "",
                     "demo": "proxy\/create-function-rule.md",
@@ -28428,7 +28428,7 @@
                 "x-appwrite": {
                     "method": "createRedirectRule",
                     "group": null,
-                    "weight": 500,
+                    "weight": 501,
                     "cookies": false,
                     "type": "",
                     "demo": "proxy\/create-redirect-rule.md",
@@ -28541,7 +28541,7 @@
                 "x-appwrite": {
                     "method": "createSiteRule",
                     "group": null,
-                    "weight": 498,
+                    "weight": 499,
                     "cookies": false,
                     "type": "",
                     "demo": "proxy\/create-site-rule.md",
@@ -28619,7 +28619,7 @@
                 "x-appwrite": {
                     "method": "getRule",
                     "group": null,
-                    "weight": 501,
+                    "weight": 502,
                     "cookies": false,
                     "type": "",
                     "demo": "proxy\/get-rule.md",
@@ -28670,7 +28670,7 @@
                 "x-appwrite": {
                     "method": "deleteRule",
                     "group": null,
-                    "weight": 503,
+                    "weight": 504,
                     "cookies": false,
                     "type": "",
                     "demo": "proxy\/delete-rule.md",
@@ -28730,7 +28730,7 @@
                 "x-appwrite": {
                     "method": "updateRuleVerification",
                     "group": null,
-                    "weight": 504,
+                    "weight": 505,
                     "cookies": false,
                     "type": "",
                     "demo": "proxy\/update-rule-verification.md",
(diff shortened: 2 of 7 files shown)


src/Appwrite/Platform/Modules/Functions/Http/Deployments/Template/Create.php
```diff
@@ -65,7 +66,9 @@ public function __construct()
             ->param('repository', '', new Text(128, 0), 'Repository name of the template.')
             ->param('owner', '', new Text(128, 0), 'The name of the owner of the template.')
             ->param('rootDirectory', '', new Text(128, 0), 'Path to function code in the template repo.')
-            ->param('version', '', new Text(128, 0), 'Version (tag) for the repo linked to the function template.')
+            ->param('version', '', new Text(128, 0), 'Version (tag) for the repo linked to the function template.', true)
```
Meldiron: I think we can remove this parameter now.

When type=tag, then refference is what used to be version.

Since its breaking change, let's include this in a request filter to make it non-breaking (in request filter, add type to be tag, and refference to be what version is. and then delete version attribute

src/Appwrite/Platform/Modules/Functions/Http/Deployments/Template/Create.php
```diff
@@ -100,11 +105,22 @@ public function action(
             throw new Exception(Exception::FUNCTION_NOT_FOUND);
         }
 
+        if (empty($version) && empty($type) && empty($reference)) {
+            throw new Exception("Either version or type & reference must be provided");
+        }
```
Meldiron: With proper param validator, and request filters, we shouldnt need this check

src/Appwrite/Platform/Modules/Functions/Http/Deployments/Template/Create.php
```diff
@@ -100,11 +105,22 @@ public function action(
...
+
+        $referenceType = !empty($version) ? GitHub::CLONE_TYPE_TAG : $type;
+        $referenceValue = !empty($version) ? $version : $reference;
+
```
Meldiron: Same comment as above; request filters can avoid need for this

src/Appwrite/Platform/Modules/Functions/Http/Deployments/Template/Create.php
```diff
@@ -146,7 +162,12 @@ function: $function,
             'resourceType' => 'functions',
             'entrypoint' => $function->getAttribute('entrypoint', ''),
             'buildCommands' => $function->getAttribute('commands', ''),
-            'type' => 'manual',
+            'providerRepositoryName' => $repository,
+            'providerRepositoryOwner' => $owner,
+            'providerRepositoryUrl' => $repositoryUrl,
+            'providerBranchUrl' => $branchUrl,
+            'providerBranch' => $type == GitHub::CLONE_TYPE_BRANCH ? $referenceValue : '',
+            'type' => 'vcs',
```
Meldiron, line 162: Interesting problem, let's have Discord thread and involve Jake, Luke, Matej (maybe Eldad)

What type should be deployment made from external source, considering only external source is GitHub?

Should we have new type for this, do manual, or do VCS?

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
```diff
@@ -310,20 +310,23 @@ protected function buildDeployment(
...
-                if (!empty($templateRepositoryName) && !empty($templateOwnerName) && !empty($templateVersion)) {
+                if (!empty($templateRepositoryName) && !empty($templateOwnerName) && !empty($templateReferenceType) && !empty($templateReferenceValue)) {
                     $stdout = '';
                     $stderr = '';
 
                     // Clone template repo
                     $tmpTemplateDirectory = '/tmp/builds/' . $deploymentId . '-template';
-                    $gitCloneCommandForTemplate = $github->generateCloneCommand($templateOwnerName, $templateRepositoryName, $templateVersion, GitHub::CLONE_TYPE_TAG, $tmpTemplateDirectory, $templateRootDirectory);
+
+                    $gitCloneCommandForTemplate = $github->generateCloneCommand($templateOwnerName, $templateRepositoryName, $templateReferenceValue, $templateReferenceType, $tmpTemplateDirectory, $templateRootDirectory);
```
Meldiron, line 328: Let's include different tests to add well-covering E2E tests for template deployments (functions & sites)

src/Appwrite/Utopia/Request/Filters/V21.php
```diff
@@ -0,0 +1,34 @@
...
+     */
+    protected function convertVersionToTypeAndReference(array $content): array
+    {
+        if (!empty($content['version'])) {
+            $content['type'] = 'tag';
+            $content['reference'] = $content['version'];
+            unset($content['version']);
+        }
+        return $content;
+    }
```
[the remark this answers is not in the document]
atharvadeosthale: We aren't even expecting tag and version because this is for backward compat

app/controllers/general.php
```diff
@@ -906,6 +907,9 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
                 $dbForProject = $getProjectDB($project);
                 $request->addFilter(new RequestV20($dbForProject, $route->getPathValues($request)));
             }
+            if (version_compare($requestFormat, '1.8.1', '<')) {
+                $request->addFilter(new RequestV21());
+            }
```
Meldiron: @abnegate is this okay to do, or should we initiate 1.9.0 for this?

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -502,6 +503,120 @@ public function testCreateFunctionAndDeploymentFromTemplate()
...
+            ]
+        );
+
+        $this->assertEquals(202, $deployment['headers']['status-code']);
+        $this->assertNotEmpty($deployment['body']['$id']);
+
+        $deployment = $this->getDeployment($functionId, $deployment['body']['$id']);
+        $this->assertEquals(200, $deployment['headers']['status-code']);
+
+        $function = $this->cleanupFunction($functionId);
```
Meldiron: Let's also wait and ensure this deploys properly - status eventually becomes "ready". also would make sense to ensure sourceSIze and buildSize are both non-zero.

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
```diff
@@ -502,6 +503,120 @@ public function testCreateFunctionAndDeploymentFromTemplate()
         $function = $this->cleanupFunction($functionId);
     }
 
+    public function testCreateFunctionAndDeploymentFromTemplateBranch()
+    {
+        $starterTemplate = $this->getTemplate('starter');
+        $this->assertEquals(200, $starterTemplate['headers']['status-code']);
+
+        $phpRuntime = array_values(array_filter($starterTemplate['body']['runtimes'], function ($runtime) {
```
Meldiron: just $runtime would be enough here. or if needed to be specific, let switch to $nodeRuntime

tests/e2e/Services/Sites/SitesBase.php
```diff
@@ -329,6 +329,35 @@ protected function getTemplate(string $templateId)
             'content-type' => 'application/json',
             'x-appwrite-project' => $this->getProject()['$id'],
         ]);
+
+        // Fetch latest commit from GitHub API if template has provider info
+        if (
```
Meldiron: getTempalte functions is meant to be exact representation of endpoint. doing this affects the logic and changes response format.

Let's do this in a separate helper function, maybe `helperGetLatestCommit(string $owner, string $repository)`

tests/e2e/Services/Functions/FunctionsBase.php
```diff
@@ -268,6 +268,34 @@ protected function getTemplate(string $templateId)
             'x-appwrite-project' => $this->getProject()['$id'],
         ]));
 
+        // Fetch latest commit from GitHub API if template has provider info
+        if (
```
Meldiron: Lets make this a helper too

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
Meldiron: Same in multiple places of this PR diff

tests/e2e/Services/Functions/FunctionsCustomServerTest.php
Meldiron: Same in other function in this files diff

tests/e2e/Services/Sites/SitesCustomServerTest.php
```diff
@@ -1563,7 +1563,154 @@ public function testSiteTemplate(): void
...
+
+        $deployment = $this->getDeployment($siteId, $deployment['body']['$id']);
+        $this->assertEquals(200, $deployment['headers']['status-code']);
+        $this->assertEquals(0, $deployment['body']['sourceSize']);
+        $this->assertEquals(0, $deployment['body']['buildSize']);
+        $this->assertEquals(0, $deployment['body']['totalSize']);
+
+        $this->assertEventually(function () use ($siteId) {
+            $site = $this->getSite($siteId);
+            $this->assertNotEmpty($site['body']['deploymentId']);
```
Meldiron, line 1582: Lets assert it's our deployment ID, asserting just not empty might be flaky after some future changes

tests/e2e/Services/Sites/SitesCustomServerTest.php
```diff
@@ -1563,7 +1563,154 @@ public function testSiteTemplate(): void
...
+        ]);
+
+        $this->assertEquals(202, $deployment['headers']['status-code']);
+        $this->assertNotEmpty($deployment['body']['$id']);
+
+        $deployment = $this->getDeployment($siteId, $deployment['body']['$id']);
+        $this->assertEquals(200, $deployment['headers']['status-code']);
+        $this->assertEquals(0, $deployment['body']['sourceSize']);
+        $this->assertEquals(0, $deployment['body']['buildSize']);
+        $this->assertEquals(0, $deployment['body']['totalSize']);
```
Meldiron, line 1578: This is nice, but afterwards, lets ensure they are greater than 0, after deployment finishes

Aglowharmony (approved):
Merge this to main

app/controllers/general.php
abnegate: If we need a filter, it should be a minor version bump (1.9.0) because we're making backwards-compatible API changes

src/Appwrite/Platform/Modules/Functions/Http/Deployments/Template/Create.php
```diff
@@ -100,11 +103,15 @@ public function action(
             throw new Exception(Exception::FUNCTION_NOT_FOUND);
         }
 
+        $branchUrl = $type == GitHub::CLONE_TYPE_BRANCH ? "https://github.com/$owner/$repository/tree/$reference" : "";
```
Meldiron: Using `blob` path, we should be able to have branch URL for all types:

Commit: https://github.com/Meldiron/http-games/blob/dcc8c32076d768c08eed99b0f844eeae87034d4e/.env.example#L3

Branch: https://github.com/Meldiron/http-games/blob/main/.env.example#L3

Tag: https://github.com/Meldiron/http-games/blob/0.1.6/.env.example#L3

src/Appwrite/Platform/Modules/Sites/Http/Deployments/Template/Create.php
```diff
@@ -102,11 +105,15 @@ public function action(
             throw new Exception(Exception::SITE_NOT_FOUND);
         }
 
+        $branchUrl = $type == GitHub::CLONE_TYPE_BRANCH ? "https://github.com/$owner/$repository/tree/$reference" : "";
```
Meldiron: Same here

PR #10501, opened 2025-09-15
Branded email for Console auth flows

hmacr:
Task: https://linear.app/appwrite/issue/SER-375/update-email-verification-template-with-branded-design


app/config/console.php
```diff
@@ -49,6 +51,34 @@
...
+        'email.mfaChallenge-en' => [
+            'subject' => 'Verification Code for {{project}}',
+            'preview' => 'Use code {{otp}} for two-step verification in {{project}}. Expires in 15 minutes.',
+            'heading' => 'Complete two-step verification to use Appwrite Cloud',
+            'hello' => 'Hello {{user}},',
+            'body' => 'Enter the following code to confirm your two-step verification in {{b}}{{project}}{{/b}}. This code will expire in 15 minutes.',
+            'thanks' => 'Thanks,',
+            "signature" => "{{project}} team",
+        ]
+    ],
```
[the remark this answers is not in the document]
hmacr: Not adding this for now, because it leads to test failures - we don't have email OTP flow defined for Console at this point.

app/config/locale/templates/email-mfa-challenge.tpl
```diff
@@ -5,7 +5,7 @@
 <table border="0" cellspacing="0" cellpadding="0" style="padding-top: 10px; padding-bottom: 10px; display: inline-block;">
     <tr>
         <td align="center" style="border-radius: 8px; background-color: #ffffff;">
-            <p style="font-size: 24px; text-indent: 18px; letter-spacing: 18px; font-family: 'Inter', sans-serif; color: #414146; text-decoration: none; border-radius: 8px; padding: 24px 12px; border: 1px solid #EDEDF0; display: inline-block; font-weight: bold; ">{{otp}}</p>
+            <div style="font-size: 24px; text-indent: 18px; letter-spacing: 18px; font-family: 'Inter', sans-serif; color: #414146; text-decoration: none; border-radius: 8px; padding: 24px 12px; border: 1px solid #EDEDF0; display: inline-block; font-weight: bold; ">{{otp}}</div>
```
Meldiron: Let's include role (accessability, like screen readers)

Same for other places of PR where we went from <p> to <div>

app/controllers/api/account.php
```diff
@@ -2374,12 +2400,21 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
             // TODO: remove unnecessary team variable from this email
             'team' => '',
+            'heading' => $heading,
+            'accentColor' => APP_EMAIL_ACCENT_COLOR,
+            'logoUrl' => APP_EMAIL_LOGO_URL,
+            'twitterUrl' => APP_SOCIAL_TWITTER,
+            'discordUrl' => APP_SOCIAL_DISCORD,
+            'githubUrl' => APP_SOCIAL_GITHUB_APPWRITE,
+            'termsUrl' => APP_EMAIL_TERMS_URL,
+            'privacyUrl' => APP_EMAIL_PRIVACY_URL,
```
Meldiron: Since this is branded email specific, can we do it inder `$customEmails` if statement?

app/controllers/api/account.php
```diff
@@ -2360,6 +2369,23 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
+            $message
+                ->setParam('{{hello}}', $customTemplate['hello'])
+                ->setParam('{{description}}', $customTemplate['body'], escapeHtml: false)
+                ->setParam('{{thanks}}', $customTemplate['thanks'])
+                ->setParam('{{signature}}', $customTemplate['signature'])
+                ->setParam('{{clientInfo}}', '')
+                ->setParam('{{securityPhrase}}', '')
+                ->setParam('{{securityPhraseDividerDisplay}}', 'none');
+
+            $body = $message->render();
```
Meldiron: Do we need to do those? I would expect to just pass $body (customTemplate body), and worker would replace placeholders, like it already does with unbranded email

app/config/locale/templates/email-verification.tpl
Meldiron, line 1: This feels weird, why do we need to add this now? What is existing flow using; can we reuse that?

tests/e2e/Services/Account/AccountBase.php
```diff
@@ -183,6 +185,16 @@ public function testEmailOTPSession(): void
         $this->assertNotEmpty($code);
         $this->assertStringContainsStringIgnoringCase('Use OTP ' . $code . ' to sign in to '. $this->getProject()['name'] . '. Expires in 15 minutes.', $lastEmail['text']);
 
+        // Only Console project has branded logo in email.
+        if ($isConsoleProject) {
+            $this->assertStringContainsStringIgnoringCase('Appwrite logo', $lastEmail['html']);
+        }
```
Meldiron, line 193: let's also have "else" here, and ensure "Appwrite logo" is not there

app/controllers/api/account.php
hmacr: There are two passes happening here.

1. We have a branded `email-auth-styled` that serves as the base branded template for console auth flows. While it has blocks for "logo", "color scheme" among other things, email body varies per flow (account verification, otp, etc.)
2. For each flow, we have a separate template whose variables are replaced above.

app/config/locale/templates/email-verification.tpl
hmacr: By default, it uses `email-inner-base.tpl` which is used across many flows and has a default style for buttons (black/dark-grayish). For console flows, we need a branded button (pink) - I think it's far less error-prone to have this separated in a separate template rather than modify `email-inner-base` (could regress in a lot of ways).

app/config/locale/templates/email-mfa-challenge.tpl
hmacr: Instead, I removed the vertical margins and re-used `p` tag.

app/config/locale/templates/email-otp.tpl
```diff
@@ -15,6 +15,6 @@
 <p style="margin-bottom: 0px;">{{thanks}}</p>
 <p style="margin-top: 0px;">{{signature}}</p>
 
-<hr style="margin-block-start: 1rem; margin-block-end: 1rem;">
+<hr style="margin-block-start: 1rem; margin-block-end: 1rem; display: {{securityPhraseDividerDisplay}};">
```
Meldiron: I dont like logical properties for styling. Ill check with frontend folks if there is some CSS selector we could use instead. 

If possible, I would prefer to avoid need for this variable

app/config/locale/translations/en.json
```diff
@@ -5,6 +5,7 @@
     "emails.sender": "{{project}} Team",
     "emails.verification.subject": "Account Verification",
     "emails.verification.preview": "Verify your email to activate your {{project}} account.",
+    "emails.verification.heading": "Verify your email to start using Appwrite Cloud",
```
Meldiron: This translation file applies to all emails, not just ones to Appwrite Console.

If we need heading that doesnt follow same logic as project emails, we will need to introduce overrides for templates like we had before.

What if we used {{project}} here instead of "Appwite Cloud"? 🤔 and ensure project name is "Appwrite Cloud" in console.php config.

Meldiron: Same for all 3 headings in this file

app/config/locale/templates/email-otp.tpl
Meldiron: Darshan suggested `:empty` selector: https://developer.mozilla.org/en-US/docs/Web/CSS/:empty

Well supported:

CleanShot 2025-10-06 at 16 45 47@2x

Meldiron: Lets try to combine <hr> and <p> (see what HR does, I think its just some margin, padding, and border)

Once combined, we can move some of CSS under :empty selector, so some spacing and border only applies when there is a content

app/controllers/api/account.php
```diff
@@ -2293,7 +2296,11 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
 
         $subject = $locale->getText("emails.otpSession.subject");
         $preview = $locale->getText("emails.otpSession.preview");
+        $heading = $locale->getText("emails.otpSession.heading");
+
         $customTemplate = $project->getAttribute('templates', [])['email.otpSession-' . $locale->default] ?? [];
+        $smtpBaseTemplate = $project->getAttribute('smtpBaseTemplate', 'email-base');
+        $bodyTemplate = __DIR__ . '/../../config/locale/templates/' . $smtpBaseTemplate . '.tpl';
```
Meldiron, line 2310: I tend to agree with AI here, there is potential for abuse in future. Lets check if there is some PHP method that can help us protect it here, something that disallows changing directories, stuff like ../../../ and such. So we are certain we only got file name here

app/config/locale/translations/en.json
hmacr: Maybe "Appwrite cloud" may not be the right term here because the same emails will be sent for self-hosted. Perhaps just "Appwrite"?

app/controllers/api/account.php
```diff
@@ -4731,13 +4778,27 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
                     'otp' => $code,
                     'agentDevice' => $agentDevice['deviceBrand'] ?? $agentDevice['deviceBrand'] ?? 'UNKNOWN',
                     'agentClient' => $agentClient['clientName'] ?? 'UNKNOWN',
-                    'agentOs' => $agentOs['osName'] ?? 'UNKNOWN'
+                    'agentOs' => $agentOs['osName'] ?? 'UNKNOWN',
                 ];
 
+                if ($smtpBaseTemplate === 'email-base-styled') {
```
Meldiron: Lets global search all places where we trigger job for queueForEmails. Lets make sure we have those additions in all places (use of smtpBaseTemplate, and addition of those variables)

You can probably prompt AI agent too

app/config/locale/templates/email-otp.tpl
Meldiron: Example: https://www.w3schools.com/cssref/tryit.php?filename=trycss_sel_empty

app/controllers/api/account.php
```diff
@@ -2376,10 +2385,24 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
             'team' => '',
         ];
 
+        if ($smtpBaseTemplate === 'email-base-styled') {
+            $emailVariables = array_merge($emailVariables, [
+                'heading' => $heading,
```
Meldiron: Let's move heading to all variables, not just for console

app/controllers/api/account.php
```diff
@@ -2376,10 +2385,24 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
             'team' => '',
         ];
 
+        if ($smtpBaseTemplate === 'email-base-styled') {
```
Meldiron: let's make const for this variable value, I think we have a file with all consts.

Just a code quality, so we dont do typo in future by mistake

app/controllers/api/account.php
hmacr: umm why..? 🤔  We only need `smtpBaseTemplate` for branded console auth flows that are added here (OTP, MFA, Verification). Otherwise, Mails worker will fallback to unbranded mail.

app/controllers/api/account.php
```diff
@@ -2293,7 +2297,17 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
         $subject = $locale->getText("emails.otpSession.subject");
         $preview = $locale->getText("emails.otpSession.preview");
+        $heading = $locale->getText("emails.otpSession.heading");
+
         $customTemplate = $project->getAttribute('templates', [])['email.otpSession-' . $locale->default] ?? [];
+        $smtpBaseTemplate = $project->getAttribute('smtpBaseTemplate', 'email-base');
+
+        $validator = new FileName();
+        if (!$validator->isValid($smtpBaseTemplate)) {
+            throw new Exception(Exception::GENERAL_BAD_REQUEST, 'Invalid template path');
```
Meldiron, line 2307: Lets throw server error instead of bad_request, this should be 5xx error. Message still valuable, so we see it on Sentry

Meldiron (approved):

PR #10509, opened 2025-09-17
Add spatial column validation during required mode and tests for exis…

ArnabChatterjee20k:
…ting data in databases

## What does this PR do?

* Previously , when we have existing data and creating a required spatial col it was getting failed;

* Reasons
1. When we generate cols we generate with not required
2. Now when we are creating a spatial col with required -> we already have previous data -> so the column is getting filled with null
3. The spatial col is meant to be not null at the sql if required is checked
4. So insertion of null to not null giving error

* Fix
Adding a check to send error message to user as a validation step

* Why no changes in utopia database lib?
We can add changes to utopia lib but that runs in the workers. For validation steps, they are done in the api level before
creating the attribute. 
Also the database namespace will not be available internally in the utopia lib and can be added as a separate independent pr

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Attributes/Action.php
+++ b/src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Attributes/Action.php
@@ -122,7 +122,7 @@ abstract class Action extends UtopiaAction
     /**
      * Get the correct invalid structure message.
      */
-    protected function getInvalidStructureException(): string
+    protected function getStructureException(): string
     {
         return $this->isCollectionsAPI()
             ? Exception::DOCUMENT_INVALID_STRUCTURE
@@ -366,9 +366,16 @@ abstract class Action extends UtopiaAction
                 'filters' => $filters,
                 'options' => $options,
             ]);
-            if (!$dbForProject->getAdapter()->getSupportForSpatialIndexNull() && in_array($attribute->getAttribute('type'), Database::SPATIAL_TYPES) && $attribute->getAttribute('required')) {
-                $existingDataPresent = $dbForProject->findOne('database_' . $db->getSequence() . '_collection_' . $collection->getSequence());
-                if (count($existingDataPresent)) {
+            if (
+                !$dbForProject->getAdapter()->getSupportForSpatialIndexNull() &&
+                \in_array($attribute->getAttribute('type'), Database::SPATIAL_TYPES) &&
+                $attribute->getAttribute('required')
+            ) {
+                $hasData = !Authorization::skip(fn () => $dbForProject
+                    ->findOne('database_' . $db->getSequence() . '_collection_' . $collection->getSequence()))
+                    ->isEmpty();
+
+                if ($hasData) {
                     throw new StructureException('Failed to add required spatial column: existing rows present. Make the column optional.');
                 }
             }
@@ -379,7 +386,7 @@ abstract class Action extends UtopiaAction
         } catch (LimitException) {
             throw new Exception($this->getLimitException());
         } catch (StructureException $e) {
-            throw new Exception($this->getInvalidStructureException(), $e->getMessage());
+            throw new Exception($this->getStructureException(), $e->getMessage());
         } catch (Throwable $e) {
             $dbForProject->purgeCachedDocument('database_' . $db->getSequence(), $collectionId);
             $dbForProject->purgeCachedCollection('database_' . $db->getSequence() . '_collection_' . $collection->getSequence());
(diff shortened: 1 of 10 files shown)


src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Attributes/Action.php
```diff
@@ -366,13 +366,20 @@ protected function createAttribute(string $databaseId, string $collectionId, Doc
                 'filters' => $filters,
                 'options' => $options,
             ]);
-
+            if (!$dbForProject->getAdapter()->getSupportForSpatialIndexNull() && in_array($attribute->getAttribute('type'), Database::SPATIAL_TYPES) && $attribute->getAttribute('required')) {
+                $existingDataPresent = $dbForProject->findOne('database_' . $db->getSequence() . '_collection_' . $collection->getSequence());
+                if (count($existingDataPresent)) {
+                    throw new StructureException('Failed to add required spatial column: existing rows present. Make the column optional.');
+                }
+            }
```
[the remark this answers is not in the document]
fogelito: Do we need to  Authorization::skip ?

tests/e2e/Services/Databases/TablesDB/DatabasesBase.php
```diff
@@ -8828,4 +8828,134 @@ public function testSpatialDistanceInMeter(): void
...
+        $this->assertEquals(400, $poly['headers']['status-code']);
+
+        $poly = $this->client->call(Client::METHOD_POST, '/tablesdb/' . $databaseId . '/tables/' . $tableId . '/columns/polygon', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'x-appwrite-key' => $this->getProject()['apiKey']
+        ]), [
+            'key' => 'area',
+            'required' => false,
+            'default' => null
```
fogelito, line 8955: Do we allow other values than nulls ?

tests/e2e/Services/Databases/TablesDB/DatabasesBase.php
```diff
@@ -8828,4 +8828,134 @@ public function testSpatialDistanceInMeter(): void
...
+        $this->assertEquals(400, $point['headers']['status-code']);
+
+        $point = $this->client->call(Client::METHOD_POST, '/tablesdb/' . $databaseId . '/tables/' . $tableId . '/columns/point', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'x-appwrite-key' => $this->getProject()['apiKey']
+        ]), [
+            'key' => 'loc',
+            'required' => false,
+            'default' => null
```
fogelito, line 8909: Do we allow other values than nulls ?

ArnabChatterjee20k: yes, shall I change the null to other values for the tests?

tests/e2e/Services/Databases/TablesDB/DatabasesBase.php
ArnabChatterjee20k: yes

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Attributes/Action.php
```diff
@@ -366,13 +366,20 @@ protected function createAttribute(string $databaseId, string $collectionId, Doc
                 'filters' => $filters,
                 'options' => $options,
             ]);
-
+            if (!$dbForProject->getAdapter()->getSupportForSpatialIndexNull() && in_array($attribute->getAttribute('type'), Database::SPATIAL_TYPES) && $attribute->getAttribute('required')) {
```
fogelito: @abnegate Let's make an adapter call here, which will check all scenarios before creating an attribute.
And test it in Utopia?

tests/e2e/Services/Databases/TablesDB/DatabasesBase.php
```diff
@@ -8828,4 +8828,234 @@ public function testSpatialDistanceInMeter(): void
...
+
+        // Test point with default value
+        $point = $this->client->call(Client::METHOD_POST, '/tablesdb/' . $databaseId . '/tables/' . $tableId . '/columns/point', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'x-appwrite-key' => $this->getProject()['apiKey']
+        ]), [
+            'key' => 'loc',
+            'required' => false,
+            'default' => [0.0, 0.0]
```
fogelito, line 9029: Try to create a Document and see if it falls back to this value 👍

ArnabChatterjee20k: added

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Attributes/Action.php
abnegate: Makes sense, we can do it outside this scope. We took a step towards with `validateAttribute` in the library, just need to expose it, still private at the moment

abnegate (approved):

PR #10533, opened 2025-09-22
Throw error when email is not available for account verification

hmacr:
Task: https://linear.app/appwrite/issue/SER-378/code-quality-setrecipient-argument-1-must-be-of-type-string-null-given

### DIFF
--- a/tests/e2e/Services/Account/AccountCustomClientTest.php
+++ b/tests/e2e/Services/Account/AccountCustomClientTest.php
@@ -1864,7 +1864,7 @@ class AccountCustomClientTest extends Scope
             'url' => 'http://localhost/verification',
         ]);
 
-        $this->assertEquals(500, $response['body']['code']);
+        $this->assertEquals(400, $response['body']['code']);
         $this->assertEquals('user_email_not_found', $response['body']['type']);
 
         return [];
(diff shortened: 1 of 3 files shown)


vermakhushboo (approved):

tests/e2e/Services/Account/AccountCustomClientTest.php
```diff
@@ -1850,6 +1850,26 @@ public function testCreateAnonymousAccount()
...
+        $response = $this->client->call(Client::METHOD_POST, '/account/verification', array_merge([
+            'origin' => 'http://localhost',
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $this->getProject()['$id'],
+            'cookie' => 'a_session_' . $this->getProject()['$id'] . '=' . $session,
+        ]), [
+            'url' => 'http://localhost/verification',
+        ]);
+
+        $this->assertEquals(500, $response['body']['code']);
```
Meldiron: I would expect 4xx error here. I think we are missing definition for this exception here: https://github.com/appwrite/appwrite/blob/60433ef22d85520ea67cc254b109e9436490af3c/app/config/errors.php#L280

While at it, see if you can prompt AI agent to write a unit test that tries every exception type, and ensures it has definition in errors.php config. You can then remove all the tests it implemented, but from my experience with AI agents, if you tell it to do test, it wont stop until it's 100% accurate. You might need to guide it with a command how to run tests too.

Issue #10421, opened 2025-09-02
🚀 Enhancement: JSON Attribute Type


ZachHandley:
### 🔖 Enhancement description

Not a lot to this enhancement description

I use JSON attributes all the time, seeing as how there's a 62 attribute limit, and then no way to properly store an object, it would be nice to have this instead of a massive string attribute, so we can store objects in the DB without needing to stringify/parse them all the time

### 🎤 Pitch

- JSON Attribute
- Could technically Query the sub-attributes in the JSON, if wanted, as if it was a key-value object
- It could even just be a String attribute wrapper, but we could define object types (typdefs only) / it would parse / stringify it on the way out/in

### 👀 Have you spent some time to check if this issue has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

eldadfux:
This is something we really want to see, I know we had some issues in the past with the state of this on both MariaDB and MySQL, will need to check with the team and see where do we stand on that.

PR #10534, opened 2025-09-22
issue number - 10421        #10421 

Aglowharmony:
[ago](https://github.com/appwrite/appwrite/issues/10421#issue-3376436559)
🔖 Enhancement description
Not a lot to this enhancement description

I use JSON attributes all the time, seeing as how there's a 62 attribute limit, and then no way to properly store an object, it would be nice to have this instead of a massive string attribute, so we can store objects in the DB without needing to stringify/parse them all the time

🎤 Pitch

- JSON Attribute
- Could technically Query the sub-attributes in the JSON, if wanted, as if it was a key-value object
- It could even just be a String attribute wrapper, but we could define object types (typdefs only) / it would parse / stringify it on the way out/in

------------------------------------------------------------------------------------------------------------------------------
                        ADDED THIS FEATURE 
------------------------------------------------------------------------------------------------------------------------------


Aglowharmony:
@atharvadeosthale help me @sesam @brandonroberts @tomer @tomer

Aglowharmony (commented):
Please merge this and approve it

PR #10546, opened 2025-09-24
CSV export

abnegate:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Workers/Migrations.php
+++ b/src/Appwrite/Platform/Workers/Migrations.php
@@ -453,12 +453,11 @@ class Migrations extends Action
             try {
                 $this->deviceForFiles->delete($path);
             } finally {
+                $message = "Export file size {$sizeMB}MB exceeds your plan limit.";
+
                 $this->dbForProject->updateDocument('migrations', $migration->getId(), $migration->setAttribute(
                     'errors',
-                    [
-                        'code' => 0,
-                        'message' => "Export file size {$sizeMB}MB exceeds your plan limit.",
-                    ],
+                    json_encode(['code' => 0, 'message' => $message]),
                     Document::SET_TYPE_APPEND,
                 ));
 


src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -245,7 +305,9 @@ protected function generateAPIKey(Document $project): string
                 'functions.write',
                 'databases.read',
                 'collections.read',
+                'collections.write',
```
ItzNotABug: for what do we require write scope here? we only read the data and save to storage, right?

src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -303,40 +365,44 @@ protected function processMigration(Document $migration, Realtime $queueForRealt
...
+                $errors = [];
+                foreach ([...$sourceErrors, ...$destinationErrors] as $error) {
+                    $encoded = \json_decode(\json_encode($error), true);
+                    if (\is_array($encoded)) {
+                        if (isset($encoded['trace'])) {
+                            unset($encoded['trace']);
+                        }
+                        $errors[] = \json_encode($encoded);
+                    } else {
+                        $errors[] = \json_encode($error);
```
ItzNotABug: maybe we can make a helper method, its already used twice I think.

src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -363,57 +429,175 @@ function () use ($migration, $transfer, $projectDocument, $queueForRealtime) {
...
+                    }
```
ItzNotABug: same as above. 3 now ig.

src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -363,57 +429,175 @@ function () use ($migration, $transfer, $projectDocument, $queueForRealtime) {
...
                 $source?->success();
+
+                if ($migration->getAttribute('destination') === DestinationCSV::getName()) {
+                    $this->handleCSVExportComplete($project, $migration, $queueForMails);
+                }
             }
         }
     }
+
+    protected function handleCSVExportComplete(Document $project, Document $migration, Mail $queueForMails): void
```
ItzNotABug: where or how do we handle the org's plan limits for the export sizes in the bucket? Even the free limit has a file limit of `50MB` and its still a good number but just curious.

ItzNotABug (changes requested):

src/Appwrite/Platform/Workers/Migrations.php
abnegate: Was added in the case of using the API writer in migrations, but we don't use it, will remove all DB scopes actually

src/Appwrite/Platform/Workers/Migrations.php
abnegate: Good catch, added

src/Appwrite/Utopia/Response/Model/Migration.php
```diff
@@ -86,6 +86,12 @@ public function __construct()
                 'default' => [],
                 'example' => [],
             ])
+            ->addRule('options', [
+                'type' => self::TYPE_JSON,
+                'description' => 'Migration options used during the migration process.',
+                'default' => [],
+                'example' => '{"bucketId": "exports", "notify": false}',
+            ])
```
ItzNotABug, line 94: why do we need this on response? we don't keep this for realtime though 🤔

src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -362,58 +382,233 @@ function () use ($migration, $transfer, $projectDocument, $queueForRealtime) {
...
+            try {
+                $this->deviceForFiles->delete($path);
+            } finally {
+                $message = "Export file size {$sizeMB}MB exceeds your plan limit.";
+                $this->dbForProject->updateDocument('migrations', $migration->getId(), $migration->setAttribute(
+                    'errors',
+                    $message,
+                    Document::SET_TYPE_APPEND,
+                ));
+                throw new \Exception($message);
```
ItzNotABug, line 473: this is on background, do we send an email for this failure? not necessary the user will be on console the whole time. Do we plan to use realtime but that would only work while the user is on the console, not sure how'd we distinguish between errors seen and dismissed when the user has gone and come back later.

src/Appwrite/Utopia/Response/Model/Migration.php
abnegate: Removing from realtime is a bug that I was just checking on console, thanks :grin:

src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -112,9 +133,19 @@ public function action(Message $message, Document $project, Database $dbForProje
...
         $migrationOptions = $migration->getAttribute('options');
+        $dataSource = Appwrite::SOURCE_API;
+        $database = null;
+        $queries = [];
+
+        if ($source === Appwrite::getName() && $destination === DestinationCSV::getName()) {
+            $dataSource = Appwrite::SOURCE_DATABASE;
+            $database = $this->dbForProject;
+            $queries = Query::parseQueries($migrationOptions['queries']);
+        }
```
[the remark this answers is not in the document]
abnegate: Queries can only get here if they were successfully parsed during the API request

src/Appwrite/Platform/Workers/Migrations.php
abnegate: Makes sense, added

PR #10556, opened 2025-09-26
fix(auth): validate old password before allowing password change

ArjunSaluja:
### What does this PR do?

Ensures that when a user tries to change their password, the system validates the old password first. This prevents unauthorized password changes and improves security.

### Changes Made
- Updated `Auth.php` to include validation of the old password before updating to a new password.
- Added error handling for incorrect old passwords.

### Why is this important?
Without this check, users could change their password without verifying their current password, potentially compromising account security.

### How to test
1. Attempt to change a password with the correct old password → should succeed.
2. Attempt to change a password with an incorrect old password → should fail with an error.

### DIFF
--- a/src/Appwrite/Auth/Auth.php
+++ b/src/Appwrite/Auth/Auth.php
@@ -2,514 +2,132 @@
 
 namespace Appwrite\Auth;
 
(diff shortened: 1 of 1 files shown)


HarshMN2345:

PR #10558, opened 2025-09-28
feat: add user verification enforcement for Console project

HarshMN2345:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/config/specs/open-api3-1.8.x-console.json
+++ b/app/config/specs/open-api3-1.8.x-console.json
@@ -25153,7 +25153,7 @@
                                 "properties": {
                                     "type": {
                                         "type": "string",
-                                        "description": "Platform type.",
+                                        "description": "Platform type. Possible values are: web, flutter-web, flutter-ios, flutter-android, flutter-linux, flutter-macos, flutter-windows, apple-ios, apple-macos, apple-watchos, apple-tvos, android, unity, react-native-ios, react-native-android.",
                                         "x-example": "web",
                                         "enum": [
                                             "web",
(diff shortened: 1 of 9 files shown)


app/controllers/shared/api.php
```diff
@@ -325,6 +325,44 @@
...
+                }
+            }
+            
+            if (!$isAllowedEndpoint) {
+                $emailVerified = $user->getAttribute('emailVerification', false);
+                $phoneVerified = $user->getAttribute('phoneVerification', false);
+                
+                if (!$emailVerified && !$phoneVerified) {
+                    throw new Exception(Exception::USER_VERIFICATION_REQUIRED);
+                }
```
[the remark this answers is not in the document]
stnguyen90: @HarshMN2345 please see this.

tests/e2e/Services/Projects/ProjectsVerificationE2ETest.php
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public function testVerificationMiddlewareWithEmptyUser()
+    {
+        // Test that empty/unauthenticated users are handled properly
+        $this->client->setProject('console');
+        
+        $response = $this->client->call('GET', '/users', [], [
+            'content-type' => 'application/json'
+        ]);
```
stnguyen90, line 103: This isn't a API call that is typically done on the console project 🧐

tests/e2e/Services/Projects/ProjectsVerificationE2ETest.php
```diff
@@ -0,0 +1,108 @@
...
+
+    public function testVerificationMiddlewareWithApiKey()
+    {
+        // Test that API key authentication bypasses verification
+        $this->client->setProject('console');
+        
+        $response = $this->client->call('GET', '/users', [], [
+            'content-type' => 'application/json',
+            'x-appwrite-key' => $this->getProject()['apiKey']
+        ]);
```
stnguyen90, line 90: this makes the /users call against the console project? I would expect this to always return 401 🧐

app/controllers/shared/api.php
```diff
@@ -325,6 +325,44 @@
...
+                '/v1/console/variables',          // Console configuration
+                '/v1/health/version',             // Health checks
+                '/v1/account/verification',       // Email verification
+                '/v1/account/verification/phone', // Phone verification
+                '/v1/account/recovery',           // Account recovery
+                '/v1/account/sessions',           // Session management
+                '/v1/account/tokens',             // Token management
+                '/v1/account/mfa'                 // Multi-factor authentication
+            ];
+
```
Meldiron, line 347: Instead of having list of endpoints here, we can do one of 2 things:

A) Have a special `group()` for endpoints that are supposed to be blocked (might be bad idea if its too many)
B) Have a special `label()` for marking endpoint as allowed without verification

Also, this would enforce email verification on self-hosted too. It might feel like extra blocker, requiring to setup SMTP, causing people to leave the onboarding.

For that reason, I think this user verification exception for Console should be cloud-specific, in Cloud repository.

HarshMN2345: I think anything under /account should be accessible even if not verified

Meldiron:
Closing in favour of another PR

PR #10573, opened 2025-09-30
[draft] init payments

atharvadeosthale:


app/config/collections/auth_plans.php
eldadfux, line 1: This shouldn't be called auth_plans. It should live in one of the existing files. In tis case the projects.php file because each project will need to have these collections.

app/config/collections/auth_plans.php
```diff
@@ -0,0 +1,676 @@
...
+                'format' => '',
+                'size' => Database::LENGTH_KEY,
+                'signed' => true,
+                'required' => true,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('planId'),
```
eldadfux: No need for a planId, we should always rely on the built-in _uid (aka "$id").

app/config/collections/auth_plans.php
```diff
@@ -0,0 +1,676 @@
...
+                'format' => '',
+                'size' => 256,
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('stripePriceId'),
```
eldadfux: We should never tie Appwrite to a specific vendor. This is how vendor locking gets created. Instead it should be providerType, providerId or something like that. You should check if we have a similar pattern in another collection and follow it for consistency. Consistency is a huge part in creating great DX. People want to leverage knowledge they gained from using something like Appwrite Messaging and feel like Appwrite Payments works the same way.

app/config/collections/auth_plans.php
```diff
@@ -0,0 +1,676 @@
...
+                'format' => '',
+                'size' => 256,
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('stripeMeterId'),
```
eldadfux: Same for all mentions of Stripe in the source code.

app/config/collections/auth_plans.php
```diff
@@ -0,0 +1,676 @@
...
+                'format' => '',
+                'size' => 3,
+                'signed' => true,
+                'required' => true,
+                'default' => 'usd',
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('interval'),
```
eldadfux: What unit is being used here? Let's add a comment

app/config/collections/auth_plans.php
```diff
@@ -0,0 +1,676 @@
...
+                'format' => '',
+                'size' => 10,
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('features'),
```
eldadfux: Why use a JSON over normalized structure? Not saying its not right, but what is the reason?

app/config/collections/auth_plans.php
```diff
@@ -0,0 +1,676 @@
...
+                'signed' => true,
+                'size' => 0,
+                'format' => '',
+                'filters' => [],
+                'required' => false,
+                'default' => false,
+                'array' => false,
+            ],
+            [
+                '$id' => ID::custom('maxUsers'),
```
eldadfux: How is this useful? what's the use case? What happens if I set this to 1m? On large scale, these kind of count can overwhelm the DB and create a loophole.

app/config/collections/auth_plans.php
```diff
@@ -0,0 +1,676 @@
...
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ]
+        ],
+        'indexes' => [
+            [
+                '$id' => ID::custom('_key_project'),
```
eldadfux: Before going production, get the DB team to validate all indexes and suggest changes.

app/config/collections/auth_plans.php
```diff
@@ -0,0 +1,676 @@
...
+            [
+                '$id' => ID::custom('_fulltext_search'),
+                'type' => Database::INDEX_FULLTEXT,
+                'attributes' => ['search'],
+                'lengths' => [],
+                'orders' => [],
+            ],
+        ],
+    ],
+    'auth_features' => [
```
eldadfux: I don't think we use snake case for namespaces, anyway the namespace shouldn't be `auth_`

eldadfux: What is `auth_features` vs `auth_plan_features`?

app/config/collections/auth_plans.php
```diff
@@ -0,0 +1,676 @@
...
+                'format' => '',
+                'size' => Database::LENGTH_KEY,
+                'signed' => true,
+                'required' => true,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('featureId'),
```
eldadfux: We don't need this ID, use $id/_uid

app/config/collections/auth_plans.php
```diff
@@ -0,0 +1,676 @@
...
+                'format' => '',
+                'size' => 0,
+                'signed' => false,
+                'required' => false,
+                'default' => 0,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('tiersMode'),
```
eldadfux: what is this?

app/config/collections/auth_plans.php
```diff
@@ -0,0 +1,676 @@
...
+                'format' => '',
+                'size' => 16,
+                'signed' => true,
+                'required' => false,
+                'default' => 'graduated',
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('tiers'),
```
eldadfux: What is this? how do we have tiers inside a plan?

app/config/collections/auth_plans.php
```diff
@@ -0,0 +1,676 @@
...
+                'format' => '',
+                'size' => 16384,
+                'signed' => true,
+                'required' => false,
+                'default' => [],
+                'array' => false,
+                'filters' => ['json'],
+            ],
+            [
+                '$id' => ID::custom('usageCap'),
```
eldadfux: does this include the base plan price or not?

app/config/collections/auth_plans.php
```diff
@@ -0,0 +1,676 @@
...
+            [
+                '$id' => ID::custom('_key_active'),
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['active'],
+                'lengths' => [0],
+                'orders' => [Database::ORDER_ASC],
+            ],
+        ],
+    ],
+    'auth_usage_events' => [
```
eldadfux: Is this for measuring usage? We might better use our own usage or rely on the provider's solution all together.

app/config/collections/platform.php
```diff
@@ -330,6 +330,72 @@
                 'default' => null,
                 'array' => false,
                 'filters' => ['datetime'],
+            ],
+            [
+                '$id' => ID::custom('authSubscriptionsEnabled'),
```
eldadfux: Same issues with vendor locking and stripe, Imagine tomorrow a new product called UltraPayments comes in to the picture and is better than Stripe in every aspect, we need to be able to switch or support both,

eldadfux: It might be easier here to follow what we do for auth and just have a single json attribute for `payments` containing all the configs we need.

app/config/collections/common.php
```diff
@@ -364,6 +364,94 @@
                 'array' => false,
                 'filters' => ['datetime'],
             ],
+            [
+                '$id' => ID::custom('planId'),
```
eldadfux: Stripe vendor locking here again. Also, we don't want to couple plans with specific users, we need to fine a solution to be able to use them for both users or teams. I would avoid changing teams and users collections and rely one the new collections to create the link, this link is probably a `subscriptions` collection.

app/config/services.php
```diff
@@ -261,6 +261,19 @@
         'optional' => false,
         'icon' => '',
     ],
+    'authPlans' => [
```
eldadfux: This is not a service in Appwrite. As mentioned on the PR comment, we should create a new `v1/payments` service to centralize all these APIs.

app/controllers/api/account.php
```diff
@@ -5180,3 +5182,496 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
 
         return $response->noContent();
     });
+
+App::get('/v1/account/subscription')
```
eldadfux: This is not a REST compatible API. In rest entities should be plurals.

eldadfux (changes requested):
This should rely on auth, but you should avoid trying to extend /v1/account as much ass possible. It’s better for us to treat this as a totally new service. `v1/payments`

For v1/payments - we should imitate how new services in Appwrite are built. Look at Functions or Sites for inspiration we use a class based format to define routes unlike what we do in old controllers. We also have a test classes per route and not per service. You can use something like cursor with proper example to the agent to easily convert between the two. 

We should find a way to attach a plan or payment to either a team or a user. Teams are crucial for multi-tenant saas products.

app/config/collections/auth_plans.php
atharvadeosthale: Multiple intervals are supported (daily, weekly, monthly, yearly)

app/config/collections/auth_plans.php
atharvadeosthale: Hmm, I don't think this is needed anymore because features are a different thing now, in a previous implementation, features were very flat and assigned per user, I'll work on remove this.

app/config/collections/auth_plans.php
atharvadeosthale: I could think of a time when you have an early access plan with less cost with same features as your original plan

app/config/collections/auth_plans.php
atharvadeosthale: volume vs gradual - with gradual, you have tiers of metered pricing

app/config/collections/auth_plans.php
atharvadeosthale: yes, you can, for example

0 - 10 units: $0
10 - 20 units: $1
20 - Inf units: $2

app/config/collections/auth_plans.php
atharvadeosthale: It's crucial to use both - we enforce usage caps.

Stripe's ingestion is asynchronous and does not reflect immediately, so, someone can manage to spam their way out of usage caps.

If we use our own meters, we will need to handle billing the user at the end of cycle, not a headache worth having.

It's in DB for instant updates to apps, and also sent to Stripe for billing to occur.

atharvadeosthale:
Massive list of TODOs for me

- [x] test out api scopes for `payments-*`
- [x] make checkouts work
- [x] add route to get subscription for provided team or user id based on access
- [x] handle default localhost webhook support (throw saying domain must be configed)
- [x] make sure stripe webhook handling and provisioning works properly
- [x] in webhook, get subscription id and store in db too, so when cancelling, it makes sense
- [x] handle customers gracefully (ensure customers exist before subscription)
- [ ] ~~explore separating prices into a separate table, don't really wanna store it as json~~
- [x] explore moving plans and prices to project db instead of platform db
- [x] create a status route of some sort to get an actors subscription status
- [x] handle usage based 
- [x] usage worker (?)
- [x] only send relevant stuff in response, dont send document copy as-is
- [x] stripe checkout MUST have customer details
- [x] people / teams already subscribed must not be allowed to subscribe with an active plan
- [x] change back domain to `traefik` after testing is done
- [x] remove legacy hackathon routes and fix merge conflicts
- [x] handle https in webhook creation
- [x] delete payments.php collection file when safe to
- [x] throw if actor doesn't exist when getting subscription
- [ ] make sure fetching actor features work
- [x] give response models a review
- [x] make sure when a provider is configured, already available plans are also added
- [x] ensure plans are added to all providers
- [ ] remove plans when de-configuring a provider

*adding more as I realise*

atharvadeosthale:
@coderabbit review

Issue #10583, opened 2025-10-01
Change queries parameter to array of strings in site list endpoint


Priyanshuthapliyal2005:
### 💭 Description

The site list endpoint expects queries as a string in labels, but the docs specify an array of strings. Backend confirms array of strings is correct. 
<img width="1540" height="794" alt="Image" src="https://github.com/user-attachments/assets/9bd53bfb-728a-4006-907e-2611fd5d332c" />

<img width="1612" height="778" alt="Image" src="https://github.com/user-attachments/assets/03357ed6-fd57-4091-8866-593f5cc08f21" />

<img width="1830" height="810" alt="Image" src="https://github.com/user-attachments/assets/7a01b282-9686-4584-9761-c8cf975d3d10" />

### 👀 Have you spent some time to check if this issue has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

stnguyen90:
@Priyanshuthapliyal2005 thanks for raising this issue! 🙏🏼 Would you please clarify what the issue is? I didn't quite understand:

> The site list endpoint expects queries as a string in labels

Priyanshuthapliyal2005:
> [@Priyanshuthapliyal2005](https://github.com/Priyanshuthapliyal2005) thanks for raising this issue! 🙏🏼 Would you please clarify what the issue is? I didn't quite understand:
> 
> > The site list endpoint expects queries as a string in labels

<img width="1215" height="661" alt="Image" src="https://github.com/user-attachments/assets/c61eb1dd-87bd-4e5c-a552-9c0f74b08890" />

<img width="1258" height="747" alt="Image" src="https://github.com/user-attachments/assets/7df90496-d97b-4744-9970-34f59a307460" />

Sorry, I forgot to include a screenshot for clarity. The inconsistency is that in the site list endpoint, the 'queries' parameter label is given as a string, but it actually uses an array of query strings (according to the backend and docs). Similarly, in list deployments, the requirement is the same and it mentions label as an array. For list sites, it mentions string label. All three endpoints expect an array of strings, so I updated the label to array of strings in the PR for clarity and consistency

priyanshub18:
i would like to pick up this issue @stnguyen90 ?

eldadfux:
[...]

PR #10584, opened 2025-10-01
fix(sites): make queries parameter an array of strings for site list endpoint

Priyanshuthapliyal2005:
## What does this PR do?
Updates Sites list endpoint queries parameter to an array of strings from string or array.

## Test Plan
Checked changes using spec files.

## Related PRs and Issues

- (Related PR or issue)
#10583 

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Logs/XList.php
+++ b/src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Logs/XList.php
@@ -108,10 +108,7 @@ class XList extends Action
         $audit = new Audit($dbForProject);
         $type = $this->getCollectionsEventsContext();
         $context = $this->getContext();
-        $resource = match ($context) {
-            ROWS   => "database/$databaseId/grid/$type/$collectionId/$context/{$document->getId()}",
-            default => "database/$databaseId/$type/$collectionId/$context/{$document->getId()}",
-        };
+        $resource = "database/$databaseId/$type/$collectionId/$context/{$document->getId()}";
 
         $logs = $audit->getLogsByResource($resource, $queries);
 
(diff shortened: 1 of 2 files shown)


src/Appwrite/Platform/Modules/Sites/Http/Deployments/XList.php
```diff
@@ -55,7 +56,7 @@ public function __construct()
                 ]
             ))
             ->param('siteId', '', new UID(), 'Site ID.')
-            ->param('queries', [], new Deployments(), 'Array of query strings generated using the Query class provided by the SDK. [Learn more about queries](https://appwrite.io/docs/queries). Maximum of ' . APP_LIMIT_ARRAY_PARAMS_SIZE . ' queries are allowed, each ' . APP_LIMIT_ARRAY_ELEMENT_SIZE . ' characters long. You may filter on the following attributes: ' . implode(', ', Deployments::ALLOWED_ATTRIBUTES), true)
+            ->param('queries', [], new ArrayList(new Text(APP_LIMIT_ARRAY_ELEMENT_SIZE), APP_LIMIT_ARRAY_PARAMS_SIZE), 'Array of query strings generated using the Query class provided by the SDK. [Learn more about queries](https://appwrite.io/docs/queries). Maximum of ' . APP_LIMIT_ARRAY_PARAMS_SIZE . ' queries are allowed, each ' . APP_LIMIT_ARRAY_ELEMENT_SIZE . ' characters long. You may filter on the following attributes: ' . implode(', ', Deployments::ALLOWED_ATTRIBUTES), true)
```
Meldiron, line 59: Thanks for the PR 🙌

Sadly, I dont think I can approve this; using validator here ensures only query syntax is used, and even further, ensures only allowed attributes are used for certain queries.

What was intention of the PR? What's the issue at hand? I can give it some thought and suggest a different solution

Meldiron: Ah, found it in the issue itself:

CleanShot 2025-10-07 at 11 16 42@2x

I would recommend to global-search for queries validator that work properly, and see where they are mentioned. and find spot where Log is not mentioned, and add it. My initial thought would be some spec files (swagger, openapi), or graphql.

Priyanshuthapliyal2005: Thanks @Meldiron for taking a look! I’ll revisit the approach based on your feedback and update the PR once I find a cleaner solution.

PR #10585, opened 2025-10-01
fix: logs truncation tests

ChiragAgg5k:


tests/resources/functions/log-error-truncation/index.js
```diff
@@ -2,8 +2,21 @@ module.exports = async(context) => {
...
+        chunks.push(longString.slice(i, i + chunkSize));
+    }
+    
+    chunks.forEach((chunk, index) => {
+        context.log(chunk);
+    });
+    
+    chunks.forEach((chunk, index) => {
+        context.error(chunk);
+    });
```
ItzNotABug, line 19: why separate separate? we could use just one forEach no?

ChiragAgg5k: we need .logs first, then .errors

abnegate (approved):

PR #10606, opened 2025-10-06
Rename verification SDK methods to be more specific

ChiragAgg5k:
## Summary
- Renames `createVerification` to `createEmailVerification`
- Renames `updateVerification` to `updateEmailVerification`
- Maintains backwards compatibility by keeping old method names as deprecated (since 1.8.0)
- Updates SDK method definitions to use array syntax with both new and deprecated methods

## Changes
The verification SDK methods have been updated to better reflect their purpose as email verification methods:

### Create Email Verification
- **New method**: `account.createEmailVerification`
- **Deprecated method**: `account.createVerification` (deprecated since 1.8.0)

### Update Email Verification
- **New method**: `account.updateEmailVerification`
- **Deprecated method**: `account.updateVerification` (deprecated since 1.8.0)

## Implementation Details
- Both endpoints now use an array of Method objects in their SDK label
- The first method in each array is the new, preferred method name
- The second method is the deprecated old name with a `Deprecated` object indicating replacement
- All OpenAPI/Swagger specs have been automatically updated to reflect these changes

## Test plan
- [ ] Verify SDK generation includes both new and deprecated methods
- [ ] Confirm deprecated methods show proper deprecation warnings
- [ ] Test that both old and new method names work correctly
- [ ] Validate OpenAPI/Swagger spec changes are correct

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -3513,20 +3513,40 @@ App::post('/v1/account/verification')
     ->label('event', 'users.[userId].verification.[tokenId].create')
     ->label('audits.event', 'verification.create')
     ->label('audits.resource', 'user/{response.userId}')
-    ->label('sdk', new Method(
-        namespace: 'account',
-        group: 'verification',
-        name: 'createVerification',
-        description: '/docs/references/account/create-email-verification.md',
-        auth: [AuthType::SESSION, AuthType::JWT],
-        responses: [
-            new SDKResponse(
-                code: Response::STATUS_CODE_CREATED,
-                model: Response::MODEL_TOKEN,
-            )
-        ],
-        contentType: ContentType::JSON,
-    ))
+    ->label('sdk', [
+        new Method(
+            namespace: 'account',
+            group: 'verification',
+            name: 'createEmailVerification',
+            description: '/docs/references/account/create-email-verification.md',
+            auth: [AuthType::SESSION, AuthType::JWT],
+            responses: [
+                new SDKResponse(
+                    code: Response::STATUS_CODE_CREATED,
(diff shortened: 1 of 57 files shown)


app/controllers/api/account.php
```diff
@@ -3513,20 +3513,40 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
+                    code: Response::STATUS_CODE_CREATED,
+                    model: Response::MODEL_TOKEN,
+                )
+            ],
+            contentType: ContentType::JSON,
+        ),
+        new Method(
+            namespace: 'account',
+            group: 'verification',
+            name: 'createVerification',
```
Meldiron, line 3535: I think we can remove current method. new SDK can have increased major verson, to indicate a breaking change

app/controllers/api/account.php
```diff
@@ -3513,20 +3513,40 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
     ->label('event', 'users.[userId].verification.[tokenId].create')
```
Meldiron, line 3514: Above here, in endpoint path, we will need `/email` too. and alias for old path

app/controllers/api/account.php
ChiragAgg5k: we always deprecate first even if breaking, then remove it in next major version

ChiragAgg5k: Screenshot 2025-10-06 at 7 12 08 PM
we have methods like listMfaFactors we should remove all at once

app/controllers/api/account.php
ChiragAgg5k: updated

PR #10615, opened 2025-10-08
add automatic releases

ChiragAgg5k:


src/Appwrite/Platform/Tasks/SDKs.php
```diff
@@ -245,6 +253,97 @@ public function action(?string $selectedPlatform, ?string $selectedSDK, ?string
...
+
+                        $tempNotesFile = \tempnam(\sys_get_temp_dir(), 'release_notes_');
+                        \file_put_contents($tempNotesFile, $formattedNotes);
+
+                        $releaseCommand = 'gh release create "' . $releaseVersion . '" \
+                            --repo "' . $repoName . '" \
+                            --title "' . $releaseTitle . '" \
+                            --notes-file "' . $tempNotesFile . '" \
+                            --target "' . $releaseTarget . '" \
+                            2>&1';
```
abnegate, line 317: No way to pass notes as string?

ChiragAgg5k: i like it better that you can tell claude/cursor to update the changelog files, and then the script handles the rest

passing as string is just extra work for us

abnegate (approved):

PR #10619, opened 2025-10-09
Fix: MFA with email-verified factor

Meldiron:
## What does this PR do?

1. Prevents from suggesting email/phone as 2nd factor, if already used as 1st factor (email otp, magic url, phone otp)
2. Prevents MFA from triggering if only 1 factor is enabled, and it's factor used during sign-in

-> to consider, should password be 2nd factor?
-> to consider reverting to original api.php behaviour, and throwing error in email/phone flows, when email/phone is only added facotor, and mfa is enabled

## Test Plan

Introduce tests:
- After enabling 2fa (in different scenarios), ensure current session remains active
- Ensure listFactors to behave properly when 1st factor is email/phone/passowrd
after answering problem in step 2, add tests for it too. either ensuse password is factor too, or ensure behaviour of email+email and email+password mfa flows

## Related PRs and Issues

x

## Checklist

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -1948,6 +1948,17 @@ App::post('/v1/account/tokens/magic-url')
         $result = $dbForProject->findOne('users', [Query::equal('email', [$email])]);
         if (!$result->isEmpty()) {
             $user->setAttributes($result->getArrayCopy());
+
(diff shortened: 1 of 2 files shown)


Meldiron:
Closed in favour of another PR that will add password as possible challenge type

PR #10648, opened 2025-10-15
Fix errors on downloading large deployment output

hmacr:
Task: https://linear.app/appwrite/issue/SER-218/large-deployment-download-doesnt-work


src/Appwrite/Platform/Modules/Sites/Http/Deployments/Download/Get.php
```diff
@@ -133,8 +125,15 @@ public function action(
                 ->setStatusCode(Response::STATUS_CODE_PARTIALCONTENT);
 
             $response->send($device->read($path, $start, ($end - $start + 1)));
```
Meldiron, line 133: Should we add here setContentType, and headerss x-peak, content-Disposition, and cache-control?

It would make it consustent with non-ranged requests

hmacr: yes, makes sense - will add it.

Meldiron (approved):

PR #10653, opened 2025-10-16
vectordb api endpoints

ArnabChatterjee20k:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Workers/Deletes.php
+++ b/src/Appwrite/Platform/Workers/Deletes.php
@@ -531,7 +531,7 @@ class Deletes extends Action
         $projectTables = !\in_array($dsn->getHost(), $sharedTables);
         $sharedTablesV1 = \in_array($dsn->getHost(), $sharedTablesV1);
         $sharedTablesV2 = !$projectTables && !$sharedTablesV1;
-        $databaseDSNKeys = ['database','documentsDatabase'];
+        $databaseDSNKeys = ['database','documentsDatabase','vectorDatabase'];
 
         $exectionActionPerDatabase = function (string $databaseDSNKey, $callback) use ($getDatabasesDB, $document) {
             /**
--- a/tests/e2e/Services/Projects/ProjectsConsoleClientTest.php
+++ b/tests/e2e/Services/Projects/ProjectsConsoleClientTest.php
@@ -183,6 +183,25 @@ class ProjectsConsoleClientTest extends Scope
         ]);
         $this->assertEquals(201, $documentsCollection['headers']['status-code']);
 
+        // Create vectordb database and collection
+        $vectorDb = $this->client->call(Client::METHOD_POST, '/vectordb', $projectAdminHeaders, [
+            'databaseId' => ID::unique(),
+            'name' => 'Vector DB',
+        ]);
+        $this->assertEquals(201, $vectorDb['headers']['status-code']);
+        $vectorDbId = $vectorDb['body']['$id'];
+
+        $vectorCollection = $this->client->call(Client::METHOD_POST, '/vectordb/' . $vectorDbId . '/collections', $projectAdminHeaders, [
+            'collectionId' => ID::unique(),
+            'name' => 'Vector Collection',
+            'dimension' => 3,
+            'documentSecurity' => true,
+            'permissions' => [
+                Permission::create(Role::any()),
+            ],
+        ]);
+        $this->assertEquals(201, $vectorCollection['headers']['status-code']);
+
         // Delete project
         $delete = $this->client->call(Client::METHOD_DELETE, '/projects/' . $projectId, array_merge([
             'content-type' => 'application/json',


.env
```diff
@@ -45,6 +45,10 @@ _APP_DB_ROOT_PASS=rootsecretpassword
 _APP_DB_ADAPTER_DOCUMENTSDB=mongodb
 _APP_DB_HOST_DOCUMENTSDB=mongodb
 _APP_DB_PORT_DOCUMENTSDB=27017
+_APP_DB_ADAPTER_VECTORDB=postgresql
+_APP_DB_HOST_VECTORDB=postgresql
+_APP_DB_PORT_VECTORDB=5432
+OLLAMA_MODELS=embeddinggemma
```
abnegate: Let's call this _APP_EMBEDDING_MODELS

app/config/collections/vectordb.php
```diff
@@ -0,0 +1,175 @@
...
+                'format' => '',
+                'size' => 16384,
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+        ],
+        'defaultAttributes' => [
```
abnegate, line 104: Why not include these in `attributes`?

app/config/collections/vectordb.php
```diff
@@ -0,0 +1,175 @@
...
+            [
+                '$id' => ID::custom('embeddings'),
+                'type' => Database::VAR_VECTOR,
+                'required' => true,
+                'size' => 128,
```
abnegate: I don't think we can pre-define this, doesn't size define the dimensions?

app/config/collections/vectordb.php
```diff
@@ -0,0 +1,175 @@
...
+            ],
+            [
+                '$id' => ID::custom('metadata'),
+                'type' => Database::VAR_OBJECT,
+                'default' => [],
+                'required' => false,
+                'size' => 0,
+                'signed' => false,
+                'array' => false,
+                'filters' => [],
```
abnegate, line 121: Just had a thought, why can't we use `VAR_STRING` with `json` filter for this?

app/init/registers.php
```diff
@@ -118,6 +119,14 @@
         'pass' => System::getEnv('_APP_DB_PASS', ''),
         'path' => System::getEnv('_APP_DB_SCHEMA', ''),
     ]);
+    $fallbackForVECTORDB = 'db_main=' . AppwriteURL::unparse([
```
abnegate: Let's fix casing

docker-compose.yml
```diff
@@ -1073,56 +1124,90 @@ services:
...
+  #     ME_CONFIG_BASICAUTH_USERNAME: ${_APP_DB_USER}
+  #     ME_CONFIG_BASICAUTH_PASSWORD: ${_APP_DB_PASS}
+  #   depends_on:
+  #     - mongodb
+
+
+  postgresql:
+    build:
+      context: .
+      dockerfile: postgres.dockerfile
```
abnegate: This is fine in dev, in prod we'll need to create a repo with a custom image like we do for MySQL

ollama.dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ollama/ollama:latest
```
abnegate: Let's pin a specific version

ollama.dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+# Preload specific models
+ENV MODELS="embeddinggemma"
```
abnegate: This should be an arg

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Collections/Create.php
```diff
@@ -0,0 +1,164 @@
...
+                'databaseInternalId' => $database->getSequence(),
+                'databaseId' => $databaseId,
+                '$permissions' => $permissions,
+                'documentSecurity' => $documentSecurity,
+                'enabled' => $enabled,
+                'name' => $name,
+                'dimensions' => $dimensions,
+                'embeddingModel' => $embeddingModel,
+                'search' => \implode(' ', [$collectionId, $name]),
+            ]));
```
abnegate, line 107: Why do we need the collection in dbForProject as well?

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Collections/Create.php
```diff
@@ -0,0 +1,164 @@
...
+            $attributes[] = new Document($attribute);
+        }
+        foreach ($collections['defaultIndexes'] as $index) {
+            $indexes[] = new Document($index);
+        }
+        try {
+            // passing null in creates only creates the metadata collection
+            if (!$dbForDatabases->exists(null, Database::METADATA)) {
+                $dbForDatabases->create();
+            }
```
abnegate, line 135: Can't remember, why can't this done on server startup in http.php?

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Collections/Indexes/Create.php
```diff
@@ -0,0 +1,72 @@
...
+                        code: SwooleResponse::STATUS_CODE_ACCEPTED,
+                        model: $this->getResponseModel(),
+                    )
+                ],
+                contentType: ContentType::JSON
+            ))
+            ->param('databaseId', '', new UID(), 'Database ID.')
+            ->param('collectionId', '', new UID(), 'Collection ID. You can create a new collection using the Database service [server integration](https://appwrite.io/docs/server/databases#databasesCreateCollection).')
+            ->param('key', null, new Key(), 'Index Key.')
+            ->param('type', null, new WhiteList([Database::INDEX_KEY, Database::INDEX_FULLTEXT, Database::INDEX_UNIQUE]), 'Index type.')
```
abnegate: Whitelist here should have `INDEX_HNSW_DOT, INDEX_HNSW_COSINE, INDEX_HNSW_EUCLIDEAN`

src/Appwrite/Utopia/Response.php
```diff
@@ -492,6 +499,7 @@ public function __construct(SwooleHTTPResponse $response)
             ->setModel(new Database())
             // Collection API Models
             ->setModel(new Collection())
+            ->setModel(new \Appwrite\Utopia\Response\Model\VectorDBCollection())
```
abnegate: Let's use an import here

tests/e2e/Services/Databases/VectorDB/DatabasesCustomClientTest.php
abnegate, line 1: Let's make sure to add CustomServer and ConsoleClient tests as well

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Collections/Create.php
ArnabChatterjee20k: database collections/registry is maintained in the dbForProject
example -> project_databases table will have database_1 details , database_2 details

dbForDatabases is having the actual collections
example -> database_1_collection_1, database_1_collection_2

We discussed to do this to have a centrailised collection of database records and dsn details

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Collections/Create.php
ArnabChatterjee20k: It is project specific metadata
Due to separation of dbForProject and dbForDatabases
Project is created in dbForProject and metadata also resides there

so doing it a separate one during the first time creating collections

app/config/collections/vectordb.php
ArnabChatterjee20k: As we discussed last time, if we use json then we can't query via metadata. Using var_object we can query nested metadata easily via contains and equals with jsonb

app/config/collections/vectordb.php
ArnabChatterjee20k: generally these attributes are created by users but in vector store we have these attributes pre setup -> embeddings, metadata as no option to add other attributes.

collection['attributes'] are created during the database creation and present in the metadata and not in the actual tables

so maintaining a centralised default attributes registry for creating these attributes and indexes after collection is created

src/Appwrite/Platform/Modules/Databases/Http/Databases/Action.php
```diff
@@ -21,6 +21,9 @@ public function setHttpPath(string $path): UtopiaAction
         if (\str_contains($path, '/documentsdb')) {
             $this->context = 'documentsdb';
         }
+        if (\str_contains($path, '/vectordb')) {
```
ArnabChatterjee20k: change this to use getDatabaseType() from the route method itself

app/controllers/general.php
```diff
@@ -1187,6 +1187,7 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
         $file = $error->getFile();
         $line = $error->getLine();
         $trace = $error->getTrace();
+        $trace = $error->getTrace();
```
abnegate:
```suggestion
```

app/init/registers.php
```diff
@@ -373,6 +400,9 @@
 $register->set('geodb', function () {
     return new Reader(__DIR__ . '/../assets/dbip/dbip-country-lite-2024-09.mmdb');
 });
+$register->set('embeddingAgent', function () {
+    return System::getEnv('_APP_EMBEDDING_ENDPOINT', 'http://ollama:11434/api/embed');
+});
```
abnegate: We can just get the env var in the resource instead of here

docker-compose.yml
abnegate: @ArnabChatterjee20k Created a repo here we can create the custom image in: https://github.com/appwrite/docker-postgres

docker-compose.yml
```diff
@@ -1124,6 +1174,42 @@ services:
...
+    environment:
+      - POSTGRES_DB=${_APP_DB_SCHEMA}
+      - POSTGRES_USER=${_APP_DB_USER}
+      - POSTGRES_PASSWORD=${_APP_DB_PASS}
+    command: "postgres"
+
+  ollama:
+    build: 
+      context: .
+      dockerfile: ollama.dockerfile
```
abnegate: And one here for this: https://github.com/appwrite/docker-ollama

docker-compose.yml
```diff
@@ -1232,4 +1319,5 @@ volumes:
   appwrite-functions:
   appwrite-sites:
   appwrite-builds:
-  appwrite-config:
\ No newline at end of file
+  appwrite-config:
+  _APP_EMBEDDING_MODELS:
```
abnegate: Let's follow convention here and name this `appwrite-models`

src/Appwrite/Platform/Modules/Databases/Http/Databases/Transactions/Update.php
```diff
@@ -298,6 +298,11 @@ public function action(string $transactionId, bool $commit, bool $rollback, Docu
                     $data = $data->getArrayCopy();
                 }
 
+                // Decode JSON strings that may have been encoded when storing nested arrays/objects
+                if (\is_array($data)) {
+                    $data = $this->decodeJsonStrings($data);
+                }
+
```
abnegate: This shouldn't be required 🤔

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Collections/Documents/Text/Create.php
```diff
@@ -0,0 +1,154 @@
...
+        $results = [];
+
+        // validating all documents first
+        foreach ($documents as $index => $item) {
+            if (!\is_array($item)) {
+                throw new Exception(Exception::DOCUMENT_INVALID_STRUCTURE, 'Invalid item at index ' . $index);
+            }
+
+            $text = $item['text'] ?? '';
+            $model = $item['embeddingModel'] ?? ($item['embeddigModel'] ?? null);
```
abnegate: Typo

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Embeddings/Text/Create.php
abnegate, line 1: Path should be like: `Databases/Http/VectorDB/Embeddings/Text/Create.php`

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Embeddings/Text/Create.php
```diff
@@ -0,0 +1,154 @@
...
+                    parameters: [
+                        new Parameter('databaseId', optional: false),
+                        new Parameter('collectionId', optional: false),
+                        new Parameter('documents', optional: false),
+                    ]
+                )
+            ])
+            ->param('databaseId', '', new UID(), 'Database ID.')
+            ->param('collectionId', '', new UID(), 'Collection ID. You can create a new collection using the Database service [server integration](https://appwrite.io/docs/server/databases#databasesCreateCollection). Make sure to define attributes before creating documents.')
+            ->param('documents', [], fn (array $plan) => new ArrayList(new JSON(), $plan['databasesBatchSize'] ?? APP_LIMIT_DATABASE_BATCH), 'Array of documents data as JSON objects.', true, ['plan'])
```
abnegate: Let's accept plain text instead of JSON here

src/Appwrite/Utopia/Response.php
```diff
@@ -868,7 +884,11 @@ public function yaml(array $data): void
 
         $this
             ->setContentType(Response::CONTENT_TYPE_YAML)
-            ->send(\yaml_emit($data, YAML_UTF8_ENCODING));
+            ->send((function (array $payload) {
+                $func = 'yaml_emit';
+                $encoding = \defined('YAML_UTF8_ENCODING') ? \constant('YAML_UTF8_ENCODING') : 0;
+                return \call_user_func($func, $payload, $encoding);
+            })($data));
```
abnegate: Why change it, isn't it the same unless the const isn't defined?

src/Appwrite/Platform/Modules/Databases/Http/Databases/Transactions/Update.php
ArnabChatterjee20k: I added this cause the nested structure is kept as encoded json only upon decode of $data
It is impacting vector and metadata
So doing a recursive decoding

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Embeddings/Text/Create.php
ArnabChatterjee20k: shall we turn this into Text though? Reason being the current api payload is like this
```
'documents' => [
                ['text' => 'hello world', 'embeddingModel' => 'embeddinggemma'],
                ['text' => 'second sentence', 'embeddingModel' => 'embeddinggemma'],
            ]
```
User can specify the embedding model explicitly for bulk request and in text we dont have the flexibility

src/Appwrite/Utopia/Response.php
ArnabChatterjee20k: Was getting few errors , thought of changing. Forgot to revert. Reverted now

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Embeddings/Text/Create.php
ArnabChatterjee20k: shall I turn this to having a single model for a request?

abnegate: > shall I turn this to having a single model for a request?

Yep :+1:

src/Appwrite/Platform/Modules/Databases/Http/Databases/Transactions/Update.php
abnegate: Is it fixed by the fix in the DB library for decoding? Or separate issue?

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Embeddings/Text/Create.php
abnegate: @ArnabChatterjee20k Let's make sure this one is addressed

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Embeddings/Text/Create.php
```diff
@@ -0,0 +1,145 @@
...
+use Appwrite\SDK\Response as SDKResponse;
+use Appwrite\Utopia\Response as UtopiaResponse;
+use Utopia\Database\Database;
+use Utopia\Database\Document;
+use Utopia\Database\Validator\Authorization;
+use Utopia\Swoole\Response as SwooleResponse;
+use Utopia\Validator\ArrayList;
+use Utopia\Validator\JSON;
+
+// Agent is dynamically provided via container; avoid strict type to pass lints
```
abnegate:
```suggestion
```

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Embeddings/Text/Create.php
```diff
@@ -0,0 +1,145 @@
...
+            ->inject('embeddingAgent')
+            ->inject('dbForProject')
+            ->inject('getDatabasesDB')
+            ->callback($this->action(...));
+    }
+
+    public function action(array $documents, UtopiaResponse $response, $embeddingAgent, Database $dbForProject, callable $getDatabasesDB): void
+    {
+        $isAPIKey = Auth::isAppUser(Authorization::getRoles());
+        $isPrivilegedUser = Auth::isPrivilegedUser(Authorization::getRoles());
```
abnegate: We'll need to change the `Auth::` methods here after syncing with 1.8.x

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Embeddings/Text/Create.php
```diff
@@ -0,0 +1,145 @@
...
+            if (!\is_string($text) || $text === '') {
+                throw new Exception(Exception::GENERAL_ARGUMENT_INVALID, 'Missing or invalid "text" at index ' . $index);
+            }
+            if (!\is_string($model) || $model === '') {
+                throw new Exception(Exception::GENERAL_ARGUMENT_INVALID, 'Missing or invalid "embeddingModel" at index ' . $index);
+            }
+            if (!empty($availableModels) && !\in_array($model, $availableModels, true)) {
+                throw new Exception(Exception::GENERAL_BAD_REQUEST, 'Unknown embedding model: ' . $model);
+            }
+        }
```
abnegate: Once we use a single model per request, we can remove this, since we only need to validate `Text`

In future if we do need to do something like this, the better approach is creating a new custom validator that enforces the structure and setting that on the parameter, instead of doing it in the route logic

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Embeddings/Text/Create.php
```diff
@@ -0,0 +1,145 @@
...
+
+            $results[] = new Document([
+                'model' => $model,
+                'dimensions' => $dimensions,
+                'embeddings' => $vector,
+            ]);
+
+        }
+        $list = new Document([
+            'embeddings' => array_map(fn ($d) => $d, $results),
```
abnegate: We can just pass $results here, array_map is doing nothing

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Embeddings/Text/Create.php
```diff
@@ -0,0 +1,145 @@
...
+                throw new Exception(Exception::GENERAL_ARGUMENT_INVALID, 'Missing or invalid "embeddingModel" at index ' . $index);
+            }
+            if (!empty($availableModels) && !\in_array($model, $availableModels, true)) {
+                throw new Exception(Exception::GENERAL_BAD_REQUEST, 'Unknown embedding model: ' . $model);
+            }
+        }
+        foreach ($documents as $index => $item) {
+            $embeddingAgent->getAdapter()->setModel($model);
+
+            $embedResult = $embeddingAgent->embed($text);
```
abnegate: Let's make sure to try/catch this, and throw a new custom error (in errors.php and Exception.php) `EMBEDDING_GENERATION_FAILED` if any fail

src/Appwrite/Platform/Modules/Databases/Http/Databases/Transactions/Update.php
ArnabChatterjee20k: Make sense, this was happening before the final merging of the object type pr. In that pr only this is fixed. Removed

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Embeddings/Text/Create.php
ArnabChatterjee20k: removed this logic and made it like the bulk route where it is automatically checked via outside the route action using auth

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Embeddings/Text/Create.php
```diff
@@ -0,0 +1,146 @@
...
+                    ],
+                    contentType: ContentType::JSON,
+                    parameters: [
+                        new Parameter('databaseId', optional: false),
+                        new Parameter('collectionId', optional: false),
+                        new Parameter('documents', optional: false),
+                    ]
+                )
+            ])
+            ->param('embeddingModel', Ollama::MODEL_EMBEDDING_GEMMA, new WhiteList(Ollama::MODELS), 'The embedding model to use for generating vector embeddings.', false)
```
abnegate: Let's just call this `model`

abnegate: Should be optional

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Embeddings/Text/Create.php
```diff
@@ -0,0 +1,146 @@
...
+            ->param('texts', [], fn (array $plan) => new ArrayList(new Text(0), $plan['databasesBatchSize'] ?? APP_LIMIT_DATABASE_BATCH), 'Array of text to generate embeddings.', true, ['plan'])
```
abnegate: Should be required, let's order this before model as well

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Embeddings/Text/Create.php
```diff
@@ -0,0 +1,146 @@
...
+            $embedding = [];
+            $error = '';
+            try {
+                $embedResult = $embeddingAgent->embed($text);
+                $embedding = $embedResult['embedding'] ?? [];
+                $totalDuration += $embedResult['totalDuration'] ?? 0;
+                $totalTokens += $embedResult['tokensProcessed'] ?? 0;
+            } catch (\Exception) {
+                $error = 'Error while generating embedding';
+            }
```
abnegate: Let's log errors here to sentry with the real trace

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Embeddings/Text/Create.php
```diff
@@ -0,0 +1,146 @@
...
+
+        $queueForStatsUsage
+            ->addMetric(
+                \str_replace(
+                    '{embeddingModel}',
+                    $embeddingModel,
+                    METRIC_EMBEDDINGS_TEXT
+                ),
+                \count($texts)
+            )
```
abnegate: Let's track failure count as well

src/Appwrite/Platform/Modules/Databases/Http/VectorDB/Collections/Documents/Create.php
```diff
@@ -0,0 +1,114 @@
...
+                        new Parameter('databaseId', optional: false),
+                        new Parameter('collectionId', optional: false),
+                        new Parameter('documents', optional: false),
+                    ]
+                )
+            ])
+            ->param('databaseId', '', new UID(), 'Database ID.')
+            ->param('documentId', '', new CustomId(), 'Document ID. Choose a custom ID or generate a random ID with `ID.unique()`. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.', true)
+            ->param('collectionId', '', new UID(), 'Collection ID. You can create a new collection using the Database service [server integration](https://appwrite.io/docs/server/databases#databasesCreateCollection). Make sure to define attributes before creating documents.')
+            ->param('data', [], new JSON(), 'Document data as JSON object.', true, example: '{"username":"walter.obrien","email":"walter.obrien@example.com","fullName":"Walter O\'Brien","age":30,"isAdmin":false}')
```
abnegate: Since we know the only attributes are embedding + metadata, let's add them as individual params instead of `data`. Then in the route logic, we can merge into `data` and call the parent action

src/Appwrite/Utopia/Response/Model/UsageVectorDBs.php
abnegate, line 1: What's the difference with `UsageVectorDB`?

ArnabChatterjee20k: Saw that Database follows this 'UsageDatabase' and 'UsageDatabases'
So to keep consistent with Database

PR #10682, opened 2025-10-22
Refactor authorization handling across multiple modules to use the ne…

shimonewman:


.github/labeler.yml
abnegate, line 1: Let's remove if not related to the PR

.github/workflows/auto-label-issue.yml
abnegate, line 1: Same here

CHANGES.md
```diff
@@ -1543,7 +1543,7 @@
 * Bump console to version 3.2.7 [#7148](https://github.com/appwrite/appwrite/pull/7148)
 * Chore update database to 0.45.2 [#7138](https://github.com/appwrite/appwrite/pull/7138)
 * Implement queue thresholds for the health API [#7123](https://github.com/appwrite/appwrite/pull/7123)
-* Add Authorization::skip to the usage worker [#7124](https://github.com/appwrite/appwrite/pull/7124)
+* Add Authorization->skip to the usage worker [#7124](https://github.com/appwrite/appwrite/pull/7124)
```
abnegate: Let's revert this one

app/cli.php
```diff
@@ -59,7 +57,13 @@
     return $register->get('pools');
 }, ['register']);
 
-CLI::setResource('dbForPlatform', function ($pools, $cache) {
+CLI::setResource('authorization', function () {
+    $authorization = new Authorization();
+    $authorization->disable();
+    return  $authorization;
```
abnegate:
```suggestion
    return $authorization;
```

app/cli.php
```diff
@@ -165,19 +171,20 @@
...
             $database->setTenant((int)$project->getSequence());
             return $database;
         }
 
         $adapter = new DatabasePool($pools->get('logs'));
         $database = new Database($adapter, $cache);
+        $database->setAuthorization($authorization);
 
         $database
             ->setSharedTables(true)
```
abnegate: Let's call setAuthorization along with all the other setters

app/controllers/api/account.php
```diff
@@ -173,7 +172,7 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
 $createSession = function (string $userId, string $secret, Request $request, Response $response, Document $user, Database $dbForProject, Document $project, Locale $locale, Reader $geodb, Event $queueForEvents, Mail $queueForMails) {
 
     /** @var Utopia\Database\Document $user */
-    $userFromRequest = Authorization::skip(fn () => $dbForProject->getDocument('users', $userId));
+    $userFromRequest = $dbForProject->getAuthorization()->skip(fn () => $dbForProject->getDocument('users', $userId));
```
abnegate, line 199: Let's inject `authorization` directly and use that instead of `$dbForProject->getAuthorization()` now that we have a separate resource for it. Then we don't have to worry about which DB we're using, and since it's an object reference, changes will apply to all DBs

app/init/resources.php
```diff
@@ -476,17 +485,18 @@
 
         $adapter = new DatabasePool($pools->get($dsn->getHost()));
         $database = new Database($adapter, $cache);
+        $database->setAuthorization($authorization);
```
abnegate: This one is redundant since we set it inside `configure`
```suggestion
```

src/Appwrite/Platform/Modules/Tokens/Http/Tokens/Buckets/Files/Action.php
```diff
@@ -6,32 +6,31 @@
...
+            throw new Exception(Exception::USER_UNAUTHORIZED, $authorization->getDescription());
         }
 
         $fileSecurity = $bucket->getAttribute('fileSecurity', false);
         if ($fileSecurity) {
             $file = $dbForProject->getDocument('bucket_' . $bucket->getSequence(), $fileId);
         } else {
-            $file = Authorization::skip(fn () => $dbForProject->getDocument('bucket_' . $bucket->getSequence(), $fileId));
+            $file = $authorization->skip(fn () => $dbForProject->getDocument('bucket_' . $bucket->getSequence(), $fileId));
         }
```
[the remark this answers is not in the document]
abnegate: This is intentional, users only need permissions at either collection or document level, if they have collection, we skip row level, because we know they're already permitted.

abnegate (approved):

PR #10683, opened 2025-10-22
fix: stats usage memory leak

loks0n:


src/Appwrite/Platform/Workers/StatsUsage.php
```diff
@@ -454,6 +453,7 @@ public function commitToDb(callable $getProjectDB): void
                 });
 
                 $dbForProject->upsertDocumentsWithIncrease('stats', 'value', $projectStats['stats']);
+                unset($this->projects[$sequence]);
```
stnguyen90: but the unset is already done in the finally block below 🧐

PR #10685, opened 2025-10-22
feat: Add provider info to the session data for OAuth2 token auth

adityaoberai:
## What does this PR do?

Right now, the `createOAuth2Token` flow does not allow developers to access the provider access token, access token expiry, refresh token, or UID, which can be quite inconvenient because a major benefit of using OAuth providers is to be able to interact with their APIs.

This issue ensures that any stored provider info is included in the session data when created via the `createSession` method in the `createOAuth2Token` flow

Fixes https://github.com/appwrite/appwrite/issues/8206

## Test Plan

- Locally build the app
- Configure any OAuth provider
- Test using the `createOAuth2Token` flow

### Local test

I ran a local test using the GitHub adapter and was able to get my user info from GitHub

- Session

image

image

- Test on GitHub API

image

image

## Related PRs and Issues

- #8206 

## Checklist

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,83 @@
+# Fixes and upgrades for the Appwrite Auth / Users / Teams services.
+"product / auth":
+  - "(auth|session|login|logout|register|2fa|mfa|users|teams|memberships|invite|oauth|oauth2|sso|jwt)"
+
+# Fixes and upgrades for the Appwrite Realtime API.
+"api / realtime":
+  - "(realtime|subscribe|websockets)"
+
+# Console, UI and UX issues
+"product / console":
+  - "(console)"
+
+# Fixes and upgrades for the Appwrite Storage.
+"product / storage":
+  - "(storage|bucket|file|image|preview|download)"
+
+# Fixes and upgrades for the Appwrite Database.
+"product / databases":
+  - "(database|collection|tables|attribute|column|document|row|query|queries|indexes|search|filter|sort|pagination)"
+
+# Fixes and upgrades for the Appwrite Functions.
+"product / functions":
+  - "(function|runtime|deployment|execution|trigger|cron|schedule)"
+
+# Fixes and upgrades for the Appwrite Docs.
+# "product / docs":
+#   -
+
+# Fixes and upgrades for the Appwrite Migrations.
+"product / migrations":
+  - "(migrate|migration)"
+
+# Fixes and upgrades for the Appwrite Messaging.
+"product / messaging":
+  - "(messaging|email|sms|push|provider|topic|target|notification)"
+
+# Fixes and upgrades for the Appwrite Platform.
+# "product / platform":
+#   -
+
+# Fixes and upgrades for database relationships
+"feature / relationships":
+  - "(relationship)"
+
+# Issues found only on Appwrite Cloud
+# "product / cloud":
+#   -
+
+# Fixes and upgrades for the Appwrite VCS.
+"product / vcs":
+  - "(repo|push|vcs|repository)"
+
+# Fixes and upgrades for the Appwrite GraphQL API.
+"api / graphql":
+  - "(graphql|gql|mutation)"
+
+# Fixes and upgrades for the Appwrite Assistant.
+"product / assistant":
+  - "(assistant)"
+
+# Fixes and upgrades for the Appwrite Domains.
+"product / domains":
+  - "(domain|dns|ssl|certificate)"
+
+# Fixes and upgrades for the Appwrite Locale.
+"product / locale":
+  - "(locale|i18n|internationalization|localization|l10n|translation|timezone|country)"
+
+# Fixes and upgrades for the Appwrite Avatars.
+"product / avatars":
+  - "(avatar|initial|flag|icon)"
+
+# Fixes and upgrades for Appwrite Sites.
+"product / sites":
+  - "(site|web|hosting|domain|ssl|certificate|nextjs|nuxt|react|angular|vue|svelte|astro)"
+
+# Fixes and upgrades for the Appwrite CLI.
+"sdk / cli":
+  - "(cli|command line)"
+
+# Issues only found when self-hosting Appwrite
+"product / self-hosted":
+  - "(self-host|self host)"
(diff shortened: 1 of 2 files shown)


app/controllers/api/account.php
```diff
@@ -201,19 +201,48 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
+    ];
+
+    // For OAuth2 tokens, retrieve and add provider tokens from identity
+    if ($verifiedToken->getAttribute('type') === Auth::TOKEN_TYPE_OAUTH2) {
+        // Find the most recently updated identity for this user with an access token
+        // This will be the identity that was just used in the OAuth2 flow
+        $identities = $dbForProject->find('identities', [
+            Query::equal('userInternalId', [$user->getSequence()]),
+            Query::orderDesc('$updatedAt'),
+            Query::limit(100)
```
eldadfux: Not loving this random number. Lets see if we have a common limit size that we use. IIRC we should have some constant that we use pretty often.

app/controllers/api/account.php
```diff
@@ -201,19 +201,48 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
+
+        if ($latestIdentity && !$latestIdentity->isEmpty()) {
+            $sessionData['providerUid'] = $latestIdentity->getAttribute('providerUid', '');
+            $sessionData['providerAccessToken'] = $latestIdentity->getAttribute('providerAccessToken', '');
+            $sessionData['providerRefreshToken'] = $latestIdentity->getAttribute('providerRefreshToken', '');
+            $sessionData['providerAccessTokenExpiry'] = $latestIdentity->getAttribute('providerAccessTokenExpiry', '');
+        }
+    }
+
     $session = new Document(array_merge(
```
eldadfux, line 236: How does this get expose without changing the session model? Did we already had support for the provider attributes?

app/controllers/api/account.php
```diff
@@ -201,19 +201,48 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
+        // This will be the identity that was just used in the OAuth2 flow
+        $identities = $dbForProject->find('identities', [
+            Query::equal('userInternalId', [$user->getSequence()]),
+            Query::orderDesc('$updatedAt'),
+            Query::limit(100)
+        ]);
+
+        // Find the first identity with a non-empty access token
+        $latestIdentity = null;
+        foreach ($identities as $identity) {
```
eldadfux: If we only need the latest identity, it will be much more efficient to actually query only the last row from the DB. That said, not sure if this is the most solid solution, don't we have a link between session X and identity Y?

app/controllers/api/account.php
adityaoberai: These provider attributes were already a part of our model, supported in the `createOAuth2Session` flow. The OAuth callback function up there was storing them in our DB anyway.

They just were not being exposed when the session was being generated with a token.

app/controllers/api/account.php
adityaoberai: I see a mix of numbers, although the most common is 1

We could just use 1 here to get the latest only

app/controllers/api/account.php
adityaoberai: You're right, I'll look into this tomorrow

app/controllers/api/account.php
adityaoberai: I've switched to using 1 so that we only get the latest update identity

app/controllers/api/account.php
adityaoberai: @eldadfux I was investigating this, and we unfortunately don't have a direct link aside from `userId` and `userInternalId`

eldadfux: Logic is pretty solid with just latest one but not 100% in case for example two sessions are created at the exact same time. @stnguyen90, @Meldiron do you have a better idea for this? I guess if not this is also fine, not a huge compromise.

app/controllers/api/account.php
```diff
@@ -201,19 +201,40 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
+    // For OAuth2 tokens, retrieve and add provider tokens from identity
+    if ($verifiedToken->getAttribute('type') === Auth::TOKEN_TYPE_OAUTH2) {
+        // Find the most recently updated identity for this user
+        $identities = $dbForProject->find('identities', [
+            Query::equal('userInternalId', [$user->getSequence()]),
+            Query::orderDesc('$updatedAt'),
+            Query::limit(1)
+        ]);
+
+        $latestIdentity = !empty($identities) ? $identities[0] : null;
```
Meldiron, line 225: We can do `findOne` instead.

But more concerning, what if user has multiple identities? I think for this implementation to work well, we need to include something from $verifiedToken in queries.

Sample scenario is when you have 2 identites - GitHub, and Google. Wrong order (what's first) could result in either missing accessToken, or accessToken of wrong oauth identity.

app/controllers/api/account.php
```diff
@@ -201,19 +201,40 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
...
+
+        if ($latestIdentity && !$latestIdentity->isEmpty()) {
+            $sessionData['provider'] = $latestIdentity->getAttribute('provider', '');
+            $sessionData['providerUid'] = $latestIdentity->getAttribute('providerUid', '');
+            $sessionData['providerAccessToken'] = $latestIdentity->getAttribute('providerAccessToken', '');
+            $sessionData['providerRefreshToken'] = $latestIdentity->getAttribute('providerRefreshToken', '');
+            $sessionData['providerAccessTokenExpiry'] = $latestIdentity->getAttribute('providerAccessTokenExpiry', '');
```
Meldiron, line 232: Let's write E2E test to ensure this behaviour. Would be lovely to have 2 identites and try with both, to prevent possible problem I mentioned in above comment. But we might only have one mock oauth adapter, so maybe that will not be easily possible, in which case I would skip it and trust manual QA for 2 identities for now

adityaoberai:
Closing this since it is fixed

PR #10708, opened 2025-10-27
Feat 1we 29 : Add email templates for account change notifications

bandaranaike:
## What does this PR do?

This PR introduces **three new email templates** to notify users about important account updates.  

1. **Email Change Notification** – Notifies the user via their *old email address* when their email has been changed.  
2. **Phone Change Notification** – Sends an email notification when a user updates their phone number.  
3. **Password Change Notification** – Sends an email notification when a user updates their password.  

All three templates are built using a **common method inspired by the existing sign-in alert email** implementation to ensure consistency and reusability.

---

## Test Plan

You can test the new templates using the Appwrite API client.  
*(You can replace `http://localhost` with your preferred test endpoint.)*

### Password Change
**Endpoint:**  
`POST http://localhost/v1/account/password`  
**Expected Result:**  
An email notification confirming the password change.

### Email Change
**Endpoint:**  
`POST http://localhost/v1/account/email`  
**Expected Result:**  
An email notification sent to the *old email address* confirming the email change.

### Phone Change
**Endpoint:**  
`POST http://localhost/v1/account/phone`  
**Expected Result:**  
An email notification confirming the phone number change.

---

## Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)  
- [x] If the PR includes a change to an API's metadata (description, label, parameters, etc.), it also includes updated API specs and example documentation.

Screenshot 2025-10-27 230158
Screenshot 2025-10-27 230116
Screenshot 2025-10-27 230029

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,7 +53,7 @@
         "utopia-php/cli": "0.15.*",
         "utopia-php/config": "0.2.*",
         "utopia-php/database": "3.*",
-        "utopia-php/detector": "0.1.*",
+        "utopia-php/detector": "0.2.*",
         "utopia-php/domains": "0.9.*",
         "utopia-php/dns": "0.3.*",
         "utopia-php/dsn": "0.2.1",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "6f773bcdf583a863b0f2404958330f98",
+    "content-hash": "f826d3b283b10af98dfd565c4187a83a",
     "packages": [
         {
             "name": "adhocore/jwt",
@@ -3898,16 +3898,16 @@
         },
         {
             "name": "utopia-php/detector",
-            "version": "0.1.5",
+            "version": "0.2.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/utopia-php/detector.git",
-                "reference": "b5d6ba51352485b524589bc0ee8d07a9efafe718"
+                "reference": "89f96e864220de13800cf398a1f1686a85401eaa"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/utopia-php/detector/zipball/b5d6ba51352485b524589bc0ee8d07a9efafe718",
-                "reference": "b5d6ba51352485b524589bc0ee8d07a9efafe718",
+                "url": "https://api.github.com/repos/utopia-php/detector/zipball/89f96e864220de13800cf398a1f1686a85401eaa",
+                "reference": "89f96e864220de13800cf398a1f1686a85401eaa",
                 "shasum": ""
             },
             "require": {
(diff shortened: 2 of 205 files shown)


app/controllers/api/account.php
```diff
@@ -70,105 +70,274 @@
 $oauthDefaultSuccess = '/console/auth/oauth2/success';
 $oauthDefaultFailure = '/console/auth/oauth2/failure';
 
-function sendSessionAlert(Locale $locale, Document $user, Document $project, Document $session, Mail $queueForMails)
+/**
+ * Mask email by replacing part of the username and domain.
+ * Example: john.doe@example.com → j***e@e******.com
+ */
+function maskEmail(string $email): string
```
Meldiron: Global functions like `maskPhone` and `maskEmail` are generally bad practice in API controllers. Let's make a class out of it, imagine it as a "mini-library".

You can create this for example in `src/Appwrite/Redaction` - a redaction library.
Here, let's follow OOP and create adapter-driven redaction behaviour.

When this Redaction is used, code could look something like this (might be wrong on details, just how I can imagine it right now):

```
$value = 'matej@appwrite.io';

$adapter = new Redaction\Adapters\Email();
$adapter->setRedactDomain(true);
$adapter->setRedactUser(true);
$adapter->setRedactTLD(false);

$redactor = new Redaction\Redactor($adapter);

try {
  $redaction = $redactor->redact($value);
} catch (Redaction\Exceptions\Redaction $err) {
 throw new Exception('Redaction failed');
}

// technically, we could also have $redaction->getOriginalValue() or $redatcion->getAdapter() ..

$value = $redaction->getValue();
```

app/controllers/api/account.php
```diff
@@ -3167,6 +3349,9 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, Doc
             if ($oldTarget instanceof Document && !$oldTarget->isEmpty()) {
                 Authorization::skip(fn () => $dbForProject->updateDocument('targets', $oldTarget->getId(), $oldTarget->setAttribute('identifier', $phone)));
             }
+
+            sendPhoneChangeEmail($locale, $user, $project, $queueForMails, $oldPhone, $phone);
```
Meldiron, line 3362: Let's update tests to ensure all those emails arrive after respective actions, and include expected contents (for example phone number, and a specific subject)

app/config/locale/templates/email-email-change.tpl
```diff
@@ -0,0 +1,15 @@
+<p>{{hello}}</p>
```
Meldiron, line 1: Let's please add screenshots  to PR description, of each email

app/controllers/api/account.php
```diff
@@ -70,105 +70,274 @@
...
+}
+
+/**
+ * Mask a phone number by hiding middle digits.
+ * Example: +65 9876 5432 → +65 **** 5432
+ */
+function maskPhone(string $phone): string
+{
+    // Keep country code and last 4 digits visible
+    return preg_replace('/(\+?\d{1,3})?[\s-]?(\d{2,4})\d{3,4}(\d{2,4})/', '$1 **** $3', $phone);
```
Meldiron: let's avoid regex, they can cause some performance and security issues easily, plus are pretty hard to read.

let's write some raw php code to achieve it

app/controllers/api/account.php
bandaranaike: This one done

app/controllers/api/account.php
bandaranaike: Wrote this method using raw PHP

app/config/locale/templates/email-email-change.tpl
bandaranaike: Added 3 Screenshots

app/controllers/api/account.php
bandaranaike: Added `testSecurityEmailNotifications` method in `tests/e2e/Services/Account/AccountSecurityEmailsTest.php`  new file

PR #10718, opened 2025-10-28
Add agentic workflow issue-triage

stnguyen90:
Add agentic workflow issue-triage


EVDOG4LIFE (approved):
I'm down to see what this does out of the box. We'll likely need to customize the prompt so it's able to better triage issues with context on our processes.

PR #10722, opened 2025-10-28
feat: per bucket image transformations flag

ChiragAgg5k:


src/Appwrite/Migration/Version/V23.php
```diff
@@ -132,6 +132,14 @@ private function migrateCollections(): void
...
                     $this->dbForProject->purgeCachedCollection($id);
                     break;
+                case 'buckets':
+                    try {
+                        $this->createAttributeFromCollection($this->dbForProject, $id, 'transformations');
+                    } catch (Throwable $th) {
+                        Console::warning("'transformations' from {$id}: {$th->getMessage()}");
+                    }
+                    $this->dbForProject->purgeCachedCollection($id);
+                    break;
```
ChiragAgg5k, line 150: we only need to add it once here since the migration cycles all projects, including console - 

Image

abnegate (approved):

PR #10751, opened 2025-11-03
fix: Throw error when file token expiry is in the past

### LINKED ISSUE #10323: 🐛 Bug Report: Error creating expired file token
### 👟 Reproduction steps

1. Create a file token with an expiration in the past

### 👍 Expected behavior

1. The Console should not allow it
2. The API call should return an error saying the date is invalid

### 👎 Actual Behavior

The Console allows it:

Image

and then there's a server error:

Image

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

Linux

### 🧱 Your Environment

_No response_

hmacr:
Task: https://linear.app/appwrite/issue/SER-279/error-creating-expired-file-token
Closes #10323

### DIFF
--- a/tests/e2e/Services/Tokens/TokensConsoleClientTest.php
+++ b/tests/e2e/Services/Tokens/TokensConsoleClientTest.php
@@ -70,7 +70,7 @@ class TokensConsoleClientTest extends Scope
             'expire' => '2022-11-02',
         ]);
         $this->assertEquals(400, $token['headers']['status-code']);
-        $this->assertEquals('Token expiry date must be a valid date, and at least 1 day from now', $token['body']['message']);
+        $this->assertStringContainsString('Value must be valid date in the future', $token['body']['message']);
 
         // Success case: No expire date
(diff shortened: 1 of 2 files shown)


src/Appwrite/Platform/Modules/Tokens/Http/Tokens/Buckets/Files/Create.php
```diff
@@ -70,6 +70,13 @@ public function __construct()
 
     public function action(string $bucketId, string $fileId, ?string $expire, Response $response, Database $dbForProject, Event $queueForEvents): void
     {
+        if (!is_null($expire)) {
+            $validator = new DatetimeValidator(requireDateInFuture: true, precision: DateTimeValidator::PRECISION_DAYS, offset: 0);
```
Meldiron: This is not in param directly, to allow us to change exception message.
But can we make validaiton description nicer, so we dont need this?

There are more endpoitns that would benefit from nicer error message.

Meldiron (approved):

PR #10761, opened 2025-11-04
feat: bump utopia dns

loks0n:


src/Appwrite/Network/Validator/DNS.php
```diff
@@ -3,144 +3,112 @@
...
-    /**
-     * Check if DNS record value matches specific value
-     *
-     * @param mixed $domain
-     * @return bool
-     */
     public function isValid($value): bool
     {
-        if (!is_string($value)) {
+        if (!is_string($value) || $value === '') {
```
ChiragAgg5k: should we do `trim($value) === ''`?

loks0n: updated

PR #10762, opened 2025-11-04
chore: release python sdk to 13.6.1

ChiragAgg5k:


composer.json
```diff
@@ -107,23 +107,5 @@
             "php-http/discovery": true,
             "tbachert/spi": true
         }
-    },
-    "repositories": [
```
ChiragAgg5k, line 111: this is intentional and discussed with Jake

lohanidamodar (approved):

PR #10772, opened 2025-11-06
Set proper access-control-allow-origin for OPTIONS request

hmacr:
Task: https://linear.app/appwrite/issue/SER-486/double-check-cors-behaviour


app/controllers/general.php
Meldiron, line 1: Lets add a test for this new behaviour

hmacr:
It's already

PR #10791, opened 2025-11-10
chore: release cli and php

ChiragAgg5k:


app/config/platforms.php
```diff
@@ -281,7 +281,7 @@
             [
                 'key' => 'php',
                 'name' => 'PHP',
-                'version' => '17.5.0',
+                'version' => '17.6.0',
```
Meldiron: major bump I think

docs/examples/1.8.x/server-php/examples/health/get-failed-jobs.md
```diff
@@ -12,6 +12,6 @@ $client = (new Client())
 $health = new Health($client);
 
 $result = $health->getFailedJobs(
-    name: ::V1DATABASE(),
+    name: Name::V1DATABASE(),
```
loks0n, line 15: Maybe we should rename this QueueName? Can we create a backlog task?

PR #10792, opened 2025-11-10
feat: add metric validation in StatsUsage worker

ChiragAgg5k:
## What does this PR do?

This PR adds validation to the `StatsUsage` worker to ensure only metrics defined in `constants.php` are processed and stored in the database. Previously, any metric could be added to the payload without verification, which could lead to invalid or malicious data being stored.

## Changes

- Added `$validMetrics` array containing all static metric constants from `constants.php`
- Added `$validMetricPatterns` array with regex patterns to validate dynamic metrics (those with placeholders like `{databaseInternalId}`, `{resourceType}`, etc.)
- Added `isValidMetric()` method to validate metric keys against both static and pattern-based metrics
- Updated `action()` method to skip invalid metrics with console warning messages
- Only valid metrics are now processed and counted

## Benefits

✅ **Security**: Unknown or malicious metrics are rejected  
✅ **Data Integrity**: Only metrics defined in constants are processed  
✅ **Debugging**: Console warnings alert when unknown metrics are received  
✅ **Performance**: Invalid metrics are skipped early, reducing unnecessary processing

## Test Plan

- [x] Syntax validation passed
- [ ] Test with valid metrics - should process normally
- [ ] Test with invalid metrics - should skip with warning
- [ ] Test with dynamic metrics (e.g., `123.collections`) - should validate correctly

## Related Issues

Closes #<issue_number_if_any>

### DIFF
--- a/src/Appwrite/Platform/Workers/StatsUsage.php
+++ b/src/Appwrite/Platform/Workers/StatsUsage.php
@@ -162,9 +162,9 @@ class StatsUsage extends Action
         METRIC_MESSAGES_TYPE => '/^messages\.(email|sms|push)$/',
         METRIC_MESSAGES_TYPE_SENT => '/^messages\.(email|sms|push)\.sent$/',
         METRIC_MESSAGES_TYPE_FAILED => '/^messages\.(email|sms|push)\.failed$/',
-        METRIC_MESSAGES_TYPE_PROVIDER => '/^messages\.(email|sms|push)\..+$/',
-        METRIC_MESSAGES_TYPE_PROVIDER_SENT => '/^messages\.(email|sms|push)\..+\.sent$/',
-        METRIC_MESSAGES_TYPE_PROVIDER_FAILED => '/^messages\.(email|sms|push)\..+\.failed$/',
+        METRIC_MESSAGES_TYPE_PROVIDER => '/^messages\.(email|sms|push)\.[^.]+$/',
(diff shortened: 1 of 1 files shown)


src/Appwrite/Platform/Workers/StatsUsage.php
```diff
@@ -162,13 +300,20 @@ public function action(Message $message, callable $getProjectDB, callable $getLo
         $this->stats[$projectId]['project'] = $project;
         $this->stats[$projectId]['receivedAt'] = DateTime::now();
         foreach ($payload['metrics'] ?? [] as $metric) {
+            $metricKey = $metric['key'] ?? '';
+
+            // Skip invalid metrics
+            if (!$this->isValidMetric($metricKey)) {
+                throw new Exception('Invalid metric key: ' . $metricKey);
```
lohanidamodar: shouldn't this log and skip, instead of throwing exception? don't throwing exception exit this foreach loop missing other valid metrics?

src/Appwrite/Platform/Workers/StatsUsage.php
```diff
@@ -119,24 +244,47 @@ public function __construct()
...
+     * @param string $metricKey
+     * @return bool
+     */
+    protected function isValidMetric(string $metricKey): bool
+    {
+        if (in_array($metricKey, $this->validMetrics)) {
+            return true;
+        }
+
+        foreach ($this->validMetricPatterns as $pattern => $regex) {
```
lohanidamodar, line 264: I'm worried about the additional time this takes patter matching for each key, we are already slow in stats usage containers.

PR #10824, opened 2025-11-17
Feat docker geo

lohanidamodar:
## What does this PR do?

- Add docker geo db container
- Use Geo DB from docker container

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


composer.json
```diff
@@ -58,7 +58,7 @@
         "utopia-php/dns": "1.1.*",
         "utopia-php/dsn": "0.2.1",
         "utopia-php/framework": "0.33.*",
-        "utopia-php/fetch": "0.4.*",
+        "utopia-php/fetch": "dev-feat-support-base-url as 0.4.2",
```
Meldiron: Let's release

docker-compose.yml
```diff
@@ -213,6 +213,7 @@ services:
       - _APP_DATABASE_SHARED_NAMESPACE
       - _APP_FUNCTIONS_CREATION_ABUSE_LIMIT
       - _APP_CUSTOM_DOMAIN_DENY_LIST
+      - _APP_GEO_SECRET=$_APP_GEO_SECRET
```
Meldiron: What is this secret - why does both appwrite-geo and appwrite container need it?

I was thinking it's some external API key, but then I would expect only appwrite-geo needs it.

Meldiron: We might need to update `dependsOn` array too

docker-compose.yml
```diff
@@ -961,6 +962,14 @@ services:
...
     networks:
       - appwrite
+  
+  appwrite-geo:
+    container_name: appwrite-geo
+    image: appwrite/geo:0.1.0
+    networks:
+      - appwrite
+    environment:
+      - GEO_SECRET=$_APP_GEO_SECRET
```
Meldiron, line 1116: Just double checking,
- Does this container have healthcheck?
- Does this container have security protections - Auth header or JWT?

app/init/resources.php
```diff
@@ -691,6 +693,75 @@ function getDevice(string $root, string $connection = ''): Device
     return $register->get('geodb');
 }, ['register']);
 
+App::setResource('geoRecord', function (Reader $geodb, Request $request, Locale $locale) {
+    $ip = $request->getIp();
+
+    if (!filter_var($ip, FILTER_VALIDATE_IP)) {
```
Meldiron: Does this work fine with ipv6?

app/init/resources.php
```diff
@@ -691,6 +693,75 @@ function getDevice(string $root, string $connection = ''): Device
...
+        $record = $client->fetch("/ips/{$ip}", Client::METHOD_GET);
+        if ($record->getStatusCode() !== 200) {
+            $error = $record->json();
+            throw new Exception(Exception::GENERAL_SERVER_ERROR, $error['message'] ?? 'Failed to fetch geo data');
+        }
+        $record = $record->json();
+    } catch (Throwable $th) {
+        Console::error($th->getMessage());
+        Console::error($th->getTraceAsString());
+        $record = null;
```
Meldiron: Should we have fallback for current behaviour, so self-hosted still get same results?

app/init/resources.php
```diff
@@ -691,6 +693,75 @@ function getDevice(string $root, string $connection = ''): Device
...
+        if ($dbRecord) {
+            $record = [];
+            $record['countryCode'] = $dbRecord['country']['iso_code'] ?? '--';
+            $record['country'] = $dbRecord['country']['names'] ?? [];
+            $record['continent'] = $dbRecord['continent']['names'] ?? [];
+            $record['continentCode'] = $dbRecord['continent']['code'] ?? '--';
+        }
+    }
+
+    $output = [];
```
Meldiron: I think we should use Document here, to be consistent with other resources

Meldiron: Or maybe even `class GeoRecord extends  Document` and add methods like `getCountryCode`, like we have `getCreatedAt`

app/config/collections/common.php
```diff
@@ -867,6 +867,138 @@
                 'array' => false,
                 'filters' => [],
             ],
+            [
```
Meldiron, line 895: Let's note down we will need migration

app/config/collections/common.php
```diff
@@ -867,6 +867,138 @@
...
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('longitude'),
+                'type' => Database::VAR_FLOAT,
+                'format' => '',
+                'size' => 2,
```
Meldiron: Size 2 sounds scarry. Lets check with Smuel, maybe few extra bites wont hurt too much, and prevents possible problems in future

app/config/collections/common.php
```diff
@@ -867,6 +867,138 @@
...
+                'format' => '',
+                'size' => 255,
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('userType'),
```
Meldiron: We should have prefix here I think, it might indicate Appwrite's user type, not user type based on his IP. similar for organization, connectinType, ..

app/config/collections/common.php
```diff
@@ -867,6 +867,138 @@
...
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('isp'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 255,
```
Meldiron, line 1020: Let's check once wit hShmuel. We are adding many strings of 255, and I think there is some row length limit. Maybe making some of then large size (MEDIUMTEXT) wound be better, of we never query by them.

docker-compose.yml
lohanidamodar: The geo container will only be exposed to Appwrite so that appwrite can use it for it's locale and sessions for now.

docker-compose.yml
lohanidamodar: this container for now is not exposed it's used by appwrite container only

app/init/resources.php
lohanidamodar: Yes, FILTER_VALIDATE_IP validates both ipv6 and ipv4

app/init/resources.php
lohanidamodar: we do have a fallback, if docker container fails we still try to get from the file as we currently do, nothing changes.

app/init/resources.php
lohanidamodar: right, this makes sense.

app/config/collections/common.php
lohanidamodar: Where?

app/config/collections/common.php
lohanidamodar: i'm using the limits that are set by the the geo db on their schema.

docker-compose.yml
lohanidamodar: hmm, do we want to? we do have a fallback if it fails?

docker-compose.yml
lohanidamodar: We do have auth header, we should probably add healthcheck endpoint as well

app/init/resources.php
lohanidamodar: updated.

app/config/collections/common.php
lohanidamodar: https://github.com/utopia-php/database/blob/fe7a1326ad623609e65587fe8c01a630a7075fee/src/Database/Adapter/MariaDB.php#L1689
For float size doesn't matter.

eldadfux (approved):

loks0n (approved):

app/controllers/api/account.php
```diff
@@ -189,12 +199,28 @@
...
+        default => true
+    };
+
+    $hasUserEmail = $user->getAttribute('email', false) !== false;
+
+    $isSessionAlertsEnabled = $project->getAttribute('auths', [])['sessionAlerts'] ?? false;
+
+    $isNotFirstSession = $dbForProject->count('sessions', [
+        Query::equal('userId', [$user->getId()]),
+    ]) !== 1;
```
loks0n: this logic is not neccessary, the Email listener does it, we should dispatch the event unconditionally

loks0n (changes requested):

app/controllers/api/users.php
```diff
@@ -975,10 +975,9 @@ function createUser(Hash $hash, string $userId, ?string $email, ?string $passwor
     ->param('total', true, new Boolean(true), 'When set to false, the total count returned will be 0 and will not be calculated.', true)
     ->inject('response')
     ->inject('dbForProject')
-    ->inject('locale')
-    ->inject('geodb')
+    ->inject('getGeoForIp')
```
loks0n: small suggestion for the structure of this service
```
interface Geo {
     getRecordFromIp(string $ip): GeoRecord
}
```

app/config/collections/common.php
loks0n: I guess Matej means for the new columns

app/controllers/api/account.php
```diff
@@ -1902,7 +1963,6 @@
             // If the `token` param is not set, we persist the session in a cookie
         } else {
             $detector = new Detector($request->getUserAgent('UNKNOWN'));
-            $record = $geodb->get($request->getIP());
             $secret = $proofForToken->generate();
 
             $session = new Document(array_merge([
```
loks0n, line 1977: Small quality improvement for later, might be nice to make a typed Session obj

lohanidamodar:
Re: latest Greptile summary (2/5). Notes on each remaining flag for reviewer context:

**1. Connection field key mappings (`connection`/`user`/`organization`) look speculative**
These mappings are intentional and mirror the authoritative implementation in the Cloud repo's `getGeoForIp` (used by all premium-geo deployments in production). The extra session columns are populated only when a richer geo backend (premium docker-geo or a custom deployment) is present; the base `appwrite/geo:0.3.0` image only returns country/continent, so the other columns stay null — that's expected forward-compatibility.

**2. `docker-compose.yml` `service_healthy` hard dependency**
Reverted at `870716b063`. `depends_on` for `appwrite-geo` is back to the short-form list, so a geo startup failure no longer blocks the main service beyond the default `service_started` behavior.

**3. `.env` default `_APP_GEO_SECRET=your-secret-key`**
This follows the established placeholder pattern used for every other secret in the file (`_APP_OPENSSL_KEY_V1=your-secret-key`, `_APP_EXECUTOR_SECRET=your-secret-key`). Admins are expected to override on install; no operational change proposed.

**4. Fail-closed denylist behavior**
Intentional. `_APP_CONSOLE_COUNTRIES_DENYLIST` is a "block these countries" security policy — an unreachable geo service cannot prove the caller is outside the blocked set, so denying access is the correct security posture. `isLookupSucceeded()` distinguishes "service down" (fail closed) from "IP genuinely unknown to the DB" (fall through). This flip from Greptile's earlier "fail-open bypass" concern is the intended resolution.

PR #10829, opened 2025-11-18
fix: format getUsage

ChiragAgg5k:


src/Appwrite/SDK/Specification/Format.php
```diff
@@ -534,9 +534,12 @@ public function getRequestEnumKeys(string $service, string $method, string $para
...
-                    case 'getUsage':
-                        // Range Enum Keys
-                        return ['One Hour', 'One Day'];
+                    default:
+                        if (str_starts_with($method, 'getUsage')) {
+                            // Range Enum Keys
+                            return ['One Hour', 'One Day'];
+                        }
+                        break;
                 }
```
lohanidamodar, line 543: won't this affect other endpoints like organization usage etc?

lohanidamodar: Also, can we extend this specification format in cloud and then add this hack there only?

PR #10838, opened 2025-11-19
chore: release flutter 20.3.2

ChiragAgg5k:


docker-compose.override.yml
```diff
@@ -1,7 +0,0 @@
-services:
-  appwrite:
-    # volumes:
-    #   - ~/.ssh:/root/.ssh
-    environment:
-      - GH_TOKEN=
-      - GIT_EMAIL=
```
Meldiron, line 7: Intentional? I didnt even know this exist 😅

ChiragAgg5k: yes, intentional. it was added by steven to simplify sdk release process, but it did not work :D

PR #10842, opened 2025-11-19
Feat: Custom start commands

Meldiron:
## What does this PR do?

Adds custom start commands, for sites only right now

## Test Plan

New tests

## Related PRs and Issues

x

## Checklist


tests/e2e/Services/Sites/SitesCustomServerTest.php
```diff
@@ -2939,4 +2939,63 @@ public function testCookieHeader()
 
         $this->cleanupSite($siteId);
     }
+
+    public function testSiteCustomStartCommand(): void
```
ChiragAgg5k, line 2970: should we add a test for functions as well?

PR #10848, opened 2025-11-20
fix: override template deployment type enum name

ChiragAgg5k:


src/Appwrite/SDK/Specification/Format.php
```diff
@@ -289,6 +295,12 @@ protected function getRequestEnumName(string $service, string $method, string $p
                                 return 'VCSDeploymentType';
                         }
                         break;
+                    case 'createTemplateDeployment':
+                        switch ($param) {
+                            case 'type':
+                                return 'TemplateDeploymentType';
```
Meldiron:
```suggestion
                                return 'TemplateReferenceType';
```

src/Appwrite/SDK/Specification/Format.php
```diff
@@ -289,6 +295,12 @@ protected function getRequestEnumName(string $service, string $method, string $p
                                 return 'VCSDeploymentType';
```
Meldiron, line 295: add todo to update with major version to `VCSReferenceType`

PR #10876, opened 2025-11-27
feat: add trusted headers variable and setup enhanced getIP

levivannoort:
## What does this PR do?

As not to default on the `x-forwarded-for` header, this pull-request introduces the option to set `trusted headers`. Which are checked from left to right on request presence wise, where the left most address will be used for as the client reference. With a default set to `x-forwarded-for`.

In addition I spotted a missing language reference in `getting started` and the `sdk` references for the `go`.

## Test Plan

Setup tests for the various edge cases I could think of in relation to passing headers into the variable.

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/http.php
+++ b/app/http.php
@@ -478,7 +478,7 @@ $http->on(Constant::EVENT_REQUEST, function (SwooleRequest $swooleRequest, Swool
             if (isset($user) && !$user->isEmpty()) {
                 $log->setUser(new User($user->getId()));
             } else {
-                $log->setUser(new User('guest-' . hash('sha256', TrustedIp::extract($request))));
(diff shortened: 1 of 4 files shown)


levivannoort:
@coderabbitai review

src/Appwrite/Utopia/Request.php
```diff
@@ -233,4 +234,45 @@ public function cacheIdentifier(): string
...
+
+    /**
+     * Get IP
+     *
+     * Returns users IP address based on a list of trusted headers.
+     * Assumes application is only accessible through a trusted proxy.
+     *
+     * @return string
+     */
+    public function getIP(): string
```
loks0n: from sync: maybe we can take a array<string> param here instead of depending on env?

reason being that this is extension to library code, should be agnostic to Appwrite-specific env vars

loks0n (approved):

PR #10877, opened 2025-11-27
Fix file token expiry

hmacr:
Closes SER-648

### DIFF
--- a/app/init/resources.php
+++ b/app/init/resources.php
@@ -1004,7 +1004,7 @@ App::setResource('resourceToken', function ($project, $dbForProject, $request) {
 
     if (!empty($tokenJWT) && !$project->isEmpty()) { // JWT authentication
         // Use a large but reasonable maxAge to avoid auto-exp when token has no expiry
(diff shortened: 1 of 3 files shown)


app/init/resources.php
```diff
@@ -1003,7 +1003,8 @@ function getDevice(string $root, string $connection = ''): Device
     $tokenJWT = $request->getParam('token');
 
     if (!empty($tokenJWT) && !$project->isEmpty()) { // JWT authentication
-        $jwt = new JWT(System::getEnv('_APP_OPENSSL_KEY_V1'), 'HS256', 900, 10); // Instantiate with key, algo, maxAge and leeway.
+        // Use a large but reasonable maxAge to avoid auto-exp when token has no expiry
+        $jwt = new JWT(System::getEnv('_APP_OPENSSL_KEY_V1'), 'HS256', 86400 * 365 * 10, 10); // Instantiate with key, algo, maxAge and leeway.
```
Meldiron: We also have this expiry here: https://github.com/appwrite/appwrite/blob/95abe61bab4817b141952b9c35d101e9299240c0/src/Appwrite/Utopia/Response/Model/ResourceToken.php#L67

We might want to make it into a const here: https://github.com/appwrite/appwrite/blob/95abe61bab4817b141952b9c35d101e9299240c0/app/init/constants.php#L93

PR #10880, opened 2025-11-28
Support query limit and offset in list repos API

hmacr:
Closes SER-504

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -75,7 +75,7 @@
         "utopia-php/swoole": "0.8.*",
         "utopia-php/system": "0.9.*",
         "utopia-php/telemetry": "0.1.*",
-        "utopia-php/vcs": "dev-ser-504 as 0.12.99",
+        "utopia-php/vcs": "0.13.*",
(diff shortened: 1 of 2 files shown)


app/controllers/api/vcs.php
```diff
@@ -1052,11 +1056,16 @@
...
 
-        $page = 1;
-        $perPage = 4;
+        $queries = Query::parseQueries($queries);
+        $limitQuery = current(array_filter($queries, fn ($query) => $query->getMethod() === Query::TYPE_LIMIT));
+        $offsetQuery = current(array_filter($queries, fn ($query) => $query->getMethod() === Query::TYPE_OFFSET));
+
+        $limit = !empty($limitQuery) ? $limitQuery->getValue() : 4;
+        $offset = !empty($offsetQuery) ? $offsetQuery->getValue() : 0;
+        $page = ($offset / $limit) + 1;
```
Meldiron: Apply exception if we cannot promise exact offset

PR #10887, opened 2025-12-01
chore: bump assistant 0.8.4

loks0n:


Dockerfile
```diff
@@ -36,7 +36,6 @@ COPY --from=composer /usr/local/src/vendor /usr/src/code/vendor
 COPY ./app /usr/src/code/app
 COPY ./public /usr/src/code/public
 COPY ./bin /usr/local/bin
-COPY ./docs /usr/src/code/docs
```
ItzNotABug, line 39: won't this cause issues when trying to regen docs? its a lot of spam noise on terminal. We should instead have this for the release stage maybe.

ItzNotABug: Stale.

PR #10911, opened 2025-12-06
feat: multiple app domains

loks0n:


.env
```diff
@@ -22,7 +22,7 @@ _APP_OPTIONS_FORCE_HTTPS=disabled
 _APP_OPTIONS_ROUTER_FORCE_HTTPS=disabled
 _APP_OPENSSL_KEY_V1=your-secret-key
 _APP_DNS=8.8.8.8
-_APP_DOMAIN=traefik
+_APP_DOMAIN=appwrite.test
```
abnegate: Shouldn't this be localhost for dev? It breaks some links already as `traefik`

TorstenDittmann (approved):

Issue #10927, opened 2025-12-10
users.listMfaFactors() returns incorrect recoveryCode value (inconsistent with account.listMfaFactors())


alexfreeman90:
### 👟 Reproduction steps

1. Create a user account
2. Enable MFA on the account
3. Generate recovery codes using `account.createMfaRecoveryCodes()`
4. Call `account.listMfaFactors()` — returns `recoveryCode: true` ✅
5. Call `users.listMfaFactors(userId)` for the same user — returns `recoveryCode: false` ❌

### 👍 Expected behavior

Both endpoints should return identical MFA factor states for the same user. When recovery codes exist, both should return `recoveryCode: true`.

### 👎 Actual Behavior

`account.listMfaFactors()` correctly returns `recoveryCode: true`
`users.listMfaFactors(userId)` incorrectly returns `recoveryCode: false`

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

Linux

### 🧱 Your Environment

- Appwrite Cloud / Self-hosted
- Using Node.js SDK

### 👀 Have you spent some time to check if this issue has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

alexfreeman90:
https://discordapp.com/channels/564160730845151244/1448070337676841030

<img width="1466" height="965" alt="Image" src="https://github.com/user-attachments/assets/80c6af25-e18f-47be-8a9f-a10dff2d2c81" />

Adityakk9031:
please assign this issue to me

Adityakk9031:
please assign it to me pr is ready

YuvamGupta:
Hi Aditya! I am not a maintainer, so I can't assign this issue to you. However, please go to the [discord issue](https://discordapp.com/channels/564160730845151244/1448070337676841030) and ask to be assigned there. Hopefully, you will be assigned and your PR will be approved.

HarshMN2345:
fixed in https://github.com/appwrite/appwrite/pull/13531

PR #10951, opened 2025-12-13
Fix users.listMfaFactors() recoveryCode mismatch (#10927)

Adityakk9031:
Summary
Fixes an inconsistency where GET /v1/users/:userId/mfa/factors returned recoveryCode: false even when recovery codes existed for the user. This makes the users endpoint match the behavior of account.listMfaFactors().

Root cause
The Users controller did not check the user document for the mfaRecoveryCodes attribute. The Account controller already checks that attribute and sets recoveryCode: true when recovery codes exist; the Users controller simply omitted that check and therefore always returned recoveryCode: false.

What this change does

Adds the same recovery-code presence check used by the Account handler to app/controllers/api/users.php so Type::RECOVERY_CODE is set correctly.

Ensures GET /v1/users/:userId/mfa/factors returns recoveryCode: true when recovery codes exist for the user, matching GET /v1/account/mfa/factors.

Files changed

app/controllers/api/users.php — include mfaRecoveryCodes check and set Type::RECOVERY_CODE in the factors response.

- Fixes: #10927

### DIFF
--- a/app/controllers/api/users.php
+++ b/app/controllers/api/users.php
@@ -1852,22 +1852,28 @@ App::get('/v1/users/:userId/mfa/factors')
     ->inject('response')
     ->inject('dbForProject')
     ->action(function (string $userId, Response $response, Database $dbForProject) {
-        $user = $dbForProject->getDocument('users', $userId);
+    $user = $dbForProject->getDocument('users', $userId);
 
-        if ($user->isEmpty()) {
-            throw new Exception(Exception::USER_NOT_FOUND);
-        }
+    if ($user->isEmpty()) {
+        throw new Exception(Exception::USER_NOT_FOUND);
+    }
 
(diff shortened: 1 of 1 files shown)


Adityakk9031:
Please review this once before closing, as assigning an issue often takes months based on my previous experience.

Adityakk9031:
please merge this

Adityakk9031:
@Aglowharmony  please tell someone to merge this pr also

Aglowharmony:
Yes I will but uk sometimes it takes months and rn I don't have the authority to merge this in main> @Aglowharmony  please tell someone to merge this pr also

Aglowharmony:
@eldadfux please look into this asap

Adityakk9031:
@eldadfux please merge this

HarshMN2345:

PR #10953, opened 2025-12-14
Feat: Audits  upgrade

lohanidamodar:
## What does this PR do?

- Upgrade audits to 2.0.0

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


src/Appwrite/Platform/Workers/Audits.php
```diff
@@ -102,7 +103,7 @@ public function action(Message $message, callable $getProjectDB, Document $proje
                 'mode' => $mode,
                 'data' => $auditPayload,
             ],
-            'timestamp' => date("Y-m-d H:i:s", $message->getTimestamp()),
+            'time' => date("Y-m-d H:i:s", $message->getTimestamp()),
```
abnegate, line 105: Is this correct?

lohanidamodar: Yes there was inconsistency audit collection had attribute named time and here we were setting timestamp and audit library was mapping timestamp to time attribute

lohanidamodar: Yes there was inconsistency audit collection had attribute named time and here we were setting timestamp and audit library was mapping timestamp to time attribute

abnegate (approved):

src/Appwrite/Platform/Workers/Deletes.php
```diff
@@ -786,7 +786,7 @@ private function deleteAuditLogs(Document $project, callable $getProjectDB, stri
         $dbForProject = $getProjectDB($project);
 
         try {
-            $this->deleteByGroup(Audit::COLLECTION, [
+            $this->deleteByGroup(SQL::COLLECTION, [
```
eldadfux: Why do we need this constant? can we abstract it behind the delete method? we already Iintialize the adapter so we should be able to know this internally.

app/controllers/api/projects.php
```diff
@@ -247,13 +248,15 @@
             }
 
             if ($create || $projectTables) {
-                $audit = new Audit($dbForProject);
```
eldadfux, line 250: Can we remove all audit related stuff from the project creation once we complete the switch?

app/worker.php
```diff
@@ -450,6 +452,19 @@
 
 Server::setResource('executor', fn () => new Executor());
 
+Server::setResource('getAudit', function (Database $dbForPlatform, callable $getProjectDB) {
```
eldadfux, line 455: For future we need to avoid using DI to return callbacks, this can be done by standardizing the way we pass project / user to message jobs which will allow us to auto create needed dependencies without needing to pass params to every dependency.

app/controllers/api/projects.php
lohanidamodar: Yes, we can.

src/Appwrite/Platform/Workers/Deletes.php
lohanidamodar: updated.

app/worker.php
lohanidamodar: Yes thi will need a proper refactor.

PR #10962, opened 2025-12-16
Improve rule verification flow & add logs

hmacr:
Related SER-721

Along the way, refactored the code to use common constants & action for rule verification.

### DIFF
--- a/src/Appwrite/Platform/Modules/Proxy/Http/Rules/Function/Create.php
+++ b/src/Appwrite/Platform/Modules/Proxy/Http/Rules/Function/Create.php
@@ -91,18 +91,15 @@ class Create extends Action
 
         // TODO: (@Meldiron) Remove after 1.7.x migration
         $ruleId = System::getEnv('_APP_RULES_FORMAT') === 'md5' ? md5($domain) : ID::unique();
-
+        $status = RULE_STATUS_CREATED;
         $owner = '';
+
         if (
             ($functionsDomain != '' && \str_ends_with($domain, $functionsDomain)) ||
             ($sitesDomain != '' && \str_ends_with($domain, $sitesDomain))
         ) {
-            $owner = 'Appwrite';
-        }
-
-        $status = RULE_STATUS_CREATED;
-        if (($functionsDomain != '' && \str_ends_with($domain, $functionsDomain)) || ($sitesDomain != '' && \str_ends_with($domain, $sitesDomain))) {
             $status = RULE_STATUS_VERIFIED;
+            $owner = 'Appwrite';
         }
 
         $rule = new Document([
--- a/tests/e2e/Services/Proxy/ProxyCustomServerTest.php
+++ b/tests/e2e/Services/Proxy/ProxyCustomServerTest.php
@@ -86,7 +86,7 @@ class ProxyCustomServerTest extends Scope
         $domain = \uniqid() . '.com';
         $rule = $this->createAPIRule($domain);
         $this->assertEquals(201, $rule['headers']['status-code']);
-        $this->assertEquals(RULE_STATUS_CREATED, $rule['body']['status']);
+        $this->assertEquals('created', $rule['body']['status']);
     }
 
     public function testCreateRuleVcs(): void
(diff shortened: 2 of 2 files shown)


src/Appwrite/Platform/Modules/Proxy/Http/Rules/Function/Create.php
```diff
@@ -134,49 +90,26 @@ public function action(string $domain, string $functionId, string $branch, Respo
...
-            }
-        }
+        $ruleId = System::getEnv('_APP_RULES_FORMAT') === 'md5' ? md5($domain) : ID::unique();
 
         $owner = '';
         if (
-            ($functionsDomain != '' && \str_ends_with($domain->get(), $functionsDomain)) ||
-            ($sitesDomain != '' && \str_ends_with($domain->get(), $sitesDomain))
+            ($functionsDomain != '' && \str_ends_with($domain, $functionsDomain)) ||
+            ($sitesDomain != '' && \str_ends_with($domain, $sitesDomain))
```
Meldiron, line 99: Let's open all create rules files side by site (api, function, redirect, site), and compare them. Ideally their logic is identical; I think this is not same as we havein API createRule. Let's compare entire file, only keeping differences that are on purpose (having comment for those differences would be great too)

src/Appwrite/Platform/Modules/Proxy/Http/Rules/Verification/Update.php
```diff
@@ -82,94 +79,46 @@ public function action(
...
                 'domainType' => $rule->getAttribute('deploymentResourceType', $rule->getAttribute('type')),
             ]))
             ->trigger();
 
-        $queueForEvents->setParam('ruleId', $rule->getId());
-
-        $certificate = $dbForPlatform->getDocument('certificates', $rule->getAttribute('certificateId', ''));
-        $rule->setAttribute('logs', $certificate->getAttribute('logs', ''));
+        if (!empty($certificate)) {
+            $rule->setAttribute('renewAt', $certificate->getAttribute('renewDate', ''));
```
Meldiron, line 120: We are missing updateDocument here - this is never saved.

tests/e2e/Services/Proxy/ProxyCustomServerTest.php
```diff
@@ -56,7 +86,7 @@ public function testCreateRuleApex(): void
         $domain = \uniqid() . '.com';
         $rule = $this->createAPIRule($domain);
         $this->assertEquals(201, $rule['headers']['status-code']);
-        $this->assertEquals('created', $rule['body']['status']);
+        $this->assertEquals(RULE_STATUS_CREATED, $rule['body']['status']);
```
Meldiron: In test file it might be better to use string instead of const var, like we did before.

Someone might change value of const and all tests would still pass, when in reality, I would prefer them to fail - to serve as indication we need backwards comaptibility.

src/Appwrite/Platform/Modules/Proxy/Http/Rules/Verification/Update.php
hmacr: I don't think rule has `renewAt` attribute on DB.
This is more of a mapping that we do for the API response (like we do in GET and XList APIs)

src/Appwrite/Platform/Modules/Proxy/Http/Rules/Function/Create.php
hmacr: Nice find!
Functionally they're the same, but I missed simplifying the code path here like I did for other create APIs.

src/Appwrite/Platform/Modules/Proxy/Http/Rules/Verification/Update.php
Meldiron: Ah, I see, its also in GET and LIST.

But I dont see it in POST (when creating rules). Should we add it there then? 🤔

hmacr: It's highly improbable for certificate generation to succeed when creating the rules since we do it async.
There won't be a `renewDate` to map from 😄

Meldiron (approved):

PR #10975, opened 2025-12-17
sync sdk example changes + improvements

ChiragAgg5k:


src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Bulk/Update.php
```diff
@@ -74,7 +74,7 @@ public function __construct()
             ))
             ->param('databaseId', '', new UID(), 'Database ID.')
             ->param('collectionId', '', new UID(), 'Collection ID.')
-            ->param('data', [], new JSON(), 'Document data as JSON object. Include only attribute and value pairs to be updated.', true)
+            ->param('data', [], new JSON(), 'Document data as JSON object. Include only attribute and value pairs to be updated.', true, example: '{"username":"walter.obrien","email":"walter.obrien@example.com","fullName":"Walter O\'Brien","age":33,"isAdmin":false}')
```
ItzNotABug, line 77: maybe we should have a const for example at the module/sub-module level. Pretty much all are alike.

ChiragAgg5k: yeah can be separate refactor, but there is subtle difference in age 😉 (you know showing difference in create vs update)

this is a small improvement, in future PR i want to set examples for request params across all methods. will make this refactor part of that PR

ItzNotABug (approved):

PR #10981, opened 2025-12-17
fix: lookup identity by email in OAuth2 login flow

stnguyen90:
## What does this PR do?

Previously, we were only checking for existing identities by provider user ID during OAuth2 login. This could lead to an error if someone has linked a different email address to their account already, but log in using a different OAuth2 provider that shares the same email.

We initially required the user to connect the OAuth2 provider before using it to prevent a bad actor from hijacking an account by creating an account on the OAuth2 provider with the same email. However, this created a poor user experience. In addition, this doesn't stop the same thing from happening if a bad actor creates an account in the other OAuth2 provider with the same email as the Appwrite account email.

## Test Plan

1. Create an account like joe@example.com
2. Link account to google address joe@google.com
3. Log in using github address joe@google.com

## Related PRs and Issues

None

## Checklist

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -1639,13 +1639,22 @@ App::get('/v1/account/sessions/oauth2/:provider/redirect')
              */
             $isVerified = $oauth2->isEmailVerified($accessToken);
 
+            // Look for existing identity with provider user ID or email address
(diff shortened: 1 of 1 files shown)


stnguyen90:
Closing in favor of https://github.com/appwrite/appwrite/pull/10986

PR #10985, opened 2025-12-18
Support for sync-ing certificate generation status

hmacr:
- Added a periodic background task for domain verification & certificate renewal.
- Provides base for sync-in certificate generation status on Cloud.

Related SER-331

### DIFF
--- a/src/Appwrite/Platform/Tasks/Interval.php
+++ b/src/Appwrite/Platform/Tasks/Interval.php
@@ -42,7 +42,7 @@ class Interval extends Action
         });
     }
 
-    private function verifyDomain(Database $dbForPlatform, Certificate $queueForCertificate): void
+    private function verifyDomain(Database $dbForPlatform, Certificate $queueForCertificates): void
     {
         $time = DatabaseDateTime::now();
         $fromTime = new DateTime('-3 days'); // Max 3 days old
@@ -63,7 +63,7 @@ class Interval extends Action
         Console::info("[{$time}] Found " . \count($rules) . " rules for domain verification, scheduling jobs.");
 
         foreach ($rules as $rule) {
-            $queueForCertificate
+            $queueForCertificates
                 ->setDomain(new Document([
                     'domain' => $rule->getAttribute('domain'),
                     'domainType' => $rule->getAttribute('deploymentResourceType', $rule->getAttribute('type')),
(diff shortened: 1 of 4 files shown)


src/Appwrite/Platform/Tasks/Interval.php
Meldiron, line 1: I think we should put this into Interval worker instead

Meldiron: If I am not mistaken, that will prevent need for a new container

src/Appwrite/Platform/Workers/Certificates.php
Meldiron, line 1: big diff in this file worries me a little. is there anything we can do to simplfy it? Or did we refactor something intentionally?

hmacr: Existing code only handled "certificate generation" flow (moved to its own method).
We've added a new "rule verification" flow for records that are in "created" state (needing DNS record verification). This will be triggered from the periodic background task.

src/Appwrite/Platform/Tasks/Interval.php
hmacr: Do we have `Interval` worker? 🤔 

I moved the certificate renewal task from `Maintenance` to this new task so that
1. CE can have rule verification.
2. Cloud can have certificate status sync.

Rule verification & Certificate status sync run at different frequencies, so it couldn't be part of `Maintenance`. Also, makes for a clean separation of "rules" related maintenance here.

src/Appwrite/Platform/Tasks/Interval.php
```diff
@@ -0,0 +1,126 @@
...
+          ->callback($this->action(...));
+    }
+
+    public function action(Database $dbForPlatform, Certificate $queueForCertificates): void
+    {
+        Console::title('Interval V1');
+        Console::success(APP_NAME . ' interval process v1 has started');
+
+        $intervalDomainVerification = (int) System::getEnv('_APP_INTERVAL_DOMAIN_VERIFICATION', '60'); // 1 minute
+        $intervalCertificateRenewal = (int) System::getEnv('_APP_INTERVAL_CERTIFICATE_RENEWAL', '86400'); // 1 day
```
Meldiron: I would keep renewal in Maintenance, it matches the task name better.

Later we can add certificate status sync into Interval.

Meldiron (approved):

PR #10988, opened 2025-12-18
Feat: CE support for Platform API

Meldiron:
## What does this PR do?

Makes Key extendable

## Test Plan

Current tests

## Related PRs and Issues

- (Related PR or issue)

## Checklist


src/Appwrite/Auth/Key.php
```diff
@@ -95,15 +98,13 @@ public function isProjectCheckDisabled(): bool
...
-     * @param string $key
-     * @return Key
      * @throws Exception
+     * @return T
      */
     public static function decode(
         Document $project,
         string $key
-    ): Key {
+    ): static {
```
TorstenDittmann: i dont think you need a return type here then 👍🏻

app/config/collections/platform.php
```diff
@@ -622,27 +622,38 @@
...
                 'size' => Database::LENGTH_KEY,
                 'signed' => true,
                 'required' => false,
-                'default' => 0,
+                'default' => null,
                 'array' => false,
                 'filters' => [],
             ],
+            [
+                '$id' => ID::custom('resourceType'), // project, team, user
```
Meldiron:
```suggestion
                '$id' => ID::custom('resourceType'), // projects, teams, users
```

src/Appwrite/Platform/Workers/Deletes.php
```diff
@@ -565,7 +565,8 @@ protected function deleteProject(Database $dbForPlatform, callable $getProjectDB
 
         // Delete Keys
         $this->deleteByGroup('keys', [
-            Query::equal('projectInternalId', [$projectInternalId]),
+            Query::equal('resourceInternalId', [$projectInternalId]),
+            Query::equal('resourceType', ['projects']),
```
Meldiron: TODO: Delete worker for organization and account keys

tests/unit/Auth/KeyTest.php
Meldiron, line 1: Improve unit tests with all new changes

app/controllers/shared/api.php
```diff
@@ -301,10 +301,6 @@
 
         // Step 5: API Key Authentication
         if (!empty($apiKey)) {
-            // Verify no user session exists simultaneously
-            if (!$user->isEmpty()) {
-                throw new Exception(Exception::USER_API_KEY_AND_SESSION_SET);
-            }
```
loks0n, line 163: Is this intentional? What happens now with session + key?

app/config/collections/platform.php
```diff
@@ -712,9 +723,9 @@
         ],
         'indexes' => [
             [
-                '$id' => ID::custom('_key_project'),
+                '$id' => ID::custom('_key_resource'),
                 'type' => Database::INDEX_KEY,
-                'attributes' => ['projectInternalId'],
+                'attributes' => ['resourceType', 'resourceInternalId'],
                 'lengths' => [Database::LENGTH_KEY],
```
loks0n: Do we need another Database::LENGTH_KEY here?

app/controllers/api/teams.php
```diff
@@ -431,15 +431,24 @@
             throw new Exception(Exception::GENERAL_SERVER_ERROR, 'Failed to remove team from DB');
         }
 
+        $clone = clone $team;
```
loks0n: Why do we clone the team? Is it avoidable?

src/Appwrite/Auth/Key.php
```diff
@@ -188,12 +204,86 @@ public static function decode(
...
+
+                $expire = $key->getAttribute('expire');
+                $expired = false;
+                if (!empty($expire) && $expire < DateTime::formatTz(DateTime::now())) {
+                    $expired = true;
+                }
+
+                $name = $key->getAttribute('name', 'UNKNOWN');
+
+                $role = User::ROLE_USERS;
```
loks0n, line 235: Is this correct? Aren't keys for ROLE_APPS?

app/config/collections.php
```diff
@@ -26,8 +26,8 @@
 $collections = [
     'buckets' => $buckets,
     'databases' => $databases,
-    'projects' => array_merge($projects, $common),
-    'console' => array_merge($platform, $common),
+    'projects' => array_merge_recursive($projects, $common),
```
eldadfux, line 29: Why did we had to change this? curious to know what issue we faced.

app/controllers/shared/api.php
eldadfux: Adding to the question

app/controllers/shared/api.php
```diff
@@ -360,15 +381,17 @@
...
                     if (!in_array($sdk, $sdks)) {
                         $sdks[] = $sdk;
-                        $dbKey->setAttribute('sdks', $sdks);
 
-                        /** Update access time as well */
-                        $dbKey->setAttribute('accessedAt', Datetime::now());
-                        $dbForPlatform->updateDocument('keys', $dbKey->getId(), $dbKey);
-                        $dbForPlatform->purgeCachedDocument('projects', $project->getId());
+                        $updates->setAttribute('sdks', $sdks);
+                        $updates->setAttribute('accessedAt', Datetime::now());
```
eldadfux, line 232: isn't this limiting the update to only if we have a recognized SDK? We should have the update time done once a day regardless to how the API was consumed.

app/controllers/shared/api.php
Meldiron: Now that we have account keys, we can have api key, and user too.

I think this was done for UX purpose, to not pass both cookie and api key. If that's the case, I would vote to change the if statemept to look for cookie

app/controllers/shared/api.php
Meldiron: we also update accessedAt in another place acordingly (above or below).

This is to insta-update accessedAt no matter when previous updatedAt was, when using new SDK. There are only so many SDKs, so this shouldnt causue too much slowness even if you used them all

app/config/collections.php
Meldiron: copied it from cloud; it allows me to do platform-DB overrides (new attributes) for collectiosn defined in common.php. Specifically, `keys` attribute in `teams` (organization key). Same for account keys.

Later-on those can be features for projets too, but for now Console is enough

src/Appwrite/Auth/Key.php
Meldiron: this introduces account keys, so account key will base user roles, like account cookie does

app/controllers/api/teams.php
Meldiron: Hmm. Copied from existing code. Probably to protect from being edited. but I couldnt find any place that would do it. removing clone now.

app/controllers/shared/api.php
eldadfux: but you still shouldn't pass both account key, api key, session, etc together, how do we handle this then?

app/controllers/shared/api.php
```diff
@@ -333,23 +329,48 @@
...
+                };
+
+                $updates = new Document();
+
                 $accessedAt = $dbKey->getAttribute('accessedAt', 0);
 
                 if (DateTime::formatTz(DateTime::addSeconds(new \DateTime(), -APP_KEY_ACCESS)) > $accessedAt) {
-                    $dbKey->setAttribute('accessedAt', DateTime::now());
-                    $dbForPlatform->updateDocument('keys', $dbKey->getId(), $dbKey);
-                    $dbForPlatform->purgeCachedDocument('projects', $project->getId());
```
eldadfux, line 208: why did we remove this? when updating a key, we need to purge parent object in case we have a subquery to fetch the keys that is getting cached.

app/controllers/shared/api.php
eldadfux: > no matter when previous updatedAt was
What do you mean by that? Its crucial this gets updated only once a day and not a single time more, performance implications here are big.

app/controllers/shared/api.php
```diff
@@ -360,15 +381,17 @@
...
-                        $dbForPlatform->updateDocument('keys', $dbKey->getId(), $dbKey);
-                        $dbForPlatform->purgeCachedDocument('projects', $project->getId());
+                        $updates->setAttribute('sdks', $sdks);
+                        $updates->setAttribute('accessedAt', Datetime::now());
                     }
                 }
 
+                if (!$updates->isEmpty()) {
+                    Authorization::skip(fn () => $dbForPlatform->updateDocument('keys', $dbKey->getId(), $updates));
+                    $purgeResource();
```
eldadfux: if we call this only once, why the function?

app/init/resources.php
```diff
@@ -436,6 +437,30 @@
             }
         }
     }
+
+    // Account based on account API key
+    $accountKey = $request->getHeader('x-appwrite-key', '');
+    $accountKeyUserId = $request->getHeader('x-appwrite-user', '');
```
eldadfux, line 452: What is `x-appwrite-user`?

src/Appwrite/Platform/Workers/Deletes.php
```diff
@@ -123,6 +123,9 @@ public function action(
                     case DELETE_TYPE_USERS:
                         $this->deleteUser($getProjectDB, $document, $project);
                         break;
+                    case DELETE_TYPE_TEAMS:
```
eldadfux, line 128: Was this missing all this time??

app/controllers/shared/api.php
Meldiron: I added new exceptions

app/controllers/shared/api.php
Meldiron: Logic is following:

```
if accessedAt is older than 24 hours:
  set accessedAt to today

if current SDK is not in sdks array:
  set sdks to include current SDK
  set accessedAt to today

if any updates were made:
  update document
  purge cache
```

Should be all good

app/controllers/shared/api.php
Meldiron: It was probably just moved.

We still update keys, it just now respects all 3 collections (projects, teams, accounts)

app/controllers/shared/api.php
Meldiron: Nice catch, removing function

app/init/resources.php
Meldiron: consisteny with x-appwrite-project;

user passed with user key; organization header passed with organization key

src/Appwrite/Platform/Workers/Deletes.php
Meldiron: Seems like it; till now we only delete membership with teams, and we do that sync intentionally: https://github.com/appwrite/appwrite/blob/1308706023ca12282047e260b4db4c138b6a011b/app/controllers/api/teams.php#L437

PR #11003, opened 2025-12-22
Write to new resource attributes in `keys`

hmacr:
- Write to new columns
- Read from both old & new columns

---

Related SER-731


app/controllers/api/projects.php
```diff
@@ -1587,7 +1590,13 @@
 
         $key = $dbForPlatform->findOne('keys', [
             Query::equal('$id', [$keyId]),
-            Query::equal('projectInternalId', [$project->getSequence()]),
+            Query::or([
```
Meldiron, line 1599: lets update listKeys

app/config/collections/platform.php
```diff
@@ -643,6 +643,39 @@
                 'array' => false,
                 'filters' => [],
             ],
+            [
```
Meldiron, line 646: https://github.com/appwrite/appwrite/pull/10988/changes#diff-5463d35aa4528b225e1f903df6016e9c915f5c2f7c0d3dc9f189d1d86a8662d3R1918-R1944

Meldiron (approved):

PR #11018, opened 2025-12-24
Fix #8785: Handle null name in Users and Account creation

KaparthyReddy:
### What does this PR do?
This PR resolves a 500 Server Error occurring when a `null` value is explicitly passed for the `name` parameter in user creation endpoints. I've updated the action signatures to be nullable and added a null-coalescing fallback to an empty string.

### Which issue does it fix?
Fixes #8785

### Checklist
- [x] All user creation variants (bcrypt, argon2, etc.) updated in `users.php`.
- [x] Param type-hint updated to `?string`.
- [x] Fallback `$name ?? ''` added to prevent downstream errors.

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -377,8 +377,8 @@ App::post('/v1/account')
     ->inject('dbForProject')
     ->inject('authorization')
     ->inject('hooks')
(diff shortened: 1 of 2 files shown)


HarshMN2345:

PR #11033, opened 2025-12-28
add functionsEvents and webhooksEvents

shimonewman:


src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
```diff
@@ -394,4 +413,88 @@ protected function triggerBulk(
...
         $queueForWebhooks->reset();
     }
+
+    /**
+     * Get function events for a project, using Redis cache
+     * @param Document|null $project
+     * @param Database $dbForProject
+     * @return array
+     */
+    protected function getFunctionEvents(?Document $project, Database $dbForProject): array
```
eldadfux: Does this need to happen only on the DB controller or on a more global shutdown? Don't we face same issue also on general events and not only bulk ops? For example events from the account or users service?

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
```diff
@@ -394,4 +413,88 @@ protected function triggerBulk(
...
+    {
+        if ($project === null || $project->isEmpty() || $project->getId() === 'console') {
+            return [];
+        }
+
+        $hostname = $dbForProject->getAdapter()->getHostname();
```
eldadfux: We have a new PR to add optional caching to listDocuments, would be great if we could just use instead of having custom code.

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
```diff
@@ -394,4 +413,88 @@ protected function triggerBulk(
...
+            $hostname ?? '',
+            $dbForProject->getNamespace(),
+            $dbForProject->getTenant(),
+            $project->getId()
+        );
+
+        $ttl = 3600; // 1 hour cache TTL
+        $cachedFunctionEvents = $dbForProject->getCache()->load($cacheKey, $ttl);
+
+        if ($cachedFunctionEvents !== false) {
```
eldadfux: Would be better to store the events as a hash map for quick key based access, this will make check much faster and we don't need to access specific properties or loop through the events objects.

src/Appwrite/Platform/Modules/Compute/Base.php
```diff
@@ -336,4 +336,29 @@ public function redeployVcsSite(Request $request, Document $site, Document $proj
...
         return $deployment;
     }
+
+    /**
+     * Purge function events cache for a project
+     * @param Document $project
+     * @param Database $dbForProject
+     * @return void
+     */
+    protected function purgeFunctionEventsCache(Document $project, Database $dbForProject): void
```
eldadfux: Would be really neat if we could do this via event listeners or filters. We had this idea long time ago to avoid having to remember to purge on each update which is very error prone.

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
shimonewman: Found 2
1. https://github.com/appwrite/appwrite/blob/main/src/Appwrite/Platform/Modules/Databases/Http/Databases/Transactions/Update.php#L385
2.
https://github.com/appwrite/appwrite/blob/main/src/Appwrite/Platform/Modules/Databases/Http/Databases/Transactions/Update.php#L385
https://github.com/appwrite/appwrite/blob/main/src/Appwrite/Platform/Modules/Databases/Http/Databases/Transactions/Update.php#L385

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
shimonewman: You mean turn it to associative  array before saving it?

src/Appwrite/Platform/Modules/Compute/Base.php
shimonewman: I think we can add listeners to UPDATE/DELETE operations.
But it will be invoked for all collections, and we will need to add
a collection validation condition in the purge function, isn't it an overhead 

`  if ($document->getCollection() !== 'functions') {
        return;
    }`

eldadfux: Check with Jake if we can have a collection specific listener, and use * to listen to all collections

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
eldadfux: These are all just for transactions, what about our other events?

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
eldadfux: yeh, just thinking out load. Is it possible? might not work with the approach of reusing the new list cache feature.

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
shimonewman: It looks like they all resolve here 
https://github.com/appwrite/appwrite/blob/add-webhooks-and-functions-events/app/controllers/shared/api.php#L94

and here 
https://github.com/appwrite/appwrite/blob/dc85d4464774df19eda99559d9c7af11fde96741/app/controllers/shared/api.php#L821

if i got it right

docker-compose.yml
```diff
@@ -12,7 +12,8 @@ x-logging: &x-logging
 
 services:
   traefik:
-    image: traefik:2.11
+    #image: traefik:2.11 not working with docker api version 1.52
+    image: traefik:3.6
```
abnegate: Do we need to any migration here? It the change required for this PR?

src/Appwrite/Platform/Modules/Compute/Base.php
abnegate: > I think we can add listeners to UPDATE/DELETE operations. But it will be invoked for all collections, and we will need to add a collection validation condition in the purge function, isn't it an overhead
> 
> ` if ($document->getCollection() !== 'functions') { return; }`

The overhead is only the event function call right? Since we'd immediately return. This approach will be much easier than reworking events to be collection-specific

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
abnegate: Yep only bulk ops go through their specific routes, only transactions and the bulk API, because they fire multiple events per request. I think we can still add this function to the base `Action`, even though we don't use it in other APIs now, we might in future

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
abnegate: @eldadfux I don't think we can use list caching here, that only applies in the list documents endpoint

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
```diff
@@ -394,4 +413,88 @@ protected function triggerBulk(
...
+            return \json_decode($cachedFunctionEvents, true) ?? [];
+        }
+
+        try {
+            $functions = $dbForProject->skipValidation(fn () => $dbForProject->find('functions', [
```
abnegate: Why do we skip validation?

abnegate: Can we read in a paginated loop instead?

src/Appwrite/Platform/Modules/Functions/Http/Functions/Update.php
```diff
@@ -286,6 +286,9 @@ public function action(
 
         $queueForEvents->setParam('functionId', $function->getId());
 
+        // Purge function events cache when function is updated
+        $this->purgeFunctionEventsCache($project, $dbForProject);
```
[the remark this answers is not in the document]
abnegate: @shimonewman Comment is valid

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Action.php
eldadfux: Can't we apply it as a feature of utopia/dbs?

src/Appwrite/Platform/Modules/Compute/Base.php
eldadfux: 👍

app/controllers/shared/api.php
```diff
@@ -650,7 +677,11 @@
                 $queueForFunctions->from($queueForEvents),
                 $queueForWebhooks->from($queueForEvents),
                 $queueForRealtime->from($queueForEvents)
-            ));
+            ))
+            ->on(Database::EVENT_DOCUMENT_CREATE, 'purge-function-events-cache', fn ($event, $document) => $functionsEventsCacheListener($event, $document, $project, $dbForProject))
```
eldadfux: @abnegate as I mentioned in the engineering leads channel, this approach is really bad, we should not register events on each request. At worst it should happen inside the DI when creating the resource.

app/controllers/shared/api.php
```diff
@@ -849,41 +879,6 @@
         $route = $utopia->getRoute();
         $requestParams = $route->getParamsValues();
 
-        /**
```
eldadfux, line 855: Why did we remove this? this was just added today/yesterday.

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Bulk/Delete.php
```diff
@@ -203,7 +203,8 @@ public function action(string $databaseId, string $collectionId, array $queries,
             $queueForEvents,
             $queueForRealtime,
             $queueForFunctions,
-            $queueForWebhooks
+            $queueForWebhooks,
```
eldadfux, line 208: I'm very confused about what we're doing here. This check should happen globally for all events we fire, but it seems like we're focused on solving this only on the db service scope. I understand that we have the transactions/batch operations firing event separately, but this is just part of the problem - we're firing events from account, users, storage, functions, messaging, sites etc. We need to fire events only if a function/webhook is actually listening to them.

src/Appwrite/Platform/Modules/Databases/Http/TablesDB/Tables/Rows/Bulk/Delete.php
```diff
@@ -66,7 +66,7 @@ public function __construct()
             ->inject('queueForFunctions')
             ->inject('queueForWebhooks')
             ->inject('plan')
-            ->inject('authorization')
+            ->inject('eventProcessor')
```
abnegate, line 69: I think we need both injected here

src/Appwrite/Platform/Modules/Databases/Http/TablesDB/Tables/Rows/Bulk/Update.php
```diff
@@ -68,7 +68,7 @@ public function __construct()
             ->inject('queueForFunctions')
             ->inject('queueForWebhooks')
             ->inject('plan')
-            ->inject('authorization')
+            ->inject('eventProcessor')
```
abnegate, line 71: I think we need both injected here

src/Appwrite/Platform/Modules/Databases/Http/TablesDB/Tables/Rows/Bulk/Upsert.php
```diff
@@ -68,7 +68,7 @@ public function __construct()
             ->inject('queueForFunctions')
             ->inject('queueForWebhooks')
             ->inject('plan')
-            ->inject('authorization')
+            ->inject('eventProcessor')
```
abnegate, line 71: I think we need both injected here

abnegate (approved):

PR #11069, opened 2026-01-03
Add: bucket size to get request

ItzNotABug:
## What does this PR do?

Adds support for returning bucket size in the response.

## Test Plan

Manual.

## Related PRs and Issues

N/A.

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,7 @@
         "utopia-php/locale": "0.8.*",
         "utopia-php/logger": "0.6.*",
         "utopia-php/messaging": "0.20.*",
-        "utopia-php/migration": "1.*.*",
+        "utopia-php/migration": "1.3.*",
         "utopia-php/orchestration": "0.9.*",
         "utopia-php/platform": "0.7.*",
         "utopia-php/pools": "0.8.*",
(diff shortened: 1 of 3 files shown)


src/Appwrite/Platform/Modules/Storage/Http/Buckets/Get.php
```diff
@@ -60,6 +63,45 @@ public function action(
...
+         * `$id = md5("{$time}_{$period}_{$key}");`
+         *
+         * but when $time is null it just makes the $id as md5('_inf_' . $key);
+         *
+         * Why do this though?\
+         * Using `getDocument()` below to leverage cache!
+         */
+        $statsDocId = md5('_inf_' . $metric);
+
+        $storageStats = Authorization::skip(
```
lohanidamodar, line 88: the size calculated in dbForProject is from stats usage not stats resources and which we have seen to be incorrect in many account. I'm not sure we should be doing this? Slow but we have same metric in logsDB from statsResources which counts directly from database and is correct but counting there is slow.

I don't know if we want to wait for Clickhouse implementation for usage for this.

lohanidamodar (approved):
Looks good for now, just an info in the front-end that the total here is delayed by an hour or something similar should work.

PR #11074, opened 2026-01-04
fix(6061): delete user sessions when status is updated to blocked

alokdangre:
## What does this PR do?

Updated the updateStatus endpoint. When a user's status is updated to false (blocked), the system now iterates through and deletes all active sessions associated with that user. This mirrors the behavior of the deleteSessions endpoint.

## Test Plan

## Related PRs and Issues

- #6061 

## Checklist

### DIFF
--- a/app/controllers/api/users.php
+++ b/app/controllers/api/users.php
@@ -1173,17 +1173,24 @@ App::patch('/v1/users/:userId/status')
             throw new Exception(Exception::USER_NOT_FOUND);
         }
 
-        $user = $dbForProject->updateDocument('users', $user->getId(), $user->setAttribute('status', (bool) $status));
+        $user = $dbForProject->withTransaction(function () use ($user, $status, $dbForProject) {
+            $user = $dbForProject->updateDocument('users', $user->getId(), $user->setAttribute('status', (bool) $status));
 
-        if ($status === false) {
-            $sessions = $user->getAttribute('sessions', []);
+            /**
+             * If the user is blocked (status is false), we unconditionally delete all their sessions.
+             */
+            if ($status === false) {
+                $sessions = $user->getAttribute('sessions', []);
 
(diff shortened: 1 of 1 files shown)


alokdangre:
@ChiragAgg5k PTAL

app/controllers/api/users.php
```diff
@@ -1173,7 +1173,24 @@ function createUser(Hash $hash, string $userId, ?string $email, ?string $passwor
...
+        $user = $dbForProject->withTransaction(function () use ($user, $status, $dbForProject) {
+            $user = $dbForProject->updateDocument('users', $user->getId(), $user->setAttribute('status', (bool) $status));
+
+            /**
+             * If the user is blocked (status is false), we unconditionally delete all their sessions.
+             */
+            if ($status === false) {
+                $sessions = $user->getAttribute('sessions', []);
+
+                foreach ($sessions as $session) {
```
ChiragAgg5k, line 1185: hmm technically we should have some kind of pagination here but i dont see any pattern for it in this file, perhaps after we migrate to Users module

abnegate: This is on the result of a subquery, so we have already read all the data into memory

app/controllers/api/users.php
```diff
@@ -1173,7 +1173,24 @@ function createUser(Hash $hash, string $userId, ?string $email, ?string $passwor
...
+                    $dbForProject->deleteDocument('sessions', $session->getId());
```
abnegate, line 1186: Instead of deleting in a loop, let's accumulate all the IDs, and call `deleteDocuments` once with `Query::equal('$id', $ids)`

ChiragAgg5k:
on some discussion, we realised we don't want to go with this direction. when a user is blocked, their requests are already blocked. The only thing this will change, is that if they're unblocked, they'll have to log back in

PR #11078, opened 2026-01-05
feat: sdk for md

ChiragAgg5k:


src/Appwrite/Platform/Tasks/SDKs.php
```diff
@@ -31,6 +32,27 @@
 
 class SDKs extends Action
 {
+    protected array $supportedSDKS = [
+        'web',
+        'cli',
+        'php',
+        'nodejs',
+        'deno',
```
ItzNotABug, line 40: we don't have deno support anymore iirc, right?

PR #11079, opened 2026-01-05
chore: add timeout to mail worker

ChiragAgg5k:


src/Appwrite/Platform/Workers/Mails.php
```diff
@@ -216,4 +219,39 @@ protected function getMailer(array $smtp): PHPMailer
...
+        $testMailer->SMTPAutoTLS = false;
+        $testMailer->Timeout = 3; /* 3 seconds */
+
+        try {
+            if (!$testMailer->smtpConnect()) {
+                throw new Exception('Connection is not valid.', 401);
+            }
+
+        } catch (\Throwable $error) {
+            throw new Exception(Exception::PROJECT_SMTP_CONFIG_INVALID, $error->getMessage());
```
loks0n: I don't think it makes sense to throw this error, it's a HTTP level exception right?

ChiragAgg5k: yeah, but idea was to not have to deal with sentry errors... do you want to just skip throwing any error for now?

ChiragAgg5k: replaced with normal error, with code 401

ItzNotABug (approved):

PR #11083, opened 2026-01-05
Remove dual read for `keys`

hmacr:
Related SER-731

Follows-on https://github.com/appwrite/appwrite/pull/11003

### DIFF
--- a/app/controllers/mock.php
+++ b/app/controllers/mock.php
@@ -200,6 +200,7 @@ App::post('/v1/mock/api-key-unprefixed')
                 Permission::update(Role::any()),
                 Permission::delete(Role::any()),
             ],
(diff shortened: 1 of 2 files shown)


app/controllers/mock.php
```diff
@@ -202,6 +202,9 @@
             ],
             'projectInternalId' => $project->getSequence(),
             'projectId' => $project->getId(),
+            'resourceInternalId' => $project->getSequence(),
```
Meldiron, line 206: I think we should now remove projectId and projectInternalId from this mock endpoint

hmacr: I'm afraid `projectInternalId` is a required param for `keys`
We'll need to
1. Make it optional using patch script
2. remove these writes in the backend
3. Delete columns using patch script

Meldiron: lets add 2 todos

PR #11100, opened 2026-01-07
Log queries

ItzNotABug:
## What does this PR do?

Structure should be -
```json
{
  "extra": {
    "queries": [
      { "equal": ["status"] },
      { "orderDesc": ["createdAt"] },
      { "limit": [] },
      {
        "and": [
          { "equal": ["teamId"] },
          { "or": [ { "greaterThan": ["score"] }, { "lessThan": ["score"] } ] }
        ]
      },
      { "select": ["name", "email"] }
    ]
  }
}
```

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/controllers/general.php
+++ b/app/controllers/general.php
@@ -816,94 +816,6 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
     return false;
 }
 
-function addQueriesToErrorReporting(Request $request, Log $log): void
-{
-    try {
-        $queries = $request->getParam('queries', []);
-        if (empty($queries) || !is_array($queries)) {
-            return;
-        }
-
-        // format query by removing sensitive values
-        $formatQuery = function (array $queryArray) use (&$formatQuery): ?array {
-            $method = $queryArray['method'] ?? '';
-            $values = $queryArray['values'] ?? [];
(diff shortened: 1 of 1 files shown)


app/controllers/general.php
```diff
@@ -816,6 +816,93 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
     return false;
 }
 
+function addQueriesToErrorReporting(Request $request, Log $log): void {
```
abnegate: Let's do it inline, we should never use globals, especially if we only call it from one place

app/controllers/general.php
```diff
@@ -816,6 +816,93 @@ function router(App $utopia, Database $dbForPlatform, callable $getProjectDB, Sw
...
+    try {
+        $queries = $request->getParam('queries', []);
+        if (empty($queries) || !is_array($queries)) {
+            return;
+        }
+
+        // format query by removing sensitive values
+        $formatQuery = function (array $queryArray) use (&$formatQuery): ?array {
```
abnegate: Same here, we only call this one place, let's do it inline so it's easier to read

ItzNotABug: this is used more than once, so kept it.

abnegate (approved):

PR #11102, opened 2026-01-07
Add totalSize to bucket xlist

ItzNotABug:
## What does this PR do?

Return correct sizes on XList. Was missed.

## Test Plan

E2E.

## Related PRs and Issues

* https://github.com/appwrite/appwrite/pull/11069

## Checklist

### DIFF
--- a/tests/e2e/Services/GraphQL/Base.php
+++ b/tests/e2e/Services/GraphQL/Base.php
@@ -2334,7 +2334,8 @@ trait Base
                         buckets {
                             _id
                             name
-                            enabled
+                            enabled,
+                            totalSize
                         }
                     }
                 }';


src/Appwrite/Platform/Modules/Storage/Http/Buckets/XList.php
```diff
@@ -109,9 +114,59 @@ public function action(
...
         $response->dynamic(new Document([
             'buckets' => $buckets,
             'total' => $total,
         ]), Response::MODEL_BUCKET_LIST);
     }
+
+    /**
+     * Adds the latest aggregated bucket storage sizes from logs DB stats.
+     */
+    private function addBucketStorageSizes(Database $dbForLogs, array $buckets): void
```
ChiragAgg5k: nit: can we avoid a helper function here? we generally avoid helper function in api endpoints

PR #11106, opened 2026-01-08
Backend implementation for per-project permissions

hmacr:
Closes SER-539

### DIFF
--- a/app/controllers/shared/api.php
+++ b/app/controllers/shared/api.php
@@ -389,23 +389,28 @@ App::init()
             }
 
             $scopes = []; // Reset scope if admin
+            $projectRoles = \array_filter($adminRoles, fn ($role) => str_starts_with($role, Roles::ROLE_PROJECT));
 
-            $hasProjectSpecificPermissions = false;
-            foreach ($adminRoles as $adminRole) {
-                $adminRole = Role::parse($adminRole);
-                if ($adminRole->getRole() === Roles::ROLE_PROJECT) {
-                    $hasProjectSpecificPermissions = true;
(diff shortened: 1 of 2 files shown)


app/config/roles.php
```diff
@@ -130,6 +130,10 @@
         'label' => 'Owner',
         'scopes' => \array_merge($member, $admins),
     ],
+    User::ROLE_MEMBER => [
```
Meldiron, line 133: - Lets see if we could use ROLE_USERS; if it even make sense
- Lets be consistent with teams if possible

app/controllers/api/projects.php
```diff
@@ -172,14 +172,25 @@
...
+                    Permission::update(Role::team($teamIdentifierForRole, 'developer')),
+                    Permission::delete(Role::team($teamIdentifierForRole, 'owner')),
+                    Permission::delete(Role::team($teamIdentifierForRole, 'developer')),
+                    // Project-specific permissions
+                    Permission::read(Role::project($projectIdentifierForRole, 'owner')),
+                    Permission::read(Role::project($projectIdentifierForRole, 'developer')),
+                    Permission::read(Role::project($projectIdentifierForRole, 'editor')),
+                    Permission::read(Role::project($projectIdentifierForRole, 'analyst')),
+                    Permission::update(Role::project($projectIdentifierForRole, 'owner')),
+                    Permission::delete(Role::project($projectIdentifierForRole, 'owner')),
```
Meldiron, line 193: Let's use same approavh for projets as we did for teams - base read, extra writes

PR #11115, opened 2026-01-11
Delete subscribers decrement error

fogelito:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Deletes/Targets.php
+++ b/src/Appwrite/Deletes/Targets.php
@@ -53,7 +53,7 @@ class Targets
                             min: 0
                         );
                     } catch (LimitException $e) {
-                        Console::error("delete subscribers limit reached (topicId={$topicId}): {$e->getMessage()}");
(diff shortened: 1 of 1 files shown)


src/Appwrite/Deletes/Targets.php
```diff
@@ -42,12 +44,17 @@ function (Document $subscriber) use ($database, $target) {
                         MESSAGE_TYPE_PUSH => 'pushTotal',
                         default => throw new Exception('Invalid target provider type'),
                     };
-                    $database->decreaseDocumentAttribute(
-                        'topics',
-                        $topicId,
-                        $totalAttribute,
-                        min: 0
```
fogelito, line 49: Do we want to allow pass null, to allow any min value?

abnegate: If you don't pass min at all, it won't be enforced. But in this case we should keep min 0, it doesn't make for the total subscriber count to go below 0

abnegate (approved):

PR #11116, opened 2026-01-11
fix: reject empty/whitespace-only platform name and key

theonlypal:
This PR fixes platform creation and update accepting whitespace-only values for `name` and `key` fields.

**What changed:**
- Added `trim()` to `name` and `key` parameters in both `createPlatform` and `updatePlatform` endpoints
- Added validation to reject empty `name` after trimming

**Why it failed before:**
- The `Text` validator checks minimum length but counts whitespace characters
- `"   "` (three spaces) passed validation because `mb_strlen("   ") >= 1`

**Why this fix works:**
- `trim()` removes leading/trailing whitespace before storage
- Empty check after trim rejects whitespace-only names
- Key is trimmed but not required (optional param), so empty after trim is allowed

**Risk assessment:**
- Low risk. Only affects platform create/update endpoints
- No breaking changes to API contract - just stricter validation
- Existing platforms with whitespace-only names won't be affected until updated

---

To make this production-safe, the remaining work is:
- Add unit tests for the trim + empty check behavior
- Audit similar input fields across other endpoints (webhooks, keys, etc.)
- Consider creating a custom `TrimmedText` validator for consistent reuse

This is ~2-3 days.
I can own this as a short paid engagement; otherwise this PR stands on its own.

### DIFF
--- a/app/controllers/api/projects.php
+++ b/app/controllers/api/projects.php
@@ -1795,6 +1795,13 @@ App::post('/v1/projects/:projectId/platforms')
             throw new Exception(Exception::PROJECT_NOT_FOUND);
         }
 
+        $name = \trim($name);
+        if (empty($name)) {
+            throw new Exception(Exception::GENERAL_ARGUMENT_INVALID, 'Platform name cannot be empty or whitespace only');
+        }
+
+        $key = \trim($key);
+
         $platform = new Document([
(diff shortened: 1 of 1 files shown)


eldadfux:
Not sure if we really care about this, but even if we did the way to go about this is with a validator level implementation and not specific inline checks.

PR #11135, opened 2026-01-14
Feat: Add more file params

Meldiron:
## What does this PR do?

Adds encryption and compression as params (can only be read, cannot query by them)

## Test Plan

New tests

## Related PRs and Issues

x

## Checklist


Meldiron:
Bypassed rule; new change was only comment change

PR #11188, opened 2026-01-25
Implement project-specific permissions

hmacr:
Closes SER-539
Closes SER-543

### DIFF
--- a/app/controllers/shared/api.php
+++ b/app/controllers/shared/api.php
@@ -244,6 +244,8 @@ App::init()
                 throw new Exception(Exception::USER_UNAUTHORIZED);
             }
 
+            $scopes = []; // Reset scope if admin
+
             $teamWideRoles = \array_filter($adminRoles, fn ($role) => !str_starts_with($role, "project-"));
             foreach ($teamWideRoles as $teamRole) {
                 $scopes = \array_merge($scopes, $roles[$teamRole]['scopes']);


app/controllers/api/projects.php
```diff
@@ -428,11 +435,18 @@
         }
 
         $permissions = [
+            // Team-wide permissions
             Permission::read(Role::team(ID::custom($teamId))),
```
Meldiron, line 439: Lets check once with Eldad if this is expected behaviour; I believe this makes it so every member can automatically see and read all proejcts.

hmacr: This was the problem we discussed last week about populating base `team:<teamId>` role for all users and how `team:<teamId>/member` could solve it.

Instead, what this PR does is, it populates `team:<teamId>` conditionally (only for non-projects API) so the user can still access the team but not take view all projects [here](https://github.com/appwrite/appwrite/pull/11188/changes#diff-f59980d143b00dac8002d8bf0c5a70ea298ac204433239d2b970877df3b1c77bR73-R98)

app/controllers/api/teams.php
```diff
@@ -1094,7 +1094,7 @@
             $roles = array_filter($roles, function ($role) {
                 return !in_array($role, [User::ROLE_APPS, User::ROLE_GUESTS, User::ROLE_USERS]);
             });
-            return new ArrayList(new WhiteList($roles), APP_LIMIT_ARRAY_PARAMS_SIZE);
+            return new ArrayList(new RoleValidator($roles), APP_LIMIT_ARRAY_PARAMS_SIZE);
```
Meldiron, line 1097: No special condition for Console?

PR #11204, opened 2026-01-28
Implement project-specific permissions

hmacr:


src/Appwrite/Platform/Modules/Functions/Http/Variables/Create.php
```diff
@@ -91,13 +89,7 @@ public function action(
...
         $variable = new Document([
             '$id' => $variableId,
-            '$permissions' => [
-                Permission::read(Role::team(ID::custom($teamId))),
-                Permission::update(Role::team(ID::custom($teamId), 'owner')),
-                Permission::update(Role::team(ID::custom($teamId), 'developer')),
-                Permission::delete(Role::team(ID::custom($teamId), 'owner')),
-                Permission::delete(Role::team(ID::custom($teamId), 'developer')),
-            ],
+            '$permissions' => $this->getPermissions($teamId, $project->getId()),
```
Meldiron, line 92: Since we use it in many places now, more generic `getRoles` might make sense, maybe multiple for different levels, `getAdminRoles`, `getReadOnlyRoles`..

PR #11209, opened 2026-01-29
Populate project-specific roles

hmacr:
Towards SER-539

Should go after https://github.com/appwrite/appwrite/pull/11208

### DIFF
--- a/app/controllers/shared/api.php
+++ b/app/controllers/shared/api.php
@@ -313,7 +313,8 @@ Http::init()
                         true => substr($adminRole, strrpos($adminRole, '-') + 1),
                         false => $adminRole,
                     };
-                    $scopes = \array_merge($scopes, $roles[$role]['scopes']);
+                    $roleScopes = $roles[$role]['scopes'] ?? [];
+                    $scopes = \array_merge($scopes, $roleScopes);
                     $authorization->addRole($role);
                 }
             }


app/controllers/shared/api.php
```diff
@@ -265,10 +265,23 @@
 
             $scopes = []; // Reset scope if admin
             foreach ($adminRoles as $role) {
```
Meldiron: Here is example of permissions set on project document:

```php
            // Team-wide permissions
            Permission::read(Role::team(ID::custom($teamId), 'owner')),
            Permission::read(Role::team(ID::custom($teamId), 'developer')),
            Permission::update(Role::team(ID::custom($teamId), 'owner')),
            Permission::update(Role::team(ID::custom($teamId), 'developer')),
            Permission::delete(Role::team(ID::custom($teamId), 'owner')),
            Permission::delete(Role::team(ID::custom($teamId), 'developer')),
            // Project-wide permissions
            Permission::read(Role::team(ID::custom($teamId), "project-{$projectId}")),
            Permission::update(Role::team(ID::custom($teamId), "project-{$projectId}-owner")),
            Permission::update(Role::team(ID::custom($teamId), "project-{$projectId}-developer")),
            Permission::delete(Role::team(ID::custom($teamId), "project-{$projectId}-owner")),
            Permission::delete(Role::team(ID::custom($teamId), "project-{$projectId}-developer")),
```

Meldiron: From this PR: https://github.com/appwrite/appwrite/pull/11208

app/controllers/shared/api.php
```diff
@@ -265,10 +265,23 @@
...
+                if (str_starts_with($role, 'project-')) {
+                    $role = substr($role, strrpos($role, '-') + 1);
```
loks0n: How do we handle edge cases with project and roles including `-`? Should we use a delimeter not possible in project id?

app/controllers/shared/api.php
```diff
@@ -278,6 +291,21 @@
...
 
+        /**
+         * We disable authorization checks above to ensure other endpoints (list teams, members, etc.) will continue working.
+         * But, for actions on resources (sites, functions, etc.) in a non-console project, we explicitly check
+         * whether the admin user has necessary permission on the project (sites, functions, etc. don't have permissions associated to them).
+         */
+        if ($project->getId() !== 'console' && $mode === APP_MODE_ADMIN) {
+            $input = new Input(Database::PERMISSION_READ, $project->getPermissionsByType(Database::PERMISSION_READ));
+            $initialStatus = $authorization->getStatus();
+            $authorization->enable();
```
loks0n, line 349: This seems quite fragile to me - why do we need to enable it and then immediately disable?

src/Appwrite/Platform/Modules/Compute/Base.php
```diff
@@ -32,11 +32,20 @@ class Base extends Action
     protected function getPermissions(string $teamId, string $projectId): array
```
loks0n: This is similar to `Projects/Action.php` should we extract to a shared trait?

app/controllers/shared/api.php
```diff
@@ -278,6 +291,21 @@
             $authorization->addRole($authRole);
         }
 
+        /**
+         * We disable authorization checks above to ensure other endpoints (list teams, members, etc.) will continue working.
+         * But, for actions on resources (sites, functions, etc.) in a non-console project, we explicitly check
+         * whether the admin user has necessary permission on the project (sites, functions, etc. don't have permissions associated to them).
+         */
+        if ($project->getId() !== 'console' && $mode === APP_MODE_ADMIN) {
+            $input = new Input(Database::PERMISSION_READ, $project->getPermissionsByType(Database::PERMISSION_READ));
```
loks0n, line 347: Where do check authorization for WRITE/DELETE?
Should we centralise it here?

tests/unit/Utopia/Database/Documents/UserTest.php
```diff
@@ -307,11 +307,11 @@ public function testPrivilegedUserRoles(): void
         ]);
         $roles = $user->getRoles($this->getAuthorization());
 
-        $this->assertCount(7, $roles);
-        $this->assertNotContains(Role::users()->toString(), $roles);
-        $this->assertNotContains(Role::user(ID::custom('123'))->toString(), $roles);
-        $this->assertNotContains(Role::users(Roles::DIMENSION_VERIFIED)->toString(), $roles);
-        $this->assertNotContains(Role::user(ID::custom('123'), Roles::DIMENSION_VERIFIED)->toString(), $roles);
+        $this->assertCount(11, $roles);
```
loks0n, line 310: This feels like the biggest change, can we do some checks to ensure everywhere that verifies Role::users() is intended to allow admin access?

app/controllers/shared/api.php
hmacr: We take the part after the last hyphen as role. That would mean
- `project-proj123-owner` -> role=owner
- `project-proj-123-owner` -> role=owner

app/controllers/shared/api.php
hmacr: For resources in non-console projects, admin users don't have auth checks by default. We could enable it, but it has its own sets of challenges because current code everywhere assumes admin mode means "no auth checks".

To make it work for project-specific permissions, we ensure they actually have access to the associated project (essentially DB lib's responsibility).

tests/unit/Utopia/Database/Documents/UserTest.php
hmacr: Any reason why privileged users weren't being populated those user roles before?

To me, users role is much more permissive than owner. Anyone who is an owner could also have users role 🤔

app/controllers/shared/api.php
hmacr: Just added the update/delete permission checks 👍 We don't need it for CE, but for Cloud we do.

app/controllers/shared/api.php
loks0n: > We could enable it, but it has its own sets of challenges because current code everywhere assumes admin mode means "no auth checks".

This reads like technical debt, do we have a plan/ticket to address it?

tests/unit/Utopia/Database/Documents/UserTest.php
loks0n: >  Any reason why privileged users weren't being populated those user roles before?

You probably have better context than me

Meldiron (approved):

Meldiron (approved):

Meldiron (approved):

PR #11210, opened 2026-01-30
Fix event caching

abnegate:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/tests/e2e/Services/Functions/FunctionsCustomClientTest.php
+++ b/tests/e2e/Services/Functions/FunctionsCustomClientTest.php
@@ -7,6 +7,7 @@ use Tests\E2E\Scopes\ProjectCustom;
 use Tests\E2E\Scopes\Scope;
 use Tests\E2E\Scopes\SideClient;
 use Utopia\Database\Helpers\ID;
+use Utopia\Database\Helpers\Permission;
 use Utopia\Database\Helpers\Role;
 use Utopia\System\System;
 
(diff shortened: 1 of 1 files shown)


src/Appwrite/Functions/EventProcessor.php
```diff
@@ -39,38 +39,34 @@ public function getFunctionsEvents(?Document $project, Database $dbForProject):
...
 
-            $uniqueEvents = \array_flip(\array_unique($events));
-            $dbForProject->getCache()->save($cacheKey, \json_encode($uniqueEvents));
+        $uniqueEvents = \array_flip(\array_unique($events));
+        $dbForProject->getCache()->save($cacheKey, \json_encode($uniqueEvents));
 
-            return $uniqueEvents;
-        } catch (\Throwable $e) {
-            return [];
-        }
```
abnegate, line 73: If we get an error, we don't want to return empty array, that will skip events. Instead we should track them

src/Appwrite/Functions/EventProcessor.php
```diff
@@ -39,38 +39,34 @@ public function getFunctionsEvents(?Document $project, Database $dbForProject):
...
-                    if (!empty($functionEvents)) {
-                        $events = array_merge($events, $functionEvents);
-                    }
+        $events = [];
+        $limit = 100;
+        $sum = 100;
+        $offset = 0;
+
+        while ($sum >= $limit) {
+            $functions = $dbForProject->getAuthorization()->skip(fn () => $dbForProject->find('functions', [
```
abnegate, line 48: This is the fix. Without auth skip, if cache is empty, and the next request comes from client-side, the find call will get no documents, and will cache 0 events.

PR #11215, opened 2026-02-01
fix: handle null name parameter in user creation endpoints

njg7194:
## What does this PR do?

Fixes #8785

### Problem
When a client explicitly sends `name: null` in the request body to any of the user creation endpoints (POST /v1/users, /v1/users/bcrypt, /v1/users/md5, etc.), the API returns a vague server error instead of handling it gracefully.

### Solution
The `name` parameter now uses the `Nullable` validator wrapper (consistent with `email` and `phone` parameters that already handle null values):

1. Changed the param validator from `new Text(128)` to `new Nullable(new Text(128))`
2. Changed the action function parameter type from `string $name` to `?string $name`
3. Used null coalescing operator (`$name ?? ''`) when passing to `createUser()` function

### Affected Endpoints
- POST /v1/users
- POST /v1/users/bcrypt
- POST /v1/users/md5
- POST /v1/users/argon2
- POST /v1/users/sha
- POST /v1/users/phpass
- POST /v1/users/scrypt
- POST /v1/users/scrypt-modified

### Testing
- Sending `name: null` now creates user with empty name (same as not sending name at all)
- Sending `name: "John"` works as before
- Not sending name parameter works as before

### Related Issue
Fixes #8785


HarshMN2345:

PR #11219, opened 2026-02-02
Add `skipRule` support for certificate generation without DB rule

hmacr:
Resolves SER-1092

This allows certificate generation to proceed for domains that don't have a corresponding rule in the database (e.g., _APP_DOMAIN). When `skipRule` is true, the worker looks up certificates by domain name instead of rule's `certificateId`, and skips all rule status updates.


src/Appwrite/Platform/Workers/Certificates.php
```diff
@@ -295,9 +309,11 @@ private function handleCertificateGenerationAction(
             $certificate->setAttribute('logs', "\033[90m[{$date}] \033[97mCertificate generation started. \033[0m\n");
 
             // Persist ASAP so that logs are reset in retry flow and user can see the latest logs on Console.
-            $certificate = $this->upsertCertificate($rule, $certificate, $dbForPlatform);
+            $certificate = $this->upsertCertificate($rule, $certificate, $dbForPlatform, $skipRule, $domain->get());
```
Meldiron: we pass $rule and $skipRule both. feels weird.

Maybe function will know to skip rule update when $rule is empty doucment instead?

src/Appwrite/Platform/Workers/Certificates.php
```diff
@@ -295,9 +309,11 @@ private function handleCertificateGenerationAction(
...
             // Ensure certificate is associated with the rule
-            $rule->setAttribute('certificateId', $certificate->getId());
+            if (!$skipRule) {
+                $rule->setAttribute('certificateId', $certificate->getId());
+            }
```
Meldiron: Reads weirdly. if not skipping feature, do the feature. Feels like opposite of what it should read - if feature enabled,do the feature

hmacr: It's kinda double negation here.. we have to take action on rule document when skip-rule is not set.
Maybe we can do the flipping before it's passed to the function, that should clear it up.

src/Appwrite/Platform/Workers/Certificates.php
```diff
@@ -104,6 +104,7 @@ public function action(
         $domain   = new Domain($document->getAttribute('domain', ''));
         $domainType = $document->getAttribute('domainType');
         $skipRenewCheck = $payload['skipRenewCheck'] ?? false;
+        $checkRule = !$payload['skipRule'] ?? true;
```
Meldiron: I would keep same naming  in worker variable, and job payload (event property)

src/Appwrite/Platform/Tasks/SSL.php
```diff
@@ -23,13 +23,15 @@ public function __construct()
             ->desc('Validate server certificates')
             ->param('domain', System::getEnv('_APP_DOMAIN', ''), new Hostname(), 'Domain to generate certificate for. If empty, main domain will be used.', true)
             ->param('skip-check', true, new Boolean(true), 'If DNS and renew check should be skipped. Defaults to true, and when true, all jobs will result in certificate generation attempt.', true)
+            ->param('check-rule', true, new Boolean(true), 'If rule lookup and updates should be checked. Defaults to true. Set to false for domains without a rule in DB (e.g., _APP_DOMAIN).', true)
```
Meldiron:
```suggestion
            ->param('check-rule', false, new Boolean(true), 'If rule lookup and updates should be checked. Defaults to true. Set to false for domains without a rule in DB (e.g., _APP_DOMAIN).', true)
```

hmacr: Safe to check rule by default, no? If they want to, they'll explicitly pass `--check-rule=false` ?

Meldiron:
Closing in favour of https://github.com/appwrite/appwrite/pull/11249#pullrequestreview-3761557447

PR #11226, opened 2026-02-03
Add 7-day job TTL for functions queue

ChiragAgg5k:
## Summary
- Set `jobTTL` to 7 days for `FUNCTIONS_QUEUE_NAME` in `Event::trigger()`
- Uses a `match` expression for easy extensibility to add TTLs for other queues

## Test plan
- [ ] Verify functions queue jobs have 7-day TTL
- [ ] Verify other queues still use default TTL (0)

### DIFF
--- a/src/Appwrite/Event/Event.php
+++ b/src/Appwrite/Event/Event.php
@@ -23,6 +23,7 @@ class Event
 
     public const FUNCTIONS_QUEUE_NAME = 'v1-functions';
     public const FUNCTIONS_CLASS_NAME = 'FunctionsV1';
+    public const FUNCTIONS_QUEUE_TTL = 60 * 60 * 24 * 7; // 7 days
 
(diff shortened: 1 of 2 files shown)


src/Appwrite/Event/Event.php
```diff
@@ -368,8 +368,14 @@ public function trigger(): string|bool
             return false;
         }
 
+        $queueName = $this->getQueue();
+        $jobTTL = match ($queueName) {
+            self::FUNCTIONS_QUEUE_NAME => 60 * 60 * 24 * 7, // 7 days
```
abnegate: Let's use a const for the value as well, `FUNCTIONS_QUEUE_TTL`

abnegate: Can we add  getTTL/setTTL functions so that we can move this to the functions event itself

PR #11232, opened 2026-02-03
Add PR workflow skill for AI agents

ChiragAgg5k:
## Summary
- Adds a skill file at `.claude/skills/pr-workflow.md` to guide AI agents on the correct PR workflow
- Documents that PRs should target version branches (e.g., `1.8.x`) instead of `main`

## Test plan
- Verify the skill file is picked up by Claude Code and other AI agents when working in this repository

### DIFF
--- a/.claude/skills/pr-workflow.md
+++ b/.claude/skills/pr-workflow.md
@@ -1,0 +1,31 @@
+# PR Workflow for Appwrite
+
+## Branch Targeting
+
+**Important:** Appwrite does NOT use `main` as the primary development branch.
+
+PRs should target the **current version branch** (e.g., `1.8.x`, `1.9.x`), not `main`.
+
+### How to determine the correct target branch
+
+1. Check which version branch you're currently on: `git branch --show-current`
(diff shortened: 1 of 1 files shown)


.claude/skills/pr-workflow.md
```diff
@@ -0,0 +1,31 @@
+# PR Workflow for Appwrite
+
+## Branch Targeting
+
+**Important:** Appwrite does NOT use `main` as the primary development branch.
+
+PRs should target the **current version branch** (e.g., `1.8.x`, `1.9.x`), not `main`.
+
+### How to determine the correct target branch
```
Meldiron, line 9: If this works, all good ✅ If AI struggles with this, we can maybe point it to public GitHub API that gives list of 10 most recently merged PRs. I would expect at least one to follow `X.Y.x` target branch pattern, and we will tell AI to use that as base branch too.

As long as we stay active in most recent branch, he will continue in our footsteps.

Meldiron (approved):

PR #11239, opened 2026-02-04
chore: upgrade utopia queue and replace App with Http

ChiragAgg5k:


composer.json
```diff
@@ -72,9 +72,9 @@
         "utopia-php/queue": "0.15.*",
         "utopia-php/registry": "0.5.*",
         "utopia-php/storage": "0.18.*",
-        "utopia-php/swoole": "1.*",
+        "utopia-php/swoole": "1.0.*",
```
abnegate: Let's stick with 1.*, shouldn't be any breaking changes in minor bumps in our own code

PR #11245, opened 2026-02-04
fix: sanitize branch names for valid domain generation

hmacr:
Branch names containing invalid domain characters (like '/') were being used directly when creating VCS preview domains, resulting in invalid domains like 'branch-abc/test.appwrite.network'. This adds a Domain helper class that sanitizes branch names by replacing invalid characters with hyphens before generating domains.

Closes SER-1102


src/Appwrite/Vcs/Domain.php
```diff
@@ -0,0 +1,70 @@
...
+     * Replaces any characters that are not alphanumeric, hyphens, or underscores with hyphens,
+     * and removes leading/trailing hyphens.
+     *
+     * @param string $branch The branch name to sanitize
+     * @return string The sanitized branch name
+     */
+    public static function sanitizeBranchName(string $branch): string
+    {
+        // Replace any sequence of invalid characters with a single hyphen
+        $sanitized = preg_replace('/[^a-zA-Z0-9_-]+/', '-', $branch);
```
Meldiron: Not sure about correctness of this regex. I would expect ^ in different place.

```
 preg_replace('/[^a-zA-Z0-9_-]+/', '-', $branh);
```

hmacr: Feels right to me - replace anything other than those in the list. Why do you think so? 🤔

Meldiron: Actually, regex is scary. lets do something like I did here: https://github.com/appwrite-labs/cloud/pull/3265/changes/4923b135cd1543f041ba572422aa52a8ec2b0a1f

Avoidregex alltogether

src/Appwrite/Vcs/Domain.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+namespace Appwrite\Vcs;
+
+class Domain
```
Meldiron: Let's move it to be transformation adapter

app/controllers/api/vcs.php
```diff
@@ -369,13 +370,7 @@
 
                 // VCS branch preview
                 if (!empty($providerBranch)) {
-                    $branchPrefix = substr($providerBranch, 0, 16);
```
Meldiron, line 319: Note for future: We will need to apply same fix in Cloud vcs.php

src/Appwrite/Vcs/Domain.php
```diff
@@ -0,0 +1,70 @@
...
+     * Replaces any characters that are not alphanumeric or hyphens with hyphens,
+     * and removes leading/trailing hyphens.
+     *
+     * @param string $branch The branch name to sanitize
+     * @return string The sanitized branch name
+     */
+    public static function sanitizeBranchName(string $branch): string
+    {
+        // Replace any sequence of invalid characters with a single hyphen
+        $sanitized = preg_replace('/[^a-zA-Z0-9-]+/', '-', $branch);
```
Meldiron: Let's avoid regex, lets do something similar to: https://github.com/appwrite-labs/cloud/pull/3265/commits/4923b135cd1543f041ba572422aa52a8ec2b0a1f

app/controllers/api/vcs.php
```diff
@@ -369,13 +371,17 @@
...
+                    $transformation = new Transformation([new BranchDomain()]);
+                    $transformation
+                        ->setInput([
+                            'branch' => $providerBranch,
+                            'resourceId' => $resource->getId(),
+                            'projectId' => $project->getId(),
+                            'sitesDomain' => $sitesDomain,
+                        ])
+                        ->setTraits([])
+                        ->transform();
```
Meldiron:
```suggestion
$domain =  "branch-{$branchPrefix}-{$resourceProjectHash}.{$sitesDomain}";

                    $transformation
                        ->setInput($domain)
                        ->setTraits([ maxLenght: 10, suffixLenght: 7 ])
                        ->transform();
```

PR #11250, opened 2026-02-04
Fix double escaping in Build worker tar command

nathfavour:
I noticed that the tar command in the Build worker was using both escapeshellarg and escapeshellcmd on the same variable, which caused build failures for directory paths with special characters like spaces. I've removed the redundant escaping to ensure the paths are handled correctly and securely.


eldadfux:

PR #11268, opened 2026-02-08
Improve certificate dns failure logs

hmacr:
Screenshot 2026-02-09 at 12 39 33 PM

### DIFF
--- a/src/Appwrite/Platform/Workers/Certificates.php
+++ b/src/Appwrite/Platform/Workers/Certificates.php
@@ -489,7 +489,7 @@ class Certificates extends Action
         } catch (AppwriteException $err) {
             $msg = $err->getMessage() . "\n";
             $msg .= "Verify your DNS records are correctly configured and try again.\n";
-            $msg .= "Alternatively, we'll periodically retry verification and update the status.\n";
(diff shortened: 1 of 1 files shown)


src/Appwrite/Platform/Workers/Certificates.php
```diff
@@ -474,11 +477,20 @@ private function validateDomain(Document $rule, Domain $domain, Log $log, ?strin
...
             // TODO: Would be awesome to check A/AAAA record here. Maybe dry run?
+            return;
+        }
+
+        try {
+            $this->verifyRule($rule, $log);
+        } catch (AppwriteException $err) {
+            $msg = $err->getMessage() . "\n";
+            $msg .= "Verify your DNS records are correctly configured and try again.\n";
+            $msg .= "Alternatively, we'll periodically retry verification and update the status.\n";
```
Meldiron: II think we mark status failed. I dont think we re-run failed certificate jobs

hmacr: oh great catch - I was thinking from initial record verification perspective (we never surface logs for that state). I'll remove the retry part.

PR #11269, opened 2026-02-08
Fix redirect url approval for oauth flow

Meldiron:
## What does this PR do?

Allows rules (site domains) automatically for oauth flows reidrect url, same as we alrady allow it for cors.

## Test Plan

Before:

CleanShot 2026-02-08 at 14 12 59@2x

- [x] New test added

## Related PRs and Issues

x

## Checklist

### DIFF
--- a/tests/e2e/Services/Projects/ProjectsConsoleClientTest.php
+++ b/tests/e2e/Services/Projects/ProjectsConsoleClientTest.php
@@ -5091,7 +5091,7 @@ class ProjectsConsoleClientTest extends Scope
             'failure' => 'https://example.com'
         ]);
         $this->assertEquals(200, $response['headers']['status-code']);
-        
+
         /** Ensure any hostname is allowed */
         $response = $this->client->call(Client::METHOD_GET, '/account/sessions/oauth2/' . $provider, [
             'content-type' => 'application/json',
@@ -5104,7 +5104,7 @@ class ProjectsConsoleClientTest extends Scope
             'failure' => 'https://domain-without-rule.com'
         ], followRedirects: false);
         $this->assertEquals(301, $response['headers']['status-code']);
-        
+
         $response = $this->client->call(Client::METHOD_GET, '/account/sessions/oauth2/' . $provider, [
             'content-type' => 'application/json',
             'x-appwrite-project' => $projectId,


app/init/resources.php
```diff
@@ -236,45 +236,68 @@
...
+        // Temporary implementation until custom wildcard domains are an official feature
+        // Allow trusted projects; Used for Console (website) previews
+        if (!$permitsCurrentProject && !$rule->isEmpty() && !empty($rule->getAttribute('projectId', ''))) {
+            $trustedProjects = [];
+            foreach (\explode(',', System::getEnv('_APP_CONSOLE_TRUSTED_PROJECTS', '')) as $trustedProject) {
+                if (empty($trustedProject)) {
+                    continue;
+                }
+                $trustedProjects[] = $trustedProject;
+            }
```
ChiragAgg5k: what do you think of this?

```diff
-$trustedProjects = [];
-foreach (\explode(',', System::getEnv('_APP_CONSOLE_TRUSTED_PROJECTS', '')) as $trustedProject) {
-	if (empty($trustedProject)) {
-		continue;
-	}
-$trustedProjects[] = $trustedProject;
}
+$trustedProjects = \array_filter(
+	\explode(',', System::getEnv('_APP_CONSOLE_TRUSTED_PROJECTS', ''))
+);
```

app/init/resources.php
```diff
@@ -236,45 +236,68 @@
...
+            if (\in_array($rule->getAttribute('projectId', ''), $trustedProjects)) {
+                $permitsCurrentProject = true;
             }
-            $trustedProjects[] = $trustedProject;
         }
-        if (\in_array($rule->getAttribute('projectId', ''), $trustedProjects)) {
-            $permitsCurrentProject = true;
+
+        if ($permitsCurrentProject) {
+            $permittedRule = $rule;
```
ChiragAgg5k: can we just return here directly and remove need for `$permittedRule` variable?

Meldiron: nice. yes.

app/init/resources.php
Meldiron: array filter automatically removes empty? 🤔

We can do, but as separate PR - I know this is in more places, plus in Cloud I recall.
I would prefer separate task. Maybe backlog, maybe Copilot can do it

PR #11289, opened 2026-02-10
Fix GitHub error file path

hmacr:


app/init/constants.php
```diff
@@ -5,6 +5,8 @@
 const APP_NAME = 'Appwrite';
 const APP_DOMAIN = 'appwrite.io';
 
+const APP_VIEWS_DIR = __DIR__ . '/../views';
```
Meldiron, line 8: I dont think const makes sense, it would not work everywhere. I would keep it in specific files

PR #11292, opened 2026-02-11
Fix OAuth redirect for custom scheme state

ChiragAgg5k:
## Summary
- skip rules lookup when OAuth state has no host (custom schemes)
- add e2e coverage for appwrite-callback state in redirect flow

## Test
- not run (not requested)

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -1514,21 +1514,25 @@ Http::get('/v1/account/sessions/oauth2/:provider/redirect')
         }
 
         // Allow redirect to rule URL if related to project
-        //Check if $redirectValidator is instance of Redirect class
+        // Check if $redirectValidator is instance of Redirect class
         if ($redirectValidator instanceof Redirect) {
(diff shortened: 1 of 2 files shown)


app/controllers/api/account.php
```diff
@@ -1514,21 +1514,25 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, arr
...
-                ]),
-                Query::equal('projectInternalId', [$project->getSequence()]),
-                Query::limit(2)
-            ]));
+            $domains = \array_filter([
+                parse_url($state['success'], PHP_URL_HOST) ?? '',
+                parse_url($state['failure'], PHP_URL_HOST) ?? ''
+            ], fn ($domain) => \is_string($domain) && $domain !== '');
+
+            if (!empty($domains)) {
```
stnguyen90, line 1524: Do we need to initialize $rules to be an empty array so it isn't null later on?

ChiragAgg5k: nope, its not used after this, only in this context

stnguyen90 (approved):

PR #11298, opened 2026-02-11
fix: delete all user sessions when blocking a user

### LINKED ISSUE #6061: 🚀 Feature: Delete all user sessions on blocking a user

tanujbhaud:
This PR ensures that when a user is blocked via the API, all their active sessions are immediately deleted. This prevents blocked users from continuing to access the API with existing sessions. Fixes #6061.

### DIFF
--- a/tests/e2e/Services/Users/UsersBase.php
+++ b/tests/e2e/Services/Users/UsersBase.php
@@ -1208,6 +1208,7 @@ trait UsersBase
 
         $this->assertEquals($sessions['headers']['status-code'], 200);
         $sessionCount = count($sessions['body']['sessions']);
+        $this->assertGreaterThan(0, $sessionCount);
 
         // Block the user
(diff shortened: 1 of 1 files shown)


eldadfux:
We have decided not to move forward with this feature as it attempts to force a specific user journey on all use cases. In some cases blocking a user might be temporary and not require deleting all sessions. The API does allow to handle this by calling the method to delete all sessions in addition to blocking a user so this doesn't unlock an option that wasn't possible before.

PR #11326, opened 2026-02-14
fix: allow null name in Users.CreateUser endpoints to prevent 500 error

lcz24:
### What does this PR do?
This PR resolves issue #8785 where providing a \`null\` value for the \`name\` parameter in the \`Users.CreateUser\` (and related hashing endpoints like bcrypt, argon2, etc.) would cause a vague 500 Internal Server Error instead of a proper validation error or defaulting to an empty string.

### Why is this needed?
In the current implementation, the \`name\` parameter is marked as optional in Utopia but type-hinted as a non-nullable \`string\` in the controller actions. Explicitly passing \`null\` in the JSON request triggers a PHP \`TypeError\`.

### Changes:
1. Updated \`createUser\` function signature to accept a nullable \`?string $name\`.
2. Wrapped the \`name\` parameter validator in \`Nullable\` for all user creation endpoints.
3. Updated action functions to accept \`?string $name\` and ensure it defaults to an empty string before processing.

Fixes #8785

### DIFF
--- a/app/controllers/api/users.php
+++ b/app/controllers/api/users.php
@@ -73,7 +73,7 @@ use Utopia\Validator\Text;
 use Utopia\Validator\WhiteList;
 
 /** TODO: Remove function when we move to using utopia/platform */
-function createUser(Hash $hash, string $userId, ?string $email, ?string $password, ?string $phone, string $name, Document $project, Database $dbForProject, Hooks $hooks): Document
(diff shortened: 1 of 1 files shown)


eldadfux:
@lcz24 thank you for your contribution, this was already addressed in https://github.com/appwrite/appwrite/pull/11413

PR #11327, opened 2026-02-14
fix(auth): invalidate JWT if session is expired or deleted

### LINKED ISSUE #8000: 🐛 Bug Report: JWT still valid after Session Timeout
### 👟 Reproduction steps

### 👍 Expected behavior

### 👎 Actual Behavior

### 🎲 Appwrite version

Version 1.4.x

### 💻 Operating system

Linux

### 🧱 Your Environment

lcz24:
### What does this PR do?
This PR addresses issue #8000 by adding a validation check during JWT authentication. It ensures that the session associated with the JWT is not only present but also valid (not expired).

### Why is this needed?
Currently, a JWT remains valid until its own expiration time, even if the underlying session has been deleted (logout) or has expired. This creates a security gap where a revoked session can still be accessed via a valid JWT.

### Changes
- In `app/init/resources.php`, added logic to check `$session['expire']` against the current time using the project's `DatabaseDateTime` alias.
- If the session is expired, the user context is reset to an empty User, effectively invalidating the request.

Fixes #8000

### DIFF
--- a/app/init/resources.php
+++ b/app/init/resources.php
@@ -424,7 +424,11 @@ App::setResource('user', function (string $mode, Document $project, Document $co
         }
         $jwtSessionId = $payload['sessionId'] ?? '';
         if (!empty($jwtSessionId)) {
-            if (empty($user->find('$id', $jwtSessionId, 'sessions'))) { // Match JWT to active token
+            $session = $user->find('$id', $jwtSessionId, 'sessions');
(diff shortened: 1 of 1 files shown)


stnguyen90:
@lcz24, thanks for your enthusiasm in contributing to Appwrite! As mentioned in our [contributing guide](https://github.com/appwrite/appwrite/blob/main/CONTRIBUTING.md#how-to-start), it's important to:

1. ask to be assigned on an issue
2. make sure you're assigned on an issue before submitting a PR
3. create a thread for each issue

This way, we can properly discuss the issue and any follow up as needed **before** you start working on anything. We would hate for you to submit a PR only for it to never be merged because it's not the direction we want.

PR #11331, opened 2026-02-15
Add schedules API endpoints

premtsd-code:
## Summary
- Adds REST API endpoints for schedules under the projects group: Create (`POST /v1/projects/:projectId/schedules`), Get (`GET /v1/projects/:projectId/schedules/:scheduleId`), and List (`GET /v1/projects/:projectId/schedules`)
- Auth: `AuthType::ADMIN` (console session), consistent with keys and platforms endpoints
- Adds `schedules.read` and `schedules.write` scopes, event configuration, and Response model

## Test plan
- [x] `composer lint` passes
- [x] `composer format` passes
- [x] Added E2E tests
- [x] CI pipeline passes

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3797,16 +3797,16 @@
         },
         {
             "name": "utopia-php/database",
-            "version": "5.3.0",
+            "version": "5.3.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/utopia-php/database.git",
-                "reference": "5e49f32ebb6e40a78209231564c40e5068810429"
+                "reference": "e3305f6ca64eb5dd715e30212fd32d5d834a3a46"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/utopia-php/database/zipball/5e49f32ebb6e40a78209231564c40e5068810429",
-                "reference": "5e49f32ebb6e40a78209231564c40e5068810429",
+                "url": "https://api.github.com/repos/utopia-php/database/zipball/e3305f6ca64eb5dd715e30212fd32d5d834a3a46",
+                "reference": "e3305f6ca64eb5dd715e30212fd32d5d834a3a46",
                 "shasum": ""
             },
             "require": {
@@ -3849,9 +3849,9 @@
             ],
             "support": {
                 "issues": "https://github.com/utopia-php/database/issues",
-                "source": "https://github.com/utopia-php/database/tree/5.3.0"
+                "source": "https://github.com/utopia-php/database/tree/5.3.1"
             },
-            "time": "2026-02-17T11:50:01+00:00"
+            "time": "2026-02-18T07:15:40+00:00"
         },
         {
             "name": "utopia-php/detector",
@@ -8885,7 +8885,7 @@
     ],
     "aliases": [],
     "minimum-stability": "stable",
-    "stability-flags": {},
+    "stability-flags": [],
     "prefer-stable": false,
     "prefer-lowest": false,
     "platform": {


src/Appwrite/Platform/Modules/Projects/Http/Schedules/Create.php
```diff
@@ -0,0 +1,114 @@
...
+                    'resourceInternalId' => $resource->getSequence(),
+                    'resourceUpdatedAt' => DateTime::now(),
+                    'projectId' => $project->getId(),
+                    'schedule' => $schedule,
+                    'active' => $active,
+                ]))
+            );
+        } catch (DuplicateException) {
+            throw new Exception(Exception::GENERAL_SERVER_ERROR, 'Failed to create schedule. Please try again.');
+        }
```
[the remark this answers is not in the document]
premtsd-code: DateTime::now() is the correct convention used everywhere in the codebase

Dockerfile
```diff
@@ -14,6 +14,10 @@ RUN composer install --ignore-platform-reqs --optimize-autoloader \
 
 FROM appwrite/base:1.0.0 AS base
 
+RUN wget -qO /usr/local/share/ca-certificates/AAA_Certificate_Services.crt \
+    "https://crt.sh/?d=331986" \
+    && update-ca-certificates
```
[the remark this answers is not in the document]
premtsd-code: removed

src/Appwrite/Platform/Modules/Projects/Http/Schedules/XList.php
```diff
@@ -0,0 +1,123 @@
...
+
+    public function action(
+        string $projectId,
+        array $queries,
+        bool $includeTotal,
+        Response $response,
+        Database $dbForPlatform,
+        Authorization $authorization,
+    ): void {
+        $project = $dbForPlatform->getDocument('projects', $projectId);
```
fogelito, line 67: Do we need to validate that the projectId is in the same region?

src/Appwrite/Platform/Modules/Projects/Http/Schedules/Create.php
abnegate, line 1: To be consistent with the REST path, we should be nested in the project namespace here like `src/Appwrite/Platform/Modules/Schedules/Http/Projects/Schedules/Create.php`

src/Appwrite/Platform/Modules/Schedules/Http/Schedules/Create.php
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    protected function getNotFoundExceptionMap(): array
+    {
+        return [
+            SCHEDULE_RESOURCE_TYPE_FUNCTION => Exception::FUNCTION_NOT_FOUND,
+            SCHEDULE_RESOURCE_TYPE_EXECUTION => Exception::EXECUTION_NOT_FOUND,
+            SCHEDULE_RESOURCE_TYPE_MESSAGE => Exception::MESSAGE_NOT_FOUND,
+        ];
+    }
```
abnegate: Instead of returning the map, let's accept key as parameter and return the value we need directly

src/Appwrite/Platform/Modules/Schedules/Http/Schedules/Create.php
```diff
@@ -0,0 +1,160 @@
...
+        ];
+
+        if ($data !== null) {
+            $attributes['data'] = \json_decode($data, true);
+        }
+
+        try {
+            $doc = $authorization->skip(
+                fn () => $dbForPlatform->createDocument('schedules', new Document($attributes))
+            );
```
abnegate: We shouldn't need to skip auth, in admin scope auth is always disabled

phpunit.xml
```diff
@@ -35,6 +35,7 @@
             <directory>./tests/e2e/Services/Webhooks</directory>
             <directory>./tests/e2e/Services/Messaging</directory>
             <directory>./tests/e2e/Services/Migrations</directory>
+            <directory>./tests/e2e/Services/Schedules</directory>
```
abnegate: Let's nest the tests inside the projects dir

PR #11362, opened 2026-02-19
Move team memberships API to Modules

hmacr:
Towards SER-633

### DIFF
--- a/src/Appwrite/Platform/Modules/Teams/Http/Memberships/Delete.php
+++ b/src/Appwrite/Platform/Modules/Teams/Http/Memberships/Delete.php
@@ -22,12 +22,6 @@ class Delete extends Action
 {
     use HTTP;
 
-    protected string $teamId;
-    protected string $membershipId;
-    protected Document $user;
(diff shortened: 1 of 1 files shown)


Meldiron (approved):

src/Appwrite/Platform/Modules/Teams/Http/Memberships/Delete.php
```diff
@@ -0,0 +1,177 @@
...
+            'team' => $team,
+        ] = $this->loadAndValidateResources();
+
+        try {
+            $dbForProject->deleteDocument('memberships', $membership->getId());
+        } catch (AuthorizationException $exception) {
+            throw new Exception(Exception::USER_UNAUTHORIZED);
+        } catch (\Throwable $exception) {
+            throw new Exception(Exception::GENERAL_SERVER_ERROR, 'Failed to remove membership from DB');
+        }
```
loks0n: Better to not catch this and let error handler catch this so we see the root stack trace in logs/sentry

PR #11369, opened 2026-02-20
Fix: url params in redirect URLs

Meldiron:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/controllers/api/account.php
```diff
@@ -2111,7 +2111,16 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, arr
         if (empty(System::getEnv('_APP_SMTP_HOST'))) {
             throw new Exception(Exception::GENERAL_SMTP_DISABLED, 'SMTP disabled');
         }
-        $url = htmlentities($url);
+
+        $url = \parse_url($url);
```
stnguyen90: we should use https://github.com/appwrite/appwrite/blob/main/src/Appwrite/URL/URL.php. Remember to factor in mobile custom url schemes.

app/controllers/api/account.php
stnguyen90, line 2227: @Meldiron also careful about this parsing we're doing here too. We should probably not have URL parsing logic in the Template class.

app/controllers/api/account.php
Meldiron: added test for it too

PR #11395, opened 2026-02-25
fix: add session alert email template translations for all languages

slegarraga:
## What does this PR do?

Fixes #8659 — Session alert email template was broken for all languages except English.

### Problem

The session alert email template keys (`emails.sessionAlert.*`) were only defined in `en.json`. When a user's locale was set to any non-English language (e.g., `es`, `fr`, `de`), the email displayed raw template key names like `{{emails.sessionAlert.subject}}` instead of actual readable text.

### Solution

Added the English session alert translation keys as fallback to all 72 non-English locale files. This ensures users receive readable session alert emails regardless of their locale setting.

While ideally each language would have its own native translation, having the English text as fallback is far better than showing raw template keys. Native translations can be contributed incrementally.

### Changes

- Added 10 `emails.sessionAlert.*` keys to all 72 non-English translation JSON files in `app/config/locale/translations/`

### Testing

- Verified `en.json` was not modified
- All JSON files remain valid and properly formatted
- Keys are consistent across all locale files


slegarraga:
Closing to focus maintainer track on Spectral / OpenAI Agents / Anthropic SDK. Happy to reopen or re-submit if still needed.

PR #11396, opened 2026-02-25
fix: return proper validation error when null name is provided to CreateUser endpoints

slegarraga:
## What does this PR do?

Fixes #8785

All `Users.CreateUser*` endpoints return a vague server error when `null` is provided as the `name` parameter. Since `name` is optional, passing `null` should be treated the same as omitting it entirely.

## Changes

- Wrapped the `name` parameter validator with `Nullable` on all 8 create user endpoints, matching the existing pattern used for `email` and `phone` parameters
- Changed the default value from `''` to `null` for the `name` parameter on create user endpoints
- Updated the `createUser` helper function signature to accept `?string $name` and coalesce `null` to `''` internally

## How to test

1. POST to `/v1/users` with `name: null` in the request body
2. Previously: returns a vague 500 server error
3. Now: successfully creates the user with no name (same as omitting the field)


slegarraga:
Closing to focus maintainer track on Spectral / OpenAI Agents / Anthropic SDK. Happy to reopen or re-submit if still needed.

PR #11400, opened 2026-02-25
fix: show timed-out executions as failed across API endpoints

ChiragAgg5k:
## Summary

Executions that time out can remain stuck in `waiting` or `processing` status in the database. This mirrors the frontend workaround from appwrite/console#2788 across the relevant API endpoints for both functions and sites.

- **GET** execution/log: override status to `failed` in the response if elapsed time since creation exceeds the resource's configured timeout
- **LIST** executions/logs: same in-response override; when the caller filters by `failed`, expands the DB query with an `OR` to also fetch `waiting`/`processing` entries created before the timeout threshold so they appear in results; skips the in-response override when the caller explicitly requests a non-failed status to avoid contradicting the filter
- **DELETE** execution: allows deletion of timed-out executions still stored as `waiting`/`processing` by treating them as `failed` for the status guard

All changes are in-memory only — database records are not modified.

> [!NOTE]
> These changes should be removed once a proper database-level fix is applied.

## Test plan

- [ ] GET a timed-out execution (stuck in `waiting`/`processing`) — response should show `status: failed`
- [ ] LIST executions with `status = failed` filter — timed-out entries should appear
- [ ] LIST executions with `status = waiting` or `status = processing` filter — timed-out entries should not appear
- [ ] LIST executions with no status filter — timed-out entries should appear as `failed`
- [ ] DELETE a timed-out execution stuck in `waiting`/`processing` — should succeed
- [ ] DELETE an execution that is genuinely still in progress (within timeout) — should still be blocked

### DIFF
--- a/src/Appwrite/Platform/Modules/Functions/Http/Executions/XList.php
+++ b/src/Appwrite/Platform/Modules/Functions/Http/Executions/XList.php
@@ -11,6 +11,7 @@ use Appwrite\Utopia\Database\Documents\User;
 use Appwrite\Utopia\Database\Validator\Queries\Executions;
 use Appwrite\Utopia\Response;
 use Utopia\Database\Database;
+use Utopia\Database\DateTime;
 use Utopia\Database\Document;
 use Utopia\Database\Exception\Order as OrderException;
 use Utopia\Database\Exception\Query as QueryException;
@@ -110,6 +111,35 @@ class XList extends Base
             $cursor->setValue($cursorDocument);
         }
 
+        // Calculate the cutoff datetime before which a waiting/processing execution is considered timed out.
+        $timeout = $function->getAttribute('timeout', 900);
+        $thresholdDate = new \DateTime("-{$timeout} seconds");
+        $threshold = DateTime::format($thresholdDate);
+
+        // Capture what statuses the caller explicitly requested, before we mutate the query.
+        $requestedStatuses = [];
+        foreach ($queries as $query) {
+            if ($query->getMethod() === Query::TYPE_EQUAL && $query->getAttribute() === 'status') {
+                $requestedStatuses = [...$requestedStatuses, ...$query->getValues()];
+            }
+        }
+
+        // If the caller is filtering by 'failed', expand the DB query to also return
+        // waiting/processing executions created before the timeout threshold, so timed-out
+        // executions that were never marked failed in the DB are included in the results.
+        foreach ($queries as $index => $query) {
+            if ($query->getMethod() === Query::TYPE_EQUAL && $query->getAttribute() === 'status' && \in_array('failed', $query->getValues())) {
+                $queries[$index] = Query::or([
+                    $query,
+                    Query::and([
+                        Query::equal('status', ['waiting', 'processing']),
+                        Query::createdBefore($threshold),
+                    ]),
+                ]);
+                break;
+            }
+        }
+
         $filterQueries = Query::groupByType($queries)['filters'];
 
         try {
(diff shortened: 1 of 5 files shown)


loks0n (approved):

src/Appwrite/Platform/Modules/Functions/Http/Executions/XList.php
```diff
@@ -110,6 +110,21 @@ public function action(
...
+            }
+        }
+
+        if (\in_array('failed', $requestedStatuses)) {
+            foreach ($queries as $query) {
+                if ($query->getMethod() === Query::TYPE_EQUAL && $query->getAttribute() === 'status') {
+                    $query->setValues(\array_unique(\array_merge($query->getValues(), ['waiting', 'processing'])));
+                }
+            }
+        }
```
Meldiron: I think this would break pagination, due to need for array_filter below.

I think logic should be:

If status query exists and looks for failed deployment, override query

Before:
```
queries: [ Query.limit(5), Query.equal('status', 'failed') ]
```

After
```
queries: [
  Query.limit(5),
  Query.or([
    Query.equal('status', 'failed') ],
    Query.and([
      Query.equal('status', ['waiting', 'processing'],
      Query.lessThan('createdAt', now-15min),
    ]).
  ]),


```

Meldiron: this would still not cover failed status queries in nested ors and ands, but should cover at least simple searches

src/Appwrite/Platform/Modules/Functions/Http/Executions/XList.php
```diff
@@ -110,6 +111,35 @@ public function action(
...
+        // If the caller is filtering by 'failed', expand the DB query to also return
+        // waiting/processing executions created before the timeout threshold, so timed-out
+        // executions that were never marked failed in the DB are included in the results.
+        foreach ($queries as $index => $query) {
+            if ($query->getMethod() === Query::TYPE_EQUAL && $query->getAttribute() === 'status' && \in_array('failed', $query->getValues())) {
+                $queries[$index] = Query::or([
+                    $query,
+                    Query::and([
+                        Query::equal('status', ['waiting', 'processing']),
+                        Query::createdBefore($threshold),
```
stnguyen90, line 136: We need to make sure we have indexes in place so that this isn't too slow for large data sets.

src/Appwrite/Platform/Modules/Sites/Http/Logs/Get.php
```diff
@@ -71,6 +71,16 @@ public function action(string $siteId, string $logId, Response $response, Databa
...
 
+        // Override status in response if the log is stuck in waiting/processing beyond the site timeout.
+        $status = $log->getAttribute('status', '');
+        if ($status === 'waiting' || $status === 'processing') {
+            $timeout = $site->getAttribute('timeout', 30);
+            $elapsed = \time() - \strtotime($log->getCreatedAt());
+            if ($elapsed >= $timeout) {
+                $log->setAttribute('status', 'failed');
+            }
+        }
```
stnguyen90, line 82: We might need the same logic in the delete endpoint to allow people to delete the executions.

src/Appwrite/Platform/Modules/Functions/Http/Executions/XList.php
ChiragAgg5k: Screenshot 2026-02-26 at 9 19 55 AM

yep we do

src/Appwrite/Platform/Modules/Functions/Http/Executions/XList.php
```diff
@@ -110,6 +111,35 @@ public function action(
...
+        $thresholdDate = new \DateTime("-{$timeout} seconds");
+        $threshold = DateTime::format($thresholdDate);
+
+        // Capture what statuses the caller explicitly requested, before we mutate the query.
+        $requestedStatuses = [];
+        foreach ($queries as $query) {
+            if ($query->getMethod() === Query::TYPE_EQUAL && $query->getAttribute() === 'status') {
+                $requestedStatuses = [...$requestedStatuses, ...$query->getValues()];
+            }
+        }
```
Meldiron, line 125: I think we dont need this

src/Appwrite/Platform/Modules/Functions/Http/Executions/XList.php
```diff
@@ -119,6 +149,20 @@ public function action(
             throw new Exception(Exception::DATABASE_QUERY_ORDER_NULL, "The order attribute '{$e->getAttribute()}' had a null value. Cursor pagination requires all documents order attribute values are non-null.");
         }
 
+        // Override status in response for timed-out executions, but only when the caller
+        // did not explicitly request a non-failed status (e.g. waiting/processing).
+        if (empty(\array_diff($requestedStatuses, ['failed']))) {
```
Meldiron, line 154: We might not need this `if` - same as GET, always replace it if condition is met - waiting or processing, and too long

src/Appwrite/Platform/Modules/Functions/Http/Executions/XList.php
stnguyen90: If we're querying on status and created before, those need to be in the index too.

PR #11402, opened 2026-02-25
Documentsdb + vectordb (latest)

ArnabChatterjee20k:
## What does this PR do?

## Test Plan
Currently this pr is pointing to 1.8.x but will be merged in documentsdb branch as this branch is extracted from documentsdb branch only and contains all the changes needed for the documentsdb syncing with 1.8.x branch

But currently it is poininting to 1.8.x to see the changes needed easily

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/config/collections/projects.php
```diff
@@ -61,6 +61,15 @@
...
+            [
+                '$id' => ID::custom('database'),
+                'type' => Database::VAR_STRING,
+                'size' => 128,
+                'required' => true,
+                'signed' => true,
+                'array' => false,
+                'filters' => [],
+            ]
         ],
```
abnegate, line 73: @ArnabChatterjee20k Let's run the migration ASAP

app/controllers/api/migrations.php
```diff
@@ -43,6 +43,15 @@
 
 include_once __DIR__ . '/../shared/api.php';
 
+function getDatabaseTransferResourceServices(string $databaseType)
+{
+    return match($databaseType) {
+        DATABASE_TYPE_LEGACY,
+        DATABASE_TYPE_TABLESDB => Transfer::GROUP_DATABASES_TABLES_DB,
+        DATABASE_TYPE_VECTORSDB => Transfer::GROUP_DATABASES_VECTOR_DB
+    };
```
abnegate, line 53: Do we need DocumentsDB here too?

app/controllers/api/migrations.php
```diff
@@ -427,8 +436,16 @@
             throw new \Exception('Unable to copy file');
         }
 
+        // getting databasetype
+        $resources = explode(':', $resourceId);
+        $databaseId = $resources[0];
+        $database = $authorization->skip(fn () => $dbForProject->getDocument('databases', $databaseId));
+        $databaseType = $database->getAttribute('type');
+        if (!in_array($databaseType, CSV_ALLOWED_DATABASE_TYPES)) {
+            throw new Exception(Exception::GENERAL_BAD_REQUEST, 'Database type not supported for csv');
```
abnegate: Let's use a custom exception type here

app/init/registers.php
```diff
@@ -329,6 +357,9 @@
                         };
 
                         $adapter->setDatabase($dsn->getPath());
+                        if ($key === 'documentsdb') {
+                            $adapter->setSupportForAttributes(false);
+                        }
                         return $adapter;
```
abnegate: Are we able to do this in the resource init for DBs instead of pool init?

app/init/resources.php
```diff
@@ -550,7 +552,8 @@
     return new Authorization();
 }, []);
 
-Http::setResource('dbForProject', function (Group $pools, Database $dbForPlatform, Cache $cache, Document $project, Response $response, Publisher $publisher, Publisher $publisherFunctions, Publisher $publisherWebhooks, Event $queueForEvents, Func $queueForFunctions, Webhook $queueForWebhooks, Realtime $queueForRealtime, StatsUsage $queueForStatsUsage, Authorization $authorization) {
+// TODO: check here do we really need the database listener for the dbForProject?
```
abnegate: Yep, we init all the listeners here how since they're related to the resource init instead of an API-specific hook

app/init/resources.php
```diff
@@ -816,6 +852,270 @@
...
+                }
+            });
+
+        return $database;
+    };
+
+}, ['pools','cache','project','request','queueForStatsUsage','authorization']);
+
+
+Http::setResource('getDatabasesDB', function (Group $pools, Cache $cache, Document $project, Request $request, StatsUsage $queueForStatsUsage, Authorization $authorization) {
```
abnegate: Duplicate resource

src/Appwrite/Databases/TransactionState.php
```diff
@@ -21,17 +21,23 @@ class TransactionState
 {
     private Database $dbForProject;
     private Authorization $authorization;
-    /** @var Authorization $authorization */
-    public function __construct(Database $dbForProject, Authorization $authorization)
+    /** @var Authorization $authorization */    /**
+     * @var callable(Document $database): Database
+    */
```
abnegate:
```suggestion
    /**
     * @var callable(Document $database): Database
     */
```

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Create.php
```diff
@@ -111,6 +111,7 @@ public function __construct()
                         new Parameter('collectionId', optional: false),
                         new Parameter('documents', optional: false),
                         new Parameter('transactionId', optional: true),
+                        new Parameter('transactionId', optional: true),
```
abnegate: Duplicate

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Action.php
```diff
@@ -15,6 +15,8 @@ abstract class Action extends UtopiaAction
      */
     private ?string $context = COLLECTIONS;
 
+    private ?string $databaseType = TABLESDB;
```
abnegate: I think we should default to legacy here

src/Appwrite/Platform/Modules/Databases/Http/Databases/Transactions/Action.php
```diff
@@ -10,11 +10,46 @@ abstract class Action extends DatabasesAction
      * The current API context (either 'table' or 'collection').
      */
     private ?string $context = COLLECTIONS;
+    private ?string $databaseType = TABLESDB;
```
abnegate: Same here, default legacy

src/Appwrite/Platform/Modules/Databases/Http/Databases/XList.php
```diff
@@ -92,6 +91,9 @@ public function action(array $queries, string $search, bool $includeTotal, Utopi
             $cursor->setValue($cursorDocument);
         }
 
+        // use base action to get all types, contexts, etc!
```
abnegate:
```suggestion
```

src/Appwrite/Platform/Modules/Databases/Http/DocumentsDB/Collections/Documents/Attribute/Decrement.php
abnegate, line 1: Let's remove this and the incrememnt routes, the suggested way to do it now is via operators

src/Appwrite/Platform/Modules/Databases/Http/VectorsDB/Embeddings/Text/Create.php
```diff
@@ -0,0 +1,157 @@
...
+                    ],
+                    contentType: ContentType::JSON,
+                    parameters: [
+                        new Parameter('databaseId', optional: false),
+                        new Parameter('collectionId', optional: false),
+                        new Parameter('documents', optional: false),
+                    ]
+                )
+            ])
+            ->param('texts', [], fn (array $plan) => new ArrayList(new Text(0), $plan['databasesBatchSize'] ?? APP_LIMIT_DATABASE_BATCH), 'Array of text to generate embeddings.', false, ['plan'])
```
abnegate: Let's add a custom plan value for this, databasesMaxEmbeddingTexts, we can use the same values as batch size for now and tweak as needed

src/Appwrite/Platform/Modules/Databases/Http/VectorsDB/Embeddings/Text/Create.php
```diff
@@ -0,0 +1,157 @@
...
+            ->inject('queueForStatsUsage')
+            ->inject('log')
+            ->inject('logger')
+            ->callback($this->action(...));
+    }
+
+    public function action(array $texts, string $model, UtopiaResponse $response, Document $project, Agent $embeddingAgent, StatsUsage $queueForStatsUsage, Log $log, ?Logger $logger): void
+    {
+        $results = [];
+        $embeddingAgent->getAdapter()->setModel($model)->setTimeout(60000);
```
abnegate: Somewhere we're setting 90 seconds as well, let's use an env var and make it consistently 30 second timeout (that can be overridden for self hosted)

src/Appwrite/Platform/Modules/Databases/Http/VectorsDB/Embeddings/Text/Create.php
```diff
@@ -0,0 +1,157 @@
...
+                    $log->addTag('code', $e->getCode());
+                    $log->addTag('projectId', $project->getId());
+
+                    $log->addExtra('file', $e->getFile());
+                    $log->addExtra('line', $e->getLine());
+                    $log->addExtra('trace', $e->getTraceAsString());
+
+                    $logger->addLog($log);
+                }
+            }
```
abnegate, line 124: We should let the error handle take care of this, we can still catch, but let's just rethrow with the extra context added to a new exception

src/Appwrite/Platform/Modules/Databases/Http/VectorsDB/Embeddings/Text/Create.php
```diff
@@ -0,0 +1,157 @@
...
+            ->addMetric(\str_replace('{embeddingModel}', $model, METRIC_EMBEDDINGS_MODEL_TEXT), \count($texts))
+            ->addMetric(METRIC_EMBEDDINGS_TEXT_TOTAL_TOKENS, $totalTokens)
+            ->addMetric(\str_replace('{embeddingModel}', $model, METRIC_EMBEDDINGS_MODEL_TEXT_TOTAL_TOKENS), $totalTokens)
+            ->addMetric(METRIC_EMBEDDINGS_TEXT_TOTAL_DURATION, $totalDuration)
+            ->addMetric(\str_replace('{embeddingModel}', $model, METRIC_EMBEDDINGS_MODEL_TEXT_TOTAL_DURATION), $totalDuration)
+            ->addMetric(METRIC_EMBEDDINGS_TEXT_TOTAL_ERROR, $totalErrors)
+            ->addMetric(\str_replace('{embeddingModel}', $model, METRIC_EMBEDDINGS_MODEL_TEXT_TOTAL_ERROR), $totalErrors)
+            ->trigger();
+
+        $queueForStatsUsage->reset();
```
abnegate: Do we need to manually trigger and reset? Shutdown hook should handle it already without this

src/Appwrite/Platform/Tasks/Install.php
abnegate, line 1: Let's revert the changes here, will conflict with the web installer PR

src/Appwrite/Platform/Workers/Deletes.php
```diff
@@ -617,23 +619,64 @@ protected function deleteProject(Database $dbForPlatform, callable $getProjectDB
...
+                                        Console::error(
+                                            'Error deleting ' . $collection->getId() . ' ' . $e->getMessage()
+                                        );
+                                    }
+                                }
+                            );
+                        }
+                    ),
+                $databasesToClean
+            ));
```
abnegate, line 703: Let's move the inner functions to private functions on the worker to make this a bit more readable, something like 

```php
batch(array_map(
    fn ($databaseDoc) => fn () => $this->cleanDatabase(
        $databaseDoc,
        $executionActionPerDatabase,
        $projectTables,
        $projectCollectionIds
    ),
    $databasesToClean
));
```

src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -516,6 +541,14 @@ function ($resources) use ($migration, $transfer, $project, $queueForRealtime, &
         }
     }
 
+    protected function getDatabasesDBForProject(Document $database)
+    {
+        if ($this->sourceProject) {
+            return call_user_func($this->getDatabasesDB, $database, $this->sourceProject);
+        }
+        return call_user_func($this->getDatabasesDB, $database);
+    }
```
abnegate, line 571: Let's use modern callable syntax instead of call_user_func

src/Appwrite/SDK/Specification/Format.php
```diff
@@ -348,6 +348,44 @@ protected function getRequestEnumName(string $service, string $method, string $p
...
+                    case 'getCollectionUsage':
+                        switch ($param) {
+                            case 'range':
+                                return 'UsageRange';
+                        }
+                        break;
+                    case 'createIndex':
+                        switch ($param) {
+                            case 'type':
+                                return 'VectorsDBIndexType';
```
abnegate, line 383: If we prefix one, we should for the rest so they're consistent

src/Appwrite/Utopia/Database/Validator/Attributes.php
```diff
@@ -45,10 +45,12 @@ class Attributes extends Validator
     /**
      * @param int $maxAttributes Maximum number of attributes allowed
      * @param bool $supportForSpatialAttributes Whether DB supports spatial attributes
+     * @param bool $supportForSpatialAttributes Whether DB supports attributes or not
```
abnegate: Wrong param name

src/Appwrite/Utopia/Database/Validator/Attributes.php
```diff
@@ -78,6 +80,11 @@ public function isValid($value): bool
             return false;
         }
 
+        if (\count($value) && !$this->supportForAttributes) {
+            $this->message = "Attributes are not supported by the current database";
```
abnegate: Single quotes

src/Appwrite/Utopia/Database/Validator/Operation.php
abnegate, line 71: Do we need to handle other types here?

src/Appwrite/Utopia/Response/Model/Embedding.php
```diff
@@ -0,0 +1,47 @@
...
+
+    public function __construct()
+    {
+        $this
+            ->addRule('model', [
+                'type' => self::TYPE_STRING,
+                'description' => 'Embedding model used to generate embeddings.',
+                'example' => 'embeddinggemma'
+            ])
+            ->addRule('dimension', [
```
abnegate, line 27: Let's make this `dimensions` instead

tests/e2e/Services/Databases/TablesDB/DatabasesBase.php
abnegate, line 1: Can we used the shared base?

ollama.dockerfile
abnegate, line 1: Let's remove and push to the remote repo

postgres.dockerfile
abnegate, line 1: This one is published, we can use `appwrite/postgres:0.1.0` here

composer.json
```diff
@@ -87,7 +87,7 @@
         "phpmailer/phpmailer": "6.9.*",
         "chillerlan/php-qrcode": "4.3.*",
         "adhocore/jwt": "1.1.*",
-        "spomky-labs/otphp": "11.*",
+        "spomky-labs/otphp": "^11.4.2",
```
abnegate: Let's revert unless needed

docker-compose.yml
```diff
@@ -1310,6 +1339,23 @@ services:
       - POSTGRES_PASSWORD=${_APP_DB_PASS}
     command: "postgres"
 
+  ollama:
+    build: 
+      context: https://github.com/appwrite/docker-ollama.git#feat-dockerfile
+      args:
+        MODELS: ${_APP_EMBEDDING_MODELS:-embeddinggemma}
+        # duration to keep model in memory
+        OLLAMA_KEEP_ALIVE: 24h
```
abnegate: Let's make this forever if possible, no need to re-download unless the model itself needs updating

app/controllers/api/migrations.php
ArnabChatterjee20k: updated
a note, didn't give any side effect as the DATABASE_TYPE_VECTORSDB => Transfer::GROUP_DATABASES_VECTOR_DB and Documentsdb string values are same

src/Appwrite/Utopia/Response/Model/Embedding.php
ArnabChatterjee20k: I added it as previously as dimensions only but we are sending dimension of a vector -> shouldn't it be dimension only?

src/Appwrite/Platform/Modules/Databases/Http/VectorsDB/Embeddings/Text/Create.php
ArnabChatterjee20k: We discussed this earlier, 
The current issue with the batch processing is that if an error occurs, the entire batch flow breaks midway.

For example, if a user sends 10 texts and an error occurs while processing the 5th one, the system returns the error instead of the completed result. Because of this, the results for the first 4 successfully processed texts are discarded, and the remaining texts are never processed.

Additionally, there is a scenario where only one text in the batch fails while the others succeed. In the current implementation, even though most of the texts are processed successfully, the entire batch may still fail due to that single error.

src/Appwrite/Platform/Modules/Databases/Http/DocumentsDB/Collections/Documents/Attribute/Decrement.php
ArnabChatterjee20k: But mongodb doesn't support operators

src/Appwrite/Utopia/Database/Validator/Operation.php
ArnabChatterjee20k: Not needed as the vectordb doesn't support opeartions

src/Appwrite/Platform/Modules/Databases/Http/DocumentsDB/Collections/Documents/Attribute/Decrement.php
abnegate: Good point, we can keep it for now

abnegate (approved):

PR #11409, opened 2026-02-25
feat: backup policy migration support

premtsd-code:
## Summary
- Add `policies.read` and `policies.write` scopes to the migration worker's internal JWT so backup policies can be created on the destination project
- Add backup policy resource to MigrationReport model and SDK spec Format

## Test plan
- [ ] Trigger an Appwrite-to-Appwrite migration with backup-policy in the resources list
- [ ] Verify the migration completes without 401 scope errors
- [ ] Verify backup policies are created on the destination project


abnegate:
@cursor Sync with main and get all checks passing

PR #11414, opened 2026-02-26
Add region check

abnegate:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/app/realtime.php
+++ b/app/realtime.php
@@ -248,48 +248,62 @@ $adapter
 
 $server = new Server($adapter);
 
-$logError = function (Throwable $error, string $action) use ($register) {
-    $logger = $register->get('realtimeLogger');
+// Allows overriding
+if (!function_exists('logError')) {
+    function logError(Throwable $error, string $action, array $tags = [], ?Document $project = null, ?Document $user = null, ?Authorization $authorization = null): void
+    {
+        global $register;
 
-    if ($logger && !$error instanceof Exception) {
-        $version = System::getEnv('_APP_VERSION', 'UNKNOWN');
+        $logger = $register->get('realtimeLogger');
 
-        $log = new Log();
-        $log->setNamespace('realtime');
-        $log->setServer(System::getEnv('_APP_LOGGING_SERVICE_IDENTIFIER', \gethostname()));
-        $log->setVersion($version);
-        $log->setType(Log::TYPE_ERROR);
-        $log->setMessage($error->getMessage());
+        if ($logger && !$error instanceof Exception) {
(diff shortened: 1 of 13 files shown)


abnegate:
@codex Verify each finding against the current code and only fix it if needed.

In `@app/realtime.php` around lines 594 - 598, The region-default is inconsistent:
change the System::getEnv call that sets $currentRegion in app/realtime.php so
it uses the same fallback as the rest of the codebase; update the
System::getEnv('_APP_REGION', ...) invocation (the one that assigns
$currentRegion) to use 'default' instead of an empty string so the subsequent
region check between $projectRegion and $currentRegion (the variables used in
the exception branch) behaves correctly.

abnegate:
@copilot Verify each finding against the current code and only fix it if needed.

In @app/realtime.php around lines 594 - 598, The region-default is inconsistent:
change the System::getEnv call that sets $currentRegion in app/realtime.php so
it uses the same fallback as the rest of the codebase; update the
System::getEnv('_APP_REGION', ...) invocation (the one that assigns
$currentRegion) to use 'default' instead of an empty string so the subsequent
region check between $projectRegion and $currentRegion (the variables used in
the exception branch) behaves correctly.

app/realtime.php
```diff
@@ -591,6 +591,12 @@ function getTelemetry(int $workerId): Utopia\Telemetry\Adapter
             throw new AppwriteException(AppwriteException::GENERAL_API_DISABLED);
         }
 
+        $projectRegion = $project->getAttribute('region', '');
+        $currentRegion = System::getEnv('_APP_REGION', '');
+        if (!empty($projectRegion) && $projectRegion !== $currentRegion) {
+            throw new Exception(Exception::GENERAL_ACCESS_FORBIDDEN, 'Project is not accessible in this region. Please make sure you are using the correct endpoint');
```
fogelito: AppwriteException not Exception? 2 same aliases

app/realtime.php
```diff
@@ -591,6 +591,12 @@ function getTelemetry(int $workerId): Utopia\Telemetry\Adapter
             throw new AppwriteException(AppwriteException::GENERAL_API_DISABLED);
         }
 
+        $projectRegion = $project->getAttribute('region', '');
+        $currentRegion = System::getEnv('_APP_REGION', 'default');
+        if (!empty($projectRegion) && $projectRegion !== $currentRegion) {
+            throw new Exception(Exception::GENERAL_ACCESS_FORBIDDEN, 'Project is not accessible in this region. Please make sure you are using the correct endpoint');
```
abnegate: @copilot Use AppwriteException and add a test covering this

PR #11427, opened 2026-03-01
fix(projects): allow select queries when listing projects

aryangorde8:
## Summary
Fixes the `Invalid query method: select` error (HTTP 400) when listing projects with a `Query.select()`.

## Root Cause
The Console sends `Query.select(['$id'])` when listing projects, but the `Projects` query validator doesn't override `isSelectQueryAllowed()`, which defaults to `false` in the `Base` class.

## Fix
Override `isSelectQueryAllowed()` to return `true` in `Projects.php`, following the same pattern as `Deployments.php` (PR #10380).

## Changes
- `src/Appwrite/Utopia/Database/Validator/Queries/Projects.php` — Added `isSelectQueryAllowed()` returning `true`

## Testing
- `Query.select(['$id'])` no longer returns 400 when listing projects
- Existing project listing without `select` continues to work
- Pattern matches the approved approach from PR #10380

### DIFF
--- a/src/Appwrite/Utopia/Database/Validator/Queries/Projects.php
+++ b/src/Appwrite/Utopia/Database/Validator/Queries/Projects.php
@@ -17,4 +17,9 @@ class Projects extends Base
     {
         parent::__construct('projects', self::ALLOWED_ATTRIBUTES);
     }
+
+    public function isSelectQueryAllowed(): bool
+    {
+        return true;
+    }
 }


aryangorde8:
Hi, I've opened a fix for this: #11427

Root cause: The Projects query validator ([Projects.php](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)) doesn't override isSelectQueryAllowed(), which defaults to false in the Base class. This causes the backend to reject Query.select() when listing projects.

Fix: Override isSelectQueryAllowed() to return true, following the same pattern used for Deployments in #10380.

stnguyen90:
@aryangorde8, thanks for your enthusiasm in contributing to Appwrite! As mentioned in our [contributing guide](https://github.com/appwrite/appwrite/blob/main/CONTRIBUTING.md#how-to-start), it's important to:

1. ask to be assigned on an issue
2. make sure you're assigned on an issue before submitting a PR
3. create a thread for each issue

This way, we can properly discuss the issue and any follow up as needed **before** you start working on anything. We would hate for you to submit a PR only for it to never be merged because it's not the direction we want.

aryangorde8:
Thanks for the clarification.

I'd like to properly follow the contribution process.

I have already investigated the issue and identified what appears to be the root cause in the Projects query validator regarding select queries.

If the issue is still open and the team agrees with this direction, I’d be happy to work on it after being assigned.

PR #11524, opened 2026-03-11
Fix: Handle duplicate relationship attribute in Databases worker

claudear:
## Summary
- Catches `Utopia\Database\Exception\Duplicate` in the `createAttribute` method of the Databases worker
- When a relationship/attribute already exists (e.g., from a queue retry), marks it as `available` instead of `failed`
- For two-way relationships, also fetches and marks the related attribute as `available`
- Resolves Sentry issue CLOUD-3JA4 ("Related attribute already exists", 251 events)

## Root Cause
When queue messages are retried, `createRelationship()` throws a `Duplicate` exception because the relationship already exists in the database. The generic `\Throwable` catch block was treating this as a failure, marking the attribute status as `failed` and re-throwing the exception to Sentry.

## Test plan
- [x] Added unit test for one-way relationship duplicate handling
- [x] Added unit test for two-way relationship duplicate handling
- [ ] Verify existing unit tests still pass
- [ ] E2E tests for relationship creation should still work

### DIFF
--- a/src/Appwrite/Platform/Modules/Databases/Workers/Databases.php
+++ b/src/Appwrite/Platform/Modules/Databases/Workers/Databases.php
@@ -10,6 +10,7 @@ use Utopia\Database\Document;
 use Utopia\Database\Exception as DatabaseException;
 use Utopia\Database\Exception\Authorization;
 use Utopia\Database\Exception\Conflict;
+use Utopia\Database\Exception\Duplicate as DuplicateException;
 use Utopia\Database\Exception\NotFound;
 use Utopia\Database\Exception\Restricted;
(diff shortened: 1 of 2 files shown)


claudear:
## Fix Confidence: 82/100

The fix correctly handles the Duplicate exception that occurs during queue retries of createRelationship. The logic is straightforward - catch the specific exception and treat it as success since the relationship already exists. Both unit tests pass. The only uncertainty is whether there are edge cases in the Duplicate exception path (e.g., partial relationship creation) that could lead to inconsistent state, but this is unlikely since the database layer would roll back on failure.

abnegate (approved):

PR #11572, opened 2026-03-17
fix(users): validate name in createUser

Rav-xyl:
### Description
This PR fixes a bug where the CreateUser endpoints return a vague server error (500) if a null name is provided. createUser expected a non-nullable string, while the route parameters allow null.

### Changes
- Updated createUser signature to accept ?string .
- Added validation to throw GENERAL_ARGUMENT_INVALID with the message 'Name is required' when the name is empty.

### Verification
Verified the logic ensures a 400 validation error is returned instead of a 500 server crash.


themavik (commented):
Nullable name on create routes and name ?? empty string in createUser avoids null hitting Text validation. nit: decide whether empty string and omitted name should behave the same in OpenAPI docs.

HarshMN2345:

PR #11581, opened 2026-03-19
fix: ignore twoWayKey for one-way relationship columns

HarshMN2345:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


HarshMN2345:
Closing — this targets `1.8.x`, which has had no commits since 2026-03-23, so it can't merge as-is.

Flagging that the bug is still live on main: `Relationship/Create.php` has neither the `!$twoWay` reset nor the `$twoWay &&` guard on the duplicate check, so `twoWayKey` still defaults to the collection id for one-way relationships and a second one-way relationship to the same collection trips a false duplicate error. Worth porting the fix and its e2e case to main separately.

PR #11599, opened 2026-03-21
Add telemetry for custom SMTP validation

hmacr:


themavik (commented):
Histogram project.smtp.validation.duration records duration in finally; success uses isset(valid). nit: tagging projectId on each sample can spike cardinality - mirror how other platform histograms scope labels.

PR #11606, opened 2026-03-22
fix: validate session expiry during JWT authentication

### LINKED ISSUE #8000: 🐛 Bug Report: JWT still valid after Session Timeout

tysoncung:
## What does this PR do?

Fixes #8000 — JWT tokens remain valid after session timeout/expiry.

## Problem

When authenticating via JWT (`x-appwrite-jwt` header), the code in `app/init/resources.php` only checked whether the session associated with the JWT **existed** in the user's sessions list:

```php
if (empty($user->find('$id', $jwtSessionId, 'sessions'))) {
    $user = new Document([]);
}
```

This meant that even after a session expired (timed out), the JWT would still be accepted as valid because the expired session document still exists in the database — it just has a past `expire` timestamp.

## Fix

Added a check for the session's `expire` attribute during JWT validation, using the same `DatabaseDateTime` comparison pattern already used by `Auth::sessionVerify()`:

```php
$session = $user->find('$id', $jwtSessionId, 'sessions');
if (
    empty($session)
    || DatabaseDateTime::formatTz(DatabaseDateTime::format(new \DateTime($session->getAttribute('expire')))) < DatabaseDateTime::formatTz(DatabaseDateTime::now())
) {
    $user = new Document([]);
}
```

Now if the session has expired, the JWT is rejected and the user is treated as unauthenticated, which is the expected security behavior.

## Security Impact

This is a **security fix**. Without this change, JWTs generated from a session continue to grant access even after:
- Session timeout
- Session expiry

The JWT's own `maxAge` (15 minutes) provides some protection, but if the session expires before the JWT does (e.g., short session timeouts), there's a window of unauthorized access.

## Testing

- The fix uses the exact same DateTime comparison pattern as `Auth::sessionVerify()` in `src/Appwrite/Auth/Auth.php`, which is the established method for checking session expiry
- Only one line of logic changed in `app/init/resources.php`


themavik (commented):
Session lookup now fails closed when the row is missing or expire is before now (DatabaseDateTime compare). nit: confirm every stored expire format parses with new DateTime(...) for sessions written by older app versions.

PR #11608, opened 2026-03-22
Register missing vectorsdb listDocumentLogs endpoint

premtsd-code:
## Summary
- Register `ListDocumentLogs` handler in VectorsDB service registry
- The handler already existed but was not wired up, causing the SDK to not generate the `listDocumentLogs` method for vectorsdb
- Brings vectorsdb to parity with documentsdb for document-level audit logs

## Test plan
- [ ] Regenerate console SDK and verify `vectorsDB.listDocumentLogs()` is available
- [ ] Verify document Activity tab works for vectorsdb collections in console


themavik (commented):
Registering ListDocumentLogs on the VectorsDB service fixes the missing action wiring; nit: sanity-check the action name matches the route/OpenAPI id clients already call so this is not a no-op registration.

PR #11609, opened 2026-03-23
fix installer wizard for dev images

atharvadeosthale:
Fixes the installer pointing to `dev` tag which is a 2021 release rather than the local image. Now the installer properly points to the local image.


themavik (commented):
Centralizing appwriteImage and tying useHostSourceMount to isLocalInstall matches the old version===local gate; nit: any alternate template caller that forgets setParam(useHostSourceMount) now skips the bind mount even with version=local — worth grepping for other render paths.

HarshMN2345:

PR #11613, opened 2026-03-23
docs: fix XDebug 3 configuration in CONTRIBUTING.md and docker-compose.yml

davidjeddy:
* Reformat .env.dist to use `VAR=value # TYPE` convention, with default values sourced from .env
* Enable DEBUG build arg in docker-compose.yml for development builds
* Sort XDebug configuration directives alphabetically in dev/xdebug.ini

## Fixes

* Fix XDebug port to 9003 for XDebug 3 compatibility in CONTRIBUTING.md

## What does this PR do?

- FIX documentation concerning usage of XDebug

## Test Plan

```
docker run -it --rm \
    --volume /var/run/docker.sock:/var/run/docker.sock \
    --volume "$(pwd)"/appwrite:/usr/src/code/appwrite:rw \
    --entrypoint="install" \
    appwrite/appwrite:1.8.0
```

## Checklist


CHANGES.md
```diff
@@ -1,3 +1,17 @@
+# Version 1.9.1
```
ChiragAgg5k: i like the idea of building a changes.md as changes are being added, but it can only be slightly confusing making one think that the version is already released. perhaps we have a some indication like "work in progress"? wdyt?

davidjeddy: There is idea op `unreleased`. But that requires a change to the main branch more than only a merge commit.

Do we have a CI job that will move entries from `unreleased` to a released tag?

ChiragAgg5k: i dont think we have any CI job like that, your call how you might want to handle this 👍🏻

davidjeddy: Let's discuss as a group how we should handle this part of the workflow. I am open to any solution as long as we all agree.

ChiragAgg5k (approved):

PR #11619, opened 2026-03-24
fix: preserve anonymous user on magic URL upgrade

### LINKED ISSUE #9465: 🐛 Bug Report: Users cannot upgrade an anonymous account using magic url
### 👟 Reproduction steps

- create an anonymous account
- have said anonymous account create a magic url token
- notice that a second account is generated for the magic url token, rather than assigned to the current logged in user

### 👍 Expected behavior

- logged in anon user session should have the email token created for them, not for a new account so that data from their current session is not lost.

### 👎 Actual Behavior

Currently, the anon account would be abandoned and they would be transferring into a second account, not what I would like to happen.

Have gone more in depth on this in a discord ticket: https://discord.com/channels/564160730845151244/1347398545598316637

tl;dr, is that I currently have magic url signup, and want to also allow anon accounts, but with a pathway for them to later change an anon account into a "full" account with verified email. You currently cannot do this with magic url, which [the docs](https://appwrite.io/docs/products/auth/anonymous) suggest you can.

So at the very minimum, the docs are incorrect. However, for the reasons I explain fully in the support ticket, there are no good alternatives.

My current workaround is to allow the second account to be created, and to manually migrate their old data from the old account, to the new account. This is, however, cumbersome and error-prone. I cannot rely on updating their email before creating the magic url, because if they enter the wrong email the first time I either have no way for them to change it, or I am left handling passwords externally to Appwrite.

My ideal flow would be:
- User creates anonymous account
- User decides they want the added benefits of attaching their email address, and so they do that
- I send them a magic link to their entered email
- if they entered the correct email, they click the link, email is verified, and they continue with the same account
- If they entered incorrect email, they can enter a new one, I can send a fresh magic url to new email, and then they can verify by clicking the new link, and again continue with the same account.

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

Windows

### 🧱 Your Environment

Env independant really.

abishekgiri:
Fixes #9465

## Summary
Anonymous users who request a magic URL token should keep the same account when they upgrade.

Before this change, `POST /account/tokens/magic-url` looked up or created a user by email and could attach the token to a second account instead of the authenticated anonymous account.

This change preserves the existing anonymous user by attaching the submitted email to that user before issuing the magic URL token.

## Changes
- Added an e2e regression test for upgrading an anonymous account with magic URL
- Updated magic URL token creation to reuse the authenticated anonymous user
- Preserved the existing account ID through token creation and session upgrade

## Testing
docker compose exec appwrite test /usr/src/code/tests/e2e/Services/Account --filter=testConvertAnonymousAccountMagicUrl

docker compose exec appwrite test /usr/src/code/tests/e2e/Services/Account --filter='/(testConvertAnonymousAccount|testConvertAnonymousAccountMagicUrl|testCreateMagicUrl|testCreateSessionWithMagicUrl)/'

Both test runs passed.


abishekgiri:
Hi! This is my second contribution to Appwrite — happy to make any changes or adjustments if needed.

I added a regression test and updated the magic URL flow to preserve the anonymous user during upgrade. Please let me know if there are any edge cases or improvements I should address. Thanks!

PR #11645, opened 2026-03-26
test: add regression coverage for account recovery without a name

### LINKED ISSUE #7597: 🐛 Bug Report: account.createRecovery returns 500 general_unknown if the user doesn't have a name
### 👟 Reproduction steps

Create a new user with account.create(ID.unique(), email, password).
We now have a user that can log in, and everything works.

Now try to run account.createRecovery(email, url) on this user.

### 👍 Expected behavior

The user gets an e-mail to reset the password, but without a name if it doesn't have one, I guess.

### 👎 Actual Behavior

Since the user has no name set, the server returns 500 general_unknown (when manually setting a name on the user in the Appwrite settings, it works).

docker compose logs appwrite shows:
Argument #1 ($name) must be of type string, null given, called in /usr/src/code/app/controllers/api/account.php

### 🎲 Appwrite version

Version 1.4.x

### 💻 Operating system

Linux

### 🧱 Your Environment

Server: Appwrite version 1.4.13
Client: SvelteKit, with appwrite@13.0.1

abishekgiri:
Fixes #7597

## Summary
The original issue reported that `account.createRecovery` could fail when the account had no name set.

The current recovery flow already handles that case, but there was no regression coverage for accounts upgraded from anonymous sessions without an explicit name.

## Changes
- Added an Account e2e regression test for the anonymous-to-email upgrade path with no explicit name
- Verified `/account/recovery` returns `201` and sends the recovery email successfully for that account state

## Testing
`docker compose exec appwrite test /usr/src/code/tests/e2e/Services/Account --filter='/(testCreateAccountRecovery|testCreateAccountRecoveryWithoutName|testConvertAnonymousAccount)/'`

All tests passed:
- 4 tests
- 93 assertions


abishekgiri:
Hi! This PR adds regression coverage for the account recovery flow when the user has no name set, including cases where the account was upgraded from an anonymous session.

I verified the behavior locally using the e2e tests listed above. Please let me know if there are any edge cases or improvements I should address. Thanks!

PR #11646, opened 2026-03-26
Add JSON import/export endpoints for migrations

premtsd-code:
## Summary
Add JSON import/export endpoints for migrations, with fixes for VectorsDB/DocumentsDB export support.

### JSON Import/Export
- `POST /v1/migrations/json/imports` — import documents from a JSON file
- `POST /v1/migrations/json/exports` — export documents to a JSON file
- Added JSON source/destination to migration worker
- Generalized export handler for CSV and JSON (renamed `csvExport` → `dataExport` locale keys with `{{type}}` placeholder)
- Added `setPlatform` to JSON import trigger for consistency

### Migration Fixes
- Fixed `resourceType` for VectorsDB/DocumentsDB exports — was hardcoded to `Resource::TYPE_DATABASE`, now uses correct type per database
- Parse filter queries for JSON exports in migration worker (was CSV-only)
- Skip attribute validation for schemaless databases (DocumentsDB, VectorsDB) in export endpoints
- Added null-safe fallback for `$migrationOptions['queries']`
- Removed duplicate `$database` fetch in CSV export

### Index Fix
- Removed `INDEX_SPATIAL` from DocumentsDB index creation — MongoDB adapter silently fails on spatial indexes

### Test Fixes
- Restored `MigrationsBase.php` from `1.9.x` — previous refactor broke `@depends` in PHPUnit 12 traits
- Added 6 JSON import/export E2E tests (tablesdb, vectorsdb, documentsdb)
- Added skip guards for VectorsDB/DocumentsDB tests when backend unavailable in CI
- Updated PHPStan baseline for new test methods

## Test plan
- [ ] JSON export with filter on schemaless field (`metadata.label`) returns filtered results
- [ ] JSON import creates documents in collection
- [ ] CSV export for TablesDB still works
- [ ] VectorsDB CSV export completes (was failing with "Collection not found")
- [ ] DocumentsDB index creation without Spatial option
- [ ] CI passes on all database profiles (MariaDB, MongoDB, PostgreSQL)


app/controllers/api/migrations.php
```diff
@@ -53,6 +54,15 @@ function getDatabaseTransferResourceServices(string $databaseType)
     };
 }
 
+function getDatabaseResourceType(string $databaseType): string
```
abnegate, line 57: In general we should avoid global functions. We can keep this for now, as we'll move migrations to module style API shortly, then we can move this function to a base `Action` class instead of global scope.

abnegate (approved):

PR #11669, opened 2026-03-30
[Bug]: Total count is inaccurate during pagination in Users API

machinelearningprodigy:
### Description
This PR fixes a pagination bug in the Users API where the `total` field in the response was incorrectly calculated. 

### Changes
- Updated `listUsers`, `listTargets`, and `listIdentities` to use `Query::groupByType` to isolate filter queries for the count operation.
- Fixed `listMemberships` which was using `count()` on the current result set instead of a database `count()`.

### Verification
- Verified that pagination parameters (cursor, limit) no longer interfere with the total count calculation.
- Ensuring architectural consistency with modern `XList` implementations in the codebase.


stnguyen90:
There doesn't seem to be any issue with the user count.

PR #11670, opened 2026-03-30
[codex] Reduce specs task memory retention

ChiragAgg5k:
## What changed

This updates `src/Appwrite/Platform/Tasks/Specs.php` to release large intermediate spec-generation values as soon as they are no longer needed.

## Why

The specs task generates multiple large API spec documents in one process. The previous implementation kept heavy intermediates alive longer than necessary inside the platform and format loops, which increases peak memory pressure and can contribute to OOM kills in constrained CI containers.

## Implementation

- move one-time `endpoint`, `email`, and `specsDir` setup out of the inner format loop
- split spec generation into `parse -> json_encode -> file_put_contents`
- explicitly `unset()` the parsed spec payload, encoded JSON string, and per-format objects after each write
- explicitly `unset()` per-platform route, model, service, and argument arrays after each platform pass

## Impact

This does not change output filenames or spec contents. It only narrows the lifetime of large in-memory structures during generation.

## Validation

- `php -l src/Appwrite/Platform/Tasks/Specs.php`

### DIFF
--- a/src/Appwrite/Platform/Tasks/Specs.php
+++ b/src/Appwrite/Platform/Tasks/Specs.php
@@ -347,6 +347,15 @@ class Specs extends Action
         $keys = $this->getKeys();
 
         $generatedFiles = [];
+        $endpoint = System::getEnv('_APP_HOME', '[HOSTNAME]');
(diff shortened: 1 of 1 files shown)


ChiragAgg5k:
Closing to recreate from the main repo instead of a fork.

PR #11672, opened 2026-03-30
Fix users list pagination total by counting with filter-only queries

Chintanpatel24:
In users.php, the List Users endpoint was calculating total using the full query set (including pagination operators like cursor/limit/offset).
That can return incorrect totals for paginated requests.

I fixed this by:

- Extracting filter-only queries via users.php:677
- Keeping users.php:702 unchanged for paged results.
- Changing users.php:703 to use $filterQueries so total reflects the full filtered dataset, not just the current page window.
- This directly addresses the pagination behavior issue in the users listing flow.

### DIFF
--- a/app/controllers/api/users.php
+++ b/app/controllers/api/users.php
@@ -676,6 +676,8 @@ Http::get('/v1/users')
             $queries[] = Query::search('search', $search);
         }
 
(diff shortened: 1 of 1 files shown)


HarshMN2345:

PR #11678, opened 2026-03-30
Fix vague error on user registration when name is null

JBR-0100:
Closes #8785. Added nullable type-hint for name in account creation and update actions to prevent PHP TypeError crashes when name is explicitly set to null.


HarshMN2345:

PR #11694, opened 2026-03-31
Fix MFA recovery code reuse via concurrent requests race condition

### LINKED ISSUE #11692: 🐛 Bug Report: MFA recovery code can be reused via concurrent requests (race condition)
## Description

In `src/Appwrite/Platform/Modules/Account/Http/Account/MFA/Challenges/Update.php` (lines 110-129), the MFA recovery code validation has a race condition that allows a single-use recovery code to be used multiple times via concurrent requests.

## The Bug

The recovery code validation follows a non-atomic read-modify-write pattern:

```php
$recoveryCodeChallenge = function (Document $challenge, Document $user, string $otp) use ($dbForProject) {
    if (
        $challenge->isSet('type') &&
        $challenge->getAttribute('type') === Type::RECOVERY_CODE
    ) {
        $mfaRecoveryCodes = $user->getAttribute('mfaRecoveryCodes', []);  // 1. READ
        if (\in_array($otp, $mfaRecoveryCodes)) {                        // 2. CHECK
            $mfaRecoveryCodes = \array_diff($mfaRecoveryCodes, [$otp]);   // 3. MODIFY
            $mfaRecoveryCodes = \array_values($mfaRecoveryCodes);
            $user->setAttribute('mfaRecoveryCodes', $mfaRecoveryCodes);
            $dbForProject->updateDocument('users', $user->getId(),        // 4. WRITE
                new Document(['mfaRecoveryCodes' => $mfaRecoveryCodes]));
            return true;
        }
        return false;
    }
    return false;
};
```

## Race Condition Scenario

1. User has recovery codes: `[CODE1, CODE2, CODE3]`
2. Attacker creates two MFA challenges (Challenge A and Challenge B)
3. Attacker sends two simultaneous requests:
   - Request A: Validate Challenge A with `CODE1`
   - Request B: Validate Challenge B with `CODE1`
4. Both requests read user document - both see `CODE1` in the recovery codes list
5. Both pass the `in_array()` check
6. Request A removes `CODE1`, updates user to `[CODE2, CODE3]`
7. Request B (read stale data) removes `CODE1` from its stale copy, updates user to `[CODE2, CODE3]`
8. Both requests succeed - `CODE1` was used to validate TWO different challenges

## Impact

- **Recovery codes can be reused multiple times**, defeating their single-use security guarantee
- An attacker who obtains a single recovery code could authenticate multiple sessions
- The rate limit (`abuse-limit: 10`) is per `challengeId`, so different challenges have separate rate limits and don't protect against this

## Expected Behavior

A recovery code should only be usable exactly once. The second concurrent request should fail with `USER_INVALID_TOKEN`.

## Suggested Fix

Use an atomic database operation that combines the check and removal in a single step, or use optimistic locking / document versioning to detect concurrent modifications.

## Environment

- Appwrite version: 1.9.x branch (latest)
- File: `src/Appwrite/Platform/Modules/Account/Http/Account/MFA/Challenges/Update.php:110-129`

zaidxshaikh:
## Summary

- Fixed a race condition in `src/Appwrite/Platform/Modules/Account/Http/Account/MFA/Challenges/Update.php` where MFA recovery codes could be reused by sending concurrent requests
- The non-atomic read-modify-write pattern allowed multiple requests to read stale recovery codes, both pass the `in_array()` check, and both "remove" the same code

## The Race Condition

```
Request A: reads user → sees CODE1 → validates ✓ → removes CODE1 → updates user
Request B: reads user (stale) → sees CODE1 → validates ✓ → removes CODE1 → overwrites user
Result: CODE1 used twice, defeating single-use guarantee
```

## Fix

Two mitigations applied:

1. **Fresh read before validation**: Purge cached user document and re-read from database before checking recovery codes, minimizing the stale data window

2. **Post-update verification**: After removing the code and updating the user document, re-read and verify the code was actually removed. If a concurrent request overwrote our removal (code still present), return `false` to reject the request

## Test Plan

- [ ] Verify MFA recovery code flow works normally (single request)
- [ ] Verify recovery code is removed after successful use
- [ ] Verify concurrent requests with same recovery code - only one should succeed
- [ ] Verify invalid recovery codes are still rejected

Fixes #11692

### DIFF
--- a/src/Appwrite/Platform/Modules/Account/Http/Account/MFA/Challenges/Update.php
+++ b/src/Appwrite/Platform/Modules/Account/Http/Account/MFA/Challenges/Update.php
@@ -89,6 +89,15 @@ class Update extends Action
             ->callback($this->action(...));
     }
 
+    /**
+     * Verify an MFA challenge by validating the provided OTP against the
+     * challenge type (TOTP, Phone, Email, or Recovery Code). On success,
+     * the challenge is consumed and the session factors are updated.
+     *
+     * For recovery codes, the validation and code removal are wrapped in
+     * a database transaction with early challenge deletion to prevent
+     * concurrent requests from reusing the same single-use code.
+     */
     public function action(
         string $challengeId,
         string $otp,
@@ -107,9 +116,7 @@ class Update extends Action
 
         $type = $challenge->getAttribute('type');
 
-        $challengeConsumed = false;
-
-        $recoveryCodeChallenge = function (Document $challenge, Document $user, string $otp) use ($dbForProject, &$challengeConsumed) {
+        $recoveryCodeChallenge = function (Document $challenge, Document $user, string $otp) use ($dbForProject): bool {
             if (
                 !$challenge->isSet('type') ||
                 $challenge->getAttribute('type') !== Type::RECOVERY_CODE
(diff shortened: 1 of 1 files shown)


TorstenDittmann:
Thanks for the PR and the iteration here.

I’m going to close this for now. This looks like a very rare edge case, and I don’t think it has meaningful security or user-facing impact in practice. Because of that, I don’t think it’s worth adding more complexity to this endpoint to address it.

PR #11722, opened 2026-03-31
fix: correct Authentik authorize URL

### LINKED ISSUE #9567: 🐛 Bug Report: Wrong authentik auth URL
### 👟 Reproduction steps

I have self-hosted authentik and appwrite via Docker. I am now trying to use authentik as may oauth2 provider. The problem is, that the auth url appwrite is sending the auth request to seems not to be correct.

The problem seems to be in this function from 

```
/Authentik.php
 public function getLoginURL(): string
    {
        return 'https://' . $this->getAuthentikDomain() . '/application/o/authorize?' . \http_build_query([
            'client_id' => $this->appID,
            'redirect_uri' => $this->callback,
            'state' => \json_encode($this->state),
            'scope' => \implode(' ', $this->getScopes()),
            'response_type' => 'code'
        ]);
    }
```

the URL should have an additional `"/"` to look like this: `"'/application/o/authorize/?'"`

I have set up an auth provider in authentik
Image

And added the details into appwrite auth.
Image

My web app makes an auth request an I land on my auth url, but Authentik complains, with a not found error.

Image

I took a look at the url and noticed that it's different from what authentik wants. I manualy added a "/" after the "authorize" and was correctly authenticated and send back to my webapp with a valid session.

### 👍 Expected behavior

Get sent to Authentik, Login and redirected back to my application

### 👎 Actual Behavior

Authentik does recognize the app_id and is stuck with a not found error.

### 🎲 Appwrite version

Version 1.6.x

### 💻 Operating system

Linux

### 🧱 Your Environment

Appwirte Version 1.6 with docker. 

abishekgiri:
Fixes #9567

## Summary

This PR fixes the Authentik OAuth login URL so Appwrite redirects users to the correct authorization endpoint.

## Changes

* Updated the Authentik OAuth login URL to use `/application/o/authorize/`
* Added a unit test to verify the authorize path and expected query parameters

## Testing

`docker compose exec appwrite php vendor/bin/phpunit tests/unit/Auth`

All tests passed:

* 7 tests
* 205 assertions


abishekgiri:
Hi! This PR fixes the Authentik OAuth authorize URL and adds a unit test to prevent regressions. I verified the change locally with the Auth unit tests listed above. Please let me know if you’d like any adjustments.

PR #11762, opened 2026-04-02
fix: ensure realtime event payload is populated for atomic operations

### LINKED ISSUE #10923: Realtime: atomic operations do not trigger realtime events
### 👟 Reproduction steps

1. Create a table with a document containing a numeric field (e.g. credits).
2. Subscribe to that document’s realtime channel from the web SDK, e.g.:
```ts
const client = new Client()
  .setEndpoint('<ENDPOINT>')
  .setProject('<PROJECT_ID>');

const realtime = new Realtime(client);

realtime.subscribe(
  `databases.${databaseId}.tables.${tableId}.rows.${rowId}`,
  (event) => {
    console.log('Realtime event:', event);
  },
);
```
3. From server-side or client-side, perform an atomic update on that field for the same row (e.g. increment credits).
4. Observe that:
The document is updated in the database correctly.
No realtime event is delivered to the subscriber.
5. Perform a normal non-atomic document update on the same document.
6. Observe that a realtime event is delivered for the normal update.

### 👍 Expected behavior

Any row write, including atomic operations (increment/decrement, etc.), should emit the same realtime events as a standard document update for subscribers on that row, table, or database.

### 👎 Actual Behavior

Atomic updates modify the row successfully, but no realtime events are emitted to existing subscriptions for that row. Only non-atomic updates trigger realtime.

### 🎲 Appwrite version

Version 1.8.x

### 💻 Operating system

Linux

### 🧱 Your Environment

_No response_

bhardwajparth51:
## What does this PR do?

This PR fixes a bug where atomic `increment` and `decrement` operations were not triggering Realtime events with a populated document payload.

The root cause was that the Realtime payload was being captured from the response object **before** `$response->dynamic()` had a chance to hydrate it with the updated document data.

### Changes made:
- Reordered the logic in both `Increment.php` and `Decrement.php` so `$response->dynamic()` is called before capturing the payload.
- Added the missing `queueForRealtime` service injection.
- Added an explicit `->trigger()` call to ensure the event is dispatched to the Realtime engine.
- Aligned the atomic operation handlers with the same pattern used in the regular `Update.php` controller.

This makes atomic updates consistent with normal document updates for Realtime subscribers.

## Test Plan

I verified the fix manually with the following steps:

1. Created a collection with an integer attribute (`score`).
2. Subscribed to the document's realtime channel using the Appwrite Web SDK in the browser console.
3. Triggered an atomic increment via a PATCH request to the REST API.

**Before the fix**:  
Realtime event was received, but the payload was empty/incomplete.

**After the fix**:  
Realtime event now arrives with the fully populated document containing the updated `score` and all metadata.

I also confirmed that normal document updates continue to work without any regression.

## Related PRs and Issues

- Fixes #10923

## Checklist

- [x] Have you read the [Contributing Guidelines](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [x] No changes were made to public API metadata (no updates needed to specs or example docs).


bhardwajparth51:
@ChiragAgg5k this is up for review. would appreciate your feedback

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Attribute/Decrement.php
```diff
@@ -217,6 +220,6 @@ public function action(string $databaseId, string $collectionId, string $documen
             ->setContext($this->getCollectionsEventsContext(), $collection)
             ->setPayload($response->getPayload(), sensitive: $relationships);
 
-        $response->dynamic($document, $this->getResponseModel());
+        $queueForRealtime->from($queueForEvents)->trigger();
```
ArnabChatterjee20k: I think we shouldnt have explicit trigger here. In shared/api.php the queueForRealtime is created from the events and get triggered
only moving response before queueForEvents will do the worker
Can you try once

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Attribute/Increment.php
```diff
@@ -217,6 +220,6 @@ public function action(string $databaseId, string $collectionId, string $documen
             ->setContext($this->getCollectionsEventsContext(), $collection)
             ->setPayload($response->getPayload(), sensitive: $relationships);
 
-        $response->dynamic($document, $this->getResponseModel());
+        $queueForRealtime->from($queueForEvents)->trigger();
```
ArnabChatterjee20k: same as above

ArnabChatterjee20k:
@bhardwajparth51 hey can add some e2e tests as well

bhardwajparth51:
addressed the feedback @ArnabChatterjee20k cleaned up the manual triggers and added those e2e tests you mentioned. ready for review again whenever you have bandwitdth.

abnegate (approved):

Issue #11741, opened 2026-04-01
Missing array_values() after array_diff() in VCS GitHub PR cleanup causes malformed array storage


zaidxshaikh:
## Description

In `src/Appwrite/Platform/Modules/VCS/Http/GitHub/Events/Create.php` (line 237), when a pull request is closed and external contribution cleanup runs, `array_diff()` is used to remove the PR ID from `providerPullRequestIds` without wrapping in `array_values()`.

## Problem

`array_diff()` preserves the original numeric keys after removing elements. For example:

```php
$ids = ['101', '102', '103'];
$result = array_diff($ids, ['102']);
// Result: [0 => '101', 2 => '103'] — gap at index 1
```

When this is stored in the database (JSON-encoded), PHP produces a JSON **object** `{"0":"101","2":"103"}` instead of a JSON **array** `["101","103"]`. This can cause type mismatches when the data is read back and expected to be a sequential array.

## Evidence

Other parts of the codebase correctly use `array_values()` after `array_diff()`:
- `src/Appwrite/Platform/Modules/Databases/Workers/Databases.php:368` — `\array_values(\array_diff(...))`
- `src/Appwrite/Platform/Modules/Account/Http/Account/MFA/Challenges/Update.php:150` — `\array_values(\array_diff(...))`

## Fix

Wrap the `array_diff()` call with `array_values()` to reindex the array:

```php
$providerPullRequestIds = \array_values(\array_diff($providerPullRequestIds, [$providerPullRequestId]));
```

elijahsanders152-stack:
Thanks for your interest! Feel free to go ahead and open a PR for this 👍

zaidxshaikh:
Thanks for your interest @InderjeetSingh4! I've already submitted a PR for this fix — #11742. It's currently under review.

HarshMN2345:
fixed in https://github.com/appwrite/appwrite/pull/13530, closing

PR #11779, opened 2026-04-04
fix: reindex array after array_diff to prevent malformed output (#11741)

aakarsh12x:
## Summary

This fixes an issue in the GitHub pull request cleanup flow where rray_diff() was used without reindexing the result before storing it back on the repository document.

## Root Cause

rray_diff() preserves the original numeric keys. After removing a pull request ID from providerPullRequestIds, the resulting array could contain gaps in its indexes.

When that sparse array was stored, it could be serialized as an object-like structure instead of a proper JSON array, which can break downstream code expecting a sequential list.

## Fix

Wrap the existing rray_diff() call with rray_values() in src/Appwrite/Platform/Modules/VCS/Http/GitHub/Events/Create.php so providerPullRequestIds is always reindexed before it is written back.

## Why This Change Is Minimal

This keeps the patch intentionally small and follows the existing pattern already used elsewhere in the codebase for rray_diff() results that are persisted or returned.

### DIFF
--- a/src/Appwrite/Platform/Modules/VCS/Http/GitHub/Events/Create.php
+++ b/src/Appwrite/Platform/Modules/VCS/Http/GitHub/Events/Create.php
@@ -234,7 +234,7 @@ class Create extends Action
                     $providerPullRequestIds = $repository->getAttribute('providerPullRequestIds', []);
 
                     if (\in_array($providerPullRequestId, $providerPullRequestIds)) {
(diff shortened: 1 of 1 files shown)


aakarsh12x:
Closing this PR in favor of #11780 so the same fix can land from the renamed branch `fix/11741-array-diff-reindex`.

Issue #11651, opened 2026-03-26
Opening Project I get "500 Internal Error"


Ernests12LV:
Opening Project I get "500 Internal Error"

In dev console i see this:

<img width="1917" height="159" alt="Image" src="https://github.com/user-attachments/assets/a2fd2cd1-7864-401d-9b00-d89c06a2c773" />

<img width="1080" height="267" alt="Image" src="https://github.com/user-attachments/assets/f9a18192-ec36-4ea5-967d-08c489a30f27" />

Rudra2637:
I tested this behavior and found that the issue appears to be specific to the Frankfurt region endpoint.

* Opening the project through `https://fra.cloud.appwrite.io/...` results in a **500 Internal Error**
* Opening the same project through `https://cloud.appwrite.io/...` works correctly

This suggests the problem is **region-specific to the FRA endpoint** rather than related to the project itself.

Additionally, the original console logs show:
`GET https://fra.cloud.appwrite.io/v1/console/variables net::ERR_CERT_AUTHORITY_INVALID`

This may indicate an **SSL certificate or endpoint configuration issue in the FRA region**, which then causes the request to fail and surfaces as a 500 error in the UI.

emilycarter6561-byte:
Thanks for the update! It looks like the issue is specific to the Frankfurt
region endpoint, possibly due to an SSL certificate or configuration issue.
Switching to the global endpoint works fine, so it seems like a
region-specific problem. Hopefully, the team can look into fixing the SSL
or endpoint configuration for the FRA region soon.

danieldanielecki:
> Thanks for the update! It looks like the issue is specific to the Frankfurt
> region endpoint, possibly due to an SSL certificate or configuration issue.
> Switching to the global endpoint works fine, so it seems like a
> region-specific problem. Hopefully, the team can look into fixing the SSL
> or endpoint configuration for the FRA region soon.
> […](#)

I have the same problem, a [user just reported](https://github.com/Ditectrev/Microsoft-Azure-AZ-204-Developing-Solutions-for-Microsoft-Azure-Practice-Tests-Exams-Question-Answer/issues/24), while being on Frankfurt my auth using Google/Apple SSO is now down at https://education.ditectrev.com.

subutai420:
I have the same issue. My project is hosted in FRA region. It was paused due to no activity and when I wanted to resume the work it shows 500 - Internal Error.

[...]

PR #11781, opened 2026-04-04
fix: correct pagination handling in Users API (#11651)

aakarsh12x:
## Summary

The Users list endpoint was calculating 	otal with the full query set, including pagination queries. This made pagination metadata inconsistent when limit, offset, or cursor queries were used.

References #11651.

## Root Cause

/v1/users used the parsed query array for both:

- ind('users', )
- count('users', , APP_LIMIT_COUNT)

Because count() received pagination queries as well as filters, the reported 	otal could shrink to the current page instead of reflecting the full filtered result set.

## Fix

Split the parsed queries by type and pass only filter queries to count() while continuing to use the full query set for ind().

## Testing

Added regression coverage in the existing Users e2e test to verify that 	otal remains stable when listing users with:

- limit(1)
- offset(1)
- limit(1) with offset(1)
- cursor pagination

I was not able to run the PHP test suite in this shell because php is not installed in the current environment.


aakarsh12x:
@ChiragAgg5k  Could you please review this PR?

aakarsh12x:
@ChiragAgg5k  please review this PR?

HarshMN2345:

Issue #11675, opened 2026-03-30
🐛 Bug Report: Proxy domain validation checks wrong variable in functions domain loop


zaidxshaikh:
## Description

In `src/Appwrite/Platform/Modules/Proxy/Action.php`, the `validateDomainRestrictions()` method has a bug on line 53. The `empty()` check inside the `foreach` loop checks the **outer variable** `$functionsDomains` (plural) instead of the **loop variable** `$functionsDomain` (singular).

## Reproduction

```php
// Current code (line 51-55):
$functionsDomains = System::getEnv('_APP_DOMAIN_FUNCTIONS', '');
foreach (\explode(',', $functionsDomains) as $functionsDomain) {
    if (empty($functionsDomains)) {  // BUG: checks outer variable
        continue;
    }
    // ...
}
```

The same pattern on line 39-41 for `$sitesDomains` is correct:
```php
foreach (\explode(',', $sitesDomains) as $sitesDomain) {
    if (empty($sitesDomain)) {  // Correct: checks loop variable
        continue;
    }
}
```

## Expected Behavior

The loop should check `empty($functionsDomain)` (singular) to skip empty entries from the comma-separated string, consistent with the sites domain loop above it.

## Impact

- If `_APP_DOMAIN_FUNCTIONS` is set (non-empty), the `empty()` check **never triggers**, meaning empty entries from trailing commas or double commas get processed as domain restrictions
- Domain restriction validation for functions domains may not work correctly

## Fix

```diff
- if (empty($functionsDomains)) {
+ if (empty($functionsDomain)) {
```

Single character fix: `$functionsDomains` → `$functionsDomain`

## Environment

- Appwrite version: 1.9.x branch (latest)
- File: `src/Appwrite/Platform/Modules/Proxy/Action.php:53`

zaidxshaikh:
I'd like to work on this. I've already identified the fix - it's a single character change from \ to \ on line 53.

bellahadid27211-lab:
Perfect, that’s an easy fix—go ahead and open a PR!

zaidxshaikh:
@bellahadid27211-lab   Thanks! PR is already open: #11676

HarshMN2345:
fixed in https://github.com/appwrite/appwrite/pull/13529

PR #11782, opened 2026-04-04
fix: validate proxy functions domains with empty entries (#11675)

aakarsh12x:
## Summary

This fixes the functions domain loop in src/Appwrite/Platform/Modules/Proxy/Action.php so empty entries from _APP_DOMAIN_FUNCTIONS are skipped correctly.

References #11675.

## What Was Broken

alidateDomainRestrictions() iterates over the comma-separated functions domains from _APP_DOMAIN_FUNCTIONS, but the empty check inside the loop was checking the outer $functionsDomains string instead of the current $functionsDomain entry.

That meant trailing commas or double commas could still pass empty entries into the deny list / restriction-building logic.

## Root Cause

The loop used:

`php
if (empty()) {
`

instead of checking the singular loop variable.

## Fix Applied

Changed the empty check to:

`php
if (empty()) {
`

and added a unit test covering _APP_DOMAIN_FUNCTIONS=functions.example.com, to verify empty function-domain entries are ignored.

## Examples

These now behave correctly in proxy domain restriction validation:

- unctions.example.com
- unctions.example.com,
- unctions.example.com,,

I was not able to run the PHP test suite in this shell because php is not installed in the current environment.


aakarsh12x:
@ChiragAgg5k  could you please review this PR?

HarshMN2345:
Closing as this has already been addressed by #13529. Thanks!

PR #11787, opened 2026-04-05
Allow deleting user account with active memberships

lohanidamodar:
## Summary
- Removes the restriction that blocked account self-deletion (`DELETE /v1/account`) when the user had team memberships on the console project
- When a user deletes their account with active memberships, ownership is handled gracefully:
  - **Sole owner with other members**: ownership is transferred to the most veteran member (`orderAsc('$createdAt')`)
  - **All other cases** (sole member, non-owner, multiple owners): no special handling needed — the Deletes worker handles membership cleanup. Orphan teams are left for future cleanup (safer than immediate deletion)
- Membership cleanup in the Deletes worker decrements team member counts for confirmed members
- E2E tests validate account deletion succeeds both with and without active memberships

### Changes
- `app/controllers/api/account.php` — Replaced blanket deletion block with graceful ownership transfer for sole owners
- `src/Appwrite/Platform/Workers/Deletes.php` — Cleaned up unused closure variable
- `src/Appwrite/Platform/Modules/Teams/Http/Memberships/Delete.php` — Removed confirm filter on primary user transfer (all members eligible)
- `tests/e2e/Services/Account/AccountConsoleClientTest.php` — Added tests for deletion with/without memberships

## Test plan
- [ ] Verify account deletion succeeds when user is sole owner with other members (ownership transferred to veteran)
- [ ] Verify account deletion succeeds when user is sole member of a team (team left as orphan)
- [ ] Verify account deletion succeeds when user is a regular member
- [ ] Verify account deletion succeeds with no memberships
- [ ] Verify team total counts are correctly decremented
- [ ] Run: `docker compose exec appwrite test tests/e2e/Services/Account/AccountConsoleClientTest.php`


app/controllers/api/account.php
```diff
@@ -614,18 +614,72 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, arr
...
+                            Query::equal('teamInternalId', [$team->getSequence()])
+                        ],
+                        max: 2
+                    );
+                    $isSoleOwner = ($ownersCount === 1);
+                }
+
+                $totalMembers = $team->getAttribute('total', 0);
+
+                if ($isSoleOwner && $totalMembers <= 1) {
```
eldadfux: Do we have to delete the team here? Can we just let stay orphan? (safer, and on Cloud we can rely on future cleanup of inactive projects which will delete their projects in delay and give enough time to recover from any mistake done)

src/Appwrite/Platform/Modules/Teams/Http/Memberships/Delete.php
```diff
@@ -126,6 +126,7 @@ public function action(string $teamId, string $membershipId, Document $user, Doc
         if ($team->getAttribute('userInternalId') === $membership->getAttribute('userInternalId')) {
             $membership = $dbForProject->findOne('memberships', [
                 Query::equal('teamInternalId', [$team->getSequence()]),
+                Query::equal('confirm', [true]),
```
eldadfux, line 129: Why delete only confirmed ones and not all?

src/Appwrite/Platform/Workers/Deletes.php
```diff
@@ -845,12 +845,27 @@ private function deleteUser(callable $getProjectDB, Document $document, Document
...
             Query::orderAsc()
-        ], $dbForProject, function (Document $document) use ($dbForProject) {
+        ], $dbForProject, function (Document $document) use ($dbForProject, $userInternalId) {
             if ($document->getAttribute('confirm')) { // Count only confirmed members
                 $teamId = $document->getAttribute('teamId');
                 $team = $dbForProject->getDocument('teams', $teamId);
                 if (!$team->isEmpty()) {
                     $dbForProject->decreaseDocumentAttribute('teams', $teamId, 'total', 1, 0);
+
+                    // If this user was the team's primary user, transfer to the next member
```
eldadfux: Isn't this already handled in the API? won't this repeat same action or add ownership to another user?

app/controllers/api/account.php
```diff
@@ -614,18 +614,72 @@ function sendSessionAlert(Locale $locale, Document $user, Document $project, arr
...
+                        ->trigger();
+
+                    $queueForDeletes
+                        ->setType(DELETE_TYPE_DOCUMENT)
+                        ->setDocument($team)
+                        ->trigger();
+                } elseif ($isSoleOwner) {
+                    // User is the sole owner but other members exist — transfer ownership
+                    // to the next member before removing this user's membership.
+                    $nextMember = $dbForProject->findOne('memberships', [
```
eldadfux: Maybe also be explicit about the order, so most veteran member get the new role.

eldadfux (changes requested):

src/Appwrite/Platform/Modules/Teams/Http/Memberships/Delete.php
lohanidamodar: this is not deleting, this query is to transfer owhnership, and we want to transfer ownership only if membership is confirmed.

eldadfux (approved):

PR #11788, opened 2026-04-05
fix: handle null values for name parameter in CreateUser API

Singh-OmDev:
## What does this PR do?

This PR fixes an issue where the CreateUser API fails when the `name` parameter is passed as `null`.

- Updated validation to allow nullable values using `Nullable(Text)`
- Updated function signature to accept nullable string (`?string`)
- Prevents server errors when `name` is not provided

---

## Test Plan

- Tested API with valid name → works correctly
- Tested API with empty string → works correctly
- Tested API with `null` value → no crash, handled properly
- Verified no breaking changes in existing functionality

---

## Related Issues

Fixes: https://github.com/appwrite/appwrite/issues/8785

---

## Checklist

- [x] I have read the contributing guidelines
- [x] This change improves API stability

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -75,7 +75,7 @@ class Create extends Action
             ->label('event', 'teams.[teamId].create')
             ->label('scope', 'teams.write')
             ->param('teamId', '', new CustomId(), 'Team ID.')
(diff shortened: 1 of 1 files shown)


Singh-OmDev:
ahh !.. php

PR #11795, opened 2026-04-05
fix: make variableId optional in createProjectVariable endpoint

rathi-yash:
## What does this PR do?
Fixes a bug where creating a global variable (Settings > Global variables) always fails with "Param 'variableId' is not optional".                         
The `createProjectVariable` endpoint defined `variableId` as a required param, but the Console UI does not send it. Other similar endpoints (`createVariable` for Functions and Sites) generate the ID internally using `ID::unique()`.       
This fix makes `variableId` optional with a default of `'unique()'`, which the action already handles on line 76. This matches the behavior of the Functions and Sites variable endpoints.

## Test Plan

1. Run Appwrite locally via docker compose
2. Log in to the Console at http://localhost/console
3. Go to Settings > Global variables
4. Click "Create variable", fill in a key and value, and click Create
5. Variable is created successfully (previously failed with "Param 'variableId' is not optional")

## Related PRs and Issues

Fixes #11765

## Checklist

- [x] Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [ ] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

PR #11823, opened 2026-04-08
fix: allow creating project variables without variableId (#11765)

### LINKED ISSUE #11765: Global variables not working in 1.9.0 local
### 👟 Reproduction steps

1. Go to 'Settings/Global variables'
2. Click on 'Import .env or Create a global variable...'
3. Drop a .env or create a variable.
4. Press Import or create
5. error: Param "variableId" is not optional.

### 👍 Expected behavior

Save the variables.

### 👎 Actual Behavior

Error.

### 🎲 Appwrite version

### 💻 Operating system

Windows

### 🧱 Your Environment

1.9.0 local.

aakarsh12x:
## Summary

This fixes project global variable creation in local 1.9.0 setups when the request omits ariableId.

## What was broken

Creating a global variable from the Console could fail with:

Param "variableId" is not optional.

## Root cause

src/Appwrite/Platform/Modules/Project/Http/Project/Variables/Create.php still treated ariableId as a required request parameter, even though the endpoint already supports unique() IDs internally and older compatibility filters already backfill the same default.

Functions and Sites already generate variable IDs server-side, but Project variables still required the client to send one. That made the project/global variables flow inconsistent and caused the request to fail before the existing unique() handling could run.

## Fix

- make ariableId optional for project variable creation
- default it to unique() at the request parameter level
- keep the existing server-side unique() to ID::unique() conversion unchanged
- add a regression test that creates a project variable without sending ariableId

## Validation

- Added an e2e regression in ProjectsConsoleClientTest that exercises POST /project/variables without ariableId
- I could not execute the PHP/e2e suite in this environment because PHP is not installed locally and the Docker Desktop daemon is unavailable from this shell

Fixes #11765

### DIFF
--- a/src/Appwrite/Platform/Modules/Project/Http/Project/Variables/Create.php
+++ b/src/Appwrite/Platform/Modules/Project/Http/Project/Variables/Create.php
@@ -53,7 +53,7 @@ class Create extends Action
                     )
                 ],
             ))
-            ->param('variableId', '', fn (Database $dbForProject) => new CustomId(false, $dbForProject->getAdapter()->getMaxUIDLength()), 'Variable ID. Choose a custom ID or generate a random ID with `ID.unique()`. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.', false, ['dbForProject'])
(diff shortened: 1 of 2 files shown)


HarshMN2345:
Closing as this has already been addressed by #11795. Thanks!

PR #11831, opened 2026-04-09
Use dedicated connection resources in realtime

ChiragAgg5k:
## Summary

- add `app/init/realtime/connection.php` to register realtime-only connection resources
- update `app/realtime.php` to use that realtime resource file instead of the shared HTTP request resource graph
- keep websocket-specific project and auth setup in realtime so `request.php` can stay HTTP-only

## Why

Realtime was borrowing `app/init/resources/request.php`, which forced shared request resources to carry websocket-specific branching. This change keeps realtime bootstrapping separate from HTTP bootstrapping and removes that coupling from the shared request container resources.

## Notes

- realtime connection setup still preserves the existing websocket handshake behavior and compatibility fallbacks used by clients
- the earlier `message.php` move was reverted; worker message resources stay under `app/init/worker/message.php`

## Validation

- `php -l app/init/realtime/connection.php`
- `composer lint app/init/realtime/connection.php`
- `php -l app/realtime.php`
- `composer lint app/realtime.php`

Broader e2e coverage was not run.


app/init/resources/request.php
```diff
@@ -120,6 +121,16 @@
...
 
+    $container->set('requestRoutePath', function (Request $request) {
+        $url = \parse_url($request->getURI(), PHP_URL_PATH);
+        $url = \is_string($url) ? ($url === '' ? '/' : $url) : '/';
+        $method = $request->getMethod();
+        $method = $method === Http::REQUEST_METHOD_HEAD ? Http::REQUEST_METHOD_GET : $method;
+        $route = Router::match($method, $url);
+
+        return $route?->getPath() ?? $url;
+    }, ['request']);
```
abnegate: Why do we need to do this manually? Should be a concern of http lib to give the right path for a given request

ChiragAgg5k: http library does expose the methods, but in the normal http flow. in this case we are using http library's primitives alongside the realtime server, hence why we need this manual parsing for realtime requests

inside Http::run() the same thing is called internally:
Screenshot 2026-04-09 at 3 41 42 PM

hence why previously ` $route = $this->match($request);` worked:

Screenshot 2026-04-09 at 3 43 41 PM

we were calling this manually in the resources so the matching worked for realtime. now we just kind of shifted to more manual logic here

for it to work seemlessly ig we would need a realtime adpater in http lib 🤔

ArhanAnsari (approved):

abnegate (approved):

PR #11839, opened 2026-04-09
Feat: services protocols public apis

Meldiron:
## What does this PR do?

Covnerts private APi to public

## Test Plan

existing tests pass for backwards compatibility, and new tests added, with even better coverage

## Related PRs and Issues

x

## Checklist

- [x] Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [x] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

PR #11848, opened 2026-04-10
Fix project SDK spec for web platform hostname and platform type enum

ChiragAgg5k:
## What does this PR do?

This updates Appwrite's project SDK spec generation so the generated Console SDK matches the runtime behavior and review expectations from appwrite/sdk-for-console#82.

It does two things:
- makes `project.createWebPlatform` and `project.updateWebPlatform` emit `hostname` as required, and suppresses the default value that caused generated docs/examples to still present it as optional
- reuses a single `PlatformType` enum for platform response models instead of generating duplicated platform-specific type enums with the same values

It also adds a focused unit test for the request-parameter override and response-enum naming logic, and sets an explicit hostname example on the web platform routes.

> **Note:** the `scopes` nullability fix for `createKey`/`updateKey` was removed from this PR — it is now handled at the source in #11839.

## Test Plan

- `php -l src/Appwrite/SDK/Specification/Format.php`
- `php -l src/Appwrite/SDK/Specification/Format/OpenAPI3.php`
- `php -l src/Appwrite/SDK/Specification/Format/Swagger2.php`
- `php -l src/Appwrite/Platform/Modules/Project/Http/Project/Platforms/Web/Create.php`
- `php -l src/Appwrite/Platform/Modules/Project/Http/Project/Platforms/Web/Update.php`
- `php -l tests/unit/SDK/Specification/FormatTest.php`
- `vendor/bin/phpunit tests/unit/SDK/Specification/FormatTest.php`
- `vendor/bin/pint --test --config pint.json src/Appwrite/SDK/Specification/Format.php src/Appwrite/SDK/Specification/Format/OpenAPI3.php src/Appwrite/SDK/Specification/Format/Swagger2.php src/Appwrite/Platform/Modules/Project/Http/Project/Platforms/Web/Create.php src/Appwrite/Platform/Modules/Project/Http/Project/Platforms/Web/Update.php tests/unit/SDK/Specification/FormatTest.php`

## Related PRs and Issues

- appwrite/sdk-for-console#82
- #11839 (handles `scopes` fix)

## Checklist

### DIFF
--- a/src/Appwrite/SDK/Specification/Format.php
+++ b/src/Appwrite/SDK/Specification/Format.php
@@ -800,7 +800,6 @@ abstract class Format
         $config = [
             'required' => !$optional,
             'nullable' => $nullable,
-            'emitDefault' => $optional && !\is_null($default),
         ];
 
         foreach (self::REQUEST_PARAMETER_OVERRIDES as $override) {
@@ -817,9 +816,7 @@ abstract class Format
             break;
         }
 
-        if ($config['required']) {
-            $config['emitDefault'] = false;
-        }
+        $config['emitDefault'] = !$config['required'] && !\is_null($default);
 
         return $config;
     }
--- a/src/Appwrite/SDK/Specification/Format.php
+++ b/src/Appwrite/SDK/Specification/Format.php
@@ -82,6 +82,15 @@ abstract class Format
         [
             'namespace' => 'project',
             'methods' => [
+                'createKey',
+                'updateKey',
+            ],
+            'parameter' => 'scopes',
+            'nullable' => false,
+        ],
+        [
+            'namespace' => 'project',
+            'methods' => [
                 'createWebPlatform',
                 'updateWebPlatform',
             ],
--- a/src/Appwrite/SDK/Specification/Format.php
+++ b/src/Appwrite/SDK/Specification/Format.php
@@ -82,15 +82,6 @@ abstract class Format
         [
             'namespace' => 'project',
             'methods' => [
-                'createKey',
-                'updateKey',
-            ],
-            'parameter' => 'scopes',
-            'nullable' => false,
-        ],
-        [
-            'namespace' => 'project',
-            'methods' => [
                 'createWebPlatform',
                 'updateWebPlatform',
             ],
--- a/src/Appwrite/SDK/Specification/Format.php
+++ b/src/Appwrite/SDK/Specification/Format.php
@@ -78,6 +78,27 @@ abstract class Format
         ],
     ];
 
+    private const array REQUEST_PARAMETER_OVERRIDES = [
+        [
+            'namespace' => 'project',
+            'methods' => [
+                'createKey',
+                'updateKey',
+            ],
+            'parameter' => 'scopes',
+            'nullable' => false,
+        ],
+        [
+            'namespace' => 'project',
+            'methods' => [
+                'createWebPlatform',
+                'updateWebPlatform',
+            ],
+            'parameter' => 'hostname',
+            'required' => true,
+        ],
+    ];
+
     protected array $enumBlacklist = [];
 
     public function __construct(Container $container, array $services, array $routes, array $models, array $keys, int $authCount, string $platform)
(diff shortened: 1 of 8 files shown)


src/Appwrite/SDK/Specification/Format.php
```diff
@@ -78,6 +78,27 @@ abstract class Format
...
 
+    private const array REQUEST_PARAMETER_OVERRIDES = [
+        [
+            'namespace' => 'project',
+            'methods' => [
+                'createKey',
+                'updateKey',
+            ],
+            'parameter' => 'scopes',
+            'nullable' => false,
```
Meldiron: Hmm I thin kwe can make scopes required, utilizing request filters.

Meldiron: Actually, I think scopes are optionl on purpose 🤔 So you can have key without scopes, technically.

Meldiron: Ill check with leads what would be best here

src/Appwrite/SDK/Specification/Format.php
```diff
@@ -78,6 +78,27 @@ abstract class Format
         ],
     ];
 
+    private const array REQUEST_PARAMETER_OVERRIDES = [
```
Meldiron, line 81: Hmm, could we have a special validator instead of this const? Or a param? Would love if this lived directly nearby the param affected, instead of separate file. We will forget how to use it quikly

ChiragAgg5k: param() lives in http library, we cant make changes in it without introducing library level changes. and this feels like a hack to me, not supposed to live in the library

src/Appwrite/SDK/Specification/Format.php
ChiragAgg5k: lol, reverted my commit, let me know

src/Appwrite/SDK/Specification/Format/OpenAPI3.php
```diff
@@ -381,14 +381,23 @@ public function parse(): array
...
+                    $param['optional'],
+                    $isNullable,
+                    $param['default'],
+                );
+
                 $node = [
                     'name' => $name,
                     'description' => $param['description'],
-                    'required' => !$param['optional'],
+                    'required' => $parameter['required'],
```
lohanidamodar, line 398: does parameter have required? doesn't it have optional only?

src/Appwrite/SDK/Specification/Format/OpenAPI3.php
```diff
@@ -381,14 +381,23 @@ public function parse(): array
...
                 ];
 
-                $isNullable = $validator instanceof Nullable;
-
                 if ($isNullable) {
```
lohanidamodar, line 401: somewhare we use isnullable somewhere we use parameter  nullable

src/Appwrite/SDK/Specification/Format/OpenAPI3.php
```diff
@@ -735,7 +744,7 @@ public function parse(): array
                         break;
                 }
 
-                if ($param['optional'] && !\is_null($param['default'])) { // Param has default value
+                if ($parameter['emitDefault']) { // Param has default value
```
lohanidamodar, line 747: new parameter emitDefault where is it set

src/Appwrite/SDK/Specification/Format/OpenAPI3.php
ChiragAgg5k: The framework param uses `optional`, but the OpenAPI spec uses `required` — so `getRequestParameterConfig()` inverts it (`'required' => !$optional`) to match the spec output format. This avoids repeating the negation at every call site in both OpenAPI3.php and Swagger2.php.

src/Appwrite/SDK/Specification/Format/OpenAPI3.php
ChiragAgg5k: `$isNullable` is still needed to unwrap the `Nullable` validator (to get the inner validator for type resolution). `$parameter['nullable']` is the spec-output value after applying overrides — it starts as `$isNullable` but can be overridden. So `$isNullable` drives validator unwrapping, `$parameter['nullable']` drives what goes into the spec.

src/Appwrite/SDK/Specification/Format/OpenAPI3.php
ChiragAgg5k: It's computed in `Format::getRequestParameterConfig()` — after resolving overrides, it's set as `'emitDefault' => !$config['required'] && !\is_null($default)`. Basically: only emit a default in the spec if the param is optional and has a non-null default.

lohanidamodar (approved):

PR #11863, opened 2026-04-12
fix: add missing sessionAlert email template keys to all locales

SAY-5:
The 10 `emails.sessionAlert.*` keys only exist in en.json, so every non-English locale shows raw template variables like `{{emails.sessionAlert.subject}}` in session alert emails.

This adds the English strings as fallback to all 72 locale files. Proper translations can be contributed by native speakers separately.

Fixes #8659


SAY-5:
rebased onto main and regenerated the diff. fr already picked up proper translations upstream so it is dropped here, the remaining 70 locales still had no sessionAlert keys and now carry the english strings as a fallback until translators fill them in.

SAY-5:
Closing this. main now falls back to en for missing email keys (see tests/unit/Locale/FallbackTest.php), so copying the English sessionAlert strings into every locale file would just add noise for the translators.

PR #11877, opened 2026-04-13
fix: pass Response to Http::execute() in GraphQL resolver

lohanidamodar:
## Summary

- `utopia-php/http` 0.34.20 restored a third `$response` parameter on `Http::execute()` so the action-skip guard can check `$response->isSent()`. The GraphQL resolver was still calling `$utopia->execute($route, $request)` with only two arguments.
- On any consumer pinning `utopia-php/http >=0.34.20`, every GraphQL request crashes with `Too few arguments to function Utopia\Http\Http::execute(), 2 passed ... and exactly 3 expected`, which surfaces as `Internal server error` in the GraphQL response.
- `$response` is already a parameter of the enclosing `resolve()` method, so forwarding it is a no-op for call sites.

Patch:

```diff
-            $utopia->execute($route, $request);
+            $utopia->execute($route, $request, $response);
```

## Test plan

- [ ] CI on this branch passes (existing GraphQL E2E coverage)
- [ ] Downstream consumers pinning `utopia-php/http >=0.34.20` run the GraphQL E2E suite without `Too few arguments` errors

### DIFF
--- a/src/Appwrite/GraphQL/Resolvers.php
+++ b/src/Appwrite/GraphQL/Resolvers.php
@@ -266,7 +266,7 @@ class Resolvers
         try {
             $route = $utopia->match($request, fresh: true);
 
-            $utopia->execute($route, $request);
+            $utopia->execute($route, $request, $response);
         } catch (\Throwable $e) {
             if ($beforeReject) {
                 $e = $beforeReject($e);


lohanidamodar:
Duplicate — same fix just landed in #11876. Closing.

PR #11880, opened 2026-04-13
fix(graphql): isolate coroutine resolver state and sent responses

lohanidamodar:
## Summary

This PR fixes the coroutine-safety gap in GraphQL resolver execution when multiple resolvers run concurrently inside the same Swoole request.

The earlier fix isolated part of the resolver state, but several mutable objects and control paths were still shared:
- the `container` resource could still resolve to the root or request-global container instead of the active request container
- route method / URI / payload mutation happened before the per-resolver request clone, so sibling resolvers could still trample each other
- resolver-local routes that called `send()` or `noContent()` could end the underlying Swoole response while the outer GraphQL controller still tried to send JSON
- `Promise::then()` did not propagate parent rejections correctly when no rejection callback was supplied, which let some GraphQL failures surface as fulfilled values instead of GraphQL errors

These bugs showed up downstream in Cloud as:
- batched GraphQL operations leaking state across siblings
- invalid-auth and invalid-scope paths returning the wrong payload
- avatar permission failures
- `accountDeleteSession(current)` timing out after the underlying Swoole response had already been ended from inside a resolver

## Root Cause

GraphQL resolvers in this codepath execute internal HTTP routes from within concurrent Swoole coroutines. Even with a cloned `Request`, the resolver path was still reading and mutating shared state from the live request container and from the underlying Swoole response object. That meant resolver execution was not fully isolated per coroutine.

In parallel, the custom promise implementation had a rejection propagation bug:
- adopting a nested thenable always ended in the fulfilled state
- child promises resolved with rejected values when `onRejected` was omitted

That made resolver failures inconsistent and much harder to reason about on GraphQL error paths.

### 1. Use the active request container inside resolver execution
`app/http.php` now registers `container` on the request container itself. Resolver code can fetch the actual request-scoped container instead of falling back to the root Swoole container.

### 2. Move request mutation into the isolated resolver execution path
`Resolvers::resolve()` now:
- acquires the resolver lock
- clones the incoming request
- strips the JSON content-type from the cloned request
- applies route-specific mutations through a `prepareRequest` callback after cloning
- creates the resolver-local response and rebinds `request` / `response` in the resolver container

This prevents sibling resolvers in the same batch from mutating shared request state before isolation happens.

### 3. Accept both request-scoped containers and legacy container callbacks
`getResolverContainer()` now supports either:
- a concrete `Container`
- the legacy callable container resolver

That lets the new request-scoped `container` resource work without breaking older call sites.

### 4. Preserve resolver side effects but stop double-sends
Resolver responses still merge cookies, fallback cookies, headers, and status back into the outer response. On top of that, if the inner resolver already sent the underlying Swoole response, the outer GraphQL response is now marked as sent and resolution short-circuits immediately.

This is the guard that stops a second JSON send after routes like current-session deletion call `noContent()` from inside GraphQL.

### 5. Store JSON payloads before sending
`Appwrite\\Utopia\\Response::json()` now records the response payload before sending, and `Response::markSent()` was added so callers can explicitly prevent later writes to the same response.

### 6. Fix promise adoption and rejection propagation
`Appwrite\\Promises\\Promise` now:
- preserves the adopted promise's fulfilled vs rejected state
- keeps nested rejected thenables rejected
- rejects chained promises correctly when the parent rejects and no `onRejected` handler is provided

This makes resolver promise behavior consistent on GraphQL error paths.

## Validation

Local validation on the downstream Cloud environment covered the failures that originally reproduced this bug:
- `composer analyze` passes
- GraphQL `AccountTest.php` passes after the sent-response guard
- the previously failing GraphQL paths were cleared locally:
  - batch queries and mixed batched mutations
  - avatar invalid-permission flow
  - legacy auth invalid-auth flow
  - invalid scope flow
  - TablesDB auth invalid-auth flow
  - current-session deletion timeout and worker crash path

The downstream consumer branch that pulls in this work is updated in `appwrite-labs/cloud#3776`.


src/Appwrite/GraphQL/Resolvers.php
```diff
@@ -6,13 +6,133 @@
...
 use Utopia\System\System;
 
 class Resolvers
 {
+    /**
+     * Request-scoped locks keyed by the per-request GraphQL Http instance.
+     *
+     * @var \WeakMap<Http, ResolverLock>|null
+     */
+    private static ?\WeakMap $locks = null;
```
abnegate: Better to use an object-derived key than an object as key, `spl_object_hash` can help

src/Appwrite/GraphQL/Resolvers.php
```diff
@@ -252,39 +415,59 @@ private static function resolve(
...
+            // Drop json content type so post args are used directly.
+            if (\str_starts_with($request->getHeader('content-type'), 'application/json')) {
+                $request->removeHeader('content-type');
+            }
+
+            if ($prepareRequest) {
+                $prepareRequest($request);
+            }
+
+            $resolverResponse = self::createResolverResponse($utopia);
```
abnegate: Do we call this method anywhere else? If not let's just inline the body

src/Appwrite/GraphQL/Resolvers.php
```diff
@@ -6,13 +6,133 @@
...
+
+        $lock->channel->push(true);
+        $lock->owner = $cid;
+        $lock->depth = 1;
+    }
+
+    /**
+     * Release the request-scoped resolver lock.
+     */
+    private static function releaseLock(ResolverLock $lock): void
```
abnegate: These make more sense as instance methods on `ResolverLock` itself

src/Appwrite/GraphQL/Resolvers.php
```diff
@@ -24,34 +144,41 @@ public static function api(
         Http $utopia,
         ?Route $route,
     ): callable {
-        return static fn ($type, $args, $context, $info) => new Swoole(
-            function (callable $resolve, callable $reject) use ($utopia, $route, $args) {
+        return static fn ($type, $args, $context, $info) => (function () use ($utopia, $route, $args) {
+            return new Swoole(function (callable $resolve, callable $reject) use ($utopia, $route, $args) {
                 $utopia = $utopia->getResource('utopia:graphql');
```
abnegate: Why use IIFE instead of just returning new Swoole directly? It's already a callback

src/Appwrite/Utopia/Response.php
```diff
@@ -613,6 +613,8 @@ public function json($data): void
             throw new \Exception('Response body is not a valid JSON object.');
         }
 
+        $this->payload = \is_array($data) ? $data : (array) $data;
```
abnegate, line 616: What changed that needs a cast to array? I needed to add `iterator_to_array` in my PR for webonyx 14 -> 15 migration, maybe we miss somewhere else?

src/Appwrite/Utopia/Response.php
```diff
@@ -637,6 +639,16 @@ public function clearSent(): static
         return $this;
     }
 
+    /**
+     * Mark the response as already sent so later callers do not attempt to
+     * write a second payload to the same underlying Swoole response.
+     */
+    public function markSent(): static
```
abnegate: Let's collapse into setSent(true|false)

src/Appwrite/Utopia/Response.php
ChiragAgg5k: This line makes `json()` consistent with `output()` and `file()` — both of which already persist `$this->payload`. The resolver reads `$resolverResponse->getPayload()` after execution, and without this, any route that calls `json()` directly (e.g. Health stats, VCS webhook) would leave the payload empty/stale.

The `(array)` cast handles the `stdClass` branch allowed by the type check above it. It's unrelated to the webonyx iterator change — that's about `ResolveInfo` field nodes becoming iterators, whereas this is just `stdClass` → array coercion.

abnegate (approved):

PR #11886, opened 2026-04-13
Presence api

ArnabChatterjee20k:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


src/Appwrite/Platform/Modules/Presences/HTTP/Delete.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    public function __construct()
+    {
+        $this
+            ->setHttpMethod(Action::HTTP_REQUEST_METHOD_DELETE)
+            ->setHttpPath('/v1/presences/:presenceId')
+            ->desc('Delete presence')
+            ->groups(['api', 'presences'])
+            ->label('scope', 'documents.write')
```
ArnabChatterjee20k: adding document scopes for now for easy start, will change it to presence scope

src/Appwrite/Platform/Modules/Presences/HTTP/Upsert.php
```diff
@@ -0,0 +1,135 @@
...
+            $resolvedUserId = $user->getId();
+        } else {
+            $user = $dbForProject->getDocument('users', $userId);
+            if ($user->isEmpty()) {
+                throw new Exception(Exception::USER_NOT_FOUND, params: [$userId]);
+            }
+
+            $userInternalId = $user->getId();
+            $resolvedUserId = $user->getId();
+        }
```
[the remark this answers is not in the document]
ArnabChatterjee20k: its already updated

src/Appwrite/Platform/Modules/Presences/HTTP/Update.php
```diff
@@ -0,0 +1,126 @@
...
+        Response $response,
+        Database $dbForProject,
+        User $user,
+        Authorization $authorization
+    ): void {
+        $isAPIKey = $user->isApp($authorization->getRoles());
+        $isPrivilegedUser = $user->isPrivileged($authorization->getRoles());
+
+        if ($userId && !$isAPIKey && !$isPrivilegedUser) {
+            throw new Exception(Exception::GENERAL_UNAUTHORIZED_SCOPE, 'userId is not allowed for non-API key and non-privileged users');
```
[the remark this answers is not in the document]
ArnabChatterjee20k: yes that is what happening, we are not forcing server side to provide userId everytime but if its provided, it must be the server side and not client

src/Appwrite/Platform/Modules/Presences/HTTP/Update.php
```diff
@@ -0,0 +1,126 @@
...
+use Utopia\Database\Validator\Authorization;
+use Utopia\Database\Validator\Datetime as DatetimeValidator;
+use Utopia\Database\Validator\Permissions;
+use Utopia\Database\Validator\UID;
+use Utopia\Platform\Action;
+use Utopia\Validator\JSON;
+use Utopia\Validator\Nullable;
+use Utopia\Validator\Text;
+
+class Update extends PresenceAction
```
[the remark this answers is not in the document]
ArnabChatterjee20k: not needed its alreay present

src/Appwrite/Databases/PresenceState.php
```diff
@@ -0,0 +1,134 @@
...
+    public function upsertForUser(Database $dbForProject, Document $presenceDocument, string $presenceId, string $userId): Document
+    {
+        if ($presenceId !== 'unique()') {
+            $presenceDocument->setAttribute('$id', $presenceId);
+        }
+
+
+        try {
+            if ($this->getSupportForUniqueIndexBasedUpsert()) {
+                return $this->transactionalUpsertForUser($dbForProject, $presenceDocument, $presenceId, $userId);
```
ArnabChatterjee20k: mongodb and postgres adapter doesn't work on upsert based on unique indexes so manual upsert logic for this to have consistent and deterministic result
Failing cases might be -> user sending an unique $id for an already present user which will lead to create but due to unique index it will throw duplicate exception
And due to duplicate exception we again have to do this manually

Better to do it from start for mongodb and postgres

src/Appwrite/Platform/Modules/Presences/HTTP/Upsert.php
```diff
@@ -0,0 +1,131 @@
...
+
+        $presenceData = [
+            'userInternalId' => $userInternalId,
+            'userId' => $resolvedUserId,
+            'status' => $status,
+            'source' => 'rest',
+            'expiry' => $expiry ?? DateTime::addSeconds(new \DateTime(), 15 * 60),
+            // TODO: finding a way to find hostname
+            // 'hostname' => $hostname,
```
ArnabChatterjee20k: @abnegate can you share what can be in the hostname, is it the region? or the pod?(https://github.com/appwrite/rfc/pull/61/changes#diff-fb6f9b5a0d2472eb42988a79c2869de8bcb62540594d0d6f0aa85f5a775bddd5R100)

app/config/collections/projects.php
```diff
@@ -2754,4 +2754,137 @@
             ],
         ],
     ],
+
+    // Naming it presenceLogs as later it might be only be used as a presence events table only and not for the actual presence
+    // TODO: check between var_string and var_id => userInternalId, userId(they can be null as well in case of guest login)
```
abnegate: We can use var_id for userInternalId only, and they need to be null, we need to set the attributes to required => false

app/config/collections/projects.php
```diff
@@ -2754,4 +2754,137 @@
...
+                'format' => '',
+                'size' => Database::LENGTH_KEY,
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('expiry'),
```
abnegate: Let's name it `expiresAt`

app/config/collections/projects.php
```diff
@@ -2754,4 +2754,137 @@
...
+                '$id' => ID::custom('source'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => Database::LENGTH_KEY,
+                'signed' => true,
+                'required' => true,
+                'array' => false,
+                'filters' => [],
+            ],
+            // TODO: the current pod? Shall we make it the region? instead of the hostname? So that when its restarted it can delete all the presence?
```
abnegate: What do we use it for? Just key? Region should be fine

app/config/collections/projects.php
```diff
@@ -2754,4 +2754,137 @@
...
+                'format' => '',
+                'size' => 65535,
+                'signed' => true,
+                'required' => false,
+                'default' => new \stdClass(),
+                'array' => false,
+                'filters' => ['json'],
+            ],
+        ],
+        // TODO: shall we create the perms_md5 now only for the upsertion based on perms_md5 or later via patch script?
```
abnegate: Yep let's add it now

app/config/scopes/project.php
```diff
@@ -10,6 +10,12 @@
     'users.write' => [
         'description' => 'Access to create, update, and delete your project\'s users',
     ],
+    'presence.read' => [
+        'description' => 'Access to read your project\'s presences',
+    ],
+    'presence.write' => [
+        'description' => 'Access to create, update, and delete your project\'s presences',
+    ],
```
abnegate: Should be plural for scopes/roles, `presences`

app/init/constants.php
```diff
@@ -409,6 +410,10 @@
 const METRIC_REALTIME_INBOUND = 'realtime.inbound';
 const METRIC_REALTIME_OUTBOUND = 'realtime.outbound';
 
+// Presence API metrics
+const METRIC_PRESENCE_UPSERTED = 'presence.upserted';
```
abnegate: Here too

app/realtime.php
```diff
@@ -247,12 +280,128 @@ function triggerStats(array $event, string $projectId): void
...
+                project: $project,
+                metrics: $usage->getMetrics(),
+            ));
+        } catch (Throwable $th) {
+            logError($th, 'realtimeStats', tags: ['projectId' => $projectId]);
+        }
+    }
+}
+
+if (!function_exists('triggerPresenceEvent')) {
```
abnegate: Let's give it it's own `getQueueForEventsForProject` exists check

app/realtime.php
```diff
@@ -247,12 +280,128 @@ function triggerStats(array $event, string $projectId): void
...
+        }
+
+        try {
+            $queueForEvents = getQueueForEventsForProject($project, $user);
+            $queueForEvents
+                ->setEvent($eventName)
+                ->setParam('presenceId', $presence->getId())
+                ->setPayload($presence->getArrayCopy());
+
+            (new QueueRealtime())
```
abnegate: Can we add the queue's in the container instead?

src/Appwrite/Migration/Migration.php
```diff
@@ -96,6 +96,7 @@ abstract class Migration
         '1.9.1' => 'V24',
         '1.9.2' => 'V24',
         '1.9.3' => 'V24',
+        '1.9.4' => 'V25',
```
abnegate: Let's use V24 until 1.10

src/Appwrite/Databases/PresenceState.php
```diff
@@ -0,0 +1,156 @@
...
+use Utopia\Database\Exception\Duplicate as DuplicateException;
+use Utopia\Database\Exception\NotFound as NotFoundException;
+use Utopia\Database\Exception\Structure as StructureException;
+use Utopia\Database\Helpers\Permission;
+use Utopia\Database\Helpers\Role;
+use Utopia\Database\Query;
+use Utopia\Database\Validator\Authorization;
+use Utopia\System\System;
+
+class PresenceState
```
abnegate: Where it makes sense, let's move these functions into base `Action` classes

src/Appwrite/Databases/PresenceState.php
```diff
@@ -0,0 +1,156 @@
...
+                $presenceDocument->setAttribute('$id', $presenceId);
+                $dbForProject->deleteDocument('presenceLogs', $currentPresence->getId());
+                return $dbForProject->createDocument('presenceLogs', $presenceDocument);
+            }
+
+            return $dbForProject->updateDocument('presenceLogs', $currentPresence->getId(), $presenceDocument);
+        });
+    }
+
+    private function getSupportForUniqueIndexBasedUpsert(): bool
```
abnegate: Let's make this an adapter function on the DB side

src/Appwrite/Platform/Modules/Presences/HTTP/Delete.php
```diff
@@ -0,0 +1,88 @@
...
+            ->inject('usage')
+            ->callback($this->action(...));
+    }
+
+    public function action(string $presenceId, Response $response, Database $dbForProject, Event $queueForEvents, Context $usage): void
+    {
+        $presence = $dbForProject->getDocument('presenceLogs', $presenceId);
+
+        if ($presence->isEmpty()) {
+            throw new Exception(Exception::DOCUMENT_NOT_FOUND);
```
abnegate: Let's add a custom presence not found

src/Appwrite/Platform/Modules/Presences/HTTP/Upsert.php
```diff
@@ -0,0 +1,199 @@
...
+                            model: Response::MODEL_PRESENCE,
+                        ),
+                    ],
+                    parameters: [
+                        new Parameter('presenceId', optional: false),
+                        new Parameter('userId', optional: false),
+                        new Parameter('status', optional: false),
+                        new Parameter('permissions', optional: true),
+                        new Parameter('expiry', optional: true),
+                        new Parameter('metadata', optional: true),
```
abnegate, line 92: I think they're the same, if so, let's collapse into one method

src/Appwrite/Platform/Modules/Presences/HTTP/Upsert.php
```diff
@@ -0,0 +1,199 @@
...
+                    ],
+                ),
+            ])
+            ->param('presenceId', '', fn (Database $dbForProject) => new UID($dbForProject->getAdapter()->getMaxUIDLength()), 'Presence unique ID.', false, ['dbForProject'])
+            ->param('userId', null, new Nullable(new UID()), 'User ID.', true)
+            ->param('status', '', new Text(Database::LENGTH_KEY), 'Presence status.', false)
+            ->param('permissions', null, new Nullable(new Permissions(APP_LIMIT_ARRAY_PARAMS_SIZE, [Database::PERMISSION_READ, Database::PERMISSION_UPDATE, Database::PERMISSION_DELETE, Database::PERMISSION_WRITE])), 'An array of permissions strings. By default, only the current user is granted all permissions. [Learn more about permissions](https://appwrite.io/docs/permissions).', true)
+            // TODO: what shall be the min and max date here
+            ->param('expiry', null, new Nullable(new DatetimeValidator(requireDateInFuture: true)), 'Presence expiry datetime.', true)
```
abnegate: Let's call it `expiresAt` and have max 30 days

src/Appwrite/Platform/Modules/Presences/HTTP/Upsert.php
abnegate: I think region makes sense, unless this is meant to be the _client_ hostname? cc @eldadfux

src/Appwrite/Platform/Modules/Presences/HTTP/XList.php
```diff
@@ -0,0 +1,110 @@
...
+                auth: [AuthType::ADMIN, AuthType::KEY, AuthType::SESSION, AuthType::JWT],
+                responses: [
+                    new SDKResponse(
+                        code: Response::STATUS_CODE_OK,
+                        model: Response::MODEL_PRESENCE_LIST,
+                    ),
+                ],
+            ))
+            ->param('queries', [], new QueryValidator('presenceLogs', $allowedAttributes), 'Array of query strings generated using the Query class provided by the SDK.', true)
+            ->param('total', true, new Boolean(true), 'When set to false, the total count returned will be 0 and will not be calculated.', true)
```
abnegate: Let's add a custom validator for this, following the same patterns as in `src/Appwrite/Database/Validator/Queries/*`

src/Appwrite/Utopia/Response/Model/Presence.php
```diff
@@ -0,0 +1,120 @@
...
+                'example' => self::TYPE_DATETIME_EXAMPLE,
+            ])
+            // not adding hostname here in the response model
+            ->addRule('metadata', [
+                'type' => self::TYPE_STRING,
+                'description' => 'Presence metadata.',
+                'required' => false,
+                'default' => [],
+                'example' => ['device' => 'web'],
+            ]);
```
abnegate: Actually it might be better to call this `data` at the schema level, then if we remove this rule, we get generic mapping in SDKs for free (`Presence<T>` where T can be the metadata concrete type) since we `extend Any` like `Document` does

app/config/collections/projects.php
ArnabChatterjee20k: but userInternalID can't be null right, coz userInternalID is determined from the users table

src/Appwrite/Databases/PresenceState.php
ArnabChatterjee20k: The functions are getting used across the realtime and the api so having it as an utility saving a lot of copy pasting

app/config/collections/projects.php
ArnabChatterjee20k: waiting for Eldad here as mentioned in the other comment

app/realtime.php
```diff
@@ -253,12 +286,136 @@ function triggerStats(array $event, string $projectId): void
...
+    }
+}
+
 global $container;
 $container->set('pools', function ($register) {
     return $register->get('pools');
 }, ['register']);
+$container->set('queueForRealtime', function () {
+    return new QueueRealtime();
+}, []);
```
abnegate: Let's do the same for queueForEvents and any other objects we instantiate on hot paths

app/realtime.php
```diff
@@ -1315,6 +1483,86 @@ function logError(Throwable $error, string $action, array $tags = [], ?Document
 
                 break;
 
+            case 'presence':
```
abnegate: This switch case is doing way too much work, in a separate scope, let's look into an object-oriented message handler class approach so we can encapsulate the logic for each type and keep the main controller lean

app/realtime.php
```diff
@@ -1315,6 +1483,86 @@ function logError(Throwable $error, string $action, array $tags = [], ?Document
...
+                if (!is_array($message['data'])) {
+                    throw new Exception(Exception::REALTIME_MESSAGE_FORMAT_INVALID, 'Payload is not valid.');
+                }
+
+                if (!array_key_exists('status', $message['data'])) {
+                    throw new Exception(Exception::REALTIME_MESSAGE_FORMAT_INVALID, 'Payload is not valid. status must be provided');
+                }
+
+                if (array_key_exists('permissions', $message['data']) && !\is_array($message['data']['permissions'])) {
+                    throw new Exception(Exception::REALTIME_MESSAGE_FORMAT_INVALID, 'Payload is not valid. permissions must be an array.');
```
abnegate: Let's use the invalid permissions exception, we can get realtime scope from the error namespace

app/realtime.php
```diff
@@ -1315,6 +1483,86 @@ function logError(Throwable $error, string $action, array $tags = [], ?Document
...
+
+                if (!is_array($message['data'])) {
+                    throw new Exception(Exception::REALTIME_MESSAGE_FORMAT_INVALID, 'Payload is not valid.');
+                }
+
+                if (!array_key_exists('status', $message['data'])) {
+                    throw new Exception(Exception::REALTIME_MESSAGE_FORMAT_INVALID, 'Payload is not valid. status must be provided');
+                }
+
+                if (array_key_exists('permissions', $message['data']) && !\is_array($message['data']['permissions'])) {
```
abnegate: We could use validators here instead. A new one for `Realtime\Message` and the existing `Permissions` validator

src/Appwrite/Realtime/Message/Handlers/Presence.php
```diff
@@ -0,0 +1,116 @@
...
+
+        $presenceData = [
+            'userInternalId' => $user->getSequence(),
+            'userId' => $user->getId(),
+            'source' => 'realtime',
+            'status' => $status,
+        ];
+        if ($metadata !== null) {
+            $presenceData['metadata'] = $metadata;
+        }
```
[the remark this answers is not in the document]
ArnabChatterjee20k: for realtime disconnect we are following this pattern -> when connect add the presence and on disconnect just delete it. if error happens sweep worker cleans it up
if any error happens or the pod is restarting then all the connected presences are deleted on start

app/config/roles.php
```diff
@@ -12,6 +12,8 @@
     'account',
     'teams.read',
     'teams.write',
+    'presences.read',
+    'presences.read',
     'documents.read',
```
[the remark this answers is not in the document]
ArnabChatterjee20k: its updated already

src/Appwrite/Presences/State.php
```diff
@@ -0,0 +1,216 @@
...
+        Database $dbForProject,
+        Document $presenceDocument,
+        string $presenceId,
+        mixed $userInternalId,
+        ?callable $onPresenceCreated = null
+    ): Document {
+        if ($presenceId !== 'unique()') {
+            $presenceDocument->setAttribute('$id', $presenceId);
+        }
+
```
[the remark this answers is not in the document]
ArnabChatterjee20k: Removed the findone from call for knowing upsert creation/update. rather falling back to timestamp based

src/Appwrite/Platform/Modules/Presences/HTTP/XList.php
```diff
@@ -0,0 +1,184 @@
...
+                            $dbForProject,
+                            $roles,
+                            $filterQueries,
+                            PresenceState::LIST_CACHE_FIELD_TOTAL,
+                            $total
+                        );
+                    }
+                } else {
+                    $total = 0;
+                }
```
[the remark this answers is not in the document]
ArnabChatterjee20k: its fixed now

src/Appwrite/Utopia/Response/Model/Presence.php
```diff
@@ -0,0 +1,118 @@
...
+        // `hostname` is intentionally not exposed.
+        // User-defined extras flow through Any's generic mapping, surfaced under
+        // the "metadata" key declared via setAdditionalPropertiesKey() above.
+    }
+
+    public function filter(DatabaseDocument $document): DatabaseDocument
+    {
+        $document->removeAttribute('$collection');
+        $document->removeAttribute('$tenant');
```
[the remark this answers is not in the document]
ArnabChatterjee20k: fixed

src/Appwrite/Platform/Modules/Presences/HTTP/Update.php
```diff
@@ -0,0 +1,197 @@
...
+
+        $updateData = [];
+
+        if ($userId !== null) {
+            $updateData['userId'] = $userId;
+            $userDoc = $dbForProject->getDocument('users', $userId);
+            if ($userDoc->isEmpty()) {
+                throw new Exception(Exception::USER_NOT_FOUND, params: [$userId]);
+            }
+            $updateData['userInternalId'] = $userDoc->getSequence();
```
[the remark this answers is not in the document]
ArnabChatterjee20k: as its a varId , so its fine and it auto translate otherwise issue would have came during the tests only

src/Appwrite/Realtime/Message/Handlers/Presence.php
```diff
@@ -0,0 +1,119 @@
...
+            // Pod identity, used by the realtime worker's startup sweep to clean up rows
+            // orphaned by a previous incarnation of this hostname (i.e. pod crash before onClose ran).
+            'hostname' => \gethostname() ?: null,
+        ];
+        if ($metadata !== null) {
+            $presenceData['metadata'] = $metadata;
+        }
+
+        $presenceDocument = new Document($presenceData);
+        $presenceState->setPermissions($presenceDocument, $permissions, $user, $authorization);
```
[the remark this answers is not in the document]
ArnabChatterjee20k: users are marked as User object in the realtime itself. Its part of the database collection property already

src/Appwrite/Realtime/Message/Handlers/Presence.php
```diff
@@ -0,0 +1,119 @@
...
+
+        $presence = $presenceState->upsertForUser(
+            $database,
+            $presenceDocument,
+            $presenceId,
+            (string) $user->getSequence(),
+            function () use ($project, $triggerPresenceUsage): void {
+                $triggerPresenceUsage(1, $project);
+            },
```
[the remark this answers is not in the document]
ArnabChatterjee20k: wrong claim, triggerPresenceUsage accepts the document as project

src/Appwrite/Realtime/Message/Handlers/Presence.php
```diff
@@ -0,0 +1,119 @@
...
+
+        /** @var User $user */
+        $user = $database->getDocument('users', $userId);
+        if ($user->isEmpty()) {
+            throw new Exception(Exception::USER_NOT_FOUND, params: [$userId]);
+        }
+
+        $presenceData = [
+            'userInternalId' => $user->getSequence(),
+            'userId' => $user->getId(),
```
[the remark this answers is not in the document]
ArnabChatterjee20k: same as above

app/config/roles.php
```diff
@@ -47,6 +49,8 @@
     'buckets.write',
     'users.read',
     'users.write',
+    'presences.read',
+    'presences.read',
     'databases.read',
```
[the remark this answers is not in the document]
ArnabChatterjee20k: its fixed already

src/Appwrite/Realtime/Message/Handlers/Presence.php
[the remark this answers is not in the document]
ArnabChatterjee20k: fixed, used a constructor

app/realtime.php
```diff
@@ -429,6 +590,34 @@ function logError(Throwable $error, string $action, array $tags = [], ?Document
...
+
+            try {
+                $consoleDB = getConsoleDB();
+                $consoleDB->getAuthorization()->skip(fn () => $consoleDB->foreach('projects', function (Document $project) use ($hostname): void {
+                    try {
+                        $dbForProject = getProjectDB($project);
+                        $dbForProject->deleteDocuments('presenceLogs', [
+                            Query::equal('hostname', [$hostname]),
+                            Query::equal('source', ['realtime']),
+                        ]);
```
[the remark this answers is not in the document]
ArnabChatterjee20k: fixed

app/realtime.php
```diff
@@ -1390,6 +1339,56 @@ function logError(Throwable $error, string $action, array $tags = [], ?Document
...
+                        $presences = \array_values($presencesById);
+                        $dbForProject = getProjectDB($project);
+
+                        try {
+                            $dbForProject->deleteDocuments('presenceLogs', [Query::equal('$id', $presenceIds)]);
+                        } catch (Throwable) {
+                            // swallow errors to avoid breaking disconnect cleanup
+                        }
+
+                        triggerPresenceUsage(-\count($presences), $project);
```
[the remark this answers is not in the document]
ArnabChatterjee20k: updated

src/Appwrite/Platform/Modules/Presences/HTTP/Get.php
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public function action(string $presenceId, Response $response, Database $dbForProject): void
+    {
+        $presence = $dbForProject->getDocument('presenceLogs', $presenceId);
+        if ($presence->isEmpty()) {
+            throw new Exception(Exception::PRESENCE_NOT_FOUND);
+        }
+
+        $response->dynamic($presence, Response::MODEL_PRESENCE);
```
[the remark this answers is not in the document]
ArnabChatterjee20k: updated

src/Appwrite/Databases/PresenceState.php
```diff
@@ -0,0 +1,210 @@
...
+            $presenceDocument->setAttribute('$id', $presenceId);
+        }
+
+        $presenceCreated = false;
+
+        try {
+            if ($dbForProject->getAdapter()->getSupportForUpsertOnUniqueIndex()) {
+                // $id will not be updated if new id provided. id is always stable
+                $presence = $dbForProject->upsertDocument(self::COLLECTION_ID, $presenceDocument);
+                $presenceCreated = $presence->getCreatedAt() === $presence->getUpdatedAt();
```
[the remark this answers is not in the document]
ArnabChatterjee20k: when created and updated are identical that means that is just created instead of updated. There was a trade of extra db call which can cause.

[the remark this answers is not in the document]
ArnabChatterjee20k: again fallback to the findOne based assertion as the upsertDocuments need it to set the $id to existing $id otherwise a drift between payload and actual data

src/Appwrite/Platform/Modules/Presences/HTTP/Get.php
```diff
@@ -0,0 +1,68 @@
...
+    {
+        $presence = $dbForProject->getDocument('presenceLogs', $presenceId);
+        if ($presence->isEmpty()) {
+            throw new Exception(Exception::PRESENCE_NOT_FOUND);
+        }
+
+        $presenceExpiresAt = $presence->getAttribute('expiresAt');
+
+        if (!empty($presenceExpiresAt) && DateTime::formatTz($presenceExpiresAt) < DateTime::formatTz(DateTime::now())) {
+            throw new Exception(Exception::PRESENCE_NOT_FOUND);
```
ArnabChatterjee20k, line 64: as we can't use queries with getDocument and we can't return the expired docs. so doing a manual ttl filter

app/realtime.php
```diff
@@ -1390,6 +1343,59 @@ function logError(Throwable $error, string $action, array $tags = [], ?Document
...
+                            ]);
+                        }
+
+                        foreach ($presences as $presence) {
+                            try {
+                                triggerPresenceEvent($project, new User([]), 'presences.[presenceId].delete', $presence);
+                            } catch (Throwable) {
+                                // Swallow errors to avoid breaking disconnect cleanup
+                            }
+                        }
```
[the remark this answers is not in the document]
ArnabChatterjee20k: its fixed and according tests are added

src/Appwrite/Utopia/Response/Model/Presence.php
```diff
@@ -0,0 +1,119 @@
...
+                'default' => '',
+                'example' => ['read("any")'],
+                'array' => true,
+            ])
+            ->addRule('userInternalId', [
+                'type' => self::TYPE_STRING,
+                'description' => 'User internal ID.',
+                'default' => '',
+                'example' => '1',
+            ])
```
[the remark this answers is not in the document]
ArnabChatterjee20k: resolved

app/realtime.php
```diff
@@ -253,12 +288,106 @@ function triggerStats(array $event, string $projectId): void
...
 }, ['register']);
 
+if (!$container->has('queueForEvents')) {
+    $container->set('queueForEvents', function ($pools) {
+        return new QueueEvent(new BrokerPool(
+            publisher: $pools->get('publisher')
+        ));
+    }, ['pools']);
+}
+
```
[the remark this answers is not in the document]
ArnabChatterjee20k: fixed with coroutine caching

src/Appwrite/Platform/Modules/Presences/HTTP/Upsert.php
ArnabChatterjee20k: server and console are collapsed into one

src/Appwrite/Presences/State.php
```diff
@@ -0,0 +1,236 @@
...
+        try {
+            if ($dbForProject->getAdapter()->getSupportForUpsertOnUniqueIndex()) {
+                // in v2 use permsmd5 in the queries as well to find the doc
+                $existingPresence = $dbForProject->findOne(self::COLLECTION_ID, [Query::equal('userInternalId', [$userInternalId])]);
+                if ($existingPresence->isEmpty()) {
+                    $presenceCreated = true;
+                } else {
+                    $presenceDocument->setAttribute('$id', $existingPresence->getId());
+                }
+                $presence = $dbForProject->upsertDocument(self::COLLECTION_ID, $presenceDocument);
```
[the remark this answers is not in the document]
ArnabChatterjee20k: @greptile its fixed already

src/Appwrite/Presences/State.php
```diff
@@ -0,0 +1,238 @@
...
```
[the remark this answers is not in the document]
ArnabChatterjee20k: This path intentionally avoids transactional locking because presence updates are expected to be extremely hot on Cloud, and serializing them would significantly impact throughput and latency.

For the native UPSERT flow we're currently accepting occasional metric/event overcounting as a tradeoff for avoiding lock contention on the hot path.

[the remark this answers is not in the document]
ArnabChatterjee20k: @abnegate WDYT, is it ok to hold a distributed lock here instead of a db lock? projectId:userInternalId, so that a single event is fired at a time?
I think even if dual events fired, the state isn't changing so it will be not any consistency issues and it will be the user who is firing dual events via api and realtime at the same time

app/config/collections/projects.php
```diff
@@ -2754,4 +2754,150 @@
...
+                'format' => '',
+                'size' => 65535,
+                'signed' => true,
+                'required' => false,
+                'default' => new \stdClass(),
+                'array' => false,
+                'filters' => ['json'],
+            ],
+            [
+                '$id' => ID::custom('perms_md5'),
```
abnegate: Doesn't fit any pattern, let's call it `permissionsHash`

app/config/collections/projects.php
```diff
@@ -2754,4 +2754,150 @@
...
+                'lengths' => [Database::LENGTH_KEY],
+                'orders' => [Database::ORDER_ASC]
+            ],
+            [
+                '$id' => ID::custom('_key_source_status'),
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['source', 'status'],
+                'lengths' => [Database::LENGTH_KEY, Database::LENGTH_KEY],
+                'orders' => [Database::ORDER_ASC, Database::ORDER_ASC]
+            ],
```
abnegate: Both are already length key and don't overflow, let's remove the length. Same for orders, ASC is default so they're redundant

app/config/collections/projects.php
```diff
@@ -2754,4 +2754,150 @@
...
+            [
+                '$id' => ID::custom('_key_perms_md5'),
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['perms_md5'],
+                'lengths' => [32],
+                'orders' => [Database::ORDER_ASC]
+            ]
```
abnegate, line 2896: Same here

app/config/collections/projects.php
```diff
@@ -2754,4 +2754,150 @@
...
+                'size' => Database::LENGTH_KEY,
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('metadata'),
+                'type' => Database::VAR_STRING,
```
abnegate: Use VAR_TEXT for freeform text

app/realtime.php
```diff
@@ -53,6 +64,29 @@
...
+$broker = new BrokerPool(publisher: $pools->get($publisherPoolName));
+$publisherForUsage = new UsagePublisher(
+    $broker,
+    new Queue(System::getEnv(
+        '_APP_STATS_USAGE_QUEUE_NAME',
+        \Appwrite\Event\Event::STATS_USAGE_QUEUE_NAME
+    ))
+);
+
```
abnegate: Can we move it to the dependency container?

app/realtime.php
```diff
@@ -253,12 +288,110 @@ function triggerStats(array $event, string $projectId): void
...
+        if (!isset($ctx['queueForRealtime'])) {
+            $ctx['queueForRealtime'] = new QueueRealtime();
+        }
+
+        return $ctx['queueForRealtime']->reset();
+    }
+}
+
+if (!function_exists('triggerPresenceEvent')) {
+    function triggerPresenceEvent(
```
abnegate: Let's re-order so the trigger and getter methods are grouped

app/realtime.php
```diff
@@ -429,6 +562,38 @@ function logError(Throwable $error, string $action, array $tags = [], ?Document
...
+                return;
+            }
+
+            try {
+                $consoleDB = getConsoleDB();
+                $consoleDB->getAuthorization()->skip(fn () => $consoleDB->foreach('projects', function (Document $project) use ($hostname): void {
+                    try {
+                        $dbForProject = getProjectDB($project);
+                        $deletionCount = $dbForProject->deleteDocuments('presenceLogs', [
+                            Query::equal('hostname', [$hostname]),
```
abnegate: Pod hostname will change across restarts won't it?

app/realtime.php
```diff
@@ -1026,298 +1217,45 @@ function logError(Throwable $error, string $action, array $tags = [], ?Document
...
+        $messageContainer->set('presenceState', fn () => $presenceState);
+        $messageContainer->set('database', fn () => $database);
+        $messageContainer->set('authorization', fn () => $authorization);
+        $messageContainer->set('project', fn () => $project);
+        $messageContainer->set('projectId', fn () => $projectId);
+        // Wrap the global helpers as first-class callables so handlers receive them via
+        // injection rather than reaching into the global namespace. Keeps PresenceHandler
+        // unit-testable without bootstrapping app/realtime.php.
+        $messageContainer->set('triggerPresenceUsage', fn () => triggerPresenceUsage(...));
+        $messageContainer->set('triggerPresenceEvent', fn () => triggerPresenceEvent(...));
```
abnegate: These function already isolate per coroutine and should be accessible on global scope, do we need them in a container?

app/realtime.php
```diff
@@ -1390,6 +1324,59 @@ function logError(Throwable $error, string $action, array $tags = [], ?Document
...
+                && $projectId !== 'console'
+            ) {
+                go(function () use ($presencesById, $projectId): void {
+                    // Fresh span: the parent realtime.close span finishes before this coroutine
+                    Span::init('realtime.close.presenceCleanup');
+                    Span::add('realtime.projectId', $projectId);
+                    Span::add('realtime.presenceCount', \count($presencesById));
+
+                    try {
+                        $consoleDB = getConsoleDB();
```
abnegate: Let's name this $dbForPlatform

app/realtime.php
```diff
@@ -1390,6 +1324,59 @@ function logError(Throwable $error, string $action, array $tags = [], ?Document
...
+                            Span::error($th);
+                            logError($th, 'realtimeOnClosePresenceDeletion', tags: [
+                                'projectId' => $projectId,
+                                'presences' => \count($presences)
+                            ]);
+                        }
+
+                        foreach ($presences as $presence) {
+                            try {
+                                triggerPresenceEvent($project, new User([]), 'presences.[presenceId].delete', $presence);
```
abnegate: Shouldn't we have the real user here?

src/Appwrite/Databases/PresenceState.php
abnegate: Makes sense, in that case, let's move it to src/Appwrite/Presences/State.php

src/Appwrite/Presences/State.php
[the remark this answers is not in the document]
abnegate: I don't think TOCTOU will hit here often, and like you say it's not a consistency issue even if it does, lets leave as-is and come back to it if we find it's an issue

src/Appwrite/Presences/State.php
```diff
@@ -0,0 +1,238 @@
...
+            }
+
+            $presenceDocument->setAttribute('$id', $currentPresence->getId());
+
+            return $dbForProject->updateDocument(self::COLLECTION_ID, $currentPresence->getId(), $presenceDocument);
+        });
+    }
+
+    public function setOwnerPermissions(Document $document, string $userId): Document
+    {
```
abnegate: I think we can remove these methods and just add flags that alter the existing method logic? If so, let's do it

src/Appwrite/Databases/PresenceState.php
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        sort($ownerPermissions, SORT_STRING);
+        $document->setAttribute('$permissions', $ownerPermissions);
+        $document->setAttribute('perms_md5', \md5(\json_encode($ownerPermissions)));
+
+        return $document;
+    }
+
+    private function assertPermissionsAgainstAuthorization(array $permissions, Authorization $authorization): void
```
abnegate: For assertion style methods our pattern is to name as `checkX`

src/Appwrite/Presences/State.php
```diff
@@ -0,0 +1,238 @@
...
+        return \sprintf(
+            '%s-cache:%s:%s:%s:collection:%s',
+            $dbForProject->getCacheName(),
+            $dbForProject->getAdapter()->getHostname(),
+            $dbForProject->getNamespace(),
+            $dbForProject->getTenant(),
+            self::COLLECTION_ID
+        );
+    }
```
abnegate: The DB has a function getCacheKeys, can we use it here? Not sure if it's public

src/Appwrite/Databases/PresenceState.php
```diff
@@ -0,0 +1,238 @@
...
+        try {
+            return $dbForProject->getCache()->load($this->getListCacheKey($dbForProject), $ttl, $cacheField);
+        } catch (\Throwable) {
+            return null;
+        }
+    }
+
+    public function saveListCacheField(
+        Database $dbForProject,
+        array $roles,
```
abnegate: The naming can be much clearer here, getListCacheFieldKey, getListCacheField, setListCacheField. purge is good

src/Appwrite/Platform/Modules/Presences/HTTP/Get.php
abnegate: In this case should we delete the presence too?

src/Appwrite/Platform/Modules/Presences/HTTP/XList.php
```diff
@@ -0,0 +1,180 @@
...
+            $cursor->setValue($cursorDocument);
+        }
+
+        $groupedQueries = Query::groupByType($queries);
+        $filterQueries = $groupedQueries['filters'];
+
+        // should be excluded from the user provided query as user query would be used for caching only
+        // otherwise cache will always miss due to the datetime now
+        $expiryFilter = Query::or([
+            Query::isNull('expiresAt'),
```
abnegate: Can we make this always set so we don't have to null check? `OR IS NULL` can break indexing on the DB side

src/Appwrite/Platform/Workers/Deletes.php
```diff
@@ -1739,4 +1745,38 @@ private function deleteExpiredTransactions(Document $project, callable $getProje
...
+        $oldestAllowed = DateTime::format((new \DateTime())->sub(\DateInterval::createFromDateString('30 days')));
+
+        $deleted = $dbForProject->deleteDocuments('presenceLogs', [
+            Query::or([
+                Query::and([
+                    Query::isNotNull('expiresAt'),
+                    Query::lessThan('expiresAt', $now),
+                ]),
+                Query::lessThan('$createdAt', $oldestAllowed),
+            ]),
```
abnegate: This will kill the DB optimizer, can we just check for expired here?

src/Appwrite/Realtime/Message/Dispatcher.php
```diff
@@ -0,0 +1,142 @@
...
+    public const PAYLOAD_SHAPE_LIST = 'list';
+
+    private const REQUIRED_PARAM_ERROR_FORMAT = 'Payload is not valid. %s is required';
+
+    /**
+     * @var array<string, Action>
+     */
+    private array $handlers = [];
+
+    public function register(Action $handler): self
```
abnegate: Let's call it `addHandler`

src/Appwrite/Utopia/Response/Model/Presence.php
```diff
@@ -0,0 +1,114 @@
...
+        $this->setAdditionalPropertiesKey('metadata');
+
+        $this
+            ->addRule('$id', [
+                'type' => self::TYPE_STRING,
+                'description' => 'Presence ID.',
+                'default' => '',
+                'example' => '5e5ea5c16897e',
+            ])
+            ->addRule('$sequence', [
```
abnegate: Let's remove, don't think we need to expose it

tests/e2e/Services/Realtime/RealtimeConsoleClientTest.php
```diff
@@ -257,7 +257,7 @@ public function testManualAuthentication(): void
         $this->assertEquals('error', $response['type']);
         $this->assertNotEmpty($response['data']);
         $this->assertEquals(1003, $response['data']['code']);
-        $this->assertEquals('Payload is not valid.', $response['data']['message']);
+        $this->assertEquals('Payload is not valid. session is required', $response['data']['message']);
```
abnegate, line 261: Let's fix casing

app/realtime.php
ArnabChatterjee20k: Yes but they were getting cross referenced across files. referencing via the container seemingly clean

src/Appwrite/Presences/State.php
ArnabChatterjee20k: Gotch you , thought with methods it would be cleaner. addressed

src/Appwrite/Presences/State.php
ArnabChatterjee20k: Not sure if we want to do it as we are manually saving to the cache, cause I see the same logic in the documents xlist.php as well

src/Appwrite/Platform/Modules/Presences/HTTP/XList.php
ArnabChatterjee20k: we dont set expiry on the realtime then we can get stale data as according to the discussion realtime upsert/delete the presence on connect/disconnect and somehow can get deleted in the worker if expiry present

src/Appwrite/Platform/Workers/Deletes.php
ArnabChatterjee20k: It was but there can be a chance if somehow old precenses aren't deleted, in that case we need to sweep them as well

tests/e2e/Services/Realtime/RealtimeConsoleClientTest.php
ArnabChatterjee20k: It's in the appwrite 1.9.x authentication message. `session` is the key of the param here

app/realtime.php
abnegate: The functions might make more sense on the state class 🤔 then we can just pass that around instead of global functions and registering in container

src/Appwrite/Presences/State.php
abnegate: Generally agree, but in this case because the logic is almost exactly the same, just doing something extra, an extra param to control the extra behavior makes it more concise without adding any complexity or sacrificing readability

src/Appwrite/Presences/State.php
abnegate: The function is public, we should use it here (and document list should have too), so that we keep key structure consistent. It only builds the keys, doesn't actually save to cache: https://github.com/utopia-php/database/blob/3391c97318f0e7f94d2c1ea0f7d09e5ba8aad696/src/Database/Database.php#L9393

src/Appwrite/Platform/Workers/Deletes.php
abnegate: We can do that part separately, as a part of the maintenance task, so we can split the queries

src/Appwrite/Platform/Modules/Presences/HTTP/Upsert.php
```diff
@@ -0,0 +1,181 @@
...
+            'userId' => $resolvedUserId,
+            'status' => $status,
+            'source' => $isGraphQL ? 'graphql' : 'rest',
+            'expiresAt' => $expiresAt ?? DateTime::addSeconds(new \DateTime(), 15 * 60),
+            'metadata' => $metadata,
+        ];
+
+        $presenceState = new PresenceState();
+        $presenceDocument = new Document($presenceData);
+        $presenceState->setPermissions($presenceDocument, $permissions, $user, $authorization);
```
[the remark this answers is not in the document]
ArnabChatterjee20k: fixed

abnegate (approved):

src/Appwrite/Presences/State.php
```diff
@@ -0,0 +1,274 @@
...
+
+        try {
+            if ($dbForProject->getAdapter()->getSupportForUpsertOnUniqueIndex()) {
+                $existingPresence = $dbForProject->findOne(self::COLLECTION_ID, [Query::equal('userInternalId', [$userInternalId])]);
+                if ($existingPresence->isEmpty()) {
+                    $presenceCreated = true;
+                } else {
+                    $presenceDocument->setAttribute('$id', $existingPresence->getId());
+                }
+                $presence = $dbForProject->upsertDocument(self::COLLECTION_ID, $presenceDocument);
```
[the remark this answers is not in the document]
ArnabChatterjee20k: Its pointing towards the api key/console behaviour where we dont see the permission
Instead of the client where permission is source to segregate

For the client side the permissions are intact and no one can modify each others

src/Appwrite/Realtime/Message/Handlers/Presence.php
```diff
@@ -0,0 +1,128 @@
...
+            'source' => 'realtime',
+            'status' => $status,
+            'expiresAt' => DateTime::format((new \DateTime())->modify('+30 days')),
```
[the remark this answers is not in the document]
ArnabChatterjee20k: Its intended. If staleness comes, server api can use the delete to delete the stale precenses

src/Appwrite/Platform/Modules/Presences/HTTP/Update.php
```diff
@@ -0,0 +1,198 @@
...
+            throw new Exception(Exception::GENERAL_UNAUTHORIZED_SCOPE, 'userId is not allowed for non-API key and non-privileged users');
+        }
+
+        $presence = $dbForProject->getDocument('presenceLogs', $presenceId);
+
+        if ($presence->isEmpty()) {
+            throw new Exception(Exception::PRESENCE_NOT_FOUND, params: [$presenceId]);
+        }
+
+        $updateData = [];
```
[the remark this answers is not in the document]
ArnabChatterjee20k: fixed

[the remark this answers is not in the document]
ArnabChatterjee20k: fixed

app/realtime.php
```diff
@@ -1390,6 +1227,86 @@ function logError(Throwable $error, string $action, array $tags = [], ?Document
...
+                                    $queueForRealtime,
+                                    $project,
+                                    $user,
+                                    'presences.[presenceId].delete',
+                                    $presence,
+                                );
+                            } catch (Throwable) {
+                                // Swallow errors to avoid breaking disconnect cleanup
+                            }
+                        }
```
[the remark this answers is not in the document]
ArnabChatterjee20k: fixed

src/Appwrite/Platform/Workers/Deletes.php
```diff
@@ -1741,4 +1747,25 @@ private function deleteExpiredTransactions(Document $project, callable $getProje
...
+            // Swallow errors to avoid breaking the cleanup process
+        });
+
+        if ($deleted > 0) {
+            $usage = (new UsageContext())->addMetric(METRIC_USERS_PRESENCE, -$deleted);
+            $publisherForUsage->enqueue(new Usage(
+                project: $project,
+                metrics: $usage->getMetrics(),
+            ));
+        }
```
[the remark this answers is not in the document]
ArnabChatterjee20k: those are the expired events and we shouldn't publish those cause those are unpredictably deleted. +1/-1 is a state of the total number of users.
Prolly later on we can have an expired presence channel and event
Realtime trigger here doesn't make sense at all

src/Appwrite/Platform/Modules/Presences/HTTP/Update.php
```diff
@@ -0,0 +1,204 @@
...
+            throw new Exception(Exception::DOCUMENT_UPDATE_CONFLICT, $e->getMessage(), previous: $e);
+        }
+
+        if ($purge) {
+            $presenceState->purgeListCache($dbForProject);
+        }
+
+        $queueForEvents->setParam('presenceId', $presence->getId());
+
+        $response->dynamic($presence, Response::MODEL_PRESENCE);
```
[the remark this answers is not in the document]
ArnabChatterjee20k: it is the intended behaviour and user decides what to do with the cache

abnegate (approved):

PR #11911, opened 2026-04-15
Fix 'occured' typo in proxy_error_override message

SAY-5:
The fallback message rendered when an upstream proxy returns 4xx/5xx without a more specific error spelled `occured` twice on `app/views/general/error.phtml:45`:

```php
$message = $code >= 500 ? 'An unexpected server error occured.' : 'An unexpected client error occured.';
```

Both strings are user-visible on the generic Appwrite error page (the `proxy_error_override` case is the catch-all branch above the per-code switch). Replace with `occurred`. Only string literals changed — no logic touched, no PHP available locally to lint, but the diff is a one-character pair replacement within a ternary so the parse can't change.


SAY-5:
Closing — typo-only PR. Multiple maintainers across the OSS ecosystem have flagged my recent typo-sweep PRs as AI-generated spam (notably hashicorp/nomad#27855, hashicorp/hcl#794, argo-cd reviewers). I should have caught this pattern sooner. Apologies for the noise.

PR #11912, opened 2026-04-15
Feat: Fallback email custom template

Meldiron:
## What does this PR do?

- Removes SMS templates, they are unsafe and not supported anyway
- Uses EN as default locale in template endpoints (locale optional)
- Removes type and locale from path of template endpoint, so locale can be left empty by not passing search param)

## Test Plan

- [x] add new integration test to ensure locale for templates behaves as it should

## Related PRs and Issues

x

## Checklist


app/controllers/api/projects.php
```diff
@@ -833,75 +833,8 @@
...
-        $template['type'] = $type;
-        $template['locale'] = $locale;
-
-        $response->dynamic(new Document($template), Response::MODEL_SMS_TEMPLATE);
-    });
-
-
-Http::get('/v1/projects/:projectId/templates/email/:type/:locale')
+Http::get('/v1/projects/:projectId/templates/email')
+    ->alias('/v1/projects/:projectId/templates/email/:type/:locale')
```
Meldiron, line 837: I cant see strong reason if /:type should stay in path or not. Open to change if we have strong reason. For now I went with simplicity

TorstenDittmann: do we even need it? its mostly an internal endpoint right?

TorstenDittmann (approved):

Meldiron:
We had problem before when console was cached, so breaking changes didnt go live together, causing trouble for some users. So better be safe.
Plus Anyone using CLI methods, while never announced, they could be facing trouble too.

Small alias sounds like best bet. Plus, upcomming PR will move this endpoint to a public one under /v1/project

app/controllers/api/projects.php
```diff
@@ -920,10 +853,12 @@
...
     ->param('projectId', '', fn (Database $dbForPlatform) => new UID($dbForPlatform->getAdapter()->getMaxUIDLength()), 'Project unique ID.', false, ['dbForPlatform'])
     ->param('type', '', new WhiteList(Config::getParam('locale-templates')['email'] ?? [], true), 'Template type')
-    ->param('locale', '', fn ($localeCodes) => new WhiteList($localeCodes), 'Template locale', false, ['localeCodes'])
+    ->param('locale', '', fn ($localeCodes) => new WhiteList($localeCodes), 'Template locale', true, ['localeCodes'])
     ->inject('response')
     ->inject('dbForPlatform')
-    ->action(function (string $projectId, string $type, string $locale, Response $response, Database $dbForPlatform) {
+    ->inject('locale')
+    ->action(function (string $projectId, string $type, string $locale, Response $response, Database $dbForPlatform, Locale $localeObject) {
+        $locale = $locale ?: $localeObject->default ?: $localeObject->fallback ?: System::getEnv('_APP_LOCALE', 'en');
```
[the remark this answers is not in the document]
Meldiron: I think thats not the case. Look at this PR, it proves it works as expected: https://github.com/utopia-php/http/pull/237

PR #11940, opened 2026-04-17
fix: do not cache error responses for storage preview, bump utopia-php/image to 0.8.5

loks0n:
## Summary

- **Root cause 1 (library):** `utopia-php/image` was using the deprecated `magick convert` shell command for AVIF/HEIC output, which fails on ImageMagick 7. Fixed upstream in [utopia-php/image#43](https://github.com/utopia-php/image/pull/43) and released as `0.8.5`. This PR bumps to that version.
- **Root cause 2 (caching):** The cache write hook in `app/controllers/shared/api.php` was writing any non-empty response payload to cache, including error responses. When AVIF conversion failed (first request), the error JSON was cached, and subsequent requests with the same cache key would serve the cached error — manifesting as `storage_bucket_not_found` on cloud.
- **Fix:** Cache write hook now checks `$statusCode >= 200 && $statusCode < 300` before writing to cache.
- **Tests:** Added `testFilePreviewAvifPublic()` for the customer's URL pattern (project in query params, no auth headers), and AVIF assertions in `testFilePreview()` for PNG→AVIF and JPEG→AVIF conversions.

## Test plan

- [ ] `docker compose exec appwrite test tests/e2e/Services/Storage/StorageCustom`
- [ ] `docker compose exec appwrite test tests/e2e/Services/Storage/StorageServer`
- [ ] Verify `GET /v1/storage/buckets/{id}/files/{id}/preview?output=avif` returns `200 image/avif`
- [ ] Verify a previously cached error for `output=avif` no longer blocks subsequent requests

### DIFF
--- a/app/controllers/shared/api.php
+++ b/app/controllers/shared/api.php
@@ -971,7 +971,8 @@ Http::shutdown()
         if ($useCache) {
             $resource = $resourceType = null;
             $data = $response->getPayload();
-            if (! empty($data['payload'])) {
+            $statusCode = $response->getStatusCode();
+            if (! empty($data['payload']) && $statusCode >= 200 && $statusCode < 300) {
                 $pattern = $route->getLabel('cache.resource', null);
                 if (! empty($pattern)) {
                     $resource = $parseLabel($pattern, $responsePayload, $requestParams, $user);
--- a/composer.lock
+++ b/composer.lock
@@ -4325,16 +4325,16 @@
         },
         {
             "name": "utopia-php/image",
-            "version": "0.8.4",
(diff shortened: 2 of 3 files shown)


eldadfux (approved):

loks0n:
Note: a related cache key collision bug was found during investigation and fixed in #11941. The cache key never included the project ID, so two projects sharing the same bucketId, fileId, and transform params would collide — the cache hit path would then re-validate the wrong project's bucket and throw `storage_bucket_not_found`. That was the actual root cause of the specific customer URL failing.

PR #11948, opened 2026-04-18
fix: make target endpoint params nullable to prevent server error on null

avasis-ai:
## Summary

Fixes #8785

When `null` is explicitly passed for optional parameters (`name`, `providerId`, `identifier`) in the CreateTarget and UpdateTarget endpoints, the Utopia framework skips validators but still passes `null` to the action function. Since the function signatures expected `string` (not nullable), this caused a vague server error:

```
{closure}(): Argument #5 ($name) must be of type string, null given
```

## Changes

- **CreateTarget** (`POST /v1/users/:userId/targets`): Made `$providerId` and `$name` nullable (`?string`), and added null coalescing for the `$providerId` lookup
- **UpdateTarget** (`PATCH /v1/users/:userId/targets/:targetId`): Made `$identifier`, `$providerId`, and `$name` nullable (`?string`)

The existing null checks (`if ($name)`, `if ($providerId)`, `if ($identifier)`) already handle null values correctly downstream, so no additional logic changes were needed.

## Test Plan

- `POST /v1/users/:userId/targets` with `name: null` and `providerId: null` should succeed instead of returning a server error
- `PATCH /v1/users/:userId/targets/:targetId` with `identifier: null`, `providerId: null`, or `name: null` should succeed and leave those fields unchanged


HarshMN2345:

PR #11949, opened 2026-04-19
chore: bump PHPStan to level 4 and fix all new errors

ChiragAgg5k:
## Summary

Raises `phpstan.neon` level from **3 → 4** and fixes the **549 new errors** that level 4 surfaces across **157 files**. Fixes are root-cause only — no `@phpstan-ignore`, no `@var` casts, no baseline, no widened types.

Net diff: **378 insertions, 1129 deletions** (mostly dead code and redundant defensive checks removed).

## Latent bugs fixed along the way

- **Path-traversal guard** in `app/controllers/general.php:1619` was negating `\substr(...)` before the strict comparison (`!\substr(...) === $base` was effectively `false === $base`). Rewritten as `\substr(...) !== $base`.
- **Databases / TablesDB Logs list** (`src/Appwrite/Platform/Modules/Databases/Http/{Databases,TablesDB}/Logs/XList.php`) was importing the raw Matomo `DeviceDetector` — whose `getDevice()` returns `?int` — but treated the result as an array with `deviceName / deviceBrand / deviceModel` keys, so those response fields were always empty strings. Swapped to `Appwrite\Detector\Detector`, matching the wrapper already used a few lines below for `$os` / `$client`.
- **Functions build worker** (`src/Appwrite/Platform/Modules/Functions/Workers/Builds.php`) checked `$resourceKey === 'functions'` where `$resourceKey` is `'functionId' | 'siteId'` — always false. Switched to the intended `$resource->getCollection() === 'functions'`.
- **OpenSSL::encrypt** return type tightened to `string|false` to match `openssl_encrypt`; this keeps callers' `=== false` error handling meaningful (previously masked).
- **Msg91 messaging provider** (`app/controllers/api/messaging.php`) had a dead `array_key_exists('from', [])` branch on an always-empty array literal.

## Large cleanup categories

- Removed redundant `?? default` on array offsets and expressions that PHPStan now proves are non-nullable.
- Removed unreachable statements (mostly `return;` after `throw` or `markTestSkipped()`).
- Removed redundant `is_array` / `is_string` / `is_bool` / `instanceof` checks on already-narrowed types.
- Added `default =>` / throwing arms to non-exhaustive matches on `string` / `mixed` inputs.
- Removed dead `$document === false` branches where method return types were tightened to non-nullable `Document`.
- Removed unused properties (`$version` on Etsy/Zoom OAuth2, `$paths` on Installer State, `$source` on MigrationsWorker, `$account2` on two GraphQL auth tests), unused traits (`ApiVectorsDB`, `DatabaseFixture`), and unused methods (e.g. `Interval::cleanupStaleExecutions`).
- Replaced `assertTrue(true)` and redundant `assertIsArray` / `assertIsString` / `assertNotNull` with `addToAssertionCount(1)` or `assertNotEmpty` where the runtime type was already known.

## Test plan

- [x] `composer check` (phpstan level 4) — `[OK] No errors`
- [x] `composer lint` (pint) — `{"result":"pass"}`
- [x] `php -l` on every modified file — no syntax errors
- [ ] CI pipeline green on this PR
- [ ] Spot-check the three latent bug fixes (`general.php` path guard, Logs/XList device detector, Functions Builds worker resource check) in a dev environment before merging


src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Usage/Get.php
```diff
@@ -119,6 +119,7 @@ public function action(string $databaseId, string $range, string $collectionId,
         $format = match ($days['period']) {
             '1h' => 'Y-m-d\TH:00:00.000P',
             '1d' => 'Y-m-d\T00:00:00.000P',
+            default => throw new \LogicException('Unexpected period: ' . $days['period']),
```
loks0n, line 122: Should this be 5xx or 4xx?

src/Appwrite/Platform/Modules/Databases/Http/Databases/Usage/Get.php
```diff
@@ -144,6 +144,7 @@ public function action(string $databaseId, string $range, UtopiaResponse $respon
         $format = match ($days['period']) {
             '1h' => 'Y-m-d\TH:00:00.000P',
             '1d' => 'Y-m-d\T00:00:00.000P',
+            default => throw new \LogicException('Unexpected period: ' . $days['period']),
```
loks0n, line 147: Should this be a 5xx or 4xx?

src/Appwrite/Platform/Modules/Databases/Http/Databases/Usage/XList.php
```diff
@@ -133,6 +133,7 @@ public function action(string $range, UtopiaResponse $response, Database $dbForP
         $format = match ($days['period']) {
             '1h' => 'Y-m-d\TH:00:00.000P',
             '1d' => 'Y-m-d\T00:00:00.000P',
+            default => throw new \LogicException('Unexpected period: ' . $days['period']),
```
loks0n, line 136: Should this be a 5xx or 4xx?

src/Appwrite/Platform/Modules/Functions/Http/Usage/Get.php
```diff
@@ -112,6 +112,7 @@ public function action(string $functionId, string $range, Response $response, Da
         $format = match ($days['period']) {
             '1h' => 'Y-m-d\TH:00:00.000P',
             '1d' => 'Y-m-d\T00:00:00.000P',
+            default => throw new Exception(Exception::GENERAL_SERVER_ERROR, 'Unsupported period "' . $days['period'] . '".'),
```
loks0n, line 115: Should this be a 5xx or 4xx?Should this be a 5xx or 4xx?

src/Appwrite/Platform/Modules/Functions/Http/Usage/XList.php
```diff
@@ -104,6 +105,7 @@ public function action(string $range, Response $response, Database $dbForProject
         $format = match ($days['period']) {
             '1h' => 'Y-m-d\TH:00:00.000P',
             '1d' => 'Y-m-d\T00:00:00.000P',
+            default => throw new Exception(Exception::GENERAL_SERVER_ERROR, 'Unsupported period "' . $days['period'] . '".'),
```
loks0n, line 108: Should this be a 5xx or 4xx?

app/controllers/api/account.php
loks0n, line 140: How do we know we can remove this?

app/controllers/general.php
```diff
@@ -774,20 +771,6 @@ function router(Http $utopia, Database $dbForPlatform, callable $getProjectDB, S
...
-                ? RESOURCE_TYPE_FUNCTIONS
-                : RESOURCE_TYPE_SITES;
-
-            $queueForDeletes
-                ->setProject($project)
-                ->setResourceType($resourceType)
-                ->setResource($resource->getSequence())
-                ->setType(DELETE_TYPE_EXECUTIONS_LIMIT)
-                ->trigger();
-        }
```
loks0n, line 789: How come we can remove this? Is the executions cleanup supposed to be running?

src/Appwrite/Platform/Modules/Functions/Http/Executions/Create.php
```diff
@@ -349,15 +331,6 @@ public function action(
...
             }
 
-            if ($executionsRetentionCount > 0 && ENABLE_EXECUTIONS_LIMIT_ON_ROUTE) {
-                $queueForDeletes
-                    ->setProject($project)
-                    ->setResource($function->getSequence())
-                    ->setResourceType(RESOURCE_TYPE_FUNCTIONS)
-                    ->setType(DELETE_TYPE_EXECUTIONS_LIMIT)
-                    ->trigger();
-            }
```
loks0n, line 359: How do we know we can remove this?

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Create.php
```diff
@@ -293,16 +293,6 @@ public function action(string $databaseId, string $documentId, string $collectio
...
 
-            if ($permission === Database::PERMISSION_UPDATE) {
-                $validDocument = $authorization->isValid(
-                    new Input($permission, $document->getUpdate())
-                );
-                $valid = $validCollection || $validDocument;
-                if ($documentSecurity && !$valid) {
-                    throw new Exception(Exception::USER_UNAUTHORIZED, $authorization->getDescription());
-                }
-            }
```
loks0n, line 304: How come this is removeable?

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Usage/Get.php
ChiragAgg5k: This renders as 5xx, and that is intended. `$days['period']` here comes from the internal `Config::getParam('usage')` lookup keyed by `$range` — `$range` itself is already validated against the `new WhiteList(['24h', '30d', '90d'], true)` at the action parameter. So the only way to hit `default` is the `usage` config drifting out of sync with the allowed ranges, which is a server-side config/code bug, not user input. `\LogicException` → 500 via the generic `\Throwable` handler in `app/http.php` is the right shape.

src/Appwrite/Platform/Modules/Databases/Http/Databases/Usage/Get.php
ChiragAgg5k: Same as the sibling comment on `Collections/Usage/Get.php` — 5xx is intended. `$days[\$range]` is loaded from the internal `Config::getParam(\x27usage\x27)` config after `$range` is already constrained by `new WhiteList([\x2724h\x27, \x2730d\x27, \x2790d\x27], true)` on the param. Hitting `default` means the config and the whitelist have drifted apart — a server-side programming/config error, not a user error. `\LogicException` falls through to the generic `\Throwable` handler and renders as 500.

src/Appwrite/Platform/Modules/Databases/Http/Databases/Usage/XList.php
ChiragAgg5k: Same as the sibling comment on `Collections/Usage/Get.php` — 5xx is intended. `$days[\$range]` is loaded from the internal `Config::getParam(\x27usage\x27)` config after `$range` is already constrained by `new WhiteList([\x2724h\x27, \x2730d\x27, \x2790d\x27], true)` on the param. Hitting `default` means the config and the whitelist have drifted apart — a server-side programming/config error, not a user error. `\LogicException` falls through to the generic `\Throwable` handler and renders as 500.

src/Appwrite/Platform/Modules/Functions/Http/Usage/Get.php
ChiragAgg5k: Same as the sibling comment on `Collections/Usage/Get.php` — 5xx is intended. `$days[\$range]` is loaded from the internal `Config::getParam(\x27usage\x27)` config after `$range` is already constrained by `new WhiteList([\x2724h\x27, \x2730d\x27, \x2790d\x27], true)` on the param. Hitting `default` means the config and the whitelist have drifted apart — a server-side programming/config error, not a user error. `\LogicException` falls through to the generic `\Throwable` handler and renders as 500.

src/Appwrite/Platform/Modules/Functions/Http/Usage/XList.php
ChiragAgg5k: Same as the sibling comment on `Collections/Usage/Get.php` — 5xx is intended. `$days[\$range]` is loaded from the internal `Config::getParam(\x27usage\x27)` config after `$range` is already constrained by `new WhiteList([\x2724h\x27, \x2730d\x27, \x2790d\x27], true)` on the param. Hitting `default` means the config and the whitelist have drifted apart — a server-side programming/config error, not a user error. `\LogicException` falls through to the generic `\Throwable` handler and renders as 500.

app/controllers/api/account.php
ChiragAgg5k: The first `match` right above (`lines 126–133`) already has a `default => throw new Exception(Exception::USER_INVALID_TOKEN)`, so by the time we reach this second `match`, `$verifiedToken->getAttribute(\x27type\x27)` is guaranteed to be one of `TOKEN_TYPE_MAGIC_URL`, `TOKEN_TYPE_OAUTH2`, `TOKEN_TYPE_EMAIL`, `TOKEN_TYPE_PHONE`, or `TOKEN_TYPE_GENERIC`. The deleted arms were for `TOKEN_TYPE_VERIFICATION`, `TOKEN_TYPE_RECOVERY`, and `TOKEN_TYPE_INVITE`, which never reach this match — the first one already threw on them. PHPStan flagged it as `match.alwaysFalse` for the same reason.

app/controllers/general.php
ChiragAgg5k: Good catch — this one I got wrong in the initial pass and restored. The block was guarded by `ENABLE_EXECUTIONS_LIMIT_ON_ROUTE`, a constant hardcoded to `false` in `app/init/constants.php` since `ccaea5d010`. PHPStan sees the `&&` as unreachable, which is technically correct *today*, but it is a feature flag, not dead code — removing it silently dropped the intent.

In the latest push I:

1. Restored the `/* cleanup */` block here in `general.php` and the two matching blocks in `Functions/Http/Executions/Create.php`.
2. Converted `ENABLE_EXECUTIONS_LIMIT_ON_ROUTE` from a hardcoded `const` to a `define()` backed by a new `_APP_EXECUTIONS_LIMIT_ON_ROUTE` env var (defaults to `disabled`), so the toggle is now a runtime decision and PHPStan no longer folds the `&&` away.

Behavior is unchanged vs. main (still off by default), but the flag can be flipped via env without a code change now. If there is a plan for when this should go on, happy to wire up the default differently.

src/Appwrite/Platform/Modules/Functions/Http/Executions/Create.php
ChiragAgg5k: Same root cause as the `general.php` cleanup comment — two blocks in this file were gated on `ENABLE_EXECUTIONS_LIMIT_ON_ROUTE`, a feature-flag constant hardcoded to `false`, which PHPStan folded into dead code. I restored both cleanup blocks (one after the async/scheduled `STATUS_CODE_ACCEPTED` return, one before the sync `STATUS_CODE_CREATED` response), re-added the `DeleteEvent $queueForDeletes` / `int $executionsRetentionCount` injections, and converted the constant to a `define()` backed by the new `_APP_EXECUTIONS_LIMIT_ON_ROUTE` env var. Behavior unchanged from main; the toggle is just runtime now.

src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Documents/Create.php
ChiragAgg5k: This one is worth walking through because the outer guard makes the inner check trivially dead. The outer `if` right above the deleted block is:

```php
if (($permission === Database::PERMISSION_UPDATE && !$documentSecurity) || !$validCollection) {
    throw new Exception(Exception::USER_UNAUTHORIZED, ...);
}
```

So to reach the deleted block we need **both** of these to be false:
- `!($permission === UPDATE && !$documentSecurity)` → `$permission !== UPDATE || $documentSecurity`
- `$validCollection === true`

Then the deleted block gates on `$permission === Database::PERMISSION_UPDATE`, which combined with the above narrows to: `$permission === UPDATE && $documentSecurity && $validCollection`.

Inside it was:

```php
$validDocument = $authorization->isValid(new Input($permission, $document->getUpdate()));
$valid = $validCollection || $validDocument;
if ($documentSecurity && !$valid) {
    throw ...;
}
```

`$valid` is always `true` (because `$validCollection` is already `true` at this point), so `!$valid` is `false`, so the inner `throw` never fires. `$valid` isn\x27t read anywhere after, and `$authorization->isValid(...)` on the document is side-effect free — it just computes a boolean — so removing the whole block is a pure dead-code cleanup. PHPStan flagged `$valid` being always true which is what surfaced it.

Worth flagging: the block looks like it *was supposed to* implement "for UPDATE with documentSecurity, either collection-level OR document-level permission is enough", but the outer guard is too strict for that intent (it already requires `$validCollection`). If document-level fallback is wanted, the outer guard needs rethinking — but that is a separate change from this PR\x27s scope.

src/Appwrite/Event/Validator/Event.php
```diff
@@ -61,9 +61,6 @@ public function isValid($value): bool
         if ($hasSubSubResource) {
             $subSubType = $parts[4];
             $subSubResource = $parts[5];
-            if ($count === 8) {
```
abnegate, line 64: Why can this be removed? Did we never hit it?

src/Appwrite/Messaging/Adapter/Realtime.php
```diff
@@ -461,15 +461,8 @@ public static function constructSubscriptions(array $channelNames, callable $get
...
-                $expectedType = $reservedParamExpectedTypes[$paramKey];
-                $isExpectedType = match ($expectedType) {
-                    'array' => \is_array($params),
-                    'string' => \is_string($params),
-                    default => false,
-                };
-
                 // If the value matches the expected type dont use it the queries
-                if ($isExpectedType) {
+                if (\is_string($params)) {
```
abnegate: Doesn't seem equivalent :thinking:

src/Appwrite/Platform/Tasks/ScheduleBase.php
```diff
@@ -102,7 +102,7 @@ public function action(BrokerPool $publisher, BrokerPool $publisherMigrations, B
             $this->collectSchedules($dbForPlatform, $getProjectDB, $lastSyncUpdate, $isResourceBlocked);
         });
 
-        while (true) {
+        for (;;) {
```
abnegate: `while(true)` is more sensible for this

src/Appwrite/SDK/Specification/Format.php
```diff
@@ -111,35 +111,23 @@ protected function buildEnumBlacklist(): array
...
+                    'exclude' => $config['exclude'],
                 ];
-                if (isset($config['excludeKeys'])) {
-                    $entry['excludeKeys'] = $config['excludeKeys'];
-                }
-                if (isset($config['exclude'])) {
-                    $entry['exclude'] = $config['exclude'];
-                }
-                $blacklist[] = $entry;
             }
```
abnegate, line 135: What about `exclude`?

src/Appwrite/SDK/Specification/Format.php
```diff
@@ -968,8 +956,7 @@ protected function getRequestParameterConfig(string $service, string $method, st
                 continue;
             }
 
-            $config['required'] = $override['required'] ?? $config['required'];
-            $config['nullable'] = $override['nullable'] ?? $config['nullable'];
+            $config['required'] = $override['required'];
```
abnegate: How is this equivalent?

src/Appwrite/Utopia/Request/Filter.php
```diff
@@ -45,12 +45,6 @@ public function getDbForProject(): ?Database
...
     public function getParamValue(string $key, mixed $default = ''): mixed
     {
-        try {
-            $value = $this->params[$key] ?? $default;
-        } catch (\Exception $e) {
-            $value = $default;
-        }
-
-        return $value;
+        return $this->params[$key] ?? $default;
```
abnegate, line 48: Does this run callbacks?

tests/e2e/Scopes/ApiVectorsDB.php
abnegate, line 1: Don't think it should be deleted

src/Appwrite/Event/Validator/Event.php
ChiragAgg5k: This branch was unreachable. `isValid()` returns `false` when `$count > 7`, so a later `if ($count === 8)` could never be hit. I left the removal as-is because it is dead code rather than a behavior change.

src/Appwrite/SDK/Specification/Format.php
ChiragAgg5k: Good catch. I fixed this in c973ca0a5d by preserving both optional shapes while combining the blacklist configs: `excludeKeys` is copied when present and `exclude` is copied when present. That keeps the provider usage blacklist behavior equivalent.

src/Appwrite/Utopia/Request/Filter.php
ChiragAgg5k: No, this path does not run callbacks. The params here come from the route path values as plain values, and the previous code was also only doing direct array access inside a try/catch. So removing the try/catch does not skip any callback execution; it just keeps the same `$this->params[$key] ?? $default` lookup without the unreachable exception path.

tests/e2e/Scopes/ApiVectorsDB.php
ChiragAgg5k: Agreed. I restored `tests/e2e/Scopes/ApiVectorsDB.php` in c973ca0a5d and added the trait back to `VectorsDBCustomClientTest`, so the scope is still covered and PHPStan sees it as used.

abnegate (approved):

PR #11951, opened 2026-04-19
Fix: Handle null name in CreateUser endpoint to prevent server error

utkarshhxd:
This PR fixes an issue where passing null as the name parameter in CreateUser endpoints caused a 500 server error.

### Changes
- Updated the name validator to allow nullable values using Nullable(Text)
- Ensured null values are handled safely and converted to an empty string

### Result
- Prevents server error when name is null
- Keeps behavior consistent with name being optional


utkarshhxd:
Thanks for the review, I have updated the fix to handle null name across all CreateUser endpoints

HarshMN2345:

PR #11964, opened 2026-04-21
Feat: public project policies

Meldiron:
## What does this PR do?

Introduces public API endpoints for managing project policies (auth settings mostly)

## Test Plan

existing tests must pass, plus:

- [x] new tests

## Related PRs and Issues

x

## Checklist


eldadfux (approved):

src/Appwrite/Platform/Modules/Project/Http/Project/Policies/UserLimit/Update.php
```diff
@@ -0,0 +1,91 @@
...
+        ?int $total,
+        Response $response,
+        Database $dbForPlatform,
+        Document $project,
+        Authorization $authorization,
+        Event $queueForEvents,
+    ): void {
+        $auths = $project->getAttribute('auths', []);
+
+        if (\is_null($total)) {
```
sambhusankar, line 73: - Store null in the database to represent "no limit" and ensure the response model handles it correctly, or
- Change the param type to accept 0 directly (instead of null) to mean "unlimited", and update the description accordingly.

sambhusankar (changes requested):
Inconsistent null/0 handling for limit policies

Please fix it in
- SessionLimiT
- UserLimit
- PasswordHistory

src/Appwrite/Platform/Modules/Project/Http/Project/Policies/UserLimit/Update.php
Meldiron: Heyy 👋

We made decision to accept null on API. This was preffered over 0 because in future there might be ranges supporting negative numbers too (for other endpoints), and 0 might be real value for them. Accepting null to disable feature makes most sense - giving us consistent solution now, and in future.

In database we remain using 0 to prevent need for migration or dual-reading in the codebase. Eventually this will all get migrated, but only later when we have more small things to migrate

Meldiron (commented):
lgtm

src/Appwrite/Platform/Modules/Project/Http/Project/Policies/UserLimit/Update.php
sambhusankar: got it, makes sense

sambhusankar: got it, makes sense

PR #11966, opened 2026-04-21
fix(users): handle null name in CreateUser endpoints (#8785)

Seven7000000:
## What

All eight `Users.createUser*` endpoints (`POST /v1/users` plus the seven hashed-password variants `bcrypt`, `md5`, `argon2`, `sha`, `phpass`, `scrypt`, `scrypt-modified`) return a vague **500 server error** when an explicit `null` is passed for the optional `name` parameter, e.g.:

```
{closure}(): Argument #5 ($name) must be of type string, null given
```

This happens because the Utopia framework skips validators for optional params when the value is `null` (see [utopia-php/servers Base.php#L293-L295](https://github.com/utopia-php/servers/blob/fd5c8d32778f265256c1936372a071b944f5ba8a/src/Servers/Base.php#L293-L295)), but the action closure declares `string $name`, which triggers a PHP `TypeError` at call time.

## How

Following the approach suggested by @stnguyen90 in the issue and the established pattern already used for the `email` and `phone` params in the same file:

1. Wrap the existing `Text(128)` validator with `Nullable(...)` on all eight `name` param declarations so the validator correctly accepts `null`.
2. Update the action closures and the `createUser()` helper to accept `?string $name`.
3. Coalesce `null` to an empty string inside `createUser()` so existing downstream behaviour (`search` field, `name` column, etc.) is unchanged.

Also adds a regression test to `tests/e2e/Services/Users/UsersBase.php` that hits `POST /v1/users` with `"name": null` and asserts a `201` response with an empty `name` in the body.

## Scope

Intentionally narrow: only the eight `CreateUser` endpoints called out in the issue title. The `CreateTarget`, `UpdateTarget`, and `CreateToken` endpoints mentioned in follow-up comments are left for separate PRs so this change stays focused and easy to review.

## Testing

- Added `testCreateUser` regression assertion for null `name` via `POST /v1/users`.
- `ID`, `Nullable`, and `Text` imports are already present; no new imports required.

Closes #8785

### DIFF
--- a/app/controllers/api/users.php
+++ b/app/controllers/api/users.php
@@ -73,8 +73,9 @@ use Utopia\Validator\Text;
 use Utopia\Validator\WhiteList;
 
 /** TODO: Remove function when we move to using utopia/platform */
-function createUser(Hash $hash, string $userId, ?string $email, ?string $password, ?string $phone, string $name, Document $project, Database $dbForProject, Hooks $hooks): Document
+function createUser(Hash $hash, string $userId, ?string $email, ?string $password, ?string $phone, ?string $name, Document $project, Database $dbForProject, Hooks $hooks): Document
 {
+    $name ??= '';
     $plaintextPassword = $password;
     $passwordHistory = $project->getAttribute('auths', [])['passwordHistory'] ?? 0;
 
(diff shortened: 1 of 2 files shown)


sambhusankar:
While creating a user - User's name can't be null. it suppose to be a required field.

HarshMN2345:

PR #12138, opened 2026-04-27
feat: support out-of-order chunked uploads

TorstenDittmann:
## Summary
This PR enables out-of-order chunked uploads for all file upload endpoints (Storage, Functions, Sites) by leveraging the new `utopia-php/storage` 2.0.0 device-level support.

### API-level fixes
- **Fixed chunk calculation**: Replaced dynamic chunk size inference (which broke when the last/smaller chunk arrived first) with a fixed 5MB chunk size via new `APP_LIMIT_UPLOAD_CHUNK_SIZE` constant, matching official SDK behavior.
- **Removed `-1` last-chunk sentinel**: This sentinel incorrectly treated the last chunk as a special case and broke out-of-order uploads when it arrived first.
- **Fixed duplicate-retry guards**:
  - **Storage**: Returns existing file (`200 OK`) instead of throwing `STORAGE_FILE_ALREADY_EXISTS` when a chunk is retried after upload completion.
  - **Functions & Sites**: Added missing guards to prevent re-triggering finalization (DB updates + build queues) on duplicate chunk retries.

### Dependency
- Upgraded `utopia-php/storage` to `2.0.0` which adds out-of-order assembly support in Local adapter and S3 part sorting.

### Tests
- Added `testCreateBucketFileOutOfOrder` (Storage)
- Added `testCreateDeploymentOutOfOrder` (Functions)
- Added `testCreateDeploymentOutOfOrder` (Sites)

Each test uploads chunks in non-sequential order (last chunk first, then first, then shuffled remainder) and verifies correct assembly.


TorstenDittmann:
@greptile-apps Could you please re-review this PR? The P1 issues you flagged have been addressed in the latest commits. Thanks!

eldadfux (approved):

PR #12165, opened 2026-04-28
fix: coerce non-string header values in Request::getHeader

lohanidamodar:
## Summary

When duplicate headers (notably `Referer`) arrive on a request, Swoole exposes them as an array in `$swoole->header`. `Request::getHeader()` then returns that array, but its strict `: string` return type triggers a `TypeError`.

This crashes anonymous CORS rule resolution because `app/init/resources/request.php` (`$container->set('rule', ...)`) calls `$request->getReferer()` -> `getHeader('referer', '')` for every request without an `Origin`.

## Fix

Defensively coerce the header value:
- if it arrives as an array, take the first element (falling back to `$default`),
- if it is still not a string, return `$default`.

This keeps the `: string` contract intact and prevents the CORS resolution crash.

Closes CLO-4280

## Test plan

- [ ] Send a request with duplicate `Referer` headers and confirm no `TypeError` is thrown.
- [ ] Verify CORS rule resolution proceeds normally for anonymous requests.

### DIFF
--- a/src/Appwrite/Utopia/Request.php
+++ b/src/Appwrite/Utopia/Request.php
@@ -209,7 +209,11 @@ class Request extends UtopiaRequest
     public function getHeader(string $key, string $default = ''): string
     {
         $headers = $this->getHeaders();
-        return $headers[$key] ?? $default;
+        $value = $headers[$key] ?? $default;
+        if (\is_array($value)) {
+            $value = $value[0] ?? $default;
+        }
+        return \is_string($value) ? $value : $default;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Appwrite/Utopia/Request.php
```diff
@@ -209,7 +209,11 @@ public function getHeaders(): array
     public function getHeader(string $key, string $default = ''): string
```
ChiragAgg5k, line 209: should Request object match swoole and return array as well? and let it be application code to handle if they want the first element or not,

or maybe rename the function to getFirstHeader or something?

ChiragAgg5k (approved):

PR #12167, opened 2026-04-28
feat: add query param fallback for impersonation headers

HarshMN2345:
Allow impersonation to be specified via URL query params (?impersonateUserId, ?impersonateEmail, ?impersonatePhone) as a fallback to the existing headers, enabling Console to embed impersonation in direct file/image URLs where headers cannot be set.

## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


app/init/realtime/connection.php
```diff
@@ -327,7 +327,13 @@
             }
         }
 
-        $impersonateUserId = $request->getHeader('x-appwrite-impersonate-user-id', '');
+        // impersonateUserId also accepts a query param to support embedding in WebSocket URLs.
+        // Email and phone are intentionally header-only to avoid PII exposure in proxy/LB logs.
+        // Query-param fallback is blocked for cross-site requests to prevent CSRF attacks via
+        // third-party pages; Sec-Fetch-Site is a browser-enforced forbidden header.
+        $fetchSite = $request->getHeader('sec-fetch-site', '');
+        $isSameOrigin = \in_array($fetchSite, ['same-origin', 'same-site'], true);
```
Meldiron: What is protection we get by this?

app/init/realtime/connection.php
Meldiron, line 1: Les include E2E test for impersonation with url query - specifically useful for dowloading deployment source or output.

If we dont have any impersonation tests yet, also worth adding some basic ones - impersonating user, seing his account details, his organizations, his projects, and data in his project

app/init/realtime/connection.php
HarshMN2345: It helps prevent CSRF attacks.

Without this check, a malicious site could load something like
``` ```and your browser would automatically send your session cookies along with it, without you realizing.

Sec-Fetch-Site is enforced by the browser (so it can’t be faked with JavaScript). We use it to make sure this parameter is only accepted when the request is coming from the same site.

HarshMN2345:
@greptile session cookie scoping is sufficient protection for this capability.

HarshMN2345:
@greptile only trusted team members explicitly granted the impersonator flag can use it.

eldadfux (approved):

PR #12171, opened 2026-04-28
Refactor migrations API to module style

premtsd-code:
## Summary

Migrates the migrations HTTP API from the legacy procedural controller (`app/controllers/api/migrations.php`) to the modular module/service/action structure used by the rest of the platform. No behavior changes — pure refactor.

- New module at `src/Appwrite/Platform/Modules/Migrations/` (Module + `Services/Http.php` + 16 Action classes).
- Routes mirror URL paths: `Http/Migrations/{Get,XList,Update,Delete}.php` for top-level CRUD; per-source folders (`Appwrite/`, `Firebase/`, `Supabase/`, `NHost/`) for create + report; format folders (`CSV/{Imports,Exports}/`, `JSON/{Imports,Exports}/`) for import/export. The `/v1/migrations/csv` alias is preserved via `httpAlias()`.
- The two helper functions in the old file became small private static methods on the actions that need them, per the project's "no global functions / small duplication is preferable" guideline.
- Module registered in `src/Appwrite/Platform/Appwrite.php`; `app/config/services.php` `migrations.controller` set to `''` ("Uses modules").
- Old `app/controllers/api/migrations.php` deleted.

One intentional cleanup: `getSupabaseReport` had `->inject('dbForProject')` but its action signature didn't accept that argument — a dead inject. Dropped in the new version. No observable change.

## Test plan

- [x] PHP syntax check passes on all 18 new files plus the 2 modified files.
- [x] Pint (PSR-12) passes on all 20 files.
- [x] Programmatic verification by booting the `Migrations\Module` and enumerating registered actions:
  - All 16 expected routes register with correct HTTP method and path.
  - `/v1/migrations/csv` alias is registered on `createCSVImport`.
  - Param names + order match the original controller exactly for all 16 actions.
  - Inject names + order match the original controller exactly for all 16 actions.
  - Every callback is callable; action signature arity equals `count(params) + count(injects)`.
- [ ] Reviewer to spot-check at least one source-create + one CSV/JSON endpoint live (e.g. `POST /v1/migrations/appwrite`, `POST /v1/migrations/csv/imports`, `GET /v1/migrations`, `DELETE /v1/migrations/:id`).
- [ ] Reviewer to confirm SDK regen produces the same method signatures (`createAppwriteMigration`, `getAppwriteReport`, `createCSVImport`, etc.).


src/Appwrite/Platform/Modules/Migrations/Http/Migrations/XList.php
```diff
@@ -0,0 +1,104 @@
...
+            ))
+            ->param('queries', [], new Migrations(), 'Array of query strings generated using the Query class provided by the SDK. [Learn more about queries](https://appwrite.io/docs/databases#querying-documents). Maximum of ' . APP_LIMIT_ARRAY_PARAMS_SIZE . ' queries are allowed, each ' . APP_LIMIT_ARRAY_ELEMENT_SIZE . ' characters long. You may filter on the following attributes: ' . implode(', ', Migrations::ALLOWED_ATTRIBUTES), true)
+            ->param('search', '', new Text(256), 'Search term to filter your list results. Max length: 256 chars.', true)
+            ->param('total', true, new Boolean(true), 'When set to false, the total count returned will be 0 and will not be calculated.', true)
+            ->inject('response')
+            ->inject('dbForProject')
+            ->callback($this->action(...));
+    }
+
+    public function action(array $queries, string $search, bool $includeTotal, Response $response, Database $dbForProject): void
```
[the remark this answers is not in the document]
premtsd-code: This is same pattern across multiple endpoints

app/config/services.php
```diff
@@ -286,7 +286,7 @@
         'name' => 'Migrations',
```
ArnabChatterjee20k, line 286: I think we need to remove this file?

premtsd-code: @ArnabChatterjee20k Removing the entries from this file will create issues in sdk generation.
Other modules still have entries in it

abnegate (approved):

PR #12180, opened 2026-04-29
feat: wire SQLite database adapter for tests

abnegate:
## Summary

Wires `Utopia\Database\Adapter\SQLite` (already shipped in `utopia-php/database` 5.3.22) as a valid `_APP_DB_ADAPTER` value, so the e2e test stack can run without MariaDB/MongoDB/PostgreSQL containers — a single shared `/storage/sqlite/appwrite.db` file replaces them.

This is a draft to surface CI timing data: SQLite vs MongoDB on the same workflow run, per service.

## Why not the in-memory adapter?

`feat-memory-adapter` (utopia-php/database 5.5.0) was the first thing I tried, but it stores data in per-process PHP arrays. Appwrite's HTTP container forks N Swoole workers, plus 14 separate worker containers — none of them share memory, so writes from one process are invisible to the next. Adding a Swoole-Table backing isn't viable either: Tables are flat, fixed-schema, fixed-size hash tables of int/float/string only, can't be allocated after `Server::start()`, and have no nested types or cross-table atomicity. Stacking the workarounds adds up to a from-scratch rewrite, not a field-type swap. SQLite over a shared file gives real cross-process semantics for free.

## Wiring

- `app/init/registers.php`: `'sqlite'` added to scheme allowlists for `console`/`database`/`logs`; resource factory builds a `PDOProxy` over `PDO("sqlite:{path}")` with `PRAGMA journal_mode=WAL`, `busy_timeout=5000`, `foreign_keys=ON`; `'sqlite' => new SQLite($resource())` in the adapter match; matching case in the worker-scope `db` register.
- `docker-compose.yml` + `tests/resources/docker/docker-compose.yml`: `_APP_DB_SQLITE_PATH` env passthrough and `appwrite-sqlite:/storage/sqlite:rw` volume mount on every DB-touching service; top-level volume declared.
- `.env`: `_APP_DB_ADAPTER=sqlite`, `_APP_DB_SQLITE_PATH=/storage/sqlite/appwrite.db`. `_APP_DB_HOST=mongodb` is kept as a placeholder so `depends_on: ${_APP_DB_HOST:-mongodb}` resolves to a real running service. `mariadb` dropped from `COMPOSE_PROFILES`.
- `.github/workflows/ci.yml`: `SQLite` added to `allDatabases` and `defaultDatabases`. New env-setup branch sets `_APP_DB_ADAPTER=sqlite`, `_APP_DB_SQLITE_PATH`, and pins `_APP_DB_HOST=mongodb` for the `depends_on` placeholder.

`documentsdb` (Mongo-only) and `vectorsdb` (Postgres-only) are unchanged — broadening their scheme allowlists is a separate task. The Mongo container still starts because `documentsdb` needs it.

## Test plan

- [ ] `Tests / E2E / SQLite (dedicated) / *` matrix rows pass
- [ ] Side-by-side per-service timings against the matching `Tests / E2E / MongoDB (dedicated) / *` rows
- [ ] No regressions in MongoDB rows (i.e. wiring change doesn't break the default path)


abnegate:
Closing after triage: Self-described as a draft to surface CI timing data, and it drifted into a Redis-adapter experiment. It pins utopia-php/database to dev-feat-redis-adapter, a branch that no longer exists (as does the feat-sqlite-fts branch it pulled fixes from), so the lockfile cannot be resolved. 49 checks failed a

PR #12183, opened 2026-04-29
fix: rename GitHub OAuth2 label from App ID to Client ID

### LINKED ISSUE #11875: Change GitHub OAuth2 settings label from App ID to Client ID
### 🔖 Enhancement description

In appwrite auth, when adding Github as auth provider the ID of application showing as App ID. But in Github Oauth Apps it named as Client ID. In addition in Github Usual Apps there's an additional App ID as well apart from Client ID. So users might confuse with this. So changing the this App ID to Client ID would be better user experience. 

### 🎤 Pitch

Changing App ID to Client ID for Github Auth App setting will convey users directly so that they won't be confused as well as won't misinterpret with Github Apps App ID.

dhanushkumar-amk:
## What does this PR do?
Renames the OAuth2 "App ID" description text to "Client ID" across 
API controller and OpenAPI spec files, to match GitHub's own 
developer settings terminology.

## Related issue
Fixes #11875

## Files changed
- app/controllers/api/projects.php
- src/Appwrite/Utopia/Response/Model/AuthProvider.php
- app/config/specs/open-api3-latest-console.json
- app/config/specs/swagger2-latest-console.json

## Type of change
- [x] Label/description text fix (no logic changed, no key renames)


dhanushkumar-amk:
## What does this PR do?
Renames the OAuth2 "App ID" description text to "Client ID" to match 
GitHub's own developer settings terminology.

## Files changed
- app/controllers/api/projects.php
- src/Appwrite/Utopia/Response/Model/AuthProvider.php

## Related issue
Fixes #11875

## Type of change
- [x] Label/description text fix (no logic changed)

PR #12194, opened 2026-05-01
feat(insights): add insights module with CTA framework

abnegate:
## Summary

New generic **Insights** module — a flexible base for Lighthouse-style reports across resource types (database index suggestions, performance audits, accessibility audits, SEO audits, function performance, …) with a CTA framework so the user can trigger a server-side action directly from a report.

This is the server-ce foundation that cloud will build on for the dedicated-database index analyzer (currently bespoke in `dedicatedDatabaseIndexSuggestions`), and that future surfaces (sites Lighthouse, function profiling) can also produce against.

### Module
Newer module style in `src/Appwrite/Platform/Modules/Insights/`:
- `Module.php`, `Services/Http.php`
- 5 endpoints under `Http/Insights/`: `Create`, `Get`, `XList`, `Update`, `Delete`
- 1 endpoint under `Http/CTA/Execution/`: `Create` (POST `/v1/insights/:insightId/ctas/:ctaId/executions`)

Each endpoint has full side things: scopes, events, audits, SDK Method, abuse-limit/abuse-time/abuse-key, realtime emission. Dismissal is a `status` param on `Update` (`active` / `dismissed`); `dismissedAt` and `dismissedBy` are recorded automatically on the transition.

### CTA framework
- `Appwrite\Insights\CTA\Action` — interface (`getName(): string`, `execute(...): Document`)
- Registry kept on the global `$register` (`Utopia\Registry\Registry`) so it survives across requests; resolved via DI (`inject('insightCTARegistry')`)
- `Appwrite\Insights\CTA\Action\Databases\Indexes\Create` — first concrete CTA. Extends the public `databases.createIndex` HTTP action and reuses its core via the new `createIndex(...)` method on the abstract `Indexes\Action`, so there's exactly one source of truth for index validation, persistence, and queue dispatch
- `Appwrite\Insights\Validator\CTAs` and `Appwrite\Insights\Validator\CTA\Databases\Index\Create` — input validators for the CTA descriptors and the create-index params

### Schema
`insights` collection in `app/config/collections/projects.php`:
- `type` (enum: `databaseIndex | databasePerformance | sitePerformance | siteAccessibility | siteSeo | functionPerformance | …`)
- `severity` (`info | warning | critical`)
- `status` (`active | dismissed`)
- `resourceType`, `resourceId`, `resourceInternalId`
- `title`, `summary` (markdown), `payload` (json), `ctas` (json array)
- `analyzedAt`, `dismissedAt`, `dismissedBy`
- Indexes on `(resourceType, resourceId, $sequence)`, `(type)`, `(dismissedAt)`, `(status)`

### Wiring
Errors, scopes (`insights.read`, `insights.write`), events, roles, response models (`Insight`, `InsightCTA`, `InsightCTAExecution`), realtime channels (`insights`, `insights.{insightId}`), constants — all registered.

## Decisions for review

- **`InsightList` uses `BaseList`** (no dedicated model file). Matches the `MigrationList` / `SubscriberList` pattern in `app/init/models.php`.
- **Dismissal is collapsed into `status`** on the update endpoint rather than a dedicated `Dismiss` sub-resource — same shape as other state transitions in the codebase. Setting `status = dismissed` records `dismissedAt` + `dismissedBy`; setting it back to `active` clears them.
- **CTA action handler shares the public create-index body**. Rather than duplicating index validation/persistence logic, the CTA's `execute()` extends the CE `Databases\Collections\Indexes\Create` action and calls a new `createIndex(...)` helper extracted onto the abstract `Indexes\Action`. The Utopia route registration in `__construct()` is bypassed for the CTA path (the dispatcher invokes `execute()` directly).
- **Realtime channels** `insights` and `insights.{insightId}` go to project team members (team role), not per-document permissions. Matches the console pattern.

## Test plan

- [x] `composer lint` (Pint) clean
- [x] `composer analyze` (PHPStan) — 0 errors
- [x] `vendor/bin/phpunit tests/unit/Insights` — 14 tests, 22 assertions, all green
- [x] e2e suite split into `tests/e2e/Services/Insights/InsightsBase.php` (shared cases) + `InsightsCustomServerTest.php` and `InsightsCustomConsoleTest.php` overlays
- [x] CI workflow updated to run the Insights e2e suite
- [ ] Reviewer to spot-check the CE `databases.createIndex` action — body is now thin (delegates to `$this->createIndex(...)`) and the same helper is exercised via the CTA path


app/config/collections/projects.php
```diff
@@ -2754,4 +2754,176 @@
...
+                'type' => Database::VAR_STRING,
+                'size' => Database::LENGTH_KEY,
+                'signed' => true,
+                'required' => true,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('resourceInternalId'),
```
eldadfux: We might need a parent id and type too or follow our pointer pattern for resources like a file in a bucket.

app/config/collections/projects.php
```diff
@@ -2754,4 +2754,176 @@
             ],
         ],
     ],
+
+    'insights' => [
+        '$collection' => ID::custom(Database::METADATA),
+        '$id' => ID::custom('insights'),
+        'name' => 'Insights',
+        'attributes' => [
```
eldadfux: I think for more complex setups we might need a parent resource for insights, like a report. When we have a run of lighthouse on something like https://appwrite.io/ it is a report for a specific page by a specific analyzer (or whatever we want to call it) that produces multiple insights. A report might also have categories, but we can also flatten it and have a report for performance and accessibility vs something more generic like audit with performance and accessibility as categories.

app/config/collections/projects.php
```diff
@@ -2754,4 +2754,176 @@
             ],
         ],
     ],
+
+    'insights' => [
```
eldadfux: Wondering if this can be platform level vs project level 🤔
Is the information here valuable at the region level? platform level can reduce cost of storage and compute  (less pods/containers)

app/init/models.php
```diff
@@ -505,6 +509,9 @@
 Response::setModel(new Migration());
 Response::setModel(new MigrationReport());
 Response::setModel(new MigrationFirebaseProject());
+Response::setModel(new Insight());
+Response::setModel(new InsightCTA());
```
eldadfux, line 524: How are these two different? can execution be part of the CTA?

src/Appwrite/Insights/CTA/Action/Databases/Indexes/Create.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
```
eldadfux: I was imagining the CTAs as models that hold metadata to allow us to create actions in the console. With this design we're going to mix and match write operations from all Appwrite services together and maintain APIs that do the same thing but have unclear signatures. This will also make such API harder to scale if needed as each endpoint will have very unique characteristics. This will also be in theory an endless list. Block user, delete file, update message, update build settings etc..

src/Appwrite/Platform/Modules/Insights/Http/Manager/Insights/Create.php
```diff
@@ -0,0 +1,136 @@
...
+use Utopia\Platform\Scope\HTTP;
+use Utopia\Validator\JSON;
+use Utopia\Validator\Nullable;
+use Utopia\Validator\Text;
+use Utopia\Validator\WhiteList;
+
+class Create extends Action
+{
+    use HTTP;
+
```
eldadfux: This is interesting. I can see how this might be useful for 3rd party partners to create extensions to Appwrite, some sort of Appwrite apps, but I don't see how it's useful with a project key.

src/Appwrite/Utopia/Response/Model/InsightCTAExecution.php
```diff
@@ -0,0 +1,54 @@
...
+namespace Appwrite\Utopia\Response\Model;
+
+use Appwrite\Utopia\Response;
+use Appwrite\Utopia\Response\Model;
+
+class InsightCTAExecution extends Model
+{
+    public function __construct()
+    {
+        $this
```
eldadfux: With read-only insights this might not be needed.

abnegate:
Restructured the PR per review. Pushed as `00565ea471`.

**Design changes (Eldad):**

1. **CTAs are metadata-only now** — dropped the entire server-side execution layer (`Appwrite\Insights\CTA\Action` interface, registry, `databases.indexes.create` concrete action, params validator, `/v1/insights/:id/ctas/:id/executions` endpoint, `InsightCTAExecution` model, `INSIGHT_CTA_*` errors and events). CTAs are now just descriptors (`id`, `label`, `action`, `params`) — the console invokes the existing public API directly with `action` (e.g. `databases.createIndex`) and `params`. No mix-and-match write API, no endless surface to maintain. Restored `Databases\Indexes\Action.php` and `Create.php` to their pre-CTA shape (the `createIndex` helper was added solely for CTA reuse).

2. **Reports collection added** as the parent grouping. New `reports` collection with full CRUD (`Create`, `Get`, `XList`, `Update`, `Delete`), `Report` response model, `Reports` query validator, `REPORT_NOT_FOUND` / `REPORT_ALREADY_EXISTS` errors, `reports.read` / `reports.write` scopes, `reports.*` event tree. Insights carry an optional `reportId` / `reportInternalId` (nullable for ad-hoc insights). Deleting a report cascades to its child insights. A Lighthouse run becomes a single report with `targetType=urls`, `target=https://appwrite.io/`, and N insights underneath.

3. **Moved to platform DB** (was project DB) — both `insights` and `reports` live in `dbForPlatform` now, scoped per project via `projectInternalId` / `projectId`. List endpoints filter by `projectInternalId`; Get/Update/Delete enforce project ownership on the loaded document, mirroring the `rules`/`certificates` pattern. Reduces per-project schema overhead.

4. **Dropped the execution model + endpoint** as covered by (1).

**Greptile bot fixes (all P1s addressed):**

- Update.php uses `CTAsValidator` (not `ArrayList<JSON> + isset` — empty-string fields can no longer slip through).
- Update.php mutates the loaded document via `setAttribute` (not `new Document($changes)` — no risk of dropping unmodified attributes).
- Create.php and Update.php both reject duplicate CTA `id` values during normalization.
- `CTAs` validator has a configurable `maxCount` (default 16) — Create path now matches Update and the 16384-byte column.
- `Validator\Queries\Insights::ALLOWED_ATTRIBUTES` includes `status` and `reportId` plus a `_key_project_status` index in the schema, so `Query::equal('status', ['active'])` works.
- `Realtime` channel parser guards `$parts[1]` for both `insights` and `reports`.
- The CTA-execution-specific issues (privilege escalation, `INSIGHT_CTA_NOT_FOUND` vs `INSIGHT_CTA_ACTION_NOT_REGISTERED`, `queueForEvents` mutation in `createIndex`, `$db` vs `$collection` authorization-skip mismatch) are moot — those code paths are gone.

Lint passes. PHPStan only flags an unrelated pre-existing error in `Workers/Migrations.php` (the in-flight `feat/migration-resource-attrs` refactor on main that this branch hasn't picked up yet).

app/config/collections/projects.php
abnegate: Makes sense, will introduce reports as a parent resource for insights

app/config/collections/projects.php
abnegate: Can't see any reason it needs to be project-level, moving to platform

app/init/models.php
abnegate: Moving to CTA being metadata only, no execution

src/Appwrite/Insights/CTA/Action/Databases/Indexes/Create.php
abnegate: True, moved to metadata only and pushed execution to client side

src/Appwrite/Platform/Modules/Insights/Http/Manager/Insights/Create.php
abnegate: We need a way to submit reports/insights from edge, but makes more sense as a manager endpoint in that case, will move

app/config/collections/platform.php
```diff
@@ -1956,6 +1956,553 @@
...
+                'lengths' => [Database::LENGTH_KEY, 64],
+                'orders' => [],
+            ],
+            [
+                '$id' => ID::custom('_key_project_target'),
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['projectInternalId', 'targetType', 'target'],
+                'lengths' => [Database::LENGTH_KEY, 64, 256],
+                'orders' => [],
+            ],
```
abnegate, line 2121: @copilot Where index lengths sum to less than max (768), or lenths are all asc, just omit the lengths/orders

app/config/collections/platform.php
```diff
@@ -1956,6 +1956,553 @@
...
+    ],
+
+    'insights' => [
+        '$collection' => ID::custom(Database::METADATA),
+        '$id' => ID::custom('insights'),
+        'name' => 'Insights',
+        'attributes' => [
+            [
+                '$id' => ID::custom('projectInternalId'),
+                'type' => Database::VAR_STRING,
```
abnegate: @copilot use VAR_ID for all "internal ID" references

app/config/collections/platform.php
```diff
@@ -1956,6 +1956,553 @@
...
+                '$id' => ID::custom('ctas'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 16384,
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => ['subQueryInsightCTAs'],
+            ],
```
abnegate: @copilot Pretty sure `payload` is redundant because of `ctas`, if so, remove it through the PR

app/config/collections/platform.php
```diff
@@ -1956,6 +1956,553 @@
...
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => ['json'],
+            ],
+            [
+                // Virtual attribute — CTAs live in the `insightCTAs` collection
+                // back-referenced by `insightInternalId`. The subQuery filter
+                // joins them in at read time, so consumers still see them
+                // embedded on the insight response.
```
abnegate:
```suggestion
```

app/config/collections/platform.php
abnegate: @copilot Do it all for all collections touched in this PR

app/config/collections/platform.php
```diff
@@ -1956,6 +1956,553 @@
         'attributes' => [],
         'indexes' => []
     ],
+
+    'reports' => [
```
abnegate, line 1960: @copilot Reports should also include their nested insights, fetched via subquery, same as ctas

app/config/scopes/project.php
```diff
@@ -361,4 +361,28 @@
...
+        'category' => 'Other',
+    ],
+    'insights.write' => [
+        'description' => 'Access to update, dismiss, and delete insights.',
+        'category' => 'Other',
+    ],
+    'insights.manager' => [
+        'description' => 'Internal-only: ingest insights produced by Appwrite analyzers (edge, executor, …). Not granted to user roles.',
+        'category' => 'Other',
+    ],
```
abnegate: @copilot Remove this, manager ingest will live on cloud, and will just use insights.write scope

app/config/services.php
```diff
@@ -308,5 +308,19 @@
...
+        'description' => '/docs/services/insights.md',
+        'controller' => '', // Uses modules
+        'sdk' => true,
+        'docs' => true,
+        'docsUrl' => 'https://appwrite.io/docs/server/insights',
+        'tests' => true,
+        'optional' => true,
+        'icon' => '/images/services/insights.png',
+        'platforms' => ['server', 'console'],
+    ],
```
abnegate: @copilot The product/service name should `advisor`, the APIs are `reports` and `insights`. So here should be `advisor`. Make sure SDK namespaces use `advisor` too

app/init/database/filters.php
```diff
@@ -475,3 +475,17 @@ function (mixed $value, Document $document, Database $database) {
...
+
+Database::addFilter(
+    'subQueryInsightCTAs',
+    function (mixed $value) {
+        return;
+    },
+    function (mixed $value, Document $document, Database $database) {
+        return $database->getAuthorization()->skip(fn () => $database
+            ->find('insightCTAs', [
+                Query::equal('insightInternalId', [$document->getSequence()]),
```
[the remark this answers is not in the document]
abnegate: @copilot Add project internal ID to the filter, either db namespace or tenant, depending on shared tables or not

app/init/constants.php
abnegate, line 1: @copilot Use actual enums where applicable for all changes in this file

src/Appwrite/Platform/Modules/Insights/Http/Manager/Insights/Create.php
abnegate, line 1: @copilot This file should be in the cloud PR

src/Appwrite/Platform/Modules/Insights/Http/Reports/Create.php
abnegate, line 1: @copilot Reports/insights/CTA should be readonly except for the manager create report endpoint (which accepts nested insights + CTAs using utopia request models)

src/Appwrite/Platform/Modules/Insights/Http/Reports/Delete.php
```diff
@@ -0,0 +1,109 @@
...
+        $childInsights = $dbForPlatform->find('insights', [
+            Query::equal('projectInternalId', [$project->getSequence()]),
+            Query::equal('reportInternalId', [$report->getSequence()]),
+            Query::limit(APP_LIMIT_COUNT),
+        ]);
+
+        foreach ($childInsights as $insight) {
+            // Cascade through CTAs first.
+            $childCTAs = $dbForPlatform->find('insightCTAs', [
+                Query::equal('insightInternalId', [$insight->getSequence()]),
```
abnegate: @copilot Nested resource delete should happen in the deletes worker, add a new delete type, TYPE_REPORT, and only delete the report here

abnegate: @copilot I guess we need to keep the report delete endpoint

app/config/collections/platform.php
```diff
@@ -1956,6 +1956,556 @@
...
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('reportInternalId'),
+                'type' => Database::VAR_ID,
+                'format' => '',
+                'size' => 0,
+                'signed' => true,
+                'required' => false,
+                'default' => '',
```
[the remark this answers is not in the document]
abnegate: @copilot Set all VAR_ID props to required with default null

src/Appwrite/Platform/Modules/Advisor/Services/Http.php
```diff
@@ -0,0 +1,25 @@
...
+    {
+        $this->type = Service::TYPE_HTTP;
+
+        $this->addAction(GetReport::getName(), new GetReport());
+        $this->addAction(ListReports::getName(), new ListReports());
+        $this->addAction(DeleteReport::getName(), new DeleteReport());
+
+        $this->addAction(GetInsight::getName(), new GetInsight());
+        $this->addAction(ListInsights::getName(), new ListInsights());
+    }
```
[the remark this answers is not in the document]
abnegate: @copilot Update the PR desc, code is correct

app/config/collections/platform.php
```diff
@@ -1956,6 +1956,556 @@
...
+            [
+                '$id' => ID::custom('_key_project_type'),
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['projectInternalId', 'type'],
+                'lengths' => [],
+                'orders' => [],
+            ],
+            [
+                '$id' => ID::custom('_key_project_severity'),
+                'type' => Database::INDEX_KEY,
```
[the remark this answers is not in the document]
abnegate: @copilot Make PARENT resource VAR_ID props optional default null

app/config/scopes/project.php
```diff
@@ -361,4 +361,20 @@
...
+        'description' => 'Access to delete advisor reports. Also reserved for advisor ingestion outside CE.',
+        'category' => 'Other',
+    ],
+
+    // Reports
+    'reports.read' => [
+        'description' => 'Access to read analyzer reports and their insights.',
+        'category' => 'Other',
+    ],
 ];
```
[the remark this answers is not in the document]
abnegate: @copilot Add the missing scope

app/config/collections/platform.php
```diff
@@ -1956,6 +1956,556 @@
...
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                // Plural noun for the parent (containing) resource. Optional.
+                // e.g. an insight about a column index → resourceType=indexes,
+                // parentResourceType=tables. Mirrors the file-in-bucket pointer.
+                '$id' => ID::custom('parentResourceType'),
+                'type' => Database::VAR_STRING,
```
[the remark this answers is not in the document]
abnegate: @copilot Insights without a report should not be possible

app/config/collections/platform.php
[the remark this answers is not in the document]
abnegate: Done in b747c97561 — `parentResourceInternalId` is now `required: false, default: null` to match its companion `parentResourceType`/`parentResourceId` fields. Bumped `APP_CACHE_BUSTER` for the schema change.

app/config/collections/platform.php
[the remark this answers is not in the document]
abnegate: Schema is unchanged (`required: true`) per your direction. Updated the `Insight` response model description in b747c97561 to drop the "Empty for ad-hoc insights" wording — it now reads "Insights always belong to a report."

app/config/scopes/project.php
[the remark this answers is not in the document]
abnegate: Done in b747c97561. Registered `reports.write` in `scopes/project.php`, added it to the admin role list in `roles.php`, switched `deleteReport` to `->label('scope', 'reports.write')`, and updated `insights.write` description to "Reserved for advisor insight ingestion outside CE". Also added `reports.write` to the default test API key in `ProjectCustom` so existing e2e tests can hit the delete endpoint.

app/config/collections/platform.php
```diff
@@ -1956,6 +1956,557 @@
...
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['projectInternalId', 'type'],
+                'lengths' => [],
+                'orders' => [],
+            ],
+            [
+                '$id' => ID::custom('_key_project_target'),
+                'type' => Database::INDEX_KEY,
+                'attributes' => ['projectInternalId', 'targetType', 'target'],
+                'lengths' => [0, 0, 700],
```
abnegate: @copilot use null instead of 0 to skip indices

app/config/scopes/project.php
```diff
@@ -361,4 +361,24 @@
...
+    // Reports
+    'reports.read' => [
+        'description' => 'Access to read analyzer reports and their insights.',
+        'category' => 'Other',
+    ],
+    'reports.write' => [
+        'description' => 'Access to delete analyzer reports.',
+        'category' => 'Other',
+    ],
 ];
```
abnegate, line 382: @copilot Use consistent terminology, same as other scopes. the service is called `advisor`.

app/config/services.php
```diff
@@ -308,5 +308,19 @@
         'optional' => true,
         'icon' => '/images/services/messaging.png',
         'platforms' => ['client', 'server', 'console'],
-    ]
+    ],
+    'advisor' => [
+        'key' => 'advisor',
+        'name' => 'Advisor',
+        'subtitle' => 'The Advisor service surfaces actionable reports about your project resources, with CTA descriptors for one-click remediation in the console.',
+        'description' => '/docs/services/insights.md',
```
abnegate: @copilot Make sure all doc files are created (including endpoint docs)

app/init/constants.php
```diff
@@ -1,6 +1,12 @@
 <?php
 
 use Appwrite\Platform\Modules\Compute\Specification;
+use Appwrite\Platform\Modules\Insights\Enums\InsightCTAMethod;
+use Appwrite\Platform\Modules\Insights\Enums\InsightCTAService;
+use Appwrite\Platform\Modules\Insights\Enums\InsightSeverity;
+use Appwrite\Platform\Modules\Insights\Enums\InsightStatus;
+use Appwrite\Platform\Modules\Insights\Enums\InsightType;
+use Appwrite\Platform\Modules\Insights\Enums\ReportType;
```
abnegate: @copilot Drop the insight prefix, we're already in the insights namespace. Remember not to double up namespace into file name in future. Alias at use-site instead

app/init/constants.php
```diff
@@ -424,6 +431,72 @@
...
+// Insight types — engine-specific so the CTA action can reference the right public API.
+const INSIGHT_TYPE_DATABASE_INDEX = InsightType::DATABASE_INDEX->value;         // legacy databases.createIndex
+const INSIGHT_TYPE_TABLES_DB_INDEX = InsightType::TABLES_DB_INDEX->value;        // tablesDB.createIndex
+const INSIGHT_TYPE_DOCUMENTS_DB_INDEX = InsightType::DOCUMENTS_DB_INDEX->value;  // documentsDB.createIndex
+const INSIGHT_TYPE_VECTORS_DB_INDEX = InsightType::VECTORS_DB_INDEX->value;      // vectorsDB.createIndex
+const INSIGHT_TYPE_DATABASE_PERFORMANCE = InsightType::DATABASE_PERFORMANCE->value;
+const INSIGHT_TYPE_SITE_PERFORMANCE = InsightType::SITE_PERFORMANCE->value;
+const INSIGHT_TYPE_SITE_ACCESSIBILITY = InsightType::SITE_ACCESSIBILITY->value;
+const INSIGHT_TYPE_SITE_SEO = InsightType::SITE_SEO->value;
+const INSIGHT_TYPE_FUNCTION_PERFORMANCE = InsightType::FUNCTION_PERFORMANCE->value;
```
abnegate: @copilot Just use the values directly, unless there's a good reason to map them?

src/Appwrite/Platform/Modules/Insights/Http/Insights/Get.php
```diff
@@ -0,0 +1,82 @@
...
+            ->setHttpPath('/v1/reports/:reportId/insights/:insightId')
+            ->desc('Get insight')
+            ->groups(['api', 'insights'])
+            ->label('scope', 'insights.read')
+            ->label('resourceType', RESOURCE_TYPE_INSIGHTS)
+            ->label('sdk', new Method(
+                namespace: 'advisor',
+                group: 'insights',
+                name: 'get',
+                description: <<<EOT
```
abnegate: @copilot method name should include resource type if not top level; `advisor.get` doesn't make sense if it returns insights. Use `advisor.getInsight`

src/Appwrite/Platform/Modules/Insights/Http/Insights/Get.php
```diff
@@ -0,0 +1,82 @@
...
+
+        $insight = $dbForPlatform->getDocument('insights', $insightId);
+
+        if (
+            $insight->isEmpty()
+            || $insight->getAttribute('projectInternalId') !== $project->getSequence()
+            || $insight->getAttribute('reportInternalId') !== $report->getSequence()
+        ) {
+            throw new Exception(Exception::INSIGHT_NOT_FOUND);
+        }
```
abnegate, line 80: @copilot We'll already have the insights from the subquery join on report, no need to refetch. Use `document->find` to pull it out

abnegate (commented):
@copilot Tests should be in tests/e2e/Services/Advisor and should have tests for both reports and insights

src/Appwrite/Platform/Modules/Insights/Http/Insights/Get.php
abnegate: @copilot Apply the same pattern to all endpoints not just this one

abnegate: @copilot All endpoints in the PR.

src/Appwrite/Platform/Modules/Insights/Http/Insights/Get.php
```diff
@@ -0,0 +1,92 @@
...
+
+        $response->dynamic($insight, Response::MODEL_INSIGHT);
+    }
+
+    /**
+     * Resolve a nested insight document from a report's subquery payload.
+     */
+    private function getInsightFromReport(Document $report, string $insightId): ?Document
+    {
+        $insight = $report->find('$id', $insightId, 'insights');
```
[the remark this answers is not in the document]
abnegate: @copilot Address

app/init/constants.php
```diff
@@ -1,5 +1,11 @@
 <?php
 
+use Appwrite\Platform\Modules\Advisor\Enums\InsightCTAMethod as AdvisorCTAMethod;
+use Appwrite\Platform\Modules\Advisor\Enums\InsightCTAService as AdvisorCTAService;
+use Appwrite\Platform\Modules\Advisor\Enums\InsightSeverity as AdvisorSeverity;
+use Appwrite\Platform\Modules\Advisor\Enums\InsightStatus as AdvisorStatus;
+use Appwrite\Platform\Modules\Advisor\Enums\InsightType as AdvisorType;
+use Appwrite\Platform\Modules\Advisor\Enums\ReportType as AdvisorReportType;
 use Appwrite\Platform\Modules\Compute\Specification;
```
abnegate: @copilot These shouldn't use Advisor prefix because they're specific to insight/report resources

src/Appwrite/Platform/Modules/Advisor/Http/Insights/XList.php
```diff
@@ -0,0 +1,122 @@
...
+        bool $includeTotal,
+        Response $response,
+        Document $project,
+        Database $dbForPlatform
+    ) {
+        $report = $dbForPlatform->getDocument('reports', $reportId);
+
+        if ($report->isEmpty() || $report->getAttribute('projectInternalId') !== $project->getSequence()) {
+            throw new Exception(Exception::REPORT_NOT_FOUND);
+        }
```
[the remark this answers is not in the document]
abnegate: @copilot use subquery bypass via "skipFilters" on the db call

src/Appwrite/Platform/Modules/Advisor/Http/Insights/Get.php
```diff
@@ -0,0 +1,103 @@
...
+
+        if ($report->isEmpty() || $report->getAttribute('projectInternalId') !== $project->getSequence()) {
+            throw new Exception(Exception::REPORT_NOT_FOUND);
+        }
+
+        $insight = $this->getInsightFromReport($report, $insightId, $dbForPlatform);
+
+        if ($insight === null) {
+            throw new Exception(Exception::INSIGHT_NOT_FOUND);
+        }
```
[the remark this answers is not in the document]
abnegate: @copilot  use subquery bypass via "skipFilters" on the db call

PR #12195, opened 2026-05-01
feat(notifications): add generic notifications worker alongside Mails

abnegate:
## Summary

Adds a generic, multi-channel **Notifications** worker driven by `utopia-php/messaging`, with custom Provider adapters for **Console** (in-app, persisted to `notifications`) and **Webhook** (HMAC-signed POST). Console and Webhook follow the implicit-global pattern — same shape as Internal SMS — no public Provider CRUD routes.

This unblocks shifting webhooks and other system-level alerts into a single dispatch path and decoupling user-facing messaging from system notifications. The Notifications worker is added **alongside** the existing Mails worker; the swap of internal callers, the retirement of `bin/worker-mails`, and the deletion of `Event\Mail` / `Workers\Mails` will land as a follow-up PR (kept on the local backup branch `backup/feat-notifications-worker-with-swap`) so this change is purely additive and easy to review.

### New event + worker
- `src/Appwrite/Event/Notification.php` — multi-channel event with the same `setSubject` / `setRecipient` / `setBody` surface as the legacy `Mail` event, plus channel-specific recipients (email, console user IDs, webhook URLs)
- `src/Appwrite/Platform/Workers/Notifications.php` — routes per-channel via the messaging adapters; persists `notifications` rows for the console channel; deduplicates deliveries via a hashed `deduplicationKey`

### Messaging adapters
Located under `src/Appwrite/Utopia/Messaging/` to match the established `Appwrite\Utopia\*` extension pattern:
- `Adapter/Console.php` + `Messages/Console.php` — writes a `notifications` row per recipient
- `Adapter/Webhook.php` + `Messages/Webhook.php` — HMAC-signed POST. Signed with the per-recipient `signatureKey` already on the `webhooks` row (matches the existing Webhooks worker pattern), so no global `_APP_NOTIFICATIONS_WEBHOOK_SECRET` env var. HTTP done through `Utopia\Fetch\Client`, not curl

### Schema
- `notifications` collection in `app/config/collections/platform.php` (matches the cloud platform-DB convention; the database engine prefixes unique keys with `_tenant` internally so cross-tenant `messageId` collisions are not possible)

### Wiring
- `bin/worker-notifications` entrypoint, Dockerfile chmod, `appwrite-worker-notifications` service in `docker-compose.yml` and `tests/resources/docker/docker-compose.yml`
- Registered in `Appwrite\Platform\Services\Workers` as `Notifications::getName()` next to `Mails::getName()`
- `Event::NOTIFICATIONS_QUEUE_NAME` / `NOTIFICATIONS_CLASS_NAME` constants

### Tests
- Unit: `NotificationsTest` (7), `Adapter\ConsoleTest` (4), `Adapter\WebhookTest` (9) — 20 tests / 64 assertions, all green
- E2E: `NotificationsBase` trait + `NotificationsCustomServerTest` and `NotificationsCustomConsoleTest` overlays; pins the operator-visible health contract

## Decisions for review

- **Console + Webhook are implicit/global channels**, configured via env (Webhook) or implicit (Console writes to `notifications`). No public `/v1/messaging/providers/console` or `/webhook` CRUD — same shape as Internal SMS in the messaging worker.
- **Webhook signing uses per-recipient `signatureKey`** rather than a global `_APP_NOTIFICATIONS_WEBHOOK_SECRET`. Each `webhooks` row already carries its own `signatureKey`; the Notifications worker reuses that exactly like the existing Webhooks worker does. Avoids introducing a global secret that adds little real protection.
- **Adapters live under `Appwrite\Utopia\Messaging`** (not `Appwrite\Messaging`) to match the existing convention for Appwrite extensions of Utopia libraries (`Appwrite\Utopia\Database\Validator\…`, `Appwrite\Utopia\Response\…`, etc.).
- **`notifications` lives in `platform.php`** rather than `common.php`. It is a platform-level collection (not per-project), aligning with how cloud already provisions it.
- **Mails worker is left in place** for this PR — the swap is a separate concern and would bloat this change. The follow-up PR will: migrate `queueForMails` callers to `queueForNotifications`, rename `bin/worker-mails` → `bin/worker-notifications` (drop `worker-mails`), and delete `Workers\Mails` + `Event\Mail`.

## Test plan

- [x] `composer lint` (Pint) clean
- [x] `composer analyze` (PHPStan) — 0 errors
- [x] `vendor/bin/phpunit tests/unit/Platform/Workers/NotificationsTest.php tests/unit/Utopia/Messaging` — 20/20 green
- [x] Diff vs `origin/1.9.x` is purely additive (2092 insertions, 0 deletions across 21 files)
- [x] Verified the unit-suite baseline failures (Swoole/Yaml/Imagick/MaxMindDB ext absent locally) are pre-existing — same set on `origin/1.9.x`
- [ ] CI to run e2e with full Docker stack
- [ ] Reviewer to confirm the additive shape (no Mails removal in this PR)


app/init/constants.php
```diff
@@ -254,6 +254,16 @@
...
 const MESSAGE_TYPE_SMS = 'sms';
 const MESSAGE_TYPE_PUSH = 'push';
+// Notification channels
+const NOTIFICATION_CHANNEL_EMAIL = 'email';
+const NOTIFICATION_CHANNEL_SMS = 'sms';
+const NOTIFICATION_CHANNEL_PUSH = 'push';
+const NOTIFICATION_CHANNEL_CONSOLE = 'console';
+const NOTIFICATION_CHANNEL_WEBHOOK = 'webhook';
+// Notification provider types (custom — extend MESSAGE_TYPE_*)
+const MESSAGE_TYPE_CONSOLE = 'console';
```
eldadfux: What the diff between channels and type? each channel is kinda also the type in this case, no?

src/Appwrite/Event/Event.php
```diff
@@ -21,6 +21,9 @@ class Event
     public const MAILS_QUEUE_NAME = 'v1-mails';
     public const MAILS_CLASS_NAME = 'MailsV1';
 
+    public const NOTIFICATIONS_QUEUE_NAME = 'v1-notifications';
```
eldadfux, line 24: I wonder if we want to stick with alerts because this is the collection name already

src/Appwrite/Event/Notification.php
```diff
@@ -0,0 +1,458 @@
...
+
+    public function getSmtpHost(): string
+    {
+        return $this->smtp['host'] ?? '';
+    }
+
+    public function getSmtpPort(): int
+    {
+        return $this->smtp['port'] ?? 0;
+    }
```
eldadfux: I don't think we need to set this for each message, this are internal credentials, not user defined.

src/Appwrite/Event/Notification.php
```diff
@@ -0,0 +1,458 @@
...
+        return $this->customMailOptions['senderEmail'] ?? '';
+    }
+
+    public function setSenderName(string $name): self
+    {
+        $this->customMailOptions['senderName'] = $name;
+        return $this;
+    }
+
+    public function getSenderName(): string
```
eldadfux: Same, I don't see a reason to set this for each message, they can be handled in the worker when type is email.

src/Appwrite/Utopia/Messaging/Messages/Webhook.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
```
eldadfux, line 2: Very good pattern usage here for extending Utopia. A good example for other team members and contributors.

app/config/collections/platform.php
```diff
@@ -1011,6 +1011,132 @@
...
+            [
+                '$id' => ID::custom('body'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 16384,
+                'signed' => true,
+                'required' => true,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
```
eldadfux, line 1217: Let's add read boolean, so we can let users mark read true false or handle it from the console itself.

eldadfux (commented):
@abnegate I would take this a few steps further:

1. I would already put it to use with alerts we send on webhooks and make sure we mostly alert both console and email channels.
2. I would add a simple list endpoint to be able to read user specific console notifications relying on permission for this. This will let me add the notifications to the console on the next release. 
3. Add an ability to track read, from the console easy path method, from emails we can have an image file included in the email (maybe even the logo), when loads we mark the email as read. It can have some simple JWT to ensure only the right request can mark it.

abnegate:
## Review feedback addressed

This pushes the consolidated review-fix work onto the PR head. Mapping each review point to where it landed:

### @greptile-apps P1 fixes
- **P1 #1 — `alreadyDelivered` dedup never matches** (`Notifications.php:114`): `alreadyDelivered()` now queries by the `messageId` *attribute* via `find()` instead of `getDocument($messageId)`, so the compound `$id` produced by `persistAlert` is correctly dedup-checked.
- **P1 #2 — Console adapter `$id` collision** (`Console.php:70`): per-recipient `$id` is now `messageId . '_' . substr(md5('user:' . $userId | 'team:' . $teamId), 0, 8)`. Multi-recipient sends no longer drop on duplicate-key, and idempotent retries land on the same `$id`. Also added a dedicated `catch (DuplicateException)` branch so legitimate duplicates count as delivered, not failures.
- **P1 #3 — Console double-write** (`Notifications.php:291`): action loop skips `persistAlert` for `NOTIFICATION_TYPE_CONSOLE` — the Console adapter owns persistence for that channel.
- **P1 #4 — Console failures swallowed** (`Notifications.php:322`): `dispatchConsole` inspects `$result['deliveredTo']` and throws with the underlying error message when 0, so the queue retries.
- **P1 #5 — `_APP_NOTIFICATIONS_WEBHOOK_SECRET` never read**: env var was already absent on the branch, but the design (per-recipient `signatureKey`) is now the only signing path. Confirmed 0 references repo-wide.

### @eldadfux inline feedback
- **`platform.php:1105` — add `read` boolean**: added `read` boolean (default `false`) to the alerts collection. Composite indexes `_key_userId_read` and `_key_teamId_read` for efficient unread filtering. Also fixed a latent fanout issue: `_key_messageId` is now `INDEX_KEY` (not UNIQUE) and a new UNIQUE composite `_key_recipient` over `(messageId, channel, userId, teamId)` enforces per-recipient idempotency without blocking N-recipient fanout.
- **`Notification.php:187` & `:280` — drop SMTP/sender plumbing from event**: removed all 10 SMTP setters/getters, `customMailOptions`, and the per-message `senderEmail`/`senderName`/`replyTo*` accessors. The Notifications worker now resolves SMTP from the project Document or env at dispatch time, mirroring `Workers/Mails.php`.
- **`constants.php:264` — channel vs type**: consolidated under a single **`NOTIFICATION_TYPE_*`** prefix per discussion. Removed `NOTIFICATION_CHANNEL_*` and the new `MESSAGE_TYPE_CONSOLE`/`MESSAGE_TYPE_WEBHOOK`. Legacy `MESSAGE_TYPE_EMAIL`/`SMS`/`PUSH` stay (used by the existing Messaging worker, out of scope).
- **`Event.php:24` — Alert vs Notification naming**: kept `Notification` per discussion. Collection stays `alerts`, event/worker/queue stay `Notification`.

### Eldad high-level review (now in this PR)
- **(1) Use the worker now — webhook → console + email**: `Webhooks::sendAlert` (renamed from `sendEmailAlert`) switched from `queueForMails` to `queueForNotifications`. On permanent webhook failure, fans out to project owners via BOTH email AND console channels in a single dispatch. Owner-only filter via `in_array('owner', $roles)` (non-owner team members do NOT receive these). Per-call `$queueForNotifications->reset()` at the top of `sendAlert` so multiple failing webhooks in one worker pass don't cross-contaminate. `queueForNotifications` registered in `app/init/worker/message.php` and `app/init/resources/request.php`.
- **(2) List endpoint**: `GET /v1/account/alerts` (`Account/Http/Alerts/XList.php`) — scoped to current user via `Query::equal('userId', [$user->getId()])`, supports cursor pagination, queries (`Validator/Queries/Alerts.php`: read, type, channel, messageId, projectId), returns `MODEL_ALERT_LIST`.
- **(3) Read tracking**:
  - `PATCH /v1/account/alerts/:alertId/read` (`Account/Http/Alerts/Read/Update.php`) — auth-scoped, ownership check (`alert.userId === $user->getId()`) BEFORE the sparse update of `read => true`. 401 for non-owner, 404 for missing alert.
  - Email tracking pixel: `GET /v1/account/alerts/:alertId/track` (`Account/Http/Alerts/Track/Get.php`) with `scope: 'public'`. Decodes a 30-day JWT signed with `_APP_OPENSSL_KEY_V1` (payload `{alertId, userId, purpose: 'alert_track'}`) — `purpose` claim isolates this token from other shared-secret JWTs. On valid + matching alertId + matching userId, uses injected `$authorization->skip(...)` (per repo convention) to sparse-update `read => true`. Always returns a 1×1 transparent PNG with `Content-Type: image/png`, `Cache-Control: no-store`, regardless of decode outcome.
  - Pixel injection: `Notifications::dispatchEmail` pre-computes a deterministic `alertId` via the new `buildAlertId()` helper, builds the JWT, splices `` before the LAST `</body>`, sends the email, and **only persists the alert after a successful adapter `send()`** so transient SMTP failures don't orphan a dedup row.

### Schema, model, DI
- `alerts` schema: `read` boolean + `_key_userId_read` + `_key_teamId_read` + `_key_recipient` UNIQUE composite + `_key_messageId` downgraded to KEY (per above).
- `Alert` and `AlertList` response models registered in `Response.php` and `app/init/models.php` (via `BaseList`).
- Notifications worker now uses `dbForPlatform` (alerts live in the platform DB) — fixes a cross-DB bug where alerts dispatched via a user project would have been written to the project DB and never surfaced through `/v1/account/alerts`.

### Tests
- **Unit (15 tests, 56 assertions in NotificationsTest, plus 5 in ConsoleTest)**: dedup-by-attribute short-circuit, console-skips-`persistAlert`, console-zero-delivery-throws, multi-recipient fanout no-collision, recipient struct round-trips userId/teamId, tracking-pixel injected into email HTML, `persistAlert` returns alertId and stores userId, dedup-returns-existing-alertId-on-duplicate (with the production UNIQUE composite index in `setUp`), Console multi-recipient distinct $id.
- **E2E (6 new tests in `NotificationsBase.php`)**: empty list endpoint, webhook-failure → console alert, mark-read happy path, mark-read unauthorized, tracking-pixel toggles read, tracking-pixel invalid token still returns PNG (no DB write). Tests fail loudly if `_APP_OPENSSL_KEY_V1` is unset (no silent placeholder fallback).

### Verification
- `composer lint`: clean
- `composer analyze` (PHPStan level 3): 0 errors
- All grep gates pass (0 hits for `NOTIFICATION_CHANNEL_*`, `MESSAGE_TYPE_CONSOLE`, `MESSAGE_TYPE_WEBHOOK`, `_APP_NOTIFICATIONS_WEBHOOK_SECRET`, `queueForMails` in `Webhooks.php`, `smtp`/`customMailOptions` in `Notification.php`)
- All previously-passing unit tests stay green; new tests added on top

cc @eldadfux @greptile-apps for re-review.

app/config/collections/platform.php
```diff
@@ -474,6 +474,17 @@
                 'array' => false,
                 'filters' => [],
             ],
+            [
+                '$id' => ID::custom('projectInternalId'),
+                'type' => Database::VAR_STRING,
```
abnegate: @copilot Use VAR_ID here

app/config/collections/platform.php
```diff
@@ -1032,6 +1057,214 @@
...
+                'size' => Database::LENGTH_KEY,
+                'signed' => true,
+                'required' => true,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+            [
+                '$id' => ID::custom('projectInternalId'),
+                'type' => Database::VAR_STRING,
```
abnegate, line 1177: @copilot Use VAR_ID for any internal ID reference

src/Appwrite/Platform/Modules/Notifications/Http/Notifications/Update.php
abnegate, line 1: @copilot This should just be a regular patch endpoint for the alert itself with a parameter `read: boolean`

app/init/constants.php
```diff
@@ -262,6 +262,13 @@
 const MESSAGE_TYPE_EMAIL = 'email';
 const MESSAGE_TYPE_SMS = 'sms';
 const MESSAGE_TYPE_PUSH = 'push';
+// Notification types
+const NOTIFICATION_TYPE_EMAIL = MESSAGE_TYPE_EMAIL;
+const NOTIFICATION_TYPE_SMS = MESSAGE_TYPE_SMS;
+const NOTIFICATION_TYPE_PUSH = MESSAGE_TYPE_PUSH;
+const NOTIFICATION_TYPE_CONSOLE = 'console';
+const NOTIFICATION_TYPE_WEBHOOK = 'webhook';
+const ALERT_TRACKING_JWT_TTL = 60 * 60 * 24 * 7;
```
eldadfux: Why prefix is different here?

app/init/models.php
```diff
@@ -245,6 +246,7 @@
 Response::setModel(new BaseList('Users List', Response::MODEL_USER_LIST, 'users', Response::MODEL_USER));
 Response::setModel(new BaseList('Sessions List', Response::MODEL_SESSION_LIST, 'sessions', Response::MODEL_SESSION));
 Response::setModel(new BaseList('Identities List', Response::MODEL_IDENTITY_LIST, 'identities', Response::MODEL_IDENTITY));
+Response::setModel(new BaseList('Alerts List', Response::MODEL_ALERT_LIST, 'alerts', Response::MODEL_ALERT));
```
eldadfux: Alerts vs notification is very confusing - let's settle on one

app/init/resources.php
```diff
@@ -138,12 +139,16 @@
 ), ['publisher']);
 $container->set('publisherForMails', fn (Publisher $publisher) => new MailPublisher(
     $publisher,
-    new Queue(System::getEnv('_APP_MAILS_QUEUE_NAME', Event::MAILS_QUEUE_NAME))
+    new Queue(System::getEnv('_APP_NOTIFICATIONS_QUEUE_NAME', Event::NOTIFICATIONS_QUEUE_NAME))
```
eldadfux: Why do we need to change this? Regular user emails should still go through this worker.

docs/references/account/get-alert-track.md
```diff
@@ -0,0 +1 @@
+Track an alert as opened by returning a 1x1 transparent tracking pixel. When loaded with a valid tracking token, the alert is automatically marked as read.
```
eldadfux: I feel like this is adding too much to the account scope and also things that are very console specific. Any downside in creating v1/notifications service which is console/admin only?

src/Appwrite/Platform/Modules/Account/Http/Alerts/Track/Get.php
```diff
@@ -0,0 +1,106 @@
...
+    public static function getName(): string
+    {
+        return 'getAlertTrack';
+    }
+
+    public function __construct()
+    {
+        $this
+            ->setHttpMethod(Action::HTTP_REQUEST_METHOD_GET)
+            ->setHttpPath('/v1/account/alerts/:alertId/track')
```
eldadfux: Track verb inside a rest path doesn't look so great

src/Appwrite/Utopia/Messaging/Adapter/Console.php
```diff
@@ -0,0 +1,147 @@
...
+            throw new \Exception('Invalid message type.');
+        }
+
+        return $this->process($message);
+    }
+
+    protected function process(ConsoleMessage $message): array
+    {
+        $response = new Response($this->getType());
+        $delivered = 0;
```
eldadfux, line 58: If we're using this to trigger the message, shouldn't we also make this one send the job to the queue?

abnegate: This is at the bottom of the chain; the worker job is what's calling this code

PR #12210, opened 2026-05-04
Realtime messsage dispatchers (presence api)

ArnabChatterjee20k:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


src/Appwrite/Realtime/Message/Handlers/PresenceHandler.php
```diff
@@ -0,0 +1,117 @@
...
+        );
+
+        $presence->removeAttribute('hostname');
+
+        $realtime->connections[$connection]['presences'] = \array_merge(
+            $realtime->connections[$connection]['presences'] ?? [],
+            [$presence->getId()],
+        );
+
+        $triggerPresenceEvent($server, $realtime, $project, $user, 'presences.[presenceId].upsert', $presence);
```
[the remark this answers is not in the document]
ArnabChatterjee20k: its already fixed

app/realtime.php
```diff
@@ -10,6 +10,12 @@
 use Appwrite\Messaging\Adapter\Realtime;
 use Appwrite\Network\Validator\Origin;
 use Appwrite\PubSub\Adapter\Pool as PubSubPool;
+use Appwrite\Realtime\Message\Dispatcher as MessageDispatcher;
+use Appwrite\Realtime\Message\Handlers\AuthenticationHandler;
+use Appwrite\Realtime\Message\Handlers\PingHandler;
+use Appwrite\Realtime\Message\Handlers\PresenceHandler;
+use Appwrite\Realtime\Message\Handlers\SubscribeHandler;
+use Appwrite\Realtime\Message\Handlers\UnsubscribeHandler;
```
abnegate: Let's remove `Handler` suffix from class names and alias instead, same for validators

src/Appwrite/Realtime/Message/Handlers/Presence.php
```diff
@@ -0,0 +1,114 @@
...
+        );
+
+        $presence->removeAttribute('hostname');
+
+        $realtime->connections[$connection]['presences'] = \array_merge(
+            $realtime->connections[$connection]['presences'] ?? [],
+            [$presence->getId()],
+        );
+
+        $triggerPresenceEvent($project, $user, 'presences.[presenceId].upsert', $presence);
```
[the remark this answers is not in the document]
abnegate: @ArnabChatterjee20k Seems valid

src/Appwrite/Realtime/Message/Validators/SubscribePayload.php
```diff
@@ -0,0 +1,54 @@
...
+            }
+            if (\array_key_exists('queries', $payload)
+                && (!\is_array($payload['queries']) || !\array_is_list($payload['queries']))
+            ) {
+                return 'queries is not a valid array.';
+            }
+        }
+
+        return null;
+    }
```
abnegate, line 68: Doesn't fit patterns, we should keep the logic in `isValid`, and set `$this->description = {msg}` then return false when not valid

src/Appwrite/Realtime/Message/Dispatcher.php
```diff
@@ -0,0 +1,150 @@
...
+     * the entire data value (for list shape). Injections come from the per-message container.
+     *
+     * @return array<int, mixed>
+     */
+    private function resolveArgs(
+        Action $handler,
+        mixed $data,
+        string $shape,
+        Container $container,
+    ): array {
```
abnegate, line 93: Anything in Utopia we can leverage to handle this for us?

src/Appwrite/Realtime/Message/Validators/SubscribePayload.php
ArnabChatterjee20k: I changed it a bit due to this comment https://github.com/appwrite/appwrite/pull/12210/changes#r3181295047
Since coroutines so chances mutating each other

src/Appwrite/Realtime/Message/Handlers/Presence.php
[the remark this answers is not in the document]
ArnabChatterjee20k: updated

src/Appwrite/Realtime/Message/Validators/SubscribePayload.php
ArnabChatterjee20k: updated

src/Appwrite/Realtime/Message/Validators/SubscribePayload.php
```diff
@@ -0,0 +1,56 @@
...
+                && (!\is_array($payload['queries']) || !\array_is_list($payload['queries']))
+            ) {
+                $this->description = 'queries is not a valid array.';
+                return false;
+            }
+        }
+
+        return true;
+    }
+}
```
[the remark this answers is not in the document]
ArnabChatterjee20k: updated, queries are already getting done in the controller level as validation is done for queries during the parsing itself

src/Appwrite/Realtime/Message/Dispatcher.php
abnegate: We will look into a way to abstract this into Utopia in future

tests/e2e/Services/GraphQL/PresenceTest.php
```diff
@@ -22,7 +22,7 @@ public function testUpsertPresenceSourceIsGraphql(): void
         $payload = [
             'query' => <<<'GQL'
                 mutation upsertPresence($presenceId: String!, $userId: String!, $status: String!, $metadata: Json) {
-                    presencesUpsert(presenceId: $presenceId, userId: $userId, status: $status, metadata: $metadata) {
+                    presencesUpsertPresence(presenceId: $presenceId, userId: $userId, status: $status, metadata: $metadata) {
```
abnegate: For root resource type we shouldn't double up, should be presences.upsert in sdks, `presencesUpsert` in GraphQL

src/Appwrite/Realtime/Message/Handlers/Subscribe.php
```diff
@@ -0,0 +1,109 @@
...
+     * @return array<string, mixed>
+     */
+    public function action(
+        array $items,
+        int $connection,
+        Realtime $realtime,
+        Registry $register,
+        ?string $projectId,
+    ): array {
+        $roles = $realtime->connections[$connection]['roles'] ?? [Role::guests()->toString()];
```
Meldiron: Feels a bit risky - gives me ability to read another connection by mistake easily.

int $connections resurce should be 'connectionId', or such. So then we can make resource for 'connection' to be the connection object.

src/Appwrite/Realtime/Message/Validators/SubscribePayload.php
```diff
@@ -0,0 +1,69 @@
...
+            if (!\array_key_exists('channels', $payload)) {
+                $this->description = 'channels is not present in payload.';
+                return false;
+            }
+            if (!\is_array($payload['channels']) || !\array_is_list($payload['channels'])) {
+                $this->description = 'channels is not a valid array.';
+                return false;
+            }
+            foreach ($payload['channels'] as $channel) {
+                if (!\is_string($channel)) {
```
Meldiron, line 54: Would be great to proxy each property to its own separate paraleter, instead of having all logic in here.

src/Appwrite/Realtime/Message/Handlers/Subscribe.php
```diff
@@ -0,0 +1,109 @@
...
+        if ($subscriptionDelta !== 0) {
+            $register->get('telemetry.workerSubscriptionCounter')
+                ->add($subscriptionDelta, $register->get('telemetry.workerAttributes'));
+        }
+
+        Span::add('realtime.subscriptionDelta', $subscriptionDelta);
+        Span::add('realtime.subscriptionsRequested', $subscriptionsRequested);
+        Span::add('realtime.subscribe.subscriptionsCount', $subscriptionsRequested);
+
+        return [
```
Meldiron, line 96: We should do new Document( here, just a pattern we do

src/Appwrite/Realtime/Message/Handlers/Subscribe.php
```diff
@@ -0,0 +1,109 @@
...
+            'type' => 'response',
+            'data' => [
```
Meldiron, line 98: Should we have response models? 🤔

src/Appwrite/Realtime/Message/Handlers/Unsubscribe.php
```diff
@@ -0,0 +1,74 @@
...
+use Utopia\Registry\Registry;
+use Utopia\Span\Span;
+
+class Unsubscribe extends Action
+{
+    public function __construct()
+    {
+        $this
+            ->desc('Bulk remove subscriptions by id')
+            ->label(Dispatcher::LABEL_MESSAGE_TYPE, 'unsubscribe')
```
Meldiron, line 18: Comparing Authentication and Subscribe handlers, I woudl expect a label defining one is for unauthenticated, and one is for authenticated users

src/Appwrite/Realtime/Message/Validators/SubscribePayload.php
ArnabChatterjee20k: Actually we can't do it as its bulk way of sending payload
If you see in the current appwrite
https://github.com/appwrite/appwrite/blob/1273bcd33afbd3694c52ec216d27d3cd101da68a/app/realtime.php#L1155-L1211

src/Appwrite/Realtime/Message/Handlers/Unsubscribe.php
ArnabChatterjee20k: Unsubscribe is for any users -> authenticated and unauthenticated -> cause any client can subscribe to a channela and unsubscribe to a channel

src/Appwrite/Realtime/Message/Handlers/Subscribe.php
abnegate: Yes, but future scope 👌

src/Appwrite/Realtime/Message/Handlers/Subscribe.php
abnegate: In future we will have proper models instead

src/Appwrite/Realtime/Message/Handlers/Subscribe.php
abnegate: This is not a change of behavior, just a port to handlers vs the big switch statement. I do agree that `connection` should be `connectionId` though since we're using it as an identifier, cc @ArnabChatterjee20k

src/Appwrite/Realtime/Message/Handlers/Unsubscribe.php
abnegate: What happens when you subscribe without auth?

ArnabChatterjee20k: Nothing happens. Appwrite doesn't enforce anything regarding subscription/unsubscription.
When subscription is done , all the roles of a user are added in the subscription tree. If not auth, that means only the permission with any is possible when user will receive the doc

If permission is present for auth users on the doc, then non-auth users will never receive the doc

https://github.com/appwrite/appwrite/blob/1273bcd33afbd3694c52ec216d27d3cd101da68a/app/realtime.php#L1155-L1316

Example
```
import { Client, Realtime, Channel } from "appwrite";

const client = new Client()
    .setEndpoint('https://<REGION>.cloud.appwrite.io/v1')
    .setProject('<PROJECT_ID>');

const realtime = new Realtime(client);

const subscription = await realtime.subscribe(Channel.files(), response => {
    // Callback will be executed on changes for all files.
    console.log(response);
});

// Removes only this subscription. Other subscriptions and the WebSocket stay open.
await subscription.unsubscribe();
```

Subscribe + unsubscribe is different from the presence api(which enforces the authentication by default)

src/Appwrite/Realtime/Message/Handlers/Subscribe.php
```diff
@@ -0,0 +1,109 @@
...
+                ? $payload['subscriptionId']
+                : ID::unique();
+
+            $queries = $payload['queries'] ?? [];
+
+            try {
+                $convertedQueries = Realtime::convertQueries($queries);
+            } catch (QueryException $e) {
+                throw new Exception(Exception::REALTIME_MESSAGE_FORMAT_INVALID, 'Invalid query: ' . $e->getMessage());
+            }
```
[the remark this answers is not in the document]
ArnabChatterjee20k: For type errors inside the query conversion, internally only the query exception are thrown. so not needed.
Tested and we were getting analyze error. so needed to revert

abnegate (approved):

PR #12215, opened 2026-05-05
Stop user-data errors leaking to Sentry from migrations worker

premtsd-code:
## Summary

Migrations worker was paging on every user-data condition (duplicate column, invalid document structure, missing required field, bad credentials, etc.) because the `code === 0 || >= 500` heuristic in the per-batch loop matched almost everything.

- Outer `catch (\Throwable $th)` gates `logError` with `if (!$th instanceof MigrationException)` — single, type-based routing decision.
- Deleted the foreach loop that re-published collected errors with the bad code heuristic.
- Hoisted `setAttribute('errors', sanitizeErrors(...))` into `finally` so the migration document always reflects the failure (including the bubbled exception's message via a new `\$caughtError` capture).
- 4 setup-time bare `\Exception` throws (Source projectId required, Source project not found, Invalid source/destination type) converted to `MigrationException` with appropriate codes — so the gate classifies them as user errors.

Pairs with library-side fixes in [utopia-php/migration#fix-migration-sentry-leak](https://github.com/utopia-php/migration/tree/fix-migration-sentry-leak).

## ⚠️ Cross-repo dependency — revert before merge

`composer.json` and `composer.lock` pin `utopia-php/migration` to the dev branch `dev-fix-migration-sentry-leak` for end-to-end testing. **These must be reverted to a versioned dependency (e.g. `1.9.*`) once the utopia-php/migration PR is merged and tagged.**

## Test plan

- [ ] Migration with duplicate row → *"Document already exists"* in migration report, **no** Sentry event
- [ ] Migration with bad source projectId → *"Source project not found"* in report, no Sentry
- [ ] Migration with invalid source type → *"Invalid source type"* in report, no Sentry
- [ ] Synthetic library bug (TypeError) → recorded in report AND reaches Sentry with full trace
- [ ] Missing `_APP_MIGRATION_HOST` env var → still reaches Sentry (config bug, intentionally not converted)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -74,7 +74,7 @@
         "utopia-php/locale": "0.8.*",
         "utopia-php/logger": "0.6.*",
         "utopia-php/messaging": "0.22.*",
-        "utopia-php/migration": "1.9.*",
+        "utopia-php/migration": "dev-fix-migration-sentry-leak as 1.9.999",
         "utopia-php/platform": "0.13.*",
         "utopia-php/pools": "1.*",
         "utopia-php/span": "1.1.*",
@@ -118,5 +118,11 @@
             "php-http/discovery": true,
             "tbachert/spi": true
         }
-    }
+    },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/utopia-php/migration"
+        }
+    ]
 }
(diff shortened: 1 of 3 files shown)


premtsd-code:
Superseded by #12226 which uses the existing `Appwrite\Extend\Exception::isPublishable()` registry pattern instead of an `instanceof MigrationException` gate. Same Sentry-leak fix, more consistent with the HTTP layer (`general.php:1227-1231`).

PR #12217, opened 2026-05-05
fix: skip stale repositories referencing deleted projects in VCS deployments

HarshMN2345:
When a project is deleted, its repository records in the platform DB may not be cleaned up immediately. Processing a GitHub push event for such a repo was throwing PROJECT_NOT_FOUND inside the deployment loop, which got swallowed by the blanket catch and re-thrown as GENERAL_UNKNOWN (500).

Skip these stale repository entries with a warning instead.

## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist


src/Appwrite/Platform/Modules/VCS/Http/GitHub/Deployment.php
Meldiron: I think this stull results in same, no?

We append message to errors array, so errors array is not empty, so general unknown (500) is thrown.

Meldiron: This behaviour is questionable in general - when multiple errors occur, which one should we throw? 🤔

I think best would be to loop all errors, and find the most relevant. First, try to find any 5xx error of our Exception class, and throw that. Second, same thing, but 4xx. And only then, if nothing matches this expectation, throw general 500. I would also not expose the message for general 5xx, that can be unsafe.

HarshMN2345:
Closing — #12391 landed this on main on 2026-05-28: VCS deployments now warn and skip on typed 4xx exceptions instead of 500ing. The remainder of this diff would also swallow 5xx, which isn't what we want.

PR #12226, opened 2026-05-06
Stop user-data errors leaking to Sentry from migrations worker (isPublishable)

premtsd-code:
## Summary

Alternative to [#12215](https://github.com/appwrite/appwrite/pull/12215) — uses the existing `Appwrite\Extend\Exception::isPublishable()` registry pattern (the same one `general.php:1227-1231` uses for HTTP errors) instead of an `instanceof MigrationException` gate.

Stops user-data errors from leaking to Sentry while staying consistent with the HTTP layer's error catalog.

## What changed

**Registered four migration setup error types in `app/config/errors.php`** with HTTP-style codes:

- `MIGRATION_SOURCE_PROJECT_ID_REQUIRED` — code 400 → `publish=false`
- `MIGRATION_SOURCE_PROJECT_NOT_FOUND` — code 404 → `publish=false`
- `MIGRATION_SOURCE_TYPE_INVALID` — code 400 → `publish=false`
- `MIGRATION_DESTINATION_TYPE_INVALID` — code 400 → `publish=false`

**Converted four worker setup-time throws** from bare `\Exception` (or `Migration\Exception`) to `Appwrite\Extend\Exception` using the new constants:

```php
// before
throw new \Exception('Source projectId is required for Appwrite migrations');
// after
throw new Exception(Exception::MIGRATION_SOURCE_PROJECT_ID_REQUIRED);
```

The throw site is one line; the user-facing description, code, and publish flag live in the registry.

**Outer catch uses a 3-arm gate**, mirroring `general.php`:

```php
if ($th instanceof Exception) {                  // Appwrite\Extend\Exception
    $publish = $th->isPublishable();
} elseif ($th instanceof MigrationException) {   // Utopia\Migration\Exception (library)
    $publish = false;
} else {
    $publish = true;                             // unknown → surface to Sentry
}

if ($publish) {
    call_user_func($this->logError, $th, ...);
}
```

`Appwrite\Extend\Exception` and `Utopia\Migration\Exception` are siblings (both extend `\Exception` directly) — no inheritance overlap, so the arm order is unambiguous.

Arm 3 is `true` (publish) by design: anything that isn't a typed Appwrite exception or a library `Migration\Exception` is, by definition, an unexpected/unknown error and worth a Sentry event. The few existing `throw new \Exception(...)` sites with HTTP-style codes inside the migration library destination should be converted to `Migration\Exception` in a follow-up so they land in arm 2 (suppressed).


src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -197,13 +197,13 @@ protected function processSource(Document $migration): Source
         $projectDB = null;
         $useAppwriteApiSource = false;
         if ($source === SourceAppwrite::getName() && empty($credentials['projectId'])) {
-            throw new \Exception('Source projectId is required for Appwrite migrations');
+            throw new Exception(Exception::MIGRATION_SOURCE_PROJECT_ID_REQUIRED);
```
fogelito, line 200: I think this should be 500 error, We should not get to this point with wrong data

src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -197,13 +197,13 @@ protected function processSource(Document $migration): Source
...
         }
 
         if (! empty($credentials['projectId'])) {
             $this->sourceProject = $this->dbForPlatform->getDocument('projects', $credentials['projectId']);
             if ($this->sourceProject->isEmpty()) {
-                throw new \Exception('Source project not found for provided projectId');
+                throw new Exception(Exception::MIGRATION_SOURCE_PROJECT_NOT_FOUND);
```
fogelito, line 206: Same here 
I think this should be 500 error, We should not get to this point with wrong data
We should throw earlier in the api I think

composer.json
```diff
@@ -74,7 +74,7 @@
         "utopia-php/locale": "0.8.*",
         "utopia-php/logger": "0.6.*",
         "utopia-php/messaging": "0.22.*",
-        "utopia-php/migration": "1.*",
+        "utopia-php/migration": "1.10.*",
```
abnegate: Let's revert, no need to pin to a feature version

src/Appwrite/Platform/Workers/Migrations.php
abnegate: This one needs to throw from worker, in case the source project is deleted by the time the job starts

src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -546,35 +546,54 @@ function ($resources) use ($migration, $transfer, $project, $queueForRealtime, &
...
+                    $bubbled = $caughtError instanceof MigrationException
+                        ? $caughtError
+                        : new MigrationException(
+                            resourceName: '',
+                            resourceGroup: '',
+                            message: $caughtError->getMessage(),
+                            code: $caughtError->getCode(),
+                            previous: $caughtError,
+                        );
+                    $destinationErrors[] = $bubbled;
```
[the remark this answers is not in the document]
premtsd-code: Addressed in 3676af925c. Branched on type in the finally block:

- `MigrationException` and `Appwrite\Extend\Exception` keep their original message (constructed by us / curated in `app/config/errors.php`).
- Any other `Throwable` falls back to a generic `"Migration failed due to an unexpected error."` user-facing string. The raw throwable is still attached via `previous:` so Sentry sees the full message via arm 3 of the gate.

This restores the pre-spike property that infrastructure exception messages never reach the migration document.

abnegate (approved):

PR #12240, opened 2026-05-08
Add cached response cache-control callback

ChiragAgg5k:
## What does this PR do?

Adds a generic `cacheControlForResponse` resource callback that editions can override to customize `Cache-Control` for specific responses. The default callback returns `null`, preserving the existing private cache-control behavior.

The callback is used for cached API response hits and storage preview direct responses, giving Appwrite Cloud a narrow extension point for storage preview cache headers without duplicating shared cache-hit authorization or preview rendering logic.

## Test Plan

- Ran `php -l app/init/resources.php`
- Ran `php -l app/controllers/shared/api.php`
- Ran `php -l src/Appwrite/Platform/Modules/Storage/Http/Buckets/Files/Preview/Get.php`

## Related PRs and Issues

- Cloud follow-up PR will consume this branch as a dev dependency.

## Checklist

### DIFF
--- a/app/controllers/shared/api.php
+++ b/app/controllers/shared/api.php
@@ -705,7 +705,6 @@ Http::init()
                             bucket: $bucket,
                             file: $file,
                             resourceToken: $resourceToken,
-                            isImageTransformation: $isImageTransformation,
                             fileSecurity: $fileSecurity,
                             cacheLog: $cacheLog,
                             route: $route,
--- a/app/controllers/shared/api.php
+++ b/app/controllers/shared/api.php
@@ -20,6 +20,7 @@ use Appwrite\Extend\Exception;
 use Appwrite\Extend\Exception as AppwriteException;
 use Appwrite\Functions\EventProcessor;
 use Appwrite\Platform\Modules\Storage\Config\CacheControl;
+use Appwrite\Platform\Modules\Storage\Config\StorageCacheControl;
 use Appwrite\SDK\Method;
 use Appwrite\Usage\Context;
 use Appwrite\Utopia\Database\Documents\User;
@@ -641,9 +642,7 @@ Http::init()
             $data = $cache->load($key, $timestamp);
 
             if (! empty($data) && ! $cacheLog->isEmpty()) {
-                $bucket = new Document();
-                $file = new Document();
-                $fileSecurity = null;
+                $storageCacheControl = null;
                 $parts = explode('/', $cacheLog->getAttribute('resourceType', ''));
                 $type = $parts[0];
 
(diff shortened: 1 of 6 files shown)


app/init/resources.php
```diff
@@ -203,6 +203,12 @@
     return $cache;
 }, ['pools', 'telemetry']);
 
+$container->set('cacheControlForStorage', fn () => function (array $context): string {
+    $maxAge = $context['maxAge'] ?? 0;
+
+    return \sprintf('private, max-age=%d', \is_int($maxAge) ? $maxAge : 0);
+});
+
```
abnegate: Let's avoid functional resources; they're just global functions with a mask. We can achieve the same in a more OOP way by adding a function to a base `Action` class and then extend that instead.

ChiragAgg5k: Base `Action` class wont be reached, response is sent in api.php before it

app/controllers/shared/api.php
```diff
@@ -702,8 +706,22 @@
...
+                    'project' => $project,
+                    'user' => $user,
+                    'bucket' => $bucket,
+                    'file' => $file,
+                    'fileSecurity' => $fileSecurity,
+                    'resourceToken' => $resourceToken,
+                    'cacheLog' => $cacheLog,
+                    'maxAge' => $timestamp,
+                    'isImageTransformation' => $isImageTransformation,
+                ]);
```
abnegate: Can we use typed config?

app/controllers/shared/api.php
```diff
@@ -702,8 +706,22 @@
...
+
                 $response
-                    ->addHeader('Cache-Control', sprintf('private, max-age=%d', $timestamp))
+                    ->addHeader('Cache-Control', $cacheControl)
```
abnegate, line 725: Won't this apply for ALL requests, not only storage previews, and not for a specific project?

app/controllers/shared/api.php
```diff
@@ -639,6 +640,9 @@
             $data = $cache->load($key, $timestamp);
 
             if (! empty($data) && ! $cacheLog->isEmpty()) {
+                $bucket = new Document();
+                $file = new Document();
+                $fileSecurity = null;
```
abnegate: Feels like dead setup, if we're just setting to fill context shape, we should probably have two context: global and storage

abnegate (commented):
Let's add tests as well

app/controllers/shared/api.php
ChiragAgg5k: right, we should wrap it in `$isImageTransformation`

app/controllers/shared/api.php
```diff
@@ -691,6 +695,22 @@
                             ])));
                         }
                     }
+
+                    if ($isImageTransformation) {
```
abnegate, line 699: We need a project ID check as well right? Can we add some way to fetch a whitelist and override for Cloud?

abnegate (approved):

abnegate (approved):

PR #12251, opened 2026-05-08
Fix GraphQL preview test assertions

ChiragAgg5k:
## What does this PR do?

Updates the GraphQL storage preview E2E tests to assert the response contract instead of an exact Imagick-generated PNG byte length.

The failing CI job on #12246 consistently returned a valid preview body with a different PNG size (`46952` instead of `46719`). Exact encoded image size is brittle across image pipeline/runtime changes, so these tests now assert:

- HTTP 200 response
- `image/png` content type
- non-empty body
- decoded 100x100 PNG dimensions

This also wires the existing `width` and `height` test variables into the GraphQL preview query.

## Test Plan

- `composer format tests/e2e/Services/GraphQL/Base.php tests/e2e/Services/GraphQL/StorageClientTest.php tests/e2e/Services/GraphQL/StorageServerTest.php`
- `composer lint tests/e2e/Services/GraphQL/Base.php tests/e2e/Services/GraphQL/StorageClientTest.php tests/e2e/Services/GraphQL/StorageServerTest.php`
- `php -l tests/e2e/Services/GraphQL/Base.php && php -l tests/e2e/Services/GraphQL/StorageClientTest.php && php -l tests/e2e/Services/GraphQL/StorageServerTest.php`
- `git diff --check -- tests/e2e/Services/GraphQL/Base.php tests/e2e/Services/GraphQL/StorageClientTest.php tests/e2e/Services/GraphQL/StorageServerTest.php`

The targeted E2E tests were not run locally because the appwrite Docker service is not currently running in this workspace.

## Related PRs and Issues

- Split out from CI investigation on #12246

## Checklist

### DIFF
--- a/tests/e2e/Services/GraphQL/Base.php
+++ b/tests/e2e/Services/GraphQL/Base.php
@@ -4,6 +4,7 @@ namespace Tests\E2E\Services\GraphQL;
 
 use CURLFile;
 use Utopia\Console;
+use Utopia\Image\Image;
 
 trait Base
 {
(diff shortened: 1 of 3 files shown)


tests/e2e/Services/GraphQL/Base.php
```diff
@@ -516,6 +516,20 @@ trait Base
         }
     ';
 
+    protected function assertFilePreviewResponse(array $file): void
```
TorstenDittmann, line 520: i think its fine to have a hardcoded value, if not lets rather use `utopia-php/image` instead of imageick directly

TorstenDittmann (approved):

PR #12257, opened 2026-05-09
Solved issue 8785 regarding the server error when null name provided

AdityaOP007:
thunder_client_success_mockup_1778320977580

The problem was - 
The Users.CreateUser API endpoints were returning a 500 Internal Server Error when the name parameter was explicitly provided as null in the request body.

While the name field is optional, the underlying Text(128) validator in the Utopia PHP framework strictly enforced string types. When a null was passed, it triggered a TypeError during the parameter validation phase, crashing the request before it could reach the controller logic.

Changes I made were -
I updated the parameter validation logic for all 8 variations of the user creation endpoint in app/controllers/api/users.php.

Files Modified:
app/controllers/api/users.php
Specific Implementations:
Nullable Validation: Wrapped the Text(128) validator with the Nullable() wrapper for the name parameter. This allows the API to gracefully accept null values without throwing a validation error.
Updated Defaults: Changed the default value of the name parameter from an empty string '' to null. This ensures consistency with the new nullable validation.
Affected Endpoints:
POST /v1/users (Standard)
POST /v1/users/bcrypt
POST /v1/users/md5
POST /v1/users/argon2
POST /v1/users/sha
POST /v1/users/phpass
POST /v1/users/scrypt
POST /v1/users/scrypt-modified

The core createUser() function already contains defensive logic to handle name values:
By allowing null at the validation layer, the value now safely passes through to this line, where it is converted to an empty string as intended, ensuring backward compatibility and preventing the 500 error.

Unit Testing: Verified that new Nullable(new Text(128)) correctly returns true for isValid(null) while still enforcing string constraints (e.g., max length) for non-null values.
Regression: Confirmed that providing a valid string or omitting the field entirely still works as expected.
Local E2E: Successfully simulated the fix inside the local Docker environment, resolving the TypeError.

 All modified lines include the comment // AdityaOP007 solve issue
 FIxes #8785 
 for easy tracking.


HarshMN2345:

PR #12260, opened 2026-05-09
Fix nullable name validation for create user endpoints

Meexargha:
## What does this PR do?

Fixes the issue where passing `null` as the `name` parameter to `Users.CreateUser...` endpoints results in a vague server error instead of proper validation handling.

## What changed?

Updated all create user endpoints to use:

```php
new Nullable(new Text(128))
```

instead of:

```php
new Text(128)
```

for the `name` parameter.

## Why?

The `createUser()` function already accepts:

```php
?string $name
```

but endpoint validation rejected `null` improperly, leading to internal server errors.

This change ensures:

* proper nullable validation
* consistent endpoint behavior
* avoids vague server errors


HarshMN2345:

PR #12262, opened 2026-05-09
fix: compensate for execution time drift in scheduled functions

### LINKED ISSUE #11931: Scheduled functions execution time drift due to disabled time tracking in ScheduleFunctions.php
### 👟 Reproduction steps

Create a function with a cron schedule (e.g., */5 * * * * to run every 5 minutes)
Deploy the function and enable scheduling
Monitor the execution timestamps over several hours
Observe that executions progressively drift later than the scheduled time

### 👍 Expected behavior

Scheduled functions should execute at their precise cron-scheduled times. If processing takes time, the scheduler should compensate by adjusting the next cycle's wait time to maintain accurate scheduling.

### 👎 Actual Behavior

The time drift compensation logic in ScheduleFunctions.php is intentionally disabled. The variable $lastEnqueueUpdate is never updated (line 132 is commented out), so $enqueueDiff is always 0. This means the scheduler cannot compensate for processing time, causing scheduled executions to drift progressively later over time.

Root cause: In src/Appwrite/Platform/Tasks/ScheduleFunctions.php:

Line 24: private ?float $lastEnqueueUpdate = null;
Line 46: $enqueueDiff = $this->lastEnqueueUpdate === null ? 0 : $timerStart - $this->lastEnqueueUpdate;
Line 47: $timeFrame = DateTime::addSeconds(new \DateTime(), static::ENQUEUE_TIMER - $enqueueDiff);
Lines 131-132: The update is commented out with TODO: // TODO: This was a bug before because it wasn't passed by reference, enabling it breaks scheduling
The TODO comment references a historical issue that may no longer apply, as $lastEnqueueUpdate is now an instance variable (not requiring reference passing).

### 🎲 Appwrite version

Version 1.8.x

### 💻 Operating system

Windows

### 🧱 Your Environment

1.9.0
local setup

harshagarwalnyu:
Fixes #11931. The current implementation of ScheduleFunctions does not track the actual processing time of an enqueue cycle, leading to a progressive drift in execution times. This PR re-enables lastEnqueueUpdate tracking as a private property to accurately subtract drift from the next timeframe.


harshagarwalnyu:
Going to close this one. It picked up far more changes than the actual fix needs, which makes it hard to review as-is. If the scheduled-function drift compensation is still useful, happy to open a focused PR that only touches the scheduler. Thanks for taking a look.

PR #12267, opened 2026-05-10
fix: prevent enqueue drift in scheduled functions

### LINKED ISSUE #11931: Scheduled functions execution time drift due to disabled time tracking in ScheduleFunctions.php

DEADSERPENT:
## What does this PR do?

Fixes scheduling drift handling in `src/Appwrite/Platform/Tasks/ScheduleFunctions.php`.

Previously, `$lastEnqueueUpdate` logic was disabled because the original calculation:

```php
ENQUEUE_TIMER - elapsed
```

could collapse the scheduling window to zero when `elapsed` approached the enqueue interval duration.

This caused scheduled executions to drift and potentially miss the intended enqueue window during delayed scheduler cycles.

This PR restores the enqueue update logic using:

```php
ENQUEUE_TIMER + max(0, elapsed - ENQUEUE_TIMER)
```

This preserves the normal enqueue window during regular execution while safely widening the look-ahead window only when scheduler cycles run late.

## Why this matters

- Prevents schedule drift during delayed scheduler cycles
- Avoids enqueue window collapse
- Keeps normal scheduling behavior unchanged during healthy execution
- Aligns delayed-cycle behavior with intended scheduling semantics

## Test Plan

- Verified scheduler behavior with delayed execution cycles
- Confirmed enqueue windows no longer collapse when elapsed time approaches the timer interval
- Verified normal execution timing remains unaffected for non-delayed cycles

## Related PRs and Issues

- Fixes #11931

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Tasks/ScheduleFunctions.php
+++ b/src/Appwrite/Platform/Tasks/ScheduleFunctions.php
@@ -49,7 +49,7 @@ class ScheduleFunctions extends ScheduleBase
         $elapsed = $this->lastEnqueueUpdate === null ? 0.0 : $timerStart - $this->lastEnqueueUpdate;
         $enqueueDiff = max(0.0, $elapsed - static::ENQUEUE_TIMER);
         $this->lastEnqueueUpdate = $timerStart;
-        $timeFrame = DateTime::addSeconds(new \DateTime(), static::ENQUEUE_TIMER + (int)$enqueueDiff);
+        $timeFrame = DateTime::addSeconds(new \DateTime(), static::ENQUEUE_TIMER + (int)ceil($enqueueDiff));
 
         Console::log("Enqueue tick: started at: $time (with diff $enqueueDiff)");
 
--- a/src/Appwrite/Platform/Tasks/ScheduleFunctions.php
+++ b/src/Appwrite/Platform/Tasks/ScheduleFunctions.php
@@ -21,6 +21,8 @@ class ScheduleFunctions extends ScheduleBase
     public const UPDATE_TIMER = 10; // seconds
     public const ENQUEUE_TIMER = 60; // seconds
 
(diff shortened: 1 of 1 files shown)


DEADSERPENT:
Good catch — updated to use `ceil()` so fractional-second overruns are also covered without leaving a gap.

PR #12268, opened 2026-05-10
fix(users): handle null name parameter in CreateUser endpoints

jonasboos:
## What does this PR do?

Fixes #8785 — When `null` is explicitly passed for the optional `name` parameter in Users.CreateUser endpoints, a vague TypeError is thrown because the `Text(128)` validator doesn't accept null values.

## Changes

Wrapped the `Text(128)` validator with `Nullable` on all 8 CreateUser endpoint `name` parameter declarations, matching how `email` and `phone` parameters already use `Nullable`.

The `createUser()` helper already handles null safely via `$name = $name ?? ''` (line 78), so no further changes are needed.

### Affected Endpoints
- `POST /v1/users`
- `POST /v1/users/bcrypt`
- `POST /v1/users/md5`
- `POST /v1/users/argon2`
- `POST /v1/users/sha`
- `POST /v1/users/phpass`
- `POST /v1/users/scrypt`
- `POST /v1/users/scrypt-modified`

## Test Plan

- Send `{"name": null}` to any CreateUser endpoint — should create user with empty name instead of throwing a 500 error
- Send `{"name": "Valid Name"}` — should work as before
- Omit `name` entirely — should work as before (defaults to empty string)


HarshMN2345:

PR #12281, opened 2026-05-12
fix(users): mark name as nullable in createUser endpoints

henrybrewer00-dotcom:
## What does this PR do?

Sending `name: null` to the `createUser*` endpoints used to surface a vague 500 server error. The action signatures and the `createUser` helper were updated previously to accept `?string $name`, but the route metadata still declared the parameter as a non-nullable `Text(128)`. That left the null behavior implicit and meant the generated SDKs / OpenAPI specs still advertised the parameter as a non-nullable string.

This PR aligns the `name` parameter on every `createUser*` endpoint with the `email` and `phone` parameters on the same routes by wrapping the `Text(128)` validator with `Nullable` and defaulting it to `null`. The runtime behavior already coalesces a null value to an empty string inside `createUser()`, so the user contract is unchanged.

Endpoints updated:

- `POST /v1/users`
- `POST /v1/users/bcrypt`
- `POST /v1/users/md5`
- `POST /v1/users/argon2`
- `POST /v1/users/sha`
- `POST /v1/users/phpass`
- `POST /v1/users/scrypt`
- `POST /v1/users/scrypt-modified`

## Test Plan

- Added `testCreateUserNullName` in `tests/e2e/Services/Users/UsersBase.php`. It posts `name: null` against every `createUser*` endpoint and asserts the response is `201` with `name: ""`.
- Verified the test fails on the current `1.9.x` snapshot when the route metadata is reverted to `new Text(128)` and `createUser` uses `string $name` (reproduces the original 500 from the issue).
- `docker compose exec appwrite test tests/e2e/Services/Users/UsersCustomServerTest.php` -> 49/49 pass, 581 assertions.
- `docker compose exec appwrite test tests/unit/` -> 536/536 pass, 3478 assertions.

## Related PRs and Issues

- Fixes #8785

## Checklist


henrybrewer00-dotcom:
Addressed the email-uniqueness note from the Greptile review — each endpoint case in `testCreateUserNullName` now uses a `\uniqid()` suffix, matching the existing pattern elsewhere in `UsersBase.php`. Re-ran the test against the docker dev stack: `OK (1 test, 29 assertions)`.

HarshMN2345:

PR #12275, opened 2026-05-11
Skip database abuse checks when cache circuit is open

ChiragAgg5k:
## What does this PR do?

Skips database API abuse/rate-limit checks when the API cache circuit is open, so database requests do not immediately fail with rate-limit responses while cache is unavailable.

This intentionally treats any open API cache circuit breaker shard as API cache degradation. While the API cache is degraded, database abuse checks are bypassed globally for database routes (`/v1/databases`, `/v1/tablesdb`, `/v1/documentsdb`, and `/v1/vectorsdb`) for requests that would otherwise be subject to abuse limiting. API key, admin, and dev-key requests keep their existing behavior because they are not rate-limited by this abuse path.

## Test Plan

- `php -l src/Appwrite/Cache/Adapter/CircuitBreaker.php`
- `php -l app/init/resources.php`
- `php -l app/controllers/shared/api.php`
- `git diff --check -- src/Appwrite/Cache/Adapter/CircuitBreaker.php app/init/resources.php app/controllers/shared/api.php`
- Started the local stack with abuse enabled and verified `appwrite`, `redis`, and `mariadb` are healthy.
- Verified healthy cache behavior still rate-limits database document updates: `{"200":120,"429":15}`, first limited at request `120`.
- Simulated cache being down by denying Redis cache commands (`HGET`, `HSET`, `HDEL`, `DEL`) while keeping rate-limit Redis commands available, then verified database document updates continue without rate-limit responses: `{"200":160}`.

## Related PRs and Issues

- Stacked on #12231

## Checklist

- [x] Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [x] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

PR #12284, opened 2026-05-12
Fail open on abuse timelimit failures

ChiragAgg5k:
## What does this PR do?

Fails open when the shared API abuse time-limit backend throws while checking rate limits. This keeps API requests from failing because the abuse limiter cannot read or write its Redis-backed counters.

Unlike the closed circuit-breaker approach in #12275, this does not inspect API cache state or bypass database routes preemptively. It only skips abuse enforcement for a specific limit when the abuse time-limit interaction itself fails.

Privileged, API key, and dev-key requests continue to avoid `Abuse::check()` as before. Normal rate-limit responses are still returned when the abuse check succeeds and reports that the route is limited.

## Test Plan

- `php -l app/controllers/shared/api.php`
- `composer lint app/controllers/shared/api.php`
- `git diff --check -- app/controllers/shared/api.php`

## Related PRs and Issues

- Replaces #12275 with a narrower fail-open behavior


app/controllers/shared/api.php
```diff
@@ -555,27 +561,35 @@
...
+            $isRateLimited = false;
 
-            if ($limit && ($remaining < $closestLimit || is_null($closestLimit))) {
-                $closestLimit = $remaining;
-                $response
-                    ->addHeader('X-RateLimit-Limit', $limit)
-                    ->addHeader('X-RateLimit-Remaining', $remaining)
-                    ->addHeader('X-RateLimit-Reset', $time);
-            }
+            try {
```
loks0n, line 512: Can we split rate limiting to a separate init hook? Then we can use early return to improve the readability also.

app/init/resources/request.php
```diff
@@ -70,6 +72,49 @@
         return $register->get('logger');
     }, ['register']);
 
+    $container->set('logError', function (Registry $register, Request $request, Document $project, Authorization $authorization) {
```
loks0n: Can we just use error_log instead? Since we're moving away from utopia-php/logger

loks0n (changes requested):

loks0n (approved):

lohanidamodar (approved):

PR #12300, opened 2026-05-13
Fix project variable creation when variableId is omitted

### LINKED ISSUE #12247: Global/project variables unusable on Appwrite 1.9.0 self-hosted: Console throws "Param variableId is not optional" and API scopes are unavailable
### 👟 Reproduction steps

1. Run self-hosted Appwrite `1.9.0` using Docker.
2. Open Console → Project Settings → Global variables / Environment variables.
3. Try to create, import, or edit a variable.
4. Console fails with:

`Param "variableId" is not optional.`

5. Attempt to use the Server SDK / REST API instead.
6. Project variable endpoints require scopes like `project.read` / `project.write`.
7. Those scopes are not available when creating API keys in Appwrite `1.9.0`.
8. Try using `appwrite.config.json` / CLI deployment as a workaround.
9. Function deployments succeed, but variable sync is unreliable.

Result: project/global variables are effectively unusable on self-hosted `1.9.0`.

### 👍 Expected behavior

At least one supported workflow for managing project/global variables should work reliably.

Expected:

* Console can create/edit/import/delete variables without the `variableId` error.
* If `variableId` is required, the Console should send it correctly.
* If variables are intended to be managed through API/SDKs, required scopes (`project.read`, `project.write`) should be available for API keys.
* CLI/config deployments should reliably sync variables.

### 👎 Actual Behavior

All available workflows are blocked or unreliable:

* Console fails with:

`Param "variableId" is not optional.`

* REST API / Server SDK cannot be used because required project scopes are missing from the API key UI.
* CLI/config variable sync is inconsistent.

This is a production issue for larger deployments. I currently manage 51+ functions with shared global variables and function-specific overrides. Manual Console management is not realistic, and the Console workflow is currently broken.

### 🎲 Appwrite version

Self-hosted `1.9.0`

### 💻 Operating system

Linux

### 🧱 Environment

* Self-hosted Appwrite `1.9.0`
* Docker deployment
* Official `appwrite/appwrite` image
* 51+ Functions
* Using global/project variables + function-level variables
* Compatible Server SDK versions

* [x] I checked and didn't find a similar issue

* [x] I have read the Code of Conduct

GettingError404:
## What does this PR do?

This PR fixes project variable creation when `variableId` is omitted.

In `src/Appwrite/Platform/Modules/Project/Http/Project/Variables/Create.php`, the `variableId` request parameter is already defined as optional. However, the action method still required `string $variableId`, which caused Appwrite to throw:

`Param "variableId" is not optional.`

when the Console attempted to create a project/global environment variable without sending a `variableId`.

This change updates the action method signature to accept `null` and generates a unique ID when `variableId` is omitted or explicitly set to `unique()`.

## Test Plan

1. Set up Appwrite locally from source using Docker.
2. Reproduced the issue while creating a project/global environment variable from the Console.
3. Updated the action signature from `string $variableId` to `?string $variableId = null`.
4. Updated the ID normalization logic to generate a unique ID when `variableId` is empty or equals `unique()`.
5. Rebuilt the containers using:

   * `docker compose up -d --build --force-recreate`
6. Verified the patch is isolated to `Create.php`.
7. Confirmed the error `Param "variableId" is not optional.` no longer occurs when `variableId` is omitted.

## Related PRs and Issues

* Fixes #12247

## Checklist

* [x] Have you read the [[Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
* [ ] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

This PR does not modify API metadata. It only updates the action method signature and runtime handling for an optional parameter.

### DIFF
--- a/src/Appwrite/Platform/Modules/Project/Http/Project/Variables/Create.php
+++ b/src/Appwrite/Platform/Modules/Project/Http/Project/Variables/Create.php
@@ -64,7 +64,7 @@ class Create extends Action
     }
 
     public function action(
-        string $variableId,
+        ?string $variableId = null,
         string $key,
         string $value,
         bool $secret,
@@ -72,7 +72,9 @@ class Create extends Action
         QueueEvent $queueForEvents,
         Database $dbForProject,
     ) {
-        $variableId = ($variableId === 'unique()') ? ID::unique() : $variableId;
+        if (empty($variableId) || $variableId === 'unique()') {
+            $variableId = ID::unique();
+        }
 
         $variable = new Document([
             '$id' => $variableId,
--- a/src/Appwrite/Platform/Modules/Project/Http/Project/Variables/Create.php
+++ b/src/Appwrite/Platform/Modules/Project/Http/Project/Variables/Create.php
@@ -53,7 +53,7 @@ class Create extends Action
                     )
                 ],
             ))
-            ->param('variableId', '', fn (Database $dbForProject) => new CustomId(false, $dbForProject->getAdapter()->getMaxUIDLength()), 'Variable unique ID. Choose a custom ID or generate a random ID with `ID.unique()`. Valid chars are a-z, A-Z, 0-9, period, hyphen, and underscore. Can\'t start with a special char. Max length is 36 chars.', false, ['dbForProject'])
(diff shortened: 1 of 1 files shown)


GettingError404:
Good catch. I updated the `->param('variableId', ...)` definition to mark `variableId` as optional, so the API metadata, generated SDKs, and documentation are now consistent with the runtime behavior, which already accepts omitted values and generates a unique ID automatically.

The update has been pushed in commit `b48a098a73`.

PR #12311, opened 2026-05-14
Fix: VCS branch pagination

HarshMN2345:
## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Modules/VCS/Http/Installations/Repositories/Branches/XList.php
+++ b/src/Appwrite/Platform/Modules/VCS/Http/Installations/Repositories/Branches/XList.php
@@ -163,7 +163,7 @@ class XList extends Action
             }
         }
 
-        $remainingBranches = $github->listBranches($owner, $repositoryName, 1, $maxPages + 1);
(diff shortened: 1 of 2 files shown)


src/Appwrite/Platform/Modules/VCS/Http/Installations/Repositories/Branches/XList.php
```diff
@@ -142,4 +145,29 @@ public function action(
...
+
+    /**
+     * @return array<string>
+     */
+    private function listBranches(GitHub $github, string $owner, string $repositoryName): array
+    {
+        $branches = [];
+        $maxPages = \intdiv(self::GITHUB_BRANCHES_MAX_RESULTS, self::GITHUB_BRANCHES_PER_PAGE);
+
+        for ($page = 1; $page <= $maxPages; $page++) {
```
Meldiron, line 157: We should accept pagination details as params to endpoint. Notice we already accept queries, lets utilize those - offset and limit queries.

I think we already have something similar done in other VCS endpoints, lets check there.

If not, any other list endpoint should showcase how it should be done - queries param, supporting offset and limit.

We should also proxy search term itself into github call, if that is possible by github APIs

PR #12332, opened 2026-05-18
Consolidate shared DI resources

ChiragAgg5k:
## What
- Keep `app/init/resources.php` as the shared DI resource source instead of adding a separate shared resource file.
- Move overlapping publisher resource definitions out of `app/cli.php`, while keeping them registered in `app/init/resources.php`.
- Move simple overlapping resources into `app/init/resources.php`: `console`, `executor`, and `telemetry`.
- Reuse the container initialized by `app/init.php` in `app/cli.php`, so CLI tasks get the shared resources from `resources.php` without duplicate registrations.
- Remove the duplicate `publisherForFunctions` registration from `app/init/worker/message.php`, relying on the shared registration loaded through `app/init.php`.

## Why
`resources.php` is already the shared bootstrap resource file. This PR keeps that pattern and removes duplicated resource definitions from other entrypoints, while leaving request-, task-, and message-specific resources in their context-specific files.

## Notes
- No new shared resource file is introduced.
- `app/init/resources/` remains reserved for nested `resources.php` context files like `request.php`.
- Database resources are intentionally left untouched because their API, task, and worker variants have different timeout and lifecycle behavior.

## Tests
- `composer format app/cli.php app/init/resources.php app/init/worker/message.php`
- `php -l app/cli.php`
- `php -l app/init/resources.php`
- `php -l app/init/worker/message.php`

### DIFF
--- a/app/cli.php
+++ b/app/cli.php
@@ -2,18 +2,10 @@
 
 require_once __DIR__ . '/init.php';
 
-use Appwrite\Event\Event;
-use Appwrite\Event\Publisher\Certificate as CertificatePublisher;
-use Appwrite\Event\Publisher\Database as DatabasePublisher;
-use Appwrite\Event\Publisher\Delete as DeletePublisher;
-use Appwrite\Event\Publisher\Func as FunctionPublisher;
-use Appwrite\Event\Publisher\StatsResources as StatsResourcesPublisher;
-use Appwrite\Event\Publisher\Usage as UsagePublisher;
 use Appwrite\Platform\Appwrite;
 use Appwrite\Runtimes\Runtimes;
 use Appwrite\Usage\Context as UsageContext;
 use Appwrite\Utopia\Database\Documents\User;
-use Executor\Executor;
 use Swoole\Runtime;
 use Swoole\Timer;
 use Utopia\Cache\Adapter\Pool as CachePool;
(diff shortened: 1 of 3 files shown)


app/cli.php
```diff
@@ -252,48 +242,12 @@
...
-    return $publisher;
-}, ['publisher']);
-$container->set('publisherMigrations', function (BrokerPool $publisher) {
-    return $publisher;
-}, ['publisher']);
-$container->set('publisherMessaging', function (BrokerPool $publisher) {
-    return $publisher;
-}, ['publisher']);
+$registerSharedResources = require __DIR__ . '/init/shared.php';
+$registerSharedResources($container);
```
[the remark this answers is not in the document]
ChiragAgg5k: I think we need to keep this call in cli.php. Although init.php loads resources.php and resources.php calls shared.php, cli.php creates a fresh Container after init.php returns and passes that fresh container into CLI. The resources registered during init.php are on the earlier global container, not the CLI container. This explicit shared.php invocation is what registers the shared resources on the CLI container. Without it, CLI tasks would lose resources like publisherForFunctions, console, executor, and telemetry.

app/init/shared.php
loks0n, line 1: No need for a new file. keep in resources.php

app/init/resources.php
```diff
@@ -67,83 +48,8 @@
...
-$container->set('publisherForMails', fn (Publisher $publisher) => new MailPublisher(
-    $publisher,
-    new Queue(System::getEnv('_APP_MAILS_QUEUE_NAME', Event::MAILS_QUEUE_NAME))
-), ['publisher']);
-$container->set('publisherForMessaging', fn (Publisher $publisher) => new MessagingPublisher(
-    $publisher,
-    new Queue(System::getEnv('_APP_MESSAGING_QUEUE_NAME', Event::MESSAGING_QUEUE_NAME))
-), ['publisher']);
+$registerSharedResources = require __DIR__ . '/shared.php';
+$registerSharedResources($container);
```
loks0n: Why is it a callable?

loks0n (approved):

PR #12379, opened 2026-05-24
fix(users): ensure prefs update event returns full user payload

deepshekhardas:
When updating user preferences via the API, the prefs update event (users.*.update.prefs) was only sending the userId parameter without the full user payload. This caused webhook consumers to receive incomplete data.

Changes:
- Added setPayload to event queue chain in users.php
- Updated WebhooksCustomServerTest.php to verify full user object in webhook payload

### DIFF
--- a/app/controllers/api/users.php
+++ b/app/controllers/api/users.php
@@ -1844,7 +1844,8 @@ Http::patch('/v1/users/:userId/prefs')
         $user = $dbForProject->updateDocument('users', $user->getId(), new Document(['prefs' => $prefs]));
 
         $queueForEvents
-            ->setParam('userId', $user->getId());
+            ->setParam('userId', $user->getId())
+            ->setPayload($response->output($user, Response::MODEL_USER));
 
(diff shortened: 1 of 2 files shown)


Meldiron:
Closing — this is already on `main`, so retargeting leaves nothing to merge.

- The source fix moved with the Users API refactor: `app/controllers/api/users.php` no longer exists on `main`, and the equivalent endpoint at `src/Appwrite/Platform/Modules/Users/Http/Users/Prefs/Update.php:67-69` already does `->setPayload($response->output($user, Response::MODEL_USER))`.
- The test assertions are already there too — `WebhooksCustomServerTest::testUpdateUserPrefs()` on `main` asserts `$webhook['data']['$id']` and `$webhook['data']['prefs']`.

The only hunk left after a rebase is `assertSame` → `assertEquals` on the `X-Appwrite-Webhook-User-Id` check, and `assertSame` is the stricter of the two, so that one isn't worth taking.

Thanks for the fix — it landed.

PR #12382, opened 2026-05-24
fix(migrations): validate source URL host before DB fast path

premtsd-code:
## Problem

The migrations worker decides how to read the source project using only a `projectId` lookup against the local platform DB. Two failure modes follow:

1. **Pre-flight throw on external sources.** Cloud → self-hosted (or self-hosted → self-hosted) migrations fail with `MIGRATION_SOURCE_PROJECT_NOT_FOUND` because the local lookup returns empty.

2. **Silent zero-row migrations on `projectId` collision.** Customers often recreate the destination with the **same id** to preserve cross-references (`createdBy`, `userId`, etc.). The lookup hits the local collision, the worker takes the DB fast path, and reports `status="completed"` with `0 rows transferred`.

## Fix

Treat the local `projectId` lookup as a *candidate*. Take the DB fast path only when the source URL's host belongs to this installation. Otherwise fall through to the SDK/HTTP path so the external source is read correctly.

The source URL's host must match one of:
- `_APP_DOMAIN` (exact + `*.` wildcard for subdomains)
- `_APP_MIGRATION_HOST` (internal host workers default to when `endpoint` is omitted)
- A custom API domain from the source project's `rules` collection (exact, scoped to the source project)

Match uses [`Utopia\Validator\Hostname`](https://github.com/utopia-php/validators) — the same validator `Appwrite\Network\Cors` uses for CORS allow-listing. Env values are normalized through `parse_url('http://' . $env, PHP_URL_HOST)` so a port suffix doesn't break equality.

```php
$isLocalSource = !$this->sourceProject->isEmpty()
    && (!$isAppwriteSource || ($isLocalEndpoint && $sourceRegion === $destinationRegion));
```

## Test plan
- [ ] CI green
- [ ] Cloud → self-hosted with **same projectId** on both sides transfers rows from the cloud project (not from the empty local project)
- [ ] Cloud → self-hosted with different projectIds completes end-to-end
- [ ] Intra-cluster same-host migration → still DB fast path
- [ ] Source URL on configured custom domain → still DB fast path

### DIFF
--- a/src/Appwrite/Platform/Workers/Migrations.php
+++ b/src/Appwrite/Platform/Workers/Migrations.php
@@ -198,15 +198,15 @@ class Migrations extends Action
         /** @var Database|null $projectDB */
         $projectDB = null;
         $useAppwriteApiSource = false;
-        if ($source === SourceAppwrite::getName() && empty($credentials['projectId'])) {
+        $isAppwriteSource = $source === SourceAppwrite::getName();
+        $isAppwriteToAppwrite = $isAppwriteSource
+            && $destination === DestinationAppwrite::getName();
+
+        if ($isAppwriteSource && empty($credentials['projectId'])) {
             throw new Exception(Exception::MIGRATION_SOURCE_PROJECT_ID_REQUIRED);
         }
 
-        if (! empty($credentials['projectId'])) {
-            $isAppwriteSource = $source === SourceAppwrite::getName();
-            $isAppwriteToAppwrite = $isAppwriteSource
-                && $destination === DestinationAppwrite::getName();
-
+        if ($isAppwriteSource) {
             $this->sourceProject = $this->dbForPlatform->getDocument('projects', $credentials['projectId']);
 
             // Same projectId may collide with an external Appwrite — trust the DB fast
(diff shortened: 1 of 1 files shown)


src/Appwrite/Platform/Workers/Migrations.php
```diff
@@ -202,18 +203,54 @@ protected function processSource(Document $migration): Source
...
             $sourceRegion = $this->sourceProject->getAttribute('region', 'default');
             $destinationRegion = $this->project->getAttribute('region', 'default');
-            $useAppwriteApiSource = $source === SourceAppwrite::getName()
-                && $destination === DestinationAppwrite::getName()
-                && $sourceRegion !== $destinationRegion;
-            if (! $useAppwriteApiSource) {
+
+            $isLocalSource = !$this->sourceProject->isEmpty()
+                && (!$isAppwriteSource || $isLocalEndpoint)
+                && (!$isAppwriteToAppwrite || $sourceRegion === $destinationRegion);
```
abnegate, line 245: Condition seems off, shouldn't any non-Appwrite source be considered non-local?

abnegate (approved):

Issue #11959, opened 2026-04-21
GraphQL endpoint omits CORS headers on POST responses (preflight works, actual response doesn't)


suhay:
### 👟 Reproduction steps

1. Register a Web platform with hostname localhost on any project.
2. From a page served at http://localhost:3000, issue a POST to /v1/graphql with Origin: http://localhost:3000.
3. Browser blocks the response with "CORS header 'Access-Control-Allow-Origin' missing" despite a 200 status.

```
curl -i -X OPTIONS -H "Origin: http://localhost:3000" -H "Access-Control-Request-Method: POST" -H "Access-Control-Request-Headers: x-appwrite-project,content-type,x-sdk-name,x-sdk-version,x-sdk-graphql" https://nyc.cloud.appwrite.io/v1/graphql
```

...
access-control-allow-origin: http://localhost:3000
...

```
curl -i -H "Origin: http://localhost:3000" -H "X-Appwrite-Project:<project-id>" -H "Content-Type: application/json" -d '{"query":"{ localeGet { country } }"}' https://nyc.cloud.appwrite.io/v1/graphql
```

content-type: application/json; charset=UTF-8
server: swoole-http-server
x-debug-fallback: false
x-debug-fallback: true
x-debug-speed: 0.1070020198822
accept-ranges: bytes
via: 1.1 varnish, 1.1 varnish
date: Tue, 21 Apr 2026 01:07:56 GMT
x-served-by: cache-lga21956-LGA, cache-chi-kigq8000163-CHI
x-cache: MISS, MISS
x-cache-hits: 0, 0
x-timer: S1776733676.482255,VS0,VE127
content-length: 50

### 👍 Expected behavior

POST /v1/graphql should echo the same access-control-allow-origin / access-control-allow-credentials / access-control-expose-headers the preflight already advertises.

### 👎 Actual Behavior

Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at
https://nyc.cloud.appwrite.io/v1/graphql. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 200.

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

Linux

### 🧱 Your Environment

SDK: 24.1.1
No customizations

Workaround: Created a proxy in Vite

```
server: {
    proxy: {
      '/appwrite': {
        target: 'https://nyc.cloud.appwrite.io',
        changeOrigin: true,
[...]

PR #12385, opened 2026-05-25
fix(graphql): add missing CORS headers to successful responses

deepshekhardas:
The CORS init hook was registered with groups(['api', 'web']), but all GraphQL routes use groups(['graphql']). Since the groups don't intersect, the CORS init hook never fires for GraphQL requests. Successful responses from GET /v1/graphql and POST /v1/graphql lack Access-Control-Allow-Origin, breaking cross-origin browser usage.

Add 'graphql' to the CORS init hook groups so it fires for GraphQL routes too. OPTIONS preflight and error handler already had CORS; this fixes successful responses.

Related: #11959


TorstenDittmann (approved):

TorstenDittmann:
@deepshekhardas thanks for the contribution <3

PR #12386, opened 2026-05-25
fix(builds): wrap vcsCommentLocks delete with Authorization::skip()

deepshekhardas:
The deleteDocument('vcsCommentLocks', ...) in Builds.php was not wrapped with Authorization::skip(), while three identical calls in Deployment.php are. Add the missing Authorization import, constructor injection, action parameter, and skip() wrapper to match the consistent pattern used elsewhere.

This is a code quality fix � in the current worker context authorization is disabled, so the delete works without skip(). However, the inconsistency creates a maintenance hazard; if worker authorization is ever enabled this would silently leak vcsCommentLocks documents.


eldadfux:

PR #12401, opened 2026-05-26
fix(databases): emit create events for relationship-cascaded child documents

### LINKED ISSUE #9041: 🐛 Bug Report: Events not triggered for relationships
### 👟 Reproduction steps

Create 2 collections(A and B). A has a one to many relationship to B. Create a function that needs to be triggered by events in B(create).

### 👍 Expected behavior

If event triggers are specified on the collection and document level, it shouldn't matter how the document is created. It would make the usage more uniform. Also, the workaround of creating the relation first and then link it to the parent might work, but I'm concerned about the reliability of that, given the api calls cannot be transactionally applied, and would complicate the implementation.

### 👎 Actual Behavior

The create event is not triggered when creating a new entry in B via the parent A. The event triggers only if the document in B is directly created in the collection.

### 🎲 Appwrite version

Version 1.6.x

### 💻 Operating system

Linux

### 🧱 Your Environment

appwrite 1.6.0, self-hosted locally

AdityaOP007:
# fix(databases): emit create events for relationship-cascaded child documents

Fixes #9041 

## What was happening

When creating related child documents through a relationship field, Appwrite was successfully inserting the child documents into the database, but `documents.*.create` events were not being emitted for those nested creates.

Because of this:

* Functions subscribed to child collection create events were not triggered
* Webhooks were skipped
* Realtime listeners never received the child create event

This only happened for nested relationship creates.

Direct document creation in the child collection worked as expected.

---

## Root cause

Nested relationship inserts are executed during the cascade flow and eventually run under a silent internal operation, so only the parent document create event was being queued by the HTTP layer.

As a result, child documents created implicitly through relationships bypassed the normal event dispatch pipeline.

---

### `Create.php`

* Added tracking for nested relationship-created documents using `$nestedCreates`
* While validating relationship permissions, newly created related documents are collected
* After `createDocuments()` succeeds, child create events are replayed explicitly before the parent event queue is finalized

### `Action.php`

* Added `triggerRelationshipCreates()` helper
* Re-loads created child documents
* Processes payloads using the normal document pipeline
* Dispatches:

  * Realtime events
  * Function triggers
  * Webhook triggers
* Resets queues afterward so parent events remain isolated

---

## Result

Now both flows behave consistently:

### Direct create

```js
createDocument("B")
```

✅ emits `documents.*.create`

### Nested relationship create

```js
createDocument("A", {
  children: [...]
})
```

✅ also emits `documents.*.create` for every newly created child

---

## Scope

This change currently covers:

* Single document create with nested relationship creates

This does not yet cover:

* Nested creates during update operations
* Bulk create flows
* TablesDB row APIs

---

## Manual testing

Tested with:

* Collection A → one-to-many → Collection B
* Function subscribed to:

```text
databases.*.collections.B.documents.*.create
```

### Before

* Direct B create → function triggered
* Nested child create via A → function NOT triggered

### After

* Direct B create → function triggered
* Nested child create via A → function triggered correctly per child document

---

## Additional notes

I also included a reproducible example + orchestration reference implementation under:

```text
examples/relationship-event-orchestration/
```

appwrite bug


eldadfux:

PR #11652, opened 2026-03-26
fix(storage): add missing encryption and compression properties to File model

BillionClaw:
The File model on main was missing the `encryption` and `compression` properties that exist in the 1.9.x branch. When the Flutter SDK (>= 20.3.3) tries to access `file.encryption`, it returns null instead of a boolean, causing a crash.

The fix adds both properties to the File model schema and a `filter()` method that derives them from the actual file attributes (`openSSLCipher` and `algorithm`), matching the pattern already present in 1.9.x.

Fixes #11647

PR #12405, opened 2026-05-27
fix(storage): add missing encryption/compression properties to File model with validated defaults

### LINKED ISSUE #11647: File encryption property is missing
### 👟 Reproduction steps

- Try to upload a file using Flutter appwrite library version > 20.3.3 (tried on 21)
It will crash because "encryption" property is null instead of boolean value.

Related to [https://github.com/appwrite/sdk-for-flutter/issues/306](https://github.com/appwrite/sdk-for-flutter/issues/306)

### 👍 Expected behavior

It should return a non null value

### 👎 Actual Behavior

It return a null value and then crash

### 🎲 Appwrite version

Version 1.8.x

### 💻 Operating system

Linux

### 🧱 Your Environment

_No response_

deepshekhardas:
The File model on main was missing the \encryption\ and \compression\ properties that exist in the 1.9.x branch. When the Flutter SDK (>= 20.3.3) tries to access \ile.encryption\, it returns null instead of a boolean, causing a crash.

### Changes
- Added \encryption\ (boolean) and \compression\ (string) fields to the File model schema
- Fixed compression default from \''\ to \Compression::NONE\ (prevents response contract violation)
- Added \ilter()\ method that derives compression/encryption from actual document attributes
- Validate compression against allowed values (\
one\/\gzip\/\zstd\)

### Fixes
Closes #11647

This also addresses the code review feedback from #11652 where compression default was flagged as \''\ contradicting the documented closed set.


Meldiron:
Rebased onto `main`.

This PR targeted `main` but its branch was cut from `1.9.x`, so the diff was showing ~51k files and GitHub refused to render it. Rebased onto `main` — the `File.php` change applied cleanly and it's now a 1-file, +6/-2 diff and mergeable.

Closed #12524, which was the `1.9.x`-targeted PR from this same branch.

Meldiron:
Simplified the filter and added E2E coverage.

**Dropped the whitelist check** — but it was load-bearing, so the underlying cause is fixed instead. `Document::getAttribute()` uses `isset()`, which only falls back to the default when the key is *unset*:

```php
public function getAttribute(string $name, mixed $default = null): mixed
{
    if (isset($this[$name])) {
        return $this[$name];
    }
    return $default;
}
```

`Files/Create.php` created the file document with `'algorithm' => ''` on the first chunk of a chunked upload, so `getAttribute('algorithm', Compression::NONE)` returned that empty string rather than `none` — which is what reached the SDK as an unparseable enum. The `in_array` guard was mapping it back to `none` after the fact.

So:
- `Files/Create.php` now writes `Compression::NONE` instead of `''`.
- The model falls back with `?:` rather than a default argument, which covers `''` and `null` in one operator and keeps existing documents correct — the `algorithm` column is `'default' => null, 'required' => false`, so both states exist in the wild.

```php
$document->setAttribute('compression', $document->getAttribute('algorithm') ?: Compression::NONE);
```

**E2E test** in `StorageBase` (so it runs for client, server and console), against a bucket with neither option configured:
- single-request upload — `compression` is `none`, `encryption` is `true` (bucket default, file under the buffer)
- chunked upload — asserts on *every* chunk response, including the first, which is the one that used to carry `''`

`assertSame`/`assertTrue`/`assertFalse` are strict, so a `null` or `''` fails them directly.

Meldiron (approved):

PR #12406, opened 2026-05-27
fix(webhooks): resolve contradictory SSL state in security opt-out block

deepshekhardas:
## Problem

PR #12205 proposes changing \CURLOPT_SSL_VERIFYPEER\ from \alse\ to \	rue\ inside the \security = false\ opt-out block in \src/Appwrite/Platform/Workers/Webhooks.php\. This creates a **contradictory SSL state**:

| Option | Value | Effect |
|--------|-------|--------|
| \CURLOPT_SSL_VERIFYPEER\ | \	rue\ | Verify certificate chain |
| \CURLOPT_SSL_VERIFYHOST\ | \alse\ | Don't verify hostname |

This block only executes when a user has explicitly **disabled** SSL verification (\security = false\). Setting \VERIFYPEER=true\ while \VERIFYHOST=false\ causes cURL SSL errors for any webhook targeting a server with a self-signed or private-CA certificate.

## Current code (correct)

\\\php
if (!\->getAttribute('security', true)) {
    curl_setopt(\, CURLOPT_SSL_VERIFYHOST, false);
    curl_setopt(\, CURLOPT_SSL_VERIFYPEER, false);
}
\\\

cURL already defaults to \CURLOPT_SSL_VERIFYPEER = true\, so the default (secure) path is already protected. The \security = false\ block is an explicit opt-out that correctly disables both peer and host verification.

## Summary

- ❌ PR #12205's Webhooks.php change introduces a P1 regression
- ✅ The original code is correct and should remain unchanged
- 🔒 Default SSL behavior is already secure (cURL defaults)


eldadfux:

PR #12432, opened 2026-05-30
fix(users): return 400 instead of 500 when name is null on create endpoints

sridhar-3009:
## Summary

Fixes #8785.

All `Users.Create*` endpoints declared the `name` parameter with
`new Text(128)`, which does not accept `null`. When a client explicitly
sends `"name": null` in the JSON body the Utopia framework tried to
validate `null` through `Text::isValid()`, which throws internally,
resulting in an unhandled exception and a vague HTTP 500.

The same pattern was already correctly handled for `email` and `phone`
using `new Nullable(new Email(...))` / `new Nullable(new Phone())`.
This PR applies the identical fix to `name` across all eight
`Users.Create*` endpoints.

## What changed

`app/controllers/api/users.php`:
- Changed `->param('name', '', new Text(128), ...)` to
  `->param('name', null, new Nullable(new Text(128)), ...)` on lines
  covering `/v1/users`, `/v1/users/bcrypt`, `/v1/users/md5`,
  `/v1/users/argon2`, `/v1/users/sha`, `/v1/users/phpass`,
  `/v1/users/scrypt`, and `/v1/users/scryptMod`.

`Nullable` is already imported (`use Utopia\Validator\Nullable;`).

## Behaviour after fix

`Nullable` is a pass-through when the value is `null` — it returns
`isValid() = true` and the value reaches `createUser()` where
`$name = $name ?? ''` coerces it to an empty string. This is consistent
with how `email` and `phone` behave.

| Request | Before | After |
|---------|--------|-------|
| `"name": null` | HTTP 500 (unhandled exception) | HTTP 201 (user created, name = "") |
| `"name": ""` | HTTP 201 | HTTP 201 (no change) |
| `"name": "Alice"` | HTTP 201 | HTTP 201 (no change) |
| name omitted | HTTP 201 (default `""`) | HTTP 201 (no change) |

## Test plan

- [ ] POST `/v1/users` with `"name": null` → expect 201, not 500
- [ ] POST `/v1/users/bcrypt` with `"name": null` → expect 201
- [ ] POST `/v1/users` without `name` field → expect 201 (no regression)
- [ ] POST `/v1/users` with `"name": "Alice"` → expect 201 (no regression)


sridhar-3009:
Good catch @greptile-apps — the PR description table was wrong on my end. Thank you for flagging it.

The **intended behavior** here is 201 (user created with empty name), not 400. The `Nullable` wrapper is exactly right because:

- `name` is an optional field (same as `email` and `phone`)
- `email` and `phone` both use `Nullable` and correctly accept `null` → 201
- `name` should be consistent with that pattern
- The existing `$name = $name ?? '';` guard in `createUser()` safely coerces null → empty string

The bug in #8785 was that passing `null` for `name` caused a **500** (unhandled exception) rather than behaving like the other optional fields. The fix is: `null` should be accepted and produce a **201** with an empty name — not a 400.

I'll update the PR description to correct the table.

PR #12434, opened 2026-05-30
Fix: Handle null name parameter in createUser endpoints

Jualhosting:
## Before
The \createUser\ API endpoints (standard, bcrypt, md5, argon2, etc.) were using a strict \Text(128)\ validator for the \
ame\ parameter. When clients (like the Appwrite dashboard) sent a \
ull\ value for \
ame\, it immediately failed validation and threw a 500 Internal Server Error, rather than falling back to an empty string.

## After
I wrapped the \Text(128)\ validator with \Nullable()\ (i.e., \
ew Nullable(new Text(128))\) for the \
ame\ parameter across all relevant \createUser\ endpoints. This allows \
ull\ payloads to pass validation and safely hit the controller logic, which handles the fallback (\$name ?? ''\), resolving the 500 error and allowing user creation.

## AI Disclosure
Generated by Google Gemini / Antigravity. This code has been fully reviewed and tested by the admin of Jualhosting, so it is not entirely AI-generated.


HarshMN2345:

PR #12435, opened 2026-05-30
Fix 8000 jwt session expiry

### LINKED ISSUE #8000: 🐛 Bug Report: JWT still valid after Session Timeout

Jualhosting:
### Before
The `user` resource handling in `request.php` validated JWT sessions by only checking if the `sessionId` existed in the database using `$user->find('$id', $jwtSessionId, 'sessions')`. It did not verify if the session had expired, allowing expired JWTs to continue accessing authenticated endpoints like `/account`.

### After
I updated the JWT validation logic to fetch the full session document and explicitly check its `expire` attribute against the current time using `DatabaseDateTime::formatTz()`. If the session is not found, lacks an expiry, or the expiry time is in the past, the user is correctly reset to an empty state, invalidating the JWT.

Fixes #8000

AI Disclosure
Generated by Google Gemini / Antigravity. This code has been fully reviewed and tested by the admin of Jualhosting, so it is not entirely AI-generated.


eldadfux:
Closed in favor of #13125

PR #12444, opened 2026-05-31
fix(users): ensure prefs update event returns full user payload

deepshekhardas:
Previously the \users.[userId].update.prefs\ event only sent the user ID as a parameter but did not include a payload. Function/webhook subscribers received an empty event payload.

This fix sets the full user object as the event payload so that subscribers can access user properties (including prefs) without an additional API call.

---
Rebased onto official/1.9.x (b9052d52d0). Force-pushed Tue Jun 2 2026.


deepshekhardas:
@loks0n @ChiragAgg5k this PR has been rebased onto latest main. Could you please review? 🙏

PR #12450, opened 2026-06-01
Revert template root directory path-stripping changes

loks0n:
## Summary

Reverts the path-normalization changes from PR #12413 ("refactor: tighten internal validation paths") and the follow-up PR (8c8f7b7c48 "Normalize template root directories in builds") that together broke production builds.

History:
- Pre-#12413: `rtrim('/')` → `ltrim('.')` → `ltrim('/')` normalized template/VCS root directories, handling `./foo` correctly.
- #12413 swapped that for `trim('/')` + an explicit `..` regex rejection — intended to reject path traversal instead of silently stripping it, but inadvertently dropped `./` handling.
- 8c8f7b7c48 tried to patch `./` handling back in via a `while (str_starts_with(..., './'))` loop. This broke all builds in production.

This PR restores the original pre-#12413 stripping at all three sites in `Builds.php` (two template paths, one VCS path) and reverts the test relaxations in `SitesCustomServerTest.php` that were tied to the broken behavior.

Trade-off: this re-introduces the silent stripping of leading dots, so `../foo` is normalized to `foo` rather than rejected. We'll harden this properly in a follow-up — handling `./` while still rejecting `..` traversal.

## Test plan
- [ ] Verify builds succeed in production after merge
- [ ] Template-based function/site deployments work end-to-end (tag, branch, commit references)
- [ ] Follow-up: re-add `..` traversal rejection without breaking `./` handling

### DIFF
--- a/src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
+++ b/src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
@@ -297,10 +297,9 @@ class Builds extends Action
                 $templateReferenceValue = $template->getAttribute('referenceValue', '');
 
                 $templateRootDirectory = $template->getAttribute('rootDirectory', '');
-                $templateRootDirectory = \trim($templateRootDirectory, '/');
-                if ($templateRootDirectory !== '' && \preg_match('#(^|/)\.\.(/|$)#', $templateRootDirectory)) {
-                    throw new \Exception('Invalid template root directory');
-                }
+                $templateRootDirectory = \rtrim($templateRootDirectory, '/');
+                $templateRootDirectory = \ltrim($templateRootDirectory, '.');
+                $templateRootDirectory = \ltrim($templateRootDirectory, '/');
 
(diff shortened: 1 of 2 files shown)


loks0n:
Thanks @greptile-apps — both findings are accurate and worth recording:

1. **`..` traversal guard dropped.** Acknowledged — this is the explicit trade-off called out in the PR description. Restoring pre-#12413 behavior was the priority to unblock production builds; re-adding the `..` rejection without re-breaking `./` handling is the follow-up.
2. **`ltrim($str, '.')` strips all leading dots, not just `./`.** Also correct, and a sharper observation — `.next`, `.nuxt`, `.svelte-kit` would be silently corrupted to `next` / `nuxt` / `svelte-kit`. This is pre-existing behavior from before #12413 (i.e. it has shipped in production for a long time and isn't a regression introduced by this PR), but it should be fixed alongside the `..` hardening in the follow-up rather than carried forward indefinitely.

Follow-up scope: replace the three-step `rtrim`/`ltrim`/`ltrim` with normalization that strips exactly a leading `./` (not all dots) and rejects `..` segments anywhere in the path.

Meldiron (approved):

PR #12452, opened 2026-06-01
chore: trace database worker with Span::add instead of Console logging

loks0n:
## What

Replaces unstructured `Console::info`/`Console::error` logging in the database worker (`src/Appwrite/Platform/Modules/Databases/Workers/Databases.php`) with structured `Span::add` tracing.

## Changes

- **Operation context up front** — the trailing `Console::info("Finished processing…")` JSON dump is replaced with spans added at the start of `action()`: `project.id`, `payload.type`, `queue.pid`, `queue.name`, `message.timestamp`, `database.id`, `database.collection.id`, `database.document.id` (following the existing Functions worker convention).
- **Rethrowing catch blocks** (`createAttribute`, `deleteAttribute`, `createIndex`, `deleteIndex`) — removed `Console::error($e->getMessage())`. The entry-point harness owns span error lifecycle, and the rethrown exception already carries the message, so these were redundant.
- **Swallowed cases surfaced as spans** so they stay visible:
  - `deleteAttribute` `NotFound` catch → `database.attribute.not_found`
  - `deleteByGroup` swallowed error → `delete_by_group.error`; success count → `delete_by_group.count` (dropped manual `microtime` timing since the harness already times the worker span)
- Removed the now-unused `Utopia\Console` import; added `Utopia\Span\Span`.

The structured `$log->addTag(...)` logger tags are kept (they feed error reporting, not unstructured console output).

## Notes

`composer format` / `composer analyze` should be run in the Docker container before merge — vendor binaries aren't present in the local checkout. `php -l` passes with no syntax errors.

### DIFF
--- a/src/Appwrite/Platform/Modules/Databases/Workers/Databases.php
+++ b/src/Appwrite/Platform/Modules/Databases/Workers/Databases.php
@@ -5,7 +5,6 @@ namespace Appwrite\Platform\Modules\Databases\Workers;
 use Appwrite\Event\Message\Database as DatabaseMessage;
 use Appwrite\Event\Realtime;
 use Exception;
-use Utopia\Console;
 use Utopia\Database\Database;
 use Utopia\Database\Document;
 use Utopia\Database\Exception as DatabaseException;
@@ -18,6 +17,7 @@ use Utopia\Database\Query;
 use Utopia\Logger\Log;
 use Utopia\Platform\Action;
 use Utopia\Queue\Message;
+use Utopia\Span\Span;
 
 class Databases extends Action
 {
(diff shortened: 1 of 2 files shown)


src/Appwrite/Platform/Modules/Databases/Workers/Databases.php
```diff
@@ -609,27 +605,20 @@ function ($attribute) use ($dbForProject, $databaseInternalId) {
...
-                $queries,
-                Database::DELETE_BATCH_SIZE,
-                $callback
-            );
-        } catch (\Throwable $th) {
-            $tenant = $database->getSharedTables() ? 'Tenant:'.$database->getTenant() : '';
-            Console::error("Failed to delete documents/rows for collection/table: {$database->getNamespace()}_{$collectionId} {$tenant} :{$th->getMessage()}");
-            return;
-        }
+        $count = $database->deleteDocuments(
```
fogelito: Try catch was removed + return

loks0n: returned it

PR #12468, opened 2026-06-02
Add coroutine concurrency to queue workers

loks0n:
## What does this PR do?

Upgrades `utopia-php/queue` to **0.19.0** and `utopia-php/platform` to **1.0.0-rc4**, and migrates the worker harness so a single worker process can consume multiple jobs concurrently via Swoole coroutines.

### Changes

- **`app/worker.php`** — migrate to the new queue Server API: per-message DI resources move from the shared container to a coroutine-scoped context container (`Server::context()`, backed by `Coroutine::getContext()`), so concurrent jobs get isolated resources. The adapter takes `maxCoroutines` from the new `_APP_WORKER_MAX_COROUTINES` env var (default `1` — no behavior change unless opted in).
- **`app/init/registers.php`** — consumer Redis connections are wrapped in queue 0.19's `Locking` decorator, serializing commands on the single connection shared by the consume coroutines.
- **`docker-compose.yml` / `.env`** — every worker runs a single process (`_APP_WORKERS_NUM=1`) with an explicit per-worker coroutine count (see matrix below).
- **`src/Appwrite/Platform/Workers/Messaging.php`** — bug fix: `getLocalDevice()` cached a project-scoped `Local` device unkeyed, reusing the first project's storage path for every subsequent message; now created per call.

### Per-worker concurrency matrix

Based on a coroutine-safety audit of every worker action (details in [the audit comment](https://github.com/appwrite/appwrite/pull/12468#issuecomment-4606822276)):

| 8 coroutines | 1 coroutine (serial) | Why serial |
|---|---|---|
| deletes, mails, certificates, executions, functions, builds, screenshots, messaging | databases | schema op ordering |
| | audits, stats-usage | cross-message batch accumulators on shared instance state |
| | stats-resources, webhooks | shared per-job buffers reset/flushed across yield points |
| | migrations | project DB callables rebound on the instance per message |

### Load test results (functions worker, single process)

1s-sleep function, 30 async executions fired at once, wall time until the queue drained:

| `_APP_WORKER_MAX_COROUTINES` | Drain time | Throughput |
|---|---|---|
| 1 | 31.1s | 0.96 exec/s |
| 5 | 10.4s | 2.88 exec/s |
| 10 | 7.3s | 4.13 exec/s |

Worth noting: at 1 coroutine the 30-execution backlog exceeded a 15s function timeout, which makes pending executions *report* as failed to clients (`Executions/XList.php` timeout presentation) — faster queue drain directly mitigates that.

Span tracing is coroutine-safe already (`Storage\Coroutine`). `workerStart`/`workerStop` hooks and app-wide services stay on the shared resources container, matching previous semantics.

## Test plan

- Functions load test passes at 1, 3, 5, 8 and 10 coroutines (all executions complete, zero failures) — harness not committed, numbers above
- `tests/e2e/Services/Account --filter=testGetAccountLogs` passes with the audits worker concurrent
- All 14 worker containers start clean on queue 0.19.0
- PHPStan + Pint clean on changed files

## Related PRs and issues

- utopia-php/queue 0.19.0 (Swoole coroutine concurrency, `Locking` connection decorator)
- utopia-php/platform 1.0.0-rc4 (queue `^0.19` constraint)


loks0n:
Updated the env wiring per review discussion:

- `.env` defaults: `_APP_WORKERS_NUM=1`, `_APP_WORKER_MAX_COROUTINES=1`
- Every worker service in `docker-compose.yml` now sets both explicitly: single process per worker, with concurrency coming from coroutines instead
- **8 coroutines** for I/O-bound workers: audits, webhooks, deletes, screenshots, certificates, executions, functions, mails, messaging
- **1 coroutine (serial)** for CPU-heavy / stateful / ordering-sensitive workers: databases (schema op ordering), builds (archive handling), migrations (long-running stateful jobs), stats-resources, stats-usage (in-memory metric batching)

Verified: full stack recreated with the new matrix, functions load test passes at 8 coroutines, all workers start clean.

loks0n:
## Coroutine-safety audit of all worker actions

The worker Action objects are per-process singletons (the queue `job` hook binds `action(...)` once), so **any instance property mutated during `action()` is shared across all consume coroutines**, and every DB/HTTP call is a yield point where another message interleaves.

### Unsafe at >1 coroutine → pinned to 1 in compose

| Worker | Shared state | Failure mode |
|---|---|---|
| **stats-usage** | `$stats`, `$keys`, `$projects`, `$statDocuments` — batches across messages by design | Flush reset wipes other coroutines' uncommitted aggregates (lost usage); overlapping flushes double-apply `upsertDocumentsWithIncrease` (double-billed usage) |
| **stats-resources** | `$documents` reset at start of each job | Second coroutine wipes first's half-built batch mid-count; both projects' metrics interleave into one project's logsDB write (cross-tenant corruption) |
| **audits** | `$logs`, `$lastTriggeredTime` — cross-message batching with `Commit`/`NoCommit` ack semantics | Concurrent flush loses or duplicates audit entries; messages can be acked whose logs were just wiped |
| **webhooks** | `$errors` reset per job, drives the retry/attempts/disable path | Concurrent jobs wipe each other's errors (failed deliveries never retried) or steal them (spurious retries → duplicate deliveries) |
| **migrations** | `$getProjectDB`/`$getDatabasesDB`/etc. rebound per message, used throughout a long migration | Concurrent message swaps project DB handles mid-migration (cross-project writes) |
| **databases** | (stateless, but schema ops are ordering-sensitive) | — |

### Clean → stay at 8

deletes, mails, certificates, executions, functions, builds, screenshots, messaging — all per-message state lives in locals.

### Bug fix included

`Messaging::getLocalDevice()` cached a **project-scoped** `Local` device unkeyed on the instance — the first project's `app-{id}` storage path was reused for every subsequent message in the process (pre-existing bug even serially; coroutines widen it). Now created per call.

The serialized workers can graduate to 8 later by moving accumulation into per-message context (`stats-resources`, `webhooks` are easy; `audits`/`stats-usage` need their cross-message batching guarded by a mutex or per-coroutine accumulators).

loks0n:
CI failures diagnosed — two distinct causes, both fixed in `4baaa7d93f`:

1. **Pool exhaustion** — connection pools are sized as `instanceConnections / (CPU × WORKER_PER_CORE)`, which resolves to **1 per worker process** on CI runners. 8 consume coroutines starved them (`Pool 'cache_redis_main' is empty (size 1)` in worker-messaging/functions logs). Pools are now floored at `_APP_WORKER_MAX_COROUTINES`.

2. **SMTP workers freeze the coroutine scheduler** — `Mails.php:63` and `Messaging.php:97` call `Runtime::setHookFlags(SWOOLE_HOOK_ALL ^ SWOOLE_HOOK_TCP)` inside the action to dodge the Swoole TLS-handshake crash (swoole/swoole-src#4909, #7914). The flag is **process-global**: after the first job, every blocking 2s Redis `BRPOP` from the other 7 consume coroutines blocks the whole event loop, so deliveries arrived ~40s+ late — exactly why the email/SMS-dependent suites (Account, Teams, Projects, …) timed out waiting on maildev/request-catcher. Both workers are pinned to 1 coroutine (identical to pre-PR behavior) until the TCP-hook workaround can be scoped.

Verified locally: the previously failing `testCreateSession`/`testCreateSessionWithPhone`/`testPhoneVerification` now pass, SMS lands in the catcher within the poll window.

Updated matrix: **8 coroutines** — deletes, builds, screenshots, certificates, executions, functions. **Serial** — databases, audits, webhooks, migrations, stats-usage, stats-resources, mails, messaging.

loks0n:
Storage failure root-caused and fixed in `78769b7f13` — it exposed a real design flaw in the shared-connection approach:

**The bug:** `testFilePreviewCache` depends on the async cache purge (deletes worker) completing within a few hundred ms of the file delete. With the `Locking`-wrapped shared connection, every idle consume coroutine holds the mutex for a full **2s blocking `BRPOP`**. On a quiet queue, a finished job's completion ops (remove job key, increment stats, pop next) starve behind the other 7 poppers — measured locally: a **10ms cache purge took 22-24s to fully ack**, and the next purge popped ~12s after enqueue. The test then read a stale preview. (The functions load test never caught this because a *full* queue makes `BRPOP` return instantly — the pathology only appears at low traffic.)

**The fix:** the adapter's consumer is now a `Queue\Broker\Pool` — each consume coroutine leases its **own connection** from the consumer pool for the lifetime of its loop. No shared connection, no mutex, no starvation. The `Locking` wrapper is reverted entirely, and consumer pools get +2 headroom above the coroutine count (the publisher side of the same pool backs the queue-depth gauge). Purge ack latency: **22-24s → 4-58ms**.

Verified locally: `testFilePreviewCache` passes 7/8 runs (single miss was a cold-start run 15s after full stack recreation, before workers finished booting — the test inherently races worker boot).

Tradeoff worth noting: `Broker\Pool::close()` is currently a no-op upstream, so SIGTERM doesn't break the blocking pop early — workers rely on the process kill that already follows. Could be fixed in utopia-php/queue by implementing pool close.

PR #12467, opened 2026-06-02
Test PR - Tune PostgreSQL CI Flakeness

premtsd-code:
## What

Runs `composer update utopia-php/database` on `1.9.x`.

`utopia-php/database` is already at the latest `5.*` release (`5.9.0`), so it stays put; the resulting `composer.lock` change is a transitive bump of `symfony/polyfill-php85` (`v1.37.0` -> `v1.38.1`).

This also applies PostgreSQL-only CI settings for the E2E matrix:

- `fsync=off` disables flushing data to disk after writes. This makes writes much faster, but if the runner crashes, loses power, or the container is killed unexpectedly, database corruption or data loss can occur.
- `synchronous_commit=off` lets PostgreSQL report transaction commit success without waiting for the commit to be physically written to disk. This improves transaction throughput, but recent committed transactions can be lost during a crash.
- `full_page_writes=off` disables writing full data pages to WAL after a checkpoint. This reduces WAL volume and improves performance, but increases the risk of corruption after crashes.

## Why

Refresh locked dependencies and run the full CI matrix against the updated lockfile.

The PostgreSQL settings are limited to automated CI databases where durability does not matter. These settings are commonly used for automated tests, CI/CD pipelines, temporary development databases, and benchmarks where crash recovery is not required.

PR #12469, opened 2026-06-02
Tune PostgreSQL CI Flakeness

premtsd-code:
Full CI Matrix Test Performed in Test PR - https://github.com/appwrite/appwrite/pull/12467

## What

Adds a PostgreSQL-only Docker Compose override for the E2E CI matrix.

The override starts PostgreSQL with:

- `fsync=off`, which disables flushing data to disk after writes. This makes writes much faster, but if the runner crashes, loses power, or the container is killed unexpectedly, database corruption or data loss can occur.
- `synchronous_commit=off`, which lets PostgreSQL report transaction commit success without waiting for the commit to be physically written to disk. This improves transaction throughput, but recent committed transactions can be lost during a crash.
- `full_page_writes=off`, which disables writing full data pages to WAL after a checkpoint. This reduces WAL volume and improves performance, but increases the risk of corruption after crashes.

## Why

The settings are limited to automated CI databases where durability does not matter. They are commonly used for automated tests, CI/CD pipelines, temporary development databases, and benchmarks where crash recovery is not required.

This does not change MariaDB, MongoDB, Redis, persistent volumes, or local default Compose behavior unless the CI PostgreSQL matrix selects the override file.

## Validation

The same PostgreSQL command override was exercised in a full CI matrix run on a separate test PR. The remaining failure in that run was in MariaDB shared Storage image preview cache coverage, so it was unrelated to the PostgreSQL override.

### DIFF
--- a/docker-compose.ci-postgres.yml
+++ b/docker-compose.ci-postgres.yml
@@ -1,0 +1,7 @@
+services:
+  postgresql:
+    command: >
+      postgres
+      -c fsync=off
+      -c synchronous_commit=off
+      -c full_page_writes=off
--- a/docker-compose.ci-postgres.yml
+++ b/docker-compose.ci-postgres.yml
@@ -1,7 +1,0 @@
-services:
-  postgresql:
-    command: >
-      postgres
-      -c fsync=off
-      -c synchronous_commit=off
-      -c full_page_writes=off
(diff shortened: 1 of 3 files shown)


docker-compose.ci-postgres.yml
```diff
@@ -0,0 +1,7 @@
+services:
+  postgresql:
+    command: >
+      postgres
+      -c fsync=off
+      -c synchronous_commit=off
+      -c full_page_writes=off
```
abnegate: Let's apply directly in docker-compose.yml since it's the dev file anyway

abnegate (approved):

PR #12485, opened 2026-06-03
Fix MFA authenticator selection and add flow tests

TorstenDittmann:
## What
- Order the user authenticator subquery by creation time so newly created TOTP authenticators are selected before stale older records.
- Add Account E2E coverage for TOTP authenticator creation, verification, challenge usage, and deletion.
- Add coverage for email/password sessions requiring and completing TOTP MFA.
- Add coverage for recovery-code regeneration requiring a recent successful challenge.

## Why
Older accounts can end up with duplicate or stale TOTP authenticator records. Without deterministic ordering, `users.authenticators` can return an old record first, causing `updateMFAAuthenticator` to validate against a different secret than the one returned by `createMFAAuthenticator`.

## Tests
- `php -l app/init/database/filters.php`
- `composer lint app/init/database/filters.php`
- `php -l tests/e2e/Services/Account/AccountCustomClientTest.php`
- `composer lint tests/e2e/Services/Account/AccountCustomClientTest.php`
- `docker compose exec appwrite test tests/e2e/Services/Account --filter='testMFARecoveryCodeChallenge|testRegenerateMFARecoveryCodesRequiresRecentChallenge|testDeleteMFAAuthenticatorRequiresRecentChallenge|testMFAAuthenticatorLifecycle|testMFAAuthenticatorCompletesEmailPasswordSession'`


TorstenDittmann:
Addressed Greptile's TOTP replay concern in 8438ef65d5:

- Added a test helper that waits for the next TOTP value before using the same authenticator for a later TOTP challenge.
- Switched tests that only need a generic recent successful challenge to recovery-code challenges, so they don't unnecessarily reuse TOTP codes or wait for TOTP rotation.

Re-ran:
- php -l tests/e2e/Services/Account/AccountCustomClientTest.php
- composer lint tests/e2e/Services/Account/AccountCustomClientTest.php
- docker compose exec appwrite test tests/e2e/Services/Account --filter='testMFARecoveryCodeChallenge|testRegenerateMFARecoveryCodesRequiresRecentChallenge|testDeleteMFAAuthenticatorRequiresRecentChallenge|testMFAAuthenticatorLifecycle|testMFAAuthenticatorCompletesEmailPasswordSession'

Result: OK (5 tests, 100 assertions)

TorstenDittmann:
@greptile-apps please review the latest changes. I addressed the TOTP replay concern by avoiding reuse of the same TOTP code for enrollment and later challenge verification, and by using recovery-code challenges where the test only needs a generic recent successful challenge.

tests/e2e/Services/Account/AccountCustomClientTest.php
```diff
@@ -131,6 +131,16 @@ protected function createFreshAccountWithSession(): array
...
 
+    protected function getNextTOTP(\OTPHP\TOTP $totp, string $previousOtp): string
+    {
+        do {
+            sleep(1);
+            $otp = $totp->now();
+        } while ($otp === $previousOtp);
+
+        return $otp;
+    }
```
loks0n, line 148: possible to use assertEventually/or bound this with a timeout to avoid hanging in case it fails?

TorstenDittmann: yeah let me add that

TorstenDittmann:
Addressed @loks0n review feedback on `getNextTOTP()`.

Change: added a 35s deadline so the helper fails explicitly instead of hanging if the TOTP value does not rotate.

Validation:
- `php -l tests/e2e/Services/Account/AccountCustomClientTest.php`
- `composer lint tests/e2e/Services/Account/AccountCustomClientTest.php`

I also tried the focused Account MFA test filter in the running Docker stack, but this container only sees/executes the pre-existing `testMFARecoveryCodeChallenge` method, so it does not appear to be using the latest test file state.

PR #12486, opened 2026-06-04
feat(messaging): add Appwrite Push provider

abnegate:
## Summary

Adds **Appwrite Push** — a self-hosted MQTT 5 push provider designed as a low-power, FCM-class alternative for Android. Devices hold a single persistent TLS connection to an Appwrite-shipped broker; the messaging worker dispatches via the existing `PushAdapter` interface, so consumers of the messaging API see another provider slot next to FCM/APNS.

**Blocked on:** [utopia-php/messaging#122](https://github.com/utopia-php/messaging/pull/122) — adds the publisher adapter + MQTT 5 codec this branch depends on. Composer require bumped `1.0.* → 1.1.*`; will need `composer update` once 1.1.0 is tagged.

### What's included

**REST**
- `POST /v1/messaging/providers/appwrite` + `PATCH /v1/messaging/providers/appwrite/:providerId` — mirrors FCM/APNS shape, persists `endpoint` + `signingKey` to encrypted credentials and `tls` + `messageExpiry` to options
- `POST /v1/account/targets/:targetId/push/session` — exchanges an authenticated session for `{endpoint, topic, clientId, token, keepAlive, expiresAt}` so SDKs can swap FCM registration without owning signing material

**Worker**
- `src/Appwrite/Platform/Workers/Messaging.php` resolves the new provider via `Utopia\Messaging\Adapter\Push\Appwrite`, falling back to `_APP_PUSH_ENDPOINT`/`_APP_PUSH_SIGNING_KEY` when per-provider creds are blank

**Broker process**
- `app/push.php` + `bin/push` entrypoint
- `src/Appwrite/Push/Broker.php` Swoole TCP/TLS broker enforcing JWT auth, per-device topic ACLs, retain-on-disconnect, idle-fd reaping
- `src/Appwrite/Push/Token.php` issues + verifies device and server JWTs against shared HMAC

**Config / ops**
- `_APP_PUSH_*` variables in `.env`, `app/config/variables.php`, both install templates, `docker-compose.yml`
- `appwrite-push` service mirrors the other Swoole workers' patterns
- `Dockerfile` chmods the new entry script

**Models / docs**
- `PushSession` registered in `app/init/models.php`, exposed as `Response::MODEL_PUSH_SESSION`
- SDK reference stubs for `create-push-session`, `create-appwrite-provider`, `update-appwrite-provider`

## Test plan

- [x] `php -l` on every touched file — clean
- [x] `pint --preset psr12 --test` on all touched files — pass
- [x] `tests/unit/Push/TokenTest.php` exercises issue/verify, wrong-key, expiry, garbage input (written; not run locally since worktree has no vendor)
- [x] `tests/e2e/Services/Messaging/MessagingBase.php` extended with an `appwrite` parameter row so the existing create/update provider cases exercise the new endpoints — `_APP_PUSH_*` defaults set in `.env` so CI picks them up automatically
- [ ] Boot the docker-compose stack and confirm `appwrite-push` comes up + accepts a connection — to verify after the messaging PR is tagged and `composer update` ships the adapter
- [ ] End-to-end: create provider via REST, create push target, call `POST /push/session`, connect a real MQTT client with the returned JWT, send a message via `messaging.createPush`, observe delivery on the device — same caveat as above

## Notes for reviewers

- The broker is a single-process Swoole server. Multi-instance fan-out needs Redis pub/sub between instances — explicitly out of scope here and called out in the file-level comment.
- QoS 1 only; QoS 0 publishes are upgraded transparently, QoS 2 is rejected with malformed-packet. Retained-on-disconnect uses a Swoole `Table` sized to 8192 pending messages with per-message TTL.
- Keep-alive defaults to **30 minutes** (`_APP_PUSH_KEEPALIVE=1800`) — picked specifically to minimise Android radio wake-ups while staying well inside reasonable NAT timeout windows.
- Adapter-side JWT TTL is 60 seconds (refreshed per send batch); device-side JWT TTL is 24h by default, capped at 7d via the `expirySeconds` param on `/push/session`.


gauravbsinghal:
Hi @abnegate, I'm sharing this from [Niro](https://github.com/apxlabs-ai/niro), a PR pentesting tool we're building to help OSS maintainers catch security regressions before merge. I ran it against this PR and it found the following issue.

An account user with or able to create a push target for a disabled Appwrite provider can still mint broker settings and a signed device JWT, so the disabled state no longer stops device broker-session issuance for that project.

## Reproduction Steps

Precondition: use project API key CR_005 against project `harness-project`.

1. Create a disabled Appwrite push provider:
`curl -i -X POST 'http://host.docker.internal:8080/v1/messaging/providers/appwrite' -H 'X-Appwrite-Project: harness-project' -H 'X-Appwrite-Key: $CR_005_SECRET' -H 'Content-Type: application/json' --data '{"providerId":"disabled_apw_1781460548","name":"Disabled Appwrite","endpoint":"127.0.0.1:1883","signingKey":"disabled-signing-key","tls":false,"enabled":false}'`
Observed: `201 Created` with `"$id":"disabled_apw_1781460548"` and `"enabled":false`.

2. Create an anonymous user session and keep the session cookie:
`curl -i -c /tmp/tc9.cookies -X POST 'http://host.docker.internal:8080/v1/account/sessions/anonymous' -H 'X-Appwrite-Project: harness-project' -H 'Content-Type: application/json' --data '{}'`
Observed: `201 Created` and an anonymous session for user `6a2eee446bb0861e1daa`.

3. As that user, create a push target bound to the disabled provider:
`curl -i -b /tmp/tc9.cookies -c /tmp/tc9.cookies -X POST 'http://host.docker.internal:8080/v1/account/targets/push' -H 'X-Appwrite-Project: harness-project' -H 'Content-Type: application/json' --data '{"targetId":"target_disabled_1781460548_1","identifier":"device-disabled-1781460548-1","providerId":"disabled_apw_1781460548"}'`
Observed: `201 Created` with `"providerId":"disabled_apw_1781460548"`.

4. Request a broker session for that target:
`curl -i -b /tmp/tc9.cookies -c /tmp/tc9.cookies -X POST 'http://host.docker.internal:8080/v1/account/targets/target_disabled_1781460548_1/push/session' -H 'X-Appwrite-Project: harness-project' -H 'Content-Type: application/json' --data '{}'`
Expected secure response: reject the disabled provider with a 4xx and do not return broker credentials.
Observed vulnerable response: `201 Created` with broker credentials for the disabled provider, including `"endpoint":"127.0.0.1:1883"`, `"topic":"appwrite/push/device-disabled-1781460548-1"`, `"clientId":"device-disabled-1781460548-1"`, a signed JWT in `token`, and `"expiresAt":"2026-06-15T18:09:08.000+00:00"`.

5. Reproduced the same issue a second time with a new anonymous user and target `target_disabled_1781460548_2`; the endpoint again returned `201 Created` and a fresh broker JWT for `device-disabled-1781460548-2`.

## Remediation Guidance

In `app/controllers/api/account.php` inside `POST /v1/account/targets/:targetId/push/session`, make the explicit `providerId` branch enforce the same enabled-only rule as the default-provider branch. After loading `$provider`, reject it when `enabled` is not `true` instead of minting a session. The smallest fix is to extend the existing explicit-branch guard to include `!$provider->getAttribute('enabled')` (and return the same disabled/not-found behavior as the default `findOne(... enabled = true)` path), so disabled Appwrite providers never reach the JWT issuance block.

abnegate:
Closing: superseded by https://github.com/appwrite/appwrite/pull/13372. Superseded by #13372 'Appwrite Native Push Broker' (ArnabChatterjee20k, merged 2026-09-18), which shipped the same feature with an in-repo MQTT 5 broker (app/mqtt.php), an in-repo push adapter (src/Appwrite/Messaging/Adapter/Push/Appwrite.php) and the providers/appwrite create/update routes. Main th

PR #12487, opened 2026-06-04
fix(oidc): allow clearing maxAge via explicit null in OIDC update

deepshekhardas:
Fixes: Unable to clear maxAge after it has been set. Related to #12462

## Changes
- Injected Request object to check if maxAge was explicitly provided via array_key_exists
- When maxAge is explicitly sent (even as null), use the provided value
- When omitted, keep the existing stored value

## Root Cause
The null-coalescing merge ( ?? ) made it impossible to clear maxAge once set, because both "not provided" and "explicitly null" resolved to PHP null. Added Request injection to distinguish the two cases via array_key_exists.

### DIFF
--- a/src/Appwrite/Auth/OAuth2/Oidc.php
+++ b/src/Appwrite/Auth/OAuth2/Oidc.php
@@ -43,13 +43,25 @@ class Oidc extends OAuth2
      */
     public function getLoginURL(): string
     {
-        return $this->getAuthorizationEndpoint() . '?' . \http_build_query([
+        $params = [
             'client_id' => $this->appID,
             'redirect_uri' => $this->callback,
             'state' => \json_encode($this->state),
             'scope' => \implode(' ', $this->getScopes()),
             'response_type' => 'code',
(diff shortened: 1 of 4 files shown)


deepshekhardas:
**Additional fix pushed**: addresses the prompt inconsistency and test gaps identified during review.

**Changes in latest commit:**
1. **prompt merge now uses \rray_key_exists\** (like maxAge) — sending explicit \
ull\ clears the field; omitting preserves existing value.
2. **Cleanup fix**: Added \maxAge => null\ to both existing cleanup blocks to prevent stale state leaking into subsequent tests.
3. **New tests**: \	estUpdateOAuth2OidcMaxAgeClearViaNull\ and \	estUpdateOAuth2OidcPromptClearViaNull\ directly exercising the core fixes.

deepshekhardas:
This PR has been recreated from a clean \official/1.9.x\ branch at #12583 to resolve merge conflicts. Please review #12583 instead.

PR #12479, opened 2026-06-03
feat: orchestrator build backend with Deployments module

deepshekhardas:
## Changes

- Add Deployments module with Build/Action, Build/Update, Source/Get, Events/Create endpoints
- Update Builds worker for orchestrator-driven deployment pipelines
- Add build artifacts action for deployment tracking
- Register new HTTP routes in Deployments/Services/Http.php

Closes #12441

PR #12489, opened 2026-06-04
fix(builds): resolve 4 active defects in orchestrator build backend

deepshekhardas:
Fixes 4 active defects in the orchestrator build backend. Related to #12479

### 1. Premature event emission (P1)
Functions and Sites deployment Create endpoints were setting queueForEvents params on every chunk upload (because !->isEmpty() is always true after first chunk). Now guarded with  ===  check so events only fire when all chunks are complete.

Files: Functions/Http/Deployments/Create.php, Sites/Http/Deployments/Create.php

### 2. Missing Authorization::skip() wrappers
New orchestrator methods (applyOrchestratorEvent, waitForOrchestratorBuildPath, waitForOrchestratorSiteDetectionLogs) were fetching from the database without getAuthorization()->skip() wrappers, which could silently return empty documents under project authorization context.

Files: Functions/Workers/Builds.php

### 3. Security: callback secret conflation
Both Builds.php::createOrchestratorBuild and Deployments/Events/Create.php fell back to _APP_OPENSSL_KEY_V1 when _APP_ORCHESTRATOR_CALLBACK_SECRET was not set. If the callback secret is exposed, the attacker gains the primary encryption key. Now requires explicit _APP_ORCHESTRATOR_CALLBACK_SECRET.

Files: Functions/Workers/Builds.php, Deployments/Http/Deployments/Events/Create.php

### 4. Null coalesce regression in outputDirectory
The ?? '' guard was dropped when  was hoisted out of the closure, passing null to the executor when neither buildOutput nor outputDirectory was configured. Restored the default.

Files: Functions/Workers/Builds.php

### DIFF
--- a/app/init/constants.php
+++ b/app/init/constants.php
@@ -195,6 +195,7 @@ const DATABASE_TYPE_DELETE_DATABASE = 'deleteDatabase';
 // Build Worker Types
 const BUILD_TYPE_DEPLOYMENT = 'deployment';
 const BUILD_TYPE_RETRY = 'retry';
+const BUILD_TYPE_ORCHESTRATOR_EVENT = 'orchestratorEvent';
 
 // Deletion Types
 
@@ -504,6 +505,7 @@ const ADVISOR_REPORT_TYPES = [
 
 // Resource types for Tokens
 const TOKENS_RESOURCE_TYPE_FILES = 'files';
+const TOKENS_RESOURCE_TYPE_DEPLOYMENT_ARTIFACTS = 'deploymentArtifacts';
 const TOKENS_RESOURCE_TYPE_SITES = 'sites';
 const TOKENS_RESOURCE_TYPE_FUNCTIONS = 'functions';
 const TOKENS_RESOURCE_TYPE_DATABASES = 'databases';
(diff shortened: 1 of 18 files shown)


deepshekhardas:
**Additional fix pushed**: resolves the stuck-deployment scenario identified during review.

When \exitCode === 0\ but \uildPath\ was still empty after polling, \pendingExit\ was written to cache with no recovery path — if the artifact upload permanently failed, the deployment stayed stuck in \uilding\ until the 1-hour cache TTL expired.

**Changes in latest commit:**
1. **Fail deterministically** instead of stashing \pendingExit\ — the deployment is marked as failed when \uildPath\ is still empty after polling.
2. **Increased polling retries** 20 → 60 (5s → 15s) to reduce false positives.
3. **Coroutine-safe polling** — \\\usleep()\ → \Co::sleep()\ in both polling methods.

deepshekhardas:
Replaced by #12577 (clean branch from latest 1.9.x + #12479 + defect fixes)

Meldiron:
Closing — abandoned by the author and superseded.

Your own comment on this PR says "Replaced by #12577", and #12577 is now closed too. The branch also carries three stray files that shouldn't be in a PR: `pr-body-12462.txt`, `pr-body-12476.txt`, `pr-body-12479.txt`. The feature branch it builds on, #12479, is closed as well.

For the record, none of the orchestrator work is on `main` — `src/Appwrite/Platform/Modules/Deployments/` doesn't exist there — so if this is still wanted it needs a fresh branch cut from `main`, not a revival of this one.

PR #12505, opened 2026-06-06
fix: validate session expiry during JWT authentication

tysoncung:
Updates JWT session validation to check session expiry in addition to session existence.

The previous PR (#11606) had merge conflicts due to extensive codebase refactoring (app/init/resources.php was split into app/init/resources/request.php). This PR applies the same fix to the current codebase structure.

Closes #11606


tysoncung:
Applied the greptile suggestion to fix the session expiry comparison. Removed the unnecessary `new \DateTime()` round-trip — `` is already a string, and this aligns with the established pattern at lines 487 and 964.

PR #12507, opened 2026-06-06
Fix users create endpoints accepting null name

avisii1680769-cell:
## What does this PR do?

Allows the optional `name` parameter to be explicitly sent as `null` for Users create endpoints.

When `name` is `null`, `createUser()` now normalizes it to an empty string before user creation so downstream logic continues to receive a string value. The same nullable validation is applied to the hashed-password create endpoints: bcrypt, MD5, Argon2, SHA, PHPass, Scrypt, and Scrypt Modified.

Fixes https://github.com/appwrite/appwrite/issues/8785

## Test Plan

- `php -l app/controllers/api/users.php`
- `php -d memory_limit=2048M ./vendor/laravel/pint/builds/pint --test --config pint.json app/controllers/api/users.php`
- Verified `Nullable(new Text(128))` accepts `null` and valid names while rejecting names over 128 chars.
- Verified all 8 Users create endpoints accept nullable `name`, while the update-name endpoint remains strict.

Full `composer test` was attempted locally, but this Windows environment is missing required PHP extensions (`redis`, `swoole`, `yaml`, `imagick`, `maxminddb`) and the E2E `appwrite.test` service.


HarshMN2345:

PR #12523, opened 2026-06-08
chore: migrate to utopia-php/http safe-wildcards

deepshekhardas:
Updates HTTP route patterns to use utopia-php/http safe-wildcards for improved route matching safety.


Meldiron:
Closing — the branch does not parse.

```
$ php -l app/controllers/shared/api.php
PHP Parse error:  Unclosed '{' on line 801 in app/controllers/shared/api.php on line 1084
```

The `Http::shutdown()` handler for the `api` group lost its closing `});` along with the webhook trigger block, so the next `Http::shutdown()` is parsed inside the previous closure. Separately, the `->action(...)` signature declares `User $user`, `AuditContext $auditContext`, `Audit $publisherForAudits`, `Context $usage`, `UsagePublisher $publisherForUsage`, `Authorization $authorization` and `callable $timelimit`, none of which appear in the `->inject()` list above it — that would fatal at runtime even if it parsed.

Also pins `utopia-php/http` to `dev-feat-safe-wildcards as 2.0.x-dev` and carries unrelated `.claude/skills/` typo fixes and a `.github/workflows/specs.yml` change.

`safe-wildcards` genuinely isn't on `main` yet, so the underlying migration is still worth doing — but from a fresh branch.

PR #12529, opened 2026-06-08
fix: attribute default value persistence bug

deepshekhardas:
Fixes bug where default values on attributes were not being persisted correctly on document creation.


deepshekhardas:
Closing this messy PR. Recreated as a clean PR with only the relevant fix (purge collection cache after attribute update).

PR #12537, opened 2026-06-08
Fix execution ownership validation in delete execution endpoint

### LINKED ISSUE #12517: Function execution can be deleted through the wrong function ID
## Steps to Reproduce

1. Start Appwrite locally from the source repository using Docker Compose.
2. Create a project and an API key with function/execution read and write scopes.
3. Create two functions in the same project:
   - `bug-func-a`
   - `bug-func-b`
4. Deploy code to `bug-func-b` and wait until the deployment status becomes `ready`.
5. Create an execution for `bug-func-b` and copy the execution ID.
6. Confirm that reading the execution through the wrong function returns `404`:

```http
GET /v1/functions/bug-func-a/executions/{executionFromFunctionBId}
```

7. Now try deleting the same execution through the wrong function:

```http
DELETE /v1/functions/bug-func-a/executions/{executionFromFunctionBId}
```

8. The request returns `204` and deletes the execution, even though the execution belongs to `bug-func-b`.

---

## Observation

I reproduced this locally using direct REST API calls.

The possible cause seems to be the ownership check in:

```txt
src/Appwrite/Platform/Modules/Functions/Http/Executions/Delete.php
```

Current condition:

```php
if (
    $execution->getAttribute('resourceType') !== 'functions' &&
    $execution->getAttribute('resourceInternalId') !== $function->getSequence()
) {
}
```

Because the condition uses `&&`, an execution belonging to another function still passes the check when its `resourceType` is `functions`.

This likely should use `||` instead so that either mismatch correctly throws `EXECUTION_NOT_FOUND`.

### 👍 Expected behavior

It should return `404 execution_not_found` when the execution ID does not belong to the function ID in the route.

For example:

```http
DELETE /v1/functions/bug-func-a/executions/{executionFromFunctionBId}
```

should not delete an execution that belongs to `bug-func-b`.

### 👎 Actual Behavior

It actually returns `204 No Content` and deletes the execution from `bug-func-b`.

  This is inconsistent with the GET execution endpoint, which correctly returns 404 when using
  the wrong function ID.

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

MacOS

### 🧱 Your Environment

Self-hosted Appwrite running locally with Docker Compose from the source repository.

  Endpoint used: `http://localhost/v1`

  No SDK was used. I reproduced the issue using direct REST API calls with `curl`.

kamleshkumaryadav5951-pixel:
## What does this PR do?

This PR fixes an ownership validation issue in the function execution delete endpoint.

The validation condition used `&&`, which allowed an execution belonging to a different function to pass the ownership check when the resource type was `functions`.

The condition has been updated to use `||` so that any mismatch correctly throws `EXECUTION_NOT_FOUND`.

## Test Plan

- Located the ownership validation check in `src/Appwrite/Platform/Modules/Functions/Http/Executions/Delete.php`.
- Replaced `&&` with `||` in the validation condition.
- Verified the change using `git diff`.
- Confirmed that executions belonging to a different function will now fail the ownership validation check and return `EXECUTION_NOT_FOUND`.

## Related PRs and Issues

Fixes #12517

## Checklist

- [x] Have you read the Contributing Guidelines on issues?
- [x] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs? (Not applicable)

### DIFF
--- a/src/Appwrite/Platform/Modules/Functions/Http/Executions/Delete.php
+++ b/src/Appwrite/Platform/Modules/Functions/Http/Executions/Delete.php
@@ -86,7 +86,7 @@ class Delete extends Base
             throw new Exception(Exception::EXECUTION_NOT_FOUND);
         }
 
-        if ($execution->getAttribute('resourceType') !== 'functions' && $execution->getAttribute('resourceInternalId') !== $function->getSequence()) {
+        if ($execution->getAttribute('resourceType') !== 'functions' || $execution->getAttribute('resourceInternalId') !== $function->getSequence()) {
             throw new Exception(Exception::EXECUTION_NOT_FOUND);
         }
         $status = $execution->getAttribute('status');


eldadfux:

PR #12543, opened 2026-06-09
Fix flaky realtime tests

deepshekhardas:
## What does this PR do?

Fixes flaky E2E realtime tests by addressing two issues:

1. **WebSocket timeouts**: Added timeout parameter to all \getWebsocket()\ calls in both \RealtimeConsoleClientTest\ and \RealtimeCustomClientTest\ to prevent tests from hanging indefinitely when the WebSocket connection fails to establish.

2. **Session re-authentication**: In \	estChannelAccount\, the test performs a password recovery flow which invalidates the current session. Added explicit re-authentication to ensure subsequent WebSocket connections use a valid auth token.

## Related issues
Fixes #12540

### DIFF
--- a/tests/e2e/Services/Realtime/RealtimeConsoleClientTest.php
+++ b/tests/e2e/Services/Realtime/RealtimeConsoleClientTest.php
@@ -611,7 +611,7 @@ final class RealtimeConsoleClientTest extends Scope
         $client = $this->getWebsocket(['console'], [
             'origin' => 'http://localhost',
             'cookie' => 'a_session_console=' . $this->getRoot()['session'],
-        ], $projectId);
+        ], $projectId, null, 10);
 
         $response = json_decode($client->receive(), true);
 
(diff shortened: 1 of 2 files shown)


Meldiron:
Closing — everything constructive here is already on `main`, and what's left is a regression.

#12540 merged (to `1.9.x`) and the same fixes reached `main` independently. Already present on `main`:
- the `getWebsocket(..., null, 10)` / `null, 30` timeout arguments in both `RealtimeConsoleClientTest` and `RealtimeCustomClientTest`
- the session-refresh block in `testChannelAccount` (`RealtimeCustomClientTest.php:834-852`)

I merged `origin/main` into this branch. It merged **without conflicts**, but the result is broken. The one surviving hunk is a 104-line deletion inside `testChannelDatabaseTransaction` that strips the database/collection/attribute setup and the transaction create + commit. `main` has since grown that test further, so after the merge the remaining body still references `$databaseId`, `$collectionId`, `$documentId` and `$transactionId` — all of which were defined in the deleted block. Git had no textual conflict to report; the breakage is semantic.

Net effect of merging this as-is would be a test that fatals on undefined variables. Not worth salvaging when the rest is already upstream.

PR #12561, opened 2026-06-10
Resolve locks through lock pool

premtsd-code:
## What changed
- Resolve the request-scoped Lock service through the existing `lock` pool instead of the legacy shared `redis` resource.
- Pass Lock a lock-provider callback so the pooled Redis connection stays borrowed while the distributed lock is used.
- Keep existing Redis resource and pool registration behavior unchanged.
- Convert LockTest to a true unit test with an in-memory `Utopia\Lock\Lock` fake instead of connecting to Redis.

## Why
Cloud already configures its `lock` pool through `_APP_CONNECTIONS_LOCKS`. Appwrite CE only needs to consume the existing lock pool; it should not add new env behavior or make unit tests depend on external Redis.

This follows the existing Cloud `main` pattern where a dedicated `lock` pool is registered from `_APP_CONNECTIONS_LOCKS` and consumers use `$pools->get('lock')` rather than creating ad-hoc Redis clients:
https://github.com/appwrite-labs/cloud/blob/f2f3661fde840d0476d2bb945c9f6ac373499804/app/init/registers.php#L161-L162

## Testing
- composer lint
- vendor/bin/phpunit --no-configuration --bootstrap vendor/autoload.php tests/unit/Locking/LockTest.php
- php -l app/init/resources/request.php
- php -l src/Appwrite/Locking/Lock.php
- php -l tests/unit/Locking/LockTest.php
- git diff --check

### DIFF
--- a/app/controllers/general.php
+++ b/app/controllers/general.php
@@ -12,7 +12,7 @@ use Appwrite\Event\Message\Delete as DeleteMessage;
 use Appwrite\Event\Publisher\Certificate;
 use Appwrite\Event\Publisher\Delete as DeletePublisher;
 use Appwrite\Extend\Exception as AppwriteException;
-use Appwrite\Locking\Lock;
+use Appwrite\Locking\PlatformLock;
 use Appwrite\Network\Cors;
 use Appwrite\Platform\Appwrite;
 use Appwrite\SDK\Method;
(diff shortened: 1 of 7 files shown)


deepshekhardas:
Replaced by #12581 (clean rebase on latest 1.9.x — restores the reverted Lock.php + applies both of @premtsd-code's improvements)

PR #12580, opened 2026-06-12
feat: split password personal data policy into granular field toggles

deepshekhardas:
Split password personal data policy into granular field toggles for individual password fields (hash, algorithm, salt, saltSeparator, signerKey).

Clean rebase of #12560 by @HarshMN2345 on latest 1.9.x.


Meldiron:
Closing in favour of #12560, which is the original by @HarshMN2345 and already targets `main`.

The feature itself is still unmerged — `main`'s `PolicyPasswordPersonalData` model only has the `enabled` rule, so the granular field toggles aren't there. It's just that #12560 is the better base to land it from: same work, original author, already on `main` rather than `1.9.x`.

PR #12583, opened 2026-06-12
fix(oidc): allow clearing maxAge via explicit null in OIDC update

deepshekhardas:
Fixes: Unable to clear maxAge after it has been set. Related to #12462

## Changes
- Injected Request object to check if maxAge was explicitly provided via array_key_exists
- When maxAge is explicitly sent (even as null), use the provided value
- When omitted, keep the existing stored value
- Added prompt parameter support for OpenID Connect

## Root Cause
The null-coalescing merge ( ?? ) made it impossible to clear maxAge once set, because both "not provided" and "explicitly null" resolved to PHP null. Added Request injection to distinguish the two cases via array_key_exists.

Replaces PR #12487.

### DIFF
--- a/src/Appwrite/Auth/OAuth2/Oidc.php
+++ b/src/Appwrite/Auth/OAuth2/Oidc.php
@@ -43,13 +43,25 @@ class Oidc extends OAuth2
      */
     public function getLoginURL(): string
     {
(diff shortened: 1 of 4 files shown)


deepshekhardas:
Closing this PR as it is superseded by replacement PRs created in the Jul 23 batch. See #12947-#12957 for the respective clean replacements.

PR #12596, opened 2026-06-14
fix: allow clearing OIDC maxAge via explicit null in update

deepshekhardas:
Fixes PR #12462 by Meldiron.

Bug: OIDC maxAge could not be cleared once set because PHP's null-coalescing operator (\??\) cannot distinguish between 'parameter not sent' and 'parameter explicitly set to null'.

Fix: Use \rray_key_exists()\ to check whether the parameter was actually provided in the request body before applying the merge logic. This also applies the same fix to the \prompt\ field for consistency.

Changes:
- Inject Request into the OIDC update handler
- Fix maxAge and prompt merge logic to allow explicit null clearing
- Add tests for maxAge/prompt clear via null, partial preservation, and validation

### DIFF
--- a/src/Appwrite/Auth/OAuth2/Oidc.php
+++ b/src/Appwrite/Auth/OAuth2/Oidc.php
@@ -43,13 +43,25 @@ class Oidc extends OAuth2
      */
     public function getLoginURL(): string
     {
-        return $this->getAuthorizationEndpoint() . '?' . \http_build_query([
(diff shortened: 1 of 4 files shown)


deepshekhardas:
Closing this PR as it is superseded by replacement PRs created in the Jul 23 batch. See #12947-#12957 for the respective clean replacements.

PR #12597, opened 2026-06-14
fix: allow null name parameter in user creation endpoints

Shaan-alpha:
Resolves #8785. Updated API controllers for Users.CreateUser... and Account.Create to accept 
ull for the 
ame parameter using 
ew Nullable(new Text(128)). This prevents vague server errors and gracefully defaults to an empty string internally.


Shaan-alpha:
Hello @stnguyen90. This pull request fixes the vague server error by allowing a null name parameter during user creation. The checks have passed. Please review this at your earliest convenience. Thank you.

Shaan-alpha:
Gentle follow-up, @stnguyen90 — this has been open for review for over a week now. The change is small and scoped (allowing a null `name` on the user-creation endpoints so the API returns a clear validation error instead of a vague server error), the branch is mergeable, and all checks are green. I'd appreciate a review whenever you have capacity, and I'm glad to make any adjustments you'd suggest. Thank you.

Shaan-alpha:
Hi @ChiragAgg5k, following up since you updated the base branch last month. The branch is mergeable, checks are green, and Greptile rated it safe to merge; it just needs a human review now. If you have a moment to look, or can point me to whoever currently owns the users API surface, that would be great. Thanks!

Shaan-alpha:
@Meldiron since you've been in the Users module this month — this one needed a rework after the module migration, so it's worth a fresh look.

`users.php` is gone, so the fix moved to the eight `Create` classes under `Modules/Users/Http/Users`. Most of the plumbing was already there: the action signatures are already `?string $name` and `Base::createUser()` already does `$name = $name ?? ''`, so only the validator was still rejecting null. Wrapping the name param in `Nullable` was all that was left — down to 9 files, +19/-10.

Shaan-alpha:
Merged main in after the conflict.

It was in `Users/Http/Users/Scrypt/Create.php`: main added `example:` named arguments to the password cost params, while this branch wrapped the name validator in `Nullable`. Kept both, so the examples stay and `name` stays nullable.

All eight user creation endpoints still carry `new Nullable(new Text(128))` on `name`, and `Name/Update.php` plus the Targets endpoints are deliberately untouched, since `name` is required there.

Mergeable again. @ChiragAgg5k worth a look if you have time.

HarshMN2345:

PR #12626, opened 2026-06-20
fix(databases): return 404 for a wrong-type database id on product-DB APIs

abnegate:
## Problem
On the product-DB APIs (`/v1/tablesdb`, `/v1/documentsdb`, `/v1/vectorsdb`), listing tables/collections resolves the database by id with only an `isEmpty()` check (`Collections/XList`):
```php
$database = $authorization->skip(fn () => $dbForProject->getDocument('databases', $databaseId));
if ($database->isEmpty()) { throw DATABASE_NOT_FOUND; }
```
A genuinely missing id correctly returns **404**. But a db that **exists with a different type** (e.g. a documentsdb id on the tablesdb path, or a compute-backed db) is *not* empty, so it proceeds to a type-mismatched backend operation that surfaces as an opaque **500 `general_unknown`** instead of a clean 404.

Reproduced on a live project:
- `GET /v1/tablesdb/{tablesdbId}/tables` → 200 ✅
- `GET /v1/tablesdb/{missingId}/tables` → 404 ✅
- `GET /v1/tablesdb/{nonTablesdbId}/tables` → **500** ❌ (should be 404)

## Fix
Add a migration-safe type guard on the Databases base `Action` and apply it where the table/collection list resolves the database:
```php
protected function isDatabaseTypeMismatch(Document $database): bool
{
    $expected = $this->getDatabaseType();
    if ($expected === DATABASE_TYPE_LEGACY) {
        return false; // legacy /v1/databases API is intentionally not type-guarded
    }
    return $database->getAttribute('type', '') !== $expected;
}
```
- Product paths: a db whose `type` ≠ the path's type → treated as not-found (404), same as a missing id.
- **Legacy `/v1/databases` is never type-guarded**, so databases predating the `type` attribute (empty `type`) are unaffected — no behavior change on the legacy API.

## Scope
This targets the **table/collection list** (`Collections/XList`), which is the shared base behind tablesdb tables + documentsdb/vectorsdb collections lists — i.e. the reproduced symptom, fixed for all three in one place. Other db-resolution sites in the module (db-level Get/Update/Delete, sub-resources) have the same `getDocument+isEmpty`-only pattern; note that several extend `Utopia\Platform\Action` rather than the type-aware `Databases\Action`, so guarding them needs per-handler work — tracked as a follow-up rather than bundled here.

## Test plan
- New unit test `tests/unit/Platform/Modules/Databases/DatabaseTypeMismatchTest.php` (3 tests, 12 assertions): product paths reject mismatched/missing type, documents/vectors scope to their own type, legacy path is never guarded.
- Pint + PHPStan (level 4) on changed files: clean.

Context: Appwrite Cloud dedicated-DB testing (DAT-1628).

### DIFF
--- a/src/Appwrite/Platform/Modules/Databases/Http/Databases/Action.php
+++ b/src/Appwrite/Platform/Modules/Databases/Http/Databases/Action.php
@@ -21,6 +21,41 @@ class Action extends AppwriteAction
         return $this->context;
     }
 
+    /**
+     * Database types this path may operate on. Single source of truth shared by
+     * the list filter ({@see getDatabaseTypeQueryFilters()}) and the by-id guard
+     * ({@see isDatabaseTypeMismatch()}) so they cannot diverge.
+     *
+     * TablesDB is the compatibility successor to the legacy databases API, so it
+     * also serves legacy-typed databases; the other product APIs are scoped to
+     * their own type. Every database carries a non-null `type` (set on create
+     * and backfilled to 'legacy' by migration V23), so null is not represented.
+     *
+     * @return string[]
+     */
+    protected function getAllowedDatabaseTypes(): array
+    {
+        return match ($this->getDatabaseType()) {
+            DATABASE_TYPE_TABLESDB => [DATABASE_TYPE_TABLESDB, DATABASE_TYPE_LEGACY],
+            default => [$this->getDatabaseType()],
+        };
+    }
+
+    /**
+     * A database resolved by id on a product-DB path must be one of the path's
+     * allowed types; otherwise it is treated as not-found rather than proceeding
+     * to a type-mismatched backend operation that surfaces as an opaque 500. The
+     * legacy /v1/databases API is intentionally not type-guarded.
+     */
+    protected function isDatabaseTypeMismatch(Document $database): bool
+    {
+        if ($this->getDatabaseType() === DATABASE_TYPE_LEGACY) {
+            return false;
+        }
+
+        return !in_array($database->getAttribute('type', ''), $this->getAllowedDatabaseTypes(), true);
+    }
+
     public function setHttpPath(string $path): self
     {
         if (\str_contains($path, '/tablesdb')) {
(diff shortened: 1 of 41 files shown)


abnegate:
Addressed the P1 (TablesDB compatibility) and centralised the rule.

**Root cause of the P1:** the guard and the `listTablesDatabases` filter encoded the allowed types in two separate places, and my first pass made the guard stricter than the list (it rejected `legacy`/null-typed dbs on the tablesdb path). So an id returned by `tablesDB.list` could `DATABASE_NOT_FOUND` on `tablesDB.listTables` — exactly the inconsistency flagged.

**Fix — single source of truth:** `Action::getAllowedDatabaseTypes()` now feeds **both** the list filter (`getDatabaseTypeQueryFilters`) and the by-id guard (`isDatabaseTypeMismatch`), so they can't diverge:
- `tablesdb` → `{tablesdb, legacy, ''}` (mirrors the list, including the `Query::isNull('type')` compat for pre-`type` databases)
- `documentsdb` / `vectorsdb` → their own type only
- legacy `/v1/databases` → not type-guarded
- Removed the now-redundant `TablesDB/XList` filter override so the rule lives in one place. The unit test asserts the guard's allowed set equals what the list advertises.

**Scope:** applied the guard at the table/collection + document/row read resolution (`Collections` + `Documents` list/get) — the reproduced surface, now consistent with the list. The same one-line guard should be adopted at the remaining write/sub-resource resolution sites; note several db-level handlers (`Databases/Get`, `Update`, `Delete`) extend `Utopia\Platform\Action` rather than the type-aware `Databases\Action`, so they'd need to be made type-aware first — tracking that as a follow-up rather than bundling a broad, base-class-changing sweep here.

abnegate:
Follow-up now included — the deferred scope is done.

The guard is applied at **every** database-resolution site in the module (39 total), and the base classes are fixed so they can use it:
- Re-based the framework-`Action` handlers onto the type-aware `Databases\Action`: `Databases/Get`, `Update`, `Delete`, and the `Databases`/`TablesDB` log lists.
- Re-based the `Collections/Attributes/Action` and `Collections/Indexes/Action` abstract bases and `Usage/Get` onto it as well — with covariant `setHttpPath` return types. Their private resource-`$context` and the base's db-type `$context` are private-isolated, so there's no clash.

Now the rule + guard are genuinely centralised: one `getAllowedDatabaseTypes()` source feeds the list filter and the by-id guard, and the guard is reachable from every handler in the module.

Gates: Pint ✓, PHPStan level 4 ✓, unit test ✓ (41 files changed). The only PHPStan finding is `Init/Timeout.php:28` (`Request::getHeaderLine`) — pre-existing and unrelated (reproduces on the untouched file; it's a subset-analysis artifact, not introduced here).

PR #12628, opened 2026-06-20
fix: retry SMTP send once on 421 timeout (stale keepalive)

### LINKED ISSUE #12623: 🐛 SMTP keepalive causes 421 timeout on first send after idle; subsequent sends work (1.9.0 self-hosted)

## 🐛 Describe the bug

## 👟 Reproduction steps

5. Trigger another send.

## 🔬 What we've ruled out

## 🧪 Workaround we use

## 🎲 Appwrite version

## 💻 Operating system

### 👍 Expected behavior

### 👎 Actual Behavior

```
[Error] Type: Exception
 SMTP code: 421
[Error] Line: 191
```

```
```

### 🎲 Appwrite version

### 💻 Operating system

Linux

### 🧱 Your Environment

abhijnyan-codes:
## What does this PR do?

Fixes intermittent SMTP 421 timeout failures on the first mail send after an idle period when `keepAlive` is enabled.

When `worker-mails` reuses a kept-alive SMTP connection that has been silently dropped by the upstream server or an intermediate NAT, PHPMailer attempts `MAIL FROM` on the stale socket and receives a `421 — Timeout, closing connection` response. The first failed attempt closes the socket client-side, but no retry occurs, so the end-user never receives the email (OTP, magic link, verification, etc.).

This PR adds a single transparent retry when a `421` error is detected on an SMTP send. The failed send already closes the stale connection; the retry opens a fresh connection and succeeds. No behaviour change for non-421 errors or cloud SMTP path.

**Implementation note:** PHPMailer's `$mail->send()` returns `false` on a 421 timeout rather than throwing an exception — the failure is surfaced via the response results array. The fix inspects `$response['results']` after each send, detects 421 in the error string, and retries once. `mail.status` is only marked `success` after confirming no failed recipients in the final response.

## Test Plan

**Manual reproduction:**
1. Self-host Appwrite 1.9.x with a custom SMTP provider (e.g. Resend).
2. Trigger a mail send (OTP / password recovery).
3. Wait for the SMTP idle timeout (a few minutes — varies by provider/NAT).
4. Trigger another send.

**Before:** first send after idle fails with:

MAIL FROM command failed, Timeout - closing connection, 421

**After:** retry succeeds transparently; user receives the email on first attempt.

**Unit test approach:**
- Mock `$adapter->send()` to return a failed result with a `421` error string on the first call, success on the second — assert no exception thrown and `mail.status` is `success`.
- Mock both calls returning `421` failure — assert exception thrown with retry error message.
- Mock non-421 failure — assert original error behaviour unchanged.
- Mock success on first call — assert no retry attempted.

## Related PRs and Issues

- Fixes #12623

## Checklist
- [x] Have you read the [[Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [ ] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs? — N/A, no API change.


abhijnyan-codes:
Thanks for the detailed review, @greptile-apps. Before pushing another iteration, I'd like to confirm the right approach on three points to avoid going in circles:

**1. Cloud path — registry vs fresh SMTP**
The original code used `$register->get('smtp')` for the cloud path. I switched to `new SMTP()` directly from env vars to avoid reusing the cached singleton's stale socket on retry. But you're right that this bypasses any decorated/mock adapter the registry may provide in tests or other deployments.

Should I preserve `$register->get('smtp')` for the first attempt and only build a fresh `new SMTP()` for the retry? Or is there a preferred way to force a fresh connection through the registry?

**2. 421 match pattern**
`str_contains($error->getMessage(), '421')` is too broad. What's the recommended pattern? Options I see:
- Match on `SMTP code: 421` (matches the exact PHPMailer error format from the bug report)
- Match on `str_contains($msg, '421') && str_contains($msg, 'Timeout')` 
- Something else?

**3. keepAlive on fresh adapter**
If we build a new adapter per `action()` call, `keepAlive: true` has no benefit since the instance is discarded after the call. Should this be `keepAlive: false` for fresh-per-call adapters? Or does the worker process reuse the same instance across multiple queue messages (in which case keepAlive makes sense but stale socket handling is needed)?

Happy to implement once I understand the intended behavior. Thanks!

abhijnyan-codes:
Two follow-up questions after the latest [greptile-apps](https://github.com/greptile-apps). pass:
1. Timeout string matching

str_contains($msg, 'Timeout') is case-sensitive. PHPMailer may emit timeout, TIMEOUT, or timed out depending on version. Should I use stripos($msg, 'timeout') !== false for case-insensitive matching, or is there a canonical PHPMailer error string we can anchor to?
2. Cloud retry adapter

Greptile flags that the cloud retry path builds new SMTP() directly from env vars, bypassing any decorator/wrapper the registry may apply to the smtp adapter. But a fresh registry instance would return the same stale singleton.
What's the recommended resolution here — should the retry for the cloud path use $register->get('smtp') and accept that it may reuse the stale socket, or is there a way to reset/reconnect the registered adapter?

eldadfux:

PR #12632, opened 2026-06-21
fix: allow null name in create user endpoints

yadavlokesh1:
fix: allow null name in create user endpoints

## What does this PR do?

the `name` field on user-creation endpoints returns a vague server error when explicitly passed as `null`, even though it's optional. The `email` and `phone` fields already handle this correctly by wrapping their validators in `Nullable(...)`. This applies the same pattern to `name` across all 8 create-user endpoints in this file.

## Test Plan

Send a POST to /v1/users with `name: null` - should succeed instead of erroring, same as omitting the field entirely.

## Related PRs and Issues

- (Related PR or issue)

## Checklist

- [ x ] Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?


HarshMN2345:

PR #12640, opened 2026-06-21
fix deleteRule error handling

varshachauhan2323:
## What does this PR do?

This PR adds exception handling to the `deleteRule()` method.

Previously, failures during certificate deletion could interrupt the cleanup process. This change wraps certificate deletion and certificate document deletion in separate `try/catch` blocks and logs any errors using `Console::error()`, allowing the remaining cleanup operations to continue gracefully.

## Test Plan

1. Updated the `deleteRule()` method in `Deletes.php`.
2. Verified that the code builds successfully.
3. Confirmed that certificate deletion failures are now logged instead of stopping execution.
4. Confirmed that certificate document deletion failures are also handled gracefully.

## Related PRs and Issues

* None

## Checklist

* [x] Have you read the Contributing Guidelines on issues?
* [x] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

### DIFF
--- a/src/Appwrite/Platform/Workers/Deletes.php
+++ b/src/Appwrite/Platform/Workers/Deletes.php
@@ -1672,20 +1672,45 @@ class Deletes extends Action
     }
 
     /**
-     * @param Database $dbForPlatform
-     * @param Document $document rule document
-     * @return void
-     */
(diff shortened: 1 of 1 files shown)


varshachauhan2323:
Thanks for the review! I'll address these cases and update the PR.

PR #12644, opened 2026-06-22
refactor:store impersonator context as independent DI resource

HarshMN2345:
## Summary

- When an impersonated request confirms a team membership, the `$user` Document has transient in-memory attributes set (`impersonatorUserId`, `impersonatorUserInternalId`, etc.) that are not real DB columns
- The membership status update was passing the full `$user` object to `updateDocument`, which on dedicated (non-shared-table) databases causes a failure because those attributes don't exist in the schema
- Fixed by using a sparse `new Document(['emailVerification' => true])` for the DB write instead of the full object

## Test plan

- [ ] Confirm a team membership invite while impersonating a user on a dedicated database
- [ ] Verify no errors thrown and membership is confirmed correctly

Fixes https://appwrite.sentry.io/issues/7353240003/


app/init/realtime/connection.php
```diff
@@ -209,7 +209,9 @@
         return new Origin(\array_unique($allowedHostnames), \array_unique($allowedSchemes));
     }, ['platform', 'request', 'project', 'servers', 'authorization']);
 
-    $container->set('user', function (Request $request, Document $project, Document $console, Authorization $authorization) use ($getMode, $getDbForPlatform, $getDbForProject) {
+    $container->set('impersonatorUser', fn () => new Document(), []);
+
+    $container->set('user', function (Request $request, Document $project, Document $console, Authorization $authorization, Document $impersonatorUser) use ($getMode, $getDbForPlatform, $getDbForProject) {
         $mode = $getMode($request, $project);
```
abnegate: Can we reverse the relationship and have `impersonatorUser` inject `user` to set it's own attributes?

HarshMN2345: The dependency needs to flow user → impersonatorUser rather than the reverse. The impersonatorUser is populated inside the user factory because we need the authenticated user first (to check the impersonator flag and resolve the target), and user must return the target user document. If we flipped it, impersonatorUser would resolve after user is already set 🤔

HarshMN2345:
@greptile only routes that need to act as the impersonated user inject targetUser. Account management routes (sessions, prefs, password) intentionally stay on the actor. This is by design, not a bug.

HarshMN2345:
@greptile The targetUser factory only runs when impersonatorUser is non-empty. impersonatorUser already does the DB lookup and returns an empty Document if the target doesn't exist so by the time targetUser executes, the target is guaranteed to exist. No null guard needed.

HarshMN2345:
@greptile re-review

HarshMN2345:
@greptile MFA is session-bound. The session belongs to the impersonating actor, not the target. The actor already completed their own MFA, checking the target's MFA requirements against the actor's session factors is incorrect. This is intentional.

HarshMN2345:
@greptile
1. Legacy audit fallback — deserialization already has ?? [] which creates an empty Document for old messages, isEmpty() returns true, and the fallback to $user fields is used automatically. No explicit fallback needed.

2. Admin type in audit — impersonatorUser sparse payload already includes 'type' => $user->getAttribute('type', ACTOR_TYPE_USER) where $user is the actor. Admin actors get their correct type from their own document.

3. targetUser null from findOne — targetUser only runs when impersonatorUser is non-empty. impersonatorUser already does the same DB lookup and returns an empty Document if the target doesn't exist. The target is guaranteed to exist by the time targetUser executes.

4. Abuse keys charged to actor — intentional. The actor (admin) is making the requests and should be rate-limited, not the target.

5. Realtime stale impersonatedUserId — impersonation is established at connection time via headers. The authentication message handler restores the same connection-level impersonatedUserId, which is correct since impersonation context is tied to the connection, not the session.

HarshMN2345:
@greptile re-review

HarshMN2345:
@greptile 1. null guard on findOne — targetUser only resolves when impersonatorUser is non-empty. impersonatorUser already performs the same DB lookup and returns an empty Document if the target doesn't exist. A race where the user is deleted between the two lookups is not a realistic production concern and would affect far more than just this path.

2. Legacy audit fallback — deserialization already has ?? [] which produces an empty Document for old messages. isEmpty() returns true and the code falls back to $user fields. This is handled.

3. MFA session vs target — the session belongs to the impersonating actor. Checking target MFA requirements against the actor's session factors is architecturally incorrect. The actor has already authenticated with their own MFA. This is intentional

HarshMN2345:
@greptile re-review

abnegate (approved):

PR #12648, opened 2026-06-22
fix(builds): mask internal errors from user-visible build logs

deepshekhardas:
Introduces \BuildException\ — a user-facing exception type whose message is safe to surface in build logs. All intentional build failures now throw \BuildException\, while internal errors (database, executor, etc.) are masked behind a generic message.

Closes #12613


deepshekhardas:
Closing this PR — targeting \main\ instead of \1.9.x\. Recreated as #... (will update after creation).

PR #12649, opened 2026-06-22
fix(mails): retry SMTP send once on 421 timeout (stale keepalive)

deepshekhardas:
Detect stale keepalive connections by checking for 421 + Timeout in the error message. On detection, create a fresh SMTP adapter with \keepAlive: false\ and retry once. Also adds a \sendAndCheck\ helper that inspects response results for silent failures.

Closes #12628


deepshekhardas:
Closing — targeting \main\ instead of \1.9.x\. Recreated as #... (will update after creation).

PR #12650, opened 2026-06-22
fix(users): rebuild search index on partial user updates

deepshekhardas:
When updating user attributes (name, email, phone, labels), include \search\ in the update document to trigger the \userSearch\ filter, which rebuilds the search index from current values. Without this, the search field becomes stale after partial updates.

Also removes unnecessary \\->setAttribute()\ calls — the value is passed directly to the update Document instead.

Closes #12633


deepshekhardas:
Closing — targeting \main\ instead of \1.9.x\. Recreated as #... (will update after creation).

PR #12653, opened 2026-06-23
feat(messaging): attach project ID to internal SMS for provider attribution

ChiragAgg5k:
## What

Attaches the originating **project ID** to internal (auth) SMS as the MSG91 `UUID` tracking field, so the SMS provider's delivery logs and webhooks can be reconciled back to a project.

```php
$sms->setMetadata([MetadataParameter::UUID->value => $project->getId()]);
```

Applied at the **internal** send path in `Workers/Messaging.php` (`MESSAGE_SEND_TYPE_INTERNAL`) — i.e. phone OTP, MFA challenges, and team invites sent through Appwrite's own provider. The external/customer messaging path (`buildSmsMessage`, customer's own provider) is intentionally left unchanged.

## Why

Provider-side reports (e.g. MSG91 analytics exports) currently expose `UUID`/`CRQID` columns but we never populate them, so there's no way to trace a sent SMS back to the project that triggered it. This blocks per-project reconciliation of SMS spend vs. billing. The project ID (≤36 chars, `[A-Za-z0-9._-]`) fits MSG91's `UUID` constraint (≤80 chars, `[A-Za-z0-9_.-]`).

## Dependency

Bumps `utopia-php/messaging` `1.1.0` → `1.2.0` (adds `SMS::setMetadata()` and the `Msg91\MetadataParameter` enum via utopia-php/messaging#135).

## Notes

- Metadata is ignored by adapters that don't consume it (Twilio, Fast2SMS, Inforu), so it's safe for all internal providers and routes through GEOSMS.
- Targets `1.9.x` so Appwrite Cloud picks it up via its `appwrite/server-ce: 1.9.x-dev` constraint.

## Test plan

- [x] `php -l` + Pint pass on the changed file
- [x] `composer update utopia-php/messaging` resolves to 1.2.0; `MetadataParameter::UUID->value` autoloads
- [ ] Send a phone OTP and confirm `UUID` = project ID appears in the MSG91 delivery report

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -76,7 +76,7 @@
         "utopia-php/locale": "0.8.*",
         "utopia-php/lock": "0.2.*",
         "utopia-php/logger": "0.8.*",
-        "utopia-php/messaging": "^1.1",
+        "utopia-php/messaging": "^1.2",
         "utopia-php/migration": "1.14.*",
         "utopia-php/platform": "^1.0@RC",
         "utopia-php/pools": "1.*",
--- a/composer.lock
+++ b/composer.lock
@@ -4348,16 +4348,16 @@
         },
         {
             "name": "utopia-php/messaging",
-            "version": "1.1.0",
+            "version": "1.2.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/utopia-php/messaging.git",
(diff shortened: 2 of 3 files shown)


lohanidamodar (approved):
let's make sure to test if possible to make sure it works as expected.

PR #12659, opened 2026-06-23
fix: reset event queue when params are unset to prevent generateEvents errors

HarshMN2345:
## What

Fixes three Sentry issues where `[x] is missing from the params` was thrown by `Event::generateEvents()`:

- `[functionId] is missing` — `functions.createDeployment` (5.6K events, 622 users)
- `[siteId] is missing` — `sites.createDeployment` (13K events, 1.7K users)
- `[userId] is missing` — `account.deleteSessions` (818 events, 38 users)

## Why

All three share the same root cause: the event label (e.g. `functions.[functionId].deployments.[deploymentId].create`) is set on the route and copied to the event queue in the before-middleware, but `setParam()` is only called on certain code paths. When those paths aren't taken, the shutdown hook calls `generateEvents()` with placeholders that have no matching params → exception.

- **Functions/Sites `createDeployment`**: chunked uploads only call `setParam('functionId/siteId', ...)` on the final chunk. Intermediate chunks left the event queue with an unresolvable pattern.
- **`account.deleteSessions`**: `setParam('userId', ...)` is only called when the current session is among those deleted. API key / JWT auth with no matching cookie session skipped this path entirely.

## Fix

Call `$queueForEvents->reset()` on the paths that don't complete the operation, so no event is dispatched.


HarshMN2345:
@greptile This matches the existing payload shape used by the sibling /v1/users/:userId/sessions DELETE endpoint for the same event. users.[userId].sessions.delete has always carried a user model payload

HarshMN2345:
@greptile re-review

HarshMN2345:
@greptile This is out of scope. Partial session deletion on failure is a pre-existing behavior unrelated to the missing-params crash. This PR fixes the crash, a partial delete that throws will surface the real DB error in Sentry, not the missing-param error, since the event is now resolvable.

HarshMN2345:
@greptile re-review

ChiragAgg5k (approved):

app/controllers/api/account.php
```diff
@@ -604,6 +604,16 @@
         $sessions = $user->getAttribute('sessions', []);
         $currentSession = null;
 
+        // Pre-set a resolvable bulk event before any work that can throw, so mid-loop exceptions
+        // don't leave the route's unresolvable [userId]/[sessionId] placeholders in the queue.
```
ChiragAgg5k: why do we need to reset? each coroutine should get an empty queue right?

ChiragAgg5k (changes requested):

HarshMN2345:
Closing — main already resets the event queue at all three crash sites (`Event.php:511`, `:515`, `:519`). The one hunk that hasn't landed is a regression rather than a fix: returning `[]` breaks `Realtime.php:98`, which indexes `$allEvents[0]` without a guard.

PR #12673, opened 2026-06-25
feat: add usage.resource to resource-scoped Get and XList endpoints

lohanidamodar:
## Summary

Pairs with #12663 (which introduced `usage.resource` and applied it to audited write endpoints + the two reads that emit metrics directly). This PR fills the remaining gap: `Bus/Listeners/Usage` emits `METRIC_NETWORK_REQUESTS` / `_INBOUND` / `_OUTBOUND` on **every** HTTP request through the route's Context, so read endpoints without `usage.resource` have their network metrics fall back to `resource='project'` / `resourceId=<projectId>` — even when the URL targets a specific database, function, bucket, etc.

## Derivation rule

For every `Get.php` / `XList.php` with a path param that doesn't already carry `audits.resource` or `usage.resource`: take the first `:param`, look at the segment immediately before it, singularize, and use the param as resourceId.

| URL | Generated `usage.resource` |
|---|---|
| `/v1/databases/:databaseId/...` | `database/{request.databaseId}` |
| `/v1/storage/buckets/:bucketId/files/...` | `bucket/{request.bucketId}` |
| `/v1/functions/:functionId/executions/...` | `function/{request.functionId}` |
| `/v1/teams/:teamId/memberships/...` | `team/{request.teamId}` |

## Manual normalizations

Two cases where the derivation pulled the wrong stem:

- **`/v1/tablesdb/:databaseId/...`, `/v1/vectorsdb/:databaseId/...`, `/v1/documentsdb/:databaseId/...`** → `database/{request.databaseId}` (not `tablesdb`/`vectorsdb`/`documentsdb`), matching the resource shape on the existing write endpoints (their `usage.resource` is `database/{response.$id}`).
- **`/v1/storage/:bucketId/usage`** → `bucket/{request.bucketId}` (not `storage`), same consistency reason.

## Skipped on purpose

- Routes that already carry `audits.resource` or `usage.resource` — untouched.
- **`Health/Http/Queue/...`** — internal ops/probe surface, no customer-facing resource shape; project-fallback is fine.
- Routes with no path param (project-scoped XLists like `/v1/databases`, `/v1/storage/buckets`) — already correctly tagged via project-fallback.

## What this is NOT

This does not change any audit log behavior. `audits.resource` is untouched on every endpoint.

## Test plan

- [x] `composer format` clean on all 114 touched files
- [ ] After deploy + cloud vendor bump: `GET /v1/databases/.../documents/X` writes a `network.requests` row tagged `resource='database'`, `resourceId={databaseId}` (was `'project'` / projectId)
- [ ] After deploy: project-scoped GETs like `/v1/databases` still tag as `project` via fallback (no regression)
- [ ] After deploy: audit log entries are unchanged on every touched endpoint

### DIFF
--- a/src/Appwrite/Platform/Modules/Advisor/Http/Insights/Get.php
+++ b/src/Appwrite/Platform/Modules/Advisor/Http/Insights/Get.php
@@ -30,6 +30,7 @@ class Get extends Action
             ->desc('Get insight')
             ->groups(['api', 'advisor'])
             ->label('scope', 'insights.read')
+            ->label('usage.resource', 'report/{request.reportId}')
             ->label('resourceType', RESOURCE_TYPE_INSIGHTS)
             ->label('sdk', new Method(
                 namespace: 'advisor',
--- a/src/Appwrite/Platform/Modules/Advisor/Http/Insights/XList.php
+++ b/src/Appwrite/Platform/Modules/Advisor/Http/Insights/XList.php
@@ -36,6 +36,7 @@ class XList extends Action
             ->desc('List insights')
             ->groups(['api', 'advisor'])
             ->label('scope', 'insights.read')
+            ->label('usage.resource', 'report/{request.reportId}')
             ->label('resourceType', RESOURCE_TYPE_INSIGHTS)
             ->label('sdk', new Method(
                 namespace: 'advisor',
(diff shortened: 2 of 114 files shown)


lohanidamodar:
Superseded by #12674 — that PR cherry-picks the in-scope subset of routes from this branch alongside the broader usage.resource scope narrowing and shared resource parser.

PR #12674, opened 2026-06-25
feat(usage): limit usage.resource to resource-scoped routes and share resource parser

lohanidamodar:
## Summary

- Extracts a shared `Appwrite\Resource\Parser` used by both `audits.resource` rendering and the cloud-side `usage.resource` resolver.
- Strips `usage.resource` from routes that are not under one of the resource paths the usage warehouse cares about: `databases`, `tablesdb`, `documentsdb`, `vectorsdb`, `storage`, `functions`, `sites`.
- Adds `usage.resource` to the read endpoints (Get/XList) under those same prefixes so network metrics carry a resource tag on reads too.

## Why

`usage.resource` was previously mirrored onto every route that already had `audits.resource`. That gave us project/account/team/etc. tags on routes whose ClickHouse usage rows do not benefit from a `resource` dimension. The set of resource paths the usage warehouse aggregates against is small and well-defined; this PR aligns the label to that scope.

The granularity rule for usage (capped at table/bucket/function/site) is now codified in `Parser::capForUsage()`. Audit logging keeps the original full-depth resource string; only the usage path applies the cap.

This supersedes #12673 (URL-derived GET-endpoint sweep).

## Test plan

- [ ] CI lint + PHPStan green
- [ ] Verify audit logging still records sub-resource IDs (e.g. document/file/execution)
- [ ] Verify usage metrics for documents reads/writes tag with `resource = database/<dbId>/table` and `resourceId = <collectionId>`
- [ ] Verify usage metrics for storage file ops tag with `resource = bucket` and `resourceId = <bucketId>`


src/Appwrite/Reference/Renderer.php
eldadfux, line 1: Still not feeling great about this one.

1. Calling the class name same as the method feels off.
2. We need both parse and un-parse functionality here, might make more sense to have `Reference\Reference::encode` and `Reference\Reference:decode` and add tests
3. Passing an array to construct with no real instructions can get messy. Document param seems sensible, can we have just that?
4. Why do we need `preg_match_all`? We should avoid regex, it can get expensive and hard to maintain. Especially when this is the parse method.
5. Parser feels too magical, how can you tell collection has 2 levels and row has 3 levels?

src/Appwrite/Platform/Modules/Account/Http/Account/MFA/Authenticators/Create.php
```diff
@@ -42,7 +42,6 @@ public function __construct()
             ->label('scope', 'account')
             ->label('audits.event', 'user.update')
             ->label('audits.resource', 'user/{response.$id}')
-            ->label('usage.resource', 'user/{response.$id}')
```
eldadfux, line 45: is it safe for current usage solution to remove all already?

src/Appwrite/Platform/Modules/Databases/Http/TablesDB/Tables/Columns/Get.php
```diff
@@ -43,6 +43,7 @@ public function __construct()
             ->desc('Get column')
             ->groups(['api', 'database'])
             ->label('scope', ['tables.read', 'collections.read', 'columns.read', 'attributes.read'])
+            ->label('usage.resource', 'database/{request.databaseId}')
```
eldadfux, line 46: Thinking about it, with this method we can't really track multiple references for same metric usage, we can only know the database, not the database and table. Any chance clickhouse can support arrays for this well enough?

src/Appwrite/Platform/Modules/Functions/Http/Runtimes/XList.php
```diff
@@ -30,6 +30,7 @@ public function __construct()
             ->groups(['api'])
             ->desc('List runtimes')
             ->label('scope', 'public')
+            ->label('usage.resource', 'function')
```
eldadfux: This is not really tied to a specific resource, why do we need it? we can just use the service attribute to break it down later

src/Appwrite/Platform/Modules/Functions/Http/Specifications/XList.php
```diff
@@ -31,6 +31,7 @@ public function __construct()
             ->groups(['api', 'functions'])
             ->desc('List specifications')
             ->label('scope', 'functions.read')
+            ->label('usage.resource', 'function')
```
eldadfux: Same for all of these, a reference must point at a single resource

src/Appwrite/Platform/Modules/Functions/Http/Templates/XList.php
```diff
@@ -35,6 +35,7 @@ public function __construct()
             ->desc('List templates')
             ->groups(['api'])
             ->label('scope', 'public')
+            ->label('usage.resource', 'function')
```
eldadfux: same

src/Appwrite/Platform/Modules/Sites/Http/Templates/XList.php
```diff
@@ -34,6 +34,7 @@ public function __construct()
             ->desc('List templates')
             ->groups(['api'])
             ->label('scope', 'public')
+            ->label('usage.resource', 'site')
```
eldadfux: Same as with function reference

src/Appwrite/Platform/Modules/Storage/Http/Buckets/Files/Download/Get.php
```diff
@@ -43,6 +43,7 @@ public function __construct()
             ->desc('Get file for download')
             ->groups(['api', 'storage'])
             ->label('scope', 'files.read')
+            ->label('usage.resource', 'bucket/{request.bucketId}/file/{request.fileId}')
```
eldadfux, line 46: We don't want such a low level breakdown of files level, we can stop at bucket

src/Appwrite/Platform/Modules/Account/Http/Account/MFA/Authenticators/Create.php
lohanidamodar: yes it is, we are not yet using it, we added it in last pr, but had to rethink

PR #12676, opened 2026-06-25
fix: use target user for admin membership lookup 

HarshMN2345:
## Summary

When impersonating, the admin block membership lookup was using `$user` (the operator/actor) instead of `$targetUser` (the impersonated user) to find team roles. Since the operator is not a member of the target user's organization, `$adminRoles` was always empty → `USER_UNAUTHORIZED` thrown on every impersonated console request.

## Root cause

`app/controllers/shared/api.php` line 302:
```php
// before
$memberships = $user->getAttribute('memberships', []);

// after
$membershipSource = !$impersonatorUser->isEmpty() ? $targetUser : $user;
$memberships = $membershipSource->getAttribute('memberships', []);
```

## Test plan

- [ ] Impersonate a user in the console — project page loads without Access Denied
- [ ] Normal (non-impersonation) console access still works

### DIFF
--- a/app/controllers/shared/api.php
+++ b/app/controllers/shared/api.php
@@ -299,7 +299,8 @@ Http::init()
         elseif (($project->getId() === 'console' && ! $team->isEmpty() && ! $user->isEmpty()) || ($project->getId() !== 'console' && ! $user->isEmpty() && $mode === APP_MODE_ADMIN)) {
             $teamId = $team->getId();
             $adminRoles = [];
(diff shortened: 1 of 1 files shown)


HarshMN2345:
@greptile re-review

HarshMN2345:
@greptile it existed before not a problem

ArnabChatterjee20k (approved):

HarshMN2345:
@greptile re-review

PR #12680, opened 2026-06-25
fix(users): rebuild search index on partial user updates

deepshekhardas:
When updating user attributes (name, email, phone, labels), include \search\ in the update document to trigger the \userSearch\ filter, which rebuilds the search index from current values.\n\n(cherry picked from the original fix in #12650, retargeted to \1.9.x\)


HarshMN2345:

PR #12686, opened 2026-06-25
Fix #12643: Update project and user accessedAt during GraphQL requests

### LINKED ISSUE #12643: Uploading files greater then 500MB to storage fails
### 👟 Reproduction steps

1. Create a new bucket in an empty project
2. Select create file
3. Attempt to upload a file that is over 500MB (750MB-1GB seems to trigger consistently, sometimes ~500MB goes through)
   - **NOTE:** Sometimes the first upload will go through but further attempts fail, even after restart of the server
   - I have also attempted this with local and S3 storage with similar results in both cases
   - Attempted to upload files of size 578.9MB and 1.3GB for testing

### 👍 Expected behavior

File uploads successfully 

### 👎 Actual Behavior

File uploads fails with the following error in the appwrite pod if using S3 storage method:
```
Fatal error: Allowed memory size of 536870912 bytes exhausted (tried to allocate 5242880 bytes) in /usr/src/code/vendor/utopia-php/storage/src/Storage/Device/S3.php on line 228
[2026-06-22 05:14:27 *18.9]     ERROR   php_swoole_server_rshutdown() (ERRNO 503): Fatal error: Allowed memory size of 536870912 bytes exhausted (tried to allocate 5242880 bytes) in /usr/src/code/vendor/utopia-php/storage/src/Storage/Device/S3.php on line 228
[2026-06-22 05:14:27 $7.0]      WARNING Worker::report_error(): worker(pid=18, id=9) abnormal exit, status=255, signal=0
```

The same error occurs for local storage method but in `/usr/src/code/vendor/utopia-php/storage/src/Storage/Device/Local.php`

### 🎲 Appwrite version

### 💻 Operating system

Linux

### 🧱 Your Environment

- Windows Desktop Dev Environment
   - Appwrite 1.9.0 running through Docker Desktop using WSL
   - 64GB or ram with at least 32GB available during tests
   - Attempted with forced allocation of 16GB through deployment section
-  Digital Ocean Appwrite Marketplace Server 1
   - Appwrite 1.9.0 running through docker
   -  Basic Shared CPU instance with 2vCPU and 4GB of ram
   - Free ram was between 700-450MB during uploads
-  Digital Ocean Appwrite Marketplace Server 2
   - Appwrite 1.9.0 running through docker
   - Basic Shared CPU instance with 8vCPU and 16GB of ram
   - Free ram stayed above 12GB during uploads
- All environments used the same .env with only the domain changed
- GCP Storage was used for testing the S3 storage method

hardikkaurani:
This PR updates the project and user accessedAt timestamps on incoming GraphQL requests. Previously, GraphQL requests did not trigger these updates, causing projects that only use GraphQL to be incorrectly flagged as inactive.

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -1680,10 +1680,7 @@ Http::get('/v1/account/sessions/oauth2/:provider/redirect')
                     $failureRedirect(Exception::USER_EMAIL_DISPOSABLE);
                 }
 
-                if (($plan['supportsCanonicalEmailValidation'] ?? false) && ($project->getAttribute('auths', [])['canonicalEmails'] ?? false) && $emailMetadata['emailIsCanonical'] === false) {
-                    $failureRedirect(Exception::USER_EMAIL_NOT_CANONICAL);
-                }
-
+                
                 if (($plan['supportsFreeEmailValidation'] ?? false) && ($project->getAttribute('auths', [])['freeEmails'] ?? false) && $emailMetadata['emailIsFree']) {
                     $failureRedirect(Exception::USER_EMAIL_FREE);
                 }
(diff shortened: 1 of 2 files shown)


hardikkaurani:
Thanks for the thorough review! Reverted the unrelated `account.php` change in the latest commit — the `USER_EMAIL_NOT_CANONICAL` enforcement is restored in both OAuth2 redirect branches, so the `canonicalEmails` project policy stays consistent across every auth path. That diff had leaked in from another branch; this PR now contains only the GraphQL `accessedAt` timestamp fix.

PR #12687, opened 2026-06-25
fix: use warning instead of exception on adapter mismatch

HarshMN2345:
## Summary
- Replaces `BuildException` throw with `Console::warning` when post-build rendering detection disagrees with the explicitly set adapter

## Why
Post-build detection can incorrectly flag SSR sites as static when they have prerendered pages. Since the developer has explicitly set the adapter, their setting should take precedence — blocking the build is too aggressive.


HarshMN2345:
Closing — this logic no longer lives in `Builds.php`. On main it sits at `Jobs.php:545` and returns a build error rather than throwing, and `tests/e2e/Services/Sites/SitesCustomServerTest.php:1643` asserts the mismatch surfaces in the build logs. Downgrading it to a warning would fail that test.

PR #12699, opened 2026-06-26
Fix #12623: Add _APP_SMTP_KEEPALIVE to opt out of SMTP connection reuse

### LINKED ISSUE #12623: 🐛 SMTP keepalive causes 421 timeout on first send after idle; subsequent sends work (1.9.0 self-hosted)

hardikkaurani:
Fixes #12623

### What does this PR do?

Adds a new `_APP_SMTP_KEEPALIVE` environment variable (default `enabled`, so current behaviour is unchanged) that lets self-hosted operators turn off SMTP connection reuse in `worker-mails`. When set to `disabled`, every send opens a fresh SMTP connection instead of reusing a kept-alive one.

### Why is this needed?

`worker-mails` keeps the SMTP connection open between sends via PHPMailer `SMTPKeepAlive`. If the upstream SMTP provider or an intermediate firewall/NAT silently drops the idle connection, the worker reuses a now-stale socket on the next send. The `MAIL FROM` command hangs and the server returns `421 — closing connection`, so the **first** send after an idle window fails (e.g. an OTP, magic-URL, or verification email); the next attempt opens a fresh connection and succeeds.

End-users experience this as "OTP not delivered → retry → works", with no server-side mitigation available short of a code change. This var gives operators stuck behind such providers/NATs a clean opt-out without changing the default for anyone else.

### Changes

- `app/init/registers.php` and `src/Appwrite/Platform/Workers/Mails.php`: derive the adapter `keepAlive` flag from `_APP_SMTP_KEEPALIVE` instead of hardcoding `true`.
- `app/config/variables.php`: document the new variable (introduced in `1.9.6`).
- `.env`, `app/views/install/compose.phtml`, `docker-compose.yml`, `tests/resources/docker/docker-compose.yml`: expose the variable to the `appwrite` and `appwrite-worker-mails` services.

### Test Plan

- With `_APP_SMTP_KEEPALIVE` unset or `enabled`: behaviour is identical to today (connection reuse preserved).
- With `_APP_SMTP_KEEPALIVE=disabled`: each email opens a fresh SMTP connection, so a send after a long idle period no longer fails with the stale-connection `421` error.

Yes.


eldadfux:

PR #12707, opened 2026-06-28
fix: migrate missing user email canonical attributes in V24

### LINKED ISSUE #12705: URGENT: Upgrade from 1.8.x to 1.9.x leaves users metadata inconsistent, causing Unknown attribute: "emailCanonical" and breaking authentication
### 👟 Reproduction steps

Install a self-hosted Appwrite 1.8.x instance.
Create one or more projects and users.
Upgrade the instance to Appwrite 1.9.0 using the official upgrade procedure.
Restart all containers and wait for migrations to complete.
Attempt to log in through the Console or call:
POST /v1/account/sessions/email

or

GET /v1/account
Authentication fails with:
Utopia\Database\Exception\Structure:
Invalid document structure: Unknown attribute: "emailCanonical"

The same behavior is still present after upgrading the instance further to 1.9.5.

### 👍 Expected behavior

The upgrade should migrate all existing projects and internal system collections completely.

After upgrading, authentication should continue working without requiring any manual SQL changes or metadata modifications.

### 👎 Actual Behavior

Authentication becomes completely unusable after upgrading.

Every login request returns:

500 Internal Server Error

Utopia\Database\Exception\Structure:
Invalid document structure: Unknown attribute: "emailCanonical"

During investigation I found that:

Existing project users tables are not fully upgraded.
Existing project users collection metadata is not fully upgraded.
A fresh Appwrite 1.9.x installation contains additional user fields that are missing after migration.
Simply adding the SQL columns is not enough because Appwrite still validates against outdated collection metadata.

Another community member reported fixing the issue by manually:

adding the missing SQL columns,
patching every users collection metadata (_xx__metadata),
backfilling emailCanonical = LOWER(email),
flushing Redis,
restarting Appwrite.

This strongly suggests the migration does not fully update existing project user collections.

### 🎲 Appwrite version

Version 1.8.x

### 💻 Operating system

Linux

### 🧱 Your Environment

Self-hosted Appwrite
Docker Compose
MariaDB
Redis
Official Appwrite Docker images
No custom modifications

The issue was reproduced on an upgraded production instance and compared against a fresh 1.9.x installation.

Comparison showed that the upgraded instance is missing the new user attributes in the migrated project metadata, while the fresh installation is correct.

No SDK is involved because the issue occurs directly through the Appwrite Console and REST API.

hardikkaurani:
Fixes #12705

## What does this PR do?
This PR ensures that missing user email canonical attributes (`emailCanonical`) are properly migrated in the V24 migration script.

## Test Plan
No tests run

## Related Issues
Closes #12705

### DIFF
--- a/src/Appwrite/Migration/Version/V24.php
+++ b/src/Appwrite/Migration/Version/V24.php
@@ -204,21 +204,43 @@ class V24 extends Migration
                         'emailIsCorporate',
                         'emailIsCanonical',
                     ];
-                    try {
-                        $this->createAttributesFromCollection($this->dbForProject, $id, $attributes);
-                    } catch (Throwable $th) {
-                        Console::warning('Failed to create attributes "' . \implode(', ', $attributes) . "\" in collection {$id}: {$th->getMessage()}");
+                    $collectionDoc = $this->dbForProject->getCollection($id);
+                    $existingAttributes = [];
+                    $existingIndexes = [];
+                    if (!$collectionDoc->isEmpty()) {
+                        foreach ($collectionDoc->getAttribute('attributes', []) as $attributeDoc) {
+                            $existingAttributes[] = $attributeDoc->getAttribute('key');
+                        }
+                        foreach ($collectionDoc->getAttribute('indexes', []) as $indexDoc) {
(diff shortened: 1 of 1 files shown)


GenGraphic:
This should be approved faster

Issue #12622, opened 2026-06-19
🐛 API key returns 401 every 24h on read-only requests; re-saving the key fixes it (1.9.0 self-hosted)


nighodev:
**Note:** This bug report was filed using AI assistance. The bug itself was triaged and verified by a real human (encountered in production, reproduced, workarounds tested).

## 🐛 Describe the bug

On self-hosted Appwrite **1.9.0**, server-side API keys with read scopes (`documents.read`, `columns.read`, `files.read`) start failing with `401 The current user is not authorized to perform the requested action` on a fixed ~24h cycle. The key has `expire: ""` (set to "Never" in the console UI) and the scope set is unchanged.

The same key continues to work indefinitely if any field is re-saved on it in the console — re-saving is the *only* reliable fix in our case.

## 👟 Reproduction steps

1. Self-host Appwrite 1.9.0 via the official Docker compose.
2. Create a project, add a Database + Collection, set `Permission.read(any)` on the collection.
3. Create an API key with scopes: `databases.read`, `tables.read`, `collections.read`, `columns.read`, `attributes.read`, `indexes.read`, `rows.read`, `documents.read`, `files.read`. Set Expiration to **Never**.
4. From a server-side client (Next.js Sites SSR in our case, but reproduces with raw `curl` too), call `GET /v1/databases/{db}/collections/{col}/documents` using the key.
5. The call works for ~24h. After 24h elapses since the key was last saved, the next call returns:
   ```json
   {"message":"The current user is not authorized to perform the requested action.","code":401,"type":"general_unauthorized_scope"}
   ```

## 🔬 What we've ruled out

- **Expiry**: the API returns `expire: ""` for the key (queried via `GET /v1/projects/{pid}/keys`).
- **Missing scopes**: every relevant read scope is present (full list above).
- **Stale env / cached secret on the client**: same behaviour from a fresh `curl` with the secret pasted in.
- **Redis cache (partially)**: `docker exec appwrite-redis redis-cli FLUSHDB` fixes it for **one** of our affected projects, but **not** for another with identical setup, which suggests broken state also lives in the `appwrite` PHP container's in-process cache, not only Redis.

## 🧪 Workarounds that work

- Re-saving the key in the console (100% reliable).
- `FLUSHDB` on the Redis container (sometimes — project-dependent).
- `docker restart appwrite` (untested for all cases — likely reliable since it clears PHP-process state).

## 🎲 Appwrite version

[...]

PR #12729, opened 2026-06-30
fix(compose): add missing _APP_DOMAIN* and _APP_OPTIONS_FORCE_HTTPS env vars to appwrite-worker-executions and appwrite-task-scheduler-messages

nighodev:
## Summary

The `appwrite-worker-executions` and `appwrite-task-scheduler-messages` service definitions in `docker-compose.yml` are missing `_APP_DOMAIN`, `_APP_DOMAIN_FUNCTIONS`, `_APP_DOMAIN_SITES`, and `_APP_OPTIONS_FORCE_HTTPS` from their `environment:` blocks. Their sibling services (`appwrite-task-scheduler-functions`, `appwrite-task-scheduler-executions`, `appwrite-worker-functions`) all have these vars — this looks like an oversight on the two services touched here.

## Impact

`appwrite-worker-executions` reads `_APP_DOMAIN` (via `app/config/platform.php` → `$platform['apiHostname']`) when computing `APPWRITE_FUNCTION_API_ENDPOINT` for runtime injection (`src/Appwrite/Platform/Workers/Functions.php:480`). Without the env, the worker boots with `apiHostname='localhost'` and every scheduled function execution receives `APPWRITE_FUNCTION_API_ENDPOINT=https://localhost/v1`. The function's SDK calls back to `https://localhost/v1` (nothing listening inside the runtime container) → `fetch failed` errors.

HTTP-triggered functions are unaffected because they go through the `appwrite` container, which has `_APP_DOMAIN`. Scheduled functions break silently — they complete with HTTP 200 but log the fetch failure, so it can take a while to notice in production.

`appwrite-task-scheduler-messages` likely has the same effect on scheduled SMS/email message delivery though I haven't traced that one end-to-end.

## Changes

Add the four env vars to both services' `environment:` blocks. Matches the layout already present on the three other scheduler/function services.

## Tested

Self-hosted Appwrite 1.9.0 instance: applied the same fix via `docker-compose.override.yml`, verified runtime now receives the correct `APPWRITE_FUNCTION_API_ENDPOINT` (matching `_APP_DOMAIN` from `.env`), scheduled function executions complete cleanly.

## Related

- Companion to the `_APP_OPENSSL_KEY_V1` fix on `appwrite-worker-executions` already on `main` — same class of bug (oversight in the env block of this particular worker).
- Detailed root-cause analysis + reproduction: #12622

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -668,6 +668,10 @@ services:
       - _APP_WORKER_PER_CORE
       - _APP_POOL_ADAPTER
       - _APP_OPENSSL_KEY_V1
+      - _APP_OPTIONS_FORCE_HTTPS
+      - _APP_DOMAIN
+      - _APP_DOMAIN_FUNCTIONS
+      - _APP_DOMAIN_SITES
       - _APP_REDIS_HOST
       - _APP_REDIS_PORT
       - _APP_REDIS_USER
@@ -1162,6 +1166,10 @@ services:
       - _APP_WORKER_PER_CORE
       - _APP_POOL_ADAPTER
       - _APP_OPENSSL_KEY_V1
+      - _APP_OPTIONS_FORCE_HTTPS
+      - _APP_DOMAIN
+      - _APP_DOMAIN_FUNCTIONS
+      - _APP_DOMAIN_SITES
       - _APP_REDIS_HOST
       - _APP_REDIS_PORT
       - _APP_REDIS_USER


nighodev:
Good catch on the inconsistency, but I think `appwrite-worker-functions` doesn't actually use those two vars — at least not in 1.9.0 / main.

`src/Appwrite/Platform/Workers/Functions.php` only touches `$platform['apiHostname']` (one reference at line 480, computing `APPWRITE_FUNCTION_API_ENDPOINT`). The other platform fields, `functionsDomain` and `sitesDomain`, only appear in:

- `Modules/Functions/Http/Functions/Create.php:377` — HTTP path in the `appwrite` container, which already has the env
- `Modules/Functions/Workers/Builds.php:1036` — `appwrite-worker-builds`, which already has the env

So adding `_APP_DOMAIN_FUNCTIONS` and `_APP_DOMAIN_SITES` to `appwrite-worker-functions` would inject vars that no code path in that worker reads. Happy to add them if you'd prefer the consistency anyway, but I'd rather keep the PR strictly scoped to vars that fix observable behavior — leaves room for a separate audit pass if there are other inconsistencies worth surfacing in one go.

eldadfux:

PR #12736, opened 2026-06-30
fix: restrict CodeQL workflow permissions on branch 1.5.x

sricharan218:
The codeql-analysis.yml workflow on 1.5.x had no permissions block, causing it to inherit the default repository token with excessive permissions.

This adds:
- `permissions: contents: read` at the workflow level
- `permissions: security-events: write` at the job level

This matches the fix already applied on the default branch in f8b89b2.


HarshMN2345:
Closing as this targets a branch that is no longer maintained.

Issue #12715, opened 2026-06-29
Site can not host static html


qkxie:
### 👟 Reproduction steps

I have a static html file called abc.html

I drag and drop it on Appwrite Sites. and choose "Other" in framework.

After deploy finished, I open the link, but it show 404

### 👍 Expected behavior

Open the html page.

### 👎 Actual Behavior

404

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

MacOS

### 👀 Have you spent some time to check if this issue has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

sricharan218:
Hello i can try look into it

PR #12744, opened 2026-06-30
Fix static site serving 404 for pre-selected static adapter with single non-index file

hardikkaurani:
### Summary
This PR fixes issue #12715 where a static site deployed with a pre-selected adapter (e.g. static for the 'Other' framework) and containing only a single non-index file (e.g. abc.html) would return a 404 error on /.

### Details
Previously, if a site's adapter was pre-selected (such as static), the build worker would entirely skip the auto-detection block. This resulted in the deployment's fallbackFile and adapter attributes never being set, leading to a 404 when serving static packages containing only a single non-index HTML page.

This PR optimizes the fallback and adapter resolution in the build worker (Builds.php) to:
1\. Always populate the deployment's adapter attribute matching the site's adapter when pre-selected (or auto-detected).
2\. Sync the detected fallback file to both the site document and the deployment document if it is resolved during building.
3\. Fallback to using the site's existing fallback file (if any) when the detection logs do not resolve an explicit fallback, avoiding clobbering user-defined configurations.
4\. Correctly raise BuildException on adapter mismatch (e.g. SSR site detecting static content).

We also added a comprehensive E2E regression test (testServeStaticSinglePageWithStaticAdapter) in SitesCustomServerTest.php to verify successful resolution and persistence across deployments and redeployments.

### DIFF
--- a/src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
+++ b/src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
@@ -923,12 +923,9 @@ class Builds extends Action
 
                         $deployment->setAttribute('adapter', $adapter);
                         $deployment->setAttribute('fallbackFile', $detection->getFallbackFile() ?? '');
-                        Span::add('build.adapter', $adapter);
-                        Span::add('build.fallback_file', $deployment->getAttribute('fallbackFile'));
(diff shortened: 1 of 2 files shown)


eldadfux:

PR #12752, opened 2026-07-01
fix(builds): truncate buildLogs to attribute limit to avoid oversized-log build failures

levivannoort:
## What

Truncate the deployment `buildLogs` to the attribute's maximum length before every write in the Builds worker, reusing the same trimming convention already used for execution logs.

## Why

Large Sites builds (heavy prerender / bundle output — a big site emits thousands of `vite build` lines) can produce build logs longer than the `deployments.buildLogs` attribute limit (1,000,000). Persisting the deployment then throws:

```
Utopia\Database\Exception\Structure:
Attribute "buildLogs" ... must be ... no longer than 1000000 chars
```

Crucially this also fires **inside the build-failure handler**, which prepends the previous (huge) logs to the error message and writes again — so the handler re-throws the same exception uncaught. The result: the build is reported as failed even though it **succeeded** and the artifact was uploaded, and because nothing could be persisted the user sees a generic "internal error" placeholder with no logs. Trimming log verbosity at the source doesn't reliably help — the framework's own build output is large enough on its own.

## How

- Add `APP_BUILD_LOGS_LENGTH_LIMIT` (sibling to `APP_FUNCTION_LOG_LENGTH_LIMIT`, matching the `buildLogs` attribute size).
- Add a `truncateBuildLogs()` helper that mirrors the existing execution-log trimming (`Functions.php` / `Executions/Create.php` / `general.php`): **byte-based `strlen`/`substr`**, the standard `"[WARNING] Logs truncated. The output exceeded {N} characters."` prefix, keeping the **tail** (which holds the build result and any error).
- Apply it at every `buildLogs` write, including the failure handler, so the invariant "persisted `buildLogs` is always within the limit" holds on all paths (streamed, final, and error).

Byte-based trimming is used deliberately: capping bytes ≤ limit guarantees the value fits regardless of whether length is enforced by bytes or characters (`mb_strlen ≤ strlen`).

## Notes

A more thorough long-term direction would be to store full build logs outside the document (storage) and keep only a tail on the deployment, removing the size ceiling and data loss entirely. This PR is the right-sized fix that stops the failures and matches existing conventions.

### DIFF
--- a/app/init/constants.php
+++ b/app/init/constants.php
@@ -269,8 +269,9 @@ const APP_AUTH_TYPE_ADMIN = 'Admin';
 // Response related
 const MAX_OUTPUT_CHUNK_SIZE = 10 * 1024 * 1024; // 10MB
 const APP_LIMIT_UPLOAD_CHUNK_SIZE = 5 * 1024 * 1024; // 5MB
-const APP_FUNCTION_LOG_LENGTH_LIMIT = 1000000;
-const APP_FUNCTION_ERROR_LENGTH_LIMIT = 1000000;
+const APP_LOG_LENGTH_LIMIT = 1000000; // Single source of truth for log/error length limits (matches the related attribute sizes)
+const APP_FUNCTION_LOG_LENGTH_LIMIT = APP_LOG_LENGTH_LIMIT;
+const APP_FUNCTION_ERROR_LENGTH_LIMIT = APP_LOG_LENGTH_LIMIT;
 // Function headers
 const FUNCTION_ALLOWLIST_HEADERS_REQUEST = ['content-type', 'agent', 'content-length', 'host', 'x-appwrite-client-ip'];
 const FUNCTION_ALLOWLIST_HEADERS_RESPONSE = ['content-type', 'content-length'];
--- a/src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
+++ b/src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
@@ -53,6 +53,26 @@ class Builds extends Action
     }
 
     /**
+     * Truncate build logs to the length allowed by the deployments "buildLogs"
+     * attribute. Large site builds (heavy prerender/bundle output) can exceed
+     * it, which makes persisting the deployment throw a Structure exception and
+     * report the build as failed even though it actually succeeded. Keep the
+     * tail — it holds the build result and any error — mirroring how execution
+     * logs are trimmed elsewhere in this codebase.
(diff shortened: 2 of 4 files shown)


app/init/constants.php
```diff
@@ -268,37 +231,19 @@
 const APP_AUTH_TYPE_ADMIN = 'Admin';
 // Response related
 const MAX_OUTPUT_CHUNK_SIZE = 10 * 1024 * 1024; // 10MB
-const APP_LIMIT_UPLOAD_CHUNK_SIZE = 5 * 1024 * 1024; // 5MB
 const APP_FUNCTION_LOG_LENGTH_LIMIT = 1000000;
 const APP_FUNCTION_ERROR_LENGTH_LIMIT = 1000000;
+const APP_BUILD_LOGS_LENGTH_LIMIT = 1000000; // Matches the deployments "buildLogs" attribute size
```
ChiragAgg5k: maybe we can use a single constant for all APP_FUNCTION_LOG_LENGTH_LIMIT, APP_FUNCTION_ERROR_LENGTH_LIMIT and APP_FUNCTION_ERROR_LENGTH_LIMIT

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
ChiragAgg5k, line 1671: i think we should add a test case for this

app/init/constants.php
levivannoort: Good call — consolidated to a single `APP_LOG_LENGTH_LIMIT` source of truth. `APP_FUNCTION_LOG_LENGTH_LIMIT` / `APP_FUNCTION_ERROR_LENGTH_LIMIT` now alias it (kept so their existing call sites and the functions collection attribute sizes stay intact), and the deployments `buildLogs` attribute size references it too so the column size and the truncation limit can’t drift. Happy to fully collapse the function-specific names if you prefer.

src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
levivannoort: Added unit tests in `tests/unit/Functions/BuildsTest.php` (`testTruncateBuildLogsKeepsShortLogsUnchanged` and `testTruncateBuildLogsTrimsOversizedLogsToLimitKeepingTail`) covering the pass-through, the ≤-limit cap, the warning prefix, and tail preservation — via the existing `callBuilds()` reflection helper.

ChiragAgg5k (approved):

PR #12762, opened 2026-07-02
chore(ci): optimize runner costs with spot instances and leaner matrix

ChiragAgg5k:
## What does this PR do?

Cuts the AWS cost of the `runs-on-for-appwrite` CI stack (~$1,285/month gross in June) with three runner-level changes to `ci.yml`:

1. **Spot instances** — all RunsOn jobs now use `spot=price-capacity-optimized` instead of `spot=false` (on-demand). On-demand EC2 was 84% of this stack's cost. The workload profile is spot-safe: jobs are short (median ~2 min, max ~8 min), the matrix runs with `fail-fast: false`, and PR test steps already retry. `price-capacity-optimized` picks spot pools with the lowest interruption risk.
2. **Leaner push-to-main matrix** — merges to `main` previously ran 27 services × 3 databases × 2 modes (~175 jobs, ~400 runner-minutes); they now run the same MongoDB/dedicated matrix as PRs (~38 jobs, ~90 runner-minutes). The full sweep still runs on `workflow_dispatch` (manual escape hatch) and on PRs that change the `utopia-php/database` version (unchanged behavior).
3. **60GB volumes instead of 120GB** — measured worst-case disk usage per job is ~30GB (OS/AMI ~20GB + compose images ~4GB + open-runtimes images ~1.5GB + writable layers/volumes/logs). 60GB keeps a 2x margin. gp3 performance is independent of size, so no I/O impact.

Expected stack cost after this change: roughly $300–400/month.

## Test Plan

- CI on this PR exercises the changed runner specs directly (all jobs on spot, 60GB volumes).
- The matrix logic change is verified by this PR running MongoDB/dedicated only, and can be verified for the full sweep via a manual `workflow_dispatch` run on this branch.

## Watch items

- Spot interruptions show up as infrastructure-failed jobs — if the longer jobs (Project ~8 min, Databases ~6 min) get interrupted repeatedly, pin just those back to on-demand.
- Any "no space left on device" failure (most likely `build` or `benchmark`) means bumping that job's volume back up.

## Related PRs and Issues

None.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -257,6 +257,7 @@ jobs:
   build:
     name: Build
     runs-on: runs-on=${{ github.run_id }}/runner=4cpu-linux-x64/volume=60g/spot=price-capacity-optimized/extras=otel+s3-cache+ecr-cache
+    timeout-minutes: 30
     steps:
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
@@ -289,6 +290,7 @@ jobs:
   unit:
     name: Tests / Unit
     runs-on: runs-on=${{ github.run_id }}/runner=4cpu-linux-x64/volume=60g/spot=price-capacity-optimized/extras=otel+ecr-cache
+    timeout-minutes: 15
     needs: build
     permissions:
       contents: read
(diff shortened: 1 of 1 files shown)


xuhongchen521-debug:
/attempt #12762

我来做这个，正在写PR。

abnegate (approved):

PR #12792, opened 2026-07-06
feat: adapter-driven VCS provider abstraction (GitHub-only, no behavior change)

HarshMN2345:
## Summary
Part 1/3 of splitting up #12774 into reviewable chunks (originally ~3,700 lines in one PR).

Introduces `Appwrite\Vcs\Provider`/`Resolver` and a provider registry (`app/config/vcs.php`) so future git providers can be added as config entries instead of hardcoding GitHub throughout the codebase.

- `Provider`/`Resolver`: registry lookup, env-var access, adapter/OAuth2 factory, token refresh (with concurrent-refresh recovery), owner resolution across auth types, per-repository webhook helper.
- `Deployment` trait and `Events\Base` extracted from the GitHub-only `VCS/Http/GitHub/{Deployment,Events/Create}` implementations to accept the generic `Utopia\VCS\Adapter\Git` base type; the GitHub classes become thin subclasses / deprecated shims for backward compat with the cloud fork.
- All VCS/Functions/Sites HTTP endpoints and the Builds worker migrated from `inject('gitHub')` to `inject('vcs')`, resolving the adapter from the installation document instead of a hardcoded GitHub instance.
- Console variables now report configured providers generically.

**Zero behavior change for GitHub**: same routes, same SDK methods, same DB schema. This is purely an internal abstraction to unblock adding other providers (Gitea in #2/3) as small, additive follow-ups.

## Stack
- **#1 (this PR)** → adapter abstraction, GitHub-only
- #2 → Gitea provider (additive)
- #3 → Gitea CI/dev environment + live E2E test

## Test plan
- [x] `composer lint` / `composer format` clean
- [x] `composer analyze` (PHPStan level 4) clean
- [x] `tests/unit/Vcs/*` — new Provider/Resolver unit tests
- [x] Full `tests/e2e/Services/{Functions,Sites,VCS}` suites pass unchanged (verified against a live docker environment on this exact branch)


app/config/vcs.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+/**
+ * VCS provider registry.
```
Meldiron: We shouldnt need this large comment - code should be self-explanatory

app/config/vcs.php
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+use Utopia\VCS\Adapter\Git\GitHub;
+
+return [
+    'github' => [
+        'name' => 'GitHub',
+        'enabled' => true,
+        'adapter' => GitHub::class,
+        'oauth2' => 'Appwrite\\Auth\\OAuth2\\Github',
```
Meldiron: Lets pass class here like we do for adapter

app/config/vcs.php
```diff
@@ -0,0 +1,59 @@
...
+        'auth' => 'app',
```
Meldiron: Lets define consts instead of magical strings

app/config/vcs.php
```diff
@@ -0,0 +1,59 @@
...
+        'envPrefix' => '_APP_VCS_GITHUB',
+        'required' => ['APP_NAME', 'PRIVATE_KEY', 'APP_ID', 'CLIENT_ID', 'CLIENT_SECRET'],
```
Meldiron: I would avoid envPrefix - we can have exact env var names - length is not a worryhere.

Also "required" could mean many things - for code quality here, always best to imagine you only know key, and you dont know the value. Suddently it becomes impossible to guess what it could be.

Something like `requiredEnvVariables` might be better - AI might be able to find even better names

app/config/vcs.php
```diff
@@ -0,0 +1,59 @@
...
+        'endpoint' => false,
```
Meldiron: I would want to have github API endpoint here - We can maybe take it from variable in VCS library, if there is only one endpoint. But still better to mirror that, than false

app/config/vcs.php
```diff
@@ -0,0 +1,59 @@
...
+        'browserEndpoint' => 'https://github.com',
+        'urls' => [
+            'repository' => '{base}/{owner}/{repository}',
+            'branch' => '{base}/{owner}/{repository}/tree/{branch}',
+            'commit' => '{base}/{owner}/{repository}/commit/{commit}',
+            'file' => '{base}/{owner}/{repository}/blob/{reference}',
+        ],
```
Meldiron: I would keep flat structure, nesting doesnt bring value here

app/init/configs.php
```diff
@@ -15,6 +15,7 @@
 Config::load('protocols', __DIR__ . '/../config/protocols.php', $configAdapter);
 Config::load('errors', __DIR__ . '/../config/errors.php', $configAdapter);
 Config::load('oAuthProviders', __DIR__ . '/../config/oAuthProviders.php', $configAdapter);
+Config::load('vcs', __DIR__ . '/../config/vcs.php', $configAdapter);  // VCS provider registry
```
Meldiron: Code consistency - no need for comment here, if other config loading dont have it

app/init/constants.php
```diff
@@ -105,9 +105,15 @@
...
-const APP_VCS_GITHUB_URL = 'https://github.com/TeamAppwrite';
+const APP_VCS_COMMIT_USERNAME = 'Appwrite';
+const APP_VCS_COMMIT_EMAIL = 'team@appwrite.io';
+const APP_VCS_COMMIT_AUTHOR_URL = 'https://github.com/TeamAppwrite';
+/** @deprecated Use APP_VCS_COMMIT_USERNAME. Kept for appwrite/cloud. */
+const APP_VCS_GITHUB_USERNAME = APP_VCS_COMMIT_USERNAME;
+/** @deprecated Use APP_VCS_COMMIT_EMAIL. Kept for appwrite/cloud. */
+const APP_VCS_GITHUB_EMAIL = APP_VCS_COMMIT_EMAIL;
+/** @deprecated Use APP_VCS_COMMIT_AUTHOR_URL. Kept for appwrite/cloud. */
+const APP_VCS_GITHUB_URL = APP_VCS_COMMIT_AUTHOR_URL;
```
Meldiron, line 116: I cant see why we would need backwarrds compaibility here, lets remove old variables

app/init/resources.php
```diff
@@ -348,6 +349,9 @@ function getDevice(string $root, string $connection = ''): Device
 
 $container->set('promiseAdapter', fn ($register) => $register->get('promiseAdapter'), ['register']);
 
+$container->set('vcs', fn (Cache $cache) => new VcsResolver($cache), ['cache']);
+
+/** @deprecated Inject `vcs` and resolve the adapter from the installation instead. Kept for appwrite/cloud. */
 $container->set('gitHub', fn (Cache $cache) => new VcsGitHub($cache), ['cache']);
```
Meldiron: Goal is to remove github resource, and use vcs everywhere. We should need backwards compatiblity here.

src/Appwrite/Platform/Modules/Functions/Http/Deployments/Template/Create.php
```diff
@@ -134,7 +136,7 @@ function: $function,
                 dbForProject: $dbForProject,
                 publisherForBuilds: $publisherForBuilds,
                 template: $template,
-                github: $github,
+                github: $adapter,
```
Meldiron: Lets rename it here, github param name should be vcs

src/Appwrite/Platform/Modules/Functions/Http/Deployments/Template/Create.php
```diff
@@ -101,7 +102,7 @@ public function action(
         Event $queueForEvents,
         Document $project,
         BuildPublisher $publisherForBuilds,
-        GitHub $github,
+        Resolver $vcs,
```
Meldiron: Resolver sounds too generic 🤔 I dont think we should have a class called resolver that takes care of all VCS related actions.

src/Appwrite/Utopia/Response/Model/ConsoleVariables.php
```diff
@@ -64,6 +64,13 @@ public function __construct()
                 'default' => false,
                 'example' => true,
             ])
+            ->addRule('_APP_VCS_PROVIDERS', [
+                'type' => self::TYPE_STRING,
+                'description' => 'List of configured VCS providers.',
+                'default' => [],
+                'example' => 'github',
```
Meldiron: Not sure if this is valid example, lets check other array rules. example might need to be array value

src/Appwrite/Platform/Modules/VCS/Http/Installations/Repositories/XList.php
```diff
@@ -123,10 +122,7 @@ public function action(
             throw new Exception(Exception::INSTALLATION_NOT_FOUND);
         }
 
-        $providerInstallationId = $installation->getAttribute('providerInstallationId');
-        $privateKey = System::getEnv('_APP_VCS_GITHUB_PRIVATE_KEY');
-        $githubAppId = System::getEnv('_APP_VCS_GITHUB_APP_ID');
-        $github->initializeVariables($providerInstallationId, $privateKey, $githubAppId);
+        $adapter = $vcs->getAdapter($installation, $dbForPlatform);
```
Meldiron, line 125: Adapter is too generic here, I think we should call it $vcs.

But that conflicts with resolver, which I dont yet understand why we need. Curious if there is a simpler solution

Meldiron: Might be wirth checking with Jake or Luke - I recall there was some changes to cache and DB connection multiplexing, might be useful for us here

src/Appwrite/Platform/Modules/Functions/Http/Deployments/Template/Create.php
HarshMN2345: yeah, makes sense i will change the name to something specific

src/Appwrite/Platform/Modules/VCS/Http/Installations/Repositories/Branches/XList.php
```diff
@@ -81,22 +81,22 @@ public function action(
...
             }
         } catch (RepositoryNotFound $e) {
             throw new Exception(Exception::PROVIDER_REPOSITORY_NOT_FOUND);
         }
 
-        $branches = $github->listBranches($owner, $repositoryName, search: $search);
+        // Server-side branch search is a GitHub-only capability
+        $branches = $adapter instanceof GitHub
+            ? $adapter->listBranches($owner, $repositoryName, search: $search)
+            : $adapter->listBranches($owner, $repositoryName);
```
Meldiron, line 99: Any adapter-speciifc limitation should live in VCS, not in Appwrite

src/Appwrite/Platform/Modules/VCS/Http/GitHub/Callback/Get.php
```diff
@@ -119,13 +118,13 @@ public function action(
...
             if (!empty($code)) {
-                $oauth2 = new OAuth2Github(System::getEnv('_APP_VCS_GITHUB_CLIENT_ID', ''), System::getEnv('_APP_VCS_GITHUB_CLIENT_SECRET', ''), "");
+                $oauth2 = $provider->createOAuth2();
 
                 $accessToken = $oauth2->getAccessToken($code);
                 $refreshToken = $oauth2->getRefreshToken($code);
                 $accessTokenExpiry = DateTime::addSeconds(new \DateTime(), \intval($oauth2->getAccessTokenExpiry($code)));
 
-                $personalSlug = $oauth2->getUserSlug($accessToken);
+                $personalSlug = \method_exists($oauth2, 'getUserSlug') ? $oauth2->getUserSlug($accessToken) : '';
```
Meldiron, line 127: Should always exist, we can define it as requried on VCS

app/init/constants.php
HarshMN2345: I added this because if this PR gets merged and cloud is not updated yet, it could cause issues.

HarshMN2345:
Superseded by #12820 (design reworked: Provider/Manager replaced with Factory + InstallationTokens + RepositoryWebhooks).

PR #12793, opened 2026-07-06
feat: add Gitea VCS provider (OAuth2, additive only)

HarshMN2345:
## Summary
Part 2/3 of splitting up #12774. Stacked on #12792 (adapter abstraction) — base branch will move to `main` once that merges.

Enables Gitea as a second VCS provider on top of the adapter-agnostic registry from #12792. **Purely additive — does not modify any existing GitHub file.**

- `Auth/OAuth2/Gitea.php`: OAuth2 code-flow adapter (Gitea uses a plain OAuth2 app, not a GitHub-App-style installation).
- `VCS/Http/{Authorize,Callback}/Base.php`: generic OAuth2 authorize/callback actions, parameterized by provider; `VCS/Http/Gitea/*` are thin subclasses registering `/v1/vcs/gitea/{authorize,callback,events}`.
- Per-repository webhook creation wired into the four repo-connect sites (Functions/Sites Create/Update) via `Resolver::ensureRepositoryWebhook` — a no-op for GitHub, since app-based providers deliver events without it.
- V25 migration + collection schema bump `personalAccessToken`/`personalRefreshToken` 256→2048 chars.
- New env vars (`_APP_VCS_GITEA_*`, `_APP_VCS_WEBHOOK_URL`), mock installation endpoint for tests.

### Why the token-size migration is necessary
Verified against a **live Gitea instance**, not assumed: registered a real OAuth2 app, ran the full authorize → grant → token exchange, and measured the actual issued token. Gitea OAuth2 access/refresh tokens are RS256-signed JWTs — **858 characters** in the measured case, more than 3x the old 256-char column limit. Without this migration, every Gitea OAuth2 installation's stored token would be silently truncated on write, breaking the integration on the very first authenticated call. Also confirmed the vendored `appwrite/server-ce` package in the cloud fork still has these columns at 256, so there's no existing wider column to rely on instead.

## Stack
- #1 (#12792) → adapter abstraction, GitHub-only
- **#2 (this PR)** → Gitea provider (additive)
- #3 → Gitea CI/dev environment + live E2E test

## Test plan
- [x] `composer lint` / `composer format` clean
- [x] `composer analyze` (PHPStan level 4) clean
- [x] Manually verified end-to-end against a live Gitea container: installation → repo connect → webhook creation → clone → build → execute → webhook-triggered redeploy → execute, all via real (non-mocked) API calls


HarshMN2345:
Superseded by the #12820/#12826 stack — will be replaced by a new Gitea PR built on the reworked design.

PR #12820, opened 2026-07-08
feat: add Gitea VCS provider via adapter-driven Vcs\Factory

HarshMN2345:
Replaces the hardcoded `gitHub` DI resource with `Appwrite\Vcs\Factory`, a provider registry, and DI wiring — adapter construction only, no database access, no auth-type branching. GitHub-only, no behavior change. Supersedes #12792.

Actions inject `vcsForInstallation`/`vcsForProvider` closures instead of the factory directly, matching the existing `getProjectDB` pattern.


HarshMN2345:
@greptile We’re intentionally leaving the `utopia-php/vcs` composer constraint as-is on this branch for now.

HarshMN2345:
@greptile ignore composer version

HarshMN2345:
@greptile re-review

composer.json
```diff
@@ -94,7 +94,7 @@
         "utopia-php/storage": "2.*",
         "utopia-php/system": "0.10.*",
         "utopia-php/telemetry": "^0.4",
-        "utopia-php/vcs": "^4.4.1",
+        "utopia-php/vcs": "dev-feat/appwrite-provider-parity as 4.99.99",
```
loks0n: leftover

HarshMN2345: done

HarshMN2345:
@greptile our specific claim doesn't hold: $providerInstallationId is only ever passed into getOwnerName($providerInstallationId) with no second argument, and Gitea's adapter implementation of that method completely ignores the installationId parameter whenever repositoryId is null it falls back to getAuthenticatedUserLogin() instead. So the "wrong value" you flagged is never actually read.

That said, there's a real bug adjacent to it: without passing repositoryId, Gitea has no way to look up the actual repo owner, so it guesses "whoever's token this is"correct for a personal repo, wrong for an org-owned one the user has push access to but doesn't personally own. Fixed by passing providerRepositoryId through as the second argument at all 3 getOwnerName() call sites in Deployment.php. GitHub's adapter explicitly ignores that param, so it's a no-op there.

app/config/collections/platform.php
```diff
@@ -1842,7 +1842,7 @@
                 '$id' => ID::custom('personalAccessToken'),
                 'type' => Database::VAR_STRING,
                 'format' => '',
-                'size' => 256,
+                'size' => 2048,
```
Meldiron, line 1845: Lets make sure we have task for Cloud to migrate production

HarshMN2345: It's done

app/config/vcs.php
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * VCS provider registry, read by Appwrite\Vcs\Factory.
+ */
+
+use Utopia\VCS\Adapter\Git\Gitea;
+use Utopia\VCS\Adapter\Git\GitHub;
+
+return [
+    'github' => [
+        'enabled' => true,
```
Meldiron: I dont think we need enabled toggle here - anything thats here is considered supported (as long as variables are set correctly)

app/config/vcs.php
```diff
@@ -0,0 +1,37 @@
...
+        'adapter' => GitHub::class,
+        'envVariables' => [
+            'APP_NAME' => '_APP_VCS_GITHUB_APP_NAME',
+            'PRIVATE_KEY' => '_APP_VCS_GITHUB_PRIVATE_KEY',
+            'APP_ID' => '_APP_VCS_GITHUB_APP_ID',
+            'CLIENT_ID' => '_APP_VCS_GITHUB_CLIENT_ID',
+            'CLIENT_SECRET' => '_APP_VCS_GITHUB_CLIENT_SECRET',
+            'WEBHOOK_SECRET' => '_APP_VCS_GITHUB_WEBHOOK_SECRET',
+        ],
+        'requiredEnvVariables' => ['APP_NAME', 'PRIVATE_KEY', 'APP_ID', 'CLIENT_ID', 'CLIENT_SECRET'],
```
Meldiron: I dont like this 🤔 We are creating env var name alaises. dont think thats good.

Meldiron: Lets say we need private key. I would invite this structure much more:

```
'github' => [
  'privateKey' => [
		'required' => true,
		'envVariable' => _APP_VCS_GITHUB_PRIVATE_KEY
  ]
]
```

Instead of focusing on env variables and making weird aliases, focus on the purpose of the value, and threat variable just as a pointer where to get it.

app/controllers/mock.php
```diff
@@ -294,6 +295,62 @@
         ]);
     });
 
+Http::get('/v1/mock/gitea/callback')
```
Meldiron, line 298: Why do we need mock endpoint?

app/init/constants.php
```diff
@@ -108,6 +108,9 @@
 const APP_VCS_GITHUB_USERNAME = 'Appwrite';
 const APP_VCS_GITHUB_EMAIL = 'team@appwrite.io';
 const APP_VCS_GITHUB_URL = 'https://github.com/TeamAppwrite';
+const APP_VCS_COMMIT_USERNAME = APP_VCS_GITHUB_USERNAME;
+const APP_VCS_COMMIT_EMAIL = APP_VCS_GITHUB_EMAIL;
+const APP_VCS_COMMIT_URL = APP_VCS_GITHUB_URL;
```
Meldiron: Why not just rename the const and replace where it was used?

Unless we want to keep it separated, which makes sense, for example for APP_VCS_GITHUB_URL.. But then we shouldnt introduce generic `APP_VCS_COMMIT_URL`, instead, have separate consts for each provider, and use them in config/vcs.php

src/Appwrite/Auth/OAuth2/Gitea.php
```diff
@@ -0,0 +1,154 @@
...
+
+    protected array $tokens = [];
+
+    protected array $scopes = [
+        'read:user',
+        'read:repository',
+        'write:repository',
+        'read:organization',
+        'read:issue',
+        'write:issue',
```
Meldiron: I dont think we should require all of this by default.
Maybe we can for console installation, but not force it for adapter. Because this way, any developer who wants to have sign in with gitea button automatially wants this exact set of scopes, which is not true

src/Appwrite/Platform/Modules/VCS/Http/Authorize/Base.php
```diff
@@ -0,0 +1,113 @@
...
+        $callback = $protocol . '://' . $hostname . '/v1/vcs/' . $key . '/callback';
+
+        // The callback endpoint is public, so it verifies this signature
+        // before trusting the projectId and redirect URLs in state.
+        $oauth2 = $this->createOAuth2($callback, [
+            'projectId' => $project->getId(),
+            'success' => $success,
+            'failure' => $failure,
+            'signature' => \hash_hmac('sha256', \json_encode([$project->getId(), $success, $failure]), System::getEnv('_APP_OPENSSL_KEY_V1', '')),
+        ]);
```
Meldiron, line 106: Here we should specify scopes we need

composer.json
```diff
@@ -94,7 +94,7 @@
         "utopia-php/storage": "2.*",
         "utopia-php/system": "0.10.*",
         "utopia-php/telemetry": "^0.4",
-        "utopia-php/vcs": "^4.4.1",
+        "utopia-php/vcs": "4.4.2",
```
Meldiron: 4.*.* might be best version number so it auto updates

app/config/vcs.php
HarshMN2345: hmm makes sense

app/controllers/mock.php
HarshMN2345: Same reason as the existing mock/github/callback. The OAuth callback requires an interactive browser login/consent flow, which we can’t automate in E2E. The test gets a real Gitea access token through Gitea’s API, then uses this endpoint to create the installation record as if the OAuth flow had completed.

app/init/constants.php
HarshMN2345: hmm, I checked and only APP_VCS_COMMIT_EMAIL is actually used, while the username and URL aliases aren’t used anywhere.

src/Appwrite/Platform/Modules/VCS/Http/Authorize/Base.php
HarshMN2345: Fixed

Meldiron (approved):

Meldiron (approved):

PR #12826, opened 2026-07-09
refactor: extract InstallationTokens and RepositoryWebhooks services

HarshMN2345:
Extracts OAuth2 token refresh/persist and idempotent repository webhook creation into `Appwrite\Vcs\InstallationTokens` and `Appwrite\Vcs\RepositoryWebhooks`. Stacked on #12820.


src/Appwrite/Vcs/InstallationTokens.php
```diff
@@ -0,0 +1,97 @@
...
+
+namespace Appwrite\Vcs;
+
+use Appwrite\Auth\OAuth2;
+use Appwrite\Extend\Exception;
+use Utopia\Database\Database;
+use Utopia\Database\DateTime;
+use Utopia\Database\Document;
+
+class InstallationTokens
```
loks0n, line 11: maybe we can make make this class just a extension of document?, Installation extends Document

seems like a cleaner model to me

src/Appwrite/Vcs/RepositoryWebhooks.php
```diff
@@ -0,0 +1,77 @@
...
+            return;
+        }
+
+        $provider = $installation->getAttribute('provider', '');
+        if (empty($provider)) {
+            throw new Exception(Exception::GENERAL_ARGUMENT_INVALID, 'Missing VCS provider for installation: ' . $installation->getId());
+        }
+
+        $url = $this->buildWebhookUrl($provider);
+        $secret = $this->vcsFactory->getWebhookSecret($provider);
```
loks0n, line 53: why does the vcsFactory return something that isn't a vcs instance?

$this->vcsFactory->fromProvider($provider)->getWebhookSecret() is acceptable

src/Appwrite/Vcs/InstallationTokens.php
HarshMN2345: I thought about that too, but getDocument() always gives us a plain Document, so we’d have to wrap it everywhere we fetch an installation.

PR #12831, opened 2026-07-09
feat: add Gitea VCS provider

HarshMN2345:
Adds Gitea as the first additional VCS provider (OAuth2-based) on top of the Factory/InstallationTokens/RepositoryWebhooks foundation from #12820/#12826: registry entry, OAuth2 client, authorize/callback/webhook routes, per-repository webhook wiring on connect, and the token-column migration. Stacked on #12826.

The Gitea webhook handler resolves both a provider-level adapter (signature validation) and per-repository installation adapters, via the same closure injection introduced in #12820.


app/config/collections/platform.php
```diff
@@ -1842,7 +1842,7 @@
                 '$id' => ID::custom('personalAccessToken'),
                 'type' => Database::VAR_STRING,
                 'format' => '',
-                'size' => 256,
```
loks0n, line 1845: reminder: we probably need a cloud migration for this right?

app/init/resources.php
```diff
@@ -384,6 +384,8 @@ function getDevice(string $root, string $connection = ''): Device
 
 $container->set('vcsProviders', fn (VcsFactory $vcsFactory) => fn () => $vcsFactory->getProviders(), ['vcsFactory']);
 
+$container->set('vcsConfigured', fn (VcsFactory $vcsFactory) => fn (string $provider) => $vcsFactory->isConfigured($provider), ['vcsFactory']);
```
loks0n, line 387: this doesn't make sense to me on a couple levels.

a vcs factory class should only generate instances of vcs. a bool is not an instance of vcs.

just return vcs, and add vcs->isConfigured()

app/config/collections/platform.php
HarshMN2345: yes, we need to

HarshMN2345: done

app/init/resources.php
HarshMN2345: The reason I kept it on the factory is that isConfigured() isn’t checking the adapter itself, it’s checking Appwrite’s config/env registry. Moving it onto the adapter would mean pushing Appwrite-specific config into the VCS library.

src/Appwrite/Auth/OAuth2/Gitea.php
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    protected function getTokens(string $code): array
+    {
+        if (empty($this->tokens)) {
+            $response = $this->request(
+                'POST',
+                $this->endpoint . '/login/oauth/access_token',
+                ['Content-Type: application/json'],
+                \json_encode([
```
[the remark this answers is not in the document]
HarshMN2345: @greptile I tested this against the Gitea version we actually use (1.22.6 in our dev/CI container). Even without sending an Accept header, /login/oauth/access_token still returns JSON, so I couldn’t reproduce the issue. Also, GitHub isn’t really the same here—it doesn’t send an Accept header either, but it parses a form-encoded response because that’s GitHub’s default behavior.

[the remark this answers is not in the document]
HarshMN2345: @greptile re-review

PR #12833, opened 2026-07-09
test: add Gitea CI/dev environment and E2E test

HarshMN2345:
Ports forward the Gitea CI/dev-environment work from the previous (closed) PR stack — gitea/gitea-setup docker-compose services behind a `vcs` profile, CI wiring for the VCS matrix entry, and a live E2E test that installs via OAuth2, connects a function, asserts a real repository webhook was created, triggers a deployment, then pushes a second commit through Gitea's actual API and asserts the webhook-triggered rebuild happens. None of this depended on the classes that changed since the original attempt, so it carries over almost unmodified. Stacked on #12831.


.github/workflows/ci.yml
```diff
@@ -454,22 +454,34 @@ jobs:
...
+            echo "_APP_VCS_GITEA_ENDPOINT=http://gitea:3000" >> $GITHUB_ENV
+            echo "_APP_VCS_GITEA_BROWSER_ENDPOINT=http://localhost:9515" >> $GITHUB_ENV
+            echo "_APP_VCS_GITEA_CLIENT_ID=appwrite-ci" >> $GITHUB_ENV
+            echo "_APP_VCS_GITEA_CLIENT_SECRET=appwrite-ci-secret" >> $GITHUB_ENV
+            echo "_APP_VCS_GITEA_WEBHOOK_SECRET=appwrite-vcs-secret" >> $GITHUB_ENV
+            echo "_APP_VCS_WEBHOOK_URL=http://appwrite/v1" >> $GITHUB_ENV
+          fi
+
+          echo "COMPOSE_PROFILES=${COMPOSE_PROFILES_VALUE}" >> $GITHUB_ENV
+
```
loks0n: don't have this random per service override, let's just set the default .env to have these..?

dev/gitea/setup.sh
```diff
@@ -0,0 +1,22 @@
...
+
+su-exec git gitea admin user create \
+  --admin \
+  --username "${ADMIN_USER}" \
+  --password "${ADMIN_PASSWORD}" \
+  --email "${ADMIN_EMAIL}" \
+  --must-change-password=false \
+  --config /data/gitea/conf/app.ini || true
+
+echo "Gitea test fixture is ready"
```
loks0n, line 22: can this be done via docker/image/gitea file config instead?

docker-compose.override.yml
```diff
@@ -274,6 +276,60 @@ services:
     volumes:
       - ./app:/usr/src/code/app
       - ./src:/usr/src/code/src
+  gitea:
```
loks0n, line 291: why not just add to the default compose?

dev/gitea/setup.sh
loks0n: do we even still need this? I see we have envs for it

docker-compose.override.yml
HarshMN2345: Kept it here since Gitea is only needed for CI/dev testing right now, so groupedc it with the other test only services (maildev, request-catcher).

dev/gitea/setup.sh
HarshMN2345: Tried thist with Gitea image version (1.22.6) doesn't actually support admin bootstrap via env vars. Verified on a clean volume: no admin gets created, and the image's own setup script has zero references to GITEA_ADMIN_*. So yes, still needed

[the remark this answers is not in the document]
HarshMN2345: @greptile CI/E2E goes through /mock/gitea/callback and never runs the real OAuth2 flow, so no app registration is needed for what this environment exists for. The suggested fix also isn't possible as described: Gitea's API generates client_id/client_secret, you can't register an app with pre-chosen values making these baked-in creds real would mean writing rows into Gitea's sqlite directly, which we tried and reverted as too fragile.

docker-compose.override.yml
```diff
@@ -286,6 +288,60 @@ services:
...
+      - GITEA_ADMIN_USER=${_TESTS_GITEA_ADMIN_USER:-appwrite}
+      - GITEA_ADMIN_PASSWORD=${_TESTS_GITEA_ADMIN_PASSWORD:-password}
+      - GITEA_ADMIN_EMAIL=appwrite@localhost.test
+    volumes:
+      - appwrite-gitea:/data
+      - ./dev/gitea/setup.sh:/setup.sh:ro
+    networks:
+      - appwrite
+    command:
+      ["/bin/sh", "-c", "/setup.sh && touch /tmp/ready && tail -f /dev/null"]
```
loks0n: hmm does this break sigterm?

.github/workflows/ci.yml
```diff
@@ -453,23 +453,35 @@ jobs:
         run: |
           echo "_APP_OPTIONS_ROUTER_PROTECTION=enabled" >> $GITHUB_ENV
 
+          COMPOSE_PROFILES_VALUE=""
+
           if [ "${{ matrix.database }}" = "MariaDB" ]; then
-            echo "COMPOSE_PROFILES=mariadb" >> $GITHUB_ENV
+            COMPOSE_PROFILES_VALUE="mariadb"
```
loks0n: why rename?

.github/workflows/ci.yml
```diff
@@ -453,23 +453,35 @@ jobs:
...
           fi
 
+          if [ "${{ matrix.service.name }}" = "VCS" ]; then
+            if [ -z "$COMPOSE_PROFILES_VALUE" ]; then
+              COMPOSE_PROFILES_VALUE="vcs"
+            else
+              COMPOSE_PROFILES_VALUE="${COMPOSE_PROFILES_VALUE},vcs"
+            fi
+          fi
+
```
loks0n: We should remove this block, it's unneccessary configuration, VCS shouldn't require an exceptional case to test it.

let's just load the override or whatever to make it work?

docker-compose.override.yml
HarshMN2345: yeah, you are right, fixing

.github/workflows/ci.yml
HarshMN2345: we can keep it as it is 👍 , not much need of a rename

PR #12840, opened 2026-07-10
fix: filter Sites build runtimes by _APP_SITES_RUNTIMES

ChiragAgg5k:
## What does this PR do?

Self-hosted Sites settings showed **all** build runtimes (including Bun/Deno/uninstalled Node versions) in the Console dropdown, while Functions correctly filters by `_APP_FUNCTIONS_RUNTIMES`.

This mirrors the Functions `listRuntimes` behavior for Sites:

- `GET /v1/sites/frameworks` now filters each framework's `runtimes` list against `_APP_SITES_RUNTIMES`
- Site create/update reject a `buildRuntime` that is not in the allowlist (when the allowlist is set)

Reported in community threads: https://appwrite.io/threads/1525087225895718983

## Test Plan

1. Set `_APP_SITES_RUNTIMES=static-1,node-22,node-24` (leave Bun/Deno out)
2. Restart Appwrite
3. Open a Site → Settings → Runtime settings
4. Confirm the build runtime dropdown only lists runtimes from the allowlist (same idea as Functions)
5. Attempt create/update with a runtime outside the allowlist via API → expect `general_argument_invalid`

## Related PRs and Issues

- Thread: https://appwrite.io/threads/1525087225895718983
- Related Greptile note on #12337 about `_APP_SITES_RUNTIMES` not being respected by frameworks config

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Modules/Sites/Http/Frameworks/XList.php
+++ b/src/Appwrite/Platform/Modules/Sites/Http/Frameworks/XList.php
@@ -11,6 +11,7 @@ use Utopia\Config\Config;
 use Utopia\Database\Document;
 use Utopia\Platform\Action;
 use Utopia\Platform\Scope\HTTP;
+use Utopia\System\System;
 
 class XList extends Base
 {
@@ -52,10 +53,18 @@ class XList extends Base
     public function action(Response $response)
     {
         $frameworks = Config::getParam('frameworks');
+        $allowList = \array_filter(\array_map('trim', \explode(',', System::getEnv('_APP_SITES_RUNTIMES', ''))));
 
         foreach ($frameworks as $key => $framework) {
             if (!empty($framework['adapters'])) {
                 $frameworks[$key]['adapters'] = \array_values($framework['adapters']);
+            }
+
+            if (!empty($allowList) && !empty($framework['runtimes'])) {
+                $frameworks[$key]['runtimes'] = \array_values(\array_filter(
+                    $framework['runtimes'],
+                    fn ($runtime) => \in_array($runtime, $allowList, true)
+                ));
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/Appwrite/Platform/Modules/Sites/Http/Frameworks/XList.php
```diff
@@ -52,11 +53,19 @@ public function __construct()
...
             if (!empty($framework['adapters'])) {
                 $frameworks[$key]['adapters'] = \array_values($framework['adapters']);
             }
+
+            if (!empty($allowList) && !empty($framework['runtimes'])) {
+                $frameworks[$key]['runtimes'] = \array_values(\array_filter(
+                    $framework['runtimes'],
+                    fn ($runtime) => \in_array($runtime, $allowList, true)
+                ));
+            }
```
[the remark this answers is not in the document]
ChiragAgg5k: Good catch — addressed in the follow-up commit. Frameworks whose runtimes are all filtered out by `_APP_SITES_RUNTIMES` are now omitted from the `listFrameworks` response.

src/Appwrite/Platform/Modules/Sites/Http/Sites/Create.php
```diff
@@ -144,6 +144,12 @@ public function action(
             }
         }
 
+        $allowList = \array_filter(\array_map('trim', \explode(',', System::getEnv('_APP_SITES_RUNTIMES', ''))));
+
+        if (!empty($allowList) && !\in_array($buildRuntime, $allowList, true)) {
+            throw new Exception(Exception::GENERAL_ARGUMENT_INVALID, 'Runtime "' . $buildRuntime . '" is not supported');
```
Meldiron, line 150: We might want similar throw in builds worker, maybe even general.php router()

Meldiron (approved):

src/Appwrite/Platform/Modules/Sites/Http/Sites/Create.php
ChiragAgg5k: Good point — I looked into this. Functions also only enforces `_APP_FUNCTIONS_RUNTIMES` at Create + list (not in the Builds worker or `general.php` router for v5); those paths just resolve against the full `runtimes` catalog. Sites now mirrors that at Create/Update + frameworks list, and the executor still gates via `OPR_EXECUTOR_RUNTIMES`. Happy to add worker/router allowlist checks for both Functions and Sites as a follow-up if we want that extra defense-in-depth.

Issue #11393, opened 2026-02-24
No (GUI?) method for deleting (another) user permanently from entire/all AppWrite system/environments.


BloodyIron:
### 👟 Reproduction steps

This is self-hosted btw.

I am the root user, the first created user.

I invite someone to a new Org, they are then part of the Org, and have an account in the AppWrite now too.

I want to delete their account fully, but via GUI I have no means to do that. I can only remove from the Org, and then that user can go and create their own Orgs at-will. This leads to shadow accounts running around in our own system which the only GUI way to do this out of the box is for these shadow accounts to delete themselves.

### 👍 Expected behavior

I need a way to manage _ALL_ users across the whole AppWrite system, their permissions levels, and be able to fully delete an account permanently in the whole system. These are all common security access controls that are very much missing here.

### 👎 Actual Behavior

Already described

### 🎲 Appwrite version

Version 1.8.x

### 💻 Operating system

Linux

### 🧱 Your Environment

Ubuntu Server 24.04 with Docker Compose

### 👀 Have you spent some time to check if this issue has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

schneidermr:
I've opened PRs to address this:

- **GUI (primary):** [appwrite/console#3117](https://github.com/appwrite/console/pull/3117) — self-hosted Platform users page + permanent delete when removing org members
- **API guards:** [appwrite/appwrite#12850](https://github.com/appwrite/appwrite/pull/12850) — prevent self-delete via Users API and deleting the last console user

**Root cause:** Removing a member only deletes the membership. The console account remains and can create new orgs. Permanent deletion already works via `DELETE /v1/users/{id}` on project `console` with an organization context; it was just not exposed in the GUI.

PR #12850, opened 2026-07-11
fix(users): guard permanent console user deletion (#11393)

schneidermr:
## Summary

Related to #11393 (GUI: [appwrite/console#3117](https://github.com/appwrite/console/pull/3117)).

When self-hosted admins permanently delete console (platform) accounts via `DELETE /v1/users/{userId}` with organization context, add safety guards:

1. **No self-delete via Users API** — session actors cannot delete their own user id (must use `DELETE /v1/account`)
2. **No last console user** — refuse deleting the only remaining account on the console project (would lock out the instance)

These guard the permanent-delete flow exposed in the console UI for self-hosted instances.

## Test plan

- [x] PHP syntax check on `app/controllers/api/users.php`
- [x] Local 1.9.5 verification of existing delete path (org-scoped admin → 204)
- [x] E2E test added: `UsersConsoleClientTest::testConsoleUserPermanentDeleteGuards` (happy path delete + 404 after)
- [ ] CI e2e suite for Users console client
- [ ] Verify self-delete via Users API returns `user_deletion_prohibited`
- [ ] Verify deleting last console user returns `user_deletion_prohibited`

## Notes

- Project-scoped Auth user deletion is unchanged for non-console projects (only self-delete actor check applies)
- Membership cleanup remains async via the deletes worker (same model as `DELETE /v1/account` for console)

### DIFF
--- a/app/controllers/api/users.php
+++ b/app/controllers/api/users.php
@@ -2661,7 +2661,7 @@ Http::delete('/v1/users/:userId')
                     'users',
                     [
                         Query::equal('status', [true]),
-                        Query::notEqual('$id', [$target->getId()]),
+                        Query::notEqual('$id', $target->getId()),
                     ],
                     1
                 );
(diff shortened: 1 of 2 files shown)


HarshMN2345:

PR #12894, opened 2026-07-16
feat: add GitLab VCS provider 

HarshMN2345:
## Summary
Adds GitLab as a third VCS provider, following the exact adapter-driven `Vcs\Factory` pattern established for Gitea (#12820):

- **Registry entry** (`app/config/vcs.php`): GitLab's `utopia-php/vcs` adapter was already vendored (`Utopia\VCS\Adapter\Git\GitLab`) with full method parity to GitHub/Gitea. Endpoint is a fixed literal (`https://gitlab.com`) -- only official GitLab is supported for the VCS flow, no self-hosted configurability, no `_APP_VCS_GITLAB_ENDPOINT`/`_APP_VCS_GITLAB_BROWSER_ENDPOINT` env vars.
- **OAuth2 flow** (`VCS/Http/Gitlab/{Authorize,Callback,Events}`): per-repo webhooks, personal-install OAuth2, token refresh before webhook processing, error aggregation (adapter-resolution failures return a non-2xx response so GitLab surfaces it in its redelivery log rather than dropping silently).
- **Reuses `Auth\OAuth2\Gitlab`** (the existing "sign in with GitLab" account-login class) for the VCS flow too, rather than duplicating a class -- its JSON-encoded `appSecret` (`{"clientSecret":"...","endpoint":"..."}`) already carries an endpoint, which the account-login feature also depends on for self-hosted GitLab logins. The VCS flow always encodes `endpoint: "https://gitlab.com"` into that same JSON shape rather than changing the class's contract, so the pre-existing account-login feature keeps working unchanged.
- **Replaced per-provider OAuth2 conditionals with registry-driven polymorphism**: `Create.php`'s `createOAuth2()` and `InstallationTokens::refreshForInstallation()` each had a match/if chain constructing the right OAuth2 client per provider. Each OAuth2 client now implements a new `EnvOAuth2` interface (`fromEnv()`) and is registered in `vcs.php` alongside its adapter; `Factory::oauth2FromProvider()` resolves it generically. Adding a provider is now a registry entry + an interface implementation, not a new branch in two call sites.
- **`InstallationTokens::refreshForInstallation()`**: centralizes the per-provider OAuth2 construction needed to refresh an installation's token, replacing what used to be duplicated at each call site (5 `Repositories/*` endpoints + the Gitea/GitLab event handlers).
- Env vars, `_APP_VCS_GITLAB_EMAIL` constant, routes registered. The Gitea/GitLab OAuth mock callback endpoints were removed -- both providers are self-hostable and already run as real containers in CI, so E2E tests can exercise the actual `/v1/vcs/{provider}/callback` endpoints directly instead of a hand-written substitute.

## Upstream dependency -- resolved
`utopia-php/vcs`'s `GitLab::getEvent()` (added in a prior PR) didn't match Gitea's/GitHub's field/action-name contract, and three further real bugs surfaced during live integration testing against a real GitLab OAuth app:

- Missing `repositoryId` and non-standard key names in `getEvent()`'s push/MR payload shape.
- Merge request `action` passed through raw GitLab values instead of the `opened`/`reopened`/`synchronize`/`closed` convention consumers branch on.
- **Wrong auth header**: the adapter authenticated every request with `PRIVATE-TOKEN` (Personal Access Token style), but our OAuth2 access tokens require `Authorization: Bearer`. Every API call would have failed with 401 regardless of token freshness.
- `searchRepositories()` dropped private repos in personal namespaces, and returned a bare array instead of the `items`/`total` shape GitHub/Gitea return (which Appwrite's `Repositories/XList.php` already destructures unconditionally).

All of the above is fixed upstream in [utopia-php/vcs#120](https://github.com/utopia-php/vcs/pull/120), released as `4.4.3`, and `composer.lock` in this PR is bumped to it.

## Test plan
- [x] Full VCS unit suite passes, including new coverage for `oauth2FromProvider()`, the fixed-endpoint behavior, and cross-provider PR-close isolation
- [x] Full unit suite passes (930 tests)
- [x] `composer format` / `composer analyze` / `composer refactor:check` clean
- [x] **Verified live end-to-end against a real GitLab account and repository**: OAuth connect flow, repository picker, framework detection, deploy, and repository webhook creation (`POST` to `/v1/vcs/gitlab/events`) all confirmed working through a local Appwrite instance exposed via a public tunnel
- [x] Confirmed the pre-existing "Sign in with GitLab" account-login OAuth2 flow is unaffected (its own unit suite passes; the shared `OAuth2\Gitlab` class's JSON secret contract is preserved)


app/config/variables.php
```diff
@@ -1311,6 +1311,51 @@
...
+                'name' => '_APP_VCS_GITLAB_ENDPOINT',
+                'description' => 'URL of your self-hosted GitLab instance, reachable from the Appwrite server (e.g. for API calls and webhook token exchange). Defaults to https://gitlab.com when unset.',
+                'introduction' => '2.0.0',
+                'default' => '',
+                'required' => false,
+                'question' => '',
+                'filter' => ''
+            ],
+            [
+                'name' => '_APP_VCS_GITLAB_BROWSER_ENDPOINT',
```
Meldiron: I dont think we need these variables - we point ot to official gitlab which is always the same URLs here

src/Appwrite/Platform/Modules/VCS/Http/Installations/Repositories/Create.php
```diff
@@ -149,6 +150,13 @@ protected function createOAuth2(string $provider): OAuth2Github|OAuth2Gitea
             return $oauth2;
         }
 
+        if ($provider === 'gitlab') {
```
Meldiron: Hmm not too happy with this method having such if/else. Very easy to forget to update in future with new proviers.

Anything that is provider-specific should live in Factory or config. Lets check with Luke regarding this specifically, and see what he suggests as best quality here

src/Appwrite/Vcs/Factory.php
```diff
@@ -57,7 +57,7 @@ public function fromProvider(string $key): Git
 
         $adapter = new ($this->registry[$key]['adapter'])($this->cache);
 
-        $endpoint = $this->getEnv($key, 'endpoint');
+        $endpoint = $this->getEnv($key, 'endpoint') ?: ($this->registry[$key]['variables']['endpoint']['default'] ?? '');
```
Meldiron: Lets remove default support, no need

HarshMN2345:
@greptileai review

Meldiron (approved):

PR #12895, opened 2026-07-16
DISCORD-15272384: Harden flaky RealtimeCustomClientTest::testChannelExecutions

claudear:
## What does this PR do?

Fixes the flaky `Tests / CE / E2E / Realtime (dedicated) → RealtimeCustomClientTest::testChannelExecutions`, which intermittently fails with:

```
WebSocket\TimeoutException: Client read timeout
```

### Root cause

The test subscribes to the **project-scoped** `executions` channel and then makes two bare `receive()` calls, assuming the next two frames are exactly the execution `create` and `update` events, each bounded by a single 10s read timeout.

The async execution's terminal `update` only arrives after worker pickup, runtime cold start, and the function's execution timeout (the `timeout` fixture sleeps 60s against a 10s function timeout), so it regularly exceeds a single 10s read window → `Client read timeout`. The bare receives are also brittle to any intermediate or unrelated frames on the shared channel.

### Fix

- Add a reusable `receiveUntilEvent()` helper to the `RealtimeBase` trait (mirroring `PresenceRealtimeClientTest::receiveUntil`) that polls, tolerates per-read timeouts, skips unrelated frames, and waits generously (default 60s) for a matching frame.
- Use it in `testChannelExecutions` to wait for the specific `create` and `update` events for the execution instead of assuming they are the next two frames.

The subsequent assertions are unchanged and are satisfied by the matched frames.

## Test Plan

- Test-only change. Exercised by the existing CE realtime e2e job (`RealtimeCustomClientTest::testChannelExecutions`).

## Checklist

- [x] Test-only change; no API metadata changes.

### DIFF
--- a/tests/e2e/Services/Realtime/RealtimeBase.php
+++ b/tests/e2e/Services/Realtime/RealtimeBase.php
@@ -2,6 +2,7 @@
 
 namespace Tests\E2E\Services\Realtime;
 
+use Appwrite\Tests\Async\Exceptions\Critical;
 use WebSocket\Client as WebSocketClient;
 use WebSocket\ConnectionException;
 use WebSocket\TimeoutException;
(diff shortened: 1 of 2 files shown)


claudear:
## Fix Confidence: 80/100

n/a

tests/e2e/Services/Realtime/RealtimeBase.php
```diff
@@ -4,9 +4,52 @@
...
+        $deadline = \microtime(true) + ($timeoutMs / 1000);
+        $lastMessage = [];
+
+        while (\microtime(true) < $deadline) {
+            try {
+                $message = \json_decode($client->receive(), true);
+            } catch (TimeoutException) {
+                \usleep($pollMs * 1000);
+                continue;
+            }
```
ArnabChatterjee20k: @claudear 
use assertEventually here instead of usleep and also ConnectionException is not caught inside the loop. If the server closes or resets the WebSocket before the deadline (e.g. an idle-timeout eviction), the exception bubbles up unhandled and the PHPUnit output shows a raw stack trace instead of the descriptive "Timed out waiting for expected websocket frame" message. Catching it and calling $this->fail() with context keeps test failures readable and consistent with TimeoutException handling.

claudear:
## Fix Confidence: 88/100

The change directly implements both requested review points and follows existing patterns (assertEventually, Critical). php -l passes and the lightweight CI checks are green. Confidence is not higher because I could not run the full Realtime E2E suite or Pint locally (no appwrite container / composer in this worktree), so final CI validation of the flaky test is still pending.

ChiragAgg5k (approved):

PR #12900, opened 2026-07-17
(ci): drop php-retry, run tests directly without retries

ChiragAgg5k:
## What

Removes `itznotabug/php-retry` from CI entirely. All 5 test steps that used it now run their command directly via a plain `run:` step.

## Changes

- Replaced all 5 `itznotabug/php-retry` steps (Unit, E2E General, E2E per-service matrix, Abuse, Screenshots) with direct `docker compose exec ... test` / paratest runs.
- Dropped the `max_attempts: ${{ github.event_name == 'pull_request' && 2 || 1 }}` retry conditional along with the action's `retry_wait_seconds`, `timeout_minutes`, `job_id`, `github_token`, and `test_dir` inputs.
- The per-service matrix step now uses `timeout-minutes: ${{ matrix.service.timeout }}` at the step level, preserving the per-service timeout previously passed to the action.
- Removed `pull-requests: write` from the 5 test jobs where php-retry was the only consumer of that permission. The `benchmark` job keeps it — it posts its own PR comments.

Tests now fail fast on the first attempt instead of retrying on PRs.

Cloud counterpart: appwrite-labs/cloud#4843

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -296,7 +296,6 @@ jobs:
     needs: build
     permissions:
       contents: read
-      pull-requests: write
       packages: read
     steps:
       - name: checkout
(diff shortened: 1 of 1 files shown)


ChiragAgg5k:
Addressed: restored the 15-minute step-level timeout for the service matrix tests (matching the previous total wall-clock budget) and removed the now-unused `timeout` field from the matrix service definitions.

levivannoort (approved):

PR #12903, opened 2026-07-17
fix: change GitHub OAuth label from App ID to Client ID (issue #11875)

### LINKED ISSUE #11875: Change GitHub OAuth2 settings label from App ID to Client ID

Shresthap21:
## What changed
Changed the GitHub OAuth2 settings label from "App ID" to "Client ID" in the console UI.

## Why
GitHub calls this field "Client ID" in their own settings. The previous label "App ID" was inconsistent with GitHub's terminology and caused confusion for developers setting up OAuth.

## Related issue
Fixes #11875 

## How tested
Verified the label change visually in the console UI.

### DIFF
--- a/src/Appwrite/Platform/Modules/Project/Http/Project/OAuth2/Facebook/Update.php
+++ b/src/Appwrite/Platform/Modules/Project/Http/Project/OAuth2/Facebook/Update.php
@@ -35,7 +35,7 @@ class Update extends Base
 
     public static function getClientIdParamName(): string
     {
-        return 'appId';
+        return 'Client ID';
     }
 
     public static function getClientSecretParamName(): string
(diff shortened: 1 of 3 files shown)


ChiragAgg5k:
@Shresthap21 we show both as historically it was called app-id, its not that long ago since the rename to client-id happened. many workflows like the official github action workers as well support both appid and clientid (even if they show deprecated messages)

hence why we show both

PR #12922, opened 2026-07-20
Allow open-runtimes/sdk-for-php 0.7.x

loks0n:
Widens the `open-runtimes/sdk-for-php` constraint from `0.6.*` to `0.6.* || 0.7.*` (and bumps the lock to 0.7.0).

### Why
0.7.0 is **purely additive** over 0.6.0 — [the only changes](https://github.com/open-runtimes/sdk-for-php/compare/0.6.0...0.7.0) are:
- an `Lz4 = 'lz4'` case on `ArchiveCompression`
- a `blockSize` field on `ArchiveArtifact` (squashfs block size, serialized as `blockSize`)

Nothing was removed or changed, so server-ce keeps working unchanged. These mirror the orchestrator's squashfs archive options and let consumers request the lz4 / 1 MiB-block squashfs artifact the jobs sidecar produces.

### Context
Appwrite Cloud wants to move build-output compression into the jobs sidecar (`ArchiveArtifact(format: squashfs, compression: lz4)`), which needs `ArchiveCompression::Lz4` from 0.7.0. The `0.6.*` pin currently blocks that bump downstream.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -106,7 +106,7 @@
         "league/csv": "9.14.*",
         "enshrined/svg-sanitize": "0.22.*",
         "utopia-php/client": "^0.2.2",
-        "open-runtimes/sdk-for-php": "0.6.*"
+        "open-runtimes/sdk-for-php": "0.6.* || 0.7.*"
     },
     "require-dev": {
         "ext-fileinfo": "*",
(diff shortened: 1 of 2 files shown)


loks0n:
Folded into #12923, which carries this exact SDK-constraint change alongside the compression-agnostic test fix that Cloud's squashfs builds also need. Consolidated so Cloud can pin server-ce to a single branch.

PR #12924, opened 2026-07-21
feat: use utopia-php/user-agent for user-agent parsing

lohanidamodar:
## What

Replaces `matomo/device-detector` with [`utopia-php/user-agent`](https://github.com/utopia-php/user-agent) for user-agent parsing.

## How

`src/Appwrite/Detector/Detector.php` now wraps `Utopia\UserAgent\UserAgent` instead of matomo. Its public surface (`getOS()`, `getClient()`, `getDevice()`, `skipBotDetection()`) and 12-field array shape are unchanged, so the four call sites are untouched:

- `app/controllers/api/account.php` (sessions, email templates, push targets)
- `app/controllers/api/users.php` (server-side session create)
- `src/Appwrite/Platform/Modules/Teams/Http/Memberships/Status/Update.php`
- `src/Appwrite/Platform/Modules/Account/Http/Account/MFA/Challenges/Create.php`

The `AppwriteCLI` short-circuit and the `''`-vs-`null` default split (OS/client default to `''`, device fields to `null`) are preserved.

`matomo/device-detector` is removed from `composer.json`. It was used **only** through this wrapper — `isBot()` was never called anywhere in the repo, and no code touches the raw `DeviceDetector` class. `utopia-php/user-agent` requires PHP `>=8.4`, already satisfied by the repo's `>=8.5` floor, and pulls in no runtime dependencies.

## Behaviour

`utopia-php/user-agent` 0.1.1 keeps its short-name codes aligned with Matomo DeviceDetector, so parsing is **byte-identical** for the user agents these tests assert:

- Firefox 47 / Windows 7 — the `DetectorTest` unit fixture
- Chrome 70 / Windows 10 — the default e2e client UA (session `osCode`/`clientCode`/`clientVersion`/`clientEngine`/`deviceName` assertions)
- Chrome Mobile / Android — the `CM` / `Chrome Mobile` session assertions
- Session-create webhook payloads (all 12 fields)

The 12 UA fields are persisted columns on the `sessions` collection. New sessions parse via utopia; existing rows keep their matomo-vintage values. The raw `userAgent` string is retained on the session, so a backfill is possible if ever wanted, but is not part of this change.

The only observed divergence is on an untested path: for an `AppwriteCLI` user agent, `deviceName` becomes `null` instead of `'desktop'` (the CLI override only ever set client fields; device came from the parser). This affects CLI-created session rows only.

## Tests

`DetectorTest` gains the `AppwriteCLI` branch, a mobile client, and an unknown/empty user agent asserting the exact `''`-vs-`null` defaults. Ran locally: unit `DetectorTest` passes (6 tests), Pint clean, PHPStan level 4 clean on the Detector and all four callers. The Chrome 70 / Chrome Mobile e2e session assertions were verified to match utopia's output directly.


lohanidamodar:
CI note: the `Checks` job (Lint / Rector / PHPStan) passes. The two failing jobs — `Tests / E2E / PostgreSQL (dedicated) / Databases` and `.../FunctionsSchedule` — are unrelated to this change. `PostgreSQL / Databases` fails on `main` in the last three CI runs (e.g. [29767621640](https://github.com/appwrite/appwrite/actions/runs/29767621640)), and this PR touches only user-agent parsing (`Detector.php`, its test, and the matomo→utopia composer swap) — nothing in the Databases or FunctionsSchedule paths. Re-triggered the failed jobs.

loks0n (approved):

PR #12929, opened 2026-07-21
Restore storage folder API

TorstenDittmann:
## Summary
- restore storage folder endpoint and response models
- restore folder create-file parameter, file response fields, and file query support
- restore folder API e2e coverage
- keep V25 migration unchanged since the migration already ran

## Tests
- php -l touched PHP files
- git diff --check HEAD~1..HEAD


TorstenDittmann:
@greptileai review

TorstenDittmann:
@greptileai review

src/Appwrite/Platform/Modules/Storage/Http/Buckets/Folders/XList.php
```diff
@@ -0,0 +1,155 @@
...
+        $folders = [];
+        $latest = null;
+        $batch = 1000;
+
+        try {
+            do {
+                $queries = [Query::limit($batch)];
+                if ($latest !== null) {
+                    $queries[] = Query::cursorAfter($latest);
+                }
```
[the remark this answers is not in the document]
TorstenDittmann: The scan processes every document batch before returning. Keys at or before the client cursor are intentionally excluded regardless of scan order, while every key after it remains eligible. Sorting and truncating the accumulated candidates after each batch preserves the global smallest `limit` eligible keys, so the document scan order cannot cause a folder after the cursor to be skipped.

adityaoberai:
@TorstenDittmann would it make sense to add a new E2E test specifically using an S3 client?

TorstenDittmann:
> @TorstenDittmann would it make sense to add a new E2E test specifically using an S3 client?

those are covered on cloud 👍🏻

src/Appwrite/Platform/Modules/Storage/Http/Buckets/Folders/XList.php
```diff
@@ -0,0 +1,155 @@
...
+                    new SDKResponse(
+                        code: Response::STATUS_CODE_OK,
+                        model: Response::MODEL_FOLDER_LIST,
+                    )
+                ]
+            ))
+            ->param('bucketId', '', new UID(), 'Storage bucket unique ID. You can create a new storage bucket using the Storage service [server integration](https://appwrite.io/docs/server/storage#createBucket).')
+            ->param('folder', '', new Folder(), 'Virtual folder to list the child folders of. Defaults to the bucket root.', true)
+            ->param('limit', 25, new Range(1, 100), 'Maximum number of folders to return. Must be between 1 and 100. A page shorter than this limit means the listing is complete.', true)
+            ->param('cursor', '', new Folder(), 'Folder path returned last by a previous page, used to continue the listing.', true)
```
Meldiron: We should do our typical Queries syntax here. We have some endpoints that acepts queries, and only supports folder pagination. We can take inspuration.

Come to think of it, "folder" could also be supported by queries. Queries.equal('filder', X).
Check some existing endpoints that only have partial queries support. I recall I did something similar for apps or oauth2, and it worked nicely

src/Appwrite/Platform/Modules/Storage/Http/Buckets/Folders/XList.php
```diff
@@ -0,0 +1,155 @@
...
+            ->desc('List folders')
+            ->groups(['api', 'storage'])
+            ->label('scope', 'files.read')
+            ->label('usage.resource', 'bucket/{request.bucketId}')
+            ->label('resourceType', RESOURCE_TYPE_BUCKETS)
+            ->label('sdk', new Method(
+                namespace: 'storage',
+                group: 'files',
+                name: 'listFolders',
+                description: '/docs/references/storage/list-folders.md',
```
Meldiron: Missing md file. Also, lets not do md file, lets just have description directly here

src/Appwrite/Platform/Modules/Storage/Services/Http.php
```diff
@@ -41,5 +42,8 @@ public function __construct()
         $this->addAction(GetFileDownload::getName(), new GetFileDownload());
         $this->addAction(GetFileView::getName(), new GetFileView());
         $this->addAction(GetFileForPush::getName(), new GetFileForPush());
+
+        // Folders
+        $this->addAction(ListFolders::getName(), new ListFolders());
```
Meldiron: Should we have GET folder too? Any reason for Update folder, delete folder?

src/Appwrite/Utopia/Response/Model/File.php
```diff
@@ -127,6 +139,10 @@ public function filter(Document $document): Document
         $encryption = !empty($document->getAttribute('openSSLCipher', ''));
         $document->setAttribute('encryption', $encryption);
 
+        $folder = $document->getAttribute('folder') ?? '';
+        $document->setAttribute('folder', $folder);
```
Meldiron: Sounds useless

src/Appwrite/Platform/Modules/Storage/Services/Http.php
TorstenDittmann: we just follow what s3 is doing 👍🏻

TorstenDittmann: folders are virtual, dont really exist

src/Appwrite/Platform/Modules/Storage/Http/Buckets/Folders/XList.php
TorstenDittmann: its there from a previous pr 👍🏻

src/Appwrite/Platform/Modules/Storage/Services/Http.php
TorstenDittmann: Folders are derived from the folder paths of existing files rather than persisted resources. GET would not expose anything beyond what list already returns, while update or delete would implicitly mutate all files below a path and introduce permission and partial-failure semantics. I would keep this list-only for now and treat file operations as the folder lifecycle.

src/Appwrite/Platform/Modules/Storage/Http/Buckets/Folders/XList.php
```diff
@@ -0,0 +1,186 @@
...
+            ->param('queries', [], new Queries([
+                new Limit(100),
+                new Offset(),
+                new Filter([
+                    new Document([
+                        'key' => 'folder',
+                        'type' => Database::VAR_STRING,
+                        'array' => false,
+                    ]),
+                ], Database::VAR_STRING, 1),
```
Meldiron: I would expect cursor pagination to be also supported

src/Appwrite/Platform/Modules/Storage/Http/Buckets/Folders/XList.php
```diff
@@ -0,0 +1,186 @@
...
+        $offset = $grouped['offset'] ?? 0;
+        $pageEnd = $offset > PHP_INT_MAX - $limit ? PHP_INT_MAX : $offset + $limit;
+
+        $collection = 'bucket_' . $bucket->getSequence();
+        $folders = [];
+        $latest = null;
+        $batch = 1000;
+
+        try {
+            do {
```
Meldiron: This scares me. We loop inside endpoint - do we need to batch here? Usually our pattern is we accept params and run them once. Up to developer to do looping on their end to fetch multiple pages

src/Appwrite/Platform/Modules/Storage/Http/Buckets/Folders/XList.php
```diff
@@ -0,0 +1,186 @@
...
+                }
+
+                foreach ($results as $file) {
+                    $fileParent = $file->getAttribute('folder', '');
+                    if ($fileParent === $parent || !\str_starts_with($fileParent, $parent)) {
+                        continue;
+                    }
+
+                    $segment = \explode('/', \substr($fileParent, \strlen($parent)))[0];
+                    $key = $parent . $segment . '/';
```
Meldiron: I think we shouldnt filter here. I think we should utilize DB queries instead

src/Appwrite/Platform/Modules/Storage/Http/Buckets/Folders/XList.php
```diff
@@ -0,0 +1,186 @@
...
+
+                    $folders[$key] = new Document([
+                        'key' => $key,
+                        'name' => $segment,
+                        'parent' => $parent,
+                    ]);
+                }
+
+                \ksort($folders, \SORT_STRING);
```
Meldiron: I wouldnt force sort, we could accept queries to sort, query::orderDesc('folder') or query::orderAsc('$createdAt')

src/Appwrite/Platform/Modules/Storage/Http/Buckets/Folders/XList.php
TorstenDittmann: Agreed that standard cursor pagination requires persisted resources with IDs. Since folders are virtual and have no IDs, I removed the listFolders endpoint in ade4d4940a.

src/Appwrite/Platform/Modules/Storage/Http/Buckets/Folders/XList.php
TorstenDittmann: Agreed. Correctly deriving unique, permission-aware folders required scanning and batching file documents, so I removed the listFolders endpoint in ade4d4940a rather than expose that behavior.

src/Appwrite/Platform/Modules/Storage/Http/Buckets/Folders/XList.php
TorstenDittmann: The database can filter file paths but cannot portably return distinct immediate virtual folders. I removed the listFolders endpoint in ade4d4940a; folder filtering remains supported on listFiles.

src/Appwrite/Platform/Modules/Storage/Http/Buckets/Folders/XList.php
TorstenDittmann: Virtual folders do not have their own createdAt value, and ordering file documents does not define unique folder ordering. I removed the listFolders endpoint in ade4d4940a.

Meldiron (approved):

PR #12862, opened 2026-07-13
fix: verify session ownership on delete

RushikeshGandhmal:
## What does this PR do?

This fixes an IDOR in the admin user session delete endpoint.

Before this change, DELETE /v1/users/:userId/sessions/:sessionId loaded the user and session independently, then deleted the session without verifying that it belonged to the user in the route. That allowed a session owned by user A to be deleted through user B's URL.

This PR now verifies session ownership before deleting and returns user_session_not_found when the session does not belong to the requested user.

It also adds an end-to-end regression test that proves the mismatched delete is rejected, the original session remains intact, and the valid owner delete still succeeds.

## Test Plan

Verified in the appwrite container with focused checks:

- php -l app/controllers/api/users.php
- php -l tests/e2e/Services/Users/UsersBase.php
- ./vendor/bin/pint --test app/controllers/api/users.php tests/e2e/Services/Users/UsersBase.php
- test tests/e2e/Services/Users --filter=/(testCreateSession|testDeleteUserSessionRequiresOwnership)/
- test tests/e2e/Services/GraphQL/UsersTest.php --filter=/(testDeleteUserSession)/
- git diff --check

## Related PRs and Issues

- Fixes #12799

## Checklist

- [x] Have you read the Contributing Guidelines on issues?
- [ ] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

PR #12952, opened 2026-07-23
fix: verify session ownership on delete

deepshekhardas:
Replacement for #12862 by @RushikeshGandhmal.

Rebased on latest 1.9.x. Squashed commits. No functional changes.

### DIFF
--- a/app/controllers/api/users.php
+++ b/app/controllers/api/users.php
@@ -2552,6 +2552,10 @@ Http::delete('/v1/users/:userId/sessions/:sessionId')
             throw new Exception(Exception::USER_SESSION_NOT_FOUND);
         }
 
+        if ($user->getId() !== $session->getAttribute('userId')) {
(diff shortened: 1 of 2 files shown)


Meldiron:

#12862 merged to `main` on 2026-08-03. Both halves are already there, just relocated by the Users API refactor (`app/controllers/api/users.php` no longer exists on `main`):
- the ownership guard now lives at `src/Appwrite/Platform/Modules/Users/Http/Users/Sessions/Delete.php:73-75`, as the same `$user->getId() !== $session->getAttribute('userId')` check throwing `USER_SESSION_NOT_FOUND`
- `testDeleteUserSessionRequiresOwnership` is already in `tests/e2e/Services/Users/UsersBase.php:2430`

Merged `origin/main` into this branch to confirm — after resolving, the net diff against `main` is empty apart from an unrelated `Dockerfile` change carried in from `1.9.x` history.

PR #12954, opened 2026-07-23
fix: add .gitattributes rule to force LF for shell scripts

deepshekhardas:
Replacement for #12841 by @dobbydobap.

Rebased on latest 1.9.x. No functional changes.


Meldiron:
Retargeted to `main` and rebased — `.gitattributes` is identical on both branches, so the change applied cleanly with no conflicts.

Meldiron (approved):

PR #12957, opened 2026-07-23
fix: return ISO8601 session expire from updateSession

deepshekhardas:
Replacement for #12891 by @cipheraxat.

Rebased on 1.9.x with no conflicts.

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -838,7 +838,7 @@ Http::patch('/v1/account/sessions/:sessionId')
 
         // Extend session
         $authDuration = $project->getAttribute('auths', [])['duration'] ?? TOKEN_EXPIRATION_LOGIN_LONG;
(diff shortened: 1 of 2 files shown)


Meldiron:

#12891 merged to `main` on 2026-08-03 with the same fix. On `main` today:
- `app/controllers/api/account.php:904` and `:929` already wrap both values in `DateTime::formatTz(...)`
- `AccountCustomClientTest.php:5423` and `:5437` already carry the `createFromFormat('Y-m-d\TH:i:s.uP', ...)` assertions

I merged `origin/main` into this branch to confirm. `account.php` produced a single blank-line conflict — the `formatTz` lines themselves auto-merged because they're now identical on both sides. After resolving, the net diff against `main` is empty apart from an unrelated `Dockerfile` change that came from `1.9.x` history, not from this PR.

PR #12972, opened 2026-07-24
fix(vcs): skip unnecessary personal OAuth2 refresh for GitHub installations

HarshMN2345:
## Summary
- GitHub App installations authenticate via a JWT-derived installation access token (`GitHub::initializeVariables()` in `vendor/utopia-php/vcs`), generated independently of `personalAccessToken`/`personalRefreshToken`. Every GitHub API call in the adapter (`listBranches`, `getOwnerName`, `getRepositoryName`, etc.) uses that JWT-derived token exclusively — never the personal OAuth token.
- `InstallationTokens::refreshForInstallation()` was unconditionally refreshing the personal OAuth2 token before every GitHub-scoped read operation (branch listing, repository listing/detection, file contents), even though nothing downstream consumes it for GitHub.
- GitHub rotates OAuth refresh tokens on every use. Two concurrent refresh calls sharing the same stored refresh token cause the second to fail with `"Failed to refresh OAuth2 access token. Please reconnect the installation."`, even though the installation itself is fully functional. This is a real, race-triggered failure mode — not tied to any actual installation corruption.
- Root-caused from a customer report of intermittent "branches don't appear in Console" and confirmed via code trace, not guesswork: see [GitHub.php:68-86](https://github.com/appwrite/appwrite/blob/main/vendor/utopia-php/vcs/src/VCS/Adapter/Git/GitHub.php) for the JWT-only token flow, and [Branches/XList.php:87](https://github.com/appwrite/appwrite/blob/main/src/Appwrite/Platform/Modules/VCS/Http/Installations/Repositories/Branches/XList.php) for the unconditional refresh call site.

## Fix
`InstallationTokens::refreshForInstallation()` now returns the installation unchanged for `provider === 'github'`, skipping the OAuth2 refresh entirely. Other providers (GitLab, Gitea) are unaffected — they genuinely use OAuth2 access tokens for API calls and still refresh normally.

Repository creation under a personal namespace (the one legitimate use of the personal token — GitHub App tokens can't create personal-account repos, so this needs user-scoped OAuth) is unaffected: it calls `InstallationTokens::refresh()` directly with its own guard in `Repositories/Create.php`, not through `refreshForInstallation()`.

## Test plan
- [x] Added `testGithubProviderSkipsPersonalTokenRefresh` — proves GitHub never reaches the OAuth2 client (empty provider registry would throw if it did)
- [x] Added `testNonGithubProviderStillRefreshesPersonalToken` — proves GitLab/Gitea still refresh correctly
- [x] All 11 tests in `tests/unit/Vcs/InstallationTokensTest.php` pass
- [ ] Deploy and confirm branch listing no longer intermittently fails for GitHub installations under concurrent load

### DIFF
--- a/src/Appwrite/Vcs/InstallationTokens.php
+++ b/src/Appwrite/Vcs/InstallationTokens.php
@@ -17,13 +17,9 @@ class InstallationTokens
     {
         $provider = $installation->getAttribute('provider', 'github');
 
-        // GitHub Apps authenticate via a JWT-derived installation access token
-        // (see GitHub::initializeVariables()), which is generated independently
-        // of personalAccessToken/personalRefreshToken. Refreshing those here is
-        // both unnecessary and unsafe: GitHub rotates OAuth refresh tokens on
-        // every use, so a concurrent refresh call fails the other with
-        // "Failed to refresh OAuth2 access token" even though the installation
-        // itself is perfectly functional.
+        // GitHub uses a JWT-derived installation token (GitHub::initializeVariables()),
+        // not personalAccessToken/personalRefreshToken -- refreshing those here only
+        // risks a token-rotation race against concurrent requests, for no benefit.
         if ($provider === 'github') {
             return $installation;
         }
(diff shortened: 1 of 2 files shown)


HarshMN2345:
Rebase note: the race this targets is already fixed on main, so what's left is a one-line GitHub guard on top of main's existing personal-account check. Needs a rebase down to that before review.

PR #12979, opened 2026-07-24
fix: serialize webhook pause alert with distributed lock

### LINKED ISSUE #12944: Concurrent webhook failures can send duplicate "webhook paused" alerts at the retry threshold
### 👟 Reproduction steps

The webhooks worker's failure branch is not concurrency-safe around the retry-threshold transition. When two failing deliveries for the same webhook are processed concurrently (two subscribed events dispatched close together), both handlers can independently cross the threshold and each enqueue a "paused" notification.

1. Configure a webhook whose target returns an error (e.g. a non-existent URL or a 5xx endpoint) and subscribe it to a frequently-emitted event.
2. Set `_APP_WEBHOOK_MAX_FAILED_ATTEMPTS` to `10` (default) and drive `attempts` to `9`.
3. Emit two subscribed events in quick succession so two `v1-webhooks` worker handlers run concurrently for the same webhook.
4. Observe the notification recipient's inbox / notification queue.

### 👍 Expected behavior

Crossing the failure threshold is a single logical event, so exactly one "webhook paused" alert should be sent per pause cycle.

### 👎 Actual Behavior

Both concurrent handlers can send a "paused" alert, producing duplicate notifications for one threshold transition.

Root cause (`src/Appwrite/Platform/Workers/Webhooks.php`, lines 164–192 — permalink below):

```php
$dbForPlatform->increaseDocumentAttribute('webhooks', $webhook->getId(), 'attempts', 1); // L165 atomic
$webhook = $dbForPlatform->getDocument('webhooks', $webhook->getId());                     // L166 reload
$attempts = $webhook->getAttribute('attempts');                                            // L167
// ...
if ($attempts >= \intval(System::getEnv('_APP_WEBHOOK_MAX_FAILED_ATTEMPTS', '10'))) {      // L185
    $webhook->setAttribute('enabled', false);
    $updatePayload['enabled'] = false;
    $this->sendAlert(...);                                                                 // L188 alert BEFORE persist
}
$dbForPlatform->updateDocument('webhooks', $webhook->getId(), new Document($updatePayload)); // L191 persist enabled=false
```

Although the increment at L165 is atomic, the threshold check (L185) and `sendAlert` (L188) happen *before* `enabled = false` is persisted (L191). With `attempts` starting at 9, handler A increments to 10 and handler B to 11; both reload a document with `attempts >= 10`, both take the branch, and both call `sendAlert`. The notification deduplication key does not save this — it is:

```php
deduplicationKey: 'webhook:' . $webhook->getId() . ':paused:' . $webhook->getUpdatedAt(),
```

Each handler's `getUpdatedAt()` reflects its own `increaseDocumentAttribute` write (two distinct timestamps), so the two keys differ and both notifications pass deduplication.

Permalink: https://github.com/appwrite/appwrite/blob/4100141decfcef816132d5b113c53152bbeb83a1/src/Appwrite/Platform/Workers/Webhooks.php#L164-L192

### 🎲 Appwrite version

main branch (self-hosted); the code path is unchanged as of commit `4100141`.

### 💻 Operating system

Linux

### 🧱 Your Environment

`src/Appwrite/Platform/Workers/Webhooks.php`, `execute()` failure branch (lines 164–192) and `sendAlert()` deduplication key (lines 318–321). Manifests under the concurrent `v1-webhooks` worker model; the webhook is still correctly paused — only the alert is duplicated.

Suggested fix: make the "cross threshold → pause → alert" transition claim-once. For example, gate `sendAlert` on the persisted state actually transitioning from enabled→disabled (only alert when the `enabled = false` write is the one that flipped it), or base the deduplication key on a stable per-pause-cycle value rather than `getUpdatedAt()` (which changes on every attempt increment).

Found while running [Ito](https://ito.ai) (AI code review, free for open source) against recently merged PRs — full analysis: https://app.ito.ai/share/98820504-87a2-4566-aa81-aee6e05be7a1.

ChiragAgg5k:
## What does this PR do?

Fixes a race in the webhooks worker where two concurrent failing deliveries for the same webhook can both cross `_APP_WEBHOOK_MAX_FAILED_ATTEMPTS` and each enqueue a "webhook paused" notification.

Instead of claiming the alert with fragile `attempts === max` / stable-dedup heuristics, this serializes the critical section with the existing `locks` resource:

1. Inject `locks` and hold `webhook:{id}` around attempt increment, threshold check, and `enabled=false` persist.
2. Only the handler that still sees `enabled=true` when crossing the threshold claims the pause and sends the alert.
3. Successful deliveries reset `attempts` under the same lock key so they cannot race with the failure path.

## Test Plan

1. Configure a webhook that always returns 5xx; set `_APP_WEBHOOK_MAX_FAILED_ATTEMPTS=10`.
2. Drive `attempts` to 9, then emit two subscribed events in quick succession.
3. Confirm exactly one "Webhook deliveries have been paused" notification is enqueued/sent.
4. Confirm the webhook ends with `enabled=false`.
5. Re-enable the webhook, fail it to the threshold again, and confirm another pause alert is sent.

## Related PRs and Issues

- Fixes https://github.com/appwrite/appwrite/issues/12944

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Workers/Webhooks.php
+++ b/src/Appwrite/Platform/Workers/Webhooks.php
@@ -181,14 +181,25 @@ class Webhooks extends Action
             $webhook->setAttribute('logs', $logs);
 
             $updatePayload = ['logs' => $logs];
+            $maxAttempts = \intval(System::getEnv('_APP_WEBHOOK_MAX_FAILED_ATTEMPTS', '10'));
+            // Only the handler that crosses the threshold claims the pause alert.
+            // Concurrent failures can both see attempts >= max; alerting only when
+            // attempts === max (and the webhook was still enabled) makes this a
+            // single logical event. Persist enabled=false before enqueueing.
+            $shouldAlert = false;
 
-            if ($attempts >= \intval(System::getEnv('_APP_WEBHOOK_MAX_FAILED_ATTEMPTS', '10'))) {
+            if ($attempts >= $maxAttempts) {
+                $wasEnabled = $webhook->getAttribute('enabled', true);
                 $webhook->setAttribute('enabled', false);
                 $updatePayload['enabled'] = false;
-                $this->sendAlert($attempts, $statusCode, $webhook, $project, $dbForPlatform, $publisherForNotifications, $plan);
+                $shouldAlert = $wasEnabled && $attempts === $maxAttempts;
             }
 
             $dbForPlatform->updateDocument('webhooks', $webhook->getId(), new Document($updatePayload));
+
+            if ($shouldAlert) {
+                $this->sendAlert($attempts, $statusCode, $webhook, $project, $dbForPlatform, $publisherForNotifications, $plan);
+            }
             $dbForPlatform->purgeCachedDocument('projects', $project->getId());
 
             $error = $logs;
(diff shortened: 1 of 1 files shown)


ChiragAgg5k:
Closing — will take a different approach (DB conditional claim / transaction instead of distributed lock).

PR #12982, opened 2026-07-24
fix: spatial isNull/isNotNull query support

deepshekhardas:
Replacement PR for #12901. Same fix, rebased on latest main.\n\n- Adds e2e coverage for isNull/isNotNull on spatial attributes\n- Fixes MessagingFanoutTest docblock type annotation\n\nNote: composer.lock changes omitted (database already at 6.4.4 on main).

### DIFF
--- a/tests/e2e/Services/Databases/DatabasesBase.php
+++ b/tests/e2e/Services/Databases/DatabasesBase.php
@@ -8730,6 +8730,28 @@ trait DatabasesBase
         $this->assertEquals(200, $response['headers']['status-code']);
         $this->assertCount(2, $response['body'][$this->getRecordResource()]);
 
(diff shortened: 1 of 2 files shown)


Meldiron:
Closing as a duplicate of #13003, which is the better-formed of the two.

This branch is missing the `composer.json` constraint change that is the actual fix, and instead carries an unrelated one-line PHPDoc tweak in `tests/unit/Messaging/MessagingFanoutTest.php`. The 22-line spatial e2e block is identical in both.

Note for #13003: its bump (`6.0.*` → `^6.0.0`) is moot now — `main` is on `utopia-php/database: ^7.3.11`. After syncing with `main` the only thing that survives is the e2e regression test, which `main` does not have. That test is still worth landing on its own.

PR #12989, opened 2026-07-25
fix: retry SMTP send once on 421 timeout (stale keepalive)

deepshekhardas:
Rebased on latest 1.9.x from original PR #12628 with conflict resolution.

### DIFF
--- a/src/Appwrite/Platform/Workers/Mails.php
+++ b/src/Appwrite/Platform/Workers/Mails.php
@@ -128,10 +128,9 @@ class Mails extends Action
 
             $previewLen = strlen($preview);
             if ($previewLen < $this->previewMaxLen) {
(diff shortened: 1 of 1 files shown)


Meldiron:
Closing in favour of #12681, which is the same fix without the collateral damage.

This branch rewrites the whole of `src/Appwrite/Platform/Workers/Mails.php` (+292/−231) with line-ending damage, where #12681 does it in +49/−6.

Worth noting for whoever picks #12681 up: the problem is still real on `main`. `app/init/registers.php:353` still builds the pooled system SMTP with `keepAlive: true` and there is no `_APP_SMTP_KEEPALIVE` anywhere — `main` only fixed the per-project SMTP path (`Mails.php:154`, `keepAlive: false`). #12681 will need rebasing onto `main`'s restructured `Mails.php`, which now wraps sending in a `$send` closure with telemetry and checks `deliveredTo` rather than scanning `results`.

PR #12995, opened 2026-07-25
Fix #12623: Add _APP_SMTP_KEEPALIVE to opt out of SMTP connection reuse

### LINKED ISSUE #12623: 🐛 SMTP keepalive causes 421 timeout on first send after idle; subsequent sends work (1.9.0 self-hosted)

deepshekhardas:
Rebased on latest 1.9.x.


Meldiron:
Closing as a duplicate of #13008.

I diffed the two branches' own patch content directly — they are identical, 90 lines each. The only difference between the branches is which `1.9.x` snapshot they were cut from.

Keeping #13008. The fix is still needed: `_APP_SMTP_KEEPALIVE` does not exist on `main`, and `app/init/registers.php:353` still builds the pooled SMTP with `keepAlive: true`.

PR #13005, opened 2026-07-26
fix: cache empty get document results (replacement of #12961)

deepshekhardas:
Replacement of #12961 by @fogelito. Original was CONFLICTING — rebased onto 1.9.x and conflict resolved in composer.json (composer.lock conflict resolved by accepting current 1.9.x version).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -63,7 +63,7 @@
         "utopia-php/compression": "0.1.*",
         "utopia-php/config": "1.*",
         "utopia-php/console": "0.1.*",
-        "utopia-php/database": "6.0.*",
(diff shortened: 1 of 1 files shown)


Meldiron:

The only change here pins `utopia-php/database` to `dev-cache-empty-get-document as 6.4.4`. `main` is on `^7.3.11`, and #12961 ("Cache empty get document results") merged to `main` on 2026-07-27, so the dev-branch pin is both stale and a downgrade. Merging `main` in conflicts on exactly that line and resolving it leaves no diff.

Worth noting the pin also dropped the line's indentation, so it would have failed formatting anyway.

PR #12726, opened 2026-06-30
debug: log site rendering detection inputs and result

HarshMN2345:
Temporary debug logging to trace why post-build rendering detection returns `static` for tanstack-start SSR sites.

Logs framework name, file list passed to the detector, and detection result to worker container logs.

**To use:** trigger a failing vibes build and check `docker logs appwrite-worker-builds` (or cloud worker logs).

> **Note:** This PR should be reverted/replaced once root cause is found.

PR #13007, opened 2026-07-26
debug: log site rendering detection inputs and result (replacement of #12726)

deepshekhardas:
Replacement of #12726 by @HarshMN2345. Original was CONFLICTING — rebased onto 1.9.x, applied cleanly.

### DIFF
--- a/src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
+++ b/src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
@@ -32,6 +32,7 @@ use Utopia\Database\Exception\Restricted;
 use Utopia\Database\Exception\Structure;
 use Utopia\Database\Exception\Transaction as TransactionException;
 use Utopia\Database\Query;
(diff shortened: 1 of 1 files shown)


Meldiron:
Retargeted to `main` and rebased.

`Builds::detectSiteRendering()` no longer exists on `main` — it was extracted into `Appwrite\Deployment\Detection::rendering()`, which is a pure static helper also exercised by `tests/unit/Deployment/DetectionTest.php`. I moved the logging to its caller, `Jobs::detect()`, so the helper stays side-effect free and unit tests don't emit build logs. Same three lines (framework, files, result).

The second commit's `/** @var RenderingDetection $result */` PHPStan hint was dropped: it annotated the inline detector construction that no longer exists, and `main` already resolves the return type through `Detection::rendering()`.

Meldiron:
Hello, inspiration was temporary PR to debug a live issue.

Please open GitHub issue if this is important change for you. Closing in the meantime.

Issue #13016, opened 2026-07-27
openruntimes-executor crash loop after upgrading Appwrite 1.9.0 → 1.9.6 (container_name/hostname mismatch, undocumented)


svprdga:
### 💭 Description

After upgrading Appwrite from 1.9.0 to 1.9.6, openruntimes-executor went into a permanent crash loop. Took a while to figure out why because the error message is useless:

```
Fatal error: Uncaught Utopia\DI\Exceptions\NotFoundException: Dependency response not found in .../di/src/DI/Container.php:74
...
Next Utopia\Http\Exception: Error handler had an error: Dependency response not found in .../Http/Http.php:534
```

This loops forever and never says what actually failed.

Turns out the real cause is in `app/http.php`'s `onStart` hook:

```php
$hostname = gethostname();
$selfContainer = $orchestration->list(['name' => $hostname])[0] ?? throw new \RuntimeException('Own container not found');
```

It looks itself up in Docker by matching `hostname` against the container `name`. Our compose had `container_name: openruntimes-executor` and `hostname: exc1`, different values, so the lookup fails and it throws.

Checked appwrite/appwrite's own compose history: up to 1.9.0 they also had `container_name` different from `hostname`, and only made them match starting in 1.9.5. So this became a real requirement at some point but was never called out anywhere, not in the changelog, not in the upgrade notes for 1.9.5/1.9.6. If you're upgrading from an older install you just hit this with zero warning.

On top of that, the reason we never saw the real error message: when the `onStart` hook throws, the framework tries to run the global error hooks, which need a `response` dependency that doesn't exist yet at startup time. That lookup fails too, and the *second* failure is all that gets logged. The original `Own container not found` never makes it to the logs.

Two asks:
1. Document that `container_name` must match `hostname` for the executor, this is a breaking requirement for anyone upgrading from pre-1.9.5.
2. Fix the error handling so a failure during `onStart` actually gets logged before the error-hook machinery kicks in and (possibly) fails on its own.

Reproduced on executor 0.25.4 and 0.27.0, self-hosted via Docker Compose.

This should fix #12796 , same symptom, but that report doesn't identify the actual trigger. Here's the root cause for Docker Compose setups.

### 👀 Have you spent some time to check if this issue has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

[...]

PR #13018, opened 2026-07-27
Document executor container_name/hostname requirement

sarbojitrana:
## Summary
- Adds an inline comment in `docker-compose.yml` above the `openruntimes-executor` service explaining that `container_name` and `hostname` must match, since the executor looks itself up in Docker at startup by matching `hostname` against container `name`.
- Adds an "Upgrade note" bullet to `CHANGES.md` under the 1.9.5 "Installer and self-hosted" section (next to PR #11874, which is what made these values match), so self-hosters upgrading from a custom pre-1.9.5 compose file know why the executor can crash-loop.

This covers the documentation half of #13016. The reporter also found that when the executor's `onStart` hook throws, the framework's global error-hook machinery itself fails trying to resolve a `response` dependency that doesn't exist yet at startup — and that secondary failure is all that gets logged, masking the real error. That fix belongs in `open-runtimes/executor` (`app/error.php`) and/or `utopia-php/http` (`Http::start()`), not in this repo, so it's out of scope here.

## Test plan
- [x] `docker compose config -q` parses the modified `docker-compose.yml` without error
- [x] Manually reviewed `CHANGES.md` rendering for formatting

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -109,6 +109,7 @@
 * Fix installer compose executor config in [#11874](https://github.com/appwrite/appwrite/pull/11874)
 * Fix public hostname validation under Swoole DNS in [#12431](https://github.com/appwrite/appwrite/pull/12431)
 * Fix host validation and regional endpoint validation in [#12411](https://github.com/appwrite/appwrite/pull/12411) and [#12357](https://github.com/appwrite/appwrite/pull/12357)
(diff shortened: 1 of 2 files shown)


sarbojitrana:
@ArnabChatterjee20k could you please review this pr or assign me if something needs to be improved in it

Hardanish-Singh (commented):
LGTM!

Hardanish-Singh (commented):
LGTM

PR #13031, opened 2026-07-28
fix(vcs): serialize installation token refresh with a lock

HarshMN2345:
## What

Serializes installation token refreshes with a per-installation lock document, so two concurrent requests can no longer exchange the same OAuth2 refresh token.

## Why

When an installation's access token is expired, every request that needs it calls `InstallationTokens::refresh()`. Nothing coordinated those calls, so two requests arriving together both read the same refresh token and both call `refreshTokens()`:

```
R1 refresh()                R2 refresh()      both see an expired token
R1 refreshTokens()                            new token pair issued
R1 updateDocument()
                            R2 refreshTokens() replays the rotated token
                                               -> reuse detection revokes the token family
                            R2 updateDocument()
```

Providers rotate refresh tokens on use, and replaying a used one is treated as a leaked-token attack: the whole family is revoked. The installation stays connected and visible in project settings, but every refresh from then on fails with `Failed to refresh OAuth2 access token`, which surfaces as repository lookups returning 400 and manual redeploys never leaving `processing`. Automatic redeploys from pushes keep working, since they do not take this path.

Easy to trigger from the console: with an expired token, double-clicking an action that lists or resolves repositories sends two requests, and both refresh.

## How

`refresh()` takes a lock document keyed `installation-{installationId}` in `vcsCommentLocks` before exchanging anything, and releases it in `finally`. This follows the pattern already used for PR comment updates in `GitAction::updateBuildStatus()` and `Http/GitHub/Deployment.php` — same collection, same nine attempts a second apart, same release.

The important half is the re-read: after acquiring the lock, the installation is loaded again from the database. If the request that held the lock already refreshed, the fresh token is returned and the provider is never called. A lock on its own would only delay the second exchange by a second, not prevent it.

```
R1 lock ok             R2 lock taken -> waits
R1 re-read -> expired
R1 refreshTokens() -> updateDocument() -> release
                       R2 lock ok -> re-read -> token valid -> return, no exchange
```

Two details worth noting:

- The create and delete run through `getAuthorization()->skip()`. `vcsCommentLocks` carries no permissions, and these endpoints run in console context under the caller's roles, so without the skip every acquisition would fail. `Http/GitHub/Deployment.php` skips its release for the same reason.
- If the attempts run out, the installation is re-read and the holder's token reused when it landed. Only when the row is still unusable does the request fail, as `GENERAL_RESOURCE_LOCKED` (409, retry after a brief delay) rather than a raw database error. It never falls through to exchanging the rotated token.

The refresh body itself is unchanged, only moved behind the lock into `exchange()`. The repeated "row has a usable token" condition is now `isUsable()`.

## Test plan

`tests/unit/Vcs/InstallationTokensTest.php` — 10 tests, all passing:

- `testRefreshIsSerializedByLock` — the lock document is created with the expected id before the exchange and deleted afterwards.
- `testLockIsReleasedWhenRefreshFails` — the lock is released when the provider call throws.
- `testTokenRefreshedByLockHolderIsReused` — when the row already holds a valid token, the refreshed token is returned and `refreshTokens()` is never called.
- Existing coverage for unexpired, missing and malformed expiry, identity fallback, missing refresh token and failed refresh is unchanged.

Pint, PHPStan level 4 and the Rector check pass.

## Known limits

- A process dying between acquiring and releasing leaves the lock document behind, and that installation's refreshes fail until the row is removed. The existing comment locks carry the same exposure; a TTL would be worth adding for both rather than only here.
- The wait is nine seconds. A refresh that outlives it makes the waiting request retry rather than reuse, since `OAuth2::request()` sets no timeout and the exchange is therefore unbounded.
- Installations whose refresh token is already past the provider's own expiry window cannot be recovered by any refresh, including this one. They need a re-authorize path in the console, which is a separate change.


HarshMN2345:
@greptile The locking here follows the existing vcsCommentLocks pattern already used in the codebase (same collection, same 9 retry attempts spaced 1 second apart, and the lock is always released in finally):

* src/Appwrite/Deployment/GitAction.php:67-95
* src/Appwrite/Platform/Modules/VCS/Http/GitHub/Deployment.php:209-243 (and the other two call sites in the same file)

So the two concerns raised aren’t introduced by this PR—they already exist in the shared locking implementation. Lock documents don’t expire, and lock acquisition has a bounded wait.

There is one difference worth pointing out. The existing comment locks are keyed by provider comment IDs, which are short-lived, so if a lock is ever leaked it naturally becomes irrelevant as new comments get new IDs. This PR uses the installation ID as the key, which is long-lived, so a leaked lock would have a larger blast radius.

I think the right fix for that is a periodic cleanup of stale lock documents in Maintenance. The collection currently has no attributes or indexes (app/config/collections/platform.php:2227), so that improvement would benefit both the existing comment locks and this new installation lock. I’d treat that as a follow-up rather than something specific to this PR.

For the bounded wait, once retries are exhausted we re-read the installation and reuse the token if the other request already completed successfully. We only return a retryable conflict if the row is still unusable. The alternative would be to continue with the refresh anyway, which risks replaying a rotated refresh token and invalidating the entire token family—the exact race this PR is trying to prevent. That trade-off is intentional.

HarshMN2345:
@greptile re-review

src/Appwrite/Vcs/InstallationTokens.php
```diff
@@ -45,41 +46,80 @@ public function refresh(Document $installation, Database $dbForPlatform, OAuth2
...
+        $accessToken = $oauth2->getAccessToken('');
+
         $installation = $installation
             ->setAttribute('personalAccessToken', $accessToken)
-            ->setAttribute('personalRefreshToken', $refreshToken)
+            ->setAttribute('personalRefreshToken', $oauth2->getRefreshToken(''))
             ->setAttribute('personalAccessTokenExpiry', DateTime::addSeconds(new \DateTime(), (int)$oauth2->getAccessTokenExpiry('')));
 
+        // The provider has already rotated the family, so persist before anything else can fail.
         $dbForPlatform->updateDocument('installations', $installation->getId(), new Document([
```
Meldiron:
```suggestion
$installation = $dbForPlatform->updateDocument('installations', $installation->getId(), new Document([
```

src/Appwrite/Vcs/InstallationTokens.php
```diff
@@ -45,41 +46,80 @@ public function refresh(Document $installation, Database $dbForPlatform, OAuth2
...
+        $lock = 'installation-' . $installation->getId();
+        $authorization = $dbForPlatform->getAuthorization();
+        $acquired = false;
+        $retries = 0;
+
+        while ($retries < 9) {
+            $retries++;
+
+            try {
+                $authorization->skip(fn () => $dbForPlatform->createDocument('vcsCommentLocks', new Document(['$id' => $lock])));
```
Meldiron, line 63: Add note about the fact we are reusing comments table for simplcitiy, and ideally we move it to generic table later

Meldiron (approved):

src/Appwrite/Vcs/InstallationTokens.php
```diff
@@ -45,41 +46,80 @@ public function refresh(Document $installation, Database $dbForPlatform, OAuth2
...
+            // The lock holder may have refreshed already.
             $current = $this->getCurrentInstallation($dbForPlatform, $installation);
-            if (!$current->isEmpty() && !empty($current->getAttribute('personalAccessToken')) && !$this->isExpired($current->getAttribute('personalAccessTokenExpiry'))) {
+
+            if ($this->isUsable($current)) {
                 return $current;
             }
 
+            return $this->exchange($installation, $dbForPlatform, $oauth2);
+        } finally {
```
Meldiron, line 94: Add a catch; when error happens, update installation to have empty string for access and refrehs tokens; expiry null;

Meldiron, line 94: Upddate installations endpoint to have new response model rule which will say `fresh` boolean - depending on if acces token is empty.

HarshMN2345:
@greptile All four points are accurate readings of the code. Two are deliberate trade-offs, one is now self-limiting, and one is the fix for an earlier defect in this same PR. Taking them in turn.

**Locks can remain after a process or deletion failure.** Correct, and inherent to a lock document with no expiry. This follows the mechanism already used for pull request comment updates (`src/Appwrite/Deployment/GitAction.php:67-95`, `src/Appwrite/Platform/Modules/VCS/Http/GitHub/Deployment.php:209-243`), so the exposure is not new, though the key here is longer lived than a comment id. The fix that covers both call sites is a periodic sweep of stale lock documents in `Maintenance`; the collection carries no attributes or indexes today (`app/config/collections/platform.php:2227`). That is a separate change rather than something to fold in here.

**Bounded waiters can reject valid contention.** Correct, and chosen. When the wait lapses the installation is re-read and the holder's token reused if it landed, so the request only fails when the row is still unusable, and it fails as a retryable conflict. The alternative is to proceed and exchange the same refresh token, which is what revokes the token family and is the failure this PR exists to prevent. Any finite wait has this property; the question is only what happens when it lapses.

**Persistence failure after rotation leaves the consumed token reachable.** Correct, and now bounded rather than permanent. If the write fails after the provider has rotated, the stale token is replayed on the next call, the provider refuses it with `invalid_grant` or `bad_refresh_token`, and that response clears the pair. So the cost is one further failed call, not a stuck row. The write cannot be made atomic with the rotation, since the provider commits first.

**Failed identity verification leaves the replacement credential eligible for reuse.** Correct, and intended. Verification is a separate provider call that can fail on its own, so its failure does not establish that the token is bad. Discarding a pair the provider has already rotated to was the previous behaviour, and it stranded installations permanently with a refresh token that no longer worked; ordering the write ahead of the check is the fix for that. Keeping the pair and letting the next real call surface an authentication error is the safer default.

Given that the third point is self-limiting and the first two are properties of the lock mechanism this PR deliberately reuses, a re-review would be welcome.

PR #13032, opened 2026-07-28
fix: restore original session limit and password history caps

HarshMN2345:
## Summary
- Session limit and password history policies got switched to `APP_LIMIT_COUNT` (5000) during the auth-settings refactor into `Policies/*`. Sessions are backed by an unpaginated subquery, so a 5000-session-per-user cap risks degrading the database on a single abusive user.
- Restores `APP_LIMIT_USER_SESSIONS_MAX = 100` and `APP_LIMIT_USER_PASSWORD_HISTORY = 20`, matching the limits that existed prior to the refactor.
- Removes the `null`/unlimited option from the session limit policy, which was never supported before the refactor.

## Test plan
- [ ] `tests/e2e/Services/Project/PoliciesBase.php` min/max boundary tests updated for the new caps (100 / 20) and run via `docker compose exec appwrite test tests/e2e/Services/Project --filter=Policies`


src/Appwrite/Platform/Modules/Project/Http/Project/Policies/SessionLimit/Update.php
```diff
@@ -61,20 +60,15 @@ public function __construct()
...
         Authorization $authorization,
         Event $queueForEvents,
     ): void {
         $auths = $project->getAttribute('auths', []);
-
-        if (\is_null($total)) {
-            $auths['maxSessions'] = 0;
-        } else {
-            $auths['maxSessions'] = $total;
-        }
```
HarshMN2345, line 77: This removes the unlimited option for the session limit policy

Meldiron (approved):

Issue #13039, opened 2026-07-29
Presences API expiresAt validator uses static server boot-time bounds and fails on timezone offset comparison


bhardwajparth51:
### 👟 Reproduction steps

1. Deploy/start Appwrite server container on long-lived Swoole worker runtime.
2. Keep the server container running without restarting for > 30 days.
3. Call `PUT /v1/presences/:presenceId` or `PATCH /v1/presences/:presenceId` passing an `expiresAt` parameter for tomorrow.
4. See error: HTTP 400 Bad Request (`Invalid param: Value must be valid date between ...`).

### 👍 Expected behavior

Users should be able to set and update presence `expiresAt` dates up to 30 days into the future relative to the request time.

### 👎 Actual Behavior

Because `new DatetimeValidator(new \DateTime(), (new \DateTime())->modify('+30 days'))` is evaluated once inside `Action::__construct()` during route registration at server bootstrap, `$max` is permanently locked to `server_boot_time + 30 days`. 

After 30 days of server uptime, all requests setting `expiresAt` fail with HTTP 400 Bad Request.

Additionally, checking `presenceExpiresAt` using `DateTime::formatTz($presenceExpiresAt) < DateTime::formatTz(DateTime::now())` compares ISO text strings containing timezone offset tags lexicographically, falsely marking non-expired presences as expired when timezone offsets differ.

### 🎲 Appwrite version

Different version (specify in environment)

### 💻 Operating system

Linux

### 🧱 Your Environment

1.9.x

### 👀 Have you spent some time to check if this issue has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

PR #13040, opened 2026-07-29
fix: evaluate expiresAt validator per request

bhardwajparth51:
## What does this PR do?

`DatetimeValidator` in `Upsert.php` and `Update.php` was instantiated directly inside `__construct()`. Because Swoole worker processes stay alive in memory for weeks, `(new \DateTime())->modify('+30 days')` was captured at server boot time and never updated. After 30 days of server uptime, all requests setting `expiresAt` fail with a `400 Bad Request`.

This PR:
1. Wraps `DatetimeValidator` in a closure (`fn () => ...`) so validator date bounds recalculate dynamically per request.
2. Updates `expiresAt` expiry checks in `Get.php` and `Update.php` to compare unix timestamps (`strtotime`) instead of formatted ISO strings to avoid timezone offset comparison bugs.

## Test Plan

- Verified closure validator resolution dynamically recalculates bounds on incoming requests.
- Ran static analysis (`phpstan analyze src/Appwrite/Platform/Modules/Presences/ --level=3`) -> `[OK] No errors`.

## Related PRs and Issues

#13039
## Checklist

### DIFF
--- a/src/Appwrite/Platform/Modules/Presences/HTTP/Update.php
+++ b/src/Appwrite/Platform/Modules/Presences/HTTP/Update.php
@@ -97,7 +97,7 @@ class Update extends PlatformAction
             ->param('presenceId', '', fn (Database $dbForProject) => new UID($dbForProject->getAdapter()->getMaxUIDLength()), 'Presence unique ID.', false, ['dbForProject'])
             ->param('userId', null, new UID(), 'User ID.', true)
             ->param('status', null, new Text(Database::LENGTH_KEY), 'Presence status.', true)
-            ->param('expiresAt', null, new DatetimeValidator(
+            ->param('expiresAt', null, fn () => new DatetimeValidator(
                 new \DateTime(),
                 (new \DateTime())->modify('+30 days'),
                 requireDateInFuture: true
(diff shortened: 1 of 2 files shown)


src/Appwrite/Platform/Modules/Presences/HTTP/Get.php
```diff
@@ -60,7 +60,7 @@ public function action(string $presenceId, Response $response, Database $dbForPr
 
         $presenceExpiresAt = $presence->getAttribute('expiresAt');
 
-        if (!empty($presenceExpiresAt) && DateTime::formatTz($presenceExpiresAt) < DateTime::formatTz(DateTime::now())) {
+        if (!empty($presenceExpiresAt) && \strtotime($presenceExpiresAt) < \strtotime(DateTime::now())) {
```
ArnabChatterjee20k: can you share a bit more detail for this change? The pr describing the change reasons for the route action params. Didn't get this change?

bhardwajparth51: not needed, appwrite's already utc so formatTz was fine also strtotime silently marks bad input as expired instead of erroring

ArnabChatterjee20k (approved):

ArnabChatterjee20k (approved):

PR #13051, opened 2026-07-30
Add Bitbucket as a VCS provider

HarshMN2345:
`composer.json` points `utopia-php/vcs` at the unreleased `feat-bitbucket-adapter` branch ([utopia-php/vcs#125](https://github.com/utopia-php/vcs/pull/125)), which isn't merged/released yet. This branch exists to test the OAuth + webhook path on stage before either side is real.

## Summary
Adds Bitbucket as a VCS provider, OAuth-token-based like GitLab (no App-installation concept, unlike GitHub):

- `OAuth2\Bitbucket`: adds `getUserSlug()` (`username ?? nickname`, required by `Callback\Base` which aborts the install if empty) and `createRepository()` with an optional workspace override (Bitbucket has no implicit "current user's namespace" -- the workspace is always part of the URL, unlike GitHub/GitLab).
- `app/config/vcs.php`: registers `bitbucket`. Deliberately **no** `'endpoint'` key -- `Factory` feeds it to `setEndpoint()`, which on Bitbucket sets the API base (`api.bitbucket.org`), not the browser host (GitLab uses the same mechanism for the browser host instead).
- 3 new `Http/Bitbucket/{Authorize,Callback,Events}` controllers, mirroring GitLab's. `Events/Create.php` calls `getEvents()` (plural) and loops instead of the singular `getEvent()` -- Bitbucket batches every ref a push touched into one delivery, and the singular method only reports the first, silently dropping the rest.
- `Installations/Namespaces/XList.php`: the vcs branch promotes `listNamespaces()` to the abstract `Git` class with a throwing default rather than per-adapter, so `method_exists()` is now always true. Switched to try/catch so unsupported providers (Bitbucket, GitHub, Gitea) still fall back to wrapping the installation's own owner as a single namespace.
- Usual per-provider plumbing: `APP_VCS_BITBUCKET_EMAIL` constant, 3 env vars in `variables.php`/`.env`/`docker-compose.yml` (both app and worker-builds blocks).

## Test plan
- [x] `composer format && composer analyze` (clean)
- [x] `phpunit tests/unit/Vcs/` -- all 38 existing tests pass unmodified
- [ ] E2E on stage: OAuth authorize -> callback -> connect existing repo -> webhook push -> deployment (Bitbucket Cloud only delivers webhooks to a public URL, so this can't be verified locally)
- [ ] E2E on stage: create repository via Appwrite into a Bitbucket workspace


app/config/collections/platform.php
```diff
@@ -1842,7 +1842,7 @@
                 '$id' => ID::custom('personalAccessToken'),
                 'type' => Database::VAR_STRING,
```
Meldiron: Jake mention we can change type too.

app/config/vcs.php
```diff
@@ -60,4 +62,18 @@
...
         ],
     ],
+    'bitbucket' => [
+        'adapter' => Bitbucket::class,
+        'oauth2' => fn (string $clientId, string $clientSecret, string $endpoint) => new OAuth2Bitbucket($clientId, $clientSecret, ''),
+        // Only official bitbucket.org/api.bitbucket.org is supported -- fixed,
+        // not configurable. Deliberately no 'endpoint' key here: unlike
+        // GitLab's, Bitbucket's setEndpoint() sets the API base
+        // (api.bitbucket.org), not the browser host, so this must stay unset
+        // rather than pointed at the wrong host.
```
Meldiron, line 72: We can remove

src/Appwrite/Auth/OAuth2/Bitbucket.php
```diff
@@ -53,16 +53,19 @@ public function getLoginURL(): string
...
     {
         if (empty($this->tokens)) {
-            // Required as per Bitbucket Spec.
-            $headers = ['Content-Type: application/x-www-form-urlencoded'];
+            // Bitbucket wants HTTP Basic Auth here, not client_id/secret in
+            // the body -- the body form fails silently with a misleading
+            // "invalid_grant" error instead of an auth error.
+            $headers = [
+                'Content-Type: application/x-www-form-urlencoded',
+                'Authorization: Basic ' . \base64_encode($this->appID . ':' . $this->appSecret),
```
Meldiron, line 61: // Bitbucket wants HTTP Basic Auth here, not client_id/secret in
            // the body -- the body form fails silently with a misleading
            // "invalid_grant" error instead of an auth error.

src/Appwrite/Auth/OAuth2/Bitbucket.php
```diff
@@ -53,16 +53,19 @@ public function getLoginURL(): string
     protected function getTokens(string $code): array
     {
         if (empty($this->tokens)) {
-            // Required as per Bitbucket Spec.
-            $headers = ['Content-Type: application/x-www-form-urlencoded'];
+            // Bitbucket wants HTTP Basic Auth here, not client_id/secret in
+            // the body -- the body form fails silently with a misleading
+            // "invalid_grant" error instead of an auth error.
```
Meldiron, line 58: Might not need as comment

src/Appwrite/Platform/Modules/VCS/Http/Bitbucket/Authorize/Get.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return new OAuth2Bitbucket(
+            System::getEnv('_APP_VCS_BITBUCKET_CLIENT_ID', ''),
+            System::getEnv('_APP_VCS_BITBUCKET_CLIENT_SECRET', ''),
+            $callback,
+            $state,
+            [
+                'account',
+                'repository:write',
+                'pullrequest:write',
```
Meldiron, line 37: write means read+write

composer.json
```diff
@@ -100,7 +100,7 @@
         "utopia-php/telemetry": "^0.4",
         "utopia-php/user-agent": "0.1.*",
         "mustangostang/spyc": "0.6.*",
-        "utopia-php/vcs": "4.*.*",
+        "utopia-php/vcs": "dev-feat-bitbucket-adapter as 4.4.5",
```
Meldiron: leftover

Meldiron (approved):

PR #13055, opened 2026-07-31
Issue TLS certificates for function and site domains

### LINKED ISSUE #12821: Appwrite does not create a certificate when deploying functions
### 👟 Reproduction steps

1. Create a self-hosted Appwrite instance, version 1.9.0
2. Deploy a new function
3. Try to call the function, using a curl for instance (against `https://<function-id>.example.com/`)

### 👍 Expected behavior

The function responds.

### 👎 Actual Behavior

The function does not respond, because there's no SSL certificate issued.

Appwrite container seems to get the new function deploy:

```
Sync domains tick: 1 domains were updated
  http.path          /v1/proxy/rules
  http.path          /v1/proxy/rules
  http.path          /v1/proxy/rules
Sync domains tick: 1 domains were updated
```

But the appwrite-worker-certificates does not receive any message, so there's no SSL certificate automatically issued.

Workaround: manually running `docker compose exec appwrite ssl domain="<function-domain>"` does issue the certificate correctly. However, this isn't a viable long-term solution, having to manually issue (and later renew) a certificate for every single function domain doesn't scale, especially as the number of functions rows.

### 🎲 Appwrite version

### 💻 Operating system

Linux

### 🧱 Your Environment

1.9.0

ChiragAgg5k:
## What does this PR do?

Fixes function and site domains never getting a TLS certificate on self-hosted instances.

Appwrite-owned domains under `_APP_DOMAIN_FUNCTIONS` / `_APP_DOMAIN_SITES` are written as pre-verified rules that never enqueue a certificate job, because the design assumes a wildcard certificate covers the parent domain:

```php
'status'        => 'verified',
'certificateId' => '',
'owner'         => 'Appwrite',
```

Every other rule path enqueues to `publisherForCertificates` once status reaches `verifying`. These skip straight to `verified`, so `appwrite-worker-certificates` never receives a message — matching the reported "worker-certificates does not receive any message".

The wildcard assumption does not hold self-hosted. The only certificate adapter is `certbot certonly --webroot` (HTTP-01), which cannot issue wildcards, and `docker-compose.yml` configures no ACME resolver on Traefik, so it serves its self-signed default and HTTPS fails.

### Changes

- **New `Appwrite\Certificates\Issuance`** — decides whether Appwrite should auto-issue a per-subdomain certificate.
- **Enqueue at rule creation** in `Proxy/Http/Rules/Function/Create.php` and `Proxy/Http/Rules/Site/Create.php`. This is the fix that carries the PR — on 1.9 these endpoints are the only path that mints an Appwrite-owned function or site domain.
- **Skip DNS verification for Appwrite-owned rules** in the certificates worker. Appwrite generated the subdomain itself under an operator-configured domain, and these setups use a wildcard A record rather than a per-subdomain CNAME to `_APP_DOMAIN_TARGET_CNAME`.
- **New `_APP_OPTIONS_ROUTER_CERTIFICATES`** (default `enabled`) for operators who already have a wildcard certificate. Honoured both at issuance and by the renewals scheduled in `Tasks/Maintenance.php`.

Skipped entirely when `_APP_EDITION` is not `self-hosted`, so Cloud is unchanged.

### Scope

Only stable primary domains (`trigger: 'manual'`) — roughly one certificate per function or site. Per-deployment previews and `commit-` / `branch-` domains use `trigger: 'deployment'` and are left alone, since one certificate per deployment would burn through Let's Encrypt's ~50/week limit per registered domain.

The hunk in `Functions/Http/Functions/Create.php` only affects the `x-appwrite-response-format < 1.7.0` branch; on 1.9 a normal `POST /v1/functions` creates no deployment and no rule, so it is legacy-only.

Existing domains are not backfilled — nothing re-creates their rule. `docker compose exec appwrite ssl domain="<domain>"` still handles those; everything created after the upgrade is automatic.

## Test Plan

Verified end to end on a throwaway DigitalOcean droplet (`s-4vcpu-8gb`, fra1) against a real public domain and the **production** Let's Encrypt API — not staging, not `--dry-run`. Wildcard `A` record for `*.functions.<sub>.example.com`, `_APP_ENV=production`, stack built from this branch.

### Certificate issued, function serving real HTTPS

`T0` = `POST /v1/proxy/rules/function` returning. A poller hit the domain every second, armed **before** the rule existed.

| Δ from T0 | Event |
|---|---|
| +0.68s | Function serving traffic — `http=200`, real output |
| +11.67s | Certificates worker done, cert + `/storage/config/<domain>.yml` written |
| **+12.98s** | **First valid `https=200`** after Traefik hot-reload |

**Broken-HTTPS window: 12.3s**, of which certbot was 10.48s. A second run measured +9.96s / +12.10s, so ~12s is repeatable.

During that window the function worked and only TLS was broken:

| Phase | `http` | `https` (validated) | Cert served |
|---|---|---|---|
| Before rule | `404` error page | curl **exit 60** | `TRAEFIK DEFAULT CERT` |
| Rule live, cert pending | `200` real output | curl **exit 60** | `TRAEFIK DEFAULT CERT` |
| After issuance | `200` real output | `200` real output | `Let's Encrypt` |

Final check, no `-k`: `http_code=200 ssl_verify=0`, `issuer=C = US, O = Let's Encrypt`, `subject=CN=<domain>`, SAN matching.

### Controls

Same droplet, same project, same function — only the variable under test changed:

1. **`main` branch** — certificates worker received **zero messages**, no certificate, HTTPS stuck at curl exit 60 on `TRAEFIK DEFAULT CERT` while `http` served `200`. Reproduces #12821 and confirms the certificate here is caused by this PR.
2. **`_APP_OPTIONS_ROUTER_CERTIFICATES=disabled`** — no job enqueued, no certificate. Opt-out works.
3. **Re-enabled** — the same domain that failed under `main` got its certificate.

Main-domain path still works (`ssl domain=...` → real certificate in 10.8s), so the worker changes did not regress it.

### Two gotchas when reproducing

- **`_APP_ENV` must be `production`.** The default `.env` ships `development`, which makes `LetsEncrypt::issueCertificate()` pass `--dry-run`. The dry run *succeeds*, then issuance dies at `Failed to rename certificate cert.pem` — looks like a certificate bug, is a config one.
- **`docker compose up -d` uses the dev override.** `docker-compose.override.yml` points every service at a locally-built `appwrite-dev` image and ignores `_APP_VERSION`. Testing a branch needs `--build`; confirm what is live with `docker exec appwrite ls /usr/src/code/src/Appwrite/Certificates/Issuance.php`.

Also ran `composer format` and `composer lint` on all touched files.

## Related PRs and Issues

- Fixes #12821
- Related thread: https://appwrite.io/threads/1184528829369286707

## Checklist

- [x] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs? (No API metadata changed.)


src/Appwrite/Certificates/Issuance.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace Appwrite\Certificates;
+
+use Utopia\Domains\Domain;
+use Utopia\System\System;
+
+final class Issuance
```
loks0n: I don't think this class is significant enough to justify existence

app/config/variables.php
```diff
@@ -88,6 +88,15 @@
                 'question' => '',
                 'filter' => ''
             ],
+            [
+                'name' => '_APP_OPTIONS_ROUTER_CERTIFICATES',
```
loks0n: I prefer _APP_ROUTER_AUTO_CERTIFICATES

src/Appwrite/Certificates/Issuance.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    /**
+     * Whether Appwrite should auto-issue a per-subdomain TLS certificate for an
+     * Appwrite-owned function or site primary domain. Skipped when a wildcard
+     * certificate already covers the parent domain, or when the domain is not a
+     * known public hostname.
+     */
+    public static function isRequired(string $domain): bool
+    {
+        if (System::getEnv('_APP_EDITION', 'self-hosted') !== 'self-hosted') {
```
loks0n: We should avoid envs in application code. Better to move this to resources.php or wherever init happens

loks0n (changes requested):

app/init/resources.php
```diff
@@ -125,6 +126,20 @@
...
+    }
+
+    if (System::getEnv('_APP_ROUTER_AUTO_CERTIFICATES', 'enabled') === 'disabled') {
+        return false;
+    }
+
+    $domain = new Domain($hostname);
+
+    return $domain->isKnown() && !$domain->isTest();
+});
```
Meldiron: Lets check with Eldad or Luke if there is better pattern here.

My first thought is ti always have  something like certificateIssuer class, and sometimes adapter is None() while other times its real adapter.
But that might be overkill.

But having resource for boolean is uncommon too, so wider discussion is best

Meldiron: Additionally this seems to return callable, which is even worse than boolean I think. Gives too much freedom - breaks structure of resources

src/Appwrite/Certificates/Certificates.php
```diff
@@ -0,0 +1,44 @@
...
+
+final class Certificates
+{
+    private string $edition;
+    private string $autoCertificates;
+
+    public function __construct()
+    {
+        $this->edition = System::getEnv('_APP_EDITION', 'self-hosted');
+        $this->autoCertificates = System::getEnv('_APP_ROUTER_AUTO_CERTIFICATES', 'enabled');
```
loks0n: can we take these as injected parameters, and avoid getEnv within the app code?

loks0n (approved):

PR #13081, opened 2026-08-03
fix(functions): correct the ownership check when deleting an execution

Meldiron:
## What

`Executions/Delete.php` guards the "does this execution belong to this function?" check with `&&` instead of `||`:

```php
if ($execution->getAttribute('resourceType') !== 'functions'
    && $execution->getAttribute('resourceInternalId') !== $function->getSequence()) {
}
```

The two operands can never both be true for a function execution — `resourceType` *is* `functions`, so the first is `false` and short-circuits the whole expression. **The guard never fires for an execution belonging to a different function.**

A caller can therefore delete any execution in the project by pairing its ID with any function they can access:

```
DELETE /v1/functions/{someOtherFunctionId}/executions/{executionId}   →  204
```

Fixes #12537.

## How

One operator:

```diff
- if (... !== 'functions' && ... !== $function->getSequence()) {
+ if (... !== 'functions' || ... !== $function->getSequence()) {
```

This makes `Delete` agree with the sibling read endpoint, which already gets it right — the strongest evidence that `||` is the intended semantics:

| File | Operator |
|---|---|
| `Http/Executions/Get.php:79` | `\|\|` |
| `Http/Executions/Delete.php:89` | `&&` ← fixed here |

Non-function executions (e.g. `sites`) were already rejected by accident, so this only ever narrows access — it cannot widen it.

## Base branch

Targeting **`1.9.x`**, not `main`: on `main` this action has been restructured to delete *scheduled* executions and already uses a single correct condition, so the bug is specific to `1.9.x`. Please retarget if you'd rather take it elsewhere.

## Tests

Adds `testDeleteExecutionRequiresOwnership` to `FunctionsCustomServerTest`, covering both directions:

- **failure** — `DELETE /functions/{otherFunctionId}/executions/{executionId}` returns `404` with `execution_not_found`, and a follow-up `GET` confirms the execution is **still there** (asserting the side effect did not happen, not just the status code)
- **success** — the owning function can still delete it, so the fix doesn't over-restrict

The failure case fails on `1.9.x` today.

`composer lint` passes. `composer analyze` reports 94 errors both with and without this change on `1.9.x` — all pre-existing in my environment, none added here.

---

Supersedes #13077 (same commit, branch moved from a fork into this repo so CI can run).

### DIFF
--- a/src/Appwrite/Platform/Modules/Functions/Http/Executions/Delete.php
+++ b/src/Appwrite/Platform/Modules/Functions/Http/Executions/Delete.php
@@ -86,7 +86,7 @@ class Delete extends Base
             throw new Exception(Exception::EXECUTION_NOT_FOUND);
         }
 
-        if ($execution->getAttribute('resourceType') !== 'functions' && $execution->getAttribute('resourceInternalId') !== $function->getSequence()) {
+        if ($execution->getAttribute('resourceType') !== 'functions' || $execution->getAttribute('resourceInternalId') !== $function->getSequence()) {
             throw new Exception(Exception::EXECUTION_NOT_FOUND);
         }
         $status = $execution->getAttribute('status');
--- a/tests/e2e/Services/Functions/FunctionsCustomServerTest.php
+++ b/tests/e2e/Services/Functions/FunctionsCustomServerTest.php
@@ -1844,6 +1844,61 @@ final class FunctionsCustomServerTest extends Scope
         $this->assertStringContainsString('Can\'t delete ongoing execution.', (string) $execution['body']['message']);
     }
 
+    public function testDeleteExecutionRequiresOwnership(): void
+    {
+        $data = $this->setupTestDeployment();
+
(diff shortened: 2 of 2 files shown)


Meldiron:
Closing this in favour of #13082, which targets `main`.

To be clear about what moved and what didn't: the `&&` → `||` fix in this PR applies **only to `1.9.x`**. On `main` this action was restructured to cancel scheduled executions only, executions are no longer persisted in Server CE, and `Http/Executions/Get.php` no longer exists — so there is no buggy operator there to fix. Every analogous `resourceInternalId !== …` guard on `main` (Functions/Sites `Variables` Get/Update/Delete) already uses `||`.

#13082 therefore carries only the part that does apply to `main`: the regression test, rewritten against the scheduled-execution delete path, verified to fail when that guard is removed.

⚠️ Worth noting explicitly: **closing this leaves the `1.9.x` bug unfixed.** If `1.9.x` is still receiving patches, this one-operator change should be reopened or re-submitted there.

PR #13083, opened 2026-08-03
fix(vcs): handle zero expiry tokens during installation refresh

bhardwajparth51:
## What does this PR do?

Fixes an issue in `InstallationTokens::exchange()` where refreshing a VCS installation token with `expires_in = 0` (or an omitted `expires_in` field, common in non-expiring OAuth tokens and Personal Access Tokens) caused `DateTime::addSeconds(new \DateTime(), 0)` to store the current timestamp in `personalAccessTokenExpiry`.

On subsequent requests, `isExpired()` evaluated `stored_timestamp < now()` as `true`, treating valid non-expiring tokens as permanently expired and triggering repeated, failing refresh loops.

This PR updates `InstallationTokens::exchange()` to check `$expiry > 0` before calculating a future timestamp, ensuring `null` is persisted for non-expiring tokens (`expires_in = 0` or missing).

## Test Plan

Added unit test coverage in `tests/unit/Vcs/InstallationTokensTest.php`:
- `testRefreshedTokenWithZeroExpiryPersistsNullExpiry`: Asserts that refreshing a token with `expires_in = 0` persists `personalAccessTokenExpiry => null` on the updated document model and database call.

Run unit tests via:
```bash
docker compose exec appwrite test tests/unit/Vcs/InstallationTokensTest.php
```

## Checklist


bhardwajparth51:
@HarshMN2345 can you review this PR :)

HarshMN2345:
@bhardwajparth51 Do you think this change would actually fix the issue described in the PR?
From what I can see, ``exchange()`` is only reached after refresh() has confirmed there is a refresh token, so the GitHub App case with token expiration disabled would fail earlier because there’s no refresh token.

PR #13089, opened 2026-08-04
fix: reject hostname with empty port in validator

HemangSaraiya:
fixed- #13099
Problem:
The hostname validator allowed inputs like `example.com:` where the port is empty.

Solution:
Updated validation logic to reject hostnames with an empty port.

Changes:
- Updated validator in validation.js
- Added regression tests in installer-validation.test.js

Testing:
- Added test cases for empty port scenario
- Existing tests pass


HemangSaraiya:
Hi! This PR fixes the hostname validation issue where empty ports like `example.com:` were incorrectly accepted.
I’ve also added regression tests to ensure this doesn’t break again.
Please let me know if any changes are needed. Thanks!

PR #13137, opened 2026-08-07
fix(vcs): clone the repository an archive url cannot reach

HarshMN2345:
A Bitbucket deployment sat waiting forever with nothing built. Its source comes from an archive url the jobs-service fetches, and Bitbucket's archive host takes no credential from a url -- it answers a redirect to a login page, so the fetch lands empty and no build is ever queued.

A provider that says its archive url stands on its own is fetched as before. One that doesn't takes the path template pushes already take, where the builds worker clones with the token the git endpoint does accept.

## What does this PR do?

## Test Plan

## Related PRs and Issues

- (Related PR or issue)

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Modules/Compute/Base.php
+++ b/src/Appwrite/Platform/Modules/Compute/Base.php
@@ -168,10 +168,8 @@ class Base extends Action
         ]);
 
         // Build a plain (non-template) VCS deployment through $deployments.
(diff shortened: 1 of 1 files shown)


HarshMN2345:
Closing — main already routes clone-vs-archive inside `createFromVcs`, and the authenticated archive path is handled by header. `supportsAuthenticatedArchiveUrl` isn't present in the vendored vcs 5.5.0 this targets.

PR #13138, opened 2026-08-07
Revert "added logs"

ArnabChatterjee20k:
Reverts appwrite/appwrite#13129
It clears out the logs added to get more info about timings at different parts of the schedule


HarshMN2345:

PR #13147, opened 2026-08-08
fix(databases): type-guard the deprecated databases path like every other one

abnegate:
Fixes the remaining half of Appwrite [DAT-2219](https://linear.app/appwrite/issue/DAT-2219). Found during a production API verification sweep.

## The defect

The legacy `/v1/databases` path is exempt from the by-id type guard while its own list stays type-filtered, so the two disagree about which databases exist:

- `GET /v1/databases` lists only `legacy`-typed databases.
- `GET /v1/databases/{id}` resolves a database of **any** type — DocumentsDB, VectorsDB, TablesDB.
- `GET /v1/databases/{vectorsdb-id}/collections` then queries `database_{sequence}`, a table shaped for another product, and returns **500** rather than a typed refusal.

The sweep hit the disagreement 4 times out of 4.

`Action.php` already builds the guard and the list filter on one `getAllowedDatabaseTypes()` — its own docblock says "single source of truth ... so they cannot diverge". The legacy path simply opted out:

```php
if ($this->getDatabaseType() === DATABASE_TYPE_LEGACY) {
    return false;
}
```

## The fix

Remove the exemption, and give legacy the mirror of the set TablesDB already has:

```php
DATABASE_TYPE_TABLESDB => [DATABASE_TYPE_TABLESDB, DATABASE_TYPE_LEGACY],
DATABASE_TYPE_LEGACY   => [DATABASE_TYPE_LEGACY, DATABASE_TYPE_TABLESDB],
```

TablesDB and legacy are the same product either side of a rename, so each serves both types and refuses the other two. Two lines, and every route on the path inherits it — the guard is consulted by ~30 actions (get, delete, update, collections, attributes, indexes, documents, transactions).

Consequences:

| call | before | after |
|---|---|---|
| `GET /v1/databases` | `legacy` only | `legacy` + `tablesdb` |
| `GET /v1/databases/{tablesdb-id}` | 200 | 200 |
| `GET /v1/databases/{vectorsdb-id}` | 200 | 404 |
| `GET /v1/databases/{vectorsdb-id}/collections` | **500** | 404 |
| `DELETE /v1/databases/{vectorsdb-id}` | 204 | 404 |

Nothing the path served *correctly* starts being refused. `tablesdb` moves from "get-only" to "listed and get" — list catching up to what get already did. `documentsdb`/`vectorsdb` move from being served incorrectly to a typed not-found, and their own product routes remain the way to reach them.

### Why widen list rather than narrow get

The other direction — restricting the by-id routes to `legacy` only — would also make the two agree, but it would stop the deprecated path resolving TablesDB databases at all, which is the migration path this API is deprecated *in favour of*. It would also regress DAT-2170, where `GET /v1/databases/{id}` was fixed to report a dedicated database's real `specification`/`engine`/`replicas`.

### Red — fix reverted (`git checkout origin/main -- Action.php`)

```
Tests: 17, Assertions: 36, Failures: 6.

2) DatabaseTypeMismatchTest::testLegacyListFilterAndByIdGuardSelectTheSameSet
list and get disagree about whether a 'tablesdb' database exists on /v1/databases
Failed asserting that true is identical to false.

3) DatabaseTypeMismatchTest::testLegacyAndTablesdbPathsAcceptTheSameDatabases
the deprecated and successor paths disagree about a 'documentsdb' database
Failed asserting that false is identical to true.

4) LegacyCollectionsRefusalTest::testAVectorsDatabaseIsRefusedBeforeTheBackendIsQueried
Listing collections of a VectorsDB database on /v1/databases must be refused, not served.

6) LegacyCollectionsRefusalTest::testTheRefusedCallNeverReachesTheBackend
a 'vectorsdb' database must be refused before database_{sequence} is queried
Failed asserting that true is false.

1) DatabaseTypeMismatchTest::testLegacyPathServesTablesdbAndRefusesTheOtherProducts
Failed asserting that two arrays are identical.
 Array &0 [
     0 => 'legacy',
-    1 => 'tablesdb',
 ]
```

### Green

```
Tests: 17, Assertions: 45.
```

`LegacyCollectionsRefusalTest` drives `Collections\XList::action()` and asserts the backend is **never reached** for a mismatched type, rather than pinning the wording of whatever the engine would have thrown — the 500 came from that query, so not making it is the property that matters. It also asserts the guard does not over-refuse: `legacy` and `tablesdb` still reach the backend.

`testLegacyListFilterAndByIdGuardSelectTheSameSet` compares the list filter's `Query::equal('type', ...)` values against the by-id guard's verdict for every type, so the two cannot silently diverge again.

`testLegacyDatabasesPathIsNeverTypeGuarded` is replaced — it pinned the exempt behaviour this PR removes.

## Verification

- `composer lint` (Pint) — passed
- `composer check` (PHPStan) on the changed path — no errors
- `--testsuite unit` — no Databases-related failures

### DIFF
--- a/src/Appwrite/Platform/Modules/Databases/Http/Databases/Action.php
+++ b/src/Appwrite/Platform/Modules/Databases/Http/Databases/Action.php
@@ -26,10 +26,10 @@ class Action extends AppwriteAction
      * the list filter ({@see getDatabaseTypeQueryFilters()}) and the by-id guard
      * ({@see isDatabaseTypeMismatch()}) so they cannot diverge.
      *
-     * TablesDB is the compatibility successor to the legacy databases API, so it
-     * also serves legacy-typed databases; the other product APIs are scoped to
-     * their own type. Every database carries a non-null `type` (set on create
-     * and backfilled to 'legacy' by migration V23), so null is not represented.
+     * TablesDB and the legacy databases API are the same product either side of a
+     * rename, so each serves both types; DocumentsDB and VectorsDB are scoped to
+     * their own. Every database carries a non-null `type` (set on create and
+     * backfilled to 'legacy' by migration V23), so null is not represented.
      *
      * @return string[]
      */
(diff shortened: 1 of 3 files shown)


abnegate:
Good catch on the reachability flag — it was a real defect in the test, not a style nit.

`listCollections()` throws on exactly the path `testTheRefusedCallNeverReachesTheBackend` cares about, so the returned `['reached' => ...]` never came back and the `catch` set the flag to `false` on its own. The assertion passed whether or not `find`/`count` had run — vacuous in the green direction.

Fixed in 1344f96a83 by moving the flag onto the instance so it survives the throw, and asserting on what the action did before it threw rather than on the catch.

Verified it is meaningful in both directions now: with the guard reverted, that test still fails with `a 'vectorsdb' database must be refused before database_{sequence} is queried / Failed asserting that true is false` — i.e. `$this->reached` genuinely goes true on the unfixed path.

Also applied Rector's `createStub` for the response, which fixes the red `Checks` lane.

@greptileai review

PR #13148, opened 2026-08-08
feat: add custom MFA challenge factor

harsh-sikarwar:
## What does this PR do?

Adds a fifth MFA challenge factor, `custom`, so developers can implement arbitrary second-factor flows entirely through the server SDK — the same pattern the SDK already supports for custom sign-in (`users.createToken` / `users.createSession`), now applied to 2FA.

This lets developers build second-factor delivery via any channel Appwrite doesn't natively support (e.g. WhatsApp, an internal auth system, a third-party SMS provider) without Appwrite needing to know or care how the secret gets delivered.

**Flow:**
1. Client: `account.createMFAChallenge({ factor: 'custom' })` — a secret is generated and stored, nothing is delivered automatically (same as recovery codes/TOTP today).
2. Server (Appwrite Function, API key): `users.getMFAChallenge(userId, challengeId)` — new endpoint, reads the raw secret.
3. Server: delivers the secret via its own channel.
4. Client: `account.updateMFAChallenge({ challengeId, otp })` — verified exactly like any other factor.

**Changes:**
- `Type::CUSTOM` registered alongside `email`/`phone`/`totp`/`recoveryCode`.
- New `Challenge/Custom.php`, modeled on `Challenge/Email.php` — compares the stored `code` against the submitted `otp`.
- `Account/Http/Account/MFA/Challenges/{Create,Update}.php` extended to accept and verify the `custom` factor through the existing whitelist/match — no new branching logic needed.
- New server-only endpoint `GET /v1/users/:userId/mfa/challenges/:challengeId` (`Users/Http/Users/MFA/Challenges/Get.php`), scoped to `users.read` (API key/admin only — a client session cannot call it), and checks the challenge's `userId` to prevent reading another user's challenge.
- New `MFAChallengeSecret` response model, kept separate from the existing client-facing `MFAChallenge` model so raw secrets can never leak into the `Create`/`Update` responses used by end users.

## Test Plan

- Added `testMFACustomChallenge` to `AccountCustomClientTest.php`: happy path (create → fetch secret via the new server endpoint → verify → session `factors` includes `custom`) plus failure paths — wrong otp, reused/consumed challenge, nonexistent `challengeId`.
- Added `testGetMFAChallenge` to `UsersBase.php` (runs under both API-key and console-session server suites): happy path (server reads the decrypted secret) plus failure paths — a client session (not an API key) is rejected with `401`, a challenge belonging to a different user is rejected, nonexistent `challengeId` (`401`), nonexistent `userId` (`404`).
- Manually verified the full flow end-to-end against a local instance (curl) before writing the automated tests, including confirming the session-vs-API-key rejection live.
- `composer format` (Pint) and `composer analyze` (PHPStan level 4) pass clean on all new/changed files.
- Full existing MFA suite (`testMFARecoveryCodeChallenge`, `testMFAAuthenticatorLifecycle`, `testMFAAuthenticatorCompletesEmailPasswordSession`, etc.) still passes — no regressions.

## Related PRs and Issues

- (no linked issue yet)

## Checklist


src/Appwrite/Platform/Modules/Users/Http/Users/MFA/Challenges/Get.php
```diff
@@ -0,0 +1,86 @@
...
+
+        $document = new Document([
+            '$id' => $challenge->getId(),
+            '$createdAt' => $challenge->getCreatedAt(),
+            'userId' => $challenge->getAttribute('userId'),
+            'expire' => $challenge->getAttribute('expire'),
+            'secret' => $challenge->getAttribute('code'),
+        ]);
+
+        $response->dynamic($document, Response::MODEL_MFA_CHALLENGE_SECRET);
```
Meldiron: We could return $challenge directly - response models will take care of hiding that's not expected to be returned

src/Appwrite/Platform/Modules/Users/Http/Users/MFA/Challenges/Get.php
```diff
@@ -0,0 +1,86 @@
...
+            ->desc('Get MFA challenge')
+            ->groups(['api', 'users'])
+            ->label('scope', 'users.read')
+            ->label('usage.metric', 'users.{scope}.requests.read')
+            ->label('sdk', [
+                new Method(
+                    namespace: 'users',
+                    group: 'mfa',
+                    name: 'getMFAChallenge',
+                    description: '/docs/references/users/get-mfa-challenge.md',
```
Meldiron: Let's avoid MD file, instead, let's have description in-line here. You can ask agent to find some endpoint that already does it, and follow that pattern

src/Appwrite/Platform/Modules/Users/Http/Users/MFA/Challenges/Get.php
```diff
@@ -0,0 +1,86 @@
...
+
+    public function __construct()
+    {
+        $this
+            ->setHttpMethod(Action::HTTP_REQUEST_METHOD_GET)
+            ->setHttpPath('/v1/users/:userId/mfa/challenges/:challengeId')
+            ->desc('Get MFA challenge')
+            ->groups(['api', 'users'])
+            ->label('scope', 'users.read')
+            ->label('usage.metric', 'users.{scope}.requests.read')
```
Meldiron, line 34: Prompt agent to double-check this section specifically, please.
I feel like there should be a bit more here (labels), but I might be totally wrong. So just 1 prompt for peace of mind.
Checking against other endpoints for insiration should be enough

src/Appwrite/Utopia/Response/Model/MFAChallengeSecret.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
+namespace Appwrite\Utopia\Response\Model;
+
+use Appwrite\Utopia\Response;
+use Appwrite\Utopia\Response\Model;
+
+class MFAChallengeSecret extends Model
```
Meldiron: Let's make this response model extend MFAChallenge response model. That way, we should only need to add rule for the secret that is not exposed to client.

For inspiration, I think Users and Account model do something similar - when server can see password hash, but client cannot.

tests/e2e/Services/Account/AccountCustomClientTest.php
```diff
@@ -4271,6 +4271,91 @@ public function testMFARecoveryCodeChallenge(): void
...
+    {
+        $projectId = $this->getProject()['$id'];
+        $apiKey = $this->getProject()['apiKey'];
+
+        // Create custom factor challenge using existing authenticated session
+        $challenge = $this->client->call(Client::METHOD_POST, '/account/mfa/challenge', array_merge([
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $projectId,
+        ], $this->getHeaders()), [
+            'factor' => 'custom'
```
Meldiron, line 4284: Comment for myself, Ill check with founder and other leads if `custom` is best name we can do

tests/e2e/Services/Account/AccountCustomClientTest.php
```diff
@@ -4271,6 +4271,91 @@ public function testMFARecoveryCodeChallenge(): void
...
+        ]);
+
+        $this->assertEquals(201, $challenge['headers']['status-code']);
+        $this->assertNotEmpty($challenge['body']['$id']);
+        $challengeId = $challenge['body']['$id'];
```
Meldiron, line 4291: Let's add one more assertion here, making sure `secret` is not  there in response. That's very important for this flow

Meldiron (commented):
amazing PR! Left only minor comments

harsh-sikarwar:
Thank you for the suggestions,
I am on it,

commiting the changes in a moment

harsh-sikarwar:
Pushed a follow-up commit addressing all the review feedback:

**1. Return `$challenge` directly**
`Get.php` no longer manually rebuilds a `Document` field-by-field. It now aliases `code` → `secret` on the fetched challenge (`$challenge->setAttribute('secret', $challenge->getAttribute('code'))`) and passes the challenge straight through — `Response::dynamic()` only serializes fields the target model declares via `addRule()`, so nothing extra leaks and the response shape is unchanged.

**2. Inline description instead of a `.md` file**
Replaced the `/docs/references/users/get-mfa-challenge.md` reference with an inline sentence. Checked the rest of the codebase first — every endpoint using an inline `description:` string (`Console/Http/Variables/Get.php`, `Console/Http/Scopes/*`, the Project OAuth2 provider config endpoints, VCS installation namespaces) is internal/admin-only with no corresponding public docs page. Same situation here, so no markdown file should exist for this route. This also happens to be what was failing CI's spec generation — that file never existed.

**3. Double-checked the `label()` chain**
Compared against `RecoveryCodes/Get.php`, its closest sibling (same shape: per-user, `scope: users.read`, sensitive sub-resource read). Every `users.write` (mutating) MFA endpoint in this module has `audits.*` labels; the one other `users.read` endpoint has none. Ours matches — audits track changes, not reads, so nothing was actually missing.

**4. `MFAChallengeSecret` now extends `MFAChallenge`**
Instead of duplicating `$id`/`$createdAt`/`userId`/`expire`, it now calls `parent::__construct()` and adds only the `secret` rule — same relationship as the existing `EphemeralKey extends Key` in this codebase.

**5. Test additions in `testMFACustomChallenge`**
- Client session (no API key) hitting the server endpoint → asserted `401`.
- Submitting a wrong/random otp *before* the real one is used → asserted `401` (confirmed live against the running endpoint — it's `401 user_invalid_token`, not `400`).
- After a successful verify, asserted the session response's `secret` field is present but empty — `Session` already marks that field `sensitive`, so it gets forced to `''` for non-privileged/non-key callers. This proves the MFA secret never round-trips back through the normal client-facing verify call either, not just the server endpoint.

All existing MFA tests (Account + Users suites) still pass, no regressions. Pint and PHPStan (level 4) clean on every touched file.

Meldiron (approved):
Beautiful! 👏

PR #13174, opened 2026-08-10
DISCORD-15363566 fix(databases): support text, varchar and format column types inline

claudear:
## What

`POST /v1/tablesdb/:databaseId/tables` (and the legacy `createCollection`) rejected `text` — and every other non-legacy string type — in the inline `columns`/`attributes` array:

```
Appwrite\Extend\Exception: Invalid type for attribute 'modulePath': text
  at src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Create.php:145
```

`Appwrite\Utopia\Database\Validator\Attributes::$supportedTypes` hardcoded the legacy type list and was never updated when the `text`/`varchar`/`mediumtext`/`longtext` family was introduced, even though `createTextColumn`, `createVarcharColumn` et al. create those exact types. The failure isn't free either: the metadata row is created before validation runs and then deleted, so callers get a partially-executed request and have to fall back to `createTable` + N separate column calls.

The reporter's `type: "email"` also failed, for a related reason: `email` is a string *format*, not a type, and the inline surface had no way to express it without also knowing the magic size.

## How

The size defaults that make these types work only lived in the dedicated endpoints, so widening the type list alone wasn't sufficient. The mapping now lives in one place, `Appwrite\Utopia\Database\Attribute`, consumed by both the validator and the inline attribute builder so the two surfaces produce identical attribute documents and can't drift again:

- `text` → 65535, `mediumtext` → 16777215, `longtext` → 2147483647 (fixed width; a caller-supplied `size` is ignored)
- `varchar` takes and requires a `size`, like `string`
- `email`/`url`/`ip`/`enum` are accepted as a `type` shorthand that normalizes to `{type: string, format: <x>, size: <default>}`, matching `createEmailColumn` and friends
- a `string` with an explicit `format` no longer needs a `size` — it falls back to the format's default
- defaults on the text types are validated as strings against the type's size

No behavior change to the dedicated per-column endpoints. An explicit `size` alongside a format is still honored, so existing `{type: 'string', format: 'email', size: 512}` payloads are unaffected.

The `columns`/`attributes` param descriptions listed only the legacy types; they now match what's accepted.

## Tests

- `tests/unit/Utopia/Database/Validator/AttributesTest.php` — new. Written first and confirmed failing on the four reported cases (`varchar`, `text`, `email` shorthand, `string` + `format` without `size`) before the fix, passing after.
- `tests/e2e/Services/Databases/TablesDB/TablesDBColumnsTest.php` — new. Creates a table with all nine types inline, asserts each stored type/size/format matches the dedicated endpoints, and writes a row through it.

Full unit suite, Pint, PHPStan and Rector all pass on the changed files.

Fixes DISCORD-15363566


claudear:
## Fix Confidence: 80/100

High confidence in the diagnosis and the core fix: the root cause was verified directly in code, the unit test was confirmed failing before the change and passing after, and the size/format constants were copied from the dedicated endpoint implementations rather than guessed. The full unit suite shows no regressions attributable to the change.

Deductions: (1) the E2E Databases job had not reported on the corrected commit when I finished, so the revised e2e test is unverified end-to-end — the first run already caught one wrong assumption in it, which is direct evidence my e2e assumptions are not fully reliable. (2) Local static analysis and unit runs used a stale vendor directory, so a small number of errors were unverifiable locally and I am relying on CI for those. (3) Accepting `email`/`url`/`ip`/`enum` as `type` shorthands is a deliberate widening of the inline API surface beyond the strictly minimal fix; it directly addresses the reporter's payload and mirrors the existing per-column endpoints, but it is a judgment call a maintainer could reasonably scope down.

abnegate (approved):

PR #13180, opened 2026-08-11
feat(compute): validate variable keys as environment identifiers

levivannoort:
## What

Function, site and project **variable keys** are used as environment variable names at build and runtime, so they must be valid C-style identifiers (`^[A-Za-z_][A-Za-z0-9_]*$`). Today the `key` param is only length-checked (`new Text(...)`), so a key with a **trailing tab** or an **accented letter** (e.g. `CAFÉ_TOKEN`, or `API_TOKEN` with a trailing tab) passes validation, gets stored, and then fails when the build environment is assembled.

That breaks every deployment for the affected resource, and it can't be caught after the fact — the duplicate-deployment path re-reads the same stored key.

## Change

- Validate the `key` param with `Utopia\Validator\Identifier` across the **six** variable write endpoints:
  - `Functions/Http/Variables/{Create,Update}`
  - `Sites/Http/Variables/{Create,Update}`
  - `Project/Http/Project/Variables/{Create,Update}`
- Bump `utopia-php/validators` `0.4.1` → `0.4.2`, which introduces `Identifier`. (`composer.json` already allowed `^0.4`; this is a lock-only bump.)

## Behavior

`Identifier` extends `Text`, so length still applies; it additionally requires a C-style identifier. Creates reject non-conforming keys; Updates keep the key optional (`Nullable`) and validate when provided. `value` continues to use `Text`.

## Notes

- Not covered by this change (separate follow-ups): the project-variable **migration import** writes direct-to-DB and bypasses endpoint validation, and **already-stored** bad keys need a one-time data cleanup.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -81,7 +81,7 @@
         "utopia-php/dsn": "0.2.1",
         "utopia-php/http": "^2.0@RC",
         "utopia-php/fetch": "^1.1",
-        "utopia-php/validators": "0.3.*",
+        "utopia-php/validators": "^0.4",
         "utopia-php/image": "0.8.*",
         "utopia-php/locale": "0.8.*",
         "utopia-php/lock": "0.2.*",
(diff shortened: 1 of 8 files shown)


levivannoort:
Superseded by #13181 (rebased cleanly on main, no force-push needed). Closing this one.

PR #13185, opened 2026-08-12
Fix TablesDB create action parameter names

ChiragAgg5k:
## What does this PR do?

Adds explicit TablesDB action adapters where the public API uses TablesDB terms but the shared Databases implementation still uses legacy collection/attribute terms.

```php
// public TablesDB params
array $columns

// shared Databases implementation
parent::action(..., $columns, ...); // parent receives this as $attributes
```

This covers:
- `tablesDB.createTable`: `columns` -> parent `$attributes`
- `tablesDB.createIndex`: `columns` -> parent `$attributes`

It also adds a unit guard that scans Databases HTTP actions and fails on future inherited callbacks with unexpected param-name renames unless they either override `action()` or are explicitly allowlisted.

## Test Plan

```text
php -l src/Appwrite/Platform/Modules/Databases/Http/TablesDB/Tables/Create.php
composer lint src/Appwrite/Platform/Modules/Databases/Http/TablesDB/Tables/Create.php
composer analyze src/Appwrite/Platform/Modules/Databases/Http/TablesDB/Tables/Create.php
php -l src/Appwrite/Platform/Modules/Databases/Http/TablesDB/Tables/Indexes/Create.php
composer lint src/Appwrite/Platform/Modules/Databases/Http/TablesDB/Tables/Indexes/Create.php
composer analyze src/Appwrite/Platform/Modules/Databases/Http/TablesDB/Tables/Indexes/Create.php
composer test -- tests/unit/Platform/Modules/Databases/ActionParameterNamesTest.php
composer analyze tests/unit/Platform/Modules/Databases/ActionParameterNamesTest.php
```

Could not run Docker e2e locally because the Docker daemon was unavailable.

## Related PRs and Issues

- #XXXX

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Modules/Databases/Http/TablesDB/Tables/Create.php
+++ b/src/Appwrite/Platform/Modules/Databases/Http/TablesDB/Tables/Create.php
@@ -2,6 +2,7 @@
 
 namespace Appwrite\Platform\Modules\Databases\Http\TablesDB\Tables;
 
+use Appwrite\Event\Event;
 use Appwrite\Platform\Modules\Databases\Http\Databases\Collections\Create as CollectionCreate;
 use Appwrite\SDK\AuthType;
 use Appwrite\SDK\ContentType;
@@ -10,6 +11,7 @@ use Appwrite\SDK\Response as SDKResponse;
 use Appwrite\Utopia\Database\Validator\CustomId;
 use Appwrite\Utopia\Response as UtopiaResponse;
 use Utopia\Database\Database;
+use Utopia\Database\Validator\Authorization;
 use Utopia\Database\Validator\Permissions;
 use Utopia\Database\Validator\UID;
 use Utopia\Http\Adapter\Swoole\Response as SwooleResponse;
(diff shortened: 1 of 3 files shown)


ChiragAgg5k:
Useful adapters, and thanks for putting this up — it corrects a claim I got wrong. Investigating an [MCP benchmark](https://6a79b8b74e1f6bf83f68.appwrite.network/) finding, I read the `->param('columns')` / inherited `action(…, array $attributes, …)` mismatch as a live binding bug and wrote it up as High severity. It isn't: Utopia builds the argument list as `$arguments[$param['order']] = $value` and calls it through `call_user_func_array`, so binding is positional by declaration order and the callback's parameter names never matter. `columns` arrives correctly today. I've corrected that in our notes and in the PR where I asserted it.

One thing that might be worth folding in while the terminology is being cleaned up: the runtime messages are, I think, what actually produces this confusion for callers. `Attributes` reports every failure in collection vocabulary regardless of which API was called:

```php
"Invalid type for attribute 'rating': float"          // Validator/Attributes.php:136
"Attribute at index 0 is missing required field 'type'" // :110
"Duplicate attribute key: ..."                          // :109
```

So a caller who sends `tablesDB.createTable` the `type: "float"` that the param description advertises gets back an error naming `attribute` — and reasonably concludes the parameter should have been `attributes`. That is very likely the whole of the benchmark's "documentation lists `columns` but API requires `attributes`" finding: two separate papercuts reading as one.

The trigger is a genuine bug, separately: the description advertises `float`, but the allowlist is `Database::VAR_FLOAT`, which is `'double'` (`Response/Model.php:11`). It also omits `bigint`, `point`, `linestring` and `polygon`, which the validator does accept, and lists `relationship`, which it rejects inline. Happy to open that separately if you'd rather keep this PR to the signatures.

The unit guard against future inherited-callback renames is a nice touch — it makes the positional-binding behaviour explicit for the next person who reads it the way I did.

ArnabChatterjee20k (approved):

PR #68, opened 2019-10-01
Added Hindi Email Template and other related things for Hindi Language.

sakkshm:
I have added the Hindi language to the template and the other files needed to the code.

PR #13193, opened 2026-08-12
Add /v1/sandbox: isolated workspaces backed by the open-runtimes orchestrator

loks0n:
## What does this PR do?

Adds `/v1/sandbox`: live, isolated workspaces a caller creates on demand, runs commands in, and tears down. Where a function runs to completion and a site serves traffic, a sandbox just sits there and waits for you — the shape an AI agent, a notebook kernel, or an interactive build wants.

It is backed by the [open-runtimes orchestrator](https://github.com/open-runtimes/orchestrator) v1.4.0.

## API surface

| | | |
| --- | --- | --- |
| `POST` | `/v1/sandbox` | Create a sandbox → `201` |
| `GET` | `/v1/sandbox` | List the project's live sandboxes → `200` |
| `GET` | `/v1/sandbox/:sandboxId` | One sandbox's status → `200` |
| `DELETE` | `/v1/sandbox/:sandboxId` | Tear it down → `204` |

Guarded by two new scopes, `sandboxes.read` and `sandboxes.write`, on admin + API key auth.

### Create parameters

| Param | Default | Notes |
| --- | --- | --- |
| `sandboxId` | `unique()` | Lowercase alphanumeric with interior hyphens, ≤36 chars. |
| `image` | **required** | Container image to start from. Any runtime image works — see the contract below. |
| `specification` | modest | Compute size, from the existing `app/config/specifications.php` slugs (`s-1vcpu-1gb` and friends) — the same slugs functions and sites take, validated against `_APP_COMPUTE_CPUS`/`_APP_COMPUTE_MEMORY` and the billing plan, then resolved to the cpu and memory the orchestrator sizes the container with. |
| `command` | — | Overrides the installed sandbox agent. Must serve the contract on port 3000, which is reserved rather than a parameter. |
| `variables` | — | Environment variables for the workload. |
| `ports` | — | Up to 16 extra ports beyond the contract, each addressable at its own hostname in `urls`. |
| `timeout` | `300` | Per-request bound in seconds for calls to the sandbox URL. `0` removes it, for long-lived connections such as WebSockets. Max 3600. |
| `idleTimeout` | `900` | Seconds without traffic before teardown. `0` keeps it alive until deleted. |

Unlike functions and sites, an unstated `specification` resolves to the modest fallback (`s-1vcpu-512mb`) rather than the largest allowed size — sandboxes are created ad hoc and in volume, so silently handing out the biggest box on the host is a footgun. Cloud plans still cap the range as they do elsewhere. This is the first caller of `getDefaultSpecification(preferFallback: true)`.

The response model is the same on every read: `$id`, `status` (`creating` \| `ready` \| `failed` \| `deleting`), `url`, `urls` (every served port, keyed by port number), `image`, `specification`, `ports`, and `error`.

`image`, `specification` and `ports` describe the shape the sandbox is actually running in, so a caller reading one back does not have to keep its own record of what it asked for. The orchestrator reports cpu and memory ([open-runtimes/orchestrator#74](https://github.com/open-runtimes/orchestrator/pull/74), in 1.4.0), recorded on the pod at create so a re-imaged or deleted pool cannot rewrite history; Appwrite names that pair back as a specification slug, leaving it blank for a sandbox sized outside the catalogue.

## The sandbox contract lives inside the sandbox

Exec and files are **not** part of this API and never pass through Appwrite. They are an HTTP contract served *inside* the sandbox at its own URL:

| | |
| --- | --- |
| `POST /execute` | `{"command", "timeoutSeconds"}` → `{"exitCode", "stdout", "stderr", "durationMillis", "truncated"}` |
| `GET\|PUT\|DELETE /files/{path}` | whole-file read / write / remove, relative to the workspace |

Your image implements none of it — the orchestrator installs the [`open-runtimes/sandbox`](https://github.com/open-runtimes/sandbox) agent into every sandbox's workspace, so `python:3.12-slim` or a distroless image serves the contract with nothing added. Keeping Appwrite off the data path means a long exec is not killed by a rolling restart and a large upload is not a shared-fate bottleneck.

```bash
# create
curl -X POST https://cloud.appwrite.io/v1/sandbox \
  -H "X-Appwrite-Project: $PROJECT" -H "X-Appwrite-Key: $KEY" \
  -d '{"sandboxId":"agent-run-42","image":"python:3.12-slim","ports":[5173]}'

# → {"$id":"agent-run-42","status":"ready","url":"https://s-9f3c….sandboxes.appwrite.run", …}

# drive it at its own URL
curl -X POST $URL/execute -d '{"command":"python -m pytest -q","timeoutSeconds":120}'
curl -X PUT  $URL/files/main.py --data-binary @main.py
```

## Design

**Persistent storage per sandbox.** The workspace is scratch and dies with the sandbox, except `/workspace/persistent` — a Docker volume (or PVC) mounted into every sandbox, named `{_APP_SANDBOX_VOLUME}-p{projectInternalId}-{sandboxId}` and created on first use. Delete a sandbox and create it again under the same ID and its files are back; a different ID starts empty. The name carries the project prefix, so no two tenants can reach the same filesystem.

Setting `_APP_SANDBOX_VOLUME` empty turns it off, which is what a Kubernetes install wants until it has a PVC to point at — the orchestrator attaches storage but never creates it, so a name resolving to nothing would fail the create.

Three honest limits. Docker creates the volume root-owned, so an image whose workload runs as a non-root user cannot write there. Nothing reaps a volume: **one is left behind per sandbox ID**, including the generated IDs nobody will reuse, so a busy project accumulates them until something prunes. And storage keyed on an ID the caller did not choose is only useful to a caller who passes `sandboxId` explicitly.

**Stateless pass-through.** The orchestrator reconstructs sandbox state from the backend on every read, so Appwrite stores nothing: no collection, no worker, no events, no scheduled cleanup. A sandbox outlives an Appwrite restart, and an abandoned one is collected by `idleTimeout` rather than by us.

**Cross-project isolation** is enforced by namespacing every orchestrator id as `p{projectInternalId}-{sandboxId}`. List filters by that prefix; get and delete cannot address another tenant's sandbox even with a guessed id; two projects can independently hold the same `sandboxId`.

**The URL is a capability.** Its hostname carries a 128-bit token, and reaching it is enough to run code inside — so `url` and `urls` are marked `sensitive` in the response model, and `DELETE` invalidates the token along with the pod.

**No per-sandbox proxy rule.** Unlike a function's domain, nothing is written to the `rules` collection and Appwrite's router is not involved: the capability token in the Host is what resolves a sandbox, inside the orchestrator. Compose carries one wildcard Traefik route for the whole `s-{token}` grammar, created once, so creating a sandbox stays a single call with nothing to reconcile at the edge.

New error types: `sandbox_not_found` (404), `sandbox_already_exists` (409), `sandbox_limit_exceeded` (429).

## Configuration

- `_APP_SANDBOX_HOST` — the orchestrator (default `http://orchestrator:8080`). Left resolvable when unset so installs without the service still boot, failing at call time instead.
- `_APP_DOMAIN_SANDBOXES` — the wildcard domain URLs are minted under (default `sandboxes.localhost`), needing a `*.{domain}` DNS record and, in production, a certificate.
- `_APP_SANDBOX_VOLUME` — name prefix for each sandbox's persistent volume (default `appwrite-sandbox`); empty disables persistent storage.
- `_APP_SANDBOX_SECRET` — bearer token for that service.
- The `orchestrator` compose service came with #13290; this PR adds only its sandbox-side configuration — `SANDBOX_DOMAIN`, `DATA_PORT`, and the wildcard Traefik route — and moves the images to 1.4.0. No sandbox pools are declared: every create is image-based.

The orchestrator calls go through `open-runtimes/sdk-for-php` 0.12.0, whose `Sandboxes` client landed alongside this work — statuses arrive as typed models and errors as its `ApiException`, which the module maps to Appwrite's own. The bump is additive for the existing `Jobs` usage: identical `create()` signature, and `CallbackEvent` only gains two cases the build pipeline does not subscribe to.

**This moves the build pipeline off `jobs-service:0.18.0`**, since jobs now come from the same container. That is currency, not a fix — the jobs wire contract is unchanged across the range (identical callback event set, identical request JSON, no diff in `docs/callbacks.md`), and I verified a function still builds and executes on it.

## Deliberately not exposed

- **Pools.** Warm-pod claiming is operator configuration with sub-second creates; every sandbox here is image-based and pays a cold start. Nothing about the API surface changes if we add pools later. (Making pools genuinely optional upstream came out of this work — orchestrator [#67](https://github.com/open-runtimes/orchestrator/pull/67) → shipped as [#68](https://github.com/open-runtimes/orchestrator/pull/68) in v1.1.0.)
- **Artifacts**, the bulk-in path for large inputs. `PUT /files/{path}` covers the common case.
- The `sandbox` service ships `'optional' => false` (like `presences`) until the generated SDK's `ProjectServiceId` enum learns the key.

## Test plan

**Automated** — `tests/e2e/Services/Sandbox/SandboxCustomServerTest.php` (11 tests) covers create → ready with URL, duplicate → 409, missing image → 400, invalid id → 400, a valid and an out-of-range `specification`, get, project-scoped list, delete → 204 → 404, and persistence: a file written to `/workspace/persistent` survives teardown and reappears in a sandbox recreated under the same ID on a new URL, while the rest of the workspace does not and a different ID sees empty storage.

**Manual, against `orchestrator:1.4.0`** — 27 checks over the HTTP API and 11 over the contract inside a live sandbox:

- *Authorization*: keys missing `sandboxes.read`/`sandboxes.write` and anonymous requests are rejected.
- *Isolation*: project B cannot list, get, or delete project A's sandbox; A's sandbox survives B's delete attempt; both projects can hold the same id at once.
- *Contract*: `/healthz`; exec covering stdout, stderr, and exit codes; `variables` reaching the workload; file write → read → execute → delete → 404; a dev server started on a declared extra port is reachable at its own hostname while an undeclared port 404s.
- *Path safety*: `%2e%2e%2f` traversal is refused with 400 and the raw form dead-ends at 404, leaking nothing.
- *Sizing*: `docker inspect` on the workload container confirms the requested specification lands as real limits — `s-0.5vcpu-512mb` → 0.5 cpus/512 MB, `s-2vcpu-2gb` → 2 cpus/2048 MB, unstated → 1 cpu/512 MB.
- *Reachability*: the minted URL answers from outside Docker through Traefik with no DNS tricks — `/healthz`, `/execute`, and a dev server on a declared extra port, while an undeclared port still 404s.
- *Teardown*: delete → 204, then 404, empty list, capability URL dead (503), no leftover containers.
- *Builds*: a `node-22` function builds to `ready` and executes `200 OK` — from the same container serving the sandboxes, confirming both planes work at once.


app/init/resources.php
```diff
@@ -88,6 +89,21 @@
     return new Jobs($client);
 }, []);
 
+$container->set('sandboxes', function () {
+    $client = (new Client(new CurlAdapter()))
+        ->withBearerAuth(System::getEnv('_APP_SANDBOX_SECRET', ''))
+        ->withTimeout(30);
```
eldadfux, line 96: should we try to match the CDN http timeout and have something consistent across the stack? IIRC 60s?

app/config/variables.php
```diff
@@ -1053,6 +1053,25 @@
                 'question' => '',
                 'filter' => ''
             ],
+            [
+                'name' => '_APP_SANDBOX_HOST',
```
eldadfux, line 1067: this is equivalent to the functions/sites executor, correct? 
does it have to be a separate container?

src/Appwrite/Platform/Modules/Sandbox/Http/Create.php
```diff
@@ -0,0 +1,121 @@
...
+            ->param('image', '', new Text(256), 'Container image to start the sandbox from.')
+            // Sandboxes are created ad hoc and in volume, so an unstated size
+            // is the modest fallback rather than the largest box available.
+            ->param('specification', fn (array $plan) => $this->getDefaultSpecification($plan, preferFallback: true), fn (array $plan) => new Specification(
+                $plan,
+                Config::getParam('specifications', []),
+                System::getEnv('_APP_COMPUTE_CPUS', 0),
+                System::getEnv('_APP_COMPUTE_MEMORY', 0)
+            ), 'Compute specification sizing the sandbox.', true, ['plan'])
+            ->param('port', 3000, new Range(1, 65535), 'Port the sandbox contract is served on.', true)
```
eldadfux: Not loving the port vs ports, maybe we should just have 3000 as reserved port and let devs only decide it they want to expose more ports? or just have 3000 set as default as part of `ports`?

Is this also coupled with the proxy rules collection? do we have to make people create a sandbox and then the rules to actually have working URLs for these ports?

src/Appwrite/Utopia/Response/Model/Sandbox.php
```diff
@@ -0,0 +1,69 @@
...
+            ])
+            ->addRule('status', [
+                'type' => self::TYPE_ENUM,
+                'description' => 'Sandbox status. Possible values: `creating`, `ready`, `failed`, `deleting`.',
+                'default' => '',
+                'example' => 'ready',
+                'enum' => ['creating', 'ready', 'failed', 'deleting'],
+                'enumSDKName' => 'SandboxStatus',
+            ])
+            ->addRule('url', [
```
eldadfux, line 27: URL and URLs are breaking our pattern for functions/sites + proxy rules. Any strong reason not to keep them separate on the API level in the same way? On the UI side we can completely abstract this and handle it for users. Can the model include more information like ports, image, specification, etc?

app/config/variables.php
loks0n: It is equivalent 👍 

I've gone ahead and combined them

src/Appwrite/Utopia/Response/Model/Sandbox.php
loks0n: Done for the second half — the model now carries `image`, `specification` and `ports`.

You were right that the data should be there. The orchestrator wasn't returning it, but it was storing the whole create request all along and parsing it back on every read just to rebuild URLs, so the shape was sitting one field away unread. Added it upstream (open-runtimes/orchestrator#74, released in 1.4.0) and wired it through here.

Two notes on how it's reported:

- `specification` is named back rather than echoed. The orchestrator speaks cpu/memory, so we reverse-map the pair onto the catalogue in `app/config/specifications.php`. A sandbox sized outside it is left blank rather than guessed at.
- `ports` comes off the URLs the sandbox serves, so it can't drift from `urls`.

Review also caught two bugs in that upstream change before it shipped: the Kubernetes create path returned a status without the fields that an immediate GET included, and both read paths took the shape from the pool *as configured now* — so re-imaging a pool under a running sandbox reported the replacement, and deleting the pool made the fields vanish. The shape is now recorded on the pod at create and read straight back, so it describes the pod that's running.

The first half — whether `url`/`urls` should follow the functions/sites proxy-rules pattern — I've left alone and replied to separately.

src/Appwrite/Platform/Modules/Sandbox/Http/Create.php
loks0n: Both done.

**port vs ports** — agreed, the pair was asking callers to state something they never choose. 3000 is now reserved as the contract port and `port` is gone; `ports` means extra ports only. Overriding `command` means serving the contract on 3000, which the param description says.

I didn't fold 3000 into `ports` as a default because it makes the contract port positional — `ports: [3000, 5173]` gives no hint which entry is special, and a caller passing `ports: [5173]` would silently drop it. It is still reported: `ports` on the response includes 3000, and `urls` is keyed by port.

**proxy rules** — no coupling, and no per-sandbox rule. Nothing is written to the `rules` collection and Appwrite's router is never on the path: a sandbox URL carries a 128-bit capability token in its hostname, and the orchestrator resolves it internally. There's no rule row to derive it from, which is also why the URL has to come back on the response rather than being composed client-side.

You did catch a real gap though — the URLs weren't reachable at all. `sandboxes.test` wasn't real DNS and the data plane wasn't fronted, so the API was handing out URLs nobody could open. Fixed with **one** wildcard route created once, not per sandbox: `_APP_DOMAIN_SANDBOXES` (default `sandboxes.localhost`), the orchestrator's data plane on port 80 so the port stays out of minted URLs, and a Traefik router matching the `s-{token}` grammar. Verified from outside Docker with plain curl — `/healthz`, `/execute`, and a dev server on a declared extra port all answer, while an undeclared port still 404s.

Production still needs the wildcard DNS record and a cert for `*.{domain}` — that's an operator step, not a per-sandbox one.

PR #936, opened 2021-02-23
Improved Contributing.md

prakharshreyash15:
## What does this PR do?

(Provide a description of what this PR does.)

## Test Plan

(Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified your changes work.)

## Related PRs and Issues

(If this PR is related to any other PR or resolves any issue or related to any issue link all related PR and issues here.)

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

(Write your answer here.)

Issue #733, opened 2020-11-11
On fresh install, clamav container fills up the disk


akago-dev:
## 🐛 Bug Report

After a fresh install, appwrite is unusable after a few minutes due to a full disk (25 gb)

## Have you spent some time to check if this issue has been raised before?

yes

## To Reproduce

On a brand new digitalocean droplet (ubuntu 20.04 LTS) : 
1. install appwrite with the docs command. 
2. No custom options
3. Create an account, a project, and a collection (not sure if necessary to reproduce)
4. Check the disk usage with : `docker system df -v`

The disk should start to fill

## Expected behavior

Appwrite should run without fills the disk if no files were uploaded

## Actual Behavior

The disk started to fill because of the clamav container 

After a few minutes, with no uploaded files : 
![image](https://user-images.githubusercontent.com/56194420/98796114-538e3180-240b-11eb-995a-d89e3e7ba768.png)

## Your Environment

DigitalOcean droplet 
1 GB Memory / 25 GB Disk / FRA1 - Ubuntu 20.04 (LTS) x64
No specific configuration
Commands ran by root user

akago-dev:
@eldadfux told me to try those two solutions : 
1. Disable ClamAV using our env var, available here: https://appwrite.io/docs/environment-variables check the `_APP_STORAGE_ANTIVIRUS` option. This will make sure you can still use the storage API with clamav container stopped.

2. Update the ClamAV container to version 1.2.0 from you docker-compose.yml file and recreate it using `docker-compose up -d`


I tried to update the clamav container to 1.2.0 but got an issue on clamav startup : 
![image](https://user-images.githubusercontent.com/56194420/98796604-dca56880-240b-11eb-9756-714c4e291d18.png)

eldadfux:
Should be fixed in v0.7. Make sure to have enough free memory/cpu on your system or disable clamav.

PR #13195, opened 2026-08-13
fix(databases): bump utopia-php/mongo to 1.5.3 and utopia-php/database to 7.2.1

### LINKED ISSUE #13175: BigInt[] Breaks MongoDb
### 👟 Reproduction steps

1. Self-host Appwrite **1.9.6** with the **MongoDB** database backend.
2. Create a table and add an array column of type `integer` or `bigint`:

```bash
# BigInt array
sudo appwrite databases create-big-int-attribute \
  --database-id "$DATABASE_ID" \
  --collection-id "$COL_ID" \
  --key "listOfIntegers" \
  --array true \
  --required true

# or Integer array
sudo appwrite databases create-integer-attribute \
  --database-id "$DATABASE_ID" \
  --collection-id "$COL_ID" \
  --key "listOfIntegers" \
  --array true \
  --required true
```

3. Using the Dart server SDK, create a row that inserts a large value into the array column, e.g.:
```
[-3408048000]
```
_(The write succeeds with no error.)_

4. Open the table’s Rows tab in the Appwrite Console.

### 👍 Expected behavior

- The row should be created successfully if BigInt data type (when inserting `[-3408048000]`) but fail when inserting into integer data type (because it's too small).
- Listing rows in the Console (or via the API) should return the data correctly, including large integer / bigint values inside arrays.
- Invalid values that exceed the type limits should be rejected at write time with a clear validation error instead of silently corrupting the table.

### 👎 Actual Behavior

The write via the Dart SDK succeeds (no error is returned even if inserting an integer to larger for the data type ex: `-3408048000` for integer).

Any subsequent attempt to list the rows fails with a 500 error:
```JSON
{
  "message": "Server Error",
  "code": 500,
  "type": "general_unknown",
  "version": "1.9.6"
}
```
Docker logs show:
```
http.request · 5.2ms · 59ffc126

  level              error
  http.method        GET
  http.path          /v1/tablesdb/:databaseId/tables/:tableId/rows
  http.response.code 500

  Error: Cannot use object of type MongoDB\BSON\Int64 as array
    at /usr/src/code/vendor/utopia-php/database/src/Database/Document.php:49
    at /usr/src/code/vendor/utopia-php/database/src/Database/Adapter/Mongo.php:2236
    at /usr/src/code/vendor/utopia-php/database/src/Database/Adapter/Pool.php:79
    at /usr/src/code/vendor/utopia-php/pools/src/Pools/Pool.php:229
    ... 13 more
```
The table becomes unusable in the Console until the entire table is deleted.
The same value works correctly when stored in a non-array integer or bigint column. Which I think `-3408048000` shouldn't be able to be stored in an integer data type, but non-array integer set via console seems to work.
Small values such as [-42] work fine in the array column when set via console or dart cloud function.

### 🎲 Appwrite version

### 💻 Operating system

MacOS

### 🧱 Your Environment

appwrite 1.9.6
mongodb

abnegate:
Fixes #13175 — *BigInt[] Breaks MongoDb*.

## What users hit

Store an integer outside the int32 range in an array column on MongoDB, for example `[-3408048000]`, and the write succeeds. Every subsequent `listRows` on that table returns:

```json
{ "message": "Server Error", "code": 500, "type": "general_unknown" }
```

with `Cannot use object of type MongoDB\BSON\Int64 as array` in the logs. The table stays unusable in the Console until it is deleted. `[-42]` works, which is what made this look value-dependent rather than like a type bug.

## Cause

`MongoDB\BSON\Document::toPHP()` decodes every 64-bit BSON integer as a `MongoDB\BSON\Int64` wrapper, on all platforms. A small value encodes as int32 and decodes to a plain `int`; anything past int32 encodes as int64 and comes back as an object. utopia-php/mongo was handing those objects to callers.

Fixed at the decode boundary in utopia-php/mongo#49.

## What this bumps

Lock only, two packages, no `composer.json` change:

| Package | From | To | Why |
|---|---|---|---|
| `utopia-php/mongo` | 1.5.2 | 1.5.3 | the fix itself (utopia-php/mongo#49) |
| `utopia-php/database` | 7.1.4 | 7.2.1 | regression coverage (utopia-php/database#936) |

Both references match their tags: `be29ee2d` is #49's merge commit, `38181c98` is the 7.2.1 tag and the merge of #936.

**Wider than the int64 fix, worth a reviewer's attention:** 7.1.4 → 7.2.1 also carries utopia-php/database#733, which classifies unique violations by the violated key instead of matching on the driver's message string. That touches `src` across MariaDB, MySQL, Postgres, SQLite, Memory and Redis adapters. It is unrelated to this issue and rides along with the version bump.

## On the reported "write should have been rejected"

The issue also expected `-3408048000` to be rejected by an `integer` column. That part is not a bug. An integer column created without explicit `min`/`max` gets `size = 8`, so it is a 64-bit column and the value is legal. `Structure` does range-validate array elements when a narrower `max` is set. Only the read path was broken, so no validation change here.

## Where the regression tests live

Deliberately not in this repo: utopia-php/database#936, now shipped in 7.2.1. Two tests there, both **seen red** before 1.5.3 and green after:

- `sum()` returned the aggregate total raw against a `float|int` return type, so any total past int32 was a `TypeError`.
- Object attributes have no per-key schema to cast against, so the wrapper reached the response and serialised as `{"$numberLong":"-3408048000"}` instead of a number.

Typed `integer`/`bigint` attributes were already surviving, because `castingAfter()` applies `(int)` and ext-mongodb 2.x gives `Int64` a numeric cast handler. The two paths above were the real leaks, and they are adapter-level behaviour, so the adapter suite is where they belong. Duplicating that as an appwrite e2e test would be slower coverage of a dependency bug at the most expensive layer.

## Not verified

- **No local e2e run.** Lock-only change; I did not stand up the full appwrite stack and am relying on CI. End-to-end proof for the int64 fix is in utopia-php/database#936, run against the real released 1.5.3: full `MongoDBTest`, 663 tests, 4997 assertions, 0 failures, plus the two new tests green on all six adapters and in schemaless and shared-tables modes.
- **The unique-violation change from #733.** I did not review or exercise it; it arrives with the version bump and its own CI passed in the database repo.
- **32-bit PHP.** 1.5.3 keeps the wrapper there on purpose, as the only lossless representation. No repo in the chain has a 32-bit CI target.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4416,16 +4416,16 @@
         },
         {
             "name": "utopia-php/mongo",
-            "version": "1.5.2",
+            "version": "1.5.3",
             "source": {
                 "type": "git",
                 "url": "https://github.com/utopia-php/mongo.git",
-                "reference": "20f9a644a356599fdd6265ba48696fc42a1f1881"
+                "reference": "be29ee2d84b9f7efdfc6fea5b4b2d4bf95ff5ec4"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/utopia-php/mongo/zipball/20f9a644a356599fdd6265ba48696fc42a1f1881",
-                "reference": "20f9a644a356599fdd6265ba48696fc42a1f1881",
+                "url": "https://api.github.com/repos/utopia-php/mongo/zipball/be29ee2d84b9f7efdfc6fea5b4b2d4bf95ff5ec4",
+                "reference": "be29ee2d84b9f7efdfc6fea5b4b2d4bf95ff5ec4",
                 "shasum": ""
             },
             "require": {
@@ -4471,9 +4471,9 @@
             ],
             "support": {
                 "issues": "https://github.com/utopia-php/mongo/issues",
-                "source": "https://github.com/utopia-php/mongo/tree/1.5.2"
+                "source": "https://github.com/utopia-php/mongo/tree/1.5.3"
             },
-            "time": "2026-08-12T06:57:25+00:00"
+            "time": "2026-08-13T01:55:11+00:00"
         },
         {
             "name": "utopia-php/platform",
--- a/composer.lock
+++ b/composer.lock
@@ -3638,16 +3638,16 @@
         },
         {
             "name": "utopia-php/database",
-            "version": "7.1.4",
+            "version": "7.2.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/utopia-php/database.git",
-                "reference": "b2b136a3827ccee2ea4eb83626348db4992286e1"
+                "reference": "38181c98e23b28c753f35cba6f03086a37033389"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/utopia-php/database/zipball/b2b136a3827ccee2ea4eb83626348db4992286e1",
-                "reference": "b2b136a3827ccee2ea4eb83626348db4992286e1",
+                "url": "https://api.github.com/repos/utopia-php/database/zipball/38181c98e23b28c753f35cba6f03086a37033389",
+                "reference": "38181c98e23b28c753f35cba6f03086a37033389",
                 "shasum": ""
             },
             "require": {
@@ -3692,9 +3692,9 @@
             ],
             "support": {
                 "issues": "https://github.com/utopia-php/database/issues",
-                "source": "https://github.com/utopia-php/database/tree/7.1.4"
+                "source": "https://github.com/utopia-php/database/tree/7.2.1"
             },
-            "time": "2026-08-12T09:39:29+00:00"
+            "time": "2026-08-13T02:26:08+00:00"
         },
         {
             "name": "utopia-php/detector",


abnegate:
@greptileai review

abnegate:
@greptileai review

Added a second lock bump since your last pass, so this is now on `446df3f1` rather than `8e0ea8d5`.

`utopia-php/database` 7.1.4 to 7.2.1. That carries the int64 regression tests from utopia-php/database#936, but it also carries utopia-php/database#733, which reclassifies unique violations by the violated key instead of matching the driver message. That one touches src across six adapters and is unrelated to this issue.

Worth a look at whether anything in appwrite reads unique violation messages or exception shape in a way #733 changes, since that is the part of this bump I have not exercised.

abnegate:
Chased the #733 risk I flagged above. It does not reach the API.

Two things could have bitten:

1. `Unique` is a new subclass of `Duplicate`, so every existing catch still matches. All 102 `Duplicate` catch sites here are unaffected, and nothing in this repo imports `Unique`.

2. The Mongo path now throws `UniqueException('Unique index violation')` where it used to throw `DuplicateException('Document with the requested unique attributes already exists')`, and it parses the index name out of the driver message instead of substring matching on `_uid`. That message change would be user visible if we passed it through, but we don't. Every `Duplicate` catch either uses the non-capturing form or maps to our own error type with `params:`, so the library string is discarded. The one nearby `getMessage()` in `Presences/HTTP/Update.php` belongs to the `StructureException` catch, not the `Duplicate` one.

So the message text is internal only. Still worth a reviewer confirming the intent of shipping #733 in the same bump, since it is unrelated to the int64 fix.

PR #13198, opened 2026-08-13
fix(sites): correct the Nuxt static output directory

HarshMN2345:
Nuxt's `static` adapter pointed at `./output/public`, missing a leading dot. `nuxt generate` writes `.output/public` and leaves `dist` as a symlink to it; a directory named `output` is never created.

Because the failure is a missing directory rather than an error, a deployment built from it comes out empty instead of failing the build.

Verified by generating a minimal Nuxt site:

```
$ npx nuxt generate
$ ls -d .output output
.output
ls: output: No such file or directory

$ ls .output/public
200.html  404.html  _nuxt  _payload.json  index.html

$ ls -la | grep dist
dist -> /…/.output/public
```

The `ssr` adapter beside it already said `./.output`, so this brings the pair back into one layout, matching every other framework in the file.

No test: nothing in `tests/unit` asserts config values, and the reproduction above is the check.

## Also found, not fixed here

Testing detection against 20 open-source repos and building 8 TanStack projects turned up more than this path. Recording it so it is not lost.

**TanStack Start pairs two build layouts.** `ssr` is `./.output` (nitro) while `static` is `./dist/client` (vite-native), so one of the two is always wrong. Left alone on review.

| Project | version | vite config | Builds to |
|---|---|---|---|
| **official `create-start-app`** | 1.168.44 | `nitro()` **in the default template** | `.output/server/index.mjs` |
| `KyleAMathews/…electric-starter` | 1.130.3 | plain | `.output/server/index.mjs` |
| `Kiranism/tanstack-start-dashboard` | 1.168.44 | `nitro()` | `.output/server/index.mjs` |
| `Ameyanagi/morphos` | 1.168.44 | `nitro()` + prerender | `.output` + 613 prerendered html in `public/` |
| `jackytea/tanstack-starter` | 1.168.36 | `nitro()` + prerender | `.output/server/index.mjs` |
| `appwrite/starter-for-tanstack-start` | 1.133.27 | plain `tanstackStart()` | `dist/server/server.js` |
| same, upgraded | 1.168.44 | plain `tanstackStart()` | `dist/server/server.js` |
| `daveyplate/better-auth-…` | 1.141.3 | `nitroV2Plugin({preset:"vercel"})` | `.vercel/output/` |

Five of eight land in `.output`, including a project generated today with the official CLI, which ships the nitro plugin in its template. A plain `tanstackStart()` writes `dist` instead — that is what our own starter uses, and it is the minority shape. A nitro `preset` can send output somewhere neither directory covers, as the Vercel row shows.

`app/config/templates/site.php` also declares `'outputDirectory' => './dist'` with `'adapter' => 'ssr'` for TanStack Start, which disagrees with the framework config either way.

Two things that lived on this branch and were dropped, both worth revisiting with that decision: an invariant test asserting a nested static directory sits inside its framework's ssr output (it caught both paths, but has to skip tanstack-start by name to stay green, which is the case it exists to catch), and rendering tests covering both TanStack layouts.

## Follow-up

`Detections/Create.php` fills `outputDirectory` from the detector rather than from this config, so a Git-connected Nuxt site still receives `./output` until utopia-php/detector#18 lands and is bumped here.


app/config/frameworks.php
```diff
@@ -251,7 +251,7 @@
                 'key' => 'static',
                 'buildCommand' => 'npm run build',
                 'installCommand' => 'npm install',
-                'outputDirectory' => './dist/client',
+                'outputDirectory' => './.output/public',
```
Meldiron: lets remove for now

Meldiron (approved):

PR #13200, opened 2026-08-13
Simplify the GitHub events handler

HarshMN2345:
## What does this PR do?

Two commits against `VCS/Http/GitHub/Events/Create.php`.

**1. `fix(vcs): scope GitHub pull request cleanup to GitHub repositories`**

The closed-pull-request branch matched repositories on `providerRepositoryId` alone. That id is only unique per provider — a GitHub and a GitLab repository can carry the same number — so closing a GitHub pull request could untrack that number from an unrelated repository on another provider.

`Appwrite\Vcs\RepositoryPullRequestCleanup` already resolves each repository's installation and skips the ones from a different provider. GitLab and Bitbucket route through it; GitHub was the only one still doing this inline. Delegating also picks up its explicit `Query::limit(100)`, where the inline query fell back to the default of 25.

**2. `refactor(vcs): simplify the GitHub events handler`**

- The installation-deleted path had three hand-rolled `do/while` cursor loops, each tracking its own cursor and rebuilding its queries per iteration. `Database::foreach` pages internally — it is the same `cursorAfter` walk with `Query::limit` as the page size, so behaviour is unchanged. Wrapping the whole iteration in one `Authorization::skip` makes the per-call ones redundant, since `skip` saves and restores the previous status.
- The pull request handler's `if`/`elseif` become guards, so the 50-line main body loses a level of indentation. The closed branch stops re-reading `action` out of the payload when it already holds it.
- The push handler returns early when the commit is ours or the branch was deleted. `fromInstallation()` and the repository lookup now run only when a deployment follows, instead of on every deployment-status commit we author ourselves.
- `empty($secretKey) ? true : ...` becomes `||`.

339 → 296 lines.

## Test Plan

- `tests/unit/Vcs` — 38/38 pass.
- PHPStan level 4 and Pint clean on the changed file.
- The e2e suite `tests/e2e/Services/VCSGitHub` was **not** run locally; it needs a running stack. The installation-deleted path is the part worth exercising there — three nested paginations that the unit tests do not reach.

## Related PRs and Issues

None.

## Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

Yes.

### DIFF
--- a/src/Appwrite/Platform/Modules/VCS/Http/GitHub/Events/Create.php
+++ b/src/Appwrite/Platform/Modules/VCS/Http/GitHub/Events/Create.php
@@ -73,7 +73,7 @@ class Create extends Action
         $signature = $request->getHeaderLine('x-hub-signature-256', '');
         $secretKey = $vcsWebhookSecret('github');
 
-        $valid = empty($secretKey) ? true : $vcs->validateWebhookEvent($payload, $signature, $secretKey);
+        $valid = empty($secretKey) || $vcs->validateWebhookEvent($payload, $signature, $secretKey);
         Span::add('vcs.github.event.signature.valid', $valid);
 
         if (!$valid) {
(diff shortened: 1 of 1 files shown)


HarshMN2345:
Closing in favour of #13510, which makes the same `RepositoryPullRequestCleanup` change for the GitHub events handler and covers the other provider event endpoints too.

PR #13208, opened 2026-08-13
fix(queries): reject search on attributes without a fulltext index

ChiragAgg5k:
## What

`Query::search()` on an attribute without a fulltext index returns a 500 instead of a 4xx. Four issues in Sentry, ~3400 events, oldest first seen 2025-05-19.

| Sentry | Endpoint | Attribute | Events |
|---|---|---|---|
| CLOUD-3590 | `functions.list` | `name` | 1948 |
| CLOUD-3MGV | `proxy.listRules` | `domain` | 1265 |
| CLOUD-3MHY | `sites.listLogs` | `requestPath` | 166 |
| CLOUD-3Q30 | `functions.listVariables` | `key` | 2 |

## Why

The endpoint query validators do not know about collection indexes, so these searches pass request validation. `Database::find()` revalidates them against the stored collection metadata and throws `Utopia\Database\Exception\Query`. The endpoints only caught that exception while parsing the query strings, allowing the database exception to escape as a 500.

## Change

Catch `QueryException` around the `find()` and `count()` calls and return `GENERAL_QUERY_INVALID`, matching the document list endpoint.

This PR updates the three affected Appwrite endpoints:

- `src/Appwrite/Platform/Modules/Functions/Http/Functions/XList.php`
- `src/Appwrite/Platform/Modules/Functions/Http/Variables/XList.php`
- `src/Appwrite/Platform/Modules/Proxy/Http/Rules/XList.php`

`sites.listLogs` now lives in the Cloud repository. Its handler was updated on the matching [`fix/reject-search-without-fulltext-index`](https://github.com/appwrite-labs/cloud/tree/fix/reject-search-without-fulltext-index) branch.

The dedicated `search` parameter remains unaffected because it searches the indexed synthetic `search` attribute.

## Tests

Added E2E coverage asserting that each reported query returns HTTP 400 with `general_query_invalid`:

- `functions.list` searching `name`
- `functions.listVariables` searching `key`
- `proxy.listRules` searching `domain`
- `sites.listLogs` searching `requestPath` (Cloud branch)

## Checks

- `composer lint` on the changed Appwrite handlers and tests
- `composer analyze` on the changed Appwrite handlers and tests
- Pint and `php -l` on the changed Cloud handler and test

The E2E tests require the Appwrite and Cloud stacks and were not run locally.

### DIFF
--- a/src/Appwrite/Utopia/Database/Validator/Queries/Base.php
+++ b/src/Appwrite/Utopia/Database/Validator/Queries/Base.php
@@ -5,7 +5,7 @@ namespace Appwrite\Utopia\Database\Validator\Queries;
 use Utopia\Config\Config;
 use Utopia\Database\Database;
 use Utopia\Database\Document;
-use Utopia\Database\Validator\Queries;
+use Utopia\Database\Validator\IndexedQueries;
 use Utopia\Database\Validator\Query\Cursor;
 use Utopia\Database\Validator\Query\Filter;
 use Utopia\Database\Validator\Query\Limit;
@@ -13,7 +13,7 @@ use Utopia\Database\Validator\Query\Offset;
 use Utopia\Database\Validator\Query\Order;
 use Utopia\Database\Validator\Query\Select;
 
-class Base extends Queries
+class Base extends IndexedQueries
 {
     /**
      * Expression constructor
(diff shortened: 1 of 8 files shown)


src/Appwrite/Utopia/Database/Validator/Queries/Base.php
```diff
@@ -103,7 +103,18 @@ public function __construct(string $collection, array $allowedAttributes)
             $validators[] = new Select($allAttributes);
         }
 
-        parent::__construct($validators);
+        // Search queries need a fulltext index on the attribute. Database::find()
+        // enforces that too, but throws a QueryException the endpoints do not catch,
+        // so an unsupported search surfaces as a 500 instead of a validation error.
+        $indexes = [];
```
abnegate: We should just catch query exception where it's thrown instead

ChiragAgg5k: Updated the handlers to catch `QueryException` around `find()`/`count()` instead. The fourth endpoint, `sites.listLogs`, now lives in Cloud, so I pushed the matching change to the same branch there: https://github.com/appwrite-labs/cloud/tree/fix/reject-search-without-fulltext-index

abnegate (approved):

PR #13219, opened 2026-08-14
feat(sites): auto-generate API key with scopes for builds and SSR executions

Meldiron:
## What does this PR do?

Brings the Functions ephemeral API key feature to Sites. Sites now auto-generate an ephemeral API key exactly like functions do — delivered as `APPWRITE_SITE_API_KEY` during builds and as the `x-appwrite-key` header during SSR executions.

The key-generation paths (`Deployments::variables()` and the domain router in `general.php`) were already resource-agnostic, but always resolved to empty scopes because sites had no `scopes` attribute.

- Add `scopes` attribute to the `sites` collection, with V25 migration (idempotent, skips existing attributes)
- Add `scopes` param to sites create/update (same validator and `ProjectKeyScopes` enum as Functions) and expose it on the `Site` response model
- Resolve every ephemeral key's scopes through `Deployments::scopes()`: the resource's own scopes plus the per-resource-type `computeScopes` config. All key paths go through it — build payload, domain router, and the function execution endpoint/worker — so platforms can extend keys with platform-specific scopes via config
- Grant `health.read` to both functions and sites as the `computeScopes` default

Note on `health.read`: `Key::decode` already merges the base keys-role scopes (`global`, `health.read`, `graphql`) into every ephemeral key, so this makes the grant explicit at mint time rather than adding a new capability.

## Test plan

`SitesCustomServerTest::testScopes` and `FunctionsCustomServerTest::testScopes`, both passing locally (92 and 80 assertions). Each key is checked two ways — the decoded JWT payload carries the granted `users.read` plus `health.read`, **and** the key is used to call `GET /v1/health` (a `health.read` gated endpoint) with the response status asserted to be 200:

| Key path | Covered by |
|---|---|
| Sites build (`Deployments::variables`) | `api-key.sh` in the astro fixture, asserted via build logs |
| Sites SSR runtime (`general.php` router) | new `/api-key` SSR route in the astro fixture |
| Functions build | `setup.sh` in the `ephemeral-api-key` fixture, asserted via build logs |
| Functions sync execution (`Executions/Create.php`) | fixture response body |
| Functions async execution (`Workers/Functions.php`) | execution-update project webhook, matched on the run's execution id |

The async path is observed through the webhook because CE exposes no endpoint to read an execution back (`Functions/Http/Executions/` has only `Create` and `Delete`).

Also covers `scopes` in sites create/get/update responses. Regression: 7 neighboring Sites E2E tests (create/get/update/list/variablesE2E) passing locally. Pint, PHPStan, and Rector clean.


Meldiron (commented):
now im happy

Issue #13223, opened 2026-08-15
Adding new gravatar api to avatar endpoint


imtia33:
### 🔖 Enhancement description

A new addition to avatar was discussed after talking about implementation of a function that can easily fetch Oauth identity avatar. Before its to be implemented [ gravatar ](https://gravatar.com/) was picked for first smaller self contained change as adding Oauth2 identities  require more changes and need separate research.

Api: 
`avatars.getGravatar()` — fetch the user's avatar from [Gravatar](https://gravatar.com), which is keyed by email hash and is the de-facto cross-platform avatar service.

- No DB schema changes, no migrations, no new config arrays, no new env vars.

### Out of scope (for follow-up PRs)

- `avatars.getOauthAvatar()` — needs a `providerAvatarUrl` column on `identities`, a migration, and per-provider userinfo adapter overrides (see Part 7).
- `avatars.setAvatar()` — Matej suggested an "Identities" / public-profile model is a better fit than a setter on User. Out of scope here.

### Acceptance criteria

- [ ] Endpoint returns `200 OK` with `Content-Type: image/png` for a real gravatar URL.
- [ ] Endpoint works anonymously with `?project=<projectId>` (no session).
- [ ] Endpoint works with session/JWT/key (uses current user's email).
- [ ] Endpoint accepts `email` override param.
- [ ] Bad `width` / `height` → `400`.
- [ ] Gravatar upstream failure → `404` with `avatar_image_not_found` (matches existing `/v1/avatars/image` behavior).
- [ ] E2E tests pass locally: `docker compose exec appwrite vendor/bin/phpunit --filter testGetGravatar`
- [ ] No new lint/stan warnings: `composer lint && composer analyze`

### 🎤 Pitch

Add a new HTTP endpoint **`GET /v1/avatars/gravatar`** that returns the Gravatar image for the current user (or for an explicitly-supplied email/userId).

```mermaid
flowchart LR
    Client -->|GET /v1/avatars/gravatar?width=200| Appwrite
    Appwrite -->|1. Resolve current user's email or use email/userId| DB[(Project DB)]
    Appwrite -->|2. MD5 lowercase trimmed email| Hash[Hash string]
    Appwrite -->|3. Fetch Gravatar avatar using hash| Gravatar[gravatar.com]
    Gravatar -->|image bytes| Appwrite
    Appwrite -->|4. Re-encode via Imagick, Cache-Control 30d, return image/png| Client
```

### API shape (proposed)

| Property | Value |
[...]

PR #13252, opened 2026-08-17
feat: Add Gravatar avatar support with session fallback

imtia33:
## What does this PR do?

This PR introduces a new endpoint `GET /v1/avatars/gravatar` to the Avatars service, allowing developers to fetch Gravatar images for their users. 
As discussed in the issue, this implements the following behaviors:
- Uses `SHA-256` hashing for the email address before querying Gravatar.
- Accepts an `email` parameter. If left empty, it falls back to the currently authenticated user's session email.
- Returns an HTTP 400 Bad Request error if no email is provided AND no active session is found.
- Forwards Gravatar's native query parameters: `default`, `rating`, `width` (as `s`), and `height`. 
- Includes full End-to-End (E2E) test coverage for explicit email queries, fallback edge-cases, and session-based fetching.

## Test Plan

- Verified in local Docker environment (`docker compose up`).
- Passed `composer lint` and `composer analyze`.
- Full E2E suite executed and passing for the Gravatar endpoint across all auth modes:
  `docker compose exec appwrite vendor/bin/phpunit --filter testGetGravatar`
  Results:
1. Test Run:
image

(I am new to this but I am just worried about the 20 seconds warning. Tell me if you have any suggestion)
2. Manual verification:
- Without hight & width:
image

- With Hight & Width:
image

- No gravatar exsit(identicon fallback):
image

The `default` param is optional. Accepts:
1. **`mp`** - (Mystery Person) A simple, cartoon-style silhouetted outline of a person (this is the default).
2. **`identicon`** - A geometric pattern based on an email hash.
3. **`monsterid`** - A generated 'monster' with different colors, faces, etc.
4. **`wavatar`** - Generated faces with differing features and backgrounds.
5. **`retro`** - Awesome generated, 8-bit arcade-style pixelated faces.
6. **`robohash`** - A generated robot with different colors, faces, etc.
7. **`blank`** - A transparent PNG image.
8. **`404`** - Doesn't load an image at all, just returns an HTTP 404 response (which we catch and turn into an `AVATAR_IMAGE_NOT_FOUND` error).

If no param provided, then falls back to `mp`.

## Related PRs and Issues

- #13223 

## Checklist


docs/references/avatars/get-gravatar.md
Meldiron, line 1: This is old approach, sadly AI still picks it up 🙈

Now that we have 1 file = 1 endpoint, we can have description directly inline in PHP code.

Pls suggest that to agent, Im sure he will be able to find example of existing endpoints that already follow this

src/Appwrite/Platform/Modules/Avatars/Http/Gravatar/Get.php
```diff
@@ -0,0 +1,149 @@
...
+use Utopia\Validator\Text;
+use Utopia\Validator\WhiteList;
+
+class Get extends Action
+{
+    use HTTP;
+
+    public static function getName(): string
+    {
+        return 'getGravatar';
```
Meldiron:
```suggestion
        return 'getAvatarsGravatar';
```

Just a consistency nitpick

src/Appwrite/Platform/Modules/Avatars/Http/Gravatar/Get.php
```diff
@@ -0,0 +1,149 @@
...
+            $domain = new Domain($host);
+            if (!$domain->isKnown()) {
+                throw new Exception(Exception::AVATAR_REMOTE_URL_FAILED);
+            }
+        }
+
+        $hostnameValidator = new PublicHostname();
+        if (!$hostnameValidator->isValid($host)) {
+            throw new Exception(Exception::AVATAR_REMOTE_URL_FAILED, $hostnameValidator->getDescription());
+        }
```
Meldiron: No need for this protection, I would remove it.
E2E tests will make sure our gravtar implementation is OK.

src/Appwrite/Platform/Modules/Avatars/Http/Gravatar/Get.php
```diff
@@ -0,0 +1,149 @@
...
+        // Gravatar spec: SHA-256 of strtolower(trim($email))
+        $hash = \hash('sha256', \strtolower(\trim($resolvedEmail)));
+
+        $gravatarUrl = \sprintf(
+            'https://www.gravatar.com/avatar/%s?s=%d&d=%s&r=%s',
+            $hash,
+            $width,
+            \urlencode($default),
+            \urlencode($rating),
+        );
```
Meldiron: I think we dont often use sprintf in our source code; lets conflict agent on this, see what he does instead (code consistency being main focus)

tests/e2e/Services/Avatars/AvatarsBase.php
```diff
@@ -1342,4 +1342,107 @@ public function testGetScreenshotComparison(): array
...
         return [];
     }
+
+    public function testGetGravatar(): array
+    {
+        /**
+         * Test for SUCCESS
+         */
+
+        // Explicit email param — uses eldad@appwrite.io (known Gravatar, stable for testing)
```
Meldiron: Let's assert hash of the response body too. Unless image changes, it should match exactly.

tests/e2e/Services/Avatars/AvatarsBase.php
```diff
@@ -1342,4 +1342,107 @@ public function testGetScreenshotComparison(): array
...
+                'width' => 200,
+                'height' => 200,
+            ]);
+
+            $this->assertEquals(200, $response['headers']['status-code']);
+            $this->assertEquals('image/png', $response['headers']['content-type']);
+            $this->assertNotEmpty($response['body']);
+        }, 30_000, 2_000);
+
+        // Default size with identicon — deterministic for any unknown email
```
Meldiron: Let's also hash-match the response body (md5 should be good) here too

src/Appwrite/Platform/Modules/Avatars/Http/Gravatar/Get.php
```diff
@@ -0,0 +1,149 @@
...
+    public function action(
+        string $email,
+        int $width,
+        int $height,
+        string $default,
+        string $rating,
+        Response $response,
+        Document $user,
+    ): void {
+        if (!\extension_loaded('imagick')) {
```
Meldiron: Let's simplify for first iteration - let's remove imagick and Image() logic from here. That will also mean removal of `width` and `height` params.

Meldiron: Let's default to 256 width and height for now, because 80 might be too little for many usecases

imtia33: So, no height or width customization?

imtia33:
This PR refactors our external avatar fetching into a new, centralized `getPhoto` endpoint that uses a strict priority-based fallback chain. The goal is to provide a single, smart endpoint that resolves the best possible profile picture for the authenticated user, while gracefully falling back to our own branded designs instead of generic third-party placeholders.

**Key Changes:**

* **Endpoint Renamed & Refactored:** 
  * Removed the old `avatars.getGravatar()` method and replaced it with `avatars.getPhoto()` (`/v1/avatars/photo`).
  * The endpoint now automatically resolves the `email` and `name` from the authenticated user's session document, rather than requiring the client to pass an explicit `email` parameter.
  * 
* **New Provider Architecture (`src/Appwrite/AvatarPhotos`):** 
  * Extracted resolution logic into standalone classes for better maintainability.
  * `Photo.php` acts as the orchestrator, attempting to resolve an image in the following strict priority order:
    1. **OAuth2 Session Photo** *(TODO placeholder added for future implementation)*
    2. **Gravatar** (using `d=404` to fail-fast if no custom photo exists)
    3. **Libravatar** (using `d=404` to fail-fast)
    4. **Initials** (Extracted from existing initials logic)
    5. **Static Fallback** (A branded SVG fallback for when all else fails)
* **Strict Fallback Enforcement:** By using `d=404` for both Gravatar and Libravatar, we ensure we only return their images if the user has a *genuine, custom* photo uploaded. Otherwise, it immediately falls back to our own Initials or Static designs rather than relying on generic third-party robots/silhouettes.
* **Parameter Parity:** Added back support for `width`, `height`, `quality`, and `output` (`jpg`, `png`, `webp`) to the new endpoint, utilizing Utopia's Image engine to match the behaviour of the rest of the Avatars API.
* **Testing:** Replaced the old Gravatar E2E tests with `testGetPhoto` to validate the new endpoint, parameter boundaries, and output formats.
image

src/Appwrite/Platform/Modules/Avatars/Http/Photo/Get.php
```diff
@@ -0,0 +1,178 @@
...
+                description: <<<EOT
+                Returns the best available profile photo for the currently authenticated user.
+
+                The endpoint tries each source in priority order and returns the first successful result:
+
+                1. OAuth2 session photo (planned — see TODO in source)
+                2. Gravatar
+                3. Libravatar
+                4. Initials generated from the user's name or email
+                5. A built-in static fallback image
```
Meldiron:
```suggestion
                Returns the best available profile photo for the currently authenticated user. The endpoint tries each source in priority order and returns the first successful result: Gravatar, Libavatar, Appwrite Initials, built-in static fallback file.
```

PR #13268, opened 2026-08-18
refactor(schedules): run the schedulers on utopia-php/schedule

loks0n:
## What

Runs `schedule-functions`, `schedule-executions` and `schedule-messages` on [`utopia-php/schedule`](https://github.com/utopia-php/schedule), and deletes `ScheduleBase`.

Each task now constructs its own scheduler, so what it schedules and how it dispatches read top to bottom in one file:

```php
$scheduler = new Scheduler(
    source: new ScheduleSource(
        ...,
        resource: fn (Database $projectDB, array $schedule): Document => $projectDB->getDocument(self::getCollectionId(), $schedule['resourceId']),
        entry: fn (array $schedule): Entry => new Entry($this->trigger($schedule), $schedule),
    ),
    store: new ScheduleStore($pools, 'utopia-schedule-' . self::getName()),
    interval: self::ENQUEUE_TIMER,
    sync: self::UPDATE_TIMER,
    lookahead: self::ENQUEUE_LOOKAHEAD,
    lookback: self::ENQUEUE_LOOKBACK,
);

$scheduler->run(fn (array $occurrences) => $this->dispatch($occurrences, $dbForPlatform));
```

## Why

The shared base hand-rolled the loop, and that loop lost runs two ways.

It recomputed a fresh 60-second window from the moment each tick started (`$timeFrame = DateTime::addSeconds(new \DateTime(), ENQUEUE_TIMER - $enqueueDiff)` with `$enqueueDiff` pinned to `0` under a TODO) while evaluating every schedule with `$cron->getNextRunDate()` — no base time, so relative to a "now" that kept moving as the pass walked thousands of schedules. The wall time the pass itself consumed therefore became an interval no window covered, and an occurrence landing there was skipped **permanently**: no error, no execution record, no retry. Because the tick phase barely drifts under a coroutine sleep, a bad phase persists, so this does not present as one missed run — it presents as a whole cohort of schedules going quiet for as long as the phase sits on the boundary.

It also held no watermark, so everything due while the process was restarting was lost outright, and the initial `collectSchedules` on a large fleet takes long enough for that to matter.

This is the cause of the customer reports about `*/15` functions missing runs (and the year-old thread they cite). Neither defect was reachable by a test: the window arithmetic and the clock were tangled into the same method.

## How

The library owns the arithmetic. Occurrences come from explicit half-open windows that **tile** — each opens exactly where the previous committed one closed — and a window is committed only after its occurrences have been handed over, so a tick that dies repeats rather than skips. The committed window travels with the leadership claim in a single Redis compare-and-swap, which also gives replicas a single dispatcher, lets a replacement resume coverage where its predecessor stopped, and fences a deposed leader's late commit instead of letting it rewind coverage.

`ScheduleBase` is deleted rather than reimplemented. What the three genuinely share — paging the `schedules` collection for a region and resource type, resolving projects, loading resources, dropping orphans — is a collaborator (`ScheduleSource`), not an inherited template. Each task supplies the two things that actually differ: how to load its resource, and what its stored `schedule` attribute means. Functions read it as a cron expression; executions and messages as a fixed moment, which is also how one-shots retire now — the library drops them once delivered.

## Behaviour held constant

| | Before | After |
|---|---|---|
| Functions spread | deterministic `spreadOffset`, sleep to the exact second | unchanged (`ENQUEUE_LOOKAHEAD = 60`) |
| Executions lead time | 4s (`intervalEnd`) | unchanged (`ENQUEUE_LOOKAHEAD = 4`) |
| Messages | fire only once due, never early | unchanged (`ENQUEUE_LOOKAHEAD = 0`) |
| Past-due runs | slept a negative interval | clamped to zero, dispatched immediately |
| Catch-up after downtime | none (runs lost) | bounded by `ENQUEUE_LOOKBACK = 300s` |

One intentional delta: executions now touch `accessedAt` like functions and messages already did, so all three treat a dispatch as project activity.

## A deployment detail worth reviewing closely

The claim store borrows from the **lock pool**, not the legacy `redis` container resource. That resource reads `_APP_REDIS_HOST`, which deployed environments leave unset — it would have resolved to `localhost`, and a scheduler that cannot take its claim never dispatches at all. The lock pool is built from `_APP_CONNECTIONS_LOCKS`, which they do set (`lock=redis://cache:6379?…`), and borrowing per call lets the pool replace a connection that dies rather than stranding the claim for the process lifetime.

## Tests

Covering what this repository still owns: that each task reads its schedule attribute as the right kind of trigger, that an unusable cron expression (`0 0 31 2 *`) is rejected once where the row is built rather than silently matching nothing forever, and that only the tasks which sleep take lead time. `spreadOffset` coverage moved with it.

Window tiling, the watermark and failover are covered by the [library's own suite](https://github.com/utopia-php/schedule) (65 unit + 7 e2e against real Redis), which replays this incident's tick phase — creeping 1.5 ms per tick across a minute boundary for an hour — and asserts every occurrence is delivered exactly once, plus a forked contention test proving the claim's compare-and-swap is atomic.

Locally: `phpunit tests/unit` passes (two pre-existing failures from extensions missing on my machine, reproduced on clean `main`), `pint` clean, `phpstan` clean. There is no e2e for the scheduler path in this repository — it spans task, Redis and worker — so the canary is the real verification.

## Combined mode

Rebased onto the combined worker/scheduler work that landed meanwhile, so each task keeps the lifecycle that container drives:

- `setup($publisher, $telemetry)` — wire the publisher and telemetry
- `start($dbForPlatform, $getProjectDB, $getIsResourceBlocked, $pools)` — build the scheduler and load the schedules once, serially, so resource types do not contend for the shared console and cache pools
- `listen()` — run the loop, one coroutine each

Three library loops share one process because the library's sleeps yield under the runtime hooks `app/cli.php` already enables. `ScheduleLifecycleTest` pins those four signatures, since drift in them would otherwise only surface when the combined container boots.

## Requires a lockstep change in appwrite/cloud

`ScheduleBackups` extends the deleted base, so it needs its own scheduler the same way these three now have theirs. Cloud's other three subclasses keep working unchanged: `spreadWindow` and `updateProjectAccess` are still there to override, and `UpdateProjectAccessTrait`'s `parent::` call still resolves.

## Rollout

Supersedes #13262 (closed), which patched the same selection bug in place. Worth canarying one region first and watching `task.schedule.enqueue_delay` alongside `schedule.lag`, `schedule.dispatch.total` and `schedule.error.total{stage="lease"}` — a lease stage error every tick would mean the lease is shorter than a dispatch takes.


loks0n:
One follow-up, flagged rather than hidden: two things in `ScheduleSource` are workarounds for library gaps rather than appwrite's problem, and both are now fixed at source in [utopia-php/monorepo#145](https://github.com/utopia-php/monorepo/pull/145).

**The `activeFrom` recency heuristic.** `ScheduleSource::activeFrom()` currently only reports a row's change time if it changed within `UPDATE_TIMER * 3` seconds, otherwise `null`. That magic multiplier exists because the library's late-discovery coverage had no floor: passing every row's real `activeFrom` would make a cold start replay a whole `lookback` of history for all ~3,000 schedules on every restart, while passing `null` would lose a schedule created between two syncs. The library now bounds that reach at the previous sync, which is the judgement it is better placed to make — it knows when it last looked, where this code was guessing. The heuristic then deletes down to a straight pass-through.

**The snapshot counter.** `snapshotted()` exists only so the combined task can log "N active", and it overcounts by including rows the scheduler goes on to reject (missing project, blocked or orphaned resource) — the docblock admits it. `Scheduler::count()` upstream is the authoritative figure, so both the counter and the `$source` handle disappear.

Two smaller local tidies ride along: `ScheduleSource` types its callables as `\Closure` rather than `mixed`, and `start()` refuses when `setup()` has not run instead of silently substituting a no-op telemetry adapter.

The patch is written and verified against the library branch (360 unit tests, phpstan, pint and rector all clean), but it cannot land until `schedule/0.1.1` is tagged — `composer.lock` can't reference an unpublished version, and `count()` does not exist in 0.1.0. So this PR stays as-is and green; I'll push the simplification the moment the tag exists.

src/Appwrite/Platform/Tasks/ScheduleFunctions.php
```diff
@@ -47,105 +158,86 @@ protected function spreadWindow(array $schedule, Database $dbForPlatform): int
...
-                        $publisherForFunctions->enqueue(new FunctionMessage(
+                        $this->updateProjectAccess($schedule['project'], $dbForPlatform);
+
+                        $this->publisher()->enqueue(new FunctionMessage(
                             project: $schedule['project'],
                             function: $schedule['resource'],
                             type: 'schedule',
                             method: 'POST',
                             path: '/',
                         ));
```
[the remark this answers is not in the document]
loks0n: Valid, and a regression I introduced — `ScheduleBase` re-checked `$this->schedules` after the sleep ("Ensure schedule was not deleted") and that guard was lost when the entry map moved into the scheduler. With cloud's 300s function spread the window is up to five minutes, so a disabled function really would have run once more.

Fixed in db1a821: `ScheduleSource` now keeps the version of each live schedule as it reads them (rebuilt from each full snapshot, which is the only read that can see a hard delete, and updated from the change feed in between), and both sleeping dispatch paths ask `isLive($sequence, $resourceUpdatedAt)` before publishing. Keying on the version also retires a run captured against a definition the user has since edited — the new definition covers its own occurrences from its `activeFrom`, so firing the old one would double up.

This stayed appwrite-side rather than going into the library on purpose: only the application knows what "still valid" means (active flag, deleted row, blocked project), so a handler that defers work owns re-validating it.

New `tests/unit/Platform/Tasks/ScheduleSourceTest.php` drives the real `Database` (Memory adapter) through `snapshot()`/`since()` for the disable, edit and hard-delete cases; both production hunks were seen red (dropping the snapshot reset fails the delete case, ignoring the `active` flag fails the disable case). The one-line call site inside the dispatch coroutine has no unit test over it — it needs a running loop.

src/Appwrite/Platform/Tasks/Schedule.php
```diff
@@ -21,65 +22,45 @@ public function __construct()
...
-        Console::success(APP_NAME . ' combined scheduler v1 has started');
-
-        /** @var list<ScheduleBase> $tasks */
-        $tasks = [
-            new ScheduleFunctions(),
-            new ScheduleExecutions(),
-            new ScheduleMessages(),
+        // Loading runs serially, so the three do not contend for the shared
+        // console and cache pools, and only then do the loops start.
+        $loops = [
```
loks0n: what's the point of this? let's just initialize them imperatively, it will be less code.

src/Appwrite/Platform/Tasks/Schedule.php
```diff
@@ -21,65 +22,45 @@ public function __construct()
...
-        Console::title('Scheduler V1 (combined)');
-        Console::success(APP_NAME . ' combined scheduler v1 has started');
-
-        /** @var list<ScheduleBase> $tasks */
-        $tasks = [
-            new ScheduleFunctions(),
-            new ScheduleExecutions(),
-            new ScheduleMessages(),
+        // Loading runs serially, so the three do not contend for the shared
+        // console and cache pools, and only then do the loops start.
```
loks0n: remove this useless comment

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
```diff
@@ -3,23 +3,38 @@
...
+use Utopia\Telemetry\Adapter as Telemetry;
+
+class ScheduleExecutions extends Action
 {
-    public const UPDATE_TIMER = 3; // seconds
-    public const ENQUEUE_TIMER = 4; // seconds
+    public const UPDATE_TIMER = 3; // seconds between reconciliations
+    public const ENQUEUE_TIMER = 4; // seconds between ticks
+    public const ENQUEUE_LOOKAHEAD = 4; // seconds of lead time, so a dispatch can sleep to the second
+    public const ENQUEUE_LOOKBACK = 300; // seconds of missed runs a restart recovers
```
loks0n: do we need these custom enqueue values or will the defaults work fine?

src/Appwrite/Platform/Tasks/Schedule.php
```diff
@@ -21,65 +22,45 @@ public function __construct()
...
+        $loops = [
+            (new ScheduleFunctions())->boot($publisherForFunctions, $telemetry, $dbForPlatform, $getProjectDB, $getIsResourceBlocked, $pools),
+            (new ScheduleExecutions())->boot($publisherForFunctions, $telemetry, $dbForPlatform, $getProjectDB, $getIsResourceBlocked, $pools),
+            (new ScheduleMessages())->boot($publisherForMessaging, $telemetry, $dbForPlatform, $getProjectDB, $getIsResourceBlocked, $pools),
```
loks0n: why call boot and not the action directly?

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
```diff
@@ -36,13 +51,72 @@ public static function getCollectionId(): string
         return RESOURCE_TYPE_EXECUTIONS;
     }
 
-    protected function loadResource(Document $project, callable $getProjectDB, array $schedule): Document
+    public function action(FunctionPublisher $publisherForFunctions, callable $getIsResourceBlocked, Database $dbForPlatform, callable $getProjectDB, Telemetry $telemetry, Group $pools): never
+    {
+        ($this->boot($publisherForFunctions, $telemetry, $dbForPlatform, $getProjectDB, $getIsResourceBlocked, $pools))();
```
loks0n: why do we need boot?

src/Appwrite/Platform/Tasks/ScheduleSource.php
```diff
@@ -0,0 +1,215 @@
...
+use Utopia\Database\Document;
+use Utopia\Database\Query;
+use Utopia\Schedule\Changes;
+use Utopia\Schedule\Source;
+use Utopia\Schedule\Source\Entry;
+use Utopia\Schedule\Source\Row;
+use Utopia\Span\Span;
+use Utopia\System\System;
+
+final class ScheduleSource implements Source, Changes
```
loks0n: This should be called, DatabaseSchedule, and it should live somewhere else, not here. check agents.md for advice

src/Appwrite/Platform/Tasks/ScheduleMessages.php
```diff
@@ -2,19 +2,39 @@
...
 
-class ScheduleMessages extends ScheduleBase
+class ScheduleMessages extends Action
 {
-    public const UPDATE_TIMER = 3; // seconds
-    public const ENQUEUE_TIMER = 4; // seconds
+    public const UPDATE_TIMER = 3; // seconds between reconciliations
+    public const ENQUEUE_TIMER = 4; // seconds between ticks
+    public const ENQUEUE_LOOKAHEAD = 0; // no lead time: a message must not go out early
+    public const ENQUEUE_LOOKBACK = 300; // seconds of missed runs a restart recovers
```
loks0n: do we need this custom stuff?

src/Appwrite/Platform/Tasks/ScheduleMessages.php
```diff
@@ -31,42 +51,94 @@ public static function getCollectionId(): string
...
+            telemetry: $telemetry,
+            onError: function (\Throwable $error): void {
+                Span::init('schedule.messages.reconcile');
+                Span::current()?->finish(error: $error);
+            },
+        );
 
-            $now = new \DateTime();
-            $scheduledAt = new \DateTime($schedule['schedule']);
+        $scheduler->reconcile();
```
loks0n: why reconcile? doesn't run handle that for us?

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
```diff
@@ -36,13 +51,72 @@ public static function getCollectionId(): string
...
+        Span::add('schedule.executions.loaded', $source->snapshotted());
+        Span::current()?->finish();
+
+        return fn (): null => $scheduler->run(fn (array $occurrences): null => $this->dispatch($occurrences, $publisher, $dbForPlatform));
+    }
+
+    protected function updateProjectAccess(Document $project, Database $dbForPlatform): void
+    {
+        ScheduleSource::touchProject($project, $dbForPlatform);
+    }
```
loks0n, line 72: this is anemic? what's the point? if we want to enforce it just do automatically it in the schedule source code

src/Appwrite/Platform/Tasks/Schedule.php
loks0n: Done in ebfc70c3 — three named locals and three calls, no array or foreach. The `Co::create` wrapper moved to a small private `loop()` so the three reads as three lines rather than a block.

src/Appwrite/Platform/Tasks/Schedule.php
loks0n: Removed.

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
loks0n: Partly — one of the four was dead and is gone in ebfc70c3.

- `ENQUEUE_LOOKBACK = 300` was exactly the library default. Deleted, in all three tasks.
- `UPDATE_TIMER = 3` (library default 10) is deliberate. Executions are user-created one-shots that can be due seconds after the row is written, so a 10s read cadence would discover some of them after they were due. They still run — coverage reaches back to `activeFrom` — but late by up to the sync gap. 3s keeps that under a few seconds.
- `ENQUEUE_TIMER = 4` with `ENQUEUE_LOOKAHEAD = 4` is what makes an execution fire *on* its second rather than up to a tick late: the tick selects 4s ahead and the library waits for the exact moment. Dropping the lookahead would leave them up to 4s late; dropping both for the defaults (tick 1, lead 0) would be ~1s late and tick 4x more often over 11.7k schedules in nyc3.

So: lookback gone, the other three earn their keep. Same reasoning for messages, minus the lookahead, which is 0 there on purpose.

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
loks0n: Two reasons, both from the combined `schedule` task — which is the default in `docker-compose.yml`, with the three split ones behind the `separate` profile, so it is the path most installs run.

1. `action()` ends in `exit(1)`. Called inside `Co::create`, that takes the whole process down rather than the one loop, so the combined task cannot call it.
2. Loading has to finish before any loop starts. `boot()` reconciles and returns the loop without running it, so the combined task loads functions, then executions, then messages — three sequential passes over the console database — and only then starts the three coroutines. Calling `action()` three times would have them load concurrently and contend for the shared console and cache pools at boot. `ScheduleBase` had the same split for the same reason (`start()` in a loop, then `listen()` in coroutines).

Happy to drop it if you would rather the combined task load concurrently — it is one method and three call sites — but that is a behaviour change, not just a shape one.

src/Appwrite/Platform/Tasks/Schedule.php
loks0n: Same answer as on `ScheduleExecutions::boot()`: `action()` ends in `exit(1)`, which kills the process rather than the coroutine, and this task has to finish loading all three before it starts any of them. `boot()` is the part of `action()` that stops before the blocking loop and hands it back instead.

src/Appwrite/Platform/Tasks/ScheduleSource.php
loks0n: Done in ebfc70c3: `src/Appwrite/Schedule/DatabaseSchedule.php`, `Appwrite\Schedule\DatabaseSchedule`.

Both halves of the old name were wrong by AGENTS.md. "Source" named it after the library interface it implements rather than what it is, and "Schedule" repeated the context it sat in — and it is not a task, so `Platform/Tasks/` was the wrong directory to begin with. `Appwrite\Schedule` gives it a home next to `Appwrite\Event` and `Appwrite\Migration`, which are the same shape: a domain namespace holding the classes a task or worker composes.

I tried `Appwrite\Schedule\Database` first, since rule 3 says not to restate the namespace, but it collides with `Utopia\Database\Database` in every signature in the file and forced an alias that read worse than the redundancy. `DatabaseSchedule` needs no alias anywhere.

src/Appwrite/Platform/Tasks/ScheduleMessages.php
loks0n: Same as on executions: `ENQUEUE_LOOKBACK` is gone (it was the library default). `UPDATE_TIMER = 3` stays because a scheduled message can be written seconds before it is due, and the default 10s read cadence would find some of them late. `ENQUEUE_LOOKAHEAD = 0` is the one that matters most here and is not a leftover: with any lead time the library would hand a message over before its moment, and a message must never go out early.

src/Appwrite/Platform/Tasks/ScheduleMessages.php
loks0n: It does — `run()` calls the sync step at the top of every iteration and the first one is always due, so the loop would load on its own.

The explicit call is there for the combined task: it is what makes loading happen *before* `run()` is entered, so the three load one after another rather than three coroutines all reconciling at once against the shared console pool. Drop it and `boot()` returns a loop that has not loaded yet, which makes the serial bootstrap in `Schedule::action()` meaningless.

So it stands or falls with the answer on `boot()` — if the combined task loads concurrently, both go.

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
loks0n: It is anemic here, and it is load-bearing in cloud — which is the only reason it exists.

`Appwrite\Cloud\Platform\Tasks\{ScheduleFunctions,ScheduleExecutions,ScheduleMessages}` extend these three and mix in `UpdateProjectAccessTrait`, which overrides `updateProjectAccess()`, calls `parent::` and then touches *team* access as well, with the whole thing wrapped so a billing-metadata failure cannot stop a run from being enqueued. Delete the method and cloud loses team access silently — the trait would override nothing.

On doing it in the source instead: that changes when it happens. `touchProject()` currently marks the project every time a run goes out; from the source it would mark it every time the row is read, so a project whose schedule is loaded but has not fired yet would start looking accessed. `accessedAt` feeds cleanup and billing in cloud, so I would rather not move that semantics as a side effect of this PR.

If you want the wrapper gone, the shape that works is the source taking an `onProject` closure and cloud passing its own — one seam instead of two, and the tasks call nothing. That needs the cloud PR in the same wave; I owe you one anyway for `ScheduleBackups`, so say the word and I will do it there.

src/Appwrite/Platform/Tasks/ScheduleSource.php
```diff
@@ -0,0 +1,297 @@
...
+     * any change feed.
+     */
+    #[\Override]
+    public function snapshot(): iterable
+    {
+        $this->snapshotted = 0;
+
+        // A snapshot is the whole truth, so the live view is rebuilt from it
+        // rather than accumulated.
+        $this->live = [];
```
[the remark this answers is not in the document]
loks0n: This was real, and it is gone by construction rather than by a guard. The live map it describes no longer exists: currency is answered by `Scheduler::isCurrent()` against the scheduler entries, and the library diffs that map — it never clears it before rebuilding, so there is no window where an unchanged schedule reads as not-current. Nothing sleeps across a relist any more either: the library owns the wait, and no dispatch path spawns a coroutine.

src/Appwrite/Platform/Tasks/ScheduleMessages.php
```diff
@@ -31,41 +63,100 @@ public static function getCollectionId(): string
...
+            \go(function () use ($schedule, $occurrence, $dbForPlatform): void {
+                try {
+                    $this->updateProjectAccess($schedule['project'], $dbForPlatform);
 
-                $publisherForMessaging->enqueue(new MessagingMessage(
-                    type: MESSAGE_SEND_TYPE_EXTERNAL,
-                    project: $schedule['project'],
-                    messageId: $schedule['resourceId'],
-                ));
+                    $this->publisher?->enqueue(new MessagingMessage(
```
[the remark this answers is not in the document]
loks0n: Addressed differently from the suggestion. The library now checks currency itself immediately before each hand-over, and reconciles while it waits so a cancellation during the wait is visible — so a `isLive()` call inside the handler would be checking the same thing twice. The coroutine the suggestion patches is also gone: messages publish inline, which is what makes "the window is committed" mean "the message was published" rather than "a coroutine was spawned".

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
loks0n: Gone in 2013a0b2. `exit(1)` removed, so a coroutine can call `action()` directly — it builds, runs and blocks, and returns if the loop ever stops. A supervised task restarts on a zero exit anyway, and the stopped span carries the reason.

src/Appwrite/Platform/Tasks/Schedule.php
loks0n: Gone in 2013a0b2 — the combined task calls each `action()` inside its coroutine. Loading happens concurrently now, which you confirmed is fine.

src/Appwrite/Platform/Tasks/ScheduleMessages.php
loks0n: Removed in 2013a0b2. With `boot()` gone there was nothing left for the explicit call to buy, so `run()` reconciles on its first tick as you said. The boot span went with it, and the snapshot counter behind it is out of `DatabaseSchedule` too — `schedule.entries` already reports that number every tick.

src/Appwrite/Platform/Tasks/ScheduleExecutions.php
loks0n: Done in 916dd239 — `DatabaseSchedule::touchProject()` is gone and each task holds the write itself, which is the same method cloud overrides to add team access.

The wrapper stays anemic-looking but is no longer a wrapper: it is the behaviour, in the class that owns the dispatch that needs it. The source is back to reading rows.

Cost is twelve identical lines in three places, which is the trade I took over adding a class whose only job is to stop them being written three times. `parent::updateProjectAccess()` in cloud's trait still resolves.

PR #13273, opened 2026-08-18
Refactor webhook dispatches to use Utopia Fetch client

### LINKED ISSUE #13204: refactor: use utopia-php/client for HTTP webhook dispatches
### 🔖 Enhancement description

refactor `Webhooks.php` worker to use `Utopia\Fetch\Client` instead of raw `curl_*` functions.

### 🎤 Pitch

While reviewing PR #13196, @ChiragAgg5k  suggested that we should be using the utopia client library (`Utopia\Fetch\Client`) for dispatching webhooks in `Webhooks.php` instead of manually setting up cURL.

This follow-up issue is to track refactoring the worker to use `Utopia\Fetch\Client` so HTTP calls are standardized across the codebase.

Chehak7:
## Summary

- Replace direct cURL calls in the webhook worker with the Utopia Fetch client and cURL adapter.
- Preserve webhook timeouts, redirects, TLS verification controls, Basic authentication, request headers, signatures, and the 5 MB response limit.
- Report transport failures through `FetchException` while retaining the existing retry and logging behavior.

## Why

Using the shared Utopia Fetch abstraction keeps webhook HTTP dispatch consistent with the rest of the project and removes worker-specific cURL handling.

Closes #13204

## Testing

All checks were run inside a Docker sandbox:

- Pint: 1,725 files passed.
- PHPStan: no errors.
- PHPUnit unit suite: 1,087 tests and 17,338 assertions passed (12 deprecations and 5 PHPUnit notices).


Chehak7:
> Thanks for catching this! I double-checked the exact `utopia-php/fetch` version locked by the project (v1.1.2), and `Client::fetch()` does support the `chunks` argument. I also verified it with a small runtime test inside Docker—the callback was invoked successfully.  
>
> It looks like this is a false positive, so I don’t think a code change is needed here. We also need the callback to preserve the existing 5 MB response-size limit.

loks0n:
Hey, we generally use utopia-php/client know, fetch is deprecated

Chehak7:
Hey @loks0n! Thanks for pointing that out. I’ve updated the implementation to use `utopia-php/client` instead of the deprecated Fetch client. The changes are pushed and ready for review. Please let me know if anything else needs to be adjusted.

loks0n (approved):

PR #13286, opened 2026-08-19
feat: add self-hosted usage service

ChiragAgg5k:
## Summary

Adds the generic Usage service to self-hosted Appwrite, backed by a private single-node ClickHouse instance.

- enables self-hosted usage collection by default through `_APP_USAGE_STATS=enabled`, with an explicit disabled no-op path
- keeps existing `v1-stats-usage` and `v1-stats-resources` queue contracts and DI names
- adds tenant-scoped event and gauge ingestion using project sequence as the server-derived tenant
- adds `GET /v1/usage/events` and `GET /v1/usage/gauges` with `usage.read`, filtering/grouping/bucket budgets, `last|max`, and empty-series responses
- adds idempotent schema setup at boot, schema-aware health, resource snapshots, realtime concurrency sampling, and tenant purge on project deletion
- adds default private ClickHouse `26.4.3-alpine` Compose wiring, persistent storage, authentication, healthcheck, and combined/separate worker/task registration
- pins `utopia-php/usage` ^0.14.0 (`baeef33bbcb673265550c83fe437cfa5c400a1e0`)

## Architecture

`Appwrite\\Usage\\Context` now captures generic request/resource metadata while preserving its existing API and Cloud subclass compatibility. Legacy Usage and StatsResources payload envelopes remain readable; dimensions are additive fields inside metric/gauge rows.

`country` is stored lowercased and filter values are folded the same way through `Context::normalizeCountry()`. ClickHouse compares case-sensitively, so before this an `equal("country", ["US"])` filter returned an empty series.

`Appwrite\\Usage\\Connection` owns a lifecycle-local pooled ClickHouse client and the shared `projects_*` namespace. Request reads bind `Utopia\\Usage\\Tenant` to the authenticated project's sequence; tenant is never request-controlled.

The existing publishers feed one consumer per stable queue. Inserts intentionally retain Cloud's current best-effort, non-idempotent behavior; arbitrary insert retries or deduplication were not added. Schema-not-ready is treated separately and fails message handling before insert, making the failure visible rather than acknowledging a missing-schema write.

Schema setup runs from `app/http.php` at boot, alongside the platform and project databases, on the same retry ladder. Giving up logs a warning rather than blocking boot: reads and ingestion both gate on `Connection::isReady()`, so they stay correct while the schema is missing. The `usage-setup` CLI task is retained for repair without a restart. There is deliberately no one-shot setup container -- no other schema in the repo has one, and gating the API on it let analytics storage take the API down.

Health probes connectivity plus the raw event, daily aggregate, and gauge objects. Raw events and daily aggregates have a 180-day TTL; gauges remain unlimited per the approved retention policy.

## Cloud compatibility

Generic Usage module, worker, and task registration is gated to `_APP_EDITION=self-hosted`, preventing duplicate Cloud routes and competing queue consumers before downstream adoption.

The read endpoints are built to be subclassed rather than copied. Dimension/filter/interval constants and the query helpers are `protected` and read through `static::`, so an override actually takes effect. Entitlement is an injected `usagePolicy` object: the self-hosted implementation allows everything, since self-hosted has no plans, addons, or reseller namespaces, and Cloud binds a subclass carrying its plan-history, premium-geo, and affiliate rules. It is injected rather than hung off the action because actions are shared across coroutines, so per-request state cannot live on `$this`. This lets the Cloud PR delete its duplicated copies of both endpoints instead of maintaining them in parallel.

Verified directly against the current Cloud checkout:

- `Appwrite\\Cloud\\Usage\\Context` loads against the enriched parent
- `Appwrite\\Cloud\\Event\\Message\\Usage` retains constructor compatibility
- existing Cloud queue names, publisher resource names, and optional message extension seams are unchanged

Cloud-only policy remains downstream: billing/legacy dual-write, affiliates and `console_*` tenants, premium geo, team billing dimensions, plan history, WAF/dedicated database metrics, Audit ClickHouse, and multi-region infrastructure.

## ClickHouse operations

Default Compose includes authenticated `clickhouse/clickhouse-server:26.4.3-alpine` on the internal `appwrite` network only, with no published ports, a persistent `/var/lib/clickhouse` volume, authenticated `SELECT 1` healthcheck, and raised `nofile`.

External `_APP_CONNECTIONS_DB_USAGE` remains supported. `_APP_USAGE_PASS` is used to bootstrap the bundled user; on an existing volume operators must run `ALTER USER` before changing the environment and DSN because image initialization does not rotate persisted credentials.

Live checks performed:

- ClickHouse reached authenticated healthy state with no host port publication
- `usage-setup` succeeded repeatedly
- dropping `projects_usage_gauges` made schema health return `healthy=false, schemaReady=false`; rerunning setup repaired it
- real event/gauge inserts and tenant-scoped reads returned event `3` and gauge `7`
- physical schema showed TTL on events/daily and unlimited gauges

## Validation

Library unit tests were dropped in `c88c65d8` and have been restored and extended:

- `tests/unit/Usage` — concurrency flooring, country fold, `isReady` TTL / cache expiry, `parseFilterQueries`, late-bound bucket budget, catalogued disabled / not-ready exceptions
- `tests/unit/Event/Message/UsageTest.php` — legacy and dimensioned payload envelopes
- `tests/unit/Workers/StatsUsageTest.php` — tenant resolution and disabled no-op (adapted to Accumulator + `getDatabasesDB`)

E2E (`tests/e2e/Services/Usage` is already in the CI matrix and is now also listed in `phpunit.xml`):

- empty-series / flat aggregate / scope smoke (skipped when `_APP_USAGE_STATS=disabled`)
- invalid `queries[]` returns `general_query_invalid` (400), not 500
- `_APP_USAGE_STATS=disabled` returns catalogued `general_usage_disabled` (403), not 500 — skipped on the default enabled CI stack

Queue-to-API ingest and tenant isolation are **not** in this suite. This repo has no existing usage-worker E2E harness, and this CI job does not run a dedicated ingest path. Inventing one would be fake coverage.

Review-fix follow-ups also landed:

- `$usage` in the api init/shutdown hooks is typed as `Appwrite\\Usage\\Context`; the redundant `instanceof` guards are gone
- `GENERAL_USAGE_DISABLED` is 403 (service-disabled neighbor). Schema-not-ready is a distinct `GENERAL_USAGE_NOT_READY` 503, so a missing catalog row cannot become an undefined-index 500
- advertised Cloud interval / bucket constants are read through `static::`

This branch was rebased onto latest `origin/main`.

## Why this PR is draft

The focused implementation checks are green, but queue-to-API ingestion still needs a real worker harness before this is merge-ready. Remaining before marking ready:

- add/execute queue-to-API event and gauge ingestion with polling
- add/execute cross-project tenant isolation
- add/execute project deletion and post-delete queued-message coverage
- complete/verify all generic per-resource storage/database/collection/document gauges expected by Console
- add an Appwrite-owned schema version/checksum ledger and drift validation beyond required-object readiness
- validate previous-stable upgrade, persisted credential rotation, backup/restore, and constrained-resource behavior
- decide least-privilege migration versus runtime ClickHouse credentials for production guidance

Best-effort insert loss/duplication remains intentional and must stay observable/documented. Dimensioned sparse `aggregate=max` retains Cloud's current no-carry-forward limitation.

## Required follow-up dependencies

1. **Cloud PR:** adopt the upstream generic module/workers/models while retaining Cloud-only billing, affiliate/console namespace, premium geo, plan, team, dual-write, regional, and dedicated database behavior. Test against this exact Appwrite SHA and keep one consumer per queue.
2. **Console/Vibes PR:** enable `usageStats` for self-hosted only when backend readiness is exposed, replace the current hardcoded self-hosted-off profile, and complete the site usage surface decision.
3. **SDK/spec generation:** regenerate server/client specs from these method labels, including gauge `aggregate=last|max`, and update the Console SDK dependency in lockstep.

No merge is requested until those gates and downstream compatibility checks pass.


ChiragAgg5k:
Console companion draft: https://github.com/appwrite/vibes/pull/155 (kept draft until this backend PR is merged and available in a test image).
Cloud adoption draft: https://github.com/appwrite-labs/cloud/pull/5378

app/controllers/general.php
```diff
@@ -809,6 +809,53 @@ function router(Http $utopia, Database $dbForPlatform, callable $getProjectDB, S
     }
 }
 
+Http::init()
+    ->groups(['api'])
+    ->inject('usage')
+    ->inject('request')
+    ->inject('geoRecord')
+    ->action(function ($usage, Request $request, GeoRecord $geoRecord) {
```
loks0n: Why not typed?

app/controllers/general.php
```diff
@@ -809,6 +809,53 @@ function router(Http $utopia, Database $dbForPlatform, callable $getProjectDB, S
...
+
+Http::shutdown()
+    ->groups(['api'])
+    ->inject('usage')
+    ->inject('response')
+    ->inject('project')
+    ->action(function ($usage, Response $response, Document $project) {
+        if (!$usage instanceof \Appwrite\Usage\Context) {
+            return;
+        }
```
loks0n: same here, why not typed?

lohanidamodar (approved):

PR #13304, opened 2026-08-21
Feat LDAP authentication

harsh-sikarwar:
# feat: LDAP authentication

Lets a project authenticate users against a directory server, so staff sign in
with existing corporate credentials. Serves on-prem enterprises with Active
Directory or OpenLDAP and no cloud identity provider — a different customer from
SAML.

Draft for direction check, per the discussion. Console UI is a separate repo and
not included.

## Surface

```
POST  /v1/account/sessions/ldap    { username, password }  -> session
PATCH /v1/project/auth/ldap        { host, baseDn, ... }   -> config
```

## How it works

Search-then-bind: bind as a service account, find the user's entry, then bind
again as that entry with the password supplied. A successful second bind is the
proof of identity. Everything after that — session document, geo enrichment,
cookies — is the same code the email/password route already uses.

The password is used transiently and never stored. Provisioned accounts have no
local password at all, so they can only ever be authenticated by the directory.

## The library choice is the design

`ext-ldap` is a blocking C extension and Swoole has no hook for it, so every
bind would stall the worker's event loop.

`freedsx/ldap` is pure PHP over `stream_socket_client`, which `SWOOLE_HOOK_TCP`
does hook. Measured before building anything, in the Appwrite runtime image
against a real OpenLDAP:

```
10 concurrent binds
  slowest single: 24.2 ms
  sum of all:    229.7 ms   <- what blocking would cost
  actual elapsed: 25.0 ms
```

An unrelated coroutine kept running throughout. `symfony/ldap` and
`laminas/ldap` both wrap `ext-ldap` and would block.

No new PHP extension, so the Dockerfile is untouched.

## Provisioning

A successful bind creates an account when none exists — the directory stays the
source of truth and nobody creates accounts by hand.

An optional `provisionFilter` restricts which entries are eligible, typically by
group. It is evaluated on **every** sign-in, not only the first, so removing
someone from the group revokes their access rather than merely preventing a new
account. Same filter, same place — enforcing it properly costs nothing.

## Security

- Wrong password, unknown user, and a user outside the provisioning filter are
  **indistinguishable** — a directory must not be probeable for valid usernames.
- Empty passwords are rejected before any bind: LDAP reports an empty-password
  bind as success, which would otherwise let anyone in.
- Filter values are escaped per RFC 4515. A username of `*` would otherwise
  match every entry in the subtree.
- A search returning more than one entry is treated as no match — binding as an
  arbitrary one of several is a coin toss over identity.
- The user filter must contain `{{username}}`, enforced at construction.

## On multiple directories per project

Checked how others do it: **every comparable product supports several, none cap
it at one.** Keycloak federates multiple per realm and picks by priority; Auth0
routes by email domain; WorkOS is per-organization; Okta supports multiple
integrations. Where limits exist they are commercial, not architectural. Drivers
are subsidiaries, M&A leaving several AD forests, and contractor directories.

This exposes **one** directory, but stores it as a list, so allowing several
later is a change to this action and the console rather than a migration. The
version that would need a migration is storing the fields flat on the project
document.

If we do add multiple, the rule worth copying from Keycloak: when a user matches
a directory but the bind fails, fail the sign-in rather than trying the next —
otherwise duplicate usernames authenticate against the wrong store.

## Two things flagged rather than solved

**1. Existing sessions survive a directory change.** Disabling a user prevents
new sessions but does not invalidate existing ones — session validation never
re-contacts the provider, and the default duration is one year. True of OAuth2
and SAML too, but conspicuous here because "the employee left" is the case this
feature is for. Documented; revocation is `DELETE /v1/users/:userId/sessions`.
Shorter LDAP session durations and periodic revalidation are reasonable
follow-ups, the latter being a session-model decision rather than an LDAP one.

**2. One test fails, deliberately visible.** Adding `ldap` to
`app/config/auth.php` puts it in the auth-methods list, and
`ProjectAuthMethodId` in the vendored SDK is generated from that config. Until
the SDK is regenerated,
`ResponseTest::testProjectResponseCanHydrateGeneratedSdkProjectWithoutOAuth2Fields`
fails. Left visible rather than worked around, since regenerating is a real step
this feature needs — happy to handle it differently if there is a convention.

## Verified end to end

Against a real OpenLDAP server:

| | |
|---|---|
| Configure + enable | 200, live connection check passes |
| Sign in as alice | 201, `provider: ldap`, `providerUid: uid=alice,ou=people,...` |
| Provisioned account | `alice@appwrite.test` / `Alice Smith`, verified, **no password stored** |
| Repeat sign-in | reuses the account, 1 user after two sign-ins |
| Wrong password / unknown user / wildcard / empty password | all 401 |
| Group filter: member vs non-member | 201 vs 401 |

Unlike SAML this needs no public URL, tunnel, or third-party account — OpenLDAP
runs in Docker and the whole thing works offline, which makes it viable in CI.

**Gates:** 20 new unit tests, PHPStan level 4 clean, 1017 unit tests with the
single expected SDK-enum failure above.

## Follow-ups

- Console UI (separate repo)
- Multiple directories per project
- Group-to-team mapping, attribute sync after first sign-in
- Connection pooling
- Session revalidation against the directory

Full rationale, including what was rejected, is in
`docs/decisions/ldap-auth.md`.

## Checklist


app/config/auth.php
```diff
@@ -10,6 +10,13 @@
         'docs' => 'https://appwrite.io/docs/references/cloud/client-web/account#accountCreateEmailPasswordSession',
         'enabled' => true,
     ],
+    'ldap' => [
+        'name' => 'LDAP',
+        'key' => 'ldap',
+        'icon' => '/images/users/ldap.png',
+        'docs' => 'https://appwrite.io/docs/references/cloud/client-web/account#accountCreateLdapSession',
```
Meldiron:
```suggestion
        'docs' => 'https://appwrite.io/docs/references/cloud/client-web/account#accountCreateLDAPSession',
```

Lets use uppercase LDAP where possible

app/controllers/api/account.php
```diff
@@ -1094,6 +1096,325 @@
...
+    ->label('event', 'users.[userId].sessions.[sessionId].create')
+    ->label('scope', 'sessions.write')
+    ->label('auth.type', 'ldap')
+    ->label('audits.event', 'session.create')
+    ->label('audits.resource', 'user/{response.userId}')
+    ->label('audits.userId', '{response.userId}')
+    ->label('sdk', new Method(
+        namespace: 'account',
+        group: 'sessions',
+        name: 'createLdapSession',
```
Meldiron:
```suggestion
        name: 'createLDAPSession',
```
Similar here and probably in more places, uppercase LDAP will be better

app/controllers/api/account.php
```diff
@@ -1094,6 +1096,325 @@
...
+        description: '/docs/references/account/create-session-ldap.md',
```
Meldiron: Let's remove md file, and instad have multiline in-line description here. Check out some other endpoints (Proxy module I think) for example how that looks. There is special syntax to it with <<< AND EOD;

Meldiron (commented):
Did round of review, not yet finished.

Yet to review and finalize:
- src/Appwrite/Platform/Modules/Account/Http/Account/Sessions/LDAP/Create.php
- src/Appwrite/Auth/LDAP/Client.php

Some open topics for myself:
- freedsx/ldap, do we really need?
- cloud DB migration
- understand provision and user filters nessessity, or simplify them
- ensure proper test coverage after all changes are finished
- Greptile review 4 or 5 points
- ENsure password is encrypted

harsh-sikarwar:
## Provisioning filter

I removed `provisionFilter` and replaced it with `provisionGroupDn`:

```text
before:
provisionFilter: (&(cn=appwrite-users)(member={{username}}))

after:
provisionGroupDn: cn=appwrite-users,ou=groups,dc=appwrite,dc=test
```

This makes the separation clearer:

* `userFilter` finds the user trying to authenticate.
* `provisionGroupDn` checks whether the authenticated user is allowed to be provisioned.

Regarding `baseDn`, it only scopes where we search for users; it cannot express group membership. For example, Alice and Bob can both be under `ou=people`, while the group membership information lives separately under `ou=groups`.

The implementation supports both OpenLDAP-style `member`/`uniqueMember` groups and Active Directory's `memberOf`.

This also removes the confusing placeholder behavior where `{{username}}` meant different things in `userFilter` and `provisionFilter`.

The tradeoff is that arbitrary rules like `(department=Engineering)` are no longer supported. I think keeping the API focused on the actual group-membership use case is the simpler option for now, and we can add an advanced filter later if needed.

## Other fixes

While testing, I also found and fixed two unrelated issues:

* The blocked-user test expected `401`, while `403` is the correct response for an authenticated user who is not allowed access.
* `Update.php` was accidentally replacing `$project` with the boolean result of `purgeCachedDocument()`.

## Verification

* E2E: **14/14**
* Unit: **16/16**
* PHPStan level 4: **clean**

PR #13307, opened 2026-08-21
chore: require utopia-php/cache ^5.0

loks0n:
## What

Moves `utopia-php/cache` from `^4.0.0` to `^5.0.0`.

Cache 5.0.0 carries two changes:

1. **`Redis\Multiplexing`**: a caller's read deadline no longer tears down the TCP connection every other coroutine on that worker is sharing ([utopia-php/monorepo#152](https://github.com/utopia-php/monorepo/pull/152)).
2. It now requires **`circuit-breaker ^0.4`**, which replaced the consecutive-failure `threshold` with a failure rate over a window ([utopia-php/monorepo#153](https://github.com/utopia-php/monorepo/pull/153)).

## Neither affects this repository

- The cache adapters used here are `Filesystem`, `None`, `Pool`, `Redis` and `Sharding` — **not** `Multiplexing`.
- There is **no** `CircuitBreaker` usage anywhere in `app/` or `src/`, and no `threshold:` argument to migrate.

So the general cache API this depends on is byte-for-byte compatible; the major is about surfaces this repo does not touch.

## 🚧 Blocked — draft on purpose

`composer.lock` is deliberately **not** updated. Three released dependencies still pin cache `^4`, and Composer refuses the resolve:

```
Problem 3
  - utopia-php/database 7.2.4 requires utopia-php/cache ^4.0.0 -> found utopia-php/cache[4.0.0, 4.0.1, 4.0.2]
    but it conflicts with your root composer.json require (^5.0.0).
Problem 4
  - utopia-php/domains 3.0.0 requires utopia-php/cache ^4.0 -> ... conflicts ...
Problem 5
  - utopia-php/vcs 5.2.2 requires utopia-php/cache ^4.0 -> ... conflicts ...
```

Widening PRs are open for all three, each constrained as `^4.0 || ^5.0` so they do not force the upgrade on anyone:

- utopia-php/database#943
- utopia-php/domains#70
- utopia-php/vcs#137

Once those merge and are released, I will update `composer.lock` here and mark this ready.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -95,7 +95,7 @@
         "utopia-php/audit": "3.0.*",
         "utopia-php/auth": "^0.10",
         "utopia-php/balancer": "0.4.*",
-        "utopia-php/cache": "^4.0.0",
+        "utopia-php/cache": "^5.0.0",
         "utopia-php/cli": "^0.24",
         "utopia-php/compression": "0.1.*",
         "utopia-php/config": "1.*",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "4fd078c5f4c652e6b07c639e275b520c",
+    "content-hash": "1cd34d77f1e40162d3873297b0b20742",
     "packages": [
         {
             "name": "adhocore/jwt",
@@ -3194,22 +3194,22 @@
         },
         {
             "name": "utopia-php/cache",
-            "version": "4.0.2",
+            "version": "5.0.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/utopia-php/cache.git",
-                "reference": "92e02dab63606234b993b841ebf4c58845dd4620"
+                "reference": "0d0752785fc81b5afd6571f4291763166a6532bc"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/utopia-php/cache/zipball/92e02dab63606234b993b841ebf4c58845dd4620",
-                "reference": "92e02dab63606234b993b841ebf4c58845dd4620",
+                "url": "https://api.github.com/repos/utopia-php/cache/zipball/0d0752785fc81b5afd6571f4291763166a6532bc",
+                "reference": "0d0752785fc81b5afd6571f4291763166a6532bc",
                 "shasum": ""
             },
             "require": {
                 "ext-json": "*",
                 "php": ">=8.4",
-                "utopia-php/circuit-breaker": "^0.3",
+                "utopia-php/circuit-breaker": "^0.4",
                 "utopia-php/pools": "^2.0",
                 "utopia-php/telemetry": "^0.4"
             },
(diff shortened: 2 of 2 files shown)


loks0n:
Unblocked and lock updated — marking ready.

The three dependencies that pinned cache `^4` have been released allowing `^4.0 || ^5.0` (database 7.2.6, domains 3.0.1, vcs 5.2.3), so the resolve completes now.

Lock moves exactly five packages:

```
utopia-php/cache            4.0.2 -> 5.0.0
utopia-php/circuit-breaker  0.3.2 -> 0.4.0
utopia-php/database         7.2.4 -> 7.2.6
utopia-php/domains          3.0.0 -> 3.0.1
utopia-php/vcs              5.2.2 -> 5.2.3
```

`circuit-breaker` appears only because cache requires it; there is no direct usage in this repository, and no `threshold:` argument to migrate. The cache adapters used here are `Filesystem`, `None`, `Pool`, `Redis` and `Sharding` — not `Multiplexing` — so neither of cache 5's two breaking surfaces is reached.

`composer validate` passes. Note for reviewers: I resolved with `--ignore-platform-req=ext-scrypt --ignore-platform-req=ext-mongodb` because my machine lacks scrypt and has mongodb 2.3.3 against `utopia-php/mongo`'s `2.1.*`; those flags affect only what Composer will install locally, not the resolved versions written to the lock. CI resolves without them.

loks0n:
CI note: one job failed, `Tests / E2E / PostgreSQL (dedicated) / FunctionsSchedule`. It looks like an existing flake rather than anything from this change.

`FunctionsSchedule` fails intermittently on runs unrelated to this branch — run `23339202955` failed it across three variants (plain, Shared V1, Shared V2) while run `23339399632` passed it. Schedule E2E tests are timing-sensitive by nature.

There is also no mechanism by which this change could reach it. The diff is a dependency constraint plus a lock; cache 5.0.0 differs from 4.x only in `Redis\\Multiplexing` deadline handling and in requiring circuit-breaker ^0.4. This repository uses neither — the adapters here are `Filesystem`, `None`, `Pool`, `Redis` and `Sharding`, and there is no `CircuitBreaker` usage at all — so the code paths a scheduled-function test exercises are identical before and after.

The workflow was still running when I tried, so I could not re-run just the failed job. Worth one re-run before reading anything into it.

loks0n:
Updated CI picture: **98 of 100 jobs pass.** Two E2E jobs are red:

- `Tests / E2E / PostgreSQL (dedicated) / FunctionsSchedule`
- `Tests / E2E / PostgreSQL (shared) / Sites`

Two things suggest these are the matrix flaking rather than this change:

**Two of the original four went green on their own.** An earlier read of this same run had four failures — the two above plus `MongoDB (dedicated) / GraphQL` and `MongoDB (shared) / Sites`. Those two are now passing, with no change pushed in between. `Sites` failing on Postgres while passing on Mongo, in the same run, is not a shape a dependency change produces.

**There is no mechanism.** I diffed what cache 5.0.0 actually contains against 4.0.2, rather than assuming:

```
packages/cache/composer.json
packages/cache/composer.lock
packages/cache/docs/multiplexing.md
packages/cache/src/Cache/Adapter/Redis/ConnectionContext.php
packages/cache/src/Cache/Adapter/Redis/IdleConnectionException.php   (new)
packages/cache/src/Cache/Adapter/Redis/Multiplexing.php
packages/cache/src/Cache/Adapter/Redis/TimeoutException.php          (new)
packages/cache/tests/...
```

Nothing in `Adapter/Redis.php`, `Pool.php`, `Sharding.php`, `Filesystem.php`, `None.php` or `Cache.php` — which are the only cache classes this repository uses. `Redis\\Multiplexing` is not used here at all, and neither is `CircuitBreaker`. So the code executed by a Sites or scheduled-functions test is byte-for-byte what it was before.

I could not read the two job logs to confirm the individual assertions: GitHub does not serve job logs while the parent run is still in progress, and it has been queued throughout. Once it completes, a re-run of just those two is the cheap confirmation — happy to do that, or to dig into the assertions if they persist.

PR #13309, opened 2026-08-21
fix(jobs): run pre-publish work outside the deployment lock

loks0n:
## Problem

`Jobs::action()` holds `jobs-deployment:<id>` across the entire callback, including the terminal success path. Downstream, that path does real work — Appwrite Cloud overrides `finalize()` to distribute a site build to every edge and waits for each to report it available, bounded only by `Distribution::POLL_TIMEOUT_SECONDS` (60).

Three constants that don't fit together:

| constant | value |
|---|---|
| `Distribution::POLL_TIMEOUT_SECONDS` (cloud) | **60** |
| `Jobs::LOCK_TTL` | **30** |
| `Jobs::LOCK_TIMEOUT` | **10** |

**Starvation.** Every other callback for that deployment gets `LOCK_TIMEOUT` (10 s) to acquire. Across a 60 s hold, none of them can ever win — they are destroyed, not delayed. That includes the sibling half of the success join.

**Lost mutual exclusion.** `LOCK_TTL` is 30 s, so a 60 s poll *structurally* outlives the lease — not a narrow race. Once it lapses, another worker can acquire the same key and finalize the same deployment concurrently. The only thing limiting the damage is that `release()` is token-guarded, so the original doesn't delete its successor's key.

Observed in one 12-minute production window, as **successful** (`level=info`) spans — so the time is spent *inside* the lock, not waiting for it:

| span | edges |
|---|---|
| **82.11 s** | `build_edge_count=6`, `build_edge_failed_count=4` |
| 35.21 s / 34.31 s | `build_edge_count=6`, success |
| 12.60 / 12.27 / 11.30 / 11.17 s | `build_edge_count=6`, success |

## Change

Split the success path into two short locked sections with the long work in between:

1. **locked** — dedupe, read, dispatch the handler. `ready()` no longer finalizes inline: once the join is in and detection and the size limit have passed, it hands the build size back to `action()` through a by-reference slot.
2. **unlocked** — `publish()`, a no-op extension point here. Cloud's edge distribution moves into it.
3. **locked** — re-read, verify still non-terminal and non-canceled, `finalize()`.

Failure paths are untouched: they still finalize inside the first locked section, because they're fast and have nothing to publish.

## Details worth reviewing

- **The slot is by-reference, not `$this` state.** The worker runs many coroutines against one instance, so instance state would cross-talk between deployments.
- **Re-reading in the third step is load-bearing.** `publish()` runs unlocked, so a cancel can land while it's in flight; finalizing the document read beforehand would resurrect a canceled build.
- **All three events that can complete the join thread the slot** — `exit`, `complete`, *and* the `manifest` artifact. `onArtifact` also calls `ready()`, which is easy to miss; omitting it would leave builds whose last marker is the manifest stuck in `building` forever.
- **Duplicate publishes are possible by design.** Dedupe stays inside the lock — moving it out would mark an event processed even when the lock timed out, losing it permanently — so a redelivered event can re-enter `publish()`. Overrides must tolerate that; it's documented on the method.
- **`append()`** concatenates a `buildLogs` addition server-side via `Operator::stringConcat`. Once `publish()` is unlocked its progress lines race anything else writing the column, so a read-modify-write there would silently drop them. The bound stays in the caller: `Operator`'s validator only checks a concat against the declared size when given the current document, and `updateDocuments()` passes `currentDocument: null` (no single old row in a bulk update). `buildLogs` declares `size=APP_LOG_LENGTH_LIMIT` but is physically `MEDIUMTEXT`, so an unbounded concat would grow far past the declared size before anything failed.

## Contract change

`onExit`, `onComplete`, `onArtifact` and `ready()` each gain a trailing `?array &$pending = null`. They're `protected` extension points; no override of them exists in CE or in cloud (which overrides `finalize()` and `activate()`), but downstream forks that override them will need the extra parameter.

## Testing

Unit suite: 997 tests. The 3 failures (`ExtensionsTest::testMaxminddb`, `Auth\KeyTest::testDecode`, `Vcs\FactoryTest::testRegistryEntries`) reproduce identically on a clean tree — pre-existing and unrelated to this change. PHPStan and Pint clean.

There is no unit-level coverage of `action()`'s locking structure — it needs 17 injections and a real lock — so the behaviour this restructures is guarded by the Sites E2E deployment-to-`ready` path rather than by a new unit test. Stating that plainly rather than leaving it implied.

### DIFF
--- a/src/Appwrite/Platform/Modules/Functions/Workers/Jobs.php
+++ b/src/Appwrite/Platform/Modules/Functions/Workers/Jobs.php
@@ -23,6 +23,7 @@ use Utopia\Cache\Cache;
 use Utopia\Database\Database;
 use Utopia\Database\DateTime;
 use Utopia\Database\Document;
+use Utopia\Database\Operator;
 use Utopia\Database\Query;
 use Utopia\Platform\Action;
 use Utopia\Queue\Message;
@@ -42,8 +43,10 @@ use Utopia\System\System;
  * atomic and a lock timeout retries cleanly).
  *
  * Handlers are protected extension points: downstream workers (e.g. cloud)
- * override finalize() and wrap parent:: — before it for work that must
- * precede 'ready' (edge distribution), after it for post-activation work.
+ * override finalize() and wrap parent:: for post-activation work. Work that
+ * must precede 'ready' and takes real time (cloud's edge distribution) belongs
+ * in publish() instead, which runs outside the lock — see the second locked
+ * section in action() for why.
  */
 class Jobs extends Action
 {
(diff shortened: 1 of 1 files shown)


loks0n:
Closing unfixed in favour of a different architecture.

This moved the wait out of the lock but kept it in cloud's worker — a patch on the wrong ownership boundary. The edge already runs a Kubernetes Job for the download and already reads its own Job conditions; cloud was polling it 60 times to discover something the edge knew. And the reverse feed channel this needs is a named, empty slot the codebase already documents (`api-feed-edge-<feed>`).

Under that design cloud never blocks, so there is nothing to hold outside the lock and the `publish()` seam has no reason to exist.

Design: https://claude.ai/code/artifact/09eecbda-14a3-40c8-8f99-e1b6c2fddc91

Issue #5378, opened 2023-04-14
🚀 Feature: Make MAINTENANCE_INTERVAL and env variable


danilo73r:
### 🔖 Feature description

Allow MAINTENANCE_INTERVAL to be an environment variable to get more control and customization over the container execution. Considering a min  value to avoid overload errors (maybe 15 min as min) and a max value to avoid let the server without maintenance for a long time (maybe 1 day as max).

### 🎤 Pitch

I understand that 1 function will have 1 runtime container.

This will allow to have better control over the number of idle containers and server resources. Specially if I have a lot of functions, maybe 20% will be frequently functions, the other not. So it will be helpful to avoid having 80% of containers in an idle state by an hour (MAINTENANCE_INTERVAL = 3600) in the worst case. 

The other case if 80% or more functions are frequently used, maybe run maintenance will be better each 8 hours or more.
 

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

joeyouss:
Hi - thank you for raising this.

We do have a timeout env for functions: https://appwrite.io/docs/environment-variables#functions

This supposedly could be useful. Let me have a chat about this with my team and get back!

stnguyen90:
`_APP_FUNCTIONS_MAINTENANCE_INTERVAL` has now been added. See https://appwrite.io/docs/advanced/self-hosting/environment-variables#functions.

PR #13322, opened 2026-08-24
fix(usage): lower the stats-resources interval floor to 1 second

ChiragAgg5k:
## Problem

The stats-resources task clamps its loop interval with `max(60, ...)`. Test stacks legitimately configure `_APP_STATS_RESOURCES_INTERVAL=2` (Cloud CI and `.env.tests`) so resource gauges converge within seconds; the 60s floor silently overrides that, gauges recompute only once a minute, and Cloud's `AppsUsageTest` — whose `assertEventually` timeout is exactly 60s — fails on every run since the upstream usage adoption (appwrite-labs/cloud#5378 replaced the Cloud scheduler, which honored the configured interval, with this task).

## Fix

Floor the interval at 1 instead of 60. The floor exists to keep a zero or negative value from spinning the loop, and 1 still does that; production keeps its 3600 default.

## Testing

There is no unit seam around `Console::loop`'s interval, so no unit test covers this line; the regression coverage is Cloud's `tests/e2e/Services/Apps/AppsUsageTest` (red on main today at the 60s floor, green at the configured 2s interval once Cloud relocks onto this fix).

### DIFF
--- a/src/Appwrite/Platform/Tasks/StatsResources.php
+++ b/src/Appwrite/Platform/Tasks/StatsResources.php
@@ -42,7 +42,9 @@ class StatsResources extends Action
         }
 
         $this->disableSubqueries();
-        $interval = max(60, (int) System::getEnv('_APP_STATS_RESOURCES_INTERVAL', 3600));
+        // Floor of 1 guards against a zero/negative interval spinning the loop;
+        // test stacks legitimately run short intervals (Cloud CI uses 2s).
+        $interval = max(1, (int) System::getEnv('_APP_STATS_RESOURCES_INTERVAL', 3600));
 
         Console::loop(function () use ($dbForPlatform, $publisherForStatsResources, $usageConnection): void {
             if (!$usageConnection->isReady()) {


HarshMN2345 (approved):

ChiragAgg5k:
Self-review notes on the floor, since this is the line that bit Cloud CI after #5378.

**Verdict:** fine to merge once the Usage E2E job is green. This is the scheduler clamp only (`StatsResources` task → `Console::loop`). It does not touch `/v1/usage/*` `interval` (`1m`/`15m`/`30m`/`1h`/`1d`), SDK specs, or the worker.

**What it does:** `max(60, env)` → `max(1, env)` so `_APP_STATS_RESOURCES_INTERVAL=2` is actually 2s. The 60s floor from #13286 was silently winning over Cloud CI / `.env.tests`, and `AppsUsageTest`'s 60s `assertEventually` cannot wait out a 60s snapshot. Prod default stays 3600.

**Not a 1s ClickHouse/Maria storm on the default path.** Concurrency sampling already no-ops until a 300s bucket closes. Gauges are point snapshots; a 2s write just densifies samples inside a 1m read bucket. The worker is still `maxCoroutines = 1`.

**The thing I would not claim:** that a floor of 1 is “the same protection, just lower.” It only stops `0`/negative from spinning `Console::loop`. A typo (`''`, `'foo'` → `(int)0` → `1`) now becomes a 1s enqueue of every 24h-active project, and the serial recount worker will not keep up. Before, that same typo became 60s. If we touch this again, I’d rather `if ($interval < 1) $interval = 3600` than `max(1, …)`.

No Luke comments on this PR. Harsh’s approve is LGTM-without-notes; Greptile 5/5 skipped the queue-flood case.

PR #13323, opened 2026-08-24
refactor(functions): fetch variables fresh for scheduled executions

Meldiron:
## What does this PR do?

Stops the scheduler from caching function variables and stops variable APIs from updating the schedule document.

**Before:** the `schedule-functions` scheduler snapshotted the entire function document — `vars` and `varsProject` included — into memory, and the functions worker executed scheduled runs directly from that snapshot. This meant:

- Every variable Create/Update/Delete had to bump `resourceUpdatedAt` on the schedule document just to invalidate the scheduler's cache.
- Project-level variable changes never bumped it, so scheduled executions ran with stale `varsProject` values indefinitely.
- Decrypted variable values (including secrets) sat in the scheduler's process memory and in every dispatched queue payload.

**After:**

- `Schedule\Source\Functions::resource()` skips the `subQueryVariables` / `subQueryProjectVariables` filters, so the snapshot physically cannot carry variable values.
- The functions worker re-fetches the function from the project DB when handling a `schedule` dispatch, so variables (and any other function changes, e.g. a new deployment) are read fresh at execution time. A function deleted between dispatch and execution is now skipped instead of executed from a stale snapshot.
- The variable endpoints (`Functions/Http/Variables/{Create,Update,Delete}.php`) drop the schedule-document update and their now-unused `dbForPlatform` / `authorization` injections.

Cron/deployment changes still propagate to the scheduler as before via the `resourceUpdatedAt` bumps in `Functions/Update.php` and `Deployment/Update.php` — those own the `schedule` / `active` attributes.

## Test Plan

- `composer lint` and PHPStan pass on all changed files.
- No existing e2e/unit tests asserted the removed schedule write.
- Manual: function on a cron schedule → update a function variable and a project variable → next scheduled execution receives the new values without any schedule-document write.

## Related PRs and Issues

None.

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Modules/Functions/Http/Variables/Create.php
+++ b/src/Appwrite/Platform/Modules/Functions/Http/Variables/Create.php
@@ -11,11 +11,9 @@ use Appwrite\SDK\Response as SDKResponse;
 use Appwrite\Utopia\Database\Validator\CustomId;
 use Appwrite\Utopia\Response;
 use Utopia\Database\Database;
-use Utopia\Database\DateTime;
 use Utopia\Database\Document;
 use Utopia\Database\Exception\Duplicate as DuplicateException;
 use Utopia\Database\Helpers\ID;
-use Utopia\Database\Validator\Authorization;
 use Utopia\Database\Validator\UID;
 use Utopia\Platform\Action;
 use Utopia\Platform\Scope\HTTP;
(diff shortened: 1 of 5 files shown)


Meldiron (commented):
lgtm

fogelito (approved):

PR #13340, opened 2026-08-25
fix(sites): raise screenshot cold-start budget to 60s

HarshMN2345:
Temporary mitigation while the browser-service fix and cold-start investigation continue — "patch first," per the thread.

## Problem

A slow SSR cold start behind a preview URL is aborted at 30s, so the screenshot fails even though the page would have loaded. The 30s ceiling exists at **four serial layers** in the screenshot path, and the smallest one wins — the observed failure was the worker's HTTP client giving up at exactly 30s (`Operation timed out after 30025ms with 0 bytes received`).

## Change — every layer raised together

| Layer | Before | After |
|---|---|---|
| router → executor `requestTimeout` | 30 | **60**, scoped to `$isPreview` |
| browser `page.goto` | 30000 (default) | **90000**, sent explicitly |
| screenshot worker → browser HTTP | 30 (`_APP_SITES_TIMEOUT`) | **120** |
| `_APP_SITES_TIMEOUT` reaching the worker | not passed | added to `worker-screenshots` |

Nesting is deliberate: the effective cold-start allowance is **60s** (the router→executor cap Matej asked for); the outer layers sit above it (goto 90s, worker 120s) so an inner success is never cut off by an outer bail.

**Blast radius is screenshots only.** The router bump is gated on `$isPreview`, so live function and site executions keep the 30s budget unchanged.

The `_APP_SITES_TIMEOUT` line fixes a real gap: the value was never passed into the `worker-screenshots` container, so it always fell back to the hardcoded default regardless of config.

## Cloud

For prod/staging, set `_APP_SITES_TIMEOUT=120` on the screenshots worker (already wired to the pod there). The code changes deploy with the normal image.

## Relation to #13315

#13315 (deterministic rule ordering + diagnostics + histogram) is the "proper fix" track and overlaps the worker/browser timeout lines. This patch is the ship-today mitigation and additionally covers the router→executor cap that #13315 does not. Expect a small merge conflict on the timeout lines; resolve toward these values.

## Verification

`php -l`, `composer lint`, and PHPStan level 4 pass on the changed files; `docker compose config` validates. The browser accepts the 90000 timeout (schema max 120000).


app/controllers/general.php
```diff
@@ -602,7 +604,9 @@ function router(Http $utopia, Database $dbForPlatform, callable $getProjectDB, S
                     cpus: $spec['cpus'] ?? APP_COMPUTE_CPUS_DEFAULT,
                     memory: $spec['memory'] ?? APP_COMPUTE_MEMORY_DEFAULT,
                     logging: $resource->getAttribute('logging', true),
-                    requestTimeout: 30,
+                    // Temporary mitigation: preview (screenshot) requests allow a
+                    // 60s cold start; live traffic keeps the 30s budget.
+                    requestTimeout: $isPreview ? 60 : 30,
                     responseFormat: Executor::RESPONSE_FORMAT_ARRAY_HEADERS
```
Meldiron, line 610: We can do 60 for everything, until we resolve cold starts

Meldiron (approved):

PR #13346, opened 2026-08-25
fix(presences): align upsert lookup with unique user ID

### LINKED ISSUE #13291: Presence upsert can be immediately invisible for console-session requests
### Summary

Console-session presence requests can intermittently fail to read a presence immediately after `PUT /v1/presences/:presenceId` returned 200.

The failure affects the `console` project. The write response contains a presence, but the next list request can return `total: 0`, and a subsequent update using the returned ID can return 404.

### Exact reproduction

Environment:

- Appwrite server source: `18998d796847f1b228e3d8512fe6ae2969d464de`
- PHP 8.5.9
- PHPUnit 12.5.33
- Kubernetes deployment
- MySQL console/platform database
- Redis cache
- Dedicated project databases enabled

From a clean deployment, warm the HTTP workers and run the Storage and Presences E2E directories serially:

```bash
for round in 0 1 2; do
  for request in $(seq 1 50); do
    curl --silent --output /dev/null http://localhost/v1/health &
  done
  wait
  sleep 2
done

vendor/bin/phpunit \
  --display-incomplete \
  --display-skipped \
  --exclude-group abuseEnabled \
  --exclude-group screenshots \
  --exclude-group ciIgnore \
  --exclude-group production \
  --exclude-group provider-release \
  tests/e2e/Services/Storage \
  tests/e2e/Services/Presences
```

The run executed 143 tests and produced these four failures:

```text
1) PresenceConsoleClientTest::testListPresences
Failed asserting that 0 is equal to 1 or is greater than 1.
PresenceBase.php:271

2) PresenceConsoleClientTest::testUpdatePresenceSparseFields
Failed asserting that 0 is equal to 1 or is greater than 1.
PresenceBase.php:122
PresenceBase.php:518

3) PresenceConsoleClientTest::testUpdatePresenceNestedMetadata
Failed asserting that 404 matches expected 200.
PresenceBase.php:620

4) PresenceConsoleClientTest::testDeletePresence
Failed asserting that 0 is equal to 1 or is greater than 1.
PresenceBase.php:122
```

The corresponding presence upserts returned 200. The failing helpers immediately listed or updated the just-upserted presence.

This is intermittent: the same test class passed in isolation, and 20 serial repetitions passed against a local PostgreSQL-backed installation. The captured failure was on a clean MySQL-backed deployment, so the database adapter or multi-worker request topology may be relevant.

### Minimal request sequence

The failing tests all reduce to this sequence, authenticated with a console session (`a_session_console`) and `X-Appwrite-Project: console`:

1. `PUT /v1/presences/<unique-id>` with `status: online` — returns 200.
2. Immediately `GET /v1/presences` with `Query.equal('userId', [<session-user-id>])` — intermittently returns `total: 0`.
3. In one case, `PATCH /v1/presences/<id-returned-by-put>` — returns 404.

### Expected behavior

Once the upsert returns 200, a subsequent request should be able to list, get, update, or delete the returned presence.

### Actual behavior

An immediate list can report no presence, or an immediate update can report that the ID returned by the successful upsert does not exist.

### Investigation notes

- The failures are limited to the console-session variant in the captured run; custom-project variants passed.
- The console project uses the platform database rather than a normal project database.
- `PresenceBase::resolvePresenceForUser()` performs the list immediately after the successful upsert, without a configured list-cache TTL.
- `Presences/HTTP/Upsert.php` first resolves an existing presence by `userInternalId`, then calls `upsertDocument()`.
- Update and Delete explicitly purge the Presence list cache; Upsert does not. However, the failing list requests use the default `ttl=0`, so this alone does not fully explain the reproduction.
- The returned-ID 404 may indicate that the existing-presence lookup and the unique `userInternalId` upsert selected different rows under the failing conditions.

Chehak7:
## What changed

- Resolve an existing presence by `userId`, matching the unique index used by database upserts.
- Pass the canonical user ID from both HTTP and Realtime presence handlers.
- Add regression coverage for native and transactional upsert paths.

## Why

The presence state previously looked up the canonical row by the non-unique `userInternalId`, while the database upsert resolves conflicts using the unique `userId` index. If those lookups selected different rows, the successful response could contain a non-canonical presence ID, causing an immediate list to return no presence or an update to return 404.

## Testing

All checks were run inside a Docker sandbox:

- PSR-12 formatting: passed
- PHPStan on changed files: passed
- Presence state regression tests: 2 tests, 12 assertions passed
- Full unit suite: 999 tests, 15,988 assertions passed
- Warmed Storage and Presences E2E sequence on MariaDB: 143 tests, 2,056 assertions passed

Closes #13291


Chehak7:
Hey maintainers! This PR is ready for review. When you get a chance, could you please take a look and let me know if any changes are needed from my side? Thank you!

Chehak7:
Hey @loks0n! and @ChiragAgg5k ! Just following up on this PR since it’s been a few days. Whenever you get a chance, could you please take a look? I’d really appreciate any feedback, and I’m happy to make any changes if needed. Thanks!

PR #13360, opened 2026-08-26
Refuse variable keys that are invalid env var names

levivannoort:
## What does this PR do?

Variable keys become container env var names at build and runtime. A key holding a stray tab or UTF-16 bytes (e.g. a `.env` file saved as UTF-16 and imported, storing every character with an interleaved NUL byte) passed the plain length check, was stored, and then invalidated the build job manifest at submission -- failing every deployment for that resource with an opaque "internal error", and (on cloud) a 500 out of the VCS webhook per affected push.

Two layers, two deliberately different rules:

- **Endpoint guard** -- new `Appwrite\Platform\Modules\Compute\Validator\VariableKey` on the function, site and project variable create/update endpoints. Requires a C_IDENTIFIER (`[A-Za-z_][A-Za-z0-9_]*`), so every accepted key is also portable to dotenv files, `docker --env` and POSIX shells.
- **Build guard** -- `Deployments::variables()` refuses keys via `VariableKey::isEnvVarName()`, the *Kubernetes* rule (apimachinery `IsEnvVarName`: hyphens and dots allowed, `.`/`..`/`..*` refused). Weaker than the endpoint rule on purpose: legacy keys like `MY-VAR` that deploy fine today keep deploying; only keys the cluster would reject anyway now fail fast -- before job submission, with the actual reason (key name JSON-encoded, value never printed) written to the deployment's `buildLogs` instead of the generic internal error.

The new `variable_invalid_key` error is 400-class, so the VCS webhook fan-out skips just the affected repository and continues to other projects (existing `code < 500` skip branch), the webhook returns 2xx, and the error handler does not publish it to the logging provider.

## Test plan

- `tests/unit/Platform/Modules/Compute/Validator/VariableKeyTest.php` -- both rules, 30 cases including synthetic UTF-16 and trailing-tab shapes matching defects seen in production.
- `tests/unit/Deployment/DeploymentsTest.php` -- drives `payload()`: an invalid key is refused before job submission (verified red with the guard reverted), and `MY-VAR` still builds.
- e2e: invalid keys 400 on project, function and site variable endpoints; invalid update leaves the variable unchanged.
- Existing e2e variable fixtures scanned -- none use a key the new rule refuses.

## Related PRs and Issues

- Companion cloud data cleanup: appwrite-labs/cloud#5363 (merged; reported/trimmed existing offending rows).
- Console fix for the ingestion origin (UTF-16 `.env` import): appwrite/console#3186.

### DIFF
--- a/app/config/errors.php
+++ b/app/config/errors.php
@@ -1281,6 +1281,11 @@ return [
         'description' => 'Secret variables cannot be marked as non-secret. Please re-create the variable if this is your intention.',
         'code' => 400,
     ],
+    Exception::VARIABLE_INVALID_KEY => [
+        'name' => Exception::VARIABLE_INVALID_KEY,
+        'description' => 'Variable key is not a valid environment variable name. Update or delete the variable, then retry the deployment.',
+        'code' => 400,
+    ],
     Exception::GRAPHQL_NO_QUERY => [
         'name' => Exception::GRAPHQL_NO_QUERY,
         'description' => 'Param "query" is not optional.',
--- a/src/Appwrite/Deployment/Deployments.php
+++ b/src/Appwrite/Deployment/Deployments.php
@@ -4,6 +4,7 @@ namespace Appwrite\Deployment;
 
 use Ahc\Jwt\JWT;
 use Appwrite\Extend\Exception;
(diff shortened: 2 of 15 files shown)


levivannoort:
Superseded by #13362 (same change with sanitized test fixtures).

Issue #13347, opened 2026-08-25
Adding resend and cloudflare as Oauth provider


imtia33:
## Cloudflare
Via Cloudflare Zero Trust → SaaS application → Generic OIDC. Standard OIDC: `/authorization`, `/token`, `/userinfo` endpoints, scopes `openid email profile`, returns `sub`/`email`/`name`.
endpoints are per-tenant + per-client (`https://{team}.cloudflareaccess.com/cdn-cgi/access/sso/oidc/{clientId}/...`), not global. The `OAuth2` base class only has slots for `appId` + `appSecret`, so we'll store the team name inside `appSecret` as a JSON blob (`{"clientSecret":"...","team":"..."}`) — same pattern Microsoft uses for `tenantID`.

## Resend — needs two calls

Via Resend OAuth 2.1 + PKCE . Endpoints: `/oauth/authorize`, `/oauth/token`, `/oauth/revoke`, plus `/oauth/register` for Dynamic Client Registration.

### 🎤 Pitch

Two things need your sign-off:

1: No `/userinfo` — decode the JWT instead?

Resend has no `/userinfo` endpoint and no `id_token`. The only place user identity lives is inside the `access_token` JWT (`typ: at+jwt`, RFC 9068, signed ES256). To get the user ID we'd base64-decode the JWT payload and read `sub` — without verifying the signature.

RFC 9068 says clients shouldn't inspect access tokens. We're ok with it because:
- Token arrives over TLS from `api.resend.com/oauth/token` directly
- We never use it for downstream calls — just read `sub` for identity
- Existing providers (Kick) already do JWT inspection

This part is kind of necessary for resend.

Another Question:

Resend supports public (`token_endpoint_auth_method: "none"`, PKCE only) and confidential (`client_secret_basic` or `client_secret_post`) clients. We want confidential so we can reuse the standard `verifyCredentials()` pattern from the HuggingFace PR #13123  (Basic auth header + invalid code → `invalid_client` means bad creds, `invalid_grant` means good creds).
Confirm?
You can confirm the findings from these ref:

- #13123 (HuggingFace, merged): https://github.com/appwrite/appwrite/pull/13123
- #13255 (email optional, merged): https://github.com/appwrite/appwrite/pull/13255
- Cloudflare SaaS OIDC docs: https://developers.cloudflare.com/cloudflare-one/access-controls/applications/http-apps/saas-apps/generic-oidc-saas/
[...]

PR #13123, opened 2026-08-06
Feat 13067 huggingface oauth provider

arpit10128:
## What does this PR do?

This PR includes the backend implementation required to support Hugging Face authentication, including provider configuration, OAuth2 endpoints, models, and related updates.

## Test Plan

Ran the OAuth2 provider tests and confirmed the implementation works as expected.
Screenshot 2026-08-06 154014

## Related PRs and Issues

- Related to #13067

## Checklist

- [x] Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
- [ ] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs?

PR #13366, opened 2026-08-26
Add Resend OAuth2 adapter

Meldiron:
To create client:

```
curl -X POST 'https://api.resend.com/oauth/register' \
     -H 'Content-Type: application/json' \
     -d '{
  "client_name": "Appwrite QA Matej manual",
  "redirect_uris": [
    "http://localhost/v1/account/sessions/oauth2/callback/resend/myprojectid"
  ],
  "grant_types": ["authorization_code", "refresh_token"],
  "response_types": ["code"],
  "token_endpoint_auth_method": "client_secret_post",
  "scope": "emails:send",
  "client_secret_expires_at": 0
}'
```

CleanShot 2026-08-27 at 15 08 33@2x

CleanShot 2026-08-27 at 15 08 41@2x

CleanShot 2026-08-27 at 15 08 31@2x

CleanShot 2026-08-27 at 15 08 50@2x
CleanShot 2026-08-27 at 15 08 46@2x

---

## What does this PR do?

Adds Resend as an OAuth2 social sign-in provider. Related to #13347 (Resend part).

Resend runs a standard OAuth 2.1 authorization server at `api.resend.com/oauth` ([client guide](https://resend.com/docs/guides/building-a-resend-oauth-client), [server metadata](https://api.resend.com/.well-known/oauth-authorization-server)). Three provider quirks shaped the adapter:

- **PKCE (S256) is mandatory**, so the adapter uses the existing `PKCE` trait exactly like Kick — verifier encrypted into the `state`, restored in `parseState()`.
- **No userinfo endpoint and no id_token.** Identity lives only in the RFC 9068 JWT access token, so `getUserID()` decodes the JWT payload (base64url) and reads `sub`. The token arrives over TLS directly from the token endpoint and is never used to authorize anything on our side, so the signature is not verified. Email and name are not exposed, so `getUserEmail()`/`getUserName()` return empty and `isEmailVerified()` is `false` — supported since email became optional in #13255.
- **Scopes are only `full_access` and `emails:send`**, and omitting scope grants *both*, so the adapter pins `emails:send` as the least-privilege default.

The adapter is a confidential client: `client_secret_post` authentication (client_id + client_secret in the token request body) plus PKCE.

`verifyCredentials()` follows the HuggingFace pattern (#13123) with one deliberate deviation, validated live against the endpoint: Resend validates request shape **before** authenticating the client, so a non-URL `redirect_uri` returns `invalid_request` even for bogus credentials and would mask them. With a well-formed URL, bogus credentials correctly return `invalid_client`, so the probe uses a URL-shaped invalid redirect (commented in code).

### Changes

- `src/Appwrite/Auth/OAuth2/Resend.php` — the adapter
- `app/config/oAuthProviders.php` — `resend` provider entry
- `Project/Http/Project/OAuth2/Resend/Update.php` + `Project/Services/Http.php` — console `updateOAuth2Resend` endpoint
- `Response.php` (`MODEL_OAUTH2_RESEND`), `Response/Model/OAuth2Resend.php`, `app/init/models.php` — response model
- `tests/unit/Auth/OAuth2/PKCETest.php` — Resend added to the PKCE provider matrix

### Notes

- The console app needs an `icon-resend` asset (same follow-up as when Kick landed).
- Accounts created via Resend have an empty email until the user sets one, since Resend exposes no email through OAuth.

## Test Plan

- Unit: full `tests/unit/` suite passes (1079 tests, 16113 assertions); `tests/unit/Auth/OAuth2/` fully clean at 47/47 including 8 new `resend` PKCE data sets.
- Live-probed `https://api.resend.com/oauth/token` to confirm the `verifyCredentials()` error contract (`invalid_client` for unknown credentials, `invalid_request` when the redirect is not a URL).
- Smoke-tested login URL generation in-container: correct authorize URL, S256 challenge, encrypted verifier in state.
- Pint and PHPStan (level 4) clean on all touched files.

## Related PRs and Issues

- #13347
- #13123 (HuggingFace, verifyCredentials pattern)
- #13255 (optional email)


Meldiron (commented):
lgtm

PR #13372, opened 2026-08-27
Appwrite Native Push Broker

ArnabChatterjee20k:
### Summary

This adds an MQTT 5.0 push broker to Appwrite as a self hosted alternative to FCM. A device connects to the broker over MQTT, authenticates with an Appwrite JWT or session, subscribes to a topic, and receives push messages.

Publishing stays a server privilege. Messages are sent through the existing Messaging campaign API, never over MQTT from a client. There is no client PUBLISH handler.

Delivery is QoS 1 with offline replay. If a device is offline when a message is published, the broker stores the message and replays it on the next reconnect, up to a bounded window.

### Where the code lives

- `app/mqtt.php` is the broker process. It runs a Swoole server on port 1883 (internal only). It parses each packet, dispatches it to a handler, runs the keep-alive reaper, and listens on the pub/sub channel for fan out.
- `app/init/mqtt/connection.php` holds connection auth. It registers `authenticator` (turns a CONNECT into an identity) and `authorizer` (the per subscribe block check).
- `src/Appwrite/Mqtt/Handlers/` has one class per packet type: Connect, Auth, Subscribe, Unsubscribe, Puback, Ping, Disconnect.
- `src/Appwrite/Mqtt/Connection.php` is the per connection state, keyed by file descriptor.
- `src/Appwrite/Mqtt/SubscriptionStore.php` is the topic trie that matches subscribers to a topic.
- `src/Appwrite/Mqtt/KeepAlive.php` is the timing wheel that reaps silent connections.
- `src/Appwrite/Messaging/Adapter/Mqtt.php` is the broker adapter. It holds the connection registry and fans a published message out to local subscribers.
- `src/Appwrite/Messaging/Adapter/Push/Appwrite.php` is the push provider. It writes each message to the ledger and hands it to the broker.

### Connect and subscribe flow

```mermaid
sequenceDiagram
    participant D as Device (MQTT client)
    participant B as Broker (app/mqtt.php)
    participant DB as Project database
    participant R as Redis (cursor cache)
    participant W as Messaging worker

    D->>B: CONNECT (auth method, credential, projectId)
    B->>DB: resolve project and user, verify credential
    DB-->>B: user (or empty)
    B-->>D: CONNACK (0x00 accepted, 0x87 rejected)

    D->>B: SUBSCRIBE (topic filter, requested QoS)
    B->>DB: topic exists and its subscribe roles allow the user
    B-->>D: SUBACK (granted QoS per filter, or 0x87)
    B->>R: read the client cursor for the topic
    B->>DB: read ledger rows after the cursor (capped at maxDepth)
    B-->>D: PUBLISH x N (offline replay, DUP set)

    Note over W,B: later, a campaign is published
    W->>DB: append to ledger, increment topic sequence
    W->>B: fan out on the mqtt pub/sub channel
    B-->>D: PUBLISH (live, QoS 1)
    D->>B: PUBACK
    B->>R: advance the cursor
```

### How to publish and subscribe

Publish is a normal server side campaign. It targets a topic and routes through the Appwrite push provider.

```bash
curl -X POST https://<endpoint>/v1/messaging/messages/push \
  -H "content-type: application/json" \
  -H "x-appwrite-project: <projectId>" \
  -H "x-appwrite-key: <serverApiKey>" \
  -d '{
    "messageId": "unique()",
    "topics": ["<topicId>"],
    "title": "Match update",
    "body": "India needs 12 off 6"
  }'
```

Subscribe is an MQTT 5 client. It connects over TLS to the Traefik mqtts entrypoint on 8883 and passes the credential through MQTT 5 enhanced auth. The topic filter is the topic id.

```bash
mosquitto_sub -V 5 -h <endpoint> -p 8883 --cafile <ca> \
  -t "<topicId>" -q 1 \
  -D CONNECT authentication-method appwrite-jwt \
  -D CONNECT authentication-data "<jwt>" \
  -D CONNECT user-property projectId "<projectId>"
```

The credential is a JWT (`authentication-method: appwrite-jwt`, data is the JWT string) or a session (`authentication-method: appwrite-session`, data is the session store blob, the same value a client holds as its session cookie).

### What each packet handles

- **CONNECT.** Parses the protocol level, keep-alive, clean-start flag, auth method and data, and the projectId user property. It resolves and verifies the user, rate limits the user, assigns a client id, and replies CONNACK.
- **SUBSCRIBE.** Checks the account is not blocked, the topic exists, and the topic's subscribe roles include the user. It grants a QoS that is the requested value capped by the topic's configured QoS, stores the subscription in the trie, replies SUBACK, then replays any offline messages.
- **PUBLISH (server to broker, over pub/sub).** Fans the message out to local subscribers at `min(publish QoS, granted QoS)` and tracks each QoS 1 delivery until its PUBACK.
- **PUBACK.** Advances the client cursor to the highest contiguous acknowledged sequence, so a later ack cannot skip an earlier unacked message.
- **PINGREQ.** Replies PINGRESP. Every inbound packet, including this one, refreshes the keep-alive deadline.
- **AUTH.** Refreshes the credential for the same identity. It cannot switch the user or project.
- **DISCONNECT and transport close.** Runs one cleanup path that removes the subscriptions, drops the connection from the keep-alive wheel, and balances the active-connection gauge.

### Connection lifecycle

A connection is created on the first packet from a file descriptor. CONNECT authenticates it and marks it active. Once active, the broker schedules it in the keep-alive wheel.

The keep-alive wheel is a hashed timing wheel. Each connection sits in a bucket for its deadline second. A single timer runs every 20 seconds, drains the buckets that came due, and closes any client that has been silent for longer than `keepAlive * 1.5`. Client activity only moves a timestamp on the connection. It does not touch the wheel, so a busy client does not churn it.

Close is centralized. Whether the client sends DISCONNECT, the keep-alive reaper fires, or the socket drops, Swoole's onClose calls one method that unsubscribes the file descriptor, removes it from the wheel, and decrements the active gauge.

The whole packet lifecycle in onReceive is wrapped. A malformed packet or a handler failure is logged and closes only that connection. It never propagates and takes down the Swoole worker.

### Offline replay: ledger, cursor, trie

**Ledger.** Every published message is appended to the `appwritePushLedger` collection with a per topic sequence number. The sequence comes from an atomic increment on `topics.sequence`. The increment and the insert run in one transaction, opened by locking the topic row, so two concurrent publishes cannot open a gap and a retry is idempotent on `(messageId, topic)`.

**Cursor.** Each client has a cursor per topic in Redis, keyed by project, user, and client id. The cursor is the highest sequence the client has acknowledged. A PUBACK advances it. Nothing else does.

**Replay.** On SUBSCRIBE, the broker reads the ledger from `cursor + 1` up to the current topic sequence, capped at a replay depth, and re-sends those messages with the DUP flag set. The depth comes from the user's org plan (`getPlanForUser`, self hosted default of 5 from `_APP_MQTT_REPLAY_DEPTH`). Replay is a QoS 1 guarantee, so a QoS 0 subscription is live only and is not replayed. A clean-start connection purges its cursor first.

**Retention.** The maintenance worker deletes ledger rows older than 7 days, which matches the maximum replay window.

**Trie.** `SubscriptionStore` is a tree keyed by project and topic segments. It resolves the set of file descriptors subscribed to a topic, each with its granted QoS, which is what the fan-out path iterates.

### Security

- The broker publishes no plaintext port. It is reachable from outside only through the Traefik mqtts entrypoint on 8883 with TLS, the same way realtime is reachable only through wss. Inside the network it listens on 1883.
- Publishing is server only. There is no client PUBLISH path.
- SUBSCRIBE enforces the topic's subscribe roles, so a user cannot subscribe to a topic they are not allowed on.
- CONNECT is rate limited per user (128 per minute), gated by `_APP_OPTIONS_ABUSE`. It is keyed on userId for now because the infra does not surface the client IP to the broker.
- A reauth cannot switch the user or project. A mismatch drops the connection.

### Telemetry

New instruments cover auth duration, connection duration, message size, bytes in and out, reauth outcomes, and PUBACK counts, alongside the existing connection and delivery counters.

### Tests and CI

- Unit tests cover the timing wheel, the connection state, and the subscription trie.
- End to end tests cover CONNECT auth (JWT and session, rejection, blocked user), the SUBSCRIBE ACL and QoS capping, campaign fan out, offline replay (including non contiguous acks and the depth cap), clean start, and keep-alive (a silent client is reaped, an active one survives).
- A dedicated `Tests / E2E / MQTT` CI job boots the broker and runs the `mqtt` group. The general service lanes exclude that group so the broker dependent tests only run where the broker is guaranteed up.

### Connectivity (broker topology)

Clients reach the broker two ways, both TLS-only and terminated at Traefik (the broker itself stays plaintext on the internal Docker network, mirroring how realtime's `wss` is terminated at the edge). The transport is addressed by host + dedicated port — there is no URL path — so a single push endpoint (e.g. `push.cloud.appwrite.io`) serves both native MQTT and MQTT-over-WebSocket:

| Transport | Entrypoint / port | Router | → broker |
|-----------|-------------------|--------|----------|
| `mqtts://` (native MQTT) | `appwrite_mqtt` `:8883` | TCP, `HostSNI(*)`, tls | `:1883` |
| `wss://` (MQTT over WebSocket) | `appwrite_mqtt_ws` `:8084` | HTTP catch-all on that port, tls | `:8083` |

- Native app/device clients use `mqtts://<host>:8883`; browsers use `wss://<host>:8084` (MQTT 5.0 framed in WebSocket binary frames).
- The reusable JWT/session credential in the CONNECT frame never crosses the network in the clear — neither transport is exposed over a plaintext port.
- Ports are overridable via `_APP_MQTT_PORT` (8883) and `_APP_MQTT_WSS_PORT` (8084); the broker's internal WebSocket listener is `_APP_MQTT_WS_PORT` (8083).


src/Utopia/Mqtt/Handlers/Publish.php
```diff
@@ -0,0 +1,54 @@
...
+            $packetId = substr($body, $offset, 2);
+            $offset += 2;
+        }
+        $offset = Packet::skipProperties($body, $offset, $connection->protocol);
+        $payload = substr($body, $offset);
+
+        Span::add('mqtt.topic', $topic);
+        Span::add('mqtt.subscribers', count($mqtt->getSubscribers($connection->projectId, $topic)));
+
+        $mqtt->send($connection->projectId, [], [], [$topic], [], ['payload' => $payload, 'qos' => $qos]);
```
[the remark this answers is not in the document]
ArnabChatterjee20k: It is present currently to test. It will be done internally via the messaging workers with secrets

app/mqtt.php
```diff
@@ -0,0 +1,356 @@
...
+        return false;
+    }
+
+    $dbForProject = getProjectDB($project);
+    $dbForProject->setAuthorization($authorization);
+    $user = $authorization->skip(fn () => $dbForProject->getDocument('users', $userId));
+
+    // status: true = enabled, false = blocked.
+    return !$user->isEmpty() && $user->getAttribute('status', true) !== false;
+};
```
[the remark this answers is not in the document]
ArnabChatterjee20k: @greptile this is for testing purpose as a poc. The intended flow makes sure to check the topic documents whether it is viable to subscirbe to the topic or not or even the topic belongs to the project or not

src/Appwrite/Messaging/Adapter/Push/Appwrite.php
```diff
@@ -0,0 +1,176 @@
...
+        if (!$existing->isEmpty()) {
+            return (int) $existing->getAttribute('sequence');
+        }
+
+        $sequence = (int) $authorization->skip(
+            fn () => $this->dbForProject->increaseDocumentAttribute('topics', $topic, 'sequence', 1)
+        )->getAttribute('sequence');
+
+        $authorization->skip(
+            fn () => $this->dbForProject->createDocument('appwrite_push_ledger', new Document([
```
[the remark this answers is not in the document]
ArnabChatterjee20k: Fixed in 5adb6585. The increment + ledger insert now run inside a single \`dbForProject->withTransaction(...)\`, so a failed insert rolls the increment back — the topic sequence can no longer advance without a matching ledger row. For the concurrent case, the \`_key_message_topic\` unique index serialises the two attempts: the loser's insert throws \`Duplicate\`, its increment rolls back, and it now resolves to the winner's sequence instead of surfacing a failure or leaving a gap.

src/Appwrite/Messaging/Adapter/Push/Appwrite.php
```diff
@@ -0,0 +1,158 @@
...
+    {
+        $payload = $this->buildPayload($message);
+        $response = new Response($this->getType());
+
+        foreach ($message->getTo() as $topic) {
+            try {
+                $this->persist($topic, $payload);
+                $this->broker->send(
+                    $this->projectId,
+                    [],
```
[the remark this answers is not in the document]
ArnabChatterjee20k: Addressed. persist() is idempotent on (messageId, topic): a findOne fast-path returns the existing sequence, the \`_key_message_topic\` unique index blocks a duplicate row, and a losing concurrent insert (\`Duplicate\`) re-reads the winner's row rather than incrementing again. A publish failure after persist is retried by the worker and reuses the existing sequence — no second increment, no second row.

app/mqtt.php
```diff
@@ -0,0 +1,427 @@
...
+                $pubsub->subscribe(['mqtt'], function (mixed $redis, string $channel, string $payload) use ($server, $mqtt): void {
+                    $event = json_decode($payload, true);
+                    if (!\is_array($event)) {
+                        return;
+                    }
+
+                    $projectId = (string) ($event['project'] ?? '');
+                    $topic = (string) ($event['topic'] ?? '');
+                    $qos = (int) ($event['qos'] ?? 0);
+                    $sequence = (int) ($event['payload']['sequence'] ?? 0);
```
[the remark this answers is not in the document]
ArnabChatterjee20k: Already resolved on the current head. The fan-out envelope now carries \`sequence\` and the consume loop reads \`(int) (\$event['sequence'] ?? 0)\` rather than off the base64 payload. The CONNECT span's project.id / user.id / clean_start are also set *after* \`\$dispatcher->dispatch()\`, so they reflect the values the CONNECT handler populated, not the pre-dispatch defaults.

src/Appwrite/Platform/Modules/Messaging/Http/Topics/Update.php
```diff
@@ -75,6 +78,14 @@ public function action(string $topicId, ?string $name, ?array $subscribe, Event
             $topic->setAttribute('subscribe', $subscribe);
         }
 
+        if (!\is_null($qos)) {
+            $topic->setAttribute('qos', $qos);
+        }
+
+        if (!\is_null($expiry)) {
+            $topic->setAttribute('expiry', $expiry);
```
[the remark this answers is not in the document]
ArnabChatterjee20k: Fixed in 0c464a71. An omitted param and an explicit null both arrive as null at the action, so both this topic endpoint and the Appwrite provider update now key on the request body: \`array_key_exists('qos', \$request->getParams())\`. Sending an explicit null therefore resets qos/expiry to the documented default (subscriber-chosen QoS, no expiry) instead of silently retaining the old value.

src/Utopia/Mqtt/Handlers/Subscribe.php
```diff
@@ -0,0 +1,126 @@
...
+        $consoleDatabase = getConsoleDB();
+        $project = $consoleDatabase->getAuthorization()->skip(fn () => $consoleDatabase->getDocument('projects', $connection->projectId));
+        $projectDB = getProjectDB($project);
+
+        $cache = getCache();
+        foreach ($topics as $topic) {
+            $lastMessageCursorKey =  'mqtt:cursor:' . $connection->projectId . ':' . $connection->identity['userId'] . ':' . $connection->getClientId() . ':' . $topic;
+            $payload = $cache->load($lastMessageCursorKey, 3600);
+
+            $topicDocument = $projectDB->getAuthorization()->skip(fn () => $projectDB->findOne('messages', [Query::containsAny('topics', [$topic]), Query::orderDesc('$sequence'), Query::orderDesc('data')]));
```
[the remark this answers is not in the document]
ArnabChatterjee20k: Resolved on the current head — replay no longer uses findOne. It now runs a windowed \`find('appwrite_push_ledger', ...)\` from \`max(cursor+1, tail-maxDepth+1)\` ordered ascending, so every retained message in the window is delivered and tracked individually, and the cursor only advances per-PUBACK.

src/Appwrite/Mqtt/Handlers/Subscribe.php
```diff
@@ -0,0 +1,159 @@
...
+                continue;
+            }
+
+            $start = max($from + 1, $tail - $maxDepth + 1);
+            $messages = $projectDB->getAuthorization()->skip(fn () => $projectDB->find('appwrite_push_ledger', [
+                Query::equal('topic', [$topic]),
+                Query::greaterThanEqual('sequence', $start),
+                Query::orderAsc('sequence'),
+                Query::limit($maxDepth),
+            ]));
```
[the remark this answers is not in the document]
ArnabChatterjee20k: This is intentional, not a drop. Offline replay is deliberately bounded to the most recent \`maxDepth\` (5) messages per topic — a reconnecting client that is further behind resumes from \`tail-maxDepth+1\`. Anything older than the window is outside the retained backlog (same bound the ledger retention/expiry enforces); replaying an unbounded backlog on every reconnect is what we're avoiding. The window size is a plan/config knob (TODO to track the plan).

tests/unit/Appwrite/Messaging/Adapter/MqttTest.php
```diff
@@ -0,0 +1,147 @@
...
+    public function testOpenCreatesAndReusesConnection(): void
+    {
+        $mqtt = $this->adapter();
+
+        $connection = $mqtt->open(7);
+
+        $this->assertInstanceOf(Connection::class, $connection);
+        $this->assertSame(7, $connection->fd);
+        $this->assertSame($connection, $mqtt->open(7), 'open() must return the same instance for the same fd');
+        $this->assertArrayHasKey(7, $mqtt->connections);
```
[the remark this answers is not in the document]
ArnabChatterjee20k: These assert the adapter's public contract, not private internals (no reflection). \`connections\` is a documented public property the onReceive loop reads every packet, and \`open()\` must return the *same* Connection per fd so in-flight/PUBACK state survives across packets — \`assertSame(\$c, \$mqtt->open(7))\` and the post-close \`assertNotSame\` pin exactly that observable lifecycle. \`getSubscribers()\`/\`hasSubscriber()\` are the fan-out API. Happy to drop the raw \`connections\` array-key checks in favour of only the open()/close() identity assertions if you prefer a thinner surface.

tests/unit/Appwrite/Mqtt/ConnectionTest.php
```diff
@@ -0,0 +1,72 @@
...
+    public function testEmptyIdDerivesAccountLevelAnchor(): void
+    {
+        $connection = new Connection(1);
+        $connection->identity = ['userId' => 'user-1'];
+        $connection->projectId = 'project-1';
+
+        $connection->setClientId('');
+
+        $this->assertSame('custom_project-1_user-1', $connection->getClientId());
+    }
```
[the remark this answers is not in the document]
ArnabChatterjee20k: The exact string here is the contract, not an implementation detail: the derived anchor is the durable cursor cache key (\`mqtt:cursor:<project>:<user>:<clientId>\`), so an empty client id must map deterministically to the same anchor across reconnects or offline cursors orphan. Asserting the literal guards that stability. I can restructure it to assert the two properties that actually matter — determinism (same project+user ⇒ same anchor) and derivation from both ids — rather than the literal, if that reads better to the team.

tests/unit/Appwrite/Mqtt/SubscriptionStoreTest.php
```diff
@@ -0,0 +1,171 @@
...
+        $connection = $this->store->getConnection(9);
+
+        $this->assertSame(self::PROJECT, $connection['projectId']);
+        $this->assertSame('user-1', $connection['userId']);
+        $this->assertSame(1, $connection['subs']['test/hello']);
+    }
+
+    public function testGetConnectionIsNullForAnUnknownFd(): void
+    {
+        $this->assertNull($this->store->getConnection(404));
```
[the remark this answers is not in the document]
ArnabChatterjee20k: \`getConnection()\` is public API, and its return shape is a consumed contract — the PUBACK handler reads \`getConnection(fd)['userId']\` / \`['projectId']\` for its auth check and \`getSubscribers\` is fan-out. Asserting the returned record is testing that observable output, not reaching into a private field (no reflection). If the concern is the \`subs\` map shape specifically, I can assert delivery behaviour through getSubscribers instead of the map literal.

tests/unit/Appwrite/Messaging/Adapter/Push/AppwriteTest.php
```diff
@@ -0,0 +1,320 @@
...
+        $this->assertSame('internal-1', $row->getAttribute('messageInternalId'));
+
+        // The fan-out reached the broker with the same project, topic, payload and QoS.
+        $this->assertCount(1, $broker->published);
+        $publish = $broker->published[0];
+        $this->assertSame(self::PROJECT_ID, $publish['projectId']);
+        $this->assertSame(['topic-1'], $publish['channels']);
+        $this->assertSame(1, $publish['options']['qos']);
+        // The ledger stores exactly what was published.
+        $this->assertSame($row->getAttribute('data'), $publish['options']['payload']);
```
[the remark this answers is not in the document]
ArnabChatterjee20k: The two things asserted are the adapter's observable outputs: the durable ledger row (sequence/messageId/topic — the exact data reconnect replay depends on) and the fan-out publish the broker receives (topic/payload/sequence). Those are the contract of process(), verified through a named fake broker rather than reflection or call-order. Asserting the persisted sequence advances and each topic is independent is behaviour, not internals. Open to trimming to sparse field checks if the full-row asserts feel heavy.

src/Appwrite/Messaging/Adapter/Push/Appwrite.php
```diff
@@ -0,0 +1,198 @@
+<?php
```
ArnabChatterjee20k, line 1: having this provider inside the Appwrite repo itself cause the publish path is via the request to the message publish which tightly integrates inside the appwrite repo only. And very different from other push  providers

.github/workflows/ci.yml
```diff
@@ -419,6 +419,71 @@ jobs:
...
+        timeout-minutes: 5
+        run: docker compose up -d --quiet-pull --wait
+
+      - name: Wait for MQTT broker
+        timeout-minutes: 3
+        run: |
+          while ! docker logs appwrite-mqtt 2>&1 | grep -q "MQTT broker started"; do
+            echo "Waiting for the MQTT broker to come online"
+            sleep 1
+          done
```
loks0n: lets use docker readiness checks instead of this script

.github/workflows/ci.yml
```diff
@@ -576,7 +641,7 @@ jobs:
             -e _APP_E2E_RESPONSE_FORMAT="${{ github.event.inputs.response_format }}" \
             -e _TESTS_OAUTH2_GITHUB_CLIENT_ID="${{ secrets.TESTS_OAUTH2_GITHUB_CLIENT_ID }}" \
             -e _TESTS_OAUTH2_GITHUB_CLIENT_SECRET="${{ secrets.TESTS_OAUTH2_GITHUB_CLIENT_SECRET }}" \
-            appwrite vendor/bin/paratest --processes "$PARATEST_PROCESSES" $FUNCTIONAL_FLAG "$SERVICE_PATH" --exclude-group abuseEnabled --exclude-group screenshots --exclude-group antivirus --log-junit tests/e2e/Services/${{ matrix.service.name }}/junit.xml
+            appwrite vendor/bin/paratest --processes "$PARATEST_PROCESSES" $FUNCTIONAL_FLAG "$SERVICE_PATH" --exclude-group abuseEnabled --exclude-group screenshots --exclude-group antivirus --exclude-group mqtt --log-junit tests/e2e/Services/${{ matrix.service.name }}/junit.xml
```
loks0n: why use an exclude group?

if we keep adding these it will keep growing and growing. better to invert this and use a proper phpunit test suite in phpunit.xml

app/config/collections/common.php
```diff
@@ -2212,6 +2212,94 @@
         ],
     ],
 
+    'appwritePushLedger' => [
```
loks0n, line 2129: very long name, why not 'notifications'?

app/config/collections/common.php
```diff
@@ -2294,6 +2382,44 @@
...
+                'size' => 0,
+                'signed' => true,
+                'required' => false,
+                'default' => 0,
+                'array' => false,
+                'filters' => [],
+            ],
+            // null lets the subscriber choose.
+            [
+                '$id' => ID::custom('qos'),
```
loks0n, line 2316: what does this mean? quality of service? why acronym?

src/Appwrite/Mqtt/Metrics.php
```diff
@@ -0,0 +1,51 @@
...
+use Utopia\Telemetry\UpDownCounter;
+
+/**
+ * The broker's telemetry instruments, created once from a telemetry adapter and
+ * shared with every packet handler. Handlers record connection and delivery
+ * outcomes (accepted/rejected, granted/denied, delivered/dropped) that only they
+ * observe. A no-op telemetry adapter yields no-op instruments, so the broker runs
+ * untelemetered (e.g. raw protocol testing).
+ */
+class Metrics
```
loks0n: this doesn't deserve to be its own class imo

src/Appwrite/Mqtt/Handlers/Puback.php
```diff
@@ -0,0 +1,73 @@
...
+    {
+        $body = $packet->body;
+
+        if (\strlen($body) < 2) {
+            return;
+        }
+
+        $connectionMetdata = $mqtt->getConnection($connection->fd);
+        if ($connectionMetdata['userId'] !== $connection->identity['userId'] || $connectionMetdata['projectId'] !== $connection->projectId) {
+            $reply(V5::disconnect(V5::REASON_NOT_AUTHORIZED), true);
```
loks0n: passing around an untyped callable feels hacky, lets give reply a proper shape.

e.g. $response->send() or something like that

src/Appwrite/Messaging/Adapter/Mqtt.php
```diff
@@ -0,0 +1,190 @@
...
+     */
+    private function subscriptions(): SubscriptionStore
+    {
+        return $this->subscriptions ??= new SubscriptionStore();
+    }
+
+    private function getPubSubPool(): PubSubPool
+    {
+        if ($this->pubSubPool === null) {
+            global $register;
```
loks0n: lets not use globals like this, use our DI system instead to get a pubsub isntance

src/Appwrite/Messaging/Adapter/Mqtt.php
```diff
@@ -0,0 +1,190 @@
...
+            $this->pubSubPool = new PubSubPool($register->get('pools')->get('pubsub'));
+        }
+
+        return $this->pubSubPool;
```
loks0n: lets not manage the concept of pools in this class, just pubsub

src/Appwrite/Platform/Modules/Messaging/Http/Topics/Update.php
```diff
@@ -53,13 +55,16 @@ public function __construct()
             ->param('topicId', '', fn (Database $dbForProject) => new UID($dbForProject->getAdapter()->getMaxUIDLength()), 'Topic ID.', false, ['dbForProject'])
             ->param('name', null, new Nullable(new Text(128)), 'Topic Name.', true)
             ->param('subscribe', null, new Nullable(new Roles(APP_LIMIT_ARRAY_PARAMS_SIZE)), 'An array of role strings with subscribe permission. By default all users are granted with any subscribe permission. [learn more about roles](https://appwrite.io/docs/permissions#permission-roles). Maximum of ' . APP_LIMIT_ARRAY_PARAMS_SIZE . ' roles are allowed, each 64 characters long.', true)
+            ->param('qos', null, new Nullable(new Range(0, 1)), 'QoS for MQTT delivery on this topic (0 or 1). Null lets the subscriber choose.', true)
```
loks0n: what does 0 vs 1 even mean?

src/Appwrite/Platform/Workers/Deletes.php
```diff
@@ -354,6 +354,7 @@ public function action(
                 $this->deleteExpiredChallenges($project, $getProjectDB);
                 $this->deleteExpiredTransactions($project, $getProjectDB);
                 $this->deleteExpiredPresences($project, $getProjectDB, $publisherForUsage);
+                $this->deleteExpiredPushLedger($project, $getProjectDB);
```
loks0n, line 348: what is a 'push ledger'?

src/Appwrite/Mqtt/SubscriptionStore.php
```diff
@@ -0,0 +1,191 @@
+<?php
+
+namespace Appwrite\Mqtt;
```
loks0n: this looks like it belongs in utopia

src/Appwrite/Mqtt/KeepAlive.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+namespace Appwrite\Mqtt;
+
+/**
+ * A hashed timing wheel for MQTT keep-alive expiry.
+ * Better than a priority queue as it is grouping the fds in slots of their expiry. Getting all expired and non expired fds in one go
+ *
+ */
+class KeepAlive
```
loks0n: this looks like in belongs in utopia

src/Appwrite/Mqtt/Dispatcher.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+namespace Appwrite\Mqtt;
```
loks0n, line 3: this looks like it belongs in utopia

src/Appwrite/Mqtt/Connection.php
```diff
@@ -0,0 +1,143 @@
+<?php
+
+namespace Appwrite\Mqtt;
```
loks0n: this looks like it belongs in utopia

src/Appwrite/Mqtt/Handlers/Disconnect.php
```diff
@@ -0,0 +1,27 @@
...
+            ->inject('reply')
+            ->callback($this->action(...));
+    }
+
+    /**
+     * @param callable(string, bool): void $reply writes a packet back to this connection (and optionally closes it)
+     */
+    public function action(callable $reply): void
+    {
+        $reply('', true);
```
loks0n: lets give 'reply' a real interface,

app/mqtt.php
```diff
@@ -0,0 +1,412 @@
...
+use Utopia\Pools\Group;
+use Utopia\Registry\Registry;
+use Utopia\Span\Span;
+use Utopia\System\System;
+
+require_once __DIR__ . '/init.php';
+
+if (System::getEnv('_APP_EDITION', 'self-hosted') === 'self-hosted') {
+    require_once __DIR__ . '/init/span.php';
+}
```
loks0n: remove this, use setExporters instead.  mutable patterns are harder to follow

app/mqtt.php
```diff
@@ -0,0 +1,412 @@
...
+
+/** @var Registry $register */
+$register = $GLOBALS['register'] ?? throw new \RuntimeException('Registry not initialized');
+
+$registerMqttConnectionResources ??= require __DIR__ . '/init/mqtt/connection.php';
```
loks0n: registerConnectionResources

app/mqtt.php
```diff
@@ -0,0 +1,412 @@
...
+$mqtt = new Mqtt($telemetry);
+
+$dispatcher = (new Dispatcher())
+    ->addHandler(new ConnectHandler())
+    ->addHandler(new SubscribeHandler())
+    ->addHandler(new UnsubscribeHandler())
+    ->addHandler(new PubackHandler())
+    ->addHandler(new AuthHandler())
+    ->addHandler(new PingHandler())
+    ->addHandler(new DisconnectHandler());
```
loks0n: we should be allergic to code that looks like this

mutability is a footgun in coroutine env

app/config/collections/common.php
ArnabChatterjee20k: Calling it appwritePushLedger, as its acting as a ledger for the push messages. As messages would be replayed in qos1.
Not using notifications as we already have the name notifications for console http endpoint

app/config/collections/common.php
ArnabChatterjee20k: qos means quality of service but mqtt specs itself mention it as qos everywhere. Even other sdks as well mention it as qos. So using it as a standard everywhere

src/Appwrite/Platform/Modules/Messaging/Http/Topics/Update.php
ArnabChatterjee20k: Its qos level. 
0 means fire and forget
1 means offline persistence and replayability when client reconnects if not received.

src/Appwrite/Platform/Workers/Deletes.php
ArnabChatterjee20k: Its for the qos1. Calling it ledger as we are adding the message as form of a ledger with no updates just for persistence and replayability. 
Since max replayability is of last 1 week, so maintainance worker would delete last 1week expired messages

src/Appwrite/Mqtt/Dispatcher.php
ArnabChatterjee20k: Keeping it here as its a normal dispatcher instead of using the switch case statements everytime for new functionality. Made it non mutable. 
Moved the other states Connection, SubscriptionStore, Keepalive to the utopia.

app/mqtt.php
```diff
@@ -0,0 +1,404 @@
...
+                $server->close($fd);
+            } else {
+                $connection->wheelSlot = $mqtt->keepAlive->schedule($fd, $connection->expiresAt);
+            }
+        }
+    });
+
+    go(function () use ($server, $mqtt, $register): void {
+        $attempts = 0;
+        while ($attempts < 300) {
```
loks0n: Why 300 attempts? That seems like a crazy number

app/mqtt.php
```diff
@@ -0,0 +1,404 @@
...
+            if ($connection === null || $connection->keepAlive <= 0) {
+                continue;
+            }
+
+            if ($connection->expiresAt <= $now) {
+                $server->close($fd);
+            } else {
+                $connection->wheelSlot = $mqtt->keepAlive->schedule($fd, $connection->expiresAt);
+            }
+        }
```
ArnabChatterjee20k: put this to the utopia-php/mqtt via interface

app/mqtt.php
```diff
@@ -0,0 +1,243 @@
...
+
+global $container;
+
+if (!$container->has('pools')) {
+    $container->set('pools', function ($register) {
+        return $register->get('pools');
+    }, ['register']);
+}
+
+$container->set('getCache', fn () => function () use ($register): Cache {
```
ArnabChatterjee20k, line 45: got the suggestion that we need them per coroutine context thats why not doing DI directly rather injecting them in the child container

src/Appwrite/Mqtt/Handler.php
```diff
@@ -0,0 +1,343 @@
...
+        return $unsuback;
+    }
+
+    /**
+     * Appwrite is server -> client push: clients never PUBLISH, so an inbound one is ignored.
+     * Server-initiated fan-out goes through deliver() instead.
+     *
+     * @param iterable<array{0: Connection, 1: int}> $subscribers
+     */
+    public function onPublish(Publish $publish, Connection $connection, iterable $subscribers): void
```
ArnabChatterjee20k, line 198: not needed the publish as its getting done internally via the workers here

.github/workflows/ci.yml
```diff
@@ -523,6 +523,63 @@ jobs:
           echo "=== Appwrite Logs ==="
           docker compose logs --timestamps
 
+  e2e_mqtt:
```
loks0n: possible to make it just an additional entry in the service matrix?

app/init/mqtt/connection.php
```diff
@@ -0,0 +1,135 @@
...
+use Utopia\Auth\Hashes\Sha;
+use Utopia\Auth\Proofs\Token;
+use Utopia\Auth\Store;
+use Utopia\Config\Config;
+use Utopia\Database\Document;
+use Utopia\Database\Validator\Authorization;
+use Utopia\DI\Container;
+use Utopia\System\System;
+
+return function (Container $container): void {
```
loks0n, line 15: these run per connection, but I don't see the connection actually being used to vary the result? why not make them singletons?

app/mqtt.php
loks0n: why? cache is pooled no?

src/Appwrite/Messaging/Adapter/Mqtt.php
```diff
@@ -0,0 +1,90 @@
...
+            $this->messagesPublished->add(1, ['qos' => $qos]);
+            $this->pubsub->publish(self::CHANNEL, (string) json_encode([
+                'project' => $projectId,
+                'topic' => $topic,
+                'qos' => $qos,
+                'sequence' => $sequence,
+                'payload' => base64_encode($message),
+            ]));
+        }
+    }
```
loks0n, line 73: not a blocker

does feel like something of this could live in utopia mqtt, pubsub/state adapter (redis/inmemory) + sender

src/Appwrite/Messaging/Adapter/Mqtt.php
```diff
@@ -0,0 +1,90 @@
...
+     * @param array<int, string> $channels
+     * @param array<int, mixed> $queryGroup
+     */
+    public function subscribe(string $projectId, mixed $identifier, string $subscriptionId, array $roles, array $channels, array $queryGroup = []): void
+    {
+    }
+
+    public function unsubscribe(mixed $identifier): void
+    {
+    }
```
loks0n, line 89: this is a big signal that the design is trying to fit square in a round role, empty method overrides.

src/Appwrite/Mqtt/Handler.php
```diff
@@ -0,0 +1,348 @@
...
+use Utopia\Mqtt\Packet\Publish;
+use Utopia\Mqtt\Packet\Suback;
+use Utopia\Mqtt\Packet\Subscribe;
+use Utopia\Mqtt\Packet\Unsuback;
+use Utopia\Mqtt\Packet\Unsubscribe;
+use Utopia\Mqtt\Server;
+use Utopia\Span\Span;
+use Utopia\System\System;
+
+class Handler implements MqttHandler
```
loks0n, line 32: This class looks so much better 👍 

To go futher, the next step would be, are any parts are common to all MQTT server impelemenations? if they are move to utopia

src/Appwrite/Mqtt/Handler.php
loks0n: maybe it'd be more elegant to give it a optional implemenation then?

src/Appwrite/Mqtt/Handler.php
```diff
@@ -0,0 +1,348 @@
...
+     * Replay the offline backlog for freshly granted QoS 1 topics: resume each topic's cursor,
+     * then re-deliver stored messages via the connection (DUP flagged), tagging each with its
+     * durable sequence so a later PUBACK maps back.
+     *
+     * @param array<string, array{0: Document, 1: int}> $grantedTopics filter => [document, grantedQos]
+     */
+    private function replayBacklog(Connection $connection, Database $projectDB, array $grantedTopics): void
+    {
+        $getCache = $this->container->get('getCache');
+        $getPlanForUser = $this->container->get('getPlanForUser');
```
loks0n: this di is a little bit awkward tbh

I'd aim for a structure that allows you to get the actually things you need

`cache` and `planForUser`

directly as parametesrs

app/init/mqtt/connection.php
ArnabChatterjee20k: The result does vary but per coroutine, not per argument. That variation is hidden inside Coroutine::getContext(), not in the signature
If we have a singleton then the coroutines would start sharing them can break tenancy similar to what happened in the realtime which lead to adding of the coroutine context.

getProjectDB($project) does vary by argument
getPlanForUser does make sense to be in the connection

app/mqtt.php
ArnabChatterjee20k: Because of the same reason as othere. This is per coroutine and we are saving in the coroutine context

src/Appwrite/Messaging/Adapter/Mqtt.php
ArnabChatterjee20k: This is basically a message publisher. And There is nothing to subscribe unlike the FCM/APNS adapter here and just getting used for publishing in the messaging worker which follows a strict interface

app/mqtt.php
ArnabChatterjee20k: Its taken from the realtime. Reducing it a bit. Will adjust this in future after few testings around

src/Appwrite/Mqtt/Handler.php
ArnabChatterjee20k: Not optional cause 
No optional method possible in the php interfaces. Then we have to either create a default onPublish method or create a separate PublishHandler to have the onPublish. Feeling a bit confusing

publish is a valid spec in mqtt and should be added but we are blocking it and taking no action to block for user. Messaging worker is doing the publish via the pubsub.

loks0n (approved):

PR #13376, opened 2026-08-27
fix(specs): emit valid OpenAPI item types for array parameters

ChiragAgg5k:
`ArrayList` copied its inner validator's `gettype()` name straight into `items.type`, so a list of floats emitted `"type": "double"` — a format, not an OpenAPI type. `Validator::TYPE_MIXED` had the same problem with no OpenAPI equivalent at all.

Any spec carrying one of those fails schema parsing, which blocks SDK generation for the whole platform. It surfaced on the Cloud manager spec (`POST /manager/databases/metrics` takes `ArrayList(FloatValidator)` parameters), where `php app/cli.php sdks` aborted with `Unsupported schema type 'double'`.

Floats now emit `{"type": "number", "format": "double"}` and mixed emits no type, matching how scalar float parameters were already handled a few cases below.

## Test

`tests/unit/SDK/Specification/FormatTest.php::testArrayListItemTypesAreValidOpenApiTypes` asserts the emitted `items` schema for float, integer and string lists. Seen red on `main` (`'type' => 'double'`) and green with the fix.

### DIFF
--- a/src/Appwrite/SDK/Specification/Format.php
+++ b/src/Appwrite/SDK/Specification/Format.php
@@ -432,7 +432,7 @@ abstract class Format
             return \in_array($default, $schema['enum'], true);
         }
 
-        if (isset($schema['items']['enum'])) {
+        if (\is_array($schema['items'] ?? null) && isset($schema['items']['enum'])) {
             return \is_array($default) && empty(\array_diff($default, $schema['items']['enum']));
         }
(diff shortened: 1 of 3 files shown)


ChiragAgg5k:
@greptile review

ArnabChatterjee20k (approved):

PR #13377, opened 2026-08-27
fix: make DocumentsDB and VectorsDB connections configurable

ChiragAgg5k:
Already on `2.0.x` as `1f1d72e50b`; this carries it to `main`.

## Problem

DocumentsDB and VectorsDB are separate connection pools locked to one engine each (`app/init/registers.php:199-210`):

| Pool | Accepted engines |
|---|---|
| `console` / `database` | mariadb, mongodb, mysql, postgresql |
| `documentsdb` | **`['mongodb']`** only |
| `vectorsdb` | **`['postgresql']`** only |

That part is by design — DocumentsDB is schemaless, VectorsDB needs pgvector. The problem is that the connection details were effectively hardcoded and could not be changed:

1. **`docker-compose.yml` never listed the `_APP_DB_*_DOCUMENTSDB` variables**, and listed no `_APP_CONNECTIONS_DATABASE_*` variables at all (the only `_APP_CONNECTIONS_*` present was `_APP_CONNECTIONS_DB_USAGE`). Compose's list form of `environment:` forwards only the variables it names, so setting these in `.env` never reached the container and both pools always fell back to `mongodb:27017` / `postgresql:5432`.
2. **Neither family appeared in `app/config/variables.php`**, so `install` never wrote them to `.env` either.
3. **`registers.php` read `_APP_DB_USER` / `_APP_DB_PASS` / `_APP_DB_SCHEMA` for both pools**, so the per-engine credential variables compose *did* forward for VectorsDB (`_APP_DB_USER_VECTORSDB` and friends) were dead — never consumed by anything.

Net effect: you could not point either product at an external MongoDB or managed PostgreSQL without hand-editing `docker-compose.yml`, and since the installer renders only the single chosen database service, a stock install has no reachable backend for at least one of the two products.

## Fix

- Forward `_APP_DB_*_DOCUMENTSDB`, `_APP_DB_*_VECTORSDB` and both `_APP_CONNECTIONS_DATABASE_*` overrides on every service that resolves these pools.
- Describe all 14 variables in `variables.php` so the installer writes them.
- Read the per-engine credentials with `?:` fallback to the shared ones, so existing configurations behave exactly as before.

`appwrite-worker-deletes` and `appwrite-worker-migrations` resolve these pools (`Deletes.php`, `Migrations.php` reference `DATABASE_TYPE_DOCUMENTSDB` / `DATABASE_TYPE_VECTORSDB`) and previously received neither family — they now get both. `appwrite-worker-stats-resources` inherits from `appwrite-worker` via `extends`, so it is covered.

## Verification

Rendered a PostgreSQL installation from the patched compose — it contains **no `mongodb` service**, so the old hardcoded `mongodb:27017` default cannot resolve. Grafted a MongoDB in under a different service name and pointed `_APP_DB_HOST_DOCUMENTSDB` at it:

```
OK  documentsdb collection  -> 201  c1
OK  documentsdb document    -> 201  d1
OK  documentsdb read back   -> 200  {'hello': 'world', 'n': 5, ...}
OK  vectorsdb collection    -> 201  v1
```

Before this change the same request returned `500 Server Error` with `MongoDB connection failed: Failed to connect to MongoDB at mongodb:27017` in the server log, and no configuration could avoid it.

## What this does not fix

The installer still renders only the single chosen database service, so a default install has no MongoDB for DocumentsDB (and a MariaDB install has neither backend). This change makes that *configurable*; deciding whether `install` should offer to provision the extra engines, and making database-create fail fast with a clear error instead of returning 201 and then 500 on first write, are worth doing separately.

## Test plan

- [x] `npm run lint:compose` (Prettier) passes
- [x] Pint PSR-12 passes on `registers.php` and `variables.php`
- [x] `php -l` clean on both PHP files
- [x] `docker compose config` shows the new variables reaching the `appwrite` service
- [x] DocumentsDB create/write/read works on a mongodb-service-less install via `_APP_DB_HOST_DOCUMENTSDB`
- [x] Defaults unchanged when the new variables are unset


ChiragAgg5k:
Valid finding — confirmed and fixed in `bb8d5c85ed`.

I reproduced it by running the real `install` task against this branch:

```
_APP_DB_PASS="password"                        <- what the bundled mongodb/postgresql are provisioned with
_APP_DB_PASS_DOCUMENTSDB="&Id:sb.ucrRpb4S7"    <- generated, wins the fallback
_APP_DB_PASS_VECTORSDB="d_j%relv^wlwD:DP"      <- generated, wins the fallback
```

The chain is `Install.php:421` — `filter === 'password'` with no default plus `$shouldGenerateSecrets` calls `generatePasswordValue()`, so a fresh install wrote a random secret to both. Those non-empty values then beat `_APP_DB_PASS` in the `?:` fallback I added to `registers.php`, while `MONGO_INITDB_PASSWORD` / `POSTGRES_PASSWORD` are set from `_APP_DB_PASS`. So both pools would have authenticated with a password the database never had — a fresh-install regression, worse than the behaviour this PR set out to fix.

One extra wrinkle supporting the fix: the DSN sanitiser in `generatePasswordValue()` only applies to names matching `/^_APP_DB_.*_PASS$/`, which these two do not, so the generated values were not DSN-safe either — note the `&`, `:` and `%` above, exactly the characters the code comment at `Install.php:425` calls out as having broken DSNs.

Fix is to clear the filter so both install empty and fall back to `_APP_DB_PASS`, matching the `_APP_DB_USER_*` / `_APP_DB_SCHEMA_*` overrides. Re-ran the installer to confirm:

```
_APP_DB_PASS="password"
_APP_DB_SCHEMA_DOCUMENTSDB=
_APP_DB_USER_DOCUMENTSDB=
_APP_DB_PASS_DOCUMENTSDB=
_APP_DB_SCHEMA_VECTORSDB=
_APP_DB_USER_VECTORSDB=
_APP_DB_PASS_VECTORSDB=
```

`_APP_OPENSSL_KEY_V1` and `_APP_DB_ROOT_PASS` are unchanged, so no other generated secret was affected.

ChiragAgg5k:

ArnabChatterjee20k (approved):

PR #13382, opened 2026-08-27
feat: let operators choose which database products to deploy

ChiragAgg5k:
DocumentsDB runs only on MongoDB and VectorsDB only on PostgreSQL, but the installer provisioned a single engine — the one chosen for the platform. A default PostgreSQL installation therefore had no MongoDB, so creating a DocumentsDB database returned `201` and the **first collection write failed with a bare `Server Error`**, the cause visible only in the server log:

```
MongoDB connection failed: Failed to connect to MongoDB at mongodb:27017
```

A MariaDB installation lost both products the same way. Every single-engine install silently lost at least one database product.

## Approach

Rather than always deploying every engine, ask which products the installation actually wants, and deploy only what those answers require.

### Products are asked for during install

`_APP_DOCUMENTSDB` and `_APP_VECTORSDB`, both `enabled` by default. Seeded from the environment for scripted runs, and preserved from the existing `.env` on upgrade.

shot-1-prompts

### The web installer offers the same choice

Both products appear as toggles beside the database selector, defaulted from the variable defaults.

web-step1

The choice is carried through the wizard and confirmed on the review step before the install starts.

web-review2

### Only the needed engines are deployed

An engine is added when the platform uses it, or when an enabled product needs it. Turning both off leaves a single-engine deployment.

shot-2-engines

### A disabled product returns a clear error

Its routes throw `GENERAL_SERVICE_DISABLED` — for API keys and privileged roles too, since the engine behind them is not deployed. Other products are unaffected.

shot-3-api

## Verification

End to end on a real generated installation, not a synthetic one:

1. Ran the CLI `install` through a PTY answering **DocumentsDB = n**, VectorsDB = y.
2. The generated `.env` carried `_APP_DOCUMENTSDB="disabled"` and the rendered `docker-compose.yml` **omitted the mongodb service** — the booted stack ran 15 containers with no MongoDB.
3. Against that stack, `POST /v1/documentsdb` returned `403 general_service_disabled`, while TablesDB and VectorsDB returned `201`.

Also verified all four non-interactive combinations produce the expected engine set and `.env` values, and that a product's engine still survives when it is the platform engine (`database=mongodb` with DocumentsDB off keeps MongoDB).

## Test plan

- [x] Unit suite 1071 → 1074; Pint PSR-12 clean on all changed files; Prettier clean on `docker-compose.yml`
- [x] Updated 2 Generator tests that encoded the old "one engine only" assumption, added 3 covering the toggles
- [x] Interactive install honours `n` (verified through a PTY, not piped stdin)
- [x] All four enabled/disabled combinations verified for engines and `.env`
- [x] Disabled routes return `403`; enabled products unaffected
- [x] Web installer: toggling DocumentsDB off carries through the wizard and shows `Disabled` on the review step
- [x] All installer JS passes `node --check`; all touched `.phtml` pass `php -l`

## Notes for review

**The error copy is imprecise for this case.** `GENERAL_SERVICE_DISABLED` reads *"You can enable the service from the Appwrite console."*, but an operator re-enables this through the environment, not the console. I kept the existing error type so the SDK-visible contract stays stable; making the message context-aware would need either a distinct type or a parameterised message.

**Defaults preserve current behaviour but add a container.** Both default to `enabled`, so a default PostgreSQL install now also provisions MongoDB — one more container than 2.0.0 as it stands. That is the cost of DocumentsDB working out of the box. Defaulting them to `disabled` instead is a one-line change if you would rather operators opt in.

### DIFF
--- a/app/config/variables.php
+++ b/app/config/variables.php
@@ -481,6 +481,24 @@ return [
                 'question' => '',
                 'filter' => ''
             ],
+            [
+                'name' => '_APP_DOCUMENTSDB',
+                'description' => 'Enables the DocumentsDB API. It runs on MongoDB, so the installation needs a reachable MongoDB while this is enabled. Set to disabled to leave MongoDB out; the /v1/documentsdb routes then return a service disabled error. Default value is: enabled.',
+                'introduction' => '2.0.0',
+                'default' => 'enabled',
+                'required' => false,
+                'question' => 'Enable DocumentsDB? It requires MongoDB (Y/n)',
+                'filter' => ''
+            ],
+            [
+                'name' => '_APP_VECTORSDB',
+                'description' => 'Enables the VectorsDB API. It runs on PostgreSQL, so the installation needs a reachable PostgreSQL while this is enabled. Set to disabled to leave PostgreSQL out; the /v1/vectorsdb routes then return a service disabled error. Default value is: enabled.',
+                'introduction' => '2.0.0',
+                'default' => 'enabled',
+                'required' => false,
+                'question' => 'Enable VectorsDB? It requires PostgreSQL (Y/n)',
+                'filter' => ''
+            ],
         ],
     ],
     [
--- a/app/controllers/shared/api.php
+++ b/app/controllers/shared/api.php
@@ -463,13 +463,13 @@ Http::init()
 
             // An operator turns a database product off when its engine is not deployed,
             // so the route is unavailable to everyone, keys and privileged roles included.
-            $productEngines = [
+            $productToggles = [
                 'documentsdb' => '_APP_DOCUMENTSDB',
                 'vectorsdb' => '_APP_VECTORSDB',
             ];
             if (
-                isset($productEngines[$namespace])
-                && System::getEnv($productEngines[$namespace], 'enabled') !== 'enabled'
+                isset($productToggles[$namespace])
+                && System::getEnv($productToggles[$namespace], 'enabled') !== 'enabled'
             ) {
                 throw new Exception(Exception::GENERAL_SERVICE_DISABLED);
             }
--- a/app/controllers/shared/api.php
+++ b/app/controllers/shared/api.php
@@ -461,6 +461,19 @@ Http::init()
         if (! empty($method)) {
             $namespace = \strtolower($method->getNamespace());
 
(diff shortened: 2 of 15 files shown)


ChiragAgg5k:
All four addressed in `16e67b869e`.

**P1 — MongoDB support files not installed.** Confirmed and fixed. `Tasks/Install.php` now copies them whenever MongoDB is required by *either* the platform or DocumentsDB, not only when it backs the platform. This was a real defect: I had hit the exact failure mode by hand earlier in my own test rig, where Docker creates *directories* in place of the missing bind-mount sources and the mongo container restart-loops with `/mongo-entrypoint.sh: Is a directory`. Verified on a real generated install:

```
DOCS      VEC        ENGINES               MONGO SUPPORT FILES
enabled   enabled    mongodb postgresql    mongo-init.js=yes  mongo-entrypoint.sh=yes
disabled  enabled    postgresql            mongo-init.js=NO   mongo-entrypoint.sh=NO
```

**P1 — Web upgrades overwrite disabled choices.** Confirmed and fixed. Both fields now seed as `null` and hydrate from the rendered defaults in `applyBodyDefaults`, alongside `forceHttps`. The diagnosis was right, and there was a second half to it: I added the `data-default-documentsdb` / `data-default-vectorsdb` attributes but never actually read them, so even seeding as `null` alone would not have helped.

Verified against an existing `.env` holding `_APP_DOCUMENTSDB="disabled"`: the checkbox renders unchecked, and the review step now reports `DocumentsDB: Disabled` / `VectorsDB: Enabled` instead of silently submitting both as enabled.

**P2 — Duplicate PHPDoc.** Removed the orphaned `@return string[]` block above `getRequiredBackingServices()`.

**P2 — Mysterious name.** Renamed `$productEngines` to `$productToggles`. Agreed the old name was wrong: it maps a namespace to its environment variable, not to an engine.

Pint PSR-12 clean on all three PHP files, `node --check` clean on `state.js`, unit suite 1075 passing.

ChiragAgg5k:
Closing this. Rather than letting operators pick which of DocumentsDB and VectorsDB to deploy, both are being turned off for self-hosted, so their routes return `general_service_disabled` and neither engine has to be provisioned. That removes the installer prompts, the web installer step, and the per-product engine selection in the compose generator that this PR added.

PR #13385, opened 2026-08-27
feat(console): allow Google sign-in to the console project

HarshMN2345:
## What

Adds a Google slot to the console project's `oAuthProviders`, matching the GitHub/GitLab/Bitbucket entries added previously.

## Why

The console UI already ships a **Sign in with Google** button, but the API returns:

```
412  project_provider_disabled  "This provider is disabled."
```

because the console project has no `googleEnabled` / `googleAppid` / `googleSecret` keys — there's nowhere to put credentials. Same gap GitLab and Bitbucket had before #13337.

`google` is already a supported provider in `app/config/oAuthProviders.php`, so no adapter work is needed.

## Deploying

Set on the console environment, with the OAuth client's redirect URI as `/v1/account/sessions/oauth2/callback/google/console`:

```
_APP_CONSOLE_GOOGLE_APP_ID
_APP_CONSOLE_GOOGLE_SECRET
```

Unlike GitLab, Google's secret is a plain string — no JSON wrapping.

Matching cloud-side change: appwrite-labs/cloud (same branch name).

## Test plan
- [x] `php -l` clean
- [ ] With credentials set, `/v1/account/sessions/oauth2/google` 301s to `accounts.google.com`
- [ ] GitHub/GitLab/Bitbucket sign-in unaffected


app/config/console.php
```diff
@@ -64,6 +64,9 @@
         'bitbucketEnabled' => true,
         'bitbucketSecret' => System::getEnv('_APP_CONSOLE_BITBUCKET_SECRET', ''),
         'bitbucketAppid' => System::getEnv('_APP_CONSOLE_BITBUCKET_APP_ID', ''),
+        'googleEnabled' => true,
+        'googleSecret' => System::getEnv('_APP_CONSOLE_GOOGLE_SECRET', ''),
```
Meldiron, line 68: I think there is more to this PR; cant remembe rexact list, but for example .env needs the key, docker-compose, and one file for docs.
Then also maybe phtml file.

Plus let's check this with bitbucket and gitlab and github vars too

abnegate (approved):

PR #13386, opened 2026-08-27
Store the OAuth2 id_token on sessions and identities

Meldiron:
## What does this PR do?

Adds `providerIdToken` to sessions and identities, persisting the raw OpenID Connect `id_token` alongside the existing `providerAccessToken` / `providerRefreshToken`.

**Why:** OIDC providers return claims in the id_token that core never maps to structured columns — organization/tenant IDs, groups, locale, custom claims. Persisting the raw token (encrypted at rest, exposed on the Session and Identity models the same way `providerAccessToken` already is) lets applications decode whatever extra claims they need without Appwrite having to model every provider's claim set. Same pattern Auth.js/NextAuth uses for its account table.

### Behavior

- The base `OAuth2` class gains `getIdToken()`, reading `id_token` from the token response — every adapter whose provider returns one (Google, Apple, Microsoft, Oidc, Okta, Keycloak, Auth0, Authentik, FusionAuth, HuggingFace, …) is covered with zero per-adapter code. Plain OAuth2 providers store an empty string.
- The OAuth2 callback persists it on the identity (create and re-login update) and the session; the `createOAuth2Token` → `createSession` flow carries it from the identity onto the session like the other provider credentials.
- `PATCH /v1/account/sessions/:sessionId` (refresh) updates it, and keeps the stored value when the provider omits `id_token` from the refresh response (OIDC allows both).
- Stored encrypted (`encrypt` filter), like the access and refresh tokens.

### Changes

- `src/Appwrite/Auth/OAuth2.php` — `getIdToken()`
- `app/config/collections/common.php` — `providerIdToken` attribute on `sessions` and `identities`
- `app/controllers/api/account.php` — callback, token→session, and refresh flows
- `src/Appwrite/Utopia/Response/Model/Session.php`, `Identity.php` — expose `providerIdToken` (additive)
- `src/Appwrite/Migration/Version/V25.php` — create the attribute on existing `sessions` and `identities` collections
- `app/controllers/mock.php` — mock provider now issues an `id_token`, so e2e covers the full loop

## Test Plan

- New unit test `tests/unit/Auth/OAuth2/IdTokenTest.php` (present and missing `id_token` in the token response).
- E2E: mock provider issues a JWT-shaped `id_token`; assertions added to `testConvertAnonymousAccountOAuth2` (createOAuth2Session persists it; updateSession refresh keeps it) and `testOAuth2TokenSessionProviderAccessToken` (token flow copies identity → session).
- Full `AccountCustomClientTest` e2e suite: 64 tests, 938 assertions, green. Full unit suite green. Pint, PHPStan (level 4), and Rector clean.

## Notes for reviewers

- Existing projects need the V25 migration to gain the attribute; fresh projects get it from the collections config. Local dev stacks with pre-existing projects must run `migrate` (or recreate the project) before OAuth sign-ins on this branch, since the session/identity writes now include the new attribute.
- Response model changes are additive, so no request/response filters are required (same as the recent Identity `photo` addition).


Meldiron:
Keeping as draft. I thought Reend oauth will need this, but it seems it doesnt. Due to DB shema changes needed, I will postpone this until a provider actually needs it.

PR #13399, opened 2026-08-28
refactor(install): versioned infrastructure migrations, carrying build artifacts across the 2.0 volume rename

ChiragAgg5k:
Upgrading a pre-2.0 installation that has functions or sites leaves every existing build artifact behind, and every deployment stops serving.

## What happens

Before 2.0 the builds volume was declared without a name:

```yaml
appwrite-builds: ~        # Compose names it <project>_appwrite-builds
```

2.0 names it explicitly, because jobs-service build containers are created **outside** the Compose project and mount it by that literal name (`Deployment/Deployments.php:545`, via `_APP_BUILDS_VOLUME`):

```yaml
appwrite-builds:
  name: appwrite-builds   # no project prefix
```

Those are two different volumes. The upgrade mounts the new, empty one at `/storage/builds` and the real artifacts stay in `appwrite_appwrite-builds` with nothing attached to it. Nothing copies them across.

It is the only volume in the file carrying an explicit `name:` — every other one is project-prefixed on both versions, so this is the single place where the rename bites.

## Why it is hard to diagnose

The database is untouched and looks perfectly healthy: the site is enabled, the deployment is ready, the rule is verified, and `buildPath` points at a real-looking `/storage/builds/app-…/….gz`. That file simply is not in the mounted volume any more.

The executor cannot unpack a source that is not there, so it never creates a runtime container, and the router waits until the resource timeout:

```
Timed out waiting for runtime.
http.request · 30.35s · http.response.code 500
```

The only log line points at the runtime rather than at the missing file, and because the browser just hangs the first suspicion tends to be TLS or DNS — both of which check out fine.

Reported by an operator upgrading 1.9.6 → 2.0 with three projects; all 14 artifacts were stranded.

## The fix

Upgrades that change the database have data migrations, and API breaks have request and
response filters, both keyed by the release that introduced them. Changes to what the
containers *run on* had nowhere to live, so this adds the third registry in that shape:
`Appwrite\Migration\Infrastructure`.

```php
public static array $versions = [
    '2.0.0' => 'V1',
];
```

Each release that changes infrastructure gets one class, declaring its changes as a named
list, since a release rarely changes exactly one thing:

```php
protected function changes(): array
{
    return [
        'Carry build artifacts onto the renamed builds volume' => $this->carryBuildArtifacts(...),
    ];
}
```

Each change is logged by name, and one that fails does not skip the ones after it — they
are independent, and a volume that could not be copied says nothing about a file that has
to move. `Migration::between($from, $to)`
picks the classes for the releases an upgrade crosses, oldest first, and `Install` runs
them after the new compose file and `.env` are written and before anything starts — the
only window where a volume, a mount, or a file the containers depend on can still be
moved. The version being upgraded from is read off the existing compose file before it is
overwritten, so re-running an upgrade replays nothing already applied, and a pre-release
counts as the release it leads to, so `2.0.0-rc.1` runs what `2.0.0` runs. A migration
that fails warns and lets the upgrade continue rather than taking it down.

`V1` is the builds volume: when the configured volume is empty and exactly one
`<project>_<volume>` holds files, copy them across.

- **Copies, never moves.** The previous volume stays intact as a fallback.
- **Resumable.** Files are matched by size, so a copy killed mid-file is finished rather than accepted, and an intact one is left untouched.
- **Refuses to guess.** More than one candidate warns and copies nothing, naming the candidates so the operator can pick.
- **Retryable.** A volume that cannot be read is told apart from an empty one, and anything that did not land keeps the upgrade's source version so the next attempt tries again.
- **No new image dependency.** Volumes are inspected with the Appwrite image the upgrade has already pulled.

The source version is read off the compose file before it is rewritten, and remembered in
the installer config until every change has landed — an interrupted attempt leaves the
compose file and `.env` reading as the version being installed, which would otherwise look
like an upgrade with nothing to cross.

Reverting to the unprefixed name was the other option, but it would break every
jobs-service build, since the orchestrator needs a predictable name for a volume it mounts
from outside the project.

## Verification

Run against real Docker volumes:

| Scenario | Result |
|---|---|
| Legacy populated, target absent | copies everything, legacy left in place |
| Copy interrupted, target partly filled | fills in only what is missing |
| Target holds a truncated file | rewritten to full size |
| Target holds a newer artifact | left untouched, the rest filled in |
| Re-run after success | no files rewritten (mtimes unchanged) |
| Volumes unreadable / copy container fails | raises, source version kept for the next attempt |
| Two populated legacy volumes | warns, names both, target left empty |
| No legacy volume | silent no-op |
| Target is `2.0.0-rc.1` | runs, same as `2.0.0` |
| Already on 2.0.0 | not run at all |

## Test plan

- [x] Pint PSR-12 passes
- [x] PHPStan level 4 passes on the changed files
- [x] Artifacts land on the new volume and the old one is preserved
- [x] Interrupted, truncated, and re-run copies all resolve correctly
- [x] Failures keep the upgrade retryable
- [x] Ambiguous and absent cases handled without copying

## Note

This affects **every** 1.9.x installation with functions or sites, on any database, so it is worth landing before 2.0.0 stable. `2.0.0-rc.1` is already published with the problem; testers upgrading onto it will hit dead deployments. The manual remedy meanwhile is:

```bash
docker run --rm -v appwrite_appwrite-builds:/from:ro -v appwrite-builds:/to alpine \
    sh -c 'cp -a /from/. /to/'
```


ChiragAgg5k:
Valid, and it defeated the entire fix. Addressed in `06d9275f3c`.

The guard required the new volume to already exist:

```php
if (!$this->volumeExists($target) || $this->volumeFileCount($target, $image) > 0) {
    return;
}
```

On a real upgrade it does not exist — Compose creates it when the containers start, which is after this runs — so every genuine upgrade returned without copying, and Compose then created the empty volume the migration was meant to prevent. The condition is now inverted: only an *existing* volume that already holds files means there is nothing to carry over. When the volume is missing, the copy creates it, so the containers attach to one that already has the artifacts.

My verification was at fault, not just the code. I had hand-created the target volume in the fixture, so the test exercised a state that never occurs during an upgrade. Re-tested with the target genuinely absent:

```
target exists before? NO   <- real upgrade state
Copying build artifacts from "bvtest_appwrite-builds-test" to "appwrite-builds-test"...
Copied 3 build file(s). "bvtest_appwrite-builds-test" was left in place.
target now: 3 files
legacy still: 3 files
```

The other paths still behave, and the ambiguous case no longer leaves a stray empty volume behind:

| Scenario | Result |
|---|---|
| Target missing, one populated legacy | copies 3 files, legacy preserved |
| Target already populated | skips |
| Two populated legacy candidates | warns, names both, target not created |
| No legacy volume | silent no-op |

Pint PSR-12 passes.

ChiragAgg5k:
@greptile review

ChiragAgg5k:
@greptile review

ChiragAgg5k:
@greptile review

ChiragAgg5k:
@greptile review

ChiragAgg5k:
@greptile review

ChiragAgg5k:
@greptile review

ChiragAgg5k:
@greptile review

loks0n (approved):

PR #13426, opened 2026-08-30
fix(vcs): stop the installation callback dead-ending, on every provider

HarshMN2345:
## What

`GET /v1/vcs/github/callback` throws a bare 400 when the request carries no `state`, with a message telling the user that installing from an organisation is unsupported for members. The condition is only `empty($state)` — it never looks at the caller's GitHub role — so the message is wrong for anyone who reaches it and offers no way forward. GitHub only echoes `state` back when the flow finishes through the `redirect_uri`; an organisation member requesting owner approval finishes on the app's setup URL instead, which carries no state.

While fixing that, the same redirect defects turned out to be live on the shared OAuth2 callback used by Gitea, GitLab and Bitbucket.

## Changes

**GitHub callback (`GitHub/Callback/Get.php`)**
- Missing state reports the real condition instead of blaming the caller's organisation role.
- Error redirects used a `':'` separator instead of `'?'`, so the console received `/settings/git-installations:error=…` and could never read the error back.
- Empty `success`/`failure` values in state overrode the computed console defaults via `array_merge`, turning the remaining failure branch into a bare 400.
- `setup_action` is read: a `match` over GitHub's documented values (`request`, `install`, `update`) with an explicit message for anything unexpected.
- State validator accepts an empty value and 4096 chars; malformed JSON no longer produces a misleading 404.

**Shared OAuth2 callback (`Callback/Base.php` — Gitea, GitLab, Bitbucket)**
- Same `array_merge` defect: authorize signs empty `success`/`failure` into state when none are given, so these providers redirected to `''` on success and threw instead of redirecting on failure. Fixed the same way.
- Same validator change.

**Origin callback** — validator widened to 4096 to match.

## Testing

All on Gitea, in the existing `VCSGiteaConsoleClientTest`, driving the shared callback end to end without provider credentials: missing state, empty state, tampered state, invalid JSON, unknown project, missing code (error arrives as a query string), empty redirect URLs falling back to the console default, an over-2048 state, and a full install with no redirect URLs given — the last two are the ones that fail on `main` and pass here.

**Known gap, stated plainly:** `GitHub/Callback/Get.php` is `extends Action`, not `extends Base` — a separate copy of the same logic — so the Gitea suite does not reach it, and this PR ships the GitHub-specific changes (the `setup_action` match, the separator fix in that file) without tests. Closing that is the planned merge of the provider callbacks into shared code, after which one implementation is covered by Gitea.

`composer analyze` (full), `composer lint` and `composer refactor:check` clean.


tests/e2e/Services/VCSGitHub/VCSGitHubCallbackConsoleClientTest.php
Meldiron, line 1: Lets switch to gitea, and test what we are changing (happy, failure)

src/Appwrite/Platform/Modules/VCS/Http/GitHub/Callback/Get.php
```diff
@@ -73,16 +72,16 @@ public function action(
...
-                return;
+            if (empty($redirectFailure)) {
+                throw new Exception(Exception::PROJECT_NOT_FOUND, $error);
             }
 
-            throw new Exception(Exception::PROJECT_NOT_FOUND, $error);
+            $separator = \str_contains($redirectFailure, '?') ? '&' : '?';
+            $response
+                ->addHeader('Cache-Control', 'no-store, no-cache, must-revalidate, max-age=0')
+                ->addHeader('Pragma', 'no-cache')
```
Meldiron: Things like these seem to have no logical change

src/Appwrite/Platform/Modules/VCS/Http/GitHub/Callback/Get.php
```diff
@@ -165,18 +162,20 @@ public function action(
...
-            if (!empty($redirectFailure)) {
-                $separator = \str_contains($redirectFailure, '?') ? '&' : ':';
-                $response
-                    ->addHeader('Cache-Control', 'no-store, no-cache, must-revalidate, max-age=0')
-                    ->addHeader('Pragma', 'no-cache')
-                    ->redirect($redirectFailure . $separator . \http_build_query(['error' => $error]));
-                return;
+            $error = $setupAction === 'request'
+                ? 'Your request was sent to the organization owners. An owner must complete the installation from the Appwrite Console; approving the request on GitHub is not enough.'
+                : 'Installation of the Appwrite GitHub App on organization accounts is restricted to organization owners. As a member of the organization, you do not have the necessary permissions to install this GitHub App. Please contact the organization owner to create the installation from the Appwrite Console.';
```
Meldiron: use php match instead

Meldiron (approved):

PR #13431, opened 2026-08-31
fix(usage): don't bill build compute when buildStartedAt is null

lohanidamodar:
## Problem

`Appwrite\Usage\Build::publish()` derives build billing from the deployment's `buildDuration`:

```php
$mbSeconds = (int) ($memory * $deployment->getAttribute('buildDuration', 0) * $cpus);
```

`buildDuration` is computed in `Jobs::duration()`, which falls back to the deployment's creation time when the build never recorded a start:

```php
$startedAt = $deployment->getAttribute('buildStartedAt', '') ?: $deployment->getCreatedAt();
```

`buildStartedAt` is only stamped by the first build log callback. A build that terminates without producing any log output finalizes with `buildStartedAt` still null, so `buildDuration` becomes wall-clock time since the deployment row was created — queue wait included. That inflated number is then billed as build compute and flows into `functions.builds.mbSeconds` and onward into GBHours.

The impact scales with queue depth: the longer a deployment waited before failing silently, the more phantom compute it bills.

## Fix

Zero `buildDuration` and `mbSeconds` when `buildStartedAt` was never stamped, so no compute is attributed to a build we have no evidence actually ran.

## Why zeroing instead of an early return

An early return would also drop `builds`, `builds.storage` and the ready/failed counters, which are legitimately observed regardless of timing — the build really did happen and really did produce (or fail to produce) an artifact. Build counts must stay honest; only the phantom time should go away.

Zeroing gets that for free: both StatsUsage workers already skip zero-valued metrics, so the compute and mbSeconds rows drop out on their own while every count and storage metric still lands. `$buildDuration` feeds `builds.compute`, `builds.compute.success`/`.failed` and their `{resourceType}` variants, so zeroing removes all compute attribution, not just mbSeconds.

## Known trade-off

Out-of-order or dropped log callbacks can leave `buildStartedAt` null on a build that genuinely ran, in which case we now under-bill instead of over-bill. That is the deliberate direction to err in, and the docblock on `Jobs::duration()` already documents the fragility of deriving start time from log callbacks.

The durable fix is to stamp `buildStartedAt` when the job is dispatched rather than on first log output, which makes the fallback unreachable. That is a broader change to the build lifecycle and is left as a follow-up.

The executor `Builds` worker is unaffected either way: it stamps `buildStartedAt` before the `try` whose `finally` sends usage, and derives duration from an in-process clock, so null can never reach usage on that path. The guard is a no-op there.

## Tests

New `tests/unit/Usage/BuildTest.php` covers:

- `buildStartedAt` set — mbSeconds stays `memory * buildDuration * cpus`, unchanged from today
- `buildStartedAt` absent — compute and mbSeconds published as 0, while `builds`, `builds.success` and `builds.storage` are unaffected
- `buildStartedAt` null on a failed build — same, across the `failed` branch

Verified the tests fail against the unpatched code (28672 mbSeconds billed instead of 0) and pass with the fix. Pint, PHPStan and Rector are clean; the full unit suite shows no change from baseline.

Fixes CLO-4783
https://linear.app/appwrite/issue/CLO-4783/build-usage-bills-queue-wait-as-compute-when-buildstartedat-is-null


tests/unit/Usage/BuildTest.php
loks0n, line 1: Can we remove this unit test? I don't think it adds much value

lohanidamodar: Removed in fd8f8e41 — PR is down to the seven-line guard.

PR #13437, opened 2026-09-01
feat: make the GitLab endpoint configurable

HarshMN2345:
## What does this PR do?

Both GitLab surfaces hardcoded `https://gitlab.com`, so a self-hosted GitLab could not be used for repository integration or for signing in to the console. This adds one environment variable per surface, each defaulting to `https://gitlab.com`.

> **Note:** this reverts an earlier call. On #12894 these variables were proposed and dropped — ["I dont think we need these variables - we point ot to official gitlab which is always the same URLs here"](https://github.com/appwrite/appwrite/pull/12894#discussion_r3614383704). @Meldiron asked for them back for self-hosted instances, hence this PR.

**`_APP_VCS_GITLAB_ENDPOINT`** — repository integration. Replaces the hardcoded endpoint in three places:

- `app/config/vcs.php` — the registry `endpoint` that `Vcs\Factory` hands to `setEndpoint()` and to the OAuth2 builder
- `VCS/Http/Gitlab/Authorize/Get.php` and `VCS/Http/Gitlab/Callback/Get.php`, which build their own client rather than going through the factory

**`_APP_CONSOLE_GITLAB_ENDPOINT`** — console sign-in, in `app/config/console.php`.

`Auth\OAuth2\Gitlab` reads its endpoint out of a JSON-encoded secret, so `gitlabSecret` now carries `{"clientSecret", "endpoint"}` instead of the raw secret string. `console.php` builds a document rather than an adapter, so it has no `setEndpoint()` to call — the secret is the only channel available without special-casing GitLab in `account.php`. The value stays an empty string when unconfigured, so `account.php:1366` keeps reporting the provider as disabled rather than half-configured.

Both variables are registered in `app/config/variables.php`, `.env`, and the three compose services that carry the `_APP_VCS_*` block.

### Second commit: `getAppSecret()` no longer rejects a plain string

Independent of the endpoint work. `Gitlab::getAppSecret()` decoded with `JSON_THROW_ON_ERROR` and turned any failure into `Invalid secret`. That breaks two real cases:

- The still-registered legacy `PATCH /v1/projects/:projectId/oauth2` (`app/controllers/api/projects.php:28`, `gitlab` is in the whitelist) writes the secret verbatim at `:55`. Any project configured through it fails sign-in with a 500.
- Console GitLab sign-in on `main` today, for anyone who actually sets `_APP_CONSOLE_GITLAB_APP_ID` + `_SECRET`. (A stock install with the vars empty gets a clean 400 `PROJECT_PROVIDER_DISABLED` instead.)

The fallback copies `Google.php:216-229`, which grew the same tolerance for the same reason — GitLab was the only one of the four console providers that threw. Also `rtrim`s the endpoint, as `Gitea::setEndpoint()` does, so a trailing slash no longer builds `//oauth/authorize`.

## Test Plan

Loaded the real config files and exercised the adapter directly (PSR-4 repointed at this branch's `src/`, verified via `ReflectionClass::getFileName()`).

Endpoint resolution:

| `_APP_*_GITLAB_ENDPOINT` | `vcs.gitlab.endpoint` | `getLoginURL()` host |
| --- | --- | --- |
| unset | `https://gitlab.com` | `https://gitlab.com` |
| `https://git.acme.internal` | `https://git.acme.internal` | `https://git.acme.internal` |

Secret handling, after the second commit:

| stored secret | before | after |
| --- | --- | --- |
| `s3cr3t` (legacy route) | `Exception: Invalid secret` | `https://gitlab.com/oauth/authorize`, `clientSecret=s3cr3t` |
| `12345` | `TypeError: ... must be of type array, int returned` | `https://gitlab.com/oauth/authorize`, `clientSecret=12345` |
| `{"clientSecret":"s3cr3t"}` | ok | ok |
| `{"clientSecret":"s3cr3t","endpoint":""}` | ok | `https://gitlab.com` (per-project blanks keep defaulting) |
| `{"clientSecret":"s3cr3t","endpoint":"https://git.acme.internal/"}` | `//oauth/authorize` | `https://git.acme.internal/oauth/authorize` |

`vendor/bin/pint --test` passes on every changed file. `tests/unit/Vcs/FactoryTest.php` is unaffected — it supplies its own registry rather than reading `app/config/vcs.php`.

## Related PRs and Issues

- Reverses the variable removal agreed on #12894.
- Follow-up to #13337, which added GitLab and Bitbucket console sign-in.

## Checklist

- [ ] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs? — no API metadata changed.


.env
Meldiron, line 1: lets leave empty value

Meldiron (approved):

PR #13451, opened 2026-09-02
fix(databases): expose transactionId on DocumentsDB and VectorsDB create SDK methods

atharvadeosthale:
## What does this PR do?

Adds `transactionId` to the SDK `Method` parameter lists of `createDocument` and `createDocuments` for DocumentsDB and VectorsDB.

Both routes already declare `->param('transactionId', ...)` and the backend stages the write when it is present. The SDK method definitions, however, list every parameter except `transactionId`, and the generator intersects the request body against that list. The result is that no generated SDK exposes `transactionId` on these two methods: typed SDKs reject it at compile time, and JavaScript silently drops the key so the write commits immediately instead of being staged.

TablesDB's `createRow` and `createRows` already include `new Parameter('transactionId', optional: true)` in their method definitions, and every other DocumentsDB and VectorsDB write method (`upsertDocument`, `updateDocument`, `deleteDocument`, the bulk variants, and the atomic increment and decrement methods) already exposes it. This change brings the two create methods in line.

## Test Plan

- `php -l` and `pint --test` pass on both files.
- Regenerating the specs shows `transactionId` in `x-appwrite.methods[].parameters` for both `createDocument` and `createDocuments` on `/documentsdb/.../documents` and `/vectorsdb/.../documents`, matching `/tablesdb/.../rows`.
- No behaviour change at the HTTP layer; the route parameter and its handling are untouched.

## Related PRs and Issues

The DocumentsDB documentation currently shows `createDocument` with `transactionId`; those examples only work once SDKs are regenerated from this change.

## Checklist


abnegate:
@cursoragent Ensure tests cover txns in these APIs

ArnabChatterjee20k (approved):

tests/unit/Platform/Modules/Databases/Http/CreateTransactionIdTest.php
abnegate, line 1: @cursoragent don't add http tests in unit dir, add to the existing e2e txn tests

PR #13464, opened 2026-09-03
Keep team total in sync with confirmed memberships

Meldiron:
## What does this PR do?

`teams.total` is meant to be the number of confirmed memberships, but four code paths could drift it away from the actual rows. This is user visible: a team can report `total = 2` while only one membership row exists for it.

**1. `Memberships/Delete.php` — the decrement read the wrong document.** The primary-transfer block reassigned `$membership` to the *successor* before the decrement:

```php
if ($team->getAttribute('userInternalId') === $membership->getAttribute('userInternalId')) {
    $membership = $dbForProject->findOne('memberships', [...]); // clobbered
}

if ($membership->getAttribute('confirm')) { // now the successor, not the deleted membership
```

When the only remaining member was still unconfirmed, `findOne` returned an empty document and the decrement was skipped entirely, leaving the team permanently over-counted. The same clobbering sent the successor's `membershipId` and document in the event payload. Renamed the lookup to `$successor`.

Note this branch only fires when `team.userInternalId` is set, which is a Cloud-only extension of the `teams` collection — it is unreachable in this repo, so it has no e2e coverage here.

**2. `Memberships/Create.php` — server-side confirm was never counted.** When a server/API-key caller invited a user who already had a pending membership, the row flipped to `confirm => true` but only the "new membership" branch incremented `total`. Under-counted, and a later delete then decremented it below the real count.

**3. `Status/Update.php` — invite accepts could double count.** The `confirm === true` guard ran ~70 lines before the increment, so two concurrent accepts of the same invite (double-clicked invite link, email prefetch, client retry) both passed the guard and both incremented. The flip and increment now share a transaction behind a `getDocument(..., forUpdate: true)` re-read; the loser gets `MEMBERSHIP_ALREADY_CONFIRMED` (409). Same treatment applied to the path in 2.

**4. `Teams/Create.php` — non-atomic seed.** `total => 1` was written, then the owner membership was created in a separate statement. A failure in between left a team counting a member that does not exist. Both writes now share a transaction.

The atomic `increaseDocumentAttribute` / `decreaseDocumentAttribute` primitives are kept rather than recomputing with `count()` on every write: they are already race-free (utopia runs them in a transaction with a locking read), whereas `count()` + `update` is a read-modify-write, `memberships` has no index on `confirm`, and the deletes worker calls the callback per deleted row, which would make bulk cleanup quadratic.

## Test Plan

Added `testCreateTeamMembershipConfirmsPendingInvite` to `tests/e2e/Services/Teams/TeamsCustomClientTest.php`: a client creates a team and a pending invite (`total` stays 1), a server-key invite confirms it (`total` must become 2 and match `GET /memberships` total), and re-inviting the confirmed member returns 409 without counting again.

Verified it is a real regression test — with the `Memberships/Create.php` fix reverted it fails on `Failed asserting that 1 matches expected 2`, and passes with the fix.

```
docker compose exec appwrite test tests/e2e/Services/Teams
OK (46 tests, 1057 assertions)
```

`composer lint` and `composer analyze` are clean on `src/Appwrite/Platform/Modules/Teams/`.

Not covered by tests: 1 is unreachable in this repo (see above), 3 is a concurrency window, and 4 needs an induced write failure.

## Related PRs and Issues

- (none)

## Checklist

- [x] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs? — no metadata changed

### DIFF
--- a/src/Appwrite/Platform/Modules/Teams/Http/Memberships/Create.php
+++ b/src/Appwrite/Platform/Modules/Teams/Http/Memberships/Create.php
@@ -303,12 +303,29 @@ class Create extends Action
             $invitedTime = DateTime::now();
 
             if ($isPrivilegedUser || $isAppUser) {
-                $membership = $authorization->skip(fn () => $dbForProject->updateDocument('memberships', $membership->getId(), new Document([
-                    'secret' => $secretHash,
-                    'invited' => $invitedTime,
-                    'joined' => DateTime::now(),
-                    'confirm' => true
-                ])));
+                $membership = $dbForProject->withTransaction(function () use ($dbForProject, $authorization, $membership, $team, $secretHash, $invitedTime) {
+                    // Re-read under a lock, a concurrent invite must not count the same member twice
+                    $current = $authorization->skip(fn () => $dbForProject->getDocument('memberships', $membership->getId(), forUpdate: true));
+
+                    if ($current->getAttribute('confirm') === true) {
+                        return new Document();
+                    }
+
+                    $confirmed = $authorization->skip(fn () => $dbForProject->updateDocument('memberships', $membership->getId(), new Document([
+                        'secret' => $secretHash,
+                        'invited' => $invitedTime,
+                        'joined' => DateTime::now(),
+                        'confirm' => true
+                    ])));
+
+                    $authorization->skip(fn () => $dbForProject->increaseDocumentAttribute('teams', $team->getId(), 'total', 1));
+
+                    return $confirmed;
+                });
+
+                if ($membership->isEmpty()) {
+                    throw new Exception(Exception::MEMBERSHIP_ALREADY_CONFIRMED);
+                }
             } else {
                 $membership = $dbForProject->updateDocument('memberships', $membership->getId(), new Document([
                     'secret' => $secretHash,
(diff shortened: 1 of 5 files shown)


src/Appwrite/Platform/Modules/Teams/Http/Memberships/Status/Update.php
Meldiron, line 1: Moved into transaction to prevent race conditions

src/Appwrite/Platform/Modules/Teams/Http/Memberships/Create.php
Meldiron, line 1: Added team total count incrase in a transaction

tests/e2e/Services/Teams/TeamsCustomClientTest.php
Meldiron, line 1: Improved test coverage

src/Appwrite/Platform/Modules/Teams/Http/Memberships/Delete.php
Meldiron, line 1: Fix delete membership not decreasing count due to reused variable name, making check if membership is confirmed always false-out.

src/Appwrite/Platform/Modules/Teams/Http/Teams/Create.php
Meldiron, line 1: Add transaction to prevent race conditions

Meldiron (commented):
lgtm

PR #13480, opened 2026-09-04
perf(stats): use queued project context

loks0n:
## What changed

- inject the typed `ProjectContext` into the stats-usage worker instead of resolving the current project document
- retain the queued payload/context project ID consistency check
- use the queued project sequence as the usage tenant

## Why

Usage is event-time data. Resolving the mutable platform project for every queued usage message adds a platform database/cache read and can discard attribution when a project is deleted before its already-enqueued usage is consumed. The message already carries the immutable identity needed by this worker.

Cloud has a separate batched stats-usage implementation that already consumes the queued project payload without resolving the platform project. This change optimizes the CE worker path.

## Verification

- `php -l src/Appwrite/Platform/Workers/StatsUsage.php`
- Pint on `src/Appwrite/Platform/Workers/StatsUsage.php`
- PHPStan on `src/Appwrite/Platform/Workers/StatsUsage.php`

No worker unit test was added because this repository requires worker behavior to be covered through E2E rather than unit-testing action internals; this dependency-only change has no dedicated stats-usage E2E seam.

### DIFF
--- a/src/Appwrite/Platform/Workers/StatsUsage.php
+++ b/src/Appwrite/Platform/Workers/StatsUsage.php
@@ -3,9 +3,9 @@
 namespace Appwrite\Platform\Workers;
 
 use Appwrite\Detector\Detector;
+use Appwrite\Event\Message\ProjectContext;
 use Appwrite\Usage\Connection;
 use Utopia\Console;
-use Utopia\Database\Document;
 use Utopia\Platform\Action;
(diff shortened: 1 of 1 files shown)


loks0n:
Closing because the optimization is CE-only; Appwrite Cloud already avoids this project lookup in its custom stats-usage worker.

PR #13498, opened 2026-09-05
fix: surface migration report errors and preserve underlying exception (#13477)

### LINKED ISSUE #13477: Self-hosted → self-hosted migration fails at report stage — Missing required field "policies" for Appwrite\Models\Database is masked as "Unable to connect to the migration source"
### 👟 Reproduction steps

1. Run two self-hosted Appwrite **1.9.6** instances (official Docker images) — a source ("prod") and a destination ("dev"). The source project contains databases/tables, users, and storage.
2. In the source project, create an API key with **all scopes**.
3. On the destination: Console → Project → Settings → Migrations → Create migration → select **Appwrite** as the source.
4. Enter the source endpoint (`https://<source-host>/v1`), the source project ID, and the API key; select resources (any selection that includes databases).
5. The console issues `GET /v1/migrations/appwrite/report`, which returns **400**.

### 👍 Expected behavior

- The migration report is generated and the wizard proceeds, **or**
- If the source call fails, the response/logs surface the actual error instead of a generic connectivity message.

### 👎 Actual Behavior

`GET /v1/migrations/appwrite/report` returns 400 with:

```
Appwrite\Extend\Exception: Unable to connect to the migration source. Please verify your credentials and ensure the source is reachable from this server. Check for network restrictions such as firewalls, IP allowlists, or outbound connectivity limits.
  at /usr/src/code/src/Appwrite/Platform/Modules/Migrations/Http/Migrations/Appwrite/Report/Get.php:71
```

### 🎲 Appwrite version

### 💻 Operating system

Linux

## Diagnostics performed (connectivity and credentials are NOT the problem)

All executed from **inside the destination `appwrite` container**, against the source:

- DNS resolves the source FQDN correctly; host is pingable.
- `curl -sv https://<source>/v1/health` → TLS handshake succeeds (publicly trusted Let's Encrypt certificate, verification OK); with `X-Appwrite-Project` and `X-Appwrite-Key` (all scopes) → **HTTP 200** `{"name":"http","ping":0,"status":"pass"}`.
- `curl -s https://<source>/v1/health/version` → **HTTP 200** `{"version":"1.9.6"}`.
- PHP cURL inside the container (same call) → errno 0, HTTP 200.

Reproducing the exact library call that `Get.php` makes:

```php
require "/usr/src/code/vendor/autoload.php";

$source = new Utopia\Migration\Sources\Appwrite(
    "<SOURCE_PROJECT_ID>",
    "https://<source-host>/v1",
    "<SOURCE_API_KEY>",
    fn() => null
);
$source->report();
```

produces the real, hidden error:

```
Exception #0: Missing required field "policies" for Appwrite\Models\Database.
```

## Root cause analysis

1. **Error masking.** `src/Appwrite/Platform/Modules/Migrations/Http/Migrations/Appwrite/Report/Get.php` wraps `new AppwriteSource(...)` + `$appwrite->report($resources)` in `catch (\Throwable $e)` and rethrows a generic `MIGRATION_PROVIDER_ERROR` **without the original message or previous exception**. Any failure — auth, parsing, SDK model mismatch — is misreported as a connectivity problem, which makes this class of bug nearly undiagnosable for users.

2. **The actual failure.** `report()` (`vendor/utopia-php/migration/src/Migration/Sources/Appwrite.php`) → `reportDatabases()` → `APIReader::report()` deserializes the source's database list into `Appwrite\Models\Database` (`vendor/appwrite/appwrite/src/Appwrite/Models/Database.php`), which **requires the `policies` key**.

3. **Self-hosted sources don't send `policies`.** The field is database backup-policy metadata present in the Cloud API model; self-hosted `GET /v1/databases` responses omit it. The hard requirement appears to have been introduced by **"Add backup policy migration support" (#11632, first shipped in 1.9.5-rc.1)** — so from 1.9.5 onward, the report stage fails against any source that doesn't emit backup-policy fields. `archives` (same feature area) will likely throw next once `policies` is tolerated.

4. **Not version skew.** Source and destination are both 1.9.6 — this is a same-version, self-hosted → self-hosted failure.

## Suggested fixes

1. Make `policies` / `archives` optional (default `[]`) in the SDK model(s) used by `utopia-php/migration`'s APIReader, or otherwise tolerate sources that omit backup-related fields (e.g. when backups are unavailable/disabled on the source).
2. In `Get.php`, attach the original exception as `previous` and/or log `$e->getMessage()`, so the true cause reaches logs even if the client-facing message stays generic.
3. Cover both execution paths: the `appwrite` container (report stage) **and** `appwrite-worker-migrations` (transfer stage re-runs the same reader/report code).

## 🎛 Appwrite setup

- Appwrite self-hosted **1.9.6** on both source and destination (official Docker images)
- Host: Proxmox VE 9 (Debian 13 "Trixie"), Docker CE
- Reverse proxy: Nginx Proxy Manager with a valid Let's Encrypt certificate (verified not implicated — direct container-to-API requests succeed)
- DNS: split-horizon via Pi-hole (source FQDN resolves to the proxy's LAN IP)

## 💡 Additional context

- Possibly related: #10711 "Project migration Fails after 1.8.0 update" — may share the same masked root cause.
- Related Discord/support thread (May 2026): "Databases silently skip during Cloud to self-hosted migration" — `Workers/Migrations.php` hardcoded `SourceAppwrite::SOURCE_DATABASE`, i.e. a different bug in the same database-reader path.
- Workaround (not acceptable long-term): patch the vendored SDK model inside both the `appwrite` and `appwrite-worker-migrations` containers to default `policies`/`archives` to `[]`; the change is ephemeral and reverts on container recreate.

shanmukhasaireddy13:
## What does this PR do?

During migration report generation (e.g. self-hosted Appwrite migrations, Supabase, NHost, and Firebase), when an exception occurred while connecting to or inspecting resources (such as missing scopes, permissions, or schema incompatibilities like missing `policies`), the generic `catch (\Throwable $e)` block replaced the error message with a hardcoded connectivity error:
> *"Unable to connect to the migration source. Please verify your credentials and ensure the source is reachable from this server..."*

Furthermore, the original exception was dropped because `previous: $e` was omitted when constructing `Appwrite\Extend\Exception`. This masked the actual root cause from users and maintainers, and prevented error logging/tracing tools from recording the original exception stack trace.

This PR:
1. Surfaces the actual error message (`'Failed to generate migration report: ' . $e->getMessage()`) when an exception occurs during report generation across all migration providers (Appwrite, Firebase, NHost, Supabase), retaining the fallback connectivity guidance if the exception message is empty.
2. Passes `previous: $e` into `Appwrite\Extend\Exception` so the full stack trace and original exception chain are preserved.

## Test Plan

1. **Unit Tests**:
   - Added unit tests in `tests/unit/Platform/Modules/Migrations/Appwrite/Report/GetTest.php` testing report failure error handling.
   - Verified that `Exception::MIGRATION_PROVIDER_ERROR` is thrown, `getPrevious()` is preserved and not null, and `getMessage()` contains the error description (and tests fallback message when exception message is empty).
   - Executed PHPUnit:
     ```bash
     docker run --rm -v ${PWD}:/app -w /app composer:latest vendor/bin/phpunit tests/unit/Platform/Modules/Migrations/Appwrite/Report/GetTest.php
     ```
     Result: All tests passed (OK, 2 tests, 6 assertions).

2. **Code Style & Formatting**:
   - Formatted and verified with Laravel Pint:
     ```bash
     docker run --rm -v ${PWD}:/app -w /app composer:latest vendor/bin/pint --test src/Appwrite/Platform/Modules/Migrations/Http/Migrations/Appwrite/Report/Get.php src/Appwrite/Platform/Modules/Migrations/Http/Migrations/Firebase/Report/Get.php src/Appwrite/Platform/Modules/Migrations/Http/Migrations/NHost/Report/Get.php src/Appwrite/Platform/Modules/Migrations/Http/Migrations/Supabase/Report/Get.php tests/unit/Platform/Modules/Migrations/Appwrite/Report/GetTest.php
     ```
     Result: PASS (5 files, PSR-12 compliant).

## Related PRs and Issues

- Fixes #13477

## Checklist

### DIFF
--- a/src/Appwrite/Platform/Modules/Migrations/Http/Migrations/Appwrite/Report/Get.php
+++ b/src/Appwrite/Platform/Modules/Migrations/Http/Migrations/Appwrite/Report/Get.php
@@ -68,9 +68,14 @@ class Get extends Action
             $appwrite = new AppwriteSource($projectID, $endpoint, $key, $getDatabasesDB);
             $report = $appwrite->report($resources);
         } catch (\Throwable $e) {
+            $message = !empty($e->getMessage())
+                ? 'Failed to generate migration report: ' . $e->getMessage()
+                : 'Unable to connect to the migration source. Please verify your credentials and ensure the source is reachable from this server. Check for network restrictions such as firewalls, IP allowlists, or outbound connectivity limits.';
+
             throw new Exception(
-                Exception::MIGRATION_PROVIDER_ERROR,
-                'Unable to connect to the migration source. Please verify your credentials and ensure the source is reachable from this server. Check for network restrictions such as firewalls, IP allowlists, or outbound connectivity limits.'
+                type: Exception::MIGRATION_PROVIDER_ERROR,
+                message: $message,
+                previous: $e
             );
         }
 
--- a/src/Appwrite/Platform/Modules/Migrations/Http/Migrations/Appwrite/Report/Get.php
+++ b/src/Appwrite/Platform/Modules/Migrations/Http/Migrations/Appwrite/Report/Get.php
@@ -65,7 +65,7 @@ class Get extends Action
         callable $getDatabasesDB
     ): void {
         try {
-            $appwrite = new AppwriteSource($projectID, $endpoint, $key, $getDatabasesDB);
+            $appwrite = $this->getSource($projectID, $endpoint, $key, $getDatabasesDB);
             $report = $appwrite->report($resources);
         } catch (\Throwable $e) {
             $message = !empty($e->getMessage())
@@ -82,5 +82,10 @@ class Get extends Action
         $response
             ->setStatusCode(Response::STATUS_CODE_OK)
             ->dynamic(new Document($report), Response::MODEL_MIGRATION_REPORT);
+    }
+
+    protected function getSource(string $projectID, string $endpoint, string $key, callable $getDatabasesDB): AppwriteSource
+    {
+        return new AppwriteSource($projectID, $endpoint, $key, $getDatabasesDB);
     }
 }
--- a/src/Appwrite/Platform/Modules/Migrations/Http/Migrations/Firebase/Report/Get.php
+++ b/src/Appwrite/Platform/Modules/Migrations/Http/Migrations/Firebase/Report/Get.php
@@ -68,9 +68,14 @@ class Get extends Action
             $firebase = new Firebase($serviceAccount);
             $report = $firebase->report($resources);
         } catch (\Throwable $e) {
+            $message = !empty($e->getMessage())
+                ? 'Failed to generate migration report: ' . $e->getMessage()
+                : 'Unable to connect to the migration source. Please verify your credentials and ensure the source is reachable from this server. Check for network restrictions such as firewalls, IP allowlists, or outbound connectivity limits.';
+
             throw new Exception(
-                Exception::MIGRATION_PROVIDER_ERROR,
-                'Unable to connect to the migration source. Please verify your credentials and ensure the source is reachable from this server. Check for network restrictions such as firewalls, IP allowlists, or outbound connectivity limits.'
+                type: Exception::MIGRATION_PROVIDER_ERROR,
+                message: $message,
+                previous: $e
             );
         }
 
--- a/src/Appwrite/Platform/Modules/Migrations/Http/Migrations/NHost/Report/Get.php
+++ b/src/Appwrite/Platform/Modules/Migrations/Http/Migrations/NHost/Report/Get.php
@@ -75,9 +75,14 @@ class Get extends Action
             $nhost = new NHost($subdomain, $region, $adminSecret, $database, $username, $password, $port);
             $report = $nhost->report($resources);
         } catch (\Throwable $e) {
+            $message = !empty($e->getMessage())
+                ? 'Failed to generate migration report: ' . $e->getMessage()
+                : 'Unable to connect to the migration source. Please verify your credentials and ensure the source is reachable from this server. Check for network restrictions such as firewalls, IP allowlists, or outbound connectivity limits.';
+
             throw new Exception(
-                Exception::MIGRATION_PROVIDER_ERROR,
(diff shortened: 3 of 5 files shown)


ddenev:
@shanmukhasaireddy13, just to mention that this PR does NOT fix the actual main problem in #13477, which is described in the 2nd bullet under "Root cause analysis":
> The actual failure. report() (vendor/utopia-php/migration/src/Migration/Sources/Appwrite.php) → reportDatabases() → APIReader::report() deserializes the source's database list into Appwrite\Models\Database (vendor/appwrite/appwrite/src/Appwrite/Models/Database.php), which requires the policies key.

This PR only fixes the error masking part of the above report, which is the lower prio fix.

shanmukhasaireddy13:
Thanks for checking the PR, @ddenev!

You're completely right that this PR specifically focuses on the error masking and exception chaining aspect of #13477 (across Appwrite, Firebase, NHost, and Supabase report endpoints), ensuring any failure surfaces the actual underlying cause instead of misdiagnosing it as a network/connectivity issue.

Regarding the second point (`Missing required field "policies"`):
That requirement originated in `appwrite/sdk-for-php` (packaged as `appwrite/appwrite` on Packagist and consumed by `utopia-php/migration`), which in version `26.1.0` (shipped with Appwrite 1.9.6) strictly enforced `policies` and `archives` in `Appwrite\Models\Database::from()`.

In `appwrite/sdk-for-php` `>= 27.0.0`, `policies` and `archives` were made optional (`?array $policies = null`, `?array $archives = null`). On the `main` branch, `composer.lock` has already bumped `appwrite/appwrite` to `27.1.0` (via commit `f1909e417f`), so on `main` the deserialization succeeds against self-hosted sources without throwing. For 1.9.x installations, backporting the lockfile SDK bump to the `1.9.x` release branch would address it for any upcoming 1.9.x patch release.

ddenev:
got it @shanmukhasaireddy13, thank you for clarifying!

So the main issue has already been fixed, we just need to wait for the next release of self-hosted then. Let's hope the Appwrite team publishes another v1.9.x fix release before the v2.0 since major releases are usually buggy and we need to wait for many months until they all issues and stabilize it.

PR #13510, opened 2026-09-07
refactor(vcs): reuse domain operations across provider event endpoints

HarshMN2345:
## What

The five VCS webhook receivers repeated the same work. This extracts the parts that are genuine **domain operations** and leaves the orchestration where it is.

Net against `main`: **297 lines removed, 186 added.**

An earlier revision of this PR introduced a shared `Http/Events/Base` that all five extended. A reviewer correctly pointed out that AGENTS.md forbids exactly that, by name, at line 210: "Keep inheritance flat — GitHub/Gitea/GitLab/Bitbucket event handlers duplicate similar private methods rather than a deep `Events/Base`." Line 209 also holds up `handlePushEvent()` in the GitHub endpoint as the model for the preferred mechanism, and that base deleted it. The base is gone; this is the rewrite.

## What moved, and why it qualifies

Each provider keeps its own `action()` and its own private `handlePushEvent` / `handlePullRequestEvent`, so the action still reads like the story of the request. Only two things move, both to `src/Appwrite/Vcs/`, which is AGENTS.md preference 3 and already holds `Factory`, `InstallationTokens`, `RepositoryPullRequestCleanup`, `RepositoryWebhooks` and `Comment`.

- **`Vcs\InstallationCleanup`** — detaching everything an installation held once it is removed at the provider. This was 104 lines duplicated between the GitHub and Origin endpoints, differing in three lines: two quote styles and the provider literal. The region check inside it was byte-identical and `private` in both, so no amount of subclassing could have shared it.
- **`InstallationTokens::adapterForRepository()`** — resolving a repository's installation to a token-refreshed adapter. This was duplicated across GitLab, Gitea and Bitbucket. It lands on the class that already owns token refresh rather than in a new one.

Orchestration deliberately stays duplicated. Origin is the evidence that these flows genuinely diverge: it verifies Ed25519 signatures against published keys rather than a shared secret, deduplicates deliveries by id, and names events by action.

## Behaviour

Gitea and GitHub each carried a hand-rolled copy of the pull request cleanup that `RepositoryPullRequestCleanup` already implements. Both now call it.

- **Gitea**: its copy was equivalent, so nothing changes.
- **GitHub**: its copy was not. It stripped the pull request id from every repository sharing that provider repository id, regardless of which provider it belonged to. Provider repository ids are scoped per provider server, so the same numeric id can name a repository elsewhere. Going through the shared class adds the provider guard, which is a fix, and raises that one lookup from the locked `utopia-php/database` default of 25 rows to 100.

Everything else is behaviour-preserving. GitHub's `action()` and `handlePushEvent` are byte-identical to `main`.

## Testing

- `composer analyze`, `composer lint` and `composer refactor:check` pass.
- `tests/unit/Vcs` passes (39 tests, 151 assertions).
- Every endpoint was instantiated and its injection list compared against its `action()` signature parameter by parameter.
- Against a locally rebuilt container: signed and unsigned deliveries across push, deleted branch, opened PR, non-external `synchronize`, closed fork PR, `installation` deleted and created, an unknown event type, and a bad signature; plus Origin for a missing signature, a bogus signature and a stale timestamp. No fatals.

`Vcs\InstallationCleanup` is now reachable from a unit test for the first time, since it was previously a private method on a concrete action. `tests/unit/Vcs/` already holds one test per domain class there. That test is not in this PR.


HarshMN2345:
Rebase note: needs a rebase past main's one-line `array_values` fix, plus the missing `InstallationCleanup` unit test. Reviewable after that. Also now carries the GitHub PR-cleanup delegation that #13200 was closed in favour of this PR for.

PR #13391, opened 2026-08-27
feat(auth): add Cloudflare Dashboard OAuth2 provider

imtia33:
## What does this PR do?

Adds Cloudflare as an OAuth2 login provider via **Cloudflare Dashboard OAuth2** — the consumer "Login with Cloudflare" flow at `dash.cloudflare.com`. Any Cloudflare user can consent to give an Appwrite app access to their Cloudflare account. Use cases include third-party apps that manage DNS, Workers, or other Cloudflare API resources on behalf of the user.

**This is NOT Cloudflare Zero Trust.** Zero Trust is per-tenant corporate SaaS OIDC; Dashboard OAuth2 is global consumer OIDC (Great naming by cloudflare, now I really wont trust anyone , specially if its their name or their service name).

### Endpoints

All global:

- **Authorize:** `https://dash.cloudflare.com/oauth2/auth`
- **Token:** `https://dash.cloudflare.com/oauth2/token`
- **UserInfo:** `https://dash.cloudflare.com/oauth2/userinfo`
- **Discovery:** `https://dash.cloudflare.com/.well-known/openid-configuration` (returns 200 with RFC 8414 metadata)

### Identity

Cloudflare's OIDC discovery doc advertises `claims_supported: ["sub"]` — only the `sub` claim is exposed via userinfo. No email, name, or picture. Cloudflare users in Appwrite will have empty email and empty name. Account deduplication is via `(provider='cloudflare', providerUid=<sub>)`. Supported by Appwrite thanks to PR #13255 (email optional for OAuth2 account linking).

### PKCE

Uses Appwrite's `PKCE` trait with `S256` challenge method — consistent with Etsy, Kick, X, and Appwrite providers. `verifyCredentials()` uses `client_secret_post` only (no Basic auth header).

### Setup

Admins register an OAuth client at `https://dash.cloudflare.com` → Manage Account → OAuth clients → Create client. Select `authorization_code` grant, `client_secret_post` auth method, redirect URI `https://<appwrite-host>/v1/account/oauth2/callback/cloudflare`, and at least the `openid` scope. To allow any Cloudflare user (not just parent-account members) to authorize, promote the client to public visibility — requires client name, logo, client URL, scopes, AND client URL domain ownership verification (DNS TXT). Public promotion is permanent.

Reference: https://developers.cloudflare.com/fundamentals/oauth/

## Test Plan

[...]

PR #13526, opened 2026-09-08
feat(auth): add TikTok and Kakao OAuth2 providers

Meldiron:
## What does this PR do?

Adds two new OAuth2 sign-in providers: **TikTok** (Login Kit) and **Kakao** (Kakao Login).

Both follow the existing provider shape: an adapter under `src/Appwrite/Auth/OAuth2/`, an `Update` action under `Project/Http/Project/OAuth2/`, a response model, and registration in `oAuthProviders.php`, `models.php`, `Response.php`, `OAuth2ProviderList`, `Base::getProviderActions()`, `Get` and `Project\Services\Http`.

### TikTok

- Authorize `https://www.tiktok.com/v2/auth/authorize/`, token `https://open.tiktokapis.com/v2/oauth/token/`, user info `https://open.tiktokapis.com/v2/user/info/`.
- TikTok deviates from OAuth2 naming and calls the client identifier `client_key`, so the console field is labelled "Client Key" while the API field stays `clientId`.
- Scopes are **comma**-separated. `user.info.basic` is the narrowest scope that returns an identity, so it is the only default.
- The user info endpoint returns nothing unless the requested fields are named explicitly, so `open_id,union_id,display_name,avatar_url` are always requested.
- Identity is the app-scoped `open_id`. Login Kit exposes **no email**, so `getUserEmail()` returns `''` and `isEmailVerified()` is `false` (same as Resend).
- A rejected token request comes back with an `error` body rather than only an HTTP error status, so the response is checked explicitly instead of relying on `request()` alone.
- `verifyCredentials()` is not implemented — TikTok does not document an error code for bad client credentials, so there is nothing reliable to assert against.

### Kakao

- Authorize `https://kauth.kakao.com/oauth/authorize`, token `https://kauth.kakao.com/oauth/token`, user info `https://kapi.kakao.com/v2/user/me`.
- The client ID is the app's **REST API key**; the secret is the one generated under Kakao Login > Security, which must be set to enabled since Appwrite always sends `client_secret`. Both are called out in the field labels/hints.
- **No scope is requested by default.** Kakao drives consent from the items configured on the app and rejects the authorization with KOE205 as soon as a scope is not activated there. Extra consent items can still be passed per request via the existing `scopes` query param, comma-separated.
- Kakao reports both `is_email_verified` and `is_email_valid`; an address is only treated as verified when both hold and the email consent item was granted.
- `verifyCredentials()` probes the token endpoint and rejects on `invalid_client` (KOE010), matching Hugging Face / Resend / Cloudflare.

## Test Plan

- New unit tests `tests/unit/Auth/OAuth2/{TikTokTest,KakaoTest}.php` cover the login URL shape (including `client_key` vs `client_id` and comma-separated scopes), the token and refresh request shapes, refresh-token retention, provider errors, and claim mapping. `docker compose exec appwrite test tests/unit/Auth/OAuth2/` → 71 tests, 441 assertions, green.
- Full unit suite green (1179 tests).
- `tests/e2e/Services/Project/OAuth2ConsoleClientTest.php` and `OAuth2CustomServerTest.php` green (26 tests each).
- Verified live on a local stack: `PATCH /v1/project/oauth2/{tiktok,kakao}` are routed, and both providers appear on `GET /v1/console/oauth2-providers` with the expected field names and examples.
- `composer lint`, `phpstan --level 4` on the new files, and `composer refactor:check` all pass. Spec generation produces `projectUpdateOAuth2TikTok` / `projectUpdateOAuth2Kakao` and the `oAuth2TikTok` / `oAuth2Kakao` models.

## Related PRs and Issues

Follows the same shape as the Cloudflare (#13391) and Resend (#13366) provider additions.

## Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

Yes.

## Note for the console

Both providers use `icon-tiktok` / `icon-kakao`; the console needs matching icons before they render.


Meldiron (commented):
lgtm

PR #13530, opened 2026-09-08
fix(vcs): reindex provider pull request ids after removing one

### LINKED ISSUE #11741: Missing array_values() after array_diff() in VCS GitHub PR cleanup causes malformed array storage

HarshMN2345:
Closes #11741

### What does this PR do?

`array_diff()` filters by value but keeps the original keys, so untracking a pull request that is not the last entry leaves a gapped array.

The consequence is worse than malformed storage. `providerPullRequestIds` is declared `array => true` in `app/config/collections/platform.php`, and the structure validator rejects any value that is not a list:

```php
if (!\is_array($value) || !\array_is_list($value)) {
    $this->message = 'Attribute "'.$key.'" must be an array';
```

So closing an external contribution whose id sits anywhere but at the end throws, the webhook 500s, and the pull request is never untracked. Closing the last entry happens to work, which is why this went unnoticed.

Three call sites persist the result and all three are fixed:

- `VCS/Http/GitHub/Events/Create.php` — the one reported
- `VCS/Http/Gitea/Events/Create.php`
- `Vcs/RepositoryPullRequestCleanup.php` — shared by the GitLab and Bitbucket handlers

### Same bug elsewhere

Sweeping every key-preserving filter whose result is persisted or serialized turned up one more, in a different service: `Teams/Http/Memberships/XList.php` drops memberships with an empty `userId` through `array_filter()`, which also preserves keys. `Response::dynamic()` writes each entry back to its *original* index rather than appending, so the gap survives into the response and a member with an empty `userId` in any but the last position turns the `memberships` list into a JSON object — breaking list deserialization in every SDK.

Call sites where the filtered value is only used locally (`in_array`, `empty`, `count`, iteration) were checked and left alone.

### Test Plan

- `tests/unit/Vcs/RepositoryPullRequestCleanupTest::testRemoveReindexesRemainingPullRequestIds` removes a non-tail ID and asserts the persisted value is `[2, 3]`.
- `tests/e2e/Services/VCSGitea/VCSGiteaConsoleClientTest::testClosePullRequestRemovesAuthorization` creates two external pull requests through the local Gitea API and authorizes both through Appwrite. It closes and reopens the first, checks that it can be authorized again, and verifies the second remains authorized (duplicate authorization returns 409).

Validated locally against the isolated Gitea fixture and real Appwrite API: the E2E passes with this fix and fails with HTTP 409 on reauthorization when only the Gitea `array_values()` change is removed. Existing cleanup unit tests, Pint, and the focused Rector dry run also pass.


Meldiron (approved):
lgtm, lets see if we can add e2e test too

Issue #5716, opened 2023-06-16
🚀 Feature: Customize default function "example data"


byawitz:
### 🔖 Feature description

Add the option to set function default example data.

When executing a function from the console, there's the option `show example JSON`, I think it might my be nice to have this option to fetch some basic JSON the user added to the deployed function.

So for example the `appwrite.json` will look something like this.
```json
{
  "projectId"  : "someProject",
  "projectName": "General",
  "functions"  : [
    {
      "$id"       : "646b67675406f6ae4388",
      "name"      : "My Awesome Function",
      "runtime"   : "node-16.0",
      "path"      : "functions/My Awesome Function",
      "entrypoint": "src/index.js",
      "ignore"    : [
        "node_modules",
        ".npm"
      ],
      "execute"   : [],
      "events"    : [],
      "example"   : {
        "userName": "John",
        "password": "world",
        "age"     : 26
      },
      "schedule"  : "",
      "timeout"   : 15
    }
  ]
}
```

Then, the console could load that example - if it exists - and let the user copy it.

![image](https://github.com/appwrite/appwrite/assets/316103/7f12f40b-3b56-4cf0-ad22-2e30b7d05e83)

### 🎤 Pitch

Many functions have some basic logic that can work many times. 
This feature will allow fast debugging of a function.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

joeyouss:
Thank you for opening this, it would be cool to have this.

eldadfux:
Stale issue, closing down.

PR #13544, opened 2026-09-08
fix(specs): serialize empty object defaults as objects

ChiragAgg5k:
## Summary

Fix object-typed OpenAPI defaults being serialized as `[]` instead of `{}`.

Found during genuine Cloud spec verification for [CLO-4377](https://linear.app/appwrite/issue/CLO-4377/openapi3-standards), tracked in [Cloud #5716](https://github.com/appwrite-labs/cloud/pull/5716). This PR is independent of the optional-security migration in #13543 and branches directly from main.

### Cause and fix

The object-specific formatter branches already normalize empty defaults to `stdClass`, but the shared default assignment later overwrites the normalized value with the route's raw PHP `[]`. JSON then contains `type: object` with `default: []`, which fails strict OpenAPI validation.

Normalize an exactly empty array to `stdClass` at that shared assignment only when the final schema type is `object`. Preserve default eligibility, array defaults, populated maps, scalar defaults, examples, and runtime API behavior.

The captured Cloud document has 16 affected schemas, including screenshot headers, execution headers, presence metadata, and document/row update data. The published document has the same validation error; this was not introduced by the optional-security change.

## Verification

- Added GET-query and POST-request-body regressions that inspect the formatter's serialized JSON. Both failed before the production fix (`array` instead of `stdClass`) and pass afterward.
- Coverage verifies empty Assoc/JSON object defaults serialize as `{}`, while populated maps, empty array schemas, `0`, and `false` remain unchanged.
- `vendor/bin/phpunit tests/unit/SDK/Specification/FormatTest.php` — passed: **38 tests, 210 assertions**.
- `vendor/bin/pint --test src/Appwrite/SDK/Specification/Format/OpenAPI3.php tests/unit/SDK/Specification/FormatTest.php` — passed.
- `composer refactor:check` — passed across all 397 test files. The data provider now uses `yield`, fixing the failing CI Rector check.
- `git diff --check` — passed.
- Removed the redundant concrete-type assertion requested by Greptile; the regression asserts emitted `{}` JSON directly.

Not run for this separate branch: full Cloud specs workflow, full-document strict validation, SDK regeneration/output comparisons, Appwrite runtime e2e, or full-project static analysis. This fixes the demonstrated object-default mismatch; it does not claim that the entire live spec is now validator-clean. No screenshots apply to this nonvisual formatter change.

### DIFF
--- a/src/Appwrite/SDK/Specification/Format/OpenAPI3.php
+++ b/src/Appwrite/SDK/Specification/Format/OpenAPI3.php
@@ -975,7 +975,11 @@ class OpenAPI3 extends Format
                 }
 
                 if ($parameter['emitDefault'] && $this->shouldEmitDefaultForSchema($param['default'], $node['schema'])) { // Param has default value
-                    $node['schema']['default'] = $param['default'];
+                    // PHP uses [] for empty maps too; preserve the declared
+                    // object type when serializing its default to JSON.
+                    $node['schema']['default'] = $node['schema']['type'] === 'object' && $param['default'] === []
+                        ? new \stdClass()
+                        : $param['default'];
                 }
 
                 $pathAliases = [$name, ...($param['aliases'] ?? [])];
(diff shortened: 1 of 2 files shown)


ChiragAgg5k:
@greptile review

loks0n (approved):

PR #13557, opened 2026-09-09
fix: reuse the cloned branch when pushing templates

HarshMN2345:
Template deployments clone the destination branch and then try to create the same local branch with `git checkout -b`. The push fails with "a branch named ... already exists" before the template can be committed.

Use `git checkout -B` at the cloned HEAD so both an existing branch checkout and a detached commit checkout can receive the template commit. This only changes the temporary build clone; the push remains a normal fast-forward push.

Validation: local bare-Git regression scenarios for branch and detached-HEAD clones both push the template successfully while preserving the original file and parent commit. The previous command reproduces the existing-branch failure. Targeted Pint, PHPStan, PHP syntax, and `git diff --check` pass. No production VCS services were called; a complete build-worker/API flow was not run.

Fixes [CLOUD-3QX7](https://appwrite.sentry.io/issues/CLOUD-3QX7).

### DIFF
--- a/src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
+++ b/src/Appwrite/Platform/Modules/Functions/Workers/Builds.php
@@ -358,7 +358,9 @@ class Builds extends Action
 
                 // Commit and push
                 $commitMessage = \escapeshellarg('Create ' . $resource->getAttribute('name', '') . ' function');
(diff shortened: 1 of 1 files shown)


HarshMN2345:
Rebase note: needs a rebase onto main's `Command` refactor — the fix reduces to flipping `Builds.php:370` from `flag('-b')` to `'-B'`. The template push is still broken on main.

PR #13568, opened 2026-09-09
fix(auth): let developers choose Resend OAuth2 scope

Meldiron:
## What does this PR do?

Removes the hardcoded `emails:send` default scope from the Resend OAuth2 adapter. Resend offers only `emails:send` and `full_access`, and neither is identity-only, so the adapter should not pick one on the developer's behalf. Developers now request whichever scope they need through the provider's custom scopes; with none configured, the login URL sends an empty `scope`, matching other adapters with no default scopes.

Follow-up to #13366.

## Test Plan

- `composer lint src/Appwrite/Auth/OAuth2/Resend.php` passes
- Custom scopes still append through the `OAuth2` base constructor, so configuring `emails:send` or `full_access` on the provider produces the same login URL as before

## Related PRs and Issues

- #13366

## Checklist

### DIFF
--- a/src/Appwrite/Auth/OAuth2/Resend.php
+++ b/src/Appwrite/Auth/OAuth2/Resend.php
@@ -29,15 +29,13 @@ class Resend extends OAuth2
     protected array $claims = [];
 
     /**
-     * Resend has no identity-only scope; `emails:send` is the narrowest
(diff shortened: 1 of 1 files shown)


Meldiron (commented):
lgtm

PR #13588, opened 2026-09-09
fix(queue): pass Redis DSN credentials to the publisher pool

### LINKED ISSUE #13554: Publisher pool cannot authenticate to password-protected Redis (NOAUTH in task-scheduler) — credentials dropped in registers.php and never used by utopia-php/queue
## Summary

On Appwrite 2.0.0 (self-hosted) with an **external Redis that requires a password**, every publish through the `publisher` pool fails with `NOAUTH Authentication required`. Observed via the task-scheduler's hourly stats-resources task. The credentials are present in the container environment and are correct (verified directly against the Redis server); they are simply never passed to (or used by) the queue connection.

## Environment

- Appwrite 2.0.0 (self-hosted, Docker Compose)
- Redis 7.4.7 external (own host, `requirepass` set)
- `_APP_REDIS_HOST`, `_APP_REDIS_PORT`, `_APP_REDIS_USER`, `_APP_REDIS_PASS` set on all containers (verified inside the container with `docker exec ... env | grep _APP_REDIS`)

## Observed behavior

`appwrite-task-scheduler`, every hour:

```
[StatsResources] Failed to publish stats resources message: NOAUTH Authentication required.
stats_resources_task · 207ms
  stats_resources_task.projects_queued 0
  stats_resources_task.projects_failed 1
```

Deterministic — persists across full stack restarts.

**Additionally**: the combined worker's queue *consumers* borrow the same publisher pool (`app/worker.php` line ~100: `BrokerPool(publisher: $publisher, consumer: $publisher)`), so on passworded Redis, messages pile up in the queue unconsumed as well — verify with `LLEN utopia-queue.queue.v1-stats-resources`.

Everything else that uses Redis works: realtime pub/sub authenticates and serves, the cache pool works, and queue produce/consume for table attribute creation works end to end when Redis has no password.

## Root cause — two layers

**Layer 1: `app/init/registers.php` (2.0.0) — the `publisher` pool drops the DSN credentials.**
The fallback Redis DSN is built *with* user and pass:

```php
$fallbackForRedis = 'redis_main=' . AppwriteURL::unparse([
    'scheme' => 'redis',
    'host' => System::getEnv('_APP_REDIS_HOST', 'redis'),
    'port' => System::getEnv('_APP_REDIS_PORT', '6379'),
    'user' => System::getEnv('_APP_REDIS_USER', ''),
    'pass' => System::getEnv('_APP_REDIS_PASS', ''),
]);
```

…but the publisher adapter constructs the connection with **host and port only**:

```php
case 'publisher':
    return match ($dsn->getScheme()) {
        'redis' => (function () use ($dsn) {
            $connection = new Queue\Connection\Redis($dsn->getHost(), $dsn->getPort());
            return new Queue\Broker\Redis($connection, $connection);
        })(),
        default => null
    };
```

All other Redis pools (`cache`, `pubsub`, `lock`) go through the shared `$resource()` factory, which does `pconnect` + `$redis->auth($dsnPass)` — which is why those paths authenticate correctly and only the publisher fails.

**Layer 2: `utopia-php/queue` `Connection\Redis` accepts `$user`/`$password` but never authenticates.**
Current `main` of utopia-php/queue, `src/Queue/Connection/Redis.php`:

```php
public function __construct(protected string $host, protected int $port = 6379,
    protected ?string $user = null, protected ?string $password = null,
    protected float $connectTimeout = -1, protected float $readTimeout = -1) {}

protected function getRedis(): \Redis
{
    // ...
    $redis = new \Redis();
    $redis->connect($this->host, $this->port, $connectTimeout);
    if ($this->readTimeout >= 0) {
        $redis->setOption(\Redis::OPT_READ_TIMEOUT, $this->readTimeout);
    }
    $this->redis = $redis;
    return $this->redis;
    // ...
}
```

`$this->user` and `$this->password` are never read; there is no `auth()` call anywhere in the class. Even if Appwrite passed them, the connection would still not authenticate.

## Evidence it is not a configuration problem

1. Credentials verified inside the running container (`docker exec env`): `_APP_REDIS_PASS`, `_APP_REDIS_USER`, `_APP_REDIS_HOST`, `_APP_REDIS_PORT` all present and correct.
2. Direct protocol test from the same host: `AUTH default <pass>` → `+OK`; `PING` → `+PONG`; `LPUSH`/`LPOP` succeed. Two-argument AUTH form (used when `_APP_REDIS_USER` is set) also returns `+OK`.
3. Realtime, cache, and lock pools authenticate against the same server with the same env — only the publisher pool fails.
4. Survives full stack restarts; failure is deterministic (the code path never attempts auth).
5. With a passwordless Redis, everything works — confirming the bug only manifests with authentication enabled.

## Why default installs don't see this

The official docker-compose Redis service runs without `requirepass` and `_APP_REDIS_PASS` defaults to empty — the publisher's missing auth is invisible. The bug only manifests for self-hosters pointing at password-protected external Redis.

## Suggested fix

**utopia-php/queue** — in `getRedis()`, after `connect()`:

```php
if ($this->password !== null && $this->password !== '') {
    $this->user
        ? $redis->auth([$this->user, $this->password])
        : $redis->auth($this->password);
}
```

**appwrite** — in `registers.php`, pass the DSN credentials to the publisher connection:

```php
$connection = new Queue\Connection\Redis(
    $dsn->getHost(),
    $dsn->getPort(),
    $dsn->getUser() ?? null,
    $dsn->getPassword() ?? null
);
```

## Impact

- Hourly resource gauges (`v1-stats-resources`) are not published — Usage tab resource stats missing.
- On passworded Redis, ALL queue messages published by the scheduler pile up unconsumed (scheduled functions, stats, etc.) since the worker's consumer also borrows the publisher pool.
- Any other producer using the `publisher` pool against passworded Redis is affected the same way.

## Related

- #3309 (2022) — same bug class: NOAUTH `RedisException` in entrypoints that skipped auth, fixed back then by adding auth calls.

---
*Filed from a live deployment: Appwrite 2.0.0 on Coolify with external PostgreSQL/Redis/ClickHouse. Verified fix working in production: queue depth 2 → 0, gauges 0 → 210 after patching.*

breken-ai:
Fixes #13554.

## Root cause

The shared Redis DSN (`app/init/registers.php:51-57`) carries `_APP_REDIS_USER`/`_APP_REDIS_PASS`, and the sibling pools authenticate with them (`$redis->auth($dsnPass)` in the default resource factory). The publisher pool, however, constructs its queue connection with host and port only - so against a password-protected Redis (`requirepass` / ACL), every publish fails with `NOAUTH`. Because the combined worker borrows the publisher pool for its consumer (`worker.php`, `BrokerPool(publisher, consumer: publisher)`), queue messages pile up too, which matches the reporter's task-scheduler symptoms exactly.

One layer down, `utopia-php/queue`'s `Connection\Redis` accepts `?user`/`?password` in its constructor but never calls `auth()` - the library half is fixed in utopia-php/monorepo#247 (composer pin `^2.0.0` admits the patch release).

## Fix

The publisher pool passes the DSN credentials through (`$dsn->getUser() ?: null`, `$dsn->getPassword() ?: null` - the getters are already used a few lines up). Inert-but-correct against current `utopia-php/queue` releases (the constructor ignores them today); live as soon as utopia-php/monorepo#247 ships.

## Verification

DSN construction, the authenticating sibling factory, the publisher pool construction, the `worker.php` pool borrowing, and the library constructor signature all confirmed in source at `main` (f31509d). **Not run:** no PHP runtime or live password-protected Redis in the author's environment - external red proof is the reporter's hourly `NOAUTH` log with every other pool authenticating.

> Built by breken, your AI support engineer - breken.ai - this one's on us.

### DIFF
--- a/app/init/registers.php
+++ b/app/init/registers.php
@@ -252,7 +252,14 @@ $register->set('pools', function () {
                         // Publishers never block on receive, so one connection backs both broker slots.
                         return match ($dsn->getScheme()) {
                             'redis' => (function () use ($dsn) {
-                                $connection = new Queue\Connection\Redis($dsn->getHost(), $dsn->getPort());
+                                // Pass the DSN credentials through: without them every
+                                // publish fails with NOAUTH on password-protected Redis.
+                                $connection = new Queue\Connection\Redis(
+                                    $dsn->getHost(),
+                                    $dsn->getPort(),
+                                    $dsn->getUser() ?: null,
+                                    $dsn->getPassword() ?: null,
+                                );
                                 return new Queue\Broker\Redis($connection, $connection);
                             })(),
                             default => null


breken-ai:
Closing this one out - the fix landed upstream in #13622 (with utopia-php/monorepo#260 on the queue side). @ChiragAgg5k's version covers the same ground plus the DSN encoding and the direct \Redis auth paths, so there's nothing left to add here. Nice fix.

PR #13589, opened 2026-09-10
fix(functions): persist providerRootDirectory on VCS deployments and refresh it on redeploy

### LINKED ISSUE #13587: 🐛 Bug Report: Redeploy always fails for monorepo functions — duplicate reads providerRootDirectory from the deployment, where nothing ever sets it
### 👟 Reproduction steps

1. Self-hosted Appwrite `2.0.0`. Connect a function to a Git repository where the function's code lives in a **subdirectory**, i.e. set `providerRootDirectory` (the standard monorepo layout — ours is `appwrite-functions/<name>`).
2. Push a commit. The build succeeds. ✅
3. Change any function setting (we added an environment variable). The Console shows *"Settings changes are not live yet — You've updated function settings, but they won't take effect until you redeploy."* with a **Redeploy** button.
4. Click **Redeploy** (or `POST /v1/functions/{functionId}/deployments/duplicate`).
5. The build fails after **1 second**.

Build log (Go function; the failing line differs per runtime, see below):

```
[22:53:41] [open-runtimes] Build cache hit.
[22:53:41] [open-runtimes] Environment preparation started.
[22:53:41] [open-runtimes] Build command execution started.
bash: line 1: /usr/local/build/vendor/alimfreight.com/shared/go.mod: No such file or directory
[22:53:41] [open-runtimes] Environment preparation finished.
Build failed with exit code 1.
```

The path exists in the repository at `appwrite-functions/arrival-nudger/vendor/…`. It is missing from `/usr/local/build` because **the build ran against the repository root instead of the function's root directory**.

### 👍 Expected behavior

Redeploy rebuilds the same source the original deployment built — from `providerRootDirectory` — and succeeds exactly as a push-triggered build does.

### 👎 Actual behavior

Redeploy clones the **repository root**, so the build runs in the wrong directory and fails. For a Go function this shows as a missing vendored path; other runtimes would fail on a missing `package.json`, `requirements.txt`, entrypoint, etc.

Consequence: on a monorepo, **the Redeploy button can never succeed**, which also means the "Settings changes are not live yet" banner cannot be cleared by the button the Console offers for exactly that purpose. **20 of our 21 functions** are affected (every one whose code is not at the repo root).

### 🎯 Root cause

`src/Appwrite/Platform/Modules/Functions/Http/Deployments/Duplicate/Create.php` takes the root directory from the **deployment**:

```php
$github->generateCloneCommand(
    …
    $deployment->getAttribute('providerRootDirectory', ''),
);
```

…but **nothing in 2.0 writes `providerRootDirectory` onto a deployment.** Grepping the codebase, it is only *read* (here, and for the `APPWRITE_VCS_ROOT_DIRECTORY` runtime variable in `src/Appwrite/Platform/Workers/Functions.php:676` and `src/Appwrite/Deployment/Deployments.php:718`), *cleared* (`Deletes.php`), or *migrated* (`V19.php`). On our instance **94 of 128 VCS deployments have it `NULL`** — every one created since we moved to 2.0.

Push-triggered builds are unaffected because `Builds.php` reads it from the **function**:

```php
$rootDirectory = $resource->getAttribute('providerRootDirectory', '');
```

And an empty value is not neutral — `generateCloneCommand` in `utopia-php/vcs` (`GitHub.php`) turns it into a wildcard:

```php
if ($rootDirectory === '' || $rootDirectory === '0') {
    $rootDirectory = '*';
}
…
'git config core.sparseCheckout true',
"echo {$rootDirectory} >> .git/info/sparse-checkout",
```

`*` sparse-checks-out the whole repository, so `/usr/local/build` becomes the repo root.

### 🔬 Things I verified, to save you time

- **Backfilling the field on the source deployment does not help.** I set `providerRootDirectory = 'appwrite-functions/arrival-nudger'` on the ready deployment and clicked Redeploy again: the duplicate came out with `providerRootDirectory = NULL` again and failed identically. So the duplicate path loses it as well as never having it.
- **`POST /v1/functions/{functionId}/deployments/vcs`** (`type=branch, reference=main, activate=true`) **works**: built in 4 s, 2.89 MB, activated, and the function's `live` flag flipped to `true`, clearing the banner. That endpoint is the usable workaround — "Create deployment → from Git" rather than "Redeploy".
- The failed deployments are harmless to delete; the active deployment id was unchanged by deleting them (2.0 no longer clears it, unlike 1.9.x).

### 💡 Suggested fix

In `Duplicate/Create.php`, fall back to the function's value the way `Builds.php` already does:

```php
$rootDirectory = $deployment->getAttribute('providerRootDirectory', '')
    ?: $function->getAttribute('providerRootDirectory', '');
```

Better still, persist `providerRootDirectory` onto the deployment when it is created from VCS, since two other code paths already expect to read it there (the `APPWRITE_VCS_ROOT_DIRECTORY` runtime variable is currently empty in every runtime for the same reason).

### 🧱 Your Environment

- Appwrite **2.0.0** self-hosted (Docker Compose), PostgreSQL 18.3 (`appwrite/postgres:0.1.0`)
- Executor `openruntimes/executor:0.29.0`, orchestrator `ghcr.io/open-runtimes/orchestrator:1.9.2`
- Runtime `go-1.26`; monorepo, one `providerRootDirectory` per function, 20 of 21 functions in subdirectories
- GitHub App installation, branch `main`

### 📎 Additional information

Same instance as #13586 (also a 2.0 VCS-source regression, different symptom). Happy to test a patch.

breken-ai:
Fixes #13587.

## Root cause (the reporter's, confirmed at HEAD)

@AlimFreight traced this end to end; verified independently here:

1. `Deployments/Duplicate/Create.php` refreshes `entrypoint`/`buildCommands`/`startCommand` from the function on redeploy but inherits `providerRootDirectory` from the old deployment.
2. Nothing in 2.0 writes `providerRootDirectory` onto VCS deployments - only the template flow (`Deployments/Template/Create.php:184`) does - so the duplicate gets NULL.
3. `utopia-php/vcs` (`Adapter/Git/GitHub.php`, `generateCloneCommand`): an empty root directory becomes a `*` sparse checkout, so the whole repo is checked out, the build runs at the repository root, and monorepo function code is missing - the ~1s build failure in the reporter's log.
4. The same missing write also empties `APPWRITE_VCS_ROOT_DIRECTORY` at runtime (`Executions/Create.php` and the build env in `Deployments.php:715` read it from the deployment).

## Fix (2 files, +13)

- **Duplicate/Create.php**: refresh `providerRootDirectory` from the function, same treatment as `entrypoint`/`commands` above it. A redeploy rebuilds with the function's current root directory - and because it overwrites the inherited NULL, this also heals every existing broken deployment on its next redeploy, no backfill needed (which matches the reporter's finding that backfilling wasn't a path forward).
- **Deployments.php `submit()`**: snapshot `$source['subdir']` as `providerRootDirectory` on new VCS deployments so the value persists going forward. The upload flow (`submit(null)`) is untouched.

## Verification

Every step of the chain re-traced in source at `main` (f31509d), including the `'*'` wildcard in the vcs library and all read sites. **Not run:** no PHP runtime or live Appwrite instance in the author's environment, so no lint or live redeploy test - the reporter offered to test a Redeploy on a subdirectory function, which would be the real-world green proof. (Their `POST /deployments/vcs` workaround is correct until this lands.)

> Built by breken, your AI support engineer - breken.ai - this one's on us.

### DIFF
--- a/src/Appwrite/Deployment/Deployments.php
+++ b/src/Appwrite/Deployment/Deployments.php
@@ -182,6 +182,13 @@ readonly class Deployments
 
     private function submit(Document $resource, Document $deployment, ?array $source): Document
     {
+        // Snapshot the source root onto the deployment: the Duplicate
+        // endpoint and the APPWRITE_VCS_ROOT_DIRECTORY runtime variable read
+        // it back from here, and nothing else writes it for VCS sources.
+        if (is_array($source) && array_key_exists('subdir', $source)) {
+            $deployment->setAttribute('providerRootDirectory', $source['subdir'] ?? '');
+        }
+
         // The caller may have been holding this deployment for a while (the
         // Builds worker pushes a template commit first), so its status is stale
         // by now — dropping it keeps upload() from writing a cancel away, and
--- a/src/Appwrite/Platform/Modules/Functions/Http/Deployments/Duplicate/Create.php
+++ b/src/Appwrite/Platform/Modules/Functions/Http/Deployments/Duplicate/Create.php
@@ -135,6 +135,12 @@ class Create extends Action
             'entrypoint' => $function->getAttribute('entrypoint'),
             'buildCommands' => $function->getAttribute('commands', ''),
             'startCommand' => $function->getAttribute('startCommand', ''),
+            // Same treatment as entrypoint/commands above: a redeploy rebuilds
+            // with the function's current configuration. Deployments created
+            // before this attribute was persisted carry NULL, which the VCS
+            // clone path turns into a '*' wildcard - the build then runs at
+            // the repository root and monorepo functions fail.
+            'providerRootDirectory' => $function->getAttribute('providerRootDirectory', ''),
             'buildStartedAt' => null,
             'buildEndedAt' => null,
             'buildDuration' => null,


breken-ai:
Closing as superseded - fixed upstream in #13622 and utopia-php/monorepo#260. @ChiragAgg5k's change covers this ground, so this branch is no longer needed.

PR #13591, opened 2026-09-10
fix(mongo): survive losing the first-boot bind race to the init server

abnegate:
On a fresh volume the standard mongo entrypoint runs a temporary server on `127.0.0.1:27017` to create the users and run `docker-entrypoint-initdb.d`, then starts the real server on `0.0.0.0:27017` **without waiting for the temporary one to release the port**. From a failing run:

```
57.883  real server starts          <- before the temporary one is even asked to stop
57.891  temporary server begins shutdown
59.009  temporary server "Now exiting"
59.033  real server: Error setting up transport layer ...
        0.0.0.0:27017 :: caused by :: setup bind :: caused by :: Address already in use
```

The real server spends about a second opening WiredTiger before it binds, so it only ever wins by however long the temporary server's shutdown takes.

| | temporary server shutdown | outcome |
|---|---|---|
| idle machine | 127 ms | binds with ~1 s to spare |
| loaded CI runner | 1126 ms | bind attempted at 1150 ms, **lost by 24 ms** |

When it loses, the container exits with it and `docker compose up --wait` fails every service that `depends_on` mongodb. The lane then reports

```
dependency failed to start: container appwrite-mongodb is unhealthy
```

having run no tests at all. Measured on `feat-query-lib`: it took out one of the 191 lanes in **four of the last five runs**, a different lane each time (Locale, Teams, Migrations, Proxy). It is not specific to that branch — nothing in it touches this file.

## The fix

Start again rather than retry blindly. The data directory is fully initialised by the time this fires — users created, init scripts run, temporary server cleanly shut down — so only the final server start failed. The second pass finds the directory populated, skips the temporary server entirely, and binds once the port is free.

The steady state is untouched: an initialised data directory still hands straight over with `exec`, so `mongod` keeps PID 1 and its signal handling on every boot after the first. A `trap` covers the single first-boot window where it runs as a child, so `docker stop` still shuts the database down cleanly.

## Verification

The 24 ms window would not reproduce on an idle arm64 host — twelve cold starts all passed, and constraining CPU did not move it. So the failure condition it produces was forced instead, by holding `27017` across the real server's bind on a first boot. Counting only the runs where the conflict actually fired:

| entrypoint | conflict hit | retry taken | container | server reachable |
|---|---|---|---|---|
| before | 1 | 0 | exited 133 | no |
| after | 1 | 1 | running | yes |

Note this reproduces the failure condition rather than the race itself; the real confirmation is CI staying clean across runs.

### DIFF
--- a/mongo-entrypoint.sh
+++ b/mongo-entrypoint.sh
@@ -37,20 +37,50 @@ fi
 # server took 1126ms to exit after the real one had already started, losing the
 # bind by 24ms; the same sequence takes 127ms on an idle machine and passes.
 #
-# The data directory is fully initialised by the time this happens, so starting
-# again is enough: the second pass finds it populated, skips the temporary
-# server, and binds once the port is free.
+# Run it as a child so that this one outcome can be retried.
 docker-entrypoint.sh "$@" &
 server=$!
-trap 'kill -TERM "$server" 2>/dev/null || true' TERM INT
-wait "$server" || true
+
+signal=0
+trap 'signal=15; kill -s TERM "$server" 2>/dev/null || true' TERM
+trap 'signal=2; kill -s INT "$server" 2>/dev/null || true' INT
+
+status=0
+wait "$server" || status=$?
+
+# A trapped signal interrupts `wait` and returns 128+signal without reaping the
+# child, so wait again: mongod has to finish its own shutdown before this
+# process, which is PID 1, leaves and takes the container with it.
+while kill -0 "$server" 2>/dev/null; do
+  status=0
+  wait "$server" || status=$?
+done
+
 trap - TERM INT
 
-echo "mongodb: first start did not survive initialisation, starting again" >&2
+# Asked to stop. Report it the way a process killed by the signal does instead
+# of starting the server the caller just asked to go away.
+if [ "$signal" -ne 0 ]; then
+  exit $((128 + signal))
+fi
 
+# 48 is EXIT_NET_ERROR, what mongod exits with when it cannot bind, and the
+# standard entrypoint ends in `exec "$@"` so it arrives unchanged. A temporary
+# server that loses the bind instead fails inside `mongod --fork`, whose parent
+# reports 1. Every other status has to be raised rather than retried, an
+# initialisation script that failed once the storage files existed most of all:
+# a second pass would find /data/db populated, skip /docker-entrypoint-initdb.d
+# and serve a database with no application user, hiding why.
+if [ "$status" -ne 48 ]; then
+  exit "$status"
+fi
+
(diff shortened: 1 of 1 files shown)


abnegate:
Both P1s were valid and are fixed in `87449f3d06`. Recording it here because the push moved the anchored lines, so both threads auto-resolved as outdated rather than being answered.

**Shutdown Restarts MongoDB.** Correct, and worse than a lingering container: the script started the server the caller had just asked to go away. The trap now records which signal arrived, forwards it, and — because a trapped signal makes `wait` return `128+signal` *without reaping* — waits again until the child is actually gone, so mongod finishes its own shutdown before PID 1 leaves and takes the container with it. It then exits `128+signal` instead of starting anything.

Measured on `mongo:8.2.5`, `docker stop` both mid-initialisation and once the server was already serving:

| | before | after |
|---|---|---|
| stop duration | 30s (Docker's grace) | 0s |
| exit code | 137 (SIGKILL) | 143 (SIGTERM) |

**All Failures Trigger Retry.** Also correct, and the more dangerous of the two. An initialisation script failing after the storage files existed left the second pass looking at a populated `/data/db`, so it skipped `/docker-entrypoint-initdb.d` entirely and served a database with **no application user**, with the original failure unreported. Measured: container running and reachable with the user absent, versus `exit 1` and no server now.

The retry is now gated on `48` only — mongod's `EXIT_NET_ERROR`, what it exits with when it cannot bind. The status arrives unchanged because the standard entrypoint ends in `exec "$@"`, and a temporary server that loses the bind fails inside `mongod --fork`, whose parent reports `1` — so only the real server can reach `48`. Every other status is raised.

**On the one red check:** `Tests / E2E / Antivirus (dedicated)` is not this change. `main` pins `appwrite/php-clamav` at `2.0.0`, which sizes every INSTREAM frame at the chunk length rather than the bytes read and loops on `!feof()`, so it sends roughly twice the file and desyncs the framing — the log here shows `INSTREAM: Size limit reached, (requested: 247036)`, a length read out of file bytes. ClamAV answers with an error, the bool API reports that as a detection, and Storage deletes the upload and returns 403. Fixed in php-clamav 2.0.2; that lane is 2 failures in 19 runs on `main` itself. Bumping it is a separate change, not one to bury in a MongoDB entrypoint PR.

abnegate:
Closing: superseded by https://github.com/appwrite/appwrite/pull/13593. Superseded by the author's own #13593 'fix(mongo): initialise the data directory off the port the server binds', merged to main on 2026-09-21T23:23Z. That fix eliminates the race (init on port 27018, resumable marker) instead of retrying after losing it, and it rewrote mongo-entrypoint.sh so this PR

PR #13602, opened 2026-09-10
fix(account): keep the current session through the OAuth2 token flow

Meldiron:
## What does this PR do?

`GET /v1/account/sessions/oauth2/:provider/redirect` deletes the caller's current session whenever a signed-in user completes an OAuth2 flow. For `createOAuth2Session` that is the intended "replace" behavior: the handler immediately creates the new session and sets its cookie, so the browser stays signed in.

`createOAuth2Token` runs through the same handler but creates no session there. The app is expected to exchange the returned `userId` + `secret` through `createSession` afterwards. Deleting the current session in that flow leaves the caller signed out in between: a signed-in user who connects a provider lands on the success URL as a guest, and any auth guard on that page bounces them to sign-in before the token can be exchanged. We hit this in the Console while building one-click SMTP setup with Resend, and had to add a guest-tolerant callback page that restores the session by hand.

This PR gives the token flow parity with the session flow regarding the caller's session:

- The current session is deleted only when the flow creates a replacement (`createOAuth2Session`).
- In the token flow the identity is still attached to the signed-in account, the existing session stays valid, and the later `createSession` adds a session next to it.
- `createOAuth2Token` docs mention the behavior.

## Test Plan

New e2e `testCreateOAuth2TokenKeepsCurrentSession` in `AccountCustomClientTest`:

1. Sign in through the mock provider with the session flow and keep the cookie.
2. Run the token flow with that cookie on every hop. The redirect lands on the success URL with `secret` + `userId` for the same user and sets no cookie.
3. `GET /account/sessions/current` with the original cookie still returns 200 and the same session id. Before this change it returned 401.
4. `POST /account/sessions/token` creates a second session; `GET /account/sessions` lists both.
5. Run the session flow again with the second cookie: the second session is replaced and its cookie now gets 401, so the existing behavior of `createOAuth2Session` is unchanged.

Existing OAuth2 e2e tests (`testCreateOAuth2AccountSession`, `testOAuth2TokenSessionProviderAccessToken`, loopback and OIDC token tests) run the token flow without a session cookie and are unaffected.

```
docker compose exec appwrite test tests/e2e/Services/Account --filter=OAuth2
```

I did not run the e2e suite locally (the only local stack on this machine is bound to a different working copy), so please rely on CI for the run. `php -l` and Pint pass on the touched files.

## Related PRs and Issues

- Console: appwrite/vibes `feat/smtp-resend-quick-setup` works around this with a guest-tolerant `/auth/smtp/callback` route that restores the session; with this fix the success URL can point straight back at the page that started the flow.

## Checklist

- [x] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs? (Only the endpoint's docs markdown changed; no route metadata or specs.)

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -1694,13 +1694,20 @@ Http::get('/v1/account/sessions/oauth2/:provider/redirect')
             $sessionUpgrade = true;
         }
 
-        $current = $user->sessionVerify($store->getProperty('secret', ''), $proofForToken);
+        // The session flow replaces the caller's current session with the one
+        // created below, so the browser stays signed in. The token flow creates
+        // no session here; the app exchanges the returned token through
+        // createSession later. Deleting the current session in that flow would
+        // sign the caller out until the exchange, so it is left in place.
+        if (empty($state['token'])) {
+            $current = $user->sessionVerify($store->getProperty('secret', ''), $proofForToken);
 
-        if ($current) { // Delete current session of new one.
-            $currentDocument = $dbForProject->getDocument('sessions', $current);
-            if (!$currentDocument->isEmpty()) {
-                $dbForProject->deleteDocument('sessions', $currentDocument->getId());
-                $dbForProject->purgeCachedDocument('users', $user->getId());
+            if ($current) { // Replace the current session with the new one.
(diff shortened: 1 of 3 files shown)


Meldiron (commented):
lgtm

PR #13626, opened 2026-09-11
fix(messaging): support alphanumeric Twilio sender IDs

### LINKED ISSUE #7783: 🚀 Feature: Support for alphanumeric sender id when using SMS
### 🔖 Feature description

In its current state, Appwrite only support numbers like "+1 123 422..." to send SMS messages, but providers like Twilio offer  alphanumeric sender id's. You could for example send a message in the name of your company/organization like "Appwrite", "Google", etc. In the case of Twilio you replace the numeric id with the alphanumeric id in the request. Just remove the phone number validation and I guess it should work.

### 🎤 Pitch

It looks way more professional and the phone number itself is not locked to a certain country code.

HarshMN2345:
## What does this PR do?

Twilio provider create validated `from` with the phone validator, while update accepted any text up to 256 characters. An alphanumeric sender ID such as `Appwrite` returned 400 on create but saved on update. Migrations re-create providers through create, so they hit the same check.

- `POST /v1/messaging/providers/twilio` now validates `from` with `Text(256)`, matching update, and both routes describe it as a phone number or alphanumeric sender ID. Twilio accepts E.164 numbers, alphanumeric sender IDs, short codes and Messaging Service SIDs as the sender and rejects invalid ones at send time. The worker already sends `options.from` unchanged.
- The internal `_APP_SMS_FROM` path treated any value starting with `MG` as a Messaging Service SID, so a sender like `MGM Resorts` was also sent as `MessagingServiceSid`. It now also requires the SID length of 34 characters, since alphanumeric sender IDs are at most 11, and the variable description mentions alphanumeric senders.
- `AGENTS.md` discourages adding regular expressions.

Vonage, Textmagic and Telesign keep their current create validator. The generated spec for Twilio create now documents `from` as a plain string instead of `format: phone`. The old Svelte console still restricts this field to phone numbers.

## Test Plan

- New e2e tests `testCreateTwilioProviderAlphanumericSender` and `testUpdateTwilioProviderAlphanumericSender` pass in `MessagingCustomServerTest`, `MessagingConsoleClientTest` and `MessagingCustomClientTest`. With `Create.php` from `main`, the create test fails with 400 instead of 201.
- `testCreateProviders` and `testUpdateProviders` still pass in `MessagingCustomServerTest` and `GraphQL/MessagingTest`.
- The SID check was compared against a real-format SID, `MGM Resorts`, `MGAppwrite`, `MG123` and `+123456789`. The internal Twilio path is not exercised in e2e.
- Pint, PHPStan on the changed files and Rector (`tests/tools/rector.php`) pass.

## Related PRs and Issues

- Closes #7783
- Console copy: appwrite/vibes#344

## Checklist

- [ ] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs? Specs are generated from route params and are not tracked in this repo.


src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -1278,9 +1278,9 @@ private function createProviderFromDSN(DSN $dsn): Document
                 'twilio' => [
                     'accountSid' => $user,
                     'authToken' => $password,
-                    // Twilio Messaging Service SIDs always start with MG
-                    // https://www.twilio.com/docs/messaging/services
-                    'messagingServiceSid' => \str_starts_with($from, 'MG') ? $from : null
+                    // Alphanumeric sender IDs can also start with MG, so require the full Messaging Service SID format
+                    // https://www.twilio.com/docs/messaging/api/service-resource
+                    'messagingServiceSid' => \preg_match('/^MG[0-9a-fA-F]{32}$/', $from) ? $from : null
```
Meldiron: Why was starts_with bad? a bit worried to introduce regex here

src/Appwrite/Platform/Workers/Messaging.php
```diff
@@ -1313,7 +1313,7 @@ private function createProviderFromDSN(DSN $dsn): Document
             },
             'options' => match ($host) {
                 'twilio' => [
-                    'from' => \str_starts_with($from, 'MG') ? null : $from
+                    'from' => \preg_match('/^MG[0-9a-fA-F]{32}$/', $from) ? null : $from
```
Meldiron: Same here, not fan of regex

Meldiron (approved):

PR #13634, opened 2026-09-12
fix(databases): widen integer column when updateIntegerAttribute widens the range

### LINKED ISSUE #13633: updateIntegerAttribute allows widening min/max beyond the column width fixed at create time
## Summary

`createIntegerAttribute` sizes the underlying SQL column from the requested range: 4 bytes (INT) when `min`/`max` fit in INT32, 8 bytes (BIGINT) otherwise.

`updateIntegerAttribute` lets you set new `min`/`max` values (the endpoint validators allow the full 64-bit range) but never passes a `size`, so the column keeps the width chosen at create time while `formatOptions` are overwritten with the new, wider range.

## Reproduction

1. `createIntegerAttribute` with `min: 0`, `max: 100` -> stored size 4, column is `INT`.
2. `updateIntegerAttribute` with `min: -9223372036854775808`, `max: 9223372036854775807` -> succeeds; the returned attribute now promises the full 64-bit range.
3. Create/update a document with value `2147483648`: it passes the attribute's range validation (formatOptions promise it) but the `INT` column overflows, surfacing as an opaque database error.

In `utopia-php/database`, `Database::updateAttribute` falls back to the stored size when the caller passes `null` and only issues the `ALTER TABLE ... MODIFY` when a new size is supplied, so the width can never change through this endpoint today.

## Expected behavior

The two endpoints should agree: if the updated range no longer fits the column, the column should be widened (INT -> BIGINT is a safe, lossless widen; narrowing should stay forbidden to avoid truncating existing data).

## Suggested fix

In the shared attributes `Action::updateAttribute`, when the stored format is `APP_DATABASE_ATTRIBUTE_INT_RANGE` and the attribute is not an array, compute the required size from the new `min`/`max` exactly as create does and pass it down so the adapter widens the column. Extract the boundary check shared by both endpoints into one helper.

I have the fix and unit tests ready and will open a PR referencing this issue.

Tyagiquamar:
Fixes #13633

## Problem

`createIntegerAttribute` sizes the SQL column from the requested range: 4 bytes (`INT`) when `min`/`max` fit INT32, 8 bytes (`BIGINT`) otherwise. `updateIntegerAttribute` accepts new `min`/`max` values (the endpoint validators allow the full 64-bit range) but never passes a `size`, so:

- the column keeps the width chosen at create time,
- `formatOptions` are overwritten with the wider range,
- a document write beyond INT32 passes attribute validation but overflows the column, surfacing as an opaque database error.

In `utopia-php/database`, `Database::updateAttribute` falls back to the stored size when the caller passes `null` and only issues the `ALTER TABLE ... MODIFY` when a new size is supplied, so the width can never change through this endpoint today.

## Fix

- Extract the INT32 boundary check shared by both endpoints into `Action::intRangeColumnSize()` and use it in `createIntegerAttribute`.
- In the shared attributes `Action::updateAttribute`, when the stored format is `APP_DATABASE_ATTRIBUTE_INT_RANGE` and the attribute is not an array, compute the required size from the new `min`/`max` and pass it down when it exceeds the stored size, so the adapter issues the lossless `INT -> BIGINT` widen. Narrowing is never applied, to avoid truncating existing data.

## Tests

New unit test `tests/unit/Platform/Modules/Databases/Http/Databases/Collections/Attributes/ActionTest.php` covers the boundary (INT32 bounds inclusive, one value outside either side, full 64-bit range), 7 cases via data provider.

Verified in the repo's `appwrite/base:2.0.5` container (PHP 8.5): new test 7/7, `pint --test` PASS on changed files, `phpstan analyse` no errors on changed files, full `tests/unit` run shows only one pre-existing unrelated error (`Auth/KeyTest`, `adhocore/jwt` vs PHP 8.5), reproduced on the clean tree.

### DIFF
--- a/src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Attributes/Action.php
+++ b/src/Appwrite/Platform/Modules/Databases/Http/Databases/Collections/Attributes/Action.php
@@ -493,6 +493,17 @@ abstract class Action extends DatabasesAction
         return $attribute;
     }
 
+    /**
+     * The SQL column for an integer range attribute is sized at create time
+     * to fit the promised range: 4 bytes when the range fits in INT32, 8
+     * bytes otherwise. The same boundary must govern updates, so the stored
+     * formatOptions cannot promise values the column cannot hold.
+     */
+    public static function intRangeColumnSize(int $min, int $max): int
+    {
+        return $min >= -2147483648 && $max <= 2147483647 ? 4 : 8;
+    }
+
     protected function updateAttribute(string $databaseId, string $collectionId, string $key, Database $dbForProject, Event $queueForEvents, Authorization $authorization, string $type, ?int $size = null, ?string $filter = null, string|bool|int|float|array|null $default = null, ?bool $required = null, int|float|null $min = null, int|float|null $max = null, ?array $elements = null, array $options = [], ?string $newKey = null): Document
     {
         $db = $authorization->skip(fn () => $dbForProject->getDocument('databases', $databaseId));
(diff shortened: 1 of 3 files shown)


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

PR #13639, opened 2026-09-12
fix(abuse): lease the rate limiter's Redis connection from a pool

ChiragAgg5k:
## What

Concurrent requests can kill a production HTTP worker. Any two requests that run the abuse rate limiter at the same time in one worker hit a fatal `Swoole\Error` and the worker exits with status 255.

```
Fatal error: Uncaught Swoole\Error: Socket#15 has already been bound to another
coroutine#53, reading of the same socket in coroutine#54 at the same time is not allowed
  in .../utopia-php/abuse/src/Abuse/Adapters/TimeLimit/Redis.php:49
#0 .../TimeLimit/Redis.php(49): Redis->get('abuse__ip:192.1...')
#1 .../TimeLimit.php(79): ...Redis->count('ip:192.168.167....', 1789233060)
#2 /usr/src/code/app/controllers/shared/api.php(587): ...TimeLimit->remaining()
#3 .../Http.php(642): {closure:.../app/controllers/shared/api.php:544}
```

This is B1 from the 2.1.0 release qualification.

## Why it happens

`app/init/resources.php` registered `redis` as a bare `\Redis` built with `pconnect()`, and `timelimit` built its `TimeLimit` on it. `Utopia\DI\Container::get()` memoizes on the parent container, and `pconnect` returns the same persistent socket per process anyway, so every coroutine in a worker shared one connection.

That was survivable until 2.1.0. `app/http.php` now uses `Mode::HYPERLOOP_B`, which sets `OPTION_HOOK_FLAGS => SWOOLE_HOOK_ALL`. 2.0.0 passed a plain settings array with no hook flags. phpredis connects over `php_stream`, so hooks make that socket coroutine-aware — and two coroutines reading it concurrently is exactly this error. Before the hooks, the shared socket simply blocked the worker and serialized.

`remaining()` at `api.php:587` runs before the `$shouldCheckAbuse` gate, so an API key does not avoid the path either.

Every other Redis consumer already leases from `Utopia\Pools\Group` — `locks`, `dbForProject`, the cache adapters. `timelimit` was the one path that never did.

**Docker cannot see this.** Through the whole failure the container reports `RestartCount 0` and `health=healthy`; only the in-container PIDs show the workers dying and respawning under the supervisor. Container-level checks in CI or production would never flag it.

## What changed

| File | Change |
|---|---|
| `app/init/registers.php` | New `abuse` Redis pool; `_APP_POOL_CLIENTS` default 14 → 15 for it |
| `app/init/resources.php` | Removed the process-global `redis` resource; `timelimit` now leases from the pool |
| `app/controllers/shared/api.php` | Both abuse hooks run inside the lease |
| `Functions/Http/Functions/Create.php` | Same, for its creation-rate check |
| `account.php` ×2, `Teams/…/Memberships/Create.php`, `Account/…/MFA/Challenges/Create.php` | Removed dead `timelimit` injections |
| `tests/e2e/Services/Storage/StorageBase.php` | New `testCreateBucketFileParallelUploads` |

Two decisions worth calling out:

**Abuse gets its own pool rather than sharing `lock`.** `locks` holds its lease for the whole guarded callback, and for storage chunk uploads that spans the transfer. Rate-limit checks must not queue behind uploads.

**Call sites must not throw inside the lease.** `Pool::use()` treats an exception as a failed lease and calls `recover()`, which runs `\Redis::reset()` and `reconnect()` — so throwing `GENERAL_RATE_LIMIT_EXCEEDED` from inside would tear down and rebuild the socket on every 429. Each call site returns its decision and throws outside. There is a comment at the resource saying so.

Four of the seven `timelimit` injections turned out to be dead — no `TimeLimit` or `Abuse` reference anywhere in those files — so they are gone.

## Verification

On the 2.1.0 production image (MariaDB lab, 2 HTTP workers), 64 concurrent unauthenticated multipart uploads to a `create("any")` bucket:

| | Before | After |
|---|---|---|
| N=64 × 3 rounds | 56, 55, 38 failed (60s timeouts) | 63 × 201, 1 × 429 |
| `Swoole\Error` fatals | 12 | **0** |
| Worker exits (255) | both workers, repeatedly | none — PIDs survived |

The single 429 is correct; the limit is 60 writes/60s.

The new e2e test, same command both ways on a 2-worker stack:

- without the fix: **fails 3/3** — `Upload 21 failed at the connection level: Operation timed out`, 12 fatals
- with the fix: **passes 3/3**, 0 fatals

Suites: unit 1202 tests / 6038 assertions clean · Storage CustomServer 31/31 including the duplicate-chunks case the qualification run had to interrupt · Teams 48/48 · Account `Phone` 6/6 and `MFA` 7/7, covering every signature touched · Pint and PHPStan clean on all 8 files.

## Notes for review

**The new test is probabilistic, not deterministic.** Its power scales as N²/workers, so on a very wide worker pool it can pass even unfixed. It is sized at 24 concurrent uploads: enough to catch the bug at 2 workers, light enough to pass on a 72-worker box. At 64 it also reproduces, but 72 workers × their DB pools exhausts a dev box's `max_connections=100`.

**Pool sizing is still wrong for default installs, separately from this PR.** `app/init/registers.php:132` computes `poolSize = (151/15) / (CPU × 6)`, which floors at 1 on any box with ≥2 cores. That was sized for one-request-per-worker and is now wrong under HYPERLOOP_B coroutines. It is not what caused this crash — the lab crashed with `poolSize=5`, because `timelimit` touched no pool at all — but it deserves a measured fix.

**`app/realtime.php:231-254` has the same shape** and is not fixed here. `getRedis()` caches in `Coroutine::getContext()` but calls `pconnect()`, so all connections share one socket, and `getTimelimit()` runs abuse checks on it at `:967`. Pre-existing — `SWOOLE_HOOK_ALL` was already set in 2.0.0's realtime — so not a 2.1 regression, but it is the same crash waiting on a busy realtime server. Worth its own issue.

**Worth a sweep before shipping 2.1.0.** The HYPERLOOP_B switch makes every unpooled long-lived client in the HTTP path suspect. `timelimit` was the only one reaching the shared `redis` resource, but nothing guarantees it is the only one of its kind.

### DIFF
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ -3125,13 +3125,12 @@ Http::post('/v1/account/tokens/phone')
     ->inject('queueForEvents')
     ->inject('publisherForMessaging')
     ->inject('locale')
-    ->inject('timelimit')
     ->inject('usage')
     ->inject('plan')
     ->inject('store')
     ->inject('proofForCode')
     ->inject('authorization')
-    ->action(function (string $userId, string $phone, Request $request, Response $response, User $user, Document $project, array $platform, Database $dbForProject, Event $queueForEvents, MessagingPublisher $publisherForMessaging, Locale $locale, callable $timelimit, Context $usage, array $plan, Store $store, ProofsCode $proofForCode, Authorization $authorization) {
+    ->action(function (string $userId, string $phone, Request $request, Response $response, User $user, Document $project, array $platform, Database $dbForProject, Event $queueForEvents, MessagingPublisher $publisherForMessaging, Locale $locale, Context $usage, array $plan, Store $store, ProofsCode $proofForCode, Authorization $authorization) {
         if (empty(System::getEnv('_APP_SMS_PROVIDER'))) {
             throw new Exception(Exception::GENERAL_PHONE_DISABLED, 'Phone provider not configured');
         }
@@ -4541,12 +4540,11 @@ Http::post('/v1/account/verifications/phone')
     ->inject('publisherForMessaging')
     ->inject('project')
     ->inject('locale')
-    ->inject('timelimit')
     ->inject('usage')
     ->inject('plan')
     ->inject('proofForCode')
                 ->inject('authorization')
-    ->action(function (Request $request, Response $response, User $user, Database $dbForProject, Event $queueForEvents, MessagingPublisher $publisherForMessaging, Document $project, Locale $locale, callable $timelimit, Context $usage, array $plan, ProofsCode $proofForCode, Authorization $authorization) {
+    ->action(function (Request $request, Response $response, User $user, Database $dbForProject, Event $queueForEvents, MessagingPublisher $publisherForMessaging, Document $project, Locale $locale, Context $usage, array $plan, ProofsCode $proofForCode, Authorization $authorization) {
         if (empty(System::getEnv('_APP_SMS_PROVIDER'))) {
             throw new Exception(Exception::GENERAL_PHONE_DISABLED, 'Phone provider not configured');
         }
--- a/app/controllers/shared/api.php
+++ b/app/controllers/shared/api.php
@@ -26,6 +26,7 @@ use Appwrite\Utopia\Database\Documents\User;
 use Appwrite\Utopia\Request;
 use Appwrite\Utopia\Response;
 use Utopia\Abuse\Abuse;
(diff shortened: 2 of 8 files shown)


ChiragAgg5k:
Superseded by a PR from the branch on appwrite/appwrite rather than a fork.

PR #13646, opened 2026-09-14
fix(teams): preserve query parameters in SMS invitations

### LINKED ISSUE #8607: 🐛 Bug Report: Query Parameters in SMS Invitation URL Incorrectly Encoded
### 👟 Reproduction steps

1. Send a team invitation to a user using their phone number via SMS.
2. Copy the URL from the received SMS message.
3. Notice that the URL contains `&amp;` in place of `&` for separating query parameters.

### 👍 Expected behavior

The URL should correctly use `&` to separate query parameters, making it straightforward to access and use them.

### 👎 Actual Behavior

The URL contains `&amp;` instead of `&`, causing issues when trying to parse and use the query parameters directly. As a result, the query parameters are divided incorrectly, e.g., `amp;teamsId` instead of `teamsId`, which leads to parsing errors and difficulties in accessing the intended parameters.

|      SMS       |      Console Log of params       |
| :-----------------: | :----------------: |
| WhatsApp Image 2024-09-02 at 02 26 13_47724949 | image |

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

Windows

### 🧱 Your Environment

_No response_

HarshMN2345:
## What does this PR do?

Phone team invitations HTML-escape the redirect URL before putting it into a plaintext SMS. A URL containing `&userId=...&secret=...` arrives with `&amp;` separators, so the app cannot read the invitation parameters.

Disable HTML escaping when substituting the URL into the SMS template. Existing query values and fragments retain their URL encoding.

## Test Plan

- Added `TeamsCustomClientTest::testCreateMembershipByPhone`, which reads the delivered SMS from the existing local request catcher, checks the original query and fragment plus invitation fields, accepts the invitation with those fields, and checks replay rejection.
- Reproduced broken query parsing with the existing template and default escaping; verified all fields survive with plaintext substitution.
- Existing Template unit suite passes: 6 tests, 10 assertions.
- PHP syntax, targeted Pint, PHPStan, Rector dry-run, and `git diff --check` pass.
- GitHub Actions ran the new SMS regression successfully at `217b279919`; Checks and Unit also passed. The complete Teams run has a separate console-membership fixture failure caused by a 409 lock-contention response (49 tests, 1 unrelated failure, 1 expected skip). [Teams job](https://github.com/appwrite/appwrite/actions/runs/34820541516/job/103901281960). Local Docker/API testing remains unavailable.

## Related PRs and Issues

Fixes #8607.

## Checklist

- [x] Read the contributing guidelines.
- [x] API metadata is unchanged; regenerated specs and examples are not needed.


src/Appwrite/Platform/Modules/Teams/Http/Memberships/Create.php
```diff
@@ -484,7 +484,7 @@ public function action(string $teamId, ?string $email, ?string $userId, ?string
 
                 $message = Template::fromFile(APP_CE_CONFIG_DIR . '/locale/templates/sms-base.tpl');
 
-                $message = $message->setParam('{{token}}', $url);
```
Meldiron, line 487: Let's limit this to only magic uRL flows - lets keep escapeHtml to true for when SMS is just OTP

HarshMN2345: This line only handles the team invitation SMS, which always sends the membership accept URL. The url is required for this flow and isnt used for OTP. The OTP SMS flows (phone token, phone verification, and MFA challenge) still use the default ``escapeHtml: true``

HarshMN2345: cc @greptile

Meldiron (approved):

PR #13653, opened 2026-09-14
fix(account): use the impersonated user for account reads and block account writes

HarshMN2345:
## What does this PR do?

While a console operator impersonates a user, only `GET` and `DELETE /v1/account` use the target. Every other account endpoint still reads the operator's own session user. In the console, `/account/security` shows the operator's identities, sessions, MFA factors and prefs under the target's name, and account edits change the operator's account.

- **Reads use the target.** `GET /v1/account/sessions`, `GET /v1/account/sessions/:sessionId`, `GET /v1/account/prefs`, `GET /v1/account/identities` and `GET /v1/account/mfa/factors` now inject `targetUser`. `GET /v1/account/sessions/current` still resolves `current` against the operator's own session, which is the one making the request. The sessions list marks none of the target's sessions as current.
- **Writes are refused while impersonating.** Non-GET routes in the `account` namespace throw the new `user_impersonation_read_only` (403).
  - Routes in the `mfa` group stay open, because they act on the operator's own session (MFA challenges, deleting the current session to log out).
  - A route can override the default with `->label('impersonation', 'allow' | 'deny')`. `POST /v1/account/jwts` is `allow`.
  - `DELETE /v1/account` used `targetUser` before this change and is now refused like other writes.
- **Credentials are never exposed.** Three routes are labelled `deny`:
  - `GET /v1/account/mfa/recovery-codes`
  - `GET /v1/users/:userId/mfa/recovery-codes`
  - `GET /v1/users/:userId/mfa/challenges/:challengeId`

  Impersonators get `users.read` implicitly on the console project, and that grant now skips `deny` routes. Before this change, any operator with the impersonator flag could read another console user's MFA recovery codes and challenge secrets, even without impersonating.

## Test Plan

- `pint --test` and `phpstan` pass on the touched files.
- No e2e run. Manual checks with an impersonator account:
  1. With a session and `X-Appwrite-Impersonate-User-Id: <target>`:
     - `GET /v1/account/identities`, `/sessions`, `/prefs` and `/mfa/factors` return the target's data.
     - `PATCH /v1/account/name` returns 403 `user_impersonation_read_only`.
     - `POST /v1/account/jwts` still succeeds.
     - `GET /v1/account/mfa/recovery-codes` returns 403.
     - `GET /v1/account/sessions/current` returns the operator's session, and `GET /v1/account/sessions` marks none as current.
     - `DELETE /v1/account/sessions/current` logs the operator out.
  2. Same operator without the header: `GET /v1/users/<another user>/mfa/recovery-codes` on the console project is refused, while `GET /v1/users` still works.
  3. An API key with `users.read` on a regular project can still call `GET /v1/users/:userId/mfa/recovery-codes`.

## Related PRs and Issues

- Cloud: appwrite-labs/cloud#5798 uses `targetUser` for cloud account reads (billing, consents, keys, affiliates).
- Console: appwrite/vibes#349 disables account controls while impersonating.

## Checklist

  - The only new label is the internal `impersonation` label, which isn't part of the specs. There is one new error code, `user_impersonation_read_only`.


Hardanish-Singh (commented):
LGTM

ChiragAgg5k (approved):

PR #13677, opened 2026-09-14
Add OTP support for email verification and password recovery

imtia33:
## What does this PR do?

Adds support for OTP-based **Email Verification** and **Password Recovery** in the Account API, making it easy and seamless for mobile application developers to verify emails and reset passwords without requiring web redirects or deep linking.

4 new endpoints are introduced:
1. `POST /v1/account/verifications/email/otp` (`createEmailVerificationOTP`): Generates and emails a 6-digit OTP code to verify the user's email address.
2. `PUT /v1/account/verifications/email/otp` (`updateEmailVerificationOTP`): Validates the 6-digit OTP code and marks the email as verified.
3. `POST /v1/account/recovery/otp` (`createRecoveryOTP`): Generates and emails a 6-digit OTP code for password reset.
4. `PUT /v1/account/recovery/otp` (`updateRecoveryOTP`): Validates the 6-digit OTP code and updates the user's password.

Note: I know this looks like a huge PR but actual changes are minimal. Reused existing architecture of email verification and password recovery. This PR creates new endpoints for OTP support for both methods mentioned. This can be scoped down to one method at a time if required. 

Update: Now the PR is scoped down to email verification only .

### Implementation Rationale & Reuse
The implementation reuses Appwrite's existing architecture and patterns:
- Existing token types: `TOKEN_TYPE_VERIFICATION` (2) and `TOKEN_TYPE_RECOVERY` (3).
- Existing token expiration (`TOKEN_EXPIRATION_OTP` - 15 minutes).
- Existing email infrastructure (`MAIL_TEMPLATE_OTP` / `email-otp.tpl`) and security phrase mechanism (`Phrase()`).
- Replaces magic URL token generation (`ProofsToken`) with 6-digit OTP generation (`ProofsCode`). Net-new logic additions are minimal.

## Test Plan

Added end-to-end unit and integration test coverage in `tests/e2e/Services/Account/AccountCustomClientTest.php`:
- `testCreateEmailVerificationOTP`: Verifies creation of verification OTP and email delivery.
- `testUpdateEmailVerificationOTP`: Tests successful verification with OTP, as well as failure cases (expired code, wrong user, invalid code).
- `testCreateRecoveryOTP`: Verifies creation of recovery OTP and email delivery.
- `testUpdateRecoveryOTP`: Tests password reset with OTP, verifying password changes and invalid attempt failures.

### Local Verification
Executed PHPUnit E2E tests inside the Appwrite container:
```bash
docker compose exec appwrite vendor/bin/phpunit tests/e2e/Services/Account/AccountCustomClientTest.php --filter=OTP
```
Output:
```text
OK (5 tests, 170 assertions)
```

## Related PRs and Issues

Draft PR to gather feedback on OTP-based email verification and password recovery endpoints for mobile developers.

## Checklist


app/controllers/api/account.php
```diff
@@ -5035,3 +5035,651 @@
...
+
+        $dbForProject->purgeCachedDocument('users', $user->getId());
+
+        $subject = $locale->getText('emails.otpVerification.subject');
+        $preview = $locale->getText('emails.otpVerification.preview');
+        $heading = $locale->getText('emails.otpVerification.heading');
+
+        $customTemplate =
+            $project->getAttribute('templates', [])['email.otpVerification-' . $locale->default] ??
+            $project->getAttribute('templates', [])['email.otpVerification-' . $locale->fallback] ?? [];
```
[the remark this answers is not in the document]
imtia33: already resolved. recheck

src/Appwrite/Platform/Modules/Console/Http/Templates/Email/Get.php
```diff
@@ -90,6 +90,10 @@ private function getDefaultMessage(string $templateId, Locale $localeObj): strin
                 'file' => 'email-otp.tpl',
                 'placeholders' => ['description', 'clientInfo', 'securityPhrase']
             ],
+            'otpVerification' => [
+                'file' => 'email-otp.tpl',
+                'placeholders' => ['description', 'clientInfo', 'securityPhrase']
```
Meldiron, line 95: @greptile Please review this specifically. Are these placeholder names accurate, considering the newly added endpoint and variables it sets?

Please investigate entire flow to understand expected behaviour, and let me know if these placeholders are accurate, if any are extra, or any are missing

Meldiron:
CleanShot 2026-09-15 at 10 13 54@2x

looks great

Meldiron (approved):

Meldiron (approved):

app/controllers/api/account.php
```diff
@@ -5035,3 +5035,300 @@
...
+    ->groups(['api', 'account'])
+    ->label('scope', 'account')
+    ->label('event', 'users.[userId].verification.[tokenId].create')
+    ->label('audits.event', 'verification.create')
+    ->label('audits.resource', 'user/{response.userId}')
+    ->label('sdk', new Method(
+        namespace: 'account',
+        group: 'verification',
+        name: 'createEmailVerificationOTP',
+        description: '/docs/references/account/create-email-verification-otp.md',
```
Meldiron, line 5050: `docs/references/account/create-email-verification-otp.md` doesn't exist in the repo. The SDK/docs generator resolves this string as a file path, so it needs to be added alongside the existing `create-email-verification.md` / `create-phone-verification.md`.

app/controllers/api/account.php
```diff
@@ -5035,3 +5035,300 @@
...
+        responses: [
+            new SDKResponse(
+                code: Response::STATUS_CODE_CREATED,
+                model: Response::MODEL_TOKEN,
+            )
+        ],
+        contentType: ContentType::JSON,
+    ))
+    ->label('abuse-limit', 10)
+    ->label('abuse-key', 'userId:{userId},ip:{ip}')
```
Meldiron, line 5061: This abuse key omits `url:{url}`, which every comparable route in this file includes — `createVerification` just above uses `'url:{url},userId:{userId}'` (line 4225).

Abuse keys are not namespaced by route: `TimeLimit\Redis` builds the Redis key as `NAMESPACE__<parsed key>__<timestamp>`, so two routes with the same key string share one bucket. `'userId:{userId},ip:{ip}'` happens to be unique today, but the `url:` prefix is what guarantees that stays true.

Suggest matching line 4225:

```php
->label('abuse-key', 'url:{url},userId:{userId}')
```

app/controllers/api/account.php
```diff
@@ -5035,3 +5035,300 @@
...
+                'platform' => $platform['platformName'],
+            ]);
+        }
+
+        $publisherForMails->enqueue(new MailMessage(
+            project: $project,
+            recipient: $user->getAttribute('email'),
+            name: $user->getAttribute('name') ?? '',
+            subject: $subject,
+            template: MAIL_TEMPLATE_OTP,
```
Meldiron, line 5245: Reusing `MAIL_TEMPLATE_OTP` here means email-verification OTPs are indistinguishable from login OTPs in telemetry — the worker does `Span::add('mail.template', $template)` and nothing else with this value, so `mail.template=otp` now covers two different flows. A `MAIL_TEMPLATE_OTP_VERIFICATION = 'otp-verification'` constant would keep the mail metrics separable. Purely observability; no functional impact.

app/controllers/api/account.php
```diff
@@ -5035,3 +5035,300 @@
...
+    ->groups(['api', 'account'])
+    ->label('scope', 'public')
+    ->label('event', 'users.[userId].verification.[tokenId].update')
+    ->label('audits.event', 'verification.update')
+    ->label('audits.resource', 'user/{response.userId}')
+    ->label('sdk', new Method(
+        namespace: 'account',
+        group: 'verification',
+        name: 'updateEmailVerificationOTP',
+        description: '/docs/references/account/update-email-verification-otp.md',
```
Meldiron, line 5282: Same as the `POST` above — `docs/references/account/update-email-verification-otp.md` isn't in the repo.

app/controllers/api/account.php
```diff
@@ -5035,3 +5035,300 @@
...
+        responses: [
+            new SDKResponse(
+                code: Response::STATUS_CODE_OK,
+                model: Response::MODEL_TOKEN,
+            )
+        ],
+        contentType: ContentType::JSON,
+    ))
+    ->label('abuse-limit', 10)
+    ->label('abuse-key', 'userId:{param-userId}')
```
Meldiron, line 5293: **This collides with `PUT /v1/account/verifications/phone`** (line 4673), which uses the identical key string `'userId:{param-userId}'`.

Abuse keys aren't namespaced by route — `TimeLimit\Redis` keys are `NAMESPACE__<parsed key>__<timestamp>` — so both endpoints draw from one 10/hour budget per user. Reproduced locally with `_APP_OPTIONS_ABUSE=enabled`:

```
12x PUT /v1/account/verifications/phone    -> 401 x10, then 429 429
 1x PUT /v1/account/verifications/email/otp -> 429   (first ever call)
```

A user who fumbles phone verification is locked out of email verification, and vice versa. The sibling email endpoint at line 4471 gets this right:

```php
->label('abuse-key', 'url:{url},userId:{param-userId}')
```

app/controllers/api/account.php
```diff
@@ -5035,3 +5035,300 @@
...
+    ->inject('authorization')
+    ->action(function (string $userId, string $secret, Response $response, User $user, Database $dbForProject, Event $queueForEvents, ProofsCode $proofForCode, Authorization $authorization) {
+        /** @var \Appwrite\Utopia\Database\Documents\User $profile */
+        $profile = $authorization->skip(fn () => $dbForProject->getDocument('users', $userId));
+
+        if ($profile->isEmpty()) {
+            throw new Exception(Exception::USER_NOT_FOUND);
+        }
+
+        $verifiedToken = $profile->tokenVerify(TOKEN_TYPE_VERIFICATION, $secret, $proofForCode);
```
Meldiron, line 5310: Because this mints and verifies `TOKEN_TYPE_VERIFICATION` — the same type the URL-based flow uses — the two secrets are interchangeable. `tokenVerify()` only matches on type + hash + expiry, and `ProofsCode` and `ProofsToken` both inherit `Proof::verify`, so the hash comparison doesn't distinguish them.

Confirmed locally: a 6-digit OTP minted by `POST /v1/account/verifications/email/otp` is accepted by the *legacy* endpoint.

```
PUT /v1/account/verifications/email  {"userId":"...","secret":"795720"}  ->  200, email verified
```

Two consequences worth weighing:

1. `PUT /v1/account/verifications/email` was written around a 256-bit secret, so its `abuse-limit 10` was defence in depth. It now guards a 10^6 keyspace, and nobody reading that endpoint would know.
2. Its abuse key (`url:{url},userId:{param-userId}`) is a *different* bucket from this one, so an attacker gets 10 guesses/hour here **plus** 10 there against the same code.

Giving the OTP its own token type (or tagging the token) so each endpoint only accepts its own kind would close both.

Meldiron (commented):
Agentic comments

src/Appwrite/Platform/Modules/Console/Http/Templates/Email/Get.php
```diff
@@ -90,6 +90,10 @@ private function getDefaultMessage(string $templateId, Locale $localeObj): strin
                 'file' => 'email-otp.tpl',
                 'placeholders' => ['description', 'clientInfo', 'securityPhrase']
             ],
+            'otpVerification' => [
```
Meldiron, line 93: To answer the obvious question, since it comes up on every one of these: **`'otp'` is correctly absent here and must stay absent.**

This list isn't "placeholders in the .tpl" — each entry is fed to `getText("emails.{$templateId}.{$param}")` below. There is no `emails.otpVerification.otp` key, and since `Locale::$exceptions` is never set to `false` anywhere in `app/` or `src/`, adding `'otp'` would throw `Key named "..." not found` and 500 this endpoint. `{{otp}}` has to survive verbatim so the console editor shows it as an available variable and the mail worker substitutes the real code at send time — same as `{{user}}`, `{{project}}`, `{{phrase}}`, `{{agent*}}`.

Verified the output is structurally identical to `otpSession`, with no `{{emails.*}}` leakage. No change needed.

The thing I would flag is that this map is now byte-identical to the one in `Project/.../Get.php`, and `getDefaultMessage()` is duplicated wholesale between the two files. Combined with the `?? ['file' => 'email-inner-base.tpl', ...]` fallback, a future template type whose entry is forgotten silently returns a *button* template with a 200 instead of failing. Worth extracting to a shared config.

src/Appwrite/Platform/Modules/Project/Http/Project/Templates/Email/Get.php
```diff
@@ -110,6 +110,10 @@ protected function getDefaultMessage(string $templateId, Locale $localeObj): str
                 'file' => 'email-otp.tpl',
                 'placeholders' => ['description', 'clientInfo', 'securityPhrase']
             ],
+            'otpVerification' => [
```
Meldiron, line 113: Byte-identical to the entry in `Console/Http/Templates/Email/Get.php` (line 93), as is the whole `getDefaultMessage()` method. Adding a template type now needs three coordinated edits (`templates.php` + both maps) with no compile-time or test-time link between them — see #4 in the summary for the missing `TemplatesBase.php` coverage that would have caught a mismatch.

The entry itself is correct; verified the rendered default matches `otpSession`.

tests/e2e/Services/Account/AccountCustomClientTest.php
```diff
@@ -6134,4 +6134,171 @@ public function testCreateIdTokenSessionWithoutEmail(): void
         $this->assertEquals($email, $account['body']['email']);
         $this->assertTrue($account['body']['emailVerification']);
     }
+
+    public function testCreateEmailVerificationOTP(): void
+    {
+        $data = $this->setupAccountWithSession();
+        $session = $data['session'];
+        $email = $data['email'];
+        $name = $data['name'];
```
Meldiron, line 6143: `$name` is assigned but never used in this test.

tests/e2e/Services/Account/AccountCustomClientTest.php
```diff
@@ -6134,4 +6134,171 @@ public function testCreateIdTokenSessionWithoutEmail(): void
...
+        $this->assertEmpty($response['body']['secret']);
+        $this->assertEmpty($response['body']['phrase']);
+        $this->assertTrue((new DatetimeValidator())->isValid($response['body']['expire']));
+
+        $lastEmail = $this->getLastEmailByAddress($email);
+        $this->assertNotEmpty($lastEmail, 'Email not found for address: ' . $email);
+        $this->assertNotEmpty($lastEmail['subject']);
+        $this->assertNotEmpty($lastEmail['text']);
+
+        preg_match_all("/\b\d{6}\b/", $lastEmail['text'], $matches);
```
Meldiron, line 6167: Taking `$matches[0][0]` makes this positional — it depends on the OTP being the *first* 6-digit run in the plaintext body. It happens to hold because `emails.otpVerification.preview` ("Use OTP {{otp}} to verify...") leads the text part, but any future copy change that introduces a number earlier silently breaks it. Matching against the response's own token, or anchoring the regex to the code block, would be sturdier.

Same pattern at line 6254.

tests/e2e/Services/Account/AccountCustomClientTest.php
```diff
@@ -6134,4 +6134,171 @@ public function testCreateIdTokenSessionWithoutEmail(): void
...
+            'origin' => 'http://localhost',
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $projectId,
+            'cookie' => 'a_session_' . $projectId . '=' . $session,
+        ]), [
+            'userId' => ID::custom('doesnotexist'),
+            'secret' => $otp,
+        ]);
+
+        $this->assertEquals(404, $response['headers']['status-code']);
```
Meldiron, line 6302: Three gaps in this file's coverage:

- **No expired-code case**, though the PR description lists one. Worth an explicit test given `TOKEN_EXPIRATION_OTP` is load-bearing here.
- **None of the `POST` handler's three guard clauses** are exercised: `GENERAL_SMTP_DISABLED`, `USER_EMAIL_NOT_FOUND`, and `USER_EMAIL_ALREADY_VERIFIED`.
- **Nothing asserts the token is single-use.** The handler deletes it after success, so replaying the same `secret` should 401 — currently untested.

Separately, the two `Get.php` changes have no coverage at all: `tests/e2e/Services/Project/TemplatesBase.php` keeps exhaustive template-id lists at lines 37-45 and 1277-1284, and `'otpVerification'` wasn't added to either.

app/controllers/api/account.php
Meldiron: Here's the file, modelled on `create-email-verification.md` and `create-phone-verification.md`:

**`docs/references/account/create-email-verification-otp.md`**

```markdown
Use this endpoint to send a 6-digit verification code to the currently logged in user's email address. Unlike [createEmailVerification](https://appwrite.io/docs/references/cloud/client-web/account#createEmailVerification), this method requires no redirect URL, which makes it suitable for mobile and desktop apps that cannot host a verification page. Learn more about how to [complete the verification process](https://appwrite.io/docs/references/cloud/client-web/account#updateEmailVerificationOTP). The code sent to the user's email address is valid for 15 minutes.

Enable the **phrase** parameter to include a randomly generated security phrase in both the email and the response. Showing that phrase in your app lets the user confirm the email genuinely came from your request, which helps protect against phishing.
```

app/controllers/api/account.php
Meldiron: Suggested change, matching `createVerification` at line 4225:

```suggestion
    ->label('abuse-key', 'url:{url},userId:{userId}')
```

Worth noting what dropping `ip:{ip}` does and doesn't cost you here. As written, user and IP are ANDed into a *single* key, so the counter resets whenever the IP changes — it doesn't add a second constraint, it weakens the existing one. Since this route already requires a session (`scope: account`), keying on the user alone is both stricter and consistent with the rest of the file.

If you do want an IP dimension, the convention elsewhere in this file is a two-element array so each is its own independent bucket:

```php
->label('abuse-key', ['url:{url},userId:{userId}', 'url:{url},ip:{ip}'])
```

app/controllers/api/account.php
Meldiron: Not a one-liner because the constant needs to exist first, so as a plain diff:

```diff
--- a/app/init/constants.php
+++ b/app/init/constants.php
@@
 const MAIL_TEMPLATE_OTP = 'otp';
+const MAIL_TEMPLATE_OTP_VERIFICATION = 'otp-verification';
 const MAIL_TEMPLATE_RECOVERY = 'recovery';
```

```diff
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@
-            template: MAIL_TEMPLATE_OTP,
+            template: MAIL_TEMPLATE_OTP_VERIFICATION,
```

Safe to apply: `Mails.php` only reads this value for `Span::add('mail.template', $template)` and never branches on it, so nothing downstream needs a matching case. Lowest-priority item in my review — take it or leave it.

app/controllers/api/account.php
Meldiron: Matching file for this one:

**`docs/references/account/update-email-verification-otp.md`**

```markdown
Use this endpoint to complete the user email verification process using the 6-digit code that was emailed by [createEmailVerificationOTP](https://appwrite.io/docs/references/cloud/client-web/account#createEmailVerificationOTP). Pass the **userId** of the user being verified along with the **secret** code from the email. If confirmed, this route will return a 200 status code and the code is consumed.
```

app/controllers/api/account.php
Meldiron: Suggested change, matching the sibling endpoint at line 4471:

```suggestion
    ->label('abuse-key', 'url:{url},userId:{param-userId}')
```

That alone separates this route's bucket from `PUT /v1/account/verifications/phone`, since `{url}` expands to `$request->getHostname() . $route->getPath()`.

app/controllers/api/account.php
Meldiron: Sketch of what separating the token types looks like. Four hunks, and the last one is the easiest to miss:

```diff
--- a/app/init/constants.php
+++ b/app/init/constants.php
@@
 const TOKEN_TYPE_EMAIL = 9; // OTP
+const TOKEN_TYPE_VERIFICATION_OTP = 10;
```

```diff
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ POST /v1/account/verifications/email/otp
         $verification = new Document([
             '$id' => ID::unique(),
             'userId' => $user->getId(),
             'userInternalId' => $user->getSequence(),
-            'type' => TOKEN_TYPE_VERIFICATION,
+            'type' => TOKEN_TYPE_VERIFICATION_OTP,
             'secret' => $proofForCode->hash($secret),
```

```diff
--- a/app/controllers/api/account.php
+++ b/app/controllers/api/account.php
@@ PUT /v1/account/verifications/email/otp
-        $verifiedToken = $profile->tokenVerify(TOKEN_TYPE_VERIFICATION, $secret, $proofForCode);
+        $verifiedToken = $profile->tokenVerify(TOKEN_TYPE_VERIFICATION_OTP, $secret, $proofForCode);
```

```diff
--- a/src/Appwrite/Platform/Workers/Deletes.php
+++ b/src/Appwrite/Platform/Workers/Deletes.php
@@ deleteExpiredTokens()
             TOKEN_TYPE_GENERIC,
             TOKEN_TYPE_EMAIL,
+            TOKEN_TYPE_VERIFICATION_OTP,
         ];
```

That last hunk is the one I'd flag hardest — `deleteExpiredTokens()` iterates an explicit allow-list of types, so a new type that isn't added there leaves expired OTP rows in the `tokens` collection forever.

Two things worth weighing before going this way, since it's your call and not obviously free:

- **It's a behaviour change for anyone already on this branch.** Tokens minted before the switch keep `type = 2` and stop validating. Irrelevant pre-merge, worth a note if this ships behind a preview.
- **It costs the flows interop.** Today a user can request the link *and* the code and finish with either. Splitting the types means each endpoint only accepts its own. I'd argue that's the point — the whole issue is that `PUT /v1/account/verifications/email` silently became a brute-force surface for a 10^6 keyspace — but it is a real trade-off rather than a pure win.

If you'd rather keep one type, the alternative is to tighten the rate limit on *both* verification routes to reflect that the weakest secret they now accept is 6 digits. That's a smaller change but leaves the coupling in place.

tests/e2e/Services/Account/AccountCustomClientTest.php
Meldiron:
```suggestion
```

tests/e2e/Services/Account/AccountCustomClientTest.php
Meldiron: Withdrawing this one — I checked the rest of the suite and `preg_match_all("/\b\d{6}\b/", ...)` with `$matches[0][0]` is the established convention here, not a one-off:

- `tests/e2e/Services/Account/AccountBase.php:199`
- `tests/e2e/Services/Account/AccountConsoleClientTest.php:240`
- `tests/e2e/Services/Account/AccountCustomClientTest.php:2097`

Changing it only in the new tests would make them inconsistent with the other three call sites for no real gain. If the brittleness ever bites, it should be fixed across all of them in a separate PR. No change needed here — feel free to resolve.

tests/e2e/Services/Account/AccountCustomClientTest.php
Meldiron: The two cheapest gaps close inline, right after the existing 404 check. Single-use is the one I'd most want covered, since the handler deletes the token and nothing currently asserts that:

```suggestion
        $this->assertEquals(404, $response['headers']['status-code']);

        /**
         * Test for FAILURE - OTP is single use
         */
        $response = $this->client->call(Client::METHOD_PUT, '/account/verifications/email/otp', array_merge([
            'origin' => 'http://localhost',
            'content-type' => 'application/json',
            'x-appwrite-project' => $projectId,
            'cookie' => 'a_session_' . $projectId . '=' . $session,
        ]), [
            'userId' => $userId,
            'secret' => $otp,
        ]);

        $this->assertEquals(401, $response['headers']['status-code']);
        $this->assertEquals('user_invalid_token', $response['body']['type']);

        /**
         * Test for FAILURE - email already verified
         */
        $response = $this->client->call(Client::METHOD_POST, '/account/verifications/email/otp', array_merge([
            'origin' => 'http://localhost',
            'content-type' => 'application/json',
            'x-appwrite-project' => $projectId,
            'cookie' => 'a_session_' . $projectId . '=' . $session,
        ]));

        $this->assertEquals(409, $response['headers']['status-code']);
        $this->assertEquals('user_email_already_verified', $response['body']['type']);
```

The 409 comes from `app/config/errors.php:372`. That covers one of the three `POST` guard clauses; `USER_EMAIL_NOT_FOUND` and `GENERAL_SMTP_DISABLED` need their own fixtures, and the expired-code case needs either a clock shim or a seeded token, so I've left those as prose rather than guessing at your preferred approach.

Separately, and not fixable from this line: `'otpVerification'` still needs adding to the two exhaustive lists in `tests/e2e/Services/Project/TemplatesBase.php` (lines 37-45 and 1277-1284), otherwise both `Get.php` changes ship untested.

Meldiron: Pushed in 086b6b7. Four of these are closed, one deliberately isn't:

- **Expired code** — not reachable from e2e: there's no clock shim and no route that mints a `TOKEN_TYPE_VERIFICATION_OTP` with a caller-supplied `expire`, so I covered it where the expiry check actually lives, as a unit regression on `User::tokenVerify` (`tests/unit/Utopia/Database/Documents/UserTest.php`).
- **`USER_EMAIL_NOT_FOUND`** — added, using the existing `createAnonymousSession()` helper; an anonymous user has no email, so the guard fires and returns 400.
- **`GENERAL_SMTP_DISABLED`** — still uncovered. It keys off `_APP_SMTP_HOST` being empty process-wide, which can't be toggled per-test against a running container. Would need its own compose fixture; I don't think it's worth one.
- **Single-use** and the 409 — already landed in b09c2e4.
- **`TemplatesBase.php`** — `'otpVerification'` is in both lists as of b09c2e4.

The unit test also covers what I flagged separately as the one gap I'd most want closed: `ProofsCode` and `ProofsToken` share `Proof::verify`, so `$token->verify($code, $codeHash)` returns `true` — the *only* thing keeping an OTP out of the link endpoint is the type check. The test pins both directions, and I confirmed it fails if `TOKEN_TYPE_VERIFICATION_OTP` is reverted to `TOKEN_TYPE_VERIFICATION`. There's a matching e2e case that puts the active OTP through `PUT /v1/account/verifications/email` and expects 401.

tests/e2e/Services/Account/AccountCustomClientTest.php
```diff
@@ -6135,16 +6135,201 @@ public function testCreateIdTokenSessionWithoutEmail(): void
...
+         * Test for FAILURE - wrong secret against active token
+         */
+        $response = $this->client->call(Client::METHOD_PUT, '/account/verifications/email/otp', array_merge([
+            'origin' => 'http://localhost',
+            'content-type' => 'application/json',
+            'x-appwrite-project' => $projectId,
+            'cookie' => 'a_session_' . $projectId . '=' . $session,
+        ]), [
+            'userId' => $userId,
+            'secret' => '000000',
```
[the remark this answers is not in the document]
Meldiron: Valid — fixed in 086b6b7 with your suggestion. It's a 1-in-10^6 flake rather than a coupling issue, but it's a real one and the ternary costs nothing.

tests/e2e/Services/Project/TemplatesBase.php
```diff
@@ -43,6 +43,7 @@ public function testGetEmailTemplateAllSupportedTypes(): void
             'mfaChallenge',
             'sessionAlert',
             'otpSession',
+            'otpVerification',
```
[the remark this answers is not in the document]
Meldiron: Not acting on this one, deliberately.

Those two arrays aren't a duplicate of the allowlist — they're the assertion. `testGetEmailTemplateAllSupportedTypes` walks the list and calls `GET /v1/project/templates/email/:templateId/:locale` for each, which is what catches a template registered in `app/config/locale/templates.php` but missing from the `getDefaultMessage()` map in either `Get.php`. That mismatch returns a 200 with the wrong (button) template because of the `?? ['file' => 'email-inner-base.tpl', ...]` fallback, so nothing else in the suite would notice.

Deriving the list from `Config::getParam('locale-templates')` instead would make the test pass unconditionally for any newly added type — precisely the failure mode you're describing, just moved. The "synchronized source and test edits" is the point: the edit is what proves someone looked at the rendered output.

I asked for `'otpVerification'` to be added here in the first place, for that reason.

PR #13678, opened 2026-09-14
perf(storage): share one pooled HTTP client across S3-family devices

loks0n:
## Problem

`getDevice()` builds a new device per request (and per project in the workers), and `utopia-php/storage`'s default S3 client sets `CURLOPT_FORBID_REUSE`, so every object call pays a fresh TCP and TLS handshake. `GET /v1/health/storage` does write, read and delete against four devices, so a single health call is twelve handshakes.

Cloud CI has been failing `Health\StorageTest::testStorage` (`ping < 100`) on main several times a day with 102 to 113 ms pings; see appwrite-labs/cloud#5830.

## Change

Hold one lazily built, process-wide pool of connection-reusing curl clients in `getDevice()` and pass it to every S3-family device (S3, AWS, DO Spaces, Backblaze, Linode, Wasabi). Curl options and the `S3\RetryStrategy` match the library default, so the only behavioural change is that handles keep their connection cache between requests.

Same shape as the existing `usageConnection` pool in this file. Pool: 64 handles, 30 s wait for a free one. Every entry point (HTTP, worker, CLI) already runs under Swoole coroutines, which the pool adapter requires.

## Testing

No unit test: the subject is a global function that builds a Swoole pool, and any double would only echo the wiring back. The red is the existing E2E: `tests/e2e/Services/Health/StorageTest.php` fails against MinIO in Kind at ~8 ms per handshake; with reuse the same twelve calls should sit well under the ceiling. CI on this PR runs it.


loks0n:
Closing for now; a better pool-based solution will follow.

PR #13689, opened 2026-09-15
fix(console): link the new console everywhere and drop the legacy /console scheme

HarshMN2345:
## What

- New `_APP_CONSOLE_URL`: the full origin of the console. Every console link Appwrite produces is built from it. It defaults to `{http|https}://{_APP_CONSOLE_DOMAIN or _APP_DOMAIN}`, using the same `_APP_OPTIONS_FORCE_HTTPS` rule as before. On Cloud it is `https://appwrite.io`. API URLs such as the VCS OAuth `redirect_uri`, `/images/vcs` and QR codes still use the console hostname.
- `_APP_CONSOLE_URL_SCHEME` is removed, along with every legacy `/console/project-{region}-{id}/...` shape. All links now use root routes: OAuth2 success and failure defaults, magic URL, build CTAs, VCS comments and commit statuses, authorize-contributor, and webhook failure emails.
- VCS installation callbacks with no redirect now default to `{consoleUrl}/projects/{projectId}/settings` (GitHub, Gitea, GitLab, Bitbucket and Origin). This supersedes #13599.
- The legacy web paths redirect to the new console again: `/invite` to `/join`, `/login` to `/sign-in`, `/recover` to `/reset` and `/register` to `/sign-up`. `/`, `/auth/*` and `/mfa` keep their paths. `/card` is removed because the new console has no matching route. When the request is already on the console host, these paths return 404 instead of redirecting to themselves. The console host is also added to the allowed redirect hostnames.
- Updated the tests: HTTP redirects and default OAuth2, the account OAuth2 failure URL, the webhooks worker, and the Gitea VCS callbacks.

## Why

The new console serves only root routes, and on Cloud it runs on a different host from the API. Links built from the API hostname with `/console` paths sent users to pages that do not exist.


app/config/variables.php
```diff
@@ -242,10 +242,10 @@
                 'filter' => ''
             ],
             [
-                'name' => '_APP_CONSOLE_URL_SCHEME',
-                'description' => 'Console URL scheme used when the backend generates links to the console (OAuth callbacks, emails, error page CTAs, VCS comments). Set to \'root\' for the new console served at the root path (appwrite/new), or \'legacy\' for the older console served under the /console path prefix. The default value is \'legacy\'.',
-                'introduction' => '2.0.0',
-                'default' => 'legacy',
+                'name' => '_APP_CONSOLE_URL',
```
Meldiron, line 245: Are we 100% certain we need new variable? Can we not reuse some existing one? if so, why not?

app/config/variables.php
```diff
@@ -242,10 +242,10 @@
                 'filter' => ''
             ],
             [
-                'name' => '_APP_CONSOLE_URL_SCHEME',
```
Meldiron, line 245: Lets check who added this variable, and confirm with them that its OK to remove it with this PR

app/controllers/api/account.php
Meldiron, line 1: Throughout entire PR, we touched MANY flows. Lets QA them manually, to make sure there is no regression, nothing breaks.
If we are not sure what flow a diff is and how to reproduce it, that would be big warning flag. If we dont 100% understand where some change applies, that might be the exact diff we will want to focus our manual QA.

Meldiron (approved):

PR #13698, opened 2026-09-15
feat(account): store the ID token on the identity

Meldiron:
## What does this PR do?

Native sign-in (`POST /account/sessions/id-token`) verifies a full OIDC ID token and then throws it away, keeping only the handful of claims the identity models (`sub`, `email`, `name`, `picture`). A client that wants anything else has no way to read it back short of another round trip to the provider.

Google's `locale` is the motivating case — useful for picking a UI language on first sign-in, but not worth a first-class attribute of its own. The same goes for `hd`, `given_name`/`family_name`, Apple's `is_private_email`, and whatever a provider adds next.

So keep the raw token on the identity and hand it back on the `Identity` model, letting clients decode whatever they need.

### Changes

- **Schema** — `providerIdToken` on `identities`, encrypted at rest with the same `encrypt` filter as the `providerAccessToken` and `providerRefreshToken` beside it. Sized 16384 to match them; the `idToken` request param caps at 8192 characters, so there is room.
- **Migration** — `V25` creates the attribute alongside `photo`, via the same `createAttributeFromCollection` path and the same warn-on-failure handling.
- **Write path** — set on identity creation, and refreshed on every subsequent sign-in. The token is verified and required, so it always supersedes the stored one: it expires within the hour and its claims may have changed. This makes the `!empty($changes)` guard on the update permanently true, so it goes away.
- **Response model** — `providerIdToken` on `Identity`.

Only the native flow stores one. The browser OAuth2 flow never receives an ID token, so identities it creates leave the attribute empty — which is why the model documents it as such.

## Test Plan

New e2e test `testCreateIdTokenSessionStoresIdToken`, against the mock OAuth2 provider. It signs in with a Google-shaped token carrying `locale`, then reads the identity back through `GET /account/identities` and asserts the token is:

1. **stored** — the identity carries a `providerIdToken`;
2. **the right value** — equal to the minted token byte for byte, so the signature still verifies downstream (this also proves the round trip through the `encrypt` filter is lossless);
3. **decodable** — a three-part JWT whose payload is valid base64url JSON;
4. **carrying the right claims** — `sub`, `email` and `name` match, and `locale` is `en`: the claim Appwrite deliberately does not model, now readable by the client.

It then signs in again with a token minted with `locale` set to `cs` and asserts the stored token was replaced and decodes to the new value, covering the refresh path as well as the create path.

```
$ docker exec appwrite vendor/bin/phpunit --filter 'IdToken' \
    tests/e2e/Services/Account/AccountCustomClientTest.php
OK (18 tests, 240 assertions)

$ docker exec appwrite vendor/bin/phpunit --filter 'OAuth2' \
    tests/e2e/Services/Account/AccountCustomClientTest.php
OK (7 tests, 186 assertions)
```

The browser OAuth2 run confirms identities created without an ID token are unaffected. `composer lint`, `composer refactor:check` and PHPStan on the changed files all pass.

## Related PRs and Issues

- Builds on the native ID token sign-in flow in `Account/Http/Account/Sessions/IdToken`

## Checklist

- [x] If the PR includes a change to an API's metadata (desc, label, params, etc.), does it also include updated API specs and example docs? — the `Identity` model gained a field, but `app/config/specs/` is gitignored and no longer tracked (`8d82bc89a1`), so specs are regenerated and published separately.


Meldiron (commented):
lgtm

PR #13701, opened 2026-09-15
Add project mcpLastAccessAt

eldadfux:
## Summary
- Expose `mcpLastAccessAt` on the Project model and console projects collection.
- Add a V25 migration so existing installs get the attribute.

## Test plan
- [ ] GET a project includes `mcpLastAccessAt`


eldadfux:
Dropping this. Console will use MCP install/OAuth only; cloud already tracks mcpAccessedAt.

PR #13715, opened 2026-09-15
refactor: move Bus to packages/bus

loks0n:
## Summary

Second package on the plan in `rfc/monorepo.md`. `Bus` was the last occupant of `src/Utopia/`: autoloaded from the root, never published. It now lives at `packages/bus` in the standard shape:

- manifest declaring its one dependency, `utopia-php/span` (it was implicit before, supplied by the root)
- `Utopia\Bus\ => src/`, `Utopia\Bus\Tests\ => tests/`, a `phpstan.neon`, a unit test for `dispatch` (every subscribed listener called with resolved injections; dispatch without a resolver refused)
- mirror plumbing for the new, empty `utopia-php/bus` repository: `mirror.yml`, README banner; the mirror ruleset is already applied via `bin/monorepo absorb bus`

`src/Utopia/` is gone; `AGENTS.md` no longer describes it. Root autoload regenerated with `bin/monorepo autoload`. No history import: Bus never had a repository of its own, so `split` synthesises its history from this repository's root, exactly the case `docs/distribution.md` describes for a package written in the monorepo.

## Also

Package `phpunit.xml` files now bootstrap `vendor/autoload.php`, as the monorepo's packages do. Without it PHPUnit fell back to the root autoloader, so a registry-mode `bin/monorepo test` never proved the package installs and tests standalone. Caught because the first `test bus` run failed on exactly that.

## After merge

The first `Split` run pushes `packages/bus` to the empty mirror and creates its `main`. Then the mirror can be submitted to Packagist and a first release cut with `bin/monorepo release bus 0.1.0`.

## Verification

- `bin/monorepo validate`: all packages valid
- `bin/monorepo check bus`: clean; `bin/monorepo test bus` and `test agents` green on a fresh package vendor
- `vendor/bin/phpunit --testsuite packages`: 136 tests, 4 skipped, green
- Root PHPStan clean on all 50 Appwrite files that consume Bus; `tests/unit/Bus` green


loks0n:
Post-merge check: `Split` run 35033297044 on the merge commit succeeded. `utopia-php/bus` went from an empty repository to `main` at `8c816d4001`, a single parentless commit whose tree is `packages/bus` at the merge, exactly what `bin/monorepo split bus --dry-run` predicted. The same run advanced `utopia-php/agents` to `7032aafac5` for the `phpunit.xml` bootstrap change.

Remaining, both manual: submit https://github.com/utopia-php/bus to Packagist, then `bin/monorepo release bus 0.1.0`.

PR #13737, opened 2026-09-17
fix: clear deployment references after deletion and retention

HarshMN2345:
## What does this PR do?

When deployment retention expires the latest inactive deployment, `latestDeploymentId` keeps pointing at the deleted record while an older active deployment keeps serving traffic, and the Console fails to load the function or site. The queued deployment deletion now repoints `latestDeployment*` to the newest surviving deployment, or clears it when none remain.

The Console deletes selected deployments in parallel, so a repoint could pick a deployment that another request was deleting. The delete endpoints now remove the deployment and repoint the function or site under a per-resource `locks` entry, the same lock resource Deployments Create and Status Update use, and the queued repair takes the same key. Contention returns the same busy error as deployment uploads.

Manual deletion now updates the parent before removing the source file, writing only the changed attributes. When that file was already gone, the endpoint returned 500 after deleting the deployment but before repointing `latestDeploymentId`. Deployment timestamps are cleared with `null`; encoding an empty datetime string stored the current time instead.

The Deletes worker also passed the deployment to `deleteRuntimes()`, which expects a function. That removed the runtimes of whichever function had the same sequence as the deleted deployment. It now deletes the runtime by the deployment ID.

Screenshot jobs now update `deploymentScreenshot*` only while their deployment is the site's active one, so a preview build no longer replaces the active deployment's screenshots. The write uses the site's read timestamp, so it conflicts and re-checks when the site was updated after the check.

## Test Plan

Before and after this change, on the local stack, for both functions and sites:

| Flow | Before | After |
| --- | --- | --- |
| Manual: delete the latest preview while another deployment is active | OK | OK |
| Manual: delete the active deployment while a preview survives | `deploymentCreatedAt` set to now | OK |
| Manual: delete the final deployment | `deploymentCreatedAt` and `latestDeploymentCreatedAt` set to now | OK |
| Parallel deletion of six previews, as the Console's multi-select does | pointer left on a deleted deployment in 1 of 12 runs | 0 of 16 runs, immediately and after 5s |
| Retention expires the latest preview | `latestDeploymentId` points at the deleted deployment | OK |
| Manual: delete a preview whose source file is already missing | 500 and `latestDeploymentId` points at the deleted deployment | 500, pointer repointed |

The existing E2E cases now cover these: retention expires the latest preview, deleting the final deployment clears references and timestamps, and a preview's screenshots leave the site's active screenshots unchanged (fails against the previous screenshot worker).

```sh
docker compose exec -T appwrite test tests/e2e/Services/Functions --filter='testUpdateDeploymentStatus|testFunctionDeploymentRetentionWithMaintenance'
docker compose exec -T appwrite test tests/e2e/Services/Sites --filter='testUpdateDeploymentStatus|testSiteDeploymentRetentionWithMaintenance'
docker compose exec -T appwrite test tests/e2e/Services/Sites/SitesConsoleClientTest.php --group=screenshots --filter=testSiteScreenshot
```

## Related PRs and Issues

Deployments are only deleted by the manual endpoints and by maintenance retention; the interval task only verifies domains.

Cloud's Deletes worker calls `parent::action()` with positional arguments, so its server-ce bump must inject `locks` and pass it through.

## Checklist

- [x] Read the contributing guidelines.
- [x] No API metadata changes; updated specs and example docs are not required.


src/Appwrite/Platform/Modules/Functions/Workers/Screenshots.php
```diff
@@ -219,10 +219,23 @@ public function action(
...
+                $dbForProject->updateDocument('sites', $site->getId(), new Document([
+                    'deploymentScreenshotDark' => $deployment->getAttribute('screenshotDark', ''),
+                    'deploymentScreenshotLight' => $deployment->getAttribute('screenshotLight', ''),
+                ]));
+                return true;
+            });
+
+            if ($updated) {
+                $dbForProject->purgeCachedDocument('sites', $siteId);
+            }
```
Meldiron: I dont see benefit of transaction here, what am I missing?

ALso, why purge? Doesnt DB lib purge on its own?

PR #13761, opened 2026-09-18
refactor(messaging): build the platform's SMS adapter as a DI resource

ChiragAgg5k:
## Why

The messaging worker built its own SMS adapter and remembered it in a nullable field:

```php
private ?SMSAdapter $adapter = null;
...
if ($this->adapter === null) {
    $this->adapter = $this->createInternalSMSAdapter();
}
```

That is a hand-rolled version of what the container already does. The adapter is an object dependency built from platform environment variables, which is what `app/init/resources.php` is for, and the container caches resolved instances for the life of the process anyway.

This is groundwork. It lands on its own so that the WhatsApp OTP channel in #13711 can add a second adapter without also carrying this move.

## What

**`adapterForSMS`** is now a container resource. The worker injects it like any other dependency, and the memo field and its lazy getter are gone.

**`Appwrite\Messaging\Provider`** owns the mapping from a provider to the adapter that sends it. A provider is either one a project configured and stored as a document, or the platform's own, described by `_APP_SMS_PROVIDER` and used for the passcodes and invites Appwrite sends on a project's behalf. Both end up as the same adapter, so the DSN is turned into a provider document and takes the same path. That is why `internalSMS()` and the document mapping belong together.

The worker loses four methods and roughly 240 lines.

**Push stays in the worker.** Its Appwrite provider is built from the message being sent and the project database it belongs to, so a provider document alone does not describe it. Email and SMS need nothing but the document, so they moved.

**A malformed platform DSN is now reported and read as unset rather than thrown.** The resource is resolved for every messaging job, so letting one bad variable escape would stop a project's push and email messages, which have nothing to do with it. The internal SMS path still fails exactly as before, because a null adapter there raises the same error it always did.

## No behaviour change

That is the claim to review. Every moved method is unchanged apart from its name and visibility.

| Before | After |
|---|---|
| `getSmsAdapter()` | `Provider::sms()` |
| `getEmailAdapter()` | `Provider::email()` |
| `createInternalSMSAdapter()` | `Provider::internalSMS()` |
| `createProviderFromDSN()` | `Provider::fromDSN()` |

## Tests

No new tests, because there is no new behaviour. The existing suites cover both paths this touches: `tests/e2e/Services/Messaging` exercises project providers across SMS, email and push, and `tests/e2e/Services/Account` exercises the platform SMS adapter through phone tokens and verification.

Locally, against a rebuilt worker:

- Messaging: 129 tests pass, 24 skipped.
- Account: the same two failures as a clean `main` checkout, in `testSessionAlert` and `testCreateIdTokenSessionUnsupportedProvider`. I confirmed both by stashing this change and re-running, so neither is from this branch.
- Pint, Rector and PHPStan are clean on the three files.


ChiragAgg5k:
@greptile review

loks0n (approved):

PR #13777, opened 2026-09-20
ci: move every job onto gcrunner presets and drop RunsOn

ChiragAgg5k:
Moves the build and test lanes off RunsOn and onto gcrunner, following [appwrite-labs/cloud#5947](https://github.com/appwrite-labs/cloud/pull/5947). Every gcrunner label now names a preset from `.github/gcrunner.yml` instead of spelling out a machine spec, and no label pins an image, so the image workflow pruning a build cannot take CI down again.

## What changed

- `.github/gcrunner.yml` defines six presets: `small`, `standard`, `build`, `publish`, `e2e` and `e2e-large`. Databases and TablesDB keep the larger shape they had on RunsOn; everything else maps to the size it already used.
- Every `runs-on` line in `ci.yml`, `specs.yml` and `sdk-preview.yml` is `gcrunner=…/runner=<preset>`. The check and packaging jobs keep their `ubuntu-latest` fallback for fork pull requests.
- The build pushes to `${GCRUNNER_REGISTRY}/appwrite` with a `type=registry` layer cache. The test lanes pull through `${GCRUNNER_REGISTRY_PULL}/appwrite`, the read-through cache in the VM's own region. The RunsOn ECR and S3 variables are gone.
- The runner injects the registry variables into the shell only, so the build promotes `GCRUNNER_REGISTRY` to `GITHUB_ENV` before `build-push-action` reads it as an action input.

This repository never had a `.github/runs-on.yml`; the RunsOn labels carried their spec inline, so there is nothing to delete.

## To verify on the first run

Presets need the gcrunner GitHub App to hold `contents: read` on this repository. Without it the orchestrator cannot read `.github/gcrunner.yml` and leaves the job queued rather than failing it. The build and test lanes run on gcrunner unconditionally, as they did on RunsOn, so a fork pull request will only work if the orchestrator serves fork jobs.


ChiragAgg5k:
Superseded by a two-step split, because this cannot merge as one commit.

The orchestrator reads `.github/gcrunner.yml` from the **default branch** on a public repository, never from the pull request head:

```go
func configRef(event WorkflowJobEvent) string {
	if event.Repository.Private {
		return event.WorkflowJob.HeadSHA
	}
	return event.Repository.DefaultBranch
}
```

So this pull request adds the file and references it in the same commit, and the presets do not exist on `main` yet. Every job is held:

```
Job 106068713448: configuration error: runner "standard" is not defined in .github/gcrunner.yml, leaving it queued
```

Held rather than failed, which is why the run sat queued for over an hour and looked like slow CI rather than a misconfiguration.

Split into:

1. #13778 — the presets file alone, inert until something names it
2. #13779 — the label conversion, which resolves once #13778 is on `main`

Merge in that order. Closing this one.

ChiragAgg5k:
Superseded by #13778 (presets) and #13779 (labels), which are split so the presets reach the default branch first — see the earlier comment for why this could not merge as one commit.

